PR #12, opened 2015-07-28
Skybridge config added

jgfouca:
Still doesn't work because the account directive is not making it
to the slurm submit script.

[BFB]


jgfouca:
Intended to merge branch into ACME's fork.

PR #16, opened 2015-09-18
Bluewaters pgi upgrade

bertinia:
update to config_machines.xml to set pgi version to 14.10.0 for bluewaters to fix
a known compiler bug error. This fixes cime issue #170.


jedwards4b:
Sorry - wrong repo

PR #18, opened 2016-01-07
Bring ACME changes into ESMCI cime

jgfouca:


components/data_comps/datm/bld/build-namelist
```diff
@@ -381,9 +381,7 @@ if ( $GRID eq "CLM_USRDAT" && $CLM_USRDAT_NAME =~ /^(UNSET|)$/ ) {
 EOF
 }
 
-(-d $DIN_LOC_ROOT)  or  die <<"EOF";
-** $ProgName - CESM inputdata root is not a directory: \"$DIN_LOC_ROOT\" **
-EOF
+(-d $DIN_LOC_ROOT)  or mkdir $DIN_LOC_ROOT;
```
jedwards4b: I do not believe that this is an appropriate place to create the DIN_LOC_ROOT directory.   It should have already been created when this script was called and erroring out is the appropriate behavior here.

douglasjacobsen: @jedwards4b The problem is that some machines require us to put DIN_LOC_ROOT in a scratch directory, in which case it gets purged every once in a while. So, DIN_LOC_ROOT has not always been created before running this script.

Aside from that, there's no reason this script shouldn't create DIN_LOC_ROOT if it doesn't exist. The configuration files told it that this was where it should find / store input data, and erroring out (for normal users) in this case is more an annoyance than anything helpful.

driver_cpl/cimetest/testlist_drv.xml
```diff
@@ -17,6 +17,8 @@
         <machine compiler="pgi" testtype="prebeta">bluewaters</machine>
```
jedwards4b: This directory no longer exists - it should be driver_cpl/cime_config/testdefs

driver_cpl/driver/cesm_comp_mod.F90
```diff
@@ -769,12 +782,15 @@ subroutine cesm_pre_init2()
...
+   call t_adj_detailf(+1)
 
+   call t_startf('CPL:cesm_pre_init2')
    !----------------------------------------------------------
    !| Memory test
    !----------------------------------------------------------
 
-   call shr_mem_init(prt=.true.)
+!mt   call shr_mem_init(prt=.true.)
+   call shr_mem_init(prt=iamroot_CPLID)
```
jedwards4b: Remove the comment line.

driver_cpl/driver/cesm_comp_mod.F90
```diff
@@ -1056,6 +1077,9 @@ subroutine cesm_init()
    !  components will set them to true for the purposes of symmetry
    !-----------------------------------------------------------------------------
 
+   call t_startf('cesm_init')
```
jedwards4b: Timers now include the prefix related to they're model, please add back the 'CPL:' string to timers in the coupler

driver_cpl/driver/cesm_comp_mod.F90
```diff
@@ -1072,25 +1095,95 @@ subroutine cesm_init()
    call component_init_pre(ice, ICEID, CPLICEID, CPLALLICEID, infodata, ntype='ice')
    call component_init_pre(glc, GLCID, CPLGLCID, CPLALLGLCID, infodata, ntype='glc')
    call component_init_pre(wav, WAVID, CPLWAVID, CPLALLWAVID, infodata, ntype='wav')
+   call t_stopf('comp_init_pre_all')
 
 #ifdef ESMF_INTERFACE
+   call t_startf('comp_init_cc_atm')
```
jedwards4b: these timer start and stops should be inside the component_init_cc subroutine.   No need to bring them out.

driver_cpl/driver/cesm_comp_mod.F90
```diff
@@ -1760,6 +1876,9 @@ subroutine cesm_init()
    !----------------------------------------------------------
 
    if (atm_present) then
+      call t_startf('comp_init_cc_atm2')
```
jedwards4b: These should be inside the component_init_cc subroutine

rljacob:
Noting these links here for everyone.
Summary of changes from CIME2 to CIME4 (not complete):
https://docs.google.com/document/d/1nrh6r6m8IPI-xc6YaYa32kA7Smi62Uc4RJVWsD2uhOw/edit?usp=sharing 

Summary of changes from CIME2 to ACME-CIME2 (just started):
https://docs.google.com/document/d/1OKobMxOsPO1YfuCXsHKgv4PcRlOIWC9KCeXBL3FeTd8/edit?usp=sharing

components/data_comps/datm/bld/build-namelist
jedwards4b: I'm not disagreeing that the DIN_LOC_ROOT may need to be created for each
case, I am only suggesting that it should be done in one place rather than
in each of the
individual component build-namelist scripts.

On Thu, Jan 7, 2016 at 12:33 PM, Doug Jacobsen <notifications@github.com>
wrote:

> In components/data_comps/datm/bld/build-namelist
> <https://github.com/ESMCI/cime/pull/18#discussion_r49115108>:
>
> > @@ -381,9 +381,7 @@ if ( $GRID eq "CLM_USRDAT" && $CLM_USRDAT_NAME =~ /^(UNSET|)$/ ) {
> >  EOF
> >  }
> >
> > -(-d $DIN_LOC_ROOT)  or  die <<"EOF";
> > -** $ProgName - CESM inputdata root is not a directory: \"$DIN_LOC_ROOT\" **
> > -EOF
> > +(-d $DIN_LOC_ROOT)  or mkdir $DIN_LOC_ROOT;
>
> @jedwards4b <https://github.com/jedwards4b> The problem is that some
> machines require us to put DIN_LOC_ROOT in a scratch directory, in which
> case it gets purged every once in a while. So, DIN_LOC_ROOT has not always
> been created before running this script.
>
> Aside from that, there's no reason this script shouldn't create
> DIN_LOC_ROOT if it doesn't exist. The configuration files told it that this
> was where it should find / store input data, and erroring out (for normal
> users) in this case is more an annoyance than anything helpful.
>
> —
> <https://github.com/ESMCI/cime/pull/18/files#r49115108>.
>



-- 
Jim Edwards

CESM Software Engineer
Boulder, CO

douglasjacobsen: So, for example, it should be done in the setup script, or in the build script, rather than in build-namelist, correct?

I'd be fine with either of those, in which case you're right this should error, since build-namelist called by things that should have created the directory.

jedwards4b: Yes, that's it.

On Thu, Jan 7, 2016 at 12:57 PM, Doug Jacobsen <notifications@github.com>
wrote:

> In components/data_comps/datm/bld/build-namelist
> <https://github.com/ESMCI/cime/pull/18#discussion_r49118186>:
>
> > @@ -381,9 +381,7 @@ if ( $GRID eq "CLM_USRDAT" && $CLM_USRDAT_NAME =~ /^(UNSET|)$/ ) {
> >  EOF
> >  }
> >
> > -(-d $DIN_LOC_ROOT)  or  die <<"EOF";
> > -** $ProgName - CESM inputdata root is not a directory: \"$DIN_LOC_ROOT\" **
> > -EOF
> > +(-d $DIN_LOC_ROOT)  or mkdir $DIN_LOC_ROOT;
>
> So, for example, it should be done in the setup script, or in the build
> script, rather than in build-namelist, correct?
>
> I'd be fine with either of those, in which case you're right this should
> error, since build-namelist called by things that should have created the
> directory.
>
> —
> <https://github.com/ESMCI/cime/pull/18/files#r49118186>.
>



-- 
Jim Edwards

CESM Software Engineer
Boulder, CO

PR #33, opened 2016-01-14
Merge changes made on the ACME side to config_definitions.xml

jgfouca:
In the shared CIME, this file is now called config_component.xml

These changes should probably be reviewed.

### DIFF
--- a/driver_cpl/cime_config/config_component.xml
+++ b/driver_cpl/cime_config/config_component.xml
@@ -409,7 +409,7 @@
   <entry id="DOUT_S"> 
     <type>logical</type>
     <valid_values>TRUE,FALSE</valid_values>
-    <default_value>TRUE</default_value>
+    <default_value>FALSE</default_value>
     <group>run_data_archive</group>
     <file>env_run.xml</file>
     <desc>Logical to turn on short term archiving.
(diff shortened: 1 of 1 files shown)


driver_cpl/cime_config/config_component.xml
```diff
@@ -409,7 +409,7 @@
   <entry id="DOUT_S"> 
     <type>logical</type>
     <valid_values>TRUE,FALSE</valid_values>
-    <default_value>TRUE</default_value>
+    <default_value>FALSE</default_value>
```
jedwards4b: Can we leave this default set to TRUE?   We set it to FALSE for testing but generally try to have the default configuration

jgfouca: How does your testing system turn it off?

jedwards4b:
I'm a little concerned about how we are going to resolve differences in the pio namelists
but I don't think that needs to hold up this merge.

driver_cpl/cime_config/config_component.xml
mvertens: This is turned off in the csh test. As an example - look at the ERI_script
and ERR_script you'll see calls like

./xmlchange -file env_run.xml -id DOUT_S        -val FALSE

agsalin: Currently, we prefer it FALSE. Mark turned DOUT_S to false a few months ago, and seems to prefer it. His reasoning is because he prefers the organization of the log files without this step. I like it because I can run case.run, instead of case.submit (I hope I have it correct). No one has complained that I know of.

In the medium term, we will probably need to have a way of different defaults, since this is likely not the last case.

In the short term, you can pick one, and one team has to deal temporarily, but the next coin-flip goes the other way.

jedwards4b: I think that we need a way to differentiate between models in the
config_component.xml file - perhaps an additional field
<model></model>    would do the trick.   This in the case where we want
different defaults we can add a model attribute to the
<default_value> field.      This would resolve the input arguments but we
still need to figure out how to handle the differences in shr_pio_mod.

On Thu, Jan 14, 2016 at 5:06 PM, Andy Salinger <notifications@github.com>
wrote:

> In driver_cpl/cime_config/config_component.xml
> <https://github.com/ESMCI/cime/pull/33#discussion_r49805378>:
>
> > @@ -409,7 +409,7 @@
> >    <entry id="DOUT_S">
> >      <type>logical</type>
> >      <valid_values>TRUE,FALSE</valid_values>
> > -    <default_value>TRUE</default_value>
> > +    <default_value>FALSE</default_value>
>
> Currently, we prefer it FALSE. Mark turned DOUT_S to false a few months
> ago, and seems to prefer it. His reasoning is because he prefers the
> organization of the log files without this step. I like it because I can
> run case.run, instead of case.submit (I hope I have it correct). No one has
> complained that I know of.
>
> In the medium term, we will probably need to have a way of different
> defaults, since this is likely not the last case.
>
> In the short term, you can pick one, and one team has to deal temporarily,
> but the next coin-flip goes the other way.
>
> —
> <https://github.com/ESMCI/cime/pull/33/files#r49805378>.
>



-- 
Jim Edwards

CESM Software Engineer
Boulder, CO

PR #34, opened 2016-01-14
Resolved case.build conflicts

mvertens:
Introduced several specific acme project ifs to temporarily get this
also working in cesm. A more general approach will need to be implemented
after the initial merge is complete.

Also changed the cesm config_machines.xml to be able to work with the acme
changes. There are still problems with the cesm build for csm_share at this point.

### DIFF
--- a/cime_config/cesm/machines/config_machines.xml
+++ b/cime_config/cesm/machines/config_machines.xml
@@ -1758,6 +1758,9 @@
     </environment_variables>
   </machine>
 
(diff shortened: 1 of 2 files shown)


jgfouca:
@mvertens , should I be worried about your MV NEED HELP tag?

mvertens:
Not sure - this was the question I had for Rob on slack yesterday - he said to just comment it out for now and you guys would deal with it.

jgfouca:
The builds made it much farther, so I think you got this script correct. Merged.

PR #35, opened 2016-01-15
Cseg master

jedwards4b:
Removed pio and added pio1 and pio2.   Changed build structure so that one can toggle a case between the two libraries.  Bring this change the the CESM-Development trunk, then pull that to 
esmci/cseg_master branch.


jedwards4b:
This pull request is superseded by #37.

PR #38, opened 2016-01-20
Deacme py

jedwards4b:
This PR will allow the python create_test to work with cesm.  It removes 'acme' as a hardcoded path and uses a new variable model which reads an environment variable CIME_MODEL and defaults to acme if the variable doesn't exist.

### DIFF
--- a/utils/python/cime_util.py
+++ b/utils/python/cime_util.py
@@ -4,8 +4,8 @@ Common functions used by cime python scripts
 
 import sys, socket, re, os, time
 
+_MODEL = None
 _VERBOSE = False
-_MODEL = "cesm"
 _MACHINE_INFO = None
 
 # batch-system-name -> ( cmd-to-list-all-jobs-for-user, cmd-to-delete-job )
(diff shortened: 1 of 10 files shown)


utils/python/cime_util.py
```diff
@@ -493,16 +494,27 @@ def delete_jobs(jobs):
...
 def parse_config_machines():
 ###############################################################################
     """
     """
     global _MACHINE_INFO
+    model = get_model()
     if (_MACHINE_INFO is None):
         _MACHINE_INFO = {}
         import xml.etree.ElementTree as ET
-        config_machines_xml = os.path.join(get_cime_root(), "cime_config", "acme", "machines", "config_machines.xml")
```
jgfouca: I encapsulated the retrieval of the config directory for a model but forgot to use it here, I'll fix that.

utils/python/create_test_impl.py
```diff
@@ -234,12 +234,12 @@ def _create_newcase_phase(self, test_name):
...
     def _xml_phase(self, test_name):
     ###########################################################################
-        test_case, _, _, _, machine, _, _ = acme_util.parse_test_name(test_name)
+        test_case, _, _, _, machine, _, _ = cime_util.parse_test_name(test_name)
 
         xml_file = os.path.join(self._get_test_dir(test_name), "env_test.xml")
-        xml_bridge_cmd = os.path.join(acme_util.get_acme_scripts_root(), "xml_bridge")
+        xml_bridge_cmd = os.path.join(cime_util.get_acme_scripts_root(), "xml_bridge")
 
-        mach_dir = os.path.join(self._cime_root, "acme", "machines-acme")
```
jgfouca: Hmmm, I'm wondering how this worked at all. These paths are wrong. I will fix.

jedwards4b:
Redone as a branch in #40

PR #44, opened 2016-01-22
Pio1 update

jedwards4b:
This brings in the acme changes for pio1 and moves around directories to allow more code to be shared between projects.  I have run the cime_developer tests using both pio1 and pio2 successfully.  There are still some files in the acme/machines directory which are no longer used and can be deleted.


rljacob:
Could you possibly remove the cime_config/models dir and move acme and cesm back up?  I don't think we'll ever have so many models coupled with cime that we'll need to move them to a subdirectory.

jedwards4b:
I've decided to back out the directory structure changes and just do the pio1_update.
Closing this, will resubmit as a new PR

PR #48, opened 2016-01-26
Oopy

jedwards4b:
This brings the OO design into the python scripts for testing.   
cime_developer tests run on yellowstone 

A goal yet to be achieved is to move all of the existing code in utils/python directory into
the utils/python/CIME directory tree in a fully integrated manor.


scripts-python/create_test
```diff
@@ -18,7 +18,7 @@ cime_util.check_minimum_python_version(2, 7)
 import argparse, sys, os, doctest, tempfile
 import update_acme_tests, create_test_impl
 from cime_util import expect, warning, verbose_print, run_cmd
-
+from CIME.Tests import Tests
 ###############################################################################
```
jgfouca: Please keep newline separating imports from code.

scripts-python/create_test
```diff
@@ -252,6 +252,8 @@ def _main_func(description):
         return
 
     cime_util.stop_buffering_output()
+    tests  = Tests()
```
jgfouca: Shouldn't this only be done if --test was provided?

jedwards4b: Yes, I didn't intend to add that yet.   I will remove.

utils/python/CIME/utils.py
```diff
@@ -0,0 +1,65 @@
...
+        ...
+    SystemExit: FAIL: error2
+    """
+    if (not condition):
+        #traceback()
+        raise SystemExit("FAIL: %s" % error_msg)
+    
+
+def get_cime_root():
+###############################################################################
```
jgfouca: Get rid of these hash marks

utils/python/CIME/utils.py
```diff
@@ -0,0 +1,65 @@
+"""
+Common functions used by cime python scripts
+"""
+import logging
+import sys
+import os
```
jgfouca: How did you decide which things to move from cime_util to utils?

utils/python/CIME/XML/Machines.py
```diff
@@ -0,0 +1,118 @@
...
+        supported_values = supported_values.split(',')
+        if(reqval is None or reqval == "UNSET"):
+            return supported_values[0]
+        for val in supported_values:
+            if(val == reqval):
+                return reqval
+        logging.critical("%s value %s not supported for machine %s" %
+                         (listname, reqval, self.name))
+
+    def get_compiler(self, compiler=None):
```
jgfouca: I'm a little confused here, why would a user do get_compiler instead of get_value("compiler")? Is it just a convenience thing because we expect them to be asking for this a lot?

utils/python/CIME/XML/Machines.py
```diff
@@ -0,0 +1,118 @@
...
+        supported_values = self.get_value(listname)
+        if(supported_values is None):
+            logging.critical("No list found for "+listname+" on machine "+self.name)
+        supported_values = supported_values.split(',')
+        if(reqval is None or reqval == "UNSET"):
+            return supported_values[0]
+        for val in supported_values:
+            if(val == reqval):
+                return reqval
+        logging.critical("%s value %s not supported for machine %s" %
```
jgfouca: What the difference between logging.critical and expect(False, ...)?

utils/python/cime_util.py
```diff
@@ -523,42 +470,17 @@ def delete_jobs(jobs):
 def parse_config_machines():
 ###############################################################################
     """
+    Moving toward an object oriented model, replace _MACHINE_INFO dict with an object of class Machines
     """
     global _MACHINE_INFO
     if (_MACHINE_INFO is None):
```
jgfouca: Glad to see these functions getting much smaller.

utils/python/cime_util.py
```diff
@@ -490,7 +436,8 @@ def find_proc_id(proc_name=None,
 ###############################################################################
 def get_batch_system(machine=None):
 ###############################################################################
-    return get_machine_info("BATCH_SYSTEM", machine=machine)
+    return _MACHINE_INFO.get_value("BATCH_SYSTEM")
+#    return get_machine_info("BATCH_SYSTEM", machine=machine)
```
jgfouca: I'd prefer deleting over commenting-out

utils/python/cime_util.py
```diff
@@ -584,58 +506,19 @@ def get_machine_info(items, machine=None, user=None, project=None, case=None, ra
...
-
-    if (isinstance(items, str)):
-        items = [items]
 
-    rv = []
-    if (raw):
-        for item in items:
-            rv.append(_MACHINE_INFO[machine][item])
+    expect(machine is not None, "Failed to probe machine. Please provide machine to whatever script you just ran")
+#    _MACHINE_INFO.set_machine(machine)
```
jgfouca: Delete commented out line.

utils/python/cime_util.py
```diff
@@ -677,7 +561,8 @@ def does_machine_have_batch(machine=None):
     >>> does_machine_have_batch("skybridge")
     True
     """
-    batch_system = get_machine_info("BATCH_SYSTEM", machine=machine)
+#    batch_system = get_machine_info("BATCH_SYSTEM", machine=machine)
```
jgfouca: Delete commented-out line,

utils/python/create_test_impl.py
```diff
@@ -29,7 +29,10 @@ def __init__(self, test_names,
                  compiler=None,
                  baseline_root=None, baseline_name=None,
                  clean=False,
-                 compare=False, generate=False, namelists_only=False,
+
```
jgfouca: Why add blank lines?

utils/python/create_test_impl.py
```diff
@@ -348,10 +351,11 @@ def _build_phase(self, test_name):
     def _run_phase(self, test_name):
     ###########################################################################
         test_dir = self._get_test_dir(test_name)
-        if (self._no_batch):
-            return self._run_phase_command(test_name, "./case.test", RUN_PHASE, from_dir=test_dir)
-        else:
-            return self._run_phase_command(test_name, "./case.submit", RUN_PHASE, from_dir=test_dir)
+# the submit script should do the right thing in either case.
```
jgfouca: Glad to see this get simpler. Please delete old code.

utils/python/tests/test.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python
+import sys, os
+cimeroot = os.getenv("CIMEROOT",\
+        os.path.dirname(os.path.abspath(os.path.join(__file__,"..",".."))))
```
jgfouca: Why not add this to scripts_regression_tests so it can be run with everything else?

utils/python/CIME/utils.py
jedwards4b: I moved the things that I am using in other parts of CIME.

jgfouca: Any reason not to just move all of it?

utils/python/CIME/XML/Machines.py
jedwards4b: COMPILER is not defined in the config_machines.xml, only COMPILERS same with MPILIBS so 
I created special subroutines to handle these two cases.   They include a second argument so that if you want to set the compiler on yellowstone to pgi you would call self.get_compiler('pgi') and it will tell you whether pgi is a valid compiler for that machine.

utils/python/CIME/utils.py
```diff
@@ -0,0 +1,65 @@
...
+    >>> os.path.isdir(os.path.join(get_cime_root(), get_acme_scripts_location_within_cime()))
+    True
+    """
+    global _CIMEROOT
+    if (_CIMEROOT is None):
+        try:
+            _CIMEROOT = os.environ["CIMEROOT"]
+        except KeyError:
+            script_absdir = os.path.abspath(os.path.join(os.path.dirname(__file__)))
+            assert acme_script_absdir.endswith(get_python_libs_location_within_cime()), acme_script_absdir
```
jgfouca: acme_script_absdir is undefined. I can worry about getting the ACME tests running if you don't have them running on hobart yet.

utils/python/CIME/utils.py
jedwards4b: Some of the other stuff would be more appropriate to move into the proposed Test class - I think that we want to keep the number of routines defined outside of classes as small as possible.

utils/python/CIME/XML/Machines.py
jgfouca: I see, OK. My only comment is get_compiler('X') is a somewhat unexpected way to ask if X is valid. I would make a different "is_valid" function.

utils/python/CIME/XML/Machines.py
jedwards4b: Nothing except that logging gives you a standard and more flexible interface.   I see that I've mixed and matched the two and we should consistently use one or the other.    Are you okay if I replace expect with logging throughout?

utils/python/CIME/utils.py
jgfouca: I can't comment on the Test class since I'm not sure what that is yet. I will, however, caution against carried away with the OO paradigm. Free functions are a very useful concept. Their simplicity makes them preferable to OO, in my opinion, unless there's a lot of persistent/shared state involved.

utils/python/CIME/XML/Machines.py
jgfouca: I'll tentatively say yes. Does it give you the same capability as exceptions?

jgfouca: Unless I'm misunderstanding, I don't think replacing expect with logging(critical) is what we want because logging(critical) allows the program to continue to run. Instead, I think the implementation of expect should log the critical error.

jedwards4b: so it should be logging.exception
https://docs.python.org/2/library/logging.html
my point is that logging gives you a flexible interface to support several levels of verbosity and we should use it rather than creating our own.

jgfouca: But the help for that function is this: This method should only be called from an exception handler. I think that's for logging an exception, not generating one.

jedwards4b: Okay so the expect function should stay and the logging.criticals should be changed to expect.

jgfouca: I understand we don't want to reinvent the wheel, but I've yet to see anything in the standard library that does these things:
* assert-like syntax
* generate user-specified or default exception type
* logging, as you want to add
So, in this case, I think a simple util function is OK.

jgfouca: > Okay so the expect function should stay and the logging.criticals should be changed to expect.
Yep.
One thing you can get rid of if you want is cime_util.warning

jedwards4b:
I've pushed an update for your review.  There is an issue building csm_share on the acme side that I am still trying to figure out.

cime_config/acme/allactive/testlist_allactive.xml
```diff
@@ -1,289 +1,5 @@
 <?xml version="1.0"?>
 <testlist>
-  <compset name="A">
```
jgfouca: Are you deleting this stuff because ACME uses update_acme_tests to manage tests?

scripts-python/create_test
```diff
@@ -14,10 +14,16 @@ If this tool is missing any feature that you need, please notify jgfouca@sandia.
...
 cime_util.check_minimum_python_version(2, 7)
-
+import logging
 import argparse, sys, os, doctest, tempfile
 import update_acme_tests, create_test_impl
-from cime_util import expect, warning, verbose_print, run_cmd
+from cime_util import run_cmd
+from CIME.utils import expect
+
+#    FORMAT = "%(level)s %(message)s"
```
jgfouca: These comments still needed?

scripts-python/create_test
```diff
@@ -250,8 +262,9 @@ def _main_func(description):
     if ("--test" in sys.argv):
         cime_util.run_cmd("python -m doctest %s/create_test_impl.py -v" % cime_util.get_python_libs_root(), arg_stdout=None, arg_stderr=None)
         return
-
+
```
jgfouca: Please remove trailing whitespace.

utils/python/CIME/XML/Machines.py
```diff
@@ -0,0 +1,142 @@
...
+        """
+        return self.get_field_from_list('COMPILERS')
+
+    def get_default_MPIlib(self):
+        """
+        Get the MPILIB to use from the list of MPILIBS
+        """
+        return self.get_field_from_list('MPILIBS')
+
+    def is_valid_compiler(self,compiler):
```
jgfouca: Thanks.

utils/python/cime_util.py
```diff
@@ -93,7 +58,7 @@ def run_cmd(cmd, ok_to_fail=False, input_str=None, from_dir=None, verbose=None,
     >>> run_cmd('ls file_i_hope_doesnt_exist')
     Traceback (most recent call last):
         ...
```
jgfouca: This may break stuff. At one point, the perl/csh was evaluating the success of a command by grepping the output for FAIL instead of checking return code. I added "FAIL" here instead of trying to track down and fix all the places where that was happening.

utils/python/cime_util.py
```diff
@@ -671,14 +531,16 @@ def does_machine_have_batch(machine=None):
 ###############################################################################
     """
     Return if this machine has a batch system
-
```
jgfouca: Please keep newline to separate documentation from doctest unit tests.

utils/python/create_test_impl.py
```diff
@@ -425,7 +421,7 @@ def _handle_test_status_file(self, test_name, test_phase, success):
             # TODO: What to do here? This failure is very severe because the
             # only way for test results to be communicated is by the TestStatus
             # file.
-            warning("VERY BAD! Could not handle TestStatus file '%s': '%s'" %
```
jgfouca: This error should be critical. I only used warning because I couldn't use expect(False, ...) due to threading issues.

utils/python/tests/scripts_regression_tests
```diff
@@ -44,6 +44,9 @@ class RunUnitTests(unittest.TestCase):
     def test_wait_for_tests_unit_test(self):
         self.do_unit_tests("wait_for_tests")
```
jgfouca: You can delete this method. It's a wrapper around the doctests

utils/python/tests/scripts_regression_tests
```diff
@@ -664,6 +668,16 @@ class TestUpdateACMETests(unittest.TestCase):
         self.assertTrue("ERS.f19_g16_rx1.E.melvin_gnu" in tests)
 
 ###########################################################################
+    def resolve_variable_name(self):
```
jgfouca: Rename this method to "test_resolve_variable_name"

utils/python/tests/scripts_regression_tests
```diff
@@ -664,6 +668,16 @@ class TestUpdateACMETests(unittest.TestCase):
...
+        from  CIME.XML.Machines import Machines
```
jgfouca: Why not do these imports at top of file?

utils/python/tests/test.py
```diff
@@ -0,0 +1,20 @@
+#!/usr/bin/env python
```
jgfouca: You can delete this file.

cime_config/acme/allactive/testlist_allactive.xml
jedwards4b: No - it's because A and X tests are now defined in driver_cpl/cime_config/testdefs/testlist_drv.xml

jedwards4b:
Okay I think that its ready to pull.   Let me know if you find any further issues.

PR #49, opened 2016-01-28
Changes to get all tests working again on ACME side.

jgfouca:
1) Lots of places where set_verbosity was left in

Solved by encapsulating standard script setup stuff and having
every script import that.

2) Encapsulate handling of standard logging options to reduce duplication

3) Remove obsolete scripts related to original CIME merge.

4) Disable update_acme_tests regression test

5) Add full system test to script_regression_tests


jgfouca:
Going to annotate the changes I made to CIME/XML, just a moment.

utils/python/CIME/XML/GenericXML.py
```diff
@@ -31,7 +31,7 @@ def __init__(self, infile=None):
             self.filename = infile
             self.root = ET.Element(infile)
             self.root.set('version','1.0')
-
```
jgfouca: Remove trailing whitespace. I have my emacs set to highlight trailing whitespace, so I notice it right away. It's a bit of a nitpick, but it can result in problems.
http://programmers.stackexchange.com/questions/121555/why-is-trailing-whitespace-a-big-deal

utils/python/CIME/XML/GenericXML.py
```diff
@@ -84,11 +84,16 @@ def get_value(self, item):
...
         logging.debug("Get Value for "+item)
+        result = None
         if item in self.lookups.keys():
-            return self.lookups[item] 
+            result = self.lookups[item]
         if item in os.environ:
-            return os.environ.get(item)
-#        expect(False, "Not implemented")
+            result = os.environ.get(item)
+
```
jgfouca: Previously, if both if checks failed, the function just ended, giving the user a None. I think None is the right return in such a case, but figured they should probably at least get a warning.

utils/python/CIME/XML/GenericXML.py
```diff
@@ -119,6 +127,6 @@ def get_resolved_value(self, raw_value):
             ref = self.get_value(var)
             if(ref is not None):
                 logging.debug("resolve: "+ref)
-                item_data = item_data.replace(m.group(), ref)
+                item_data = item_data.replace(m.group(), self.get_resolved_value(ref))
```
jgfouca: This was the only significant behavioral change. I was getting unresolved values because ref itself had refs to other things.

utils/python/CIME/XML/Machines.py
```diff
@@ -10,8 +10,9 @@ class Machines(GenericXML):
     def __init__(self,infile):
         """ initialize an object """
         logging.info("Open file "+infile)
-        self.machine = None
         GenericXML.__init__(self,infile)
+        self.machine = None
```
jgfouca: Initialize name member in constructor so it's never undefined.

utils/python/cime_util.py
```diff
@@ -448,9 +447,8 @@ def parse_config_machines():
         config_machines = files.get_resolved_value(files.get_value('MACHINES_SPEC_FILE'))
         _MACHINE_INFO = Machines(config_machines)
 
-
 ###############################################################################
-def get_machine_info( items, machine=None, user=None, project=None, case=None, raw=False):
```
jgfouca: This will need future work.

utils/python/cime_util.py
```diff
@@ -538,8 +535,8 @@ def does_machine_have_batch(machine=None):
     True
     """
     parse_config_machines()
-    if(machine is not None):
```
jgfouca: This is another minor nitpick. I'd prefer if a space was left between "if" and the conditional expression. Always using parens in if statements is a personal style, not sure if it's PEP compliant or not.

utils/python/create_test_impl.py
```diff
@@ -347,7 +352,11 @@ def _build_phase(self, test_name):
     def _run_phase(self, test_name):
     ###########################################################################
         test_dir = self._get_test_dir(test_name)
-        return self._run_phase_command(test_name, "./case.submit", RUN_PHASE, from_dir=test_dir)
+        # TODO: Submit should work in either case.
+        if (self._no_batch):
```
jgfouca: Using submit did not work on melvin.

utils/python/CIME/XML/GenericXML.py
jedwards4b: I just added this option to my emacs:
(add-hook 'before-save-hook 'delete-trailing-whitespace)

utils/python/CIME/XML/GenericXML.py
jedwards4b: I think that this warning may be a problem unless we add a has_value function to test whether the variable exists.  The reason is that at the Case level I am always searching all of the objects until I find the variable defined, so None is the expected result pretty frequently.

utils/python/CIME/XML/GenericXML.py
jedwards4b: Thanks - I took that out at some point while debugging and neglected to put it back.

utils/python/cime_util.py
jedwards4b: Coming very soon in my next pull request :-)

utils/python/CIME/XML/GenericXML.py
jgfouca: Hmm, didn't anticipate that usecase. Well, I'm fine with deleting the warning for now.

utils/python/create_test_impl.py
jedwards4b: We can't not use submit since that is the mechanism via which archiving is handled.   The
ERR test in cime_developer tests this feature.

utils/python/cime_util.py
jgfouca: Great! I'm definitely sick of this half-baked way getting config

utils/python/create_test_impl.py
jgfouca: OK, I will debug this.

PR #50, opened 2016-01-29
Get scripts_regression_tests working on melvin with CESM model

jgfouca:
Had to add some stuff to cesm/machines to support melvin.

Improve output coming from scripts_regression_tests.


utils/python/CIME/XML/Machines.py
```diff
@@ -62,7 +62,7 @@ def set_machine(self,machine):
         if(self.machine is not None and self.name is not machine):
             self.machine = None
         mach_nodes = self.get_node('machine',{'MACH':machine})
-        expect(mach_nodes is not None,"No machine %s found" % machine)
```
jgfouca: mach_nodes appears to be empty list, not None, when there's no match.

PR #51, opened 2016-02-01
Jedwards/module restructure

jedwards4b:
Restructures python modules to fit better the OO structure.


scripts-python/bless_test_results
```diff
@@ -135,11 +138,13 @@ def bless_history(test_name, baseline_tag, baseline_dir_for_test, testcase_dir_f
...
 
     case = os.path.basename(testcase_dir_for_test)
-    cime_root = cime_util.get_cime_root()
+    cime_root = CIME.utils.get_cime_root()
     compgen = os.path.join(cime_root, "scripts", "Tools", "component_compgen_baseline.sh")
     machine_env = os.path.join(testcase_dir_for_test, ".env_mach_specific.sh")
     run_dir = os.path.join(acme_root, case, "run")
-    cprnc_loc = cime_util.get_machine_info("CPRNC_ROOT")
+    machobj = Machines()
+    machobj.probe_machine_name()
```
jgfouca: How about having Machine objects automatically probe for machine if user does not give them one?

scripts-python/create_test
```diff
@@ -135,128 +140,91 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
     if (args.no_build):
         args.no_run = True
 
-    use_batch = cime_util.does_machine_have_batch()
```
jgfouca: Can you explain why you wanted to move this stuff? It's not so bad to do argument validation in the argument parser.

scripts/Tools/component_compare.sh
```diff
@@ -179,6 +179,7 @@ test_hist=$test_hist_init
 #----------------------------------------------------------------------
 # Process command-line arguments
 #----------------------------------------------------------------------
+echo $#
```
jgfouca: Was this for debugging?

utils/python/CIME/CreateTest.py
```diff
@@ -31,28 +33,30 @@ def __init__(self, test_names,
                  clean=False,compare=False, generate=False, namelists_only=False,
                  project=None, parallel_jobs=None):
     ###########################################################################
-        self._cime_root      = cime_util.get_cime_root()
+        self._machobj = machobj
+        self._cime_root      = CIME.utils.get_cime_root()
         self._test_names     = test_names
-        self._no_build       = no_build      if not namelists_only else True
```
jgfouca: Is all this stuff checked in create_test now? It didn't look to me like it was.

scripts-python/create_test
jedwards4b: I find it very confusing to mix the argument parsing with other code, I want to know for example whether the user has input a machine or there is a machine in test request before I select a default compiler

scripts/Tools/component_compare.sh
jedwards4b: yes, i will remove.

utils/python/CIME/CreateTest.py
jedwards4b: Yes I think that it is.

utils/python/CIME/CreateTest.py
```diff
@@ -31,28 +33,30 @@ def __init__(self, test_names,
...
-        self._no_run         = no_run        if not self._no_build else True
-        self._no_batch       = no_batch      if no_batch is not None else not cime_util.does_machine_have_batch()
-        self._test_root      = test_root     if test_root is not None else cime_util.get_machine_info("CESMSCRATCHROOT")
-        self._test_id        = test_id       if test_id is not None else cime_util.get_utc_timestamp()
-        self._project        = project       if project is not None else cime_util.get_machine_project()
-        self._baseline_root  = baseline_root if baseline_root is not None else cime_util.get_machine_info("CCSM_BASELINE")
```
jgfouca: I understand not doing this work twice. Maybe we should handle all the default setting and argument validation here. It seems like now this class is less usable as a library because none of the arguments are defaultable anymore. For example, if the user fails to provide a test_root argument, which is legal because it defaults to None, this class will now try to use None as self._test_root which won't work.

jedwards4b: Okay - let me look into that.

utils/python/CIME/XML/Machines.py
```diff
@@ -65,8 +86,9 @@ def set_machine(self,machine):
         expect(mach_nodes, "No machine %s found" % machine)
         self.machine = mach_nodes[0]
         self.name = machine
+        return machine
 
-    def get_value(self,name):
+    def get_value(self,name,resolved=False):
```
jgfouca: Shouldn't this default to True? How often do we want unresolved values?

utils/python/tests/scripts_regression_tests
```diff
@@ -4,18 +4,24 @@ import unittest, sys, os, tempfile, shutil, re, threading, time, signal, glob, s
...
 
-SCRIPT_DIR = cime_util.get_acme_scripts_root()
+SCRIPT_DIR = CIME.utils.get_acme_scripts_root()
 
 DART_CONFIG = "DartConfiguration.tcl"
 DART_BACKUP = "temp_dart_backup"
 
 logging.basicConfig()
 
+#def load_tests(loader, tests, ignore):
```
jgfouca: Delete this?

utils/python/tests/scripts_regression_tests
```diff
@@ -147,19 +159,19 @@ class TestWaitForTests(unittest.TestCase):
             shutil.move(DART_CONFIG, DART_BACKUP)
 
     ###########################################################################
-    def tearDown(self):
+#    def tearDown(self):
```
jgfouca: Probably want to turn this back on now.

utils/python/tests/scripts_regression_tests
```diff
@@ -302,40 +314,41 @@ class TestCreateTestCommon(unittest.TestCase):
...
-        self._compiler          = cime_util.get_machine_info("COMPILER")
+        self._compiler          = machine.get_default_compiler()
 
         # wait_for_tests could blow away our dart config, need to back it up
         if (os.path.exists(DART_CONFIG)):
             shutil.move(DART_CONFIG, DART_BACKUP)
 
     ###########################################################################
-    def tearDown(self):
+#    def tearDown(self):
```
jgfouca: Turn back on.

utils/python/tests/scripts_regression_tests
```diff
@@ -680,20 +695,6 @@ class TestBlessTestResults(TestCreateTestCommon):
 #         self.assertTrue("ERS.f19_g16_rx1.D.melvin_gnu" in tests)
 #         self.assertTrue("ERS.f19_g16_rx1.E.melvin_gnu" in tests)
 
-###############################################################################
```
jgfouca: We need to move this somewhere before we delete it.

utils/python/update_acme_tests.py
```diff
@@ -114,9 +114,18 @@ def get_test_suite(suite, machine=None, compiler=None):
...
-    compiler = cime_util.get_machine_info("COMPILER", machine=machine)[0] if compiler is None else compiler
+    machobj = Machines()
+    if(machine is None):
+        machobj.probe_machine_name()
+    else:
+        machobj.set_machine(machine)
+    if(compiler is None):
+        compiler = machobj.get_default_compiler()
+    else:
+        if(machobj.is_valid_compiler(compiler)):
```
jgfouca: This could be slightly simpler as:
expect(machobj.is_valid_compiler(compiler) ,"Compiler %s not supported for machine %s" %(compiler,machine))
machobj.set_value('COMPILER',compiler)

utils/python/tests/scripts_regression_tests
```diff
@@ -488,12 +501,12 @@ class TestCreateTestImpl(TestCreateTestCommon):
         for test_status in test_statuses:
             status, test_name = wait_for_tests.parse_test_status_file(test_status)
             if (test_name == build_fail_test):
-                self.assertEqual(status[create_test_impl.BUILD_PHASE], wait_for_tests.TEST_FAIL_STATUS)
```
jedwards4b: BUILD_PHASE is a member of CreateTest - assigned at line 73 of CreateTest.py

utils/python/tests/scripts_regression_tests
```diff
@@ -302,40 +314,41 @@ class TestCreateTestCommon(unittest.TestCase):
...
-        self._compiler          = cime_util.get_machine_info("COMPILER")
+        self._compiler          = machine.get_default_compiler()
 
         # wait_for_tests could blow away our dart config, need to back it up
         if (os.path.exists(DART_CONFIG)):
             shutil.move(DART_CONFIG, DART_BACKUP)
 
     ###########################################################################
-    def tearDown(self):
+#    def tearDown(self):
```
jgfouca: Turn back on.

utils/python/tests/scripts_regression_tests
jgfouca: I just checked out your branch and I see BUILD_PHASE being assigned as a global on line 18 of CreateTest.py.

utils/python/update_acme_tests.py
jedwards4b: corrected

PR #57, opened 2016-02-06
Jedwards/testlists

jedwards4b:
This adds support for testlist.xml files.  It supports both the older (version 1.0) and newer (version 2.0) formats for testlist files.   version 2.0 files have an optional wallclock field which allows you to set the
wallclock time for each test thus improving throughput.   In create_test the self._test_names list of
tests was replaced by a self._tests dict.

### DIFF
--- a/scripts-python/create_test
+++ b/scripts-python/create_test
@@ -15,8 +15,6 @@ from standard_script_setup import *
 import update_acme_tests
 from CIME.create_test import CreateTest
 from CIME.utils import expect
-from CIME.XML.machines import Machines
-from CIME.XML.files import Files
 
 import argparse, doctest, tempfile
 
@@ -25,6 +23,8 @@ def parse_command_line(args, description):
 ###############################################################################
     parser = argparse.ArgumentParser(
         usage="""\n%s <TEST|SUITE> [<TEST|SUITE> ...] [--verbose]
+OR
+%s --xml_category [CATEGORY] [--xml_machine ...]
 OR
 %s --help
 OR
@@ -48,7 +48,7 @@ OR
 
     \033[1;32m# Run all tests in a suite except for tests that are in another suite \033[0m
     > %s <SUITE1> ^<SUITE2>
-""" % ((os.path.basename(args[0]), ) * 9),
+""" % ((os.path.basename(args[0]), ) * 10),
 
 description=description,
 
@@ -59,7 +59,7 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
 
     CIME.utils.setup_standard_logging_options(parser)
 
-    parser.add_argument("testargs", nargs="+", help="Tests or test suites to run. Testnames expect in form CASE.GRID.COMPSET")
+    parser.add_argument("testargs", nargs="*", help="Tests or test suites to run. Testnames expect in form CASE.GRID.COMPSET")
 
     parser.add_argument("--no-run", action="store_true",
                         help="Do not run generated tests")
@@ -99,7 +99,7 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
 
     parser.add_argument("-m", "--machine",
                         help="The machine for which to build tests, this machine must be defined in the "
-                        "config_mahines.xml file for the given model. Default is to match the name of the "
+                        "config_machines.xml file for the given model. Default is to match the name of the "
                         "machine in the test name or the name of the machine "
                         "this script is run on to the NODENAME_REGEX field in config_machines.xml"  )
 
@@ -120,6 +120,13 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
     parser.add_argument("-j", "--parallel-jobs", type=int, default=None,
                         help="Number of tasks create_test should perform simultaneously. Default "
                         "will be min(num_cores, num_tests).")
+
+
+    parser.add_argument("--xml_machine",help="Use this machine key in the lookup in testlist.xml")
+    parser.add_argument("--xml_compiler",help="Use this compiler key in the lookup in testlist.xml")
+    parser.add_argument("--xml_category",help="Use this category key in the lookup in testlist.xml")
+    parser.add_argument("--xml_testlist",help="Use this testlist to lookup tests")
+
 
     args = parser.parse_args(args[1:])
 
@@ -149,12 +156,14 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
 
     return args.testargs, args.compiler, args.machine, args.no_run, args.no_build, args.no_batch, \
         args.test_root, args.baseline_root, args.clean, args.compare, args.generate, args.baseline_name, \
-        args.namelists_only, args.project, args.test_id, args.parallel_jobs
+        args.namelists_only, args.project, args.test_id, args.parallel_jobs, args.xml_machine, args.xml_compiler,\
+        args.xml_category, args.xml_testlist
 
 ###############################################################################
 def create_test(testargs, compiler, machine_name, no_run, no_build, no_batch, test_root,
                 baseline_root, clean, compare, generate,
-                baseline_name, namelists_only, project, test_id, parallel_jobs):
+                baseline_name, namelists_only, project, test_id, parallel_jobs,
+                xml_machine, xml_compiler, xml_category, xml_testlist):
 ###############################################################################
     impl = CreateTest(testargs,
                       no_run=no_run, no_build=no_build, no_batch=no_batch,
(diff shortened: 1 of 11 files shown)


jedwards4b:
I've found some problems in scripts_regression_tests that I will need to address before completing this PR.

jedwards4b:
I've resolved the issue in scripts_regression_tests.

scripts-python/create_test
```diff
@@ -121,6 +121,13 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
                         help="Number of tasks create_test should perform simultaneously. Default "
                         "will be min(num_cores, num_tests).")
 
+
+    parser.add_argument("--xml_machine",help="Use this machine key in the lookup in testlist.xml")
```
jgfouca: You probably want --xml-something instead of --xml_something. You almost never see dashes and underscores combined and is not consistent with any of the other options.

scripts-python/create_test
```diff
@@ -48,7 +48,7 @@ OR
```
jgfouca: Might be worthwhile to provide an example of how to select test XML-style.

utils/python/CIME/XML/generic_xml.py
```diff
@@ -57,21 +63,27 @@ def get_node(self, nodename, attributes=None, root=None):
         """
         if(root is None):
             root = self.root
-
+        nodes = []
         xpath = ".//"+nodename
         if(attributes is not None):
             keys = list(attributes.keys())
-            cnt = 0
+            # xml.etree has limited support for xpath and does not allow more than
```
jgfouca: Sorry about the xpath stuff. Python's built-in library support is usually very comprehensive, so it's odd that the xpath support is limited.

utils/python/CIME/XML/machines.py
```diff
@@ -175,8 +175,7 @@ def is_valid_compiler(self,compiler):
         """
         Check the compiler is valid for the current machine
 
-        >>> machobj = Machines()
```
jgfouca: Thanks for cleaning this up. I forgot to update these tests when I changed the constructor.

utils/python/CIME/XML/testlist.py
```diff
@@ -0,0 +1,85 @@
+"""
+Interface to the config_files.xml file.  This class inherits from EntryID.py
+"""
+from standard_module_setup import *
+
+from generic_xml import GenericXML
+from CIME.utils import expect, get_cime_root, get_model
+
```
jgfouca: I'm not going to look too closely at this file since I don't think ACME will ever get it's tests this way.

utils/python/CIME/create_test.py
```diff
@@ -105,15 +114,17 @@ def __init__(self, test_names,
         # This is the only data that multiple threads will simultaneously access
         # Each test has it's own index and setting/retrieving items from a list
         # is atomic, so this should be fine to use without mutex
-        self._test_states = [ (INITIAL_PHASE, TEST_PASS_STATUS) ] * len(self._test_names)
+        for test in self._tests:
```
jgfouca: Ohhh, I wish you had asked me about this. I know the data structure looks a bit odd, but there was a reason for my madness. If you look at the comment above, you can see that I wanted the update of test_states to be atomic. This is important because the parallel code below is lock-free. IE, it relies on atomic operations instead of locks to be thread-safe. Updating two dicts is not atomic.

utils/python/CIME/create_test.py
```diff
@@ -378,32 +392,36 @@ def _nlcomp_phase(self, test_name):
...
-        return self._run_phase_command(test_name, "./case.test_build", BUILD_PHASE, from_dir=test_dir)
+        return self._run_phase_command(test, "./case.test_build", BUILD_PHASE, from_dir=test_dir)
 
     ###########################################################################
-    def _run_phase(self, test_name):
+    def _run_phase(self, test):
     ###########################################################################
-        test_dir = self._get_test_dir(test_name)
-        return self._run_phase_command(test_name, "./case.submit", RUN_PHASE, from_dir=test_dir)
+        test_dir = self._get_test_dir(test["name"])
```
jgfouca: The code below confuses me. Can you add a comment to describe what's going on here with the wallclock stuff?

utils/python/CIME/create_test.py
```diff
@@ -506,16 +523,20 @@ def _producer(self):
         while (True):
             work_to_do = False
             num_threads_launched_this_iteration = 0
-            for test_name in self._test_names:
+            for test in self._tests:
```
jgfouca: We probably want self._tests to have a consistent type.

utils/python/CIME/create_test.py
```diff
@@ -585,9 +606,12 @@ def create_test(self):
 
         # Tell user what will be run
         print "RUNNING TESTS:"
-        for test_name in self._test_names:
+        for test in self._tests:
+            if (type(test) == type(dict())):
```
jgfouca: Ditto above.

utils/python/CIME/create_test.py
```diff
@@ -157,31 +168,30 @@ def _get_test_dir(self, test_name):
         return os.path.join(self._test_root, self._get_case_id(test_name))
 
     ###########################################################################
-    def _get_test_data(self, test_name):
+    def _get_test_data(self, test):
     ###########################################################################
-        state_idx = self._test_names.index(test_name)
-        return self._test_states[state_idx]
```
jgfouca: Not atomic!

utils/python/CIME/create_test.py
```diff
@@ -600,30 +624,78 @@ def create_test(self):
...
+                print status, test["name"], phase
 
-            print "    Case dir: %s" % self._get_test_dir(test_name)
+            print "    Case dir: %s" % self._get_test_dir(test["name"])
 
         print "create_test took", time.time() - start_time, "seconds"
 
         return rv
+
+    def  _get_tests_from_xml(self,xml_machine=None,xml_category=None,xml_compiler=None, xml_testlist=None,
```
jgfouca: Can this go in utils or the Testlist class instead?

utils/python/tests/scripts_regression_tests
```diff
@@ -722,7 +722,7 @@ class FullSystemTest(TestCreateTestCommon):
                              msg="COMMAND SHOULD HAVE WORKED\wait_for_tests output:\n%s\n\nerrput:\n%s\n\ncode: %d" % (output, errput, stat))
 
 
-        stat, output, errput = run_cmd("%s/cs_status.%s" % (self._testroot, self._baseline_name), ok_to_fail=True)
```
jgfouca: Hmmm, I wonder how this passed for me? Looks like old version was broken. Sorry if I pushed broken code.

jgfouca:
Review complete. Most issues are minor but unfortunately I do have to reject the data structure change in CreateTest because it introduces race conditions. I understand the that the data structure I used must have looked odd at first glance.

utils/python/CIME/create_test.py
jedwards4b: I've read this a few times and I think I understand your concern, but I do not see the danger of which you speak.  Can you give me a concrete example of what you think could go wrong?

utils/python/CIME/create_test.py
jedwards4b: The testlist.xml includes an optional estimated wallclock time for each test,  this is where we apply it.  And this was the motivation to make each test data structure a dict.   Changing the status stuff seemed like an obvious additional cleanup.

utils/python/CIME/create_test.py
jedwards4b: I think that it does.

utils/python/CIME/create_test.py
jedwards4b: yes

jedwards4b:
I need clarification here - is the issue with the entire test dict or just
with
putting status and phase into that dict?   I still don't see where the race
condition arises?

On Tue, Feb 9, 2016 at 9:10 PM, James Foucar <notifications@github.com>
wrote:

> Review complete. Most issues are minor but unfortunately I do have to
> reject the data structure change in CreateTest because it introduces race
> conditions. I understand the that the data structure I used must have
> looked odd at first glance.
>
> —
> <https://github.com/ESMCI/cime/pull/57#issuecomment-182187502>.
>



-- 
Jim Edwards

CESM Software Engineer
Boulder, CO

utils/python/CIME/create_test.py
```diff
@@ -212,12 +221,13 @@ def _update_test_status(self, test_name, phase, status):
                    "New phase should be set to pending status")
             expect(self._phases.index(old_phase) == phase_idx - 1,
                    "Skipped phase?")
-
-        self._test_states[state_idx] = (phase, status)
+        test["phase"] = phase
```
jgfouca: Not atomic.

scripts-python/create_test
jedwards4b: I don't think that line 27 is quite correct:  there are three new flags --xml_category, --xml_machine --xml_compiler, if any one or more of them are provided then we do not expect a <TEST|SUITE> argument.  How do I express that here?

PR #60, opened 2016-02-11
Testspec

jedwards4b:
With this change we provide an initial version of the testspec.xml file used by cesm and we allow the user to optionally provide arguments to the --generate and --compare options.   generate and compare can now be used together if and only if they point to different directories.

### DIFF
--- a/utils/python/CIME/XML/generic_xml.py
+++ b/utils/python/CIME/XML/generic_xml.py
@@ -3,7 +3,7 @@ Common interface to XML files, this is an abstract class and is expected to
 be used by other XML interface modules and not directly.
 """
 from standard_module_setup import *
-
+from xml.dom import minidom
 from CIME.utils import expect, get_cime_root
 
 class GenericXML:
@@ -51,11 +51,12 @@ class GenericXML:
         """
         Write an xml file from data in self
         """
+        if(infile is None):
+            infile = self.filename
         logging.info("write: "+ infile)
-        if(infile != None):
-            self.tree.write(infile)
-        else:
-            self.tree.write(self.filename)
+        doc = minidom.parseString(ET.tostring(self.root))
+        with open(infile,'w') as xmlout:
+            doc.writexml(xmlout,addindent='  ',newl='\n')
 
     def get_node(self, nodename, attributes=None, root=None):
         """
--- a/utils/python/CIME/create_test.py
+++ b/utils/python/CIME/create_test.py
@@ -100,13 +100,12 @@ class CreateTest(object):
         self._compare = False
         self._generate = False
         if (compare or generate):
-            logging.info("compare is %s generate is %s"%(compare,generate))
             # Figure out what baseline name to use
             if (baseline_name is None):
-                if(compare is not None and compare == str(compare)):
+                if(compare is not None and isinstance(compare,str)):
                     self._baseline_cmp_name = compare
                     self._compare = True
-                if(generate is not None and generate == str(generate)):
+                if(generate is not None and isinstance(generate,str)):
                     self._baseline_gen_name = generate
                     self._generate = True
                 branch_name = CIME.utils.get_current_branch(repo=self._cime_root)
--- a/utils/python/CIME/create_test.py
+++ b/utils/python/CIME/create_test.py
@@ -13,7 +13,6 @@ from CIME.XML.env_test import EnvTest
 from CIME.XML.files import Files
 from CIME.XML.component import Component
 from CIME.XML.testlist import Testlist
-from CIME.XML.testspec import TestSpec
 import CIME.test_utils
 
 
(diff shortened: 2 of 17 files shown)


utils/python/CIME/XML/generic_xml.py
```diff
@@ -51,11 +51,12 @@ def write(self, infile=None):
         """
         Write an xml file from data in self
         """
+        if(infile is None):
+            infile = self.filename
         logging.info("write: "+ infile)
-        if(infile != None):
```
jgfouca: What was the motivation for this change? Outputted XML didn't look good?

utils/python/CIME/create_test.py
```diff
@@ -95,17 +98,33 @@ def __init__(self, test_names,
         else:
             self._parallel_jobs = parallel_jobs
 
-        if (self._compare or self._generate):
+        self._baseline_cmp_name = None
+        self._baseline_gen_name = None
 
+        if (self._compare or self._generate):
+            logging.warn("compare is %s generate is %s"%(self._compare,self._generate))
```
jgfouca: Probably should be logging.info instead of logging.warn

utils/python/CIME/create_test.py
```diff
@@ -95,17 +98,33 @@ def __init__(self, test_names,
...
             # Figure out what baseline name to use
             if (baseline_name is None):
+                if(self._compare is not None and self._compare == str(self._compare)):
```
jgfouca: _compare and _generate should always be booleans. If the incoming compare argument is a string, set self._compare to True and self._baseline_cmp_name to compare argument.

utils/python/CIME/create_test.py
```diff
@@ -228,10 +251,12 @@ def _update_test_status(self, test, phase, status):
                    "New phase should be set to pending status")
             expect(self._phases.index(old_phase) == phase_idx - 1,
                    "Skipped phase?")
-
+        if(self._cime_model == "cesm"):
```
jgfouca: Does this really need to be updated every time a state changes or can it just be updated at the end?

utils/python/wait_for_tests.py
```diff
@@ -429,6 +429,12 @@ def wait_for_tests_impl(test_paths, no_wait=False, check_throughput=False, check
 
     return test_results
 
+def update_testspec_xml(test_results):
```
jgfouca: I'm confused by this on many levels. Wait_for_tests already prints statuses. Also, why is this function called "update_testspec_xml" when it's just printing testnames?

utils/python/tests/scripts_regression_tests
```diff
@@ -416,7 +417,8 @@ class TestCreateTestImpl(TestCreateTestCommon):
         for idx, phase in enumerate(ct._phases):
             for test in ct._tests:
                 if (phase == CIME.create_test.INITIAL_PHASE):
-                    continue
+                    if (ct._cime_model == "cesm"):
```
jgfouca: Testing phases shouldn't have anything to do with testspec.

utils/python/CIME/XML/generic_xml.py
jedwards4b: Yes, when I tried to output the internally generated testspec.xml file it was alll on a single line with no formatting.  I would really like to use the xslt library here but the python interface doesn't seem to be generally available.

utils/python/CIME/create_test.py
jedwards4b: fixed

utils/python/CIME/create_test.py
jedwards4b: fixed

utils/python/wait_for_tests.py
jedwards4b: this is incomplete code - i guess I got distracted and forgot about it.  Maybe I'll pull the testspec changes out of this PR and do it separately.

jgfouca:
Review complete.

utils/python/CIME/XML/generic_xml.py
jgfouca: That's odd. When I use generic_xml to write XML files, they seem to look OK.

utils/python/CIME/create_test.py
```diff
@@ -95,17 +95,37 @@ def __init__(self, test_names,
...
-
+        self._baseline_cmp_name = None
+        self._baseline_gen_name = None
+        self._compare = False
+        self._generate = False
+        if (compare or generate):
+            logging.info("compare is %s generate is %s"%(compare,generate))
             # Figure out what baseline name to use
             if (baseline_name is None):
+                if(compare is not None and compare == str(compare)):
```
jgfouca: To check if compare is a string, you want this: "type(compare) is str"

utils/python/wait_for_tests.py
jgfouca: Did you intend to remove this in your recent remove test-spec commit?

jgfouca:
Getting close. Only string type checking in create_test and testspec in wait_for_tests need to be fixed.

utils/python/CIME/create_test.py
bandre-ucar: You should use the isinstance function: if isinstance(compare, str)

jgfouca: Yeah, pylint flags type(X) is Y and tells me to use isinstance instead. I kinda disagree for primitive types like str though, the former just reads better to me.

PR #61, opened 2016-02-11
Major progress converting perl to python: case.build and check_input_…

jgfouca:
…data

Things to note:
1) Machines will now need to specific python information in the modules
section of config_machines.xml
2) Fix a few mistakes in config_component.xml, errant ">" character
3) Implemented Case class from perl-OO branch
       -This should replace xmlquery/xmlchange
4) Had to get rid of XML version in xmlchange. This was causing XML
files to diff between locked files. This restriction can go away
once everything is using Case.
5) New python tools that need to be in caseroot are symlinked instead
of copied. This makes it so all the python/utils don't have to be copied.
6) Add classes for all the env*.xml files. They all inherit from a common
base class to reduce code duplication.
7) Add new env_module module to replace ModuleLoader.pm

### DIFF
--- a/scripts-python/case.build
+++ b/scripts-python/case.build
@@ -1,0 +1,508 @@
+#!/usr/bin/env python
+
+"""
+Script to build a case.
+"""
+
+from standard_script_setup import *
+
+from CIME.utils import expect, run_cmd
+from CIME.case import Case
+from CIME.env_module import EnvModule
+
+import argparse, doctest, shutil, glob
+
+###############################################################################
+def parse_command_line(args, description):
+###############################################################################
+    parser = argparse.ArgumentParser(
+        usage="""\n%s [<casedir>] [--verbose]
+OR
+%s --help
+OR
+%s --test
+
+\033[1mEXAMPLES:\033[0m
+    \033[1;32m# Build case \033[0m
+    > %s
+""" % ((os.path.basename(args[0]), ) * 4),
+
+description=description,
+
+formatter_class=argparse.ArgumentDefaultsHelpFormatter
+)
+
+    CIME.utils.setup_standard_logging_options(parser)
+
+    parser.add_argument("caseroot", nargs="?", default=os.getcwd(),
+                        help="Case directory to build")
+
+    parser.add_argument("-t", "--testmode", action="store_true",
+                        help="Unknown TODO")
+
+    args = parser.parse_args(args[1:])
+
+    CIME.utils.handle_standard_logging_options(args)
+
+    return args.caseroot, args.testmode
+
+###############################################################################
+def check_input_data(case):
+###############################################################################
+    run_cmd("./check_input_data --download")
+
+    get_refcase  = case.get_value("GET_REFCASE")
+    run_type     = case.get_value("RUN_TYPE")
+    continue_run = case.get_value("CONTINUE_RUN")
+
+    # JGF - I don't have the foggiest idea what this is for so this is just
+    # a 1-1 perl->python conversion
+    if (get_refcase == "TRUE" and run_type != "startup" and continue_run == "FALSE"):
+        din_loc_root = case.get_value("DIN_LOC_ROOT")
+        run_refdate  = case.get_value("RUN_REFDATE")
+        run_refcase  = case.get_value("RUN_REFCASE")
+        run_refdir   = case.get_value("RUN_REFDIR")
+        rundir       = case.get_value("RUNDIR")
+
+        refdir = os.path.join(run_refdir, run_refcase, run_refdate)
+        expect(os.path.isdir(refdir),
+"""
+*****************************************************************
+ccsm_prestage ERROR: $refdir is not on local disk
+obtain this data from the svn input data repository
+> mkdir -p %s
+> cd %s
+> cd ..
+> svn export --force https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata/%s
+or set GET_REFCASE to FALSE in env_run.xml
+and prestage the restart data to $RUNDIR manually
+*****************************************************************""" % (refdir, refdir, refdir))
+
+        logging.info(" - Prestaging REFCASE (%s) to %s" % (refdir, rundir))
+
+        # prestage the reference case's files.
+
+        if (not os.path.exists(rundir)):
+            os.makedirs(rundir)
+
+        refcasefiles = glob.glob("%s/%s/*%s*" % (din_loc_root, refdir, run_refcase))
+        for rcfile in refcasefiles:
+            rcbaseline = os.path.basename(rcfile)
+            if (not os.path.exists("%s/%s" % (rundir, rcbaseline))):
+                os.symlink(rcfile, "%s/%s" % ((rundir, rcbaseline)))
+
+            # copy the refcases' rpointer files to the run directory
+            rpointerfiles = glob.glob("%s/%s/*rpointer*" % (din_loc_root, refdir))
+            for rpointerfile in rpointerfiles:
+                shutil.copy(rpointerfile, rundir)
+
+            cam2_list = glob.glob("%s/*.cam2.*" % rundir)
+            for cam2file in cam2_list:
+                camfile = cam2file.replace("cam2", "cam")
+                os.symlink(cam2file, camfile)
+
+            allrundirfiles = glob.glob("%s/*" % rundir)
+            for runfile in allrundirfiles:
+                os.chmod(runfile, 0755)
+
+###############################################################################
+def build_checks(case, build_threaded, comp_interface, use_esmf_lib, mpilib,
+                 nthrds_cpl, nthrds_atm, nthrds_lnd, nthrds_ice, nthrds_ocn, nthrds_glc, nthrds_wav, nthrds_rof, ninst_build, smp_value):
+###############################################################################
+    ninst_atm    = int(case.get_value("NINST_ATM"))
+    ninst_lnd    = int(case.get_value("NINST_LND"))
+    ninst_ice    = int(case.get_value("NINST_ICE"))
+    ninst_ocn    = int(case.get_value("NINST_OCN"))
+    ninst_glc    = int(case.get_value("NINST_GLC"))
+    ninst_wav    = int(case.get_value("NINST_WAV"))
+    ninst_rof    = int(case.get_value("NINST_ROF"))
+    ninst_value  = case.get_value("NINST_VALUE")
+    smp_build    = int(case.get_value("SMP_BUILD"))
+    build_status = int(case.get_value("BUILD_STATUS"))
+
+    atmstr = 1 if (build_threaded == "TRUE" or nthrds_atm > 1) else 0
+    lndstr = 1 if (build_threaded == "TRUE" or nthrds_lnd > 1) else 0
+    icestr = 1 if (build_threaded == "TRUE" or nthrds_ice > 1) else 0 # not in perl
+    ocnstr = 1 if (build_threaded == "TRUE" or nthrds_ocn > 1) else 0
+    rofstr = 1 if (build_threaded == "TRUE" or nthrds_rof > 1) else 0
+    glcstr = 1 if (build_threaded == "TRUE" or nthrds_glc > 1) else 0
+    wavstr = 1 if (build_threaded == "TRUE" or nthrds_wav > 1) else 0
+    cplstr = 1 if (build_threaded == "TRUE" or nthrds_cpl > 1) else 0
+
+    if (nthrds_atm > 1 or nthrds_lnd > 1 or nthrds_ice > 1 or nthrds_ocn > 1 or
+        nthrds_rof > 1 or nthrds_glc > 1 or nthrds_wav > 1 or nthrds_cpl > 1):
+        os.environ["SMP"] = "TRUE"
+    else:
+        os.environ["SMP"] = "FALSE"
+
+    smpstr = "a%dl%dr%di%do%dg%dw%dc%d" % \
+        (atmstr, lndstr, rofstr, icestr, ocnstr,  glcstr, wavstr, cplstr)
+
+    case.set_value("SMP_VALUE", smpstr)
+    os.environ["SMP_VALUE"] = smpstr
+
+    inststr = "a%dl%dr%di%do%dg%dw%d" % \
+        (ninst_atm, ninst_lnd, ninst_rof, ninst_ice, ninst_ocn, ninst_glc, ninst_wav)
+
+    case.set_value("NINST_VALUE", inststr)
+    os.environ["NINST_VALUE"] = inststr
+
+    expect( ninst_build == ninst_value or ninst_build == "0",
+            """
+ERROR, NINST VALUES HAVE CHANGED
+  NINST_BUILD = %s
+  NINST_VALUE = %s
+  A manual clean of your obj directories is strongly recommended
+  You should execute the following:
+    ./case.clean_build
+  Then rerun the build script interactively
+  ---- OR ----
+  You can override this error message at your own risk by executing:
+    ./xmlchange -file env_build.xml -id NINST_BUILD -val 0
+  Then rerun the build script interactively
+""" % (ninst_build, ninst_value))
+
+    expect( smp_build == smpstr or smp_build == 0,
+            """
+ERROR, SMP VALUES HAVE CHANGED
+  SMP_BUILD = %s
+  SMP_VALUE = %s
+  A manual clean of your obj directories is strongly recommended
+  You should execute the following:
+    ./case.clean_build
+  Then rerun the build script interactively
+  ---- OR ----
+  You can override this error message at your own risk by executing:
+    ./xmlchange -file env_build.xml -id SMP_BUILD -val 0
+  Then rerun the build script interactively
+""" % (smp_build, smp_value))
+
+    expect(build_status == 0,
+           """
+ERROR env_build HAS CHANGED
+  A manual clean of your obj directories is required
+  You should execute the following:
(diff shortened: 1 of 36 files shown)


jgfouca:
I will annotate to help you review.

scripts-python/case.build
```diff
@@ -0,0 +1,508 @@
...
+formatter_class=argparse.ArgumentDefaultsHelpFormatter
+)
+
+    CIME.utils.setup_standard_logging_options(parser)
+
+    parser.add_argument("caseroot", nargs="?", default=os.getcwd(),
+                        help="Case directory to build")
+
+    parser.add_argument("-t", "--testmode", action="store_true",
+                        help="Unknown TODO")
```
jgfouca: Wasn't sure what this option means.

scripts-python/case.build
```diff
@@ -0,0 +1,508 @@
...
+###############################################################################
+def check_input_data(case):
+###############################################################################
+    run_cmd("./check_input_data --download")
+
+    get_refcase  = case.get_value("GET_REFCASE")
+    run_type     = case.get_value("RUN_TYPE")
+    continue_run = case.get_value("CONTINUE_RUN")
+
+    # JGF - I don't have the foggiest idea what this is for so this is just
```
jgfouca: Can you explain what this code is for?

scripts-python/case.build
```diff
@@ -0,0 +1,508 @@
+#!/usr/bin/env python
+
+"""
```
jgfouca: The implementation of this file follows the perl very closely. We can clean up more later.

scripts-python/case.build
```diff
@@ -0,0 +1,508 @@
...
+    logs = []
+    overall_smp = os.environ["SMP"]
+
+    for model, comp, config_dir, nthrds in models_build_data:
+        os.environ["MODEL"] = model
+        if (nthrds > 1 or build_threaded == "TRUE"):
+            os.environ["SMP"] = "TRUE"
+        else:
+            os.environ["SMP"] = "FALSE"
+
```
jgfouca: I punted on the special case for clm until I could get some comment from you about it.

scripts/Tools/xmlchange
```diff
@@ -272,8 +272,6 @@ foreach my $file (values (%id_file))
     my $xml = XML::LibXML->new( no_blanks => 1)->parse_file($file); 
 
     my $fh = IO::File->new($file, '>' ) or $logger->logdie ("can't open file: $file");
```
jgfouca: This xml version tag was causing lockedfiles to diff.

utils/python/CIME/XML/machines.py
```diff
@@ -235,3 +235,33 @@ def get_batch_system_type(self):
         """
         batch_system = self.get_node("batch_system")
         return batch_system[0].get("type")
+
```
jgfouca: Curious as to your opinion here. I was trying to follow the system being followed in the rest of the file.

utils/python/CIME/case.py
```diff
@@ -0,0 +1,80 @@
...
+    def get_value(self, item):
+        for env_file in self._env_files:
+            result = env_file.get_value(item)
+            if (result is not None):
+                return self.get_resolved_value(result)
+
+        logging.warning("Not able to retreive value for item '%s'" % item)
+
+    def get_resolved_value(self, item):
+        # TODO HACK - surely there is a better way?
```
jgfouca: Thoughts here?

scripts-python/case.build
jedwards4b: The idea is that we want to be able to run this script on machines other than the machine specified for the build and
so in testmode it should ignore certain errors such as input or output paths.   It's not been implemented consistently across scripts, but it is a feature that I want to support in some form.   An example of where this is useful:  using the --namelist-only flag to create_test I can test for namelist changes on any machine from a single location.

scripts-python/case.build
jedwards4b: We do not fully populate the inputdata directory on every machine and do not expect every user to download the 3TB+ of data in our inputdata repository.   This code checks for the existence of inputdata  in the local inputdata directory and attempts to download data from the server if it's needed and missing.

scripts-python/case.build
```diff
@@ -0,0 +1,508 @@
...
+            allrundirfiles = glob.glob("%s/*" % rundir)
+            for runfile in allrundirfiles:
+                os.chmod(runfile, 0755)
+
+###############################################################################
+def build_checks(case, build_threaded, comp_interface, use_esmf_lib, mpilib,
+                 nthrds_cpl, nthrds_atm, nthrds_lnd, nthrds_ice, nthrds_ocn, nthrds_glc, nthrds_wav, nthrds_rof, ninst_build, smp_value):
+###############################################################################
+    ninst_atm    = int(case.get_value("NINST_ATM"))
```
jedwards4b: I would really like to see all these things in an array or dict that we can loop over components, but that is a future cleanup.

scripts-python/case.build
```diff
@@ -0,0 +1,508 @@
...
+
+    sharedpath = os.path.join(sharedlibroot, compiler, mpilib, debugdir, threaddir)
+    os.environ["SHAREDPATH"] = sharedpath
+    shared_lib = os.path.join(sharedpath, "lib")
+    shared_inc = os.path.join(sharedpath, "include")
+    for shared_item in [shared_lib, shared_inc]:
+        if (not os.path.exists(shared_item)):
+            os.makedirs(shared_item)
+
+    libs = ["mct", "gptl", "pio", "csm_share"]
```
jedwards4b: Build order here is important and represents a dependency chain.

scripts-python/case.build
```diff
@@ -0,0 +1,508 @@
...
+###############################################################################
+    config_atm_dir = os.path.dirname(case.get_value("CONFIG_ATM_FILE"))
+    config_lnd_dir = os.path.dirname(case.get_value("CONFIG_LND_FILE"))
+    config_ice_dir = os.path.dirname(case.get_value("CONFIG_ICE_FILE"))
+    config_ocn_dir = os.path.dirname(case.get_value("CONFIG_OCN_FILE"))
+    config_glc_dir = os.path.dirname(case.get_value("CONFIG_GLC_FILE"))
+    config_wav_dir = os.path.dirname(case.get_value("CONFIG_WAV_FILE"))
+    config_rof_dir = os.path.dirname(case.get_value("CONFIG_ROF_FILE"))
+
+    # Also defines build order
```
jedwards4b: There are not dependencies between components and so order is not important and in fact we should be able to thread this area and build all at once.

scripts-python/case.build
jedwards4b: clm4.0 build differently than clm4.5 and clm5.0, we need to support both cases.

scripts-python/case.build
jgfouca: That would be great. I'll add that to the branch.

scripts-python/case.build
```diff
@@ -0,0 +1,508 @@
...
+        logs.append(file_build)
+
+    #
+    # Now build the executable
+    #
+
+    os.environ["SMP"] = overall_smp
+
+    # model is overloaded here, confusing
+    model = CIME.utils.get_model()
```
jedwards4b: call this cime_model?

scripts-python/case.build
```diff
@@ -0,0 +1,508 @@
...
+    os.environ["COMP_ROF"]             = comp_rof
+    os.environ["CLM_CONFIG_OPTS"]      = clm_config_opts     if clm_config_opts     is not None else ""
+    os.environ["CAM_CONFIG_OPTS"]      = cam_config_opts     if cam_config_opts     is not None else ""
+    os.environ["PIO_CONFIG_OPTS"]      = pio_config_opts     if pio_config_opts     is not None else ""
+    os.environ["OCN_SUBMODEL"]         = ocn_submodel        if ocn_submodel        is not None else ""
+    os.environ["PROFILE_PAPI_ENABLE"]  = profile_papi_enable if profile_papi_enable is not None else ""
+    os.environ["CLM_USE_PETSC"]        = clm_use_petsc       if clm_use_petsc       is not None else ""
+    os.environ["CISM_USE_TRILINOS"]    = cism_use_trilinos   if cism_use_trilinos   is not None else ""
+    os.environ["MPASLI_USE_ALBANY"]    = mpasli_use_albany   if mpasli_use_albany   is not None else ""
+    lid                                = run_cmd("date +%y%m%d-%H%M%S") # why not testid?
```
jedwards4b: This is a timestamp for the build , not the same as the testid, and this case may not be a test anyway.   For a production experiment there may be many builds of the same case.

scripts-python/check_input_data
```diff
@@ -0,0 +1,164 @@
...
+from standard_script_setup import *
+
+from CIME.utils import expect, get_model, run_cmd
+from CIME.XML.machines import Machines
+from CIME.case import Case
+
+import argparse, doctest, fnmatch
+
+# Should probably be in XML somewhere
+SVN_LOCS = {
```
jedwards4b: maybe config_files.xml is a good place?

utils/python/CIME/XML/machines.py
jedwards4b: Looks fine - the init_paths and and cmd_paths might have an optional lang argument?

utils/python/CIME/case.py
```diff
@@ -0,0 +1,80 @@
...
+        for env_file in self._env_files_that_need_rewrite:
+            env_file.write()
+
+        self._env_files_that_need_rewrite = set()
+
+    def get_value(self, item):
+        for env_file in self._env_files:
+            result = env_file.get_value(item)
+            if (result is not None):
+                return self.get_resolved_value(result)
```
jedwards4b: The default for get_value is not resolved so I don't thiink we need the get_resolved_value call?

utils/python/CIME/case.py
jedwards4b: I think maybe we don't need this, when is the generic_entry.py get_resolved_value not enough?  I think that the issue is when you have a variable in one xml file that depends on a value in another xml file.

utils/python/CIME/env_module.py
```diff
@@ -0,0 +1,109 @@
...
+            return my_value == xml_value
+
+    def _load_module_modules(self, modules_to_load):
+        python_mod_cmd = self._cmd_paths["python"]
+        for action, argument in modules_to_load:
+            cmd = "%s %s %s" % (python_mod_cmd, action, argument)
+            py_module_code = run_cmd(cmd)
+            exec(py_module_code)
+
+    def _load_soft_modules(self, modules_to_load):
```
jedwards4b: we need softenv support I think for mira, I don't know if anyone is using dotfile any longer.   
A feature of ModuleLoader.pm that I really like is that it writes .env_mach_specific.csh and .env_machine_specific.sh files to your case directory and updates them if env_machine_specific.xml is changed.    Those hidden files are really usefull when debugging with ddt or totalview.

rljacob: Blues also uses softenv

utils/python/CIME/case.py
```diff
@@ -0,0 +1,80 @@
...
+from CIME.XML.env_case          import EnvCase
+from CIME.XML.env_mach_pes      import EnvMachPes
+from CIME.XML.env_build         import EnvBuild
+from CIME.XML.env_run           import EnvRun
+from CIME.XML.env_archive       import EnvArchive
+from CIME.XML.env_batch         import EnvBatch
+
+class Case(object):
+
+    def __init__(self, case_root=os.getcwd()):
```
jedwards4b: So we are not accounting for creating a new case yet and this __init__ will in the future allow for case_root to not exist, right?

jedwards4b:
Review complete, no major concerns.

scripts-python/case.build
jgfouca: OK, added comment.

scripts-python/case.build
jgfouca: So, no change is needed, but cannot do these in parallel?

scripts-python/case.build
jgfouca: On second thought, this PR is big already, I'll do that in separate branch.

scripts-python/case.build
jedwards4b: right

scripts-python/case.build
jedwards4b: We still need to handle the clm4 vs clm4+ case.

scripts-python/case.build
```diff
@@ -0,0 +1,508 @@
...
+
+    for model, comp, config_dir, nthrds in models_build_data:
+        os.environ["MODEL"] = model
+        if (nthrds > 1 or build_threaded == "TRUE"):
+            os.environ["SMP"] = "TRUE"
+        else:
+            os.environ["SMP"] = "FALSE"
+
+        # TODO - Special case for clm?
+        # TODO - bldroot?
```
jgfouca: I'm trying to understand how bldroot works in this algorithm. It looks like it might be left as empty string in many cases in the perl version. Is this right?

jedwards4b: yes if it's empty the default setting of EXEROOT is used

scripts-python/case.build
jgfouca: done.

scripts-python/case.build
jgfouca: Added comment

scripts-python/case.build
jgfouca: Comment added.

scripts-python/check_input_data
jgfouca: Could be. I may punt on this for now.

utils/python/CIME/XML/machines.py
jgfouca: Yeah, that's better.

utils/python/CIME/case.py
jgfouca: Correct, we can cross that bridge later.

utils/python/CIME/case.py
jgfouca: I think you're right.

utils/python/CIME/case.py
```diff
@@ -0,0 +1,80 @@
...
+        expect(os.path.isdir(case_root),
+               "Case root directory '%s' does not exist" % case_root)
+
+        self._env_files = []
+        self._env_files_that_need_rewrite = set()
+
+        self._env_files.append(EnvTest(case_root))
```
jedwards4b: EnvRun is missing

utils/python/CIME/case.py
jgfouca: There are two issues. 1) If item is not an entry but just a string referring to other entries. 2) If dependencies cross XML files

utils/python/CIME/env_module.py
jgfouca: Yeah, I wasn't suggesting these not be implemented, I was just trying to limit the scale of the PR.

utils/python/CIME/case.py
jgfouca: Thanks

scripts-python/case.build
jgfouca: Done (I think).

utils/python/CIME/case.py
jgfouca: Hmm, it looks like the cross-XML dependency is an issue here as well. When I take out the get_resolved_value, I get this error:

OSError: [Errno 2] No such file or directory: '/home/jgfouca/acme/scratch/$CASE/bld'

Because EXEROOT doesn't get fully resolved.

quantheory:
I haven't been following this too closely (don't have time), but if you're using Python 2 specifically I recommend using the shebang:
```python
#!/usr/bin/env python2
```
instead of
```python
#!/usr/bin/env python
```
There are a few weird cases (e.g. some Linux distributions) where `python` is soft-linked to `python3`. So if you're only testing these scripts with Python 2, you probably want to specify that that's what you want.

bertinia:
In the postprocessing scripts, we actually go a little further with the
following snippet of code at the
top of the script to ensure 2.7.x is being using:

#!/usr/bin/env python

from __future__ import print_function

import sys

# check the system python version and require 2.7.x or greater
if sys.hexversion < 0x02070000:
    print(70 * "*")
    print("ERROR: {0} requires python >= 2.7.x. ".format(sys.argv[0]))
    print("It appears that you are running python {0}".format(
            ".".join(str(x) for x in sys.version_info[0:3])))
    print(70 * "*")
    sys.exit(1)


On Tue, Feb 16, 2016 at 2:17 PM, Sean Patrick Santos <
notifications@github.com> wrote:

> I haven't been following this too closely (don't have time), but if you're
> using Python 2 specifically I recommend using the shebang:
>
> #!/usr/bin/env python2
>
> instead of
>
> #!/usr/bin/env python
>
> There are a few weird cases (e.g. some Linux distributions) where python
> is soft-linked to python3. So if you're only testing these scripts with
> Python 2, you probably want to specify that that's what you want.
>
> —
> <https://github.com/ESMCI/cime/pull/61#issuecomment-184878035>.
>



-- 
Alice Bertini
NCAR - CSEG Software Engineer

jedwards4b:
Superceded by PR #62

rljacob:
@quantheory or @bertinia , please make a github issue about verifying python versions so it doesn't get lost in the PR review comments.

jgfouca:
@quantheory , python2 is probably better. The one problem, which I immediately noticed, is that emacs was not smart enough to load python-mode after this change was made. It was easy to fix in my .emacs file after some googling, but it might trip up other users.

PR #62, opened 2016-02-16
Jgfouca/more xml

jedwards4b:
This request supersedes #61 and solves issues with xml Lockedfiles not matching.   I also renamed the
module CIME/create_test.py to CIME/system_test.py since a noun makes more sense to me as an object name than a verb does.


jgfouca:
This one got a little too big to review, especially with your and my changes mixed together. I tried to go through git log -p to review.

PR #65, opened 2016-02-17
need a flush before preview_namelist in case.build, need to allow for…

jedwards4b:
… finidat unset in check_input_data, improve parsing in xmlchange


scripts-python/check_input_data
```diff
@@ -117,29 +117,32 @@ def check_input_data(svn_loc, input_data_root, data_list_dir, download):
             if (line and not line.startswith("#")):
                 tokens = line.split('=')
                 description, full_path = tokens[0].strip(), tokens[1].strip()
-
+                if(full_path):
```
jgfouca: Full_path can be empty?

PR #71, opened 2016-02-26
rework logic used to determine if a component pelayout change require…

jedwards4b:
…s a rebuild.

This change in case.setup allows a component to indicate if a pelayout change will 
require a rebuild through use of xml variable $comp_PE_CHANGE_REQUIRES_REBUILD
where $comp is in (ATM LND ROF OCN ICE GLC WAV)  If a component does not define this
variable it is assumed false.

### DIFF
--- a/scripts/Tools/case.setup
+++ b/scripts/Tools/case.setup
@@ -201,6 +201,8 @@ if (! $clean ) {
 
     } else {
 
+	_check_pelayouts_require_rebuild($xmlvars{'COMP_ICE'},$xmlvars{'COMP_OCN'});
+
 	my $sysmod = "./Tools/check_lockedfiles --caseroot $caseroot";
 	system($sysmod) == 0 or $logger->logdie("ERROR case.setup: $sysmod failed: $?");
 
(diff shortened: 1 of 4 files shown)


scripts/Tools/case.setup
```diff
@@ -377,12 +371,12 @@ if ($clean) {
     $sysmod = "rm ./case.run";
     system($sysmod) == 0 or $logger->logdie("ERROR case.setup: $sysmod failed: $?");
 
-    # remove relevant files from $caseroot/LockedFiels
-    my @files = <${caseroot}/LockedFiles/env_build.xml ${caseroot}/LockedFiles/env_mach_pes.xml>;
-    foreach my $file (@files) {
-	$sysmod = "rm -f $file";
-	system($sysmod) == 0 or $logger->logdie("ERROR case.setup: $sysmod failed: $?");
-    }
```
jgfouca: Any reason to hang on to this commented-out code?

scripts/Tools/case.setup
```diff
@@ -412,6 +406,35 @@ if ($clean) {
...
+		    my $cleanflag = lc substr($comp,0,1);
+		    system("./case.clean_build -$cleanflag");
+		}
+	    }
+	}
+	unlink("LockedFiles/env_mach_pes.xml");
+    }
+
+}
+
```
jgfouca: Nitpick: no need for so many blank lines between subroutines

PR #74, opened 2016-02-29
Restore ability to share sharedlibs in create_test.

jgfouca:
Change list:
1) Split BUILD phase into MODEL_BUILD and SHAREDLIB_BUILD
2) case.build now builds all models in parallel!
3) Sharedlib builds are serialized by clever use of procs_needed
4) Crucial Makefile fix from Jim Edwards
5) Add --model-only, --sharedlib-only options to case.build
6) Add --model-only, --sharedlib-only options to case.test_build
7) Add support for model-only, sharedlib-only to all the SystemTest classes
8) New tools for debugging: case_diff, normalize_cases.
 - Used to find subte differences in infrastructure behavior at the case level

9) Move most of check_input_files into a library
10) Remove non-threadsafe things from preview_namelists
11) Encapsulate fake system tests into a common base class
12) Add some missing fake system tests (TESTBUILDFAIL, TESTRUNSLOWPASS)
13) Add some timing data to case.build to assist future optimizations
14) Add prototype replacement for preview_namelists (not yet active)

### DIFF
--- a/scripts/Tools/check_input_data
+++ b/scripts/Tools/check_input_data
@@ -63,9 +63,13 @@ def _main_func(description):
         test_results = doctest.testmod(verbose=True)
         sys.exit(1 if test_results.failed > 0 else 0)
 
-    svn_loc, input_data_root, data_list_file, download = parse_command_line(sys.argv, description)
+    svn_loc, input_data_root, data_list_dir, download = parse_command_line(sys.argv, description)
 
-    sys.exit(0 if check_input_data(svn_loc, input_data_root, data_list_file, download) else 1)
+    sys.exit(0 if check_input_data(case=None,
+                                   svn_loc=svn_loc,
+                                   input_data_root=input_data_root,
+                                   data_list_dir=data_list_dir,
+                                   download=download) else 1)
 
 ###############################################################################
 
--- a/scripts/Tools/check_input_data
+++ b/scripts/Tools/check_input_data
@@ -10,14 +10,9 @@ from standard_script_setup import *
 from CIME.utils import expect, get_model, run_cmd
 from CIME.XML.machines import Machines
 from CIME.case import Case
+from CIME.check_input_data import check_input_data, SVN_LOCS
 
 import argparse, doctest, fnmatch
-
-# Should probably be in XML somewhere
-SVN_LOCS = {
-    "acme" : "https://acme-svn2.ornl.gov/acme-repo/acme/inputdata",
-    "cesm" : "https://svn-ccsm-inputdata.cgd.ucar.edu/trunk/inputdata"
-}
 
 MACHINE = Machines()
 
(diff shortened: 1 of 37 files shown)


jgfouca:
Going to add annotations for you.

scripts/Tools/check_input_data
```diff
@@ -62,91 +57,6 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
     return args.svn_loc, args.input_data_root, args.data_list_dir, args.download
 
 ###############################################################################
```
jgfouca: Moved most of this to a library so other tools can call check_input_data without having to fork a subprocess.

utils/python/CIME/SystemTests/nck.py
```diff
@@ -41,21 +40,21 @@ def build(self):
                         self._case.set_value("ROOTPE_%s"%comp, "%s"%int(rootpe/2))
             self._case.flush()
 
-
             run_cmd("case.setup -clean -testmode")
```
jgfouca: I had to undo some of your changes to this file in order to get identical behavior in both one-phase and two-phase builds. Might be worth thinking about those changes and if we need to bring them back in some form.

utils/python/CIME/build.py
```diff
@@ -498,3 +525,14 @@ def build_libraries(exeroot, caseroot, cimeroot, libroot, sharedlibroot, compile
...
 
 ###############################################################################
+def _build_model_thread(config_dir, caseroot, bldroot, compspec, file_build,
+                        exeroot, model, comp, objdir, incroot, thread_bad_results):
+###############################################################################
+    stat = run_cmd("%s/buildlib %s %s %s >> %s 2>&1" %
+                   (config_dir, caseroot, bldroot, compspec, file_build),
+                   from_dir=os.path.join(exeroot, model), ok_to_fail=True)[0]
+    if (stat != 0):
+        thread_bad_results.append("ERROR: %s.buildlib failed, see %s" % (comp, file_build))
```
jgfouca: Append is atomic.

utils/python/CIME/system_test.py
```diff
@@ -149,12 +150,12 @@ def __init__(self, test_names,
                                                                self._baseline_gen_name)
 
             # Compute baseline_root
-            if baseline_root is None:
-                self._baseline_root = self._machobj.get_value("CCSM_BASELINE")
-            else:
-                self._baseline_root = baseline_root
+            self._baseline_root = baseline_root if baseline_root is not None \
```
jgfouca: I think this is a reasonable way to split a ternary across two lines if we must.

utils/python/CIME/utils.py
```diff
@@ -149,8 +149,7 @@ def run_cmd(cmd, ok_to_fail=False, input_str=None, from_dir=None, verbose=None,
     if (arg_stderr is _hack):
         arg_stderr = subprocess.PIPE
```
jgfouca: May need to discuss this further. If the verbose argument is True but logging.info is off, then logging.info prints nothing when it should. If the opposite is true, then we have the opposite problem.

jgfouca:
Annotations complete.

utils/python/CIME/build.py
```diff
@@ -274,25 +285,41 @@ def case_build(caseroot, testmode):
...
     env_module = EnvModule(mach, compiler, cimeroot, caseroot, mpilib, debug)
     env_module.load_env_for_case()
-    # Need to flush case xml to disk before calling preview_namelists
-    case.flush()
-    run_cmd("./preview_namelists")
 
-    build_checks(case, build_threaded, comp_interface, use_esmf_lib, mpilib,
+    if not sharedlib_only:
+        # Need to flush case xml to disk before calling preview_namelists
+        case.flush()
```
jedwards4b: I think that this case.flush() needs to happen in both passes - the case may have been updated with something needed for the sharedlib build.

utils/python/CIME/check_input_data.py
```diff
@@ -0,0 +1,104 @@
...
+        else:
+            return True
+
+def check_input_data(case=None, svn_loc=None, input_data_root=None, data_list_dir="Buildconf", download=False):
+    """
+    Return True if no files missing
+    """
+    # Fill in defaults as needed
+    case = Case() if case is None else case
+    svn_loc = SVN_LOCS[get_model()] if svn_loc is None else snv_loc
```
jedwards4b: typo here: snv_lock should be svn_lock

utils/python/CIME/check_input_data.py
```diff
@@ -0,0 +1,104 @@
...
+            line = line.strip()
+            if (line and not line.startswith("#")):
+                tokens = line.split('=')
+                description, full_path = tokens[0].strip(), tokens[1].strip()
+                if(full_path):
+                # expand xml variables
+                    full_path = case.get_resolved_value(full_path)
+                    rel_path  = full_path.replace(input_data_root, "")
+
+                    if (not os.path.exists(full_path)):
```
jedwards4b: There are some special values of rel_path that we need to ignore - some of the component models set things like
'NULL' or 'same_as_TS' - basically if rel_path does not contain '/' (a directory tree) you can assume it's a special value and ignore it (perhaps with a warning)

scripts/Tools/check_input_data
```diff
@@ -62,91 +57,6 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
...
-                else:
-                    model = os.path.basename(data_list_file).split('.')[0]
-                    logging.warning("Model %s no file specified for %s"%(model,description))
-
-    return no_files_missing
-
-###############################################################################
 def _main_func(description):
 ###############################################################################
     if ("--test" in sys.argv):
```
jedwards4b: For some reason github won't let me annotate line 68.   But the arguments to check_input_data are wrong here - should have a case first.

utils/python/CIME/utils.py
jedwards4b: I plan to rework the logging facility a bit and will address this at that point.

jedwards4b:
Review is complete - a few minor changes needed.

utils/python/CIME/build.py
jgfouca: Done. Can't hurt.

utils/python/CIME/check_input_data.py
jgfouca: Done. Must have forgotten to run pylint on this file.

utils/python/CIME/check_input_data.py
jgfouca: Done.

PR #75, opened 2016-03-02
improved logging facility

jedwards4b:
The logging facility now includes a -s, --silent option.  
When a script is run at the --debug level it also creates a file scriptname.log

So there are now 4 levels of logging:
default (no option)
--silent
--verbose
--debug

### DIFF
--- a/scripts/Tools/standard_script_setup.py
+++ b/scripts/Tools/standard_script_setup.py
@@ -3,7 +3,8 @@ Encapsulate the importing of python utils and logging setup, things
 that every script should do.
 """
 
-import sys, os, logging, doctest, argparse
+import sys, os, logging, doctest, argparse, logging.config
+import __main__ as main
 _CIMEROOT = os.environ.get("CIMEROOT")
 if(_CIMEROOT is None):
     _CIMEROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..","..")
@@ -15,5 +16,5 @@ sys.path.append(_LIB_DIR)
 import CIME.utils
 CIME.utils.check_minimum_python_version(2, 7)
 CIME.utils.stop_buffering_output()
-
-logging.basicConfig()
+# set up logging to file
+logging.basicConfig(level=logging.INFO)
--- a/utils/python/CIME/XML/component.py
+++ b/utils/python/CIME/XML/component.py
@@ -6,6 +6,8 @@ from standard_module_setup import *
 from entry_id import EntryID
 from CIME.utils import expect, get_cime_root, get_model
 
+logger = logging.getLogger(__name__)
+
 class Component(EntryID):
     def __init__(self,infile):
         """
--- a/utils/python/CIME/system_test.py
+++ b/utils/python/CIME/system_test.py
@@ -28,6 +28,8 @@ PHASES = [INITIAL_PHASE, CREATE_NEWCASE_PHASE, XML_PHASE, SETUP_PHASE,
 CONTINUE = [TEST_PASS_STATUS, NAMELIST_FAIL_STATUS]
 
 ###############################################################################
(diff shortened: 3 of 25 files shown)


scripts/Tools/standard_script_setup.py
```diff
@@ -15,5 +16,5 @@
 import CIME.utils
 CIME.utils.check_minimum_python_version(2, 7)
 CIME.utils.stop_buffering_output()
-
-logging.basicConfig()
+# set up logging to file
+logging.basicConfig(level=logging.INFO)
```
jgfouca: Doesn't this turn on --verbose always?

utils/python/CIME/XML/component.py
```diff
@@ -6,6 +6,8 @@
 from entry_id import EntryID
 from CIME.utils import expect, get_cime_root, get_model
 
+logger = logging.getLogger(__name__)
```
jgfouca: Does this mean each .py file gets its own logger?

utils/python/CIME/system_test.py
```diff
@@ -606,12 +608,12 @@ def _consumer(self, test, test_phase, phase_method):
                      (test_phase, test, elapsed_time, status)
         if not success:
             status_str += "    Case dir: %s\n" % self._get_test_dir(test)
-        sys.stdout.write(status_str)
+        logger.info(status_str)
 
         # On batch systems, we want to immediately submit to the queue, because
         # it's very cheap to submit and will get us a better spot in line
         if (success and not self._no_run and not self._no_batch and test_phase == MODEL_BUILD_PHASE):
-            sys.stdout.write("Starting %s for test %s with %d procs\n" % (RUN_PHASE, test, 1))
```
jgfouca: Not sure I agree with this. Without --verbose, create_test won't display any info. Also, the sys.stdout.write was necessary for threaded printing. Don't know if logger is thread safe.

scripts/Tools/standard_script_setup.py
jedwards4b: I moved some stuff that we had set to info into debug, the idea being that we want the default output level to be more informational than warn.

utils/python/CIME/XML/component.py
jedwards4b: yes they do, but by default all loggers go to the same handler.  try running with the different levels to see how each is treated.

utils/python/CIME/utils.py
```diff
@@ -149,7 +150,7 @@ def run_cmd(cmd, ok_to_fail=False, input_str=None, from_dir=None, verbose=None,
     if (arg_stderr is _hack):
         arg_stderr = subprocess.PIPE
 
-    logging.info("RUN: %s" % cmd)
```
jgfouca: This does not take the verbose flag to this function into account.

utils/python/CIME/system_test.py
jedwards4b: logger is thread safe and in fact at debug level will tell you which thread printed each message.

scripts/Tools/standard_script_setup.py
jgfouca: OK, so logging.info(...) is something users will always see?

jedwards4b: By default, yes.   --silent turns it off.

utils/python/CIME/utils.py
jedwards4b: I don't understand this comment?

jgfouca: If you look above, this function takes a verbose flag. The idea is, some shell commands are so important that you always want to see "verbose" information about them. In most cases, the shell commands are not interesting and you don't want to see anything with default logging.

jgfouca:
Closed, made new PR.

PR #78, opened 2016-03-03
update call to build.case_build in Tools/case.build

sarich:
case_build previously (70fe3c0) changed the protoype for build.case_build, but
wasn't updated here

### DIFF
--- a/scripts/Tools/case.build
+++ b/scripts/Tools/case.build
@@ -58,7 +58,7 @@ def _main_func(description):
     caseroot, testmode, sharedlib_only, model_only = parse_command_line(sys.argv, description)
     logging.info("calling build.case_build with caseroot=%s and testmode=%s"
                  %(caseroot, testmode))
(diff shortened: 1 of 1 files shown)


scripts/Tools/case.build
```diff
@@ -58,7 +58,7 @@ def _main_func(description):
     caseroot, testmode, sharedlib_only, model_only = parse_command_line(sys.argv, description)
     logging.info("calling build.case_build with caseroot=%s and testmode=%s"
                  %(caseroot, testmode))
-    build.case_build(caseroot, testmode, sharedlib_only, model_only)
+    build.case_build(caseroot, None, testmode, sharedlib_only, model_only)
```
jedwards4b: Good catch, thanks.  Can you make it explicit case=None, testmode=testmode ... ?

sarich:
I have added the argument names

PR #79, opened 2016-03-04
Pythonize preview_namelists

jgfouca:
1) Convert preview_namelists to python
 - Just activate the prototype library we already had

2) Fix very sinister bug where CASEROOT was being inherited through
   the environment with the wrong value in cloned cases.
3) A couple things in template.case.run where still CESM-centric
4) Upgrades to case_diff and normalize cases to work with clones
5) Add debug logging statements to preview_namelists
6) Have build.py use preview_namelists as a library

### DIFF
--- a/cime_config/acme/machines/template.case.run
+++ b/cime_config/acme/machines/template.case.run
@@ -206,7 +206,7 @@ sub doPreRunChecks()
     $logger->info(" - To prestage restarts, untar a restart.tar file into $config{'RUNDIR'}");
 
     # Run preview namelists.. should turn this into a module at some point..
-    system("./preview_namelists -loglevel $opts{loglevel} ");
+    system("./preview_namelists");
 
     if($?)
     {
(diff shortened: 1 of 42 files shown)


scripts/Tools/preview_namelists
```diff
@@ -1,258 +1,59 @@
...
-use Cwd;
-use English;
-use Getopt::Long;
-use IO::File;
-use IO::Handle;
-use File::Basename;
-use File::Copy;
-use File::Path;
+"""
+TODO - What does this do?
```
jedwards4b: I think we can remove this?

cime_config/acme/machines/template.case.run
```diff
@@ -160,6 +160,10 @@ sub doPreRunChecks()
     }
     $config{'MPILIB'} = $mpilib;
 
+    # Necessary for NCK clones to work, must override the CASEROOT that
```
jedwards4b: DOn't you mean ERI here?

jgfouca: Yeah, brain fart.

PR #80, opened 2016-03-04
move test specific stuff into test, add comment in preview_namelists

jedwards4b:


jgfouca:
Thanks for fixing the tool description. Is there any reason not to rename this tool something like "build_namelists"?
Also, I'm not sure I agree with the main change of this PR. Why would we make it so test cases have to remember to handle CASEROOT when we can just do it in case.run and it will always work?

jedwards4b:
each of the components have a buildnml script, i think that renaming this to build_namelist would create more confusion than it resolves.    The only test that has this CASEROOT issue is ERI, to me it makes sense to isolate this special case.   I think that putting it in case.run makes it hard to explain and later someone who is not familiar with that test may deem it unnessarary and remove it.

jgfouca:
Hi Jim. I think the comment I had added should have been adequate to indicate to any future developers that the setting of CASEROOT is important and why. Taking a step back, the ERI failure was basically a case of a violated assumption/precondition. It's generally good practice to enforce or at least check your preconditions.

jedwards4b:
So maybe you want to track down where that assumption was violated and why - if the caseroot is wrong isn't there a potential that other xml is wrong for the case as well?   To put that another way, is the case object being properly updated as we move from case to case in this test?   If so why is env CASEROOT not up to date?

jgfouca:
testcase.setup adds CASEROOT to the environment. Later, in ERI, inside one of the clones, preview_namelists sets up a Case object and asks for CASEROOT. Case then uses the CASEROOT from the environment instead of the XML as it's supposed to do and then preview_namelists operates on that caseroot instead of the clone. It took my many many hours to track all this down, hence why I think we need to enforce or at least check this condition all the time.

jedwards4b:
Sorry to keep harping on this but something just isn't right and I think we should correct the root problem and not just apply this "last minute" correction

jgfouca:
That would be fine with me. What do you have in mind?

jedwards4b:
when preview_namelists gets it’s CASEROOT from the clone is it correct?

I don’t see where this is happening do you > Case then uses the CASEROOT from the environment instead of the XML?
 
when I look at case.run it’s not looking in the envronment for CASEROOT

jedwards4b:
Superceeded by #82

Issue #81, opened 2016-03-08
Failed to build A case with create_newcase on edison


rljacob:
Steps:

Checkout out latest master dfb98f064198eb8b2ba590d40844dd83bdc7711b
setenv CIME_MODEL acme (OR cesm)
module load python
create_newcase -case Acase -res f45_g37_rx1 -compset A -mach edison -compiler intel -project acme
cd Acase
./case.setup

Everything above works.  Then I tried the next step...

edison09 scripts/Acase> ./case.build 
calling build.case_build with caseroot=/global/u2/r/rjacob/cimeESMCI/scripts/Acase and testmode=False
sharedlib_only is False
model_only is False
Not able to retreive value for item 'TESTCASE'
Loading input file: 'Buildconf/datm.input_data_list'
Loading input file: 'Buildconf/dlnd.input_data_list'
Loading input file: 'Buildconf/dice.input_data_list'
Loading input file: 'Buildconf/docn.input_data_list'
Loading input file: 'Buildconf/drof.input_data_list'
Loading input file: 'Buildconf/cpl.input_data_list'
Not able to retreive value for item 'CISM_USE_TRILINOS'
Not able to retreive value for item 'CLM_CONFIG_OPTS'
Not able to retreive value for item 'CAM_CONFIG_OPTS'
Not able to retreive value for item 'OCN_SUBMODEL'
ERROR: Could not find cmd_path for lang 'python'

rljacob:
Fixed by merge of PR #84

rljacob:
Tried latest master and get a new error wile running ./case.setup

edison08 scripts/Acase> ./case.setup
Creating Macros file for edison
Checking file env_case
Creating batch script case.run
Creating batch script case.st_archive
Creating batch script case.lt_archive
Locking file env_mach_pes.xml 
Creating user_nl_xxx files for components and cpl
Running preview_namelist script with loglevel INFO
Not able to fully resolve item '$EXEROOT/lib'
Not able to fully resolve item '$EXEROOT/lib/include'
Not able to retreive value for item 'TESTCASE'
See ./CaseDoc for component namelists 
If an old case build already exists, might want to run case.clean_build before building 

Not being able to resolve EXEROOT leads to bigger problems in case.build.

PR #84, opened 2016-03-08
port to edison, fix order of args

jedwards4b:


jedwards4b:
Resolves #81

jgfouca:
Looks good. Wondering how the argument order to build_checks wasn't breaking things.

rljacob:
Note:  You have to put "Resolves #81" in the commit message for the issue to close automatically.

PR #88, opened 2016-03-09
Mvclmbldtry2

jedwards4b:
mv clm build into build_libraries
prepare to move tests into python by looking for TESTNAME_script and
using the python case.test template if its not found.

Instead of having case.test_build and case.build in the case directory and
throwing an error if the user trys case.build, just replace case.build with case.test_build


utils/python/CIME/build.py
```diff
@@ -166,10 +166,6 @@ def case_build(caseroot, case=None, testmode=False, sharedlib_only=False, model_
     case = Case() if case is None else case
     testcase = case.get_value("TESTCASE")
     cimeroot = case.get_value("CIMEROOT")
```
jgfouca: case_build still takes testmode as an optional argument. Does it still need it?

utils/python/CIME/case.py
```diff
@@ -68,18 +68,20 @@ def get_value(self, item, attribute={}, resolved=True, subgroup=None):
         logging.info("Not able to retreive value for item '%s'" % item)
 
 
-    def get_resolved_value(self, item):
-        # TODO HACK - surely there is a better way?
```
jgfouca: What was the idea here?

utils/python/CIME/case.py
```diff
@@ -68,18 +68,20 @@ def get_value(self, item, attribute={}, resolved=True, subgroup=None):
...
+        if (num_unresolved > 0 and recurse < recurse_limit ):
             for env_file in self._env_entryid_files:
                 result = env_file.get_resolved_value(item)
-                if (result.count("$") < num_unresolved):
-                    num_unresolved = result.count("$")
-                    item = result
-                    if ("$" not in item):
-                        return item
-
+                num_unresolved = result.count("$")
```
jgfouca: This assignment of num_unresolved doesn't do anything.

scripts/Tools/case.test_build
```diff
@@ -17,6 +17,7 @@ from CIME.SystemTests.ncr import NCR
 from CIME.SystemTests.noc import NOC
 from CIME.SystemTests.pea import PEA
```
jgfouca: Shouldn't this file have been deleted?

jgfouca:
Review complete.

scripts/Tools/case.test_build
jedwards4b: I didn't replace case.build with case.test_build in the repo, only in the case.   We may be able to merge them as a second step, let me think about it a bit.

utils/python/CIME/build.py
jedwards4b: I think that it doesn't.

utils/python/CIME/case.py
jedwards4b: good point

utils/python/CIME/case.py
jgfouca: Still confused by this change.

jedwards4b: So this change is so that when there are recursive unresolved values as in for example EXEROOT=$SCRATCHROOT/$CASE/bld and SCRATCHROOT=/tmp/$USER  the search includes all of the xml files in case and not just the one pointed to by the current env_file object.

PR #89, opened 2016-03-09
use explicit paths, recursion in get_resolved_value

jedwards4b:
Updated to correct history issues.

### DIFF
--- a/utils/python/CIME/case.py
+++ b/utils/python/CIME/case.py
@@ -68,18 +68,20 @@ class Case(object):
         logging.info("Not able to retreive value for item '%s'" % item)
 
 
-    def get_resolved_value(self, item):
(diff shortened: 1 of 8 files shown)


utils/python/CIME/case.py
```diff
@@ -68,18 +68,20 @@ def get_value(self, item, attribute={}, resolved=True, subgroup=None):
         logging.info("Not able to retreive value for item '%s'" % item)
 
 
-    def get_resolved_value(self, item):
```
jgfouca: This may conflict with what I just pushed. Please rewrite history to remove any changes to this file.

jgfouca:
Review complete. Will merge once case.py is reverted.

jedwards4b:
Included in #88

PR #90, opened 2016-03-10
Take a close look at CIME/XML

jgfouca:
Preparing for type-aware get_value.

1) Formatting changes all over to meet agreed-upon python coding convention
2) Split get_node into get_nodes, get_node, get_optional_node
3) Only accept string ids or XML elements, never both
4) Machines refactor: call super class for 'normal' get_node behavior, call Machine.get_node for machine-specific behavior

### DIFF
--- a/utils/python/CIME/XML/entry_id.py
+++ b/utils/python/CIME/XML/entry_id.py
@@ -19,7 +19,14 @@ class EntryID(GenericXML):
         """
         Set the value of an entry to the default value for that entry
         """
-        # Need comment here
+        # Hangle this case:
+        # <entry id ...>
+        #  <values>
+        #   <value A="a1">X</value>
+        #   <value A="a2">Y</value>
+        #   <value A="a3">Z</value>
+        #  </values>
+        # </entry>
         valnodes = self.get_nodes("value", root=node)
         for valnode in valnodes:
             for att in valnode.attributes:
--- a/utils/python/CIME/XML/entry_id.py
+++ b/utils/python/CIME/XML/entry_id.py
@@ -15,92 +15,66 @@ class EntryID(GenericXML):
         GenericXML.__init__(self, infile)
         self.groups={}
 
-    def set_default_value(self, vid, attributes=None):
+    def set_default_value(self, node, attributes=None):
         """
         Set the value of an entry to the default value for that entry
-        vid can be an xml node pointer or a string identifier of a node
         """
-        value = None
-        if (type(vid) != type(str())):
-            node = vid
-            vid = node.attrib["id"]
-        else:
-            nodes = self.get_node("entry", {"id":vid})
-            if (nodes is None):
-                return
-            expect(len(nodes) == 1, "More than one match found for id " + vid)
-            node = nodes[0]
+        # Need comment here
+        valnodes = self.get_nodes("value", root=node)
+        for valnode in valnodes:
+            for att in valnode.attributes:
+                if att.key in attributes:
+                    if re.search(attributes[att.key], att.text):
+                        node.set("value", valnode.text)
+                        return valnode.text
+                        logger.info("id %s value %s" % (vid, valnode.text))
 
-        valnodes = self.get_node("value",root=node)
-        if (valnodes is not None):
-            for valnode in valnodes:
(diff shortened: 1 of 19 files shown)


utils/python/CIME/XML/entry_id.py
```diff
@@ -15,92 +15,66 @@ def __init__(self, infile=None):
...
-            return value[0].text
-
-    def set_value(self, vid, value,subgroup=None):
+        # Need comment here
+        valnodes = self.get_nodes("value", root=node)
+        for valnode in valnodes:
+            for att in valnode.attributes:
+                if att.key in attributes:
+                    if re.search(attributes[att.key], att.text):
+                        node.set("value", valnode.text)
```
jgfouca: Code above did not set value in this loop. Was that a mistake or intentional?

utils/python/CIME/XML/generic_xml.py
```diff
@@ -15,13 +15,16 @@ def __init__(self, infile=None):
         Initialize an object
         """
         self.tree = None
+
+        # Hold arbitary values? Why?
```
jgfouca: This self.lookups member has always confused me, can you explain?

utils/python/CIME/XML/generic_xml.py
```diff
@@ -30,47 +33,65 @@ def __init__(self, infile=None):
...
-        logger.debug("write: "+ outfile)
+
+        logger.debug("write: " + outfile)
         xmlstr = ET.tostring(self.root)
         doc = minidom.parseString(xmlstr)
         with open(outfile,'w') as xmlout:
             doc.writexml(xmlout,addindent='  ')
 
-
     def get_node(self, nodename, attributes=None, root=None):
```
jgfouca: I was frequently confused by get_node since the name implies one, but it could actually return 0, 1, or N. This required the client to do an error-check after every call to ensure they got back the number they expected and in many cases, we forgot to do that. With this change, the API makes it explicit how what we expect and so relieves client from lots of checking and improves readability.

utils/python/CIME/XML/headers.py
```diff
@@ -18,14 +18,12 @@ def __init__(self,infile=None):
...
-        if(infile is None):
+        if infile is None:
             files = Files()
-            infile = files.get_value('CASEFILE_HEADERS',resolved=True)
-        EntryID.__init__(self,infile)
+            infile = files.get_value('CASEFILE_HEADERS', resolved=True)
+        EntryID.__init__(self, infile)
 
     def get_header_node(self,fname):
-        fnode=self.get_node("file", attributes={"name":fname})
```
jgfouca: A good example of how the new API makes things easier.

utils/python/CIME/XML/machines.py
```diff
@@ -64,28 +72,23 @@ def probe_machine_name(self):
         """
         machine = None
         nametomatch = socket.gethostname().split(".")[0]
-        nodes = self.get_node("machine")
+        nodes = self.get_nodes("machine")
 
         for node in nodes:
             machtocheck = node.get("MACH")
             logger.debug("machine is " + machtocheck)
-            self.set_machine(machtocheck)
```
jgfouca: This set_machine looked like a dangerous and unexpected side-effect of this method.

utils/python/CIME/XML/entry_id.py
jedwards4b: mistake -  It didn't affect us since we aren't creating the case xml files here yet.

utils/python/CIME/XML/generic_xml.py
jedwards4b: In create_newcase we may set values for xml files that haven't been created yet.   We need a place to store them 
until we are ready to create the file.  At file creation we get the values for those fields from this lookup table and then remove the entry.    This was what I came up with in the perl anyway and I think that we still need it here.

jedwards4b:
Review complete, thanks for the cleanup - I don't see anything of concern - please merge when you are ready.

utils/python/CIME/XML/generic_xml.py
jgfouca: Thanks, I will add comment.

PR #92, opened 2016-03-10
Update cseg master

jedwards4b:
Update ESCMI with with changes from cime4.4.3 to cime4.4.7 from CESM-Development fork.

cime_config/cesm:
  cori queue settings
  machines files for constance
  add T85_T85 to cesm grids
  update hobart pe layout for all active f19_g16
  update config_batch.xml to remove pbs directives
  Update compsets and grids for single column pop capability.

data_comps and csm_share:
 Add C1DIAF compset / update datm streams
 Use single column datm forcing for 1D_1D res
 Split 6-hourly, daily SOFS and NCEP streams in to multiple files.

driver_cpl and csm_share:
  add BARRIER_OPTION to driver and new logic for opening and
    closing files in the streams input to improve performance.
  Add correct value of ATM_NCPL for T42 grid.
  Modify seq_io_mod.F90 so an avect field named "hgt" is never
    written to the history file.  This is a temporary workaround
    until models initialize the hgt field in the domain datatype.
  Add in drv_flds namelist definition
  address issue #386 with this new variable

testing:
  Update SMS and PFS to ouput test functionality summary to the caseroot TestStatus file.
  fix testreporter.pl to output true test functionality summar, 
  Update testreporter.pl to report teststatus as overall runtime status when
    it is DONE even if test functionality reports FAIL due to namelist comparison.
  Add testlist.xsd schema
  Remove ExpectedFails.xml file from CIME and update testreporter.pl
    in favor of having all-active expected test failures managed
    from the testdb. Components may still choose to work with an
    expected fails file for their stand-alone tests.

tools:
  new treatment of user_compset, bug fix for make clean
  Changed logger calls to prints in check_input_data to make sure the output capture in checkinputdata
     actually works
  correct syntax in clean_build, do not clean shared land library in a test
  fix clean build issue in Makefile bug in xmlchange
  create_newcase

timing tests and BatchUtils.pm:
  Allow a different RESUBMIT xml variable for use with timing tests
  improve documentation of COMP_RUN_BARRIERS variable
  A new xml variable was added, RESUBMIT_SETS_CONTINUE_RUN, to reduce ambiguity  and unexpected side effects in 
   the use of COMP_RUN_BARRIERS.   RESUBMIT_SETS_CONTINUE_RUN defaults to true, a setting of false will cause 
   subsequent runs using the RESUBMIT feature to simply repeat the previous run.   

update cime for pop single column and strm4d capabilities
Add setting of BARRIER for I compsets
correct comparison error
Change name of Solar for GSWP3 data
improve documentation of COMP_RUN_BARRIERS variable


rljacob:
I'm trying to make a better commit message that summarizes what all is in this.  Can you check the above and add what's missing?

rljacob:
Figured out which tag version this was adding by doing the merge and using git log to see what tags were new.

rljacob:
Passed  ./scripts_regression_tests --fast on edison after merging to master.

PR #95, opened 2016-03-11
Xmlbugfixes

jedwards4b:
Fix some issues with xml data typing.  Correct order of operations in setting machine and compiler for tests.

### DIFF
--- a/utils/python/CIME/XML/entry_id.py
+++ b/utils/python/CIME/XML/entry_id.py
@@ -114,6 +114,7 @@ class EntryID(GenericXML):
             return
 
         type_str = self._get_type_info(node)
+        logger.debug("vid %s type %s"%(vid,type_str))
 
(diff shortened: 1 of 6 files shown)


utils/python/CIME/XML/entry_id.py
```diff
@@ -71,6 +71,24 @@ def set_value(self, vid, value, subgroup=None):
         else:
             return None
 
+    def _set_value(self, node, vid, value, subgroup=None):
```
jgfouca: Looks like we have duplicates of these set_value methods. I'm going to assume the news ones are the right versions.

jgfouca:
Thanks for finding all these problems, sorry so many got through. Clearly there's some holes in our testing if all these got through scripts_regression_tests.

jgfouca:
I made one commit ontop of this branch. I'm going to PR my branch and close this one.

PR #98, opened 2016-03-14
Change get timing to python

sarich:


scripts/Tools/getTiming
```diff
@@ -1,669 +1,620 @@
...
-}
-
-
+#!/usr/bin/env python
+"""
+Get timing information from run
+"""
+
+from standard_script_setup import *
+import argparse, sys, os, logging
```
jedwards4b: These are already imported in standard_script_setup, no need to import them again.

scripts/Tools/getTiming
```diff
@@ -1,669 +1,620 @@
...
+
+class TimingParser:
+    def __init__(self, caseroot, lid="999999-999999"):
+        self.caseroot = caseroot
+        self.lid = lid
+        self.fin = None
+        self.fout = None
+        self.adays=0
+        self.models = {}
+        for s in ["ATM", "LND", "ROF", "WAV", "ICE", "OCN", "GLC", "CPL"]:
```
jedwards4b: These generic component names are defined in cime/driver_cpl/cime_config/config_component.xml 
and are available through the CIME.XML.component module method get_value('component")

scripts/Tools/getTiming
```diff
@@ -1,669 +1,620 @@
...
+                maxval = float(m.groups()[0])
+                minval = float(m.groups()[1])
+                found = True
+                return (minval, maxval, found)
+        return (0, 0, False)
+
+
+
+    def getTiming(self):
+        atm = self.models['ATM']
```
jedwards4b: get components from xml instead of hardcoding

scripts/Tools/getTiming
```diff
@@ -1,669 +1,620 @@
...
+        ice = self.models['ICE']
+        ocn = self.models['OCN']
+        glc = self.models['GLC']
+        cpl = self.models['CPL']
+        case = Case(self.caseroot)
+        caseid = case.get_value("CASE")
+        mach = case.get_value("MACH")
+        user = case.get_value("USER")
+        continue_run = str(case.get_value("CONTINUE_RUN")).upper()
+        rundir = case.get_value("RUNDIR")
```
jedwards4b: On the current master these variables are cast to the correct type, no need to convert them here.

jedwards4b:
Review complete - please get components (ATM, LND, ROF, ...) from components module 
and get rid of type casting as get_value now already returns the correct type.   Otherwise it looks fine.

jedwards4b:
I notice that the master has a bug in component.py at line 25 comp.text() should be comp.text

sarich:
The components listed in the xml file has "DRV" in the list instead of
"CPL". I can replace this manually, but there are also some parts of the
output where the component information is printed out in different orders,
which would require hard-coding an ordered list anyway. If we are trying to
be flexible in case the models or their abbreviations change, there would
need to be substantial changes to the script.

The type values for *_NCPL are char. I was assuming from the leading 'N'
that this should be an integer, but is a floating point value possible? The
other casts can be removed easily.

PR #109, opened 2016-03-18
rewrote manage_case in python

mvertens:
Rewrote manage_case in python. This required also adding the following new files that primarily are being used to write the output of the query commands.
utils/python/CIME/XML/compsets.py
utils/python/CIME/XML/grids.py

### DIFF
--- a/scripts/manage_case
+++ b/scripts/manage_case
@@ -1,375 +1,172 @@
-#!/usr/bin/env perl 
+#!/usr/bin/env python2.7
+#FIXME - the above needs the explicit python2.7 - this is not general ???
 
-#-----------------------------------------------------------------------------------------------
-#
-# manage_compsets
-#
-# This utility allows the users to manage and query model component sets  
-#
-#-----------------------------------------------------------------------------------------------
+from Tools.standard_script_setup import *
 
-use XML::LibXML;
-use IO::File;
-use Data::Dumper;
-use Getopt::Long;
-use POSIX qw(strftime);
-use Cwd qw(getcwd abs_path chdir);
-use File::Basename;
-use strict;
+from CIME.utils         import expect, get_model
+from CIME.XML.files     import Files
+from CIME.XML.component import Component
+from CIME.XML.compsets  import Compsets
+from CIME.XML.grids     import Grids
+from CIME.XML.machines  import Machines
 
-# Check for the existence of XML::LibXML in whatever perl distribution happens to be in use.  
-# If not found, print a warning message then exit. 
-eval {
-    require XML::LibXML;
-    XML::LibXML->import();
-};
-if($@)
-{
-    my $warning = <<END;
-WARNING:
-    The perl module XML::LibXML is needed for XML parsing in the CESM script system.  
-	Please contact your local systems administrators or IT staff and have them install it for 
-	you, or install the module locally.  
+import argparse, doctest
 
-END
-    print "$warning\n";
-    exit(1);
-}
+###############################################################################
+def query_grids():  
+###############################################################################
 
-#-----------------------------------------------------------------------------------------------
-# Setting autoflush (an IO::Handle method) on STDOUT helps in debugging.  It forces the test
-# descriptions to be printed to STDOUT before the error messages start.
-*STDOUT->autoflush();                  
+    files = Files()
+    config_file = files.get_value("GRIDS_SPEC_FILE", resolved=True)
+    expect(os.path.isfile(config_file),
+           "Cannot find config_file %s on disk" %config_file)
 
-#-----------------------------------------------------------------------------------------------
-# Set the directory that contains the CESM configuration scripts.  If the create_newcase command was
-# issued using a relative or absolute path, that path is in $ProgDir.  Otherwise assume the
-# command was issued from the current working directory.
+    parsed_xml = Grids(config_file)
+    if verbose:
+        parsed_xml.print_values(verbose=verbose)
+    else:
+        parsed_xml.print_values()
 
-(my $ProgName = $0) =~ s!(.*)/!!;      # name of this script
-my $ProgDir = $1;                      # name of directory containing this script -- may be a
-                                       # relative or absolute path, or null if the script is in
-                                       # the user's PATH
-my $cwd = getcwd();                    # current working directory
-my $cfgdir;                            # absolute pathname of directory that contains this script
-if ($ProgDir) { 
-    $cfgdir = abs_path($ProgDir);
-} else {
-    $cfgdir = $cwd;
-}
+###############################################################################
+def query_machines():  
+###############################################################################
 
-#-----------------------------------------------------------------------------------------------
-if ($#ARGV == -1) {
-    usage();
-}
-#-----------------------------------------------------------------------------------------------
+    files = Files()
+    config_file = files.get_value("MACHINES_SPEC_FILE", resolved=True)
+    expect(os.path.isfile(config_file),
+           "Cannot find config_file %s on disk" %config_file)
 
-sub usage {
-    die <<EOF;
+    parsed_xml = Machines(config_file)
+    if verbose:
+        parsed_xml.print_values(verbose=verbose)
+    else:
+        parsed_xml.print_values()
 
-SYNOPSIS
-  manage_case -model [options] -query  [options]
+###############################################################################
+def query_compsets(name):  
+###############################################################################
 
-OPTIONS
-  User supplied values are denoted in angle brackets (<>).  Any value that contains
-  white-space must be quoted.  Long option names may be supplied with either single
-  or double leading dashes.  A consequence of this is that single letter options may
-  NOT be bundled.
+    # Determine valid component values by checking the value attributes for COMPSETS_SPEC_FILE
+    files = Files()
+    components = files.get_components("COMPSETS_SPEC_FILE", resolved=True)
+    match_found = "unset"
+    for component in components:
+        if component == name:
+            match_found = component
+            break
 
-  -model [cesm, acme]
-  -query compsets -setby [cam, clm, cice, cism, pop, allactive] (if model is CESM)
-  -query component -name [cam, clm, cice, cism, pop, datm, dlnd, dice, dglc, docn] (if model is CESM)
-  -query grids
-  -query machines
+    # If name is not a valid argument - exit with error
+    expect(match_found != "unset", 
+           "Invalid input argument %s, valid input arguments are %s" % (name, components))
 
-EOF
-}
+    # Determine the config_file for the target component 
+    config_file = files.get_value("COMPSETS_SPEC_FILE", attribute={"component":name}, resolved=True)
+    expect((config_file),
+           "Cannot find any config_component.xml file for %s" %name)
 
-#-----------------------------------------------------------------------------------------------
-# Save commandline
-my $commandline = "create_newcase @ARGV";
+    # Check that file exists on disk
+    expect(os.path.isfile(config_file),
+           "Cannot find config_file %s on disk" %config_file)
 
-#-----------------------------------------------------------------------------------------------
-# Parse command-line options.
-my %opts = ( );
-GetOptions(
-    "h|help"            => \$opts{'help'},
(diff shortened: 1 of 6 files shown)


scripts/manage_case
```diff
@@ -1,375 +1,172 @@
...
-    my $xml_machine = XML::LibXML->new( no_blanks => 1)->parse_file($file);
-    foreach my $machine ($xml_machine->findnodes(".//machine")) {
-	my $name = $machine->getAttribute('MACH');
-	foreach my $node ($machine->findnodes("./DESC")) {
-	    my $desc = $node->textContent();
-	    print " $name :  \n      $desc \n";
-	}
-    }
-}
+#!/usr/bin/env python2.7
```
jedwards4b: I think that you just want python2 here not 2.7

scripts/manage_case
```diff
@@ -1,375 +1,172 @@
...
+from CIME.XML.machines  import Machines
+
+import argparse, doctest
+
+###############################################################################
+def query_grids():  
+###############################################################################
+
+    files = Files()
+    config_file = files.get_value("GRIDS_SPEC_FILE", resolved=True)
```
jedwards4b: resolved defaults to True so no need to have that argument here (it doesn't hurt anything to have it either)

scripts/manage_case
```diff
@@ -1,375 +1,172 @@
...
+
+###############################################################################
+def query_compsets(name):  
+###############################################################################
+
+    # Determine valid component values by checking the value attributes for COMPSETS_SPEC_FILE
+    files = Files()
+    components = files.get_components("COMPSETS_SPEC_FILE", resolved=True)
+    match_found = "unset"
+    for component in components:
```
jedwards4b: I think that this can be reduced to 
  if name in components:
       match_found = name

scripts/manage_case
mvertens: Yes - that is why the comment. I could not get it to work on my mac without this. Should I put in another pull request with this fix?

scripts/manage_case
```diff
@@ -1,375 +1,172 @@
...
+    expect((config_file),
+           "Cannot find any config_component.xml file for %s" %name)
+
+    # Check that file exists on disk
+    expect(os.path.isfile(config_file),
+           "Cannot find config_file %s on disk" %config_file)
+
+    # Now parse the compsets file and write out the compset alias and longname as well as the help text
+    # determine component xml content 
+    parsed_xml = Compsets(config_file)
```
jedwards4b: I think that you shouldn't resuse the same name (parsed_xml) for different xml objects - call each object what it is
(compset, machine, etc)

utils/python/CIME/XML/component.py
```diff
@@ -22,7 +22,31 @@ def get_value(self, name, attribute={}, resolved=False):
...
+            rootnode = self.get_node("help")
+            helptext = rootnode.text
+            return helptext
+        elif name == "description":
+            rootnode = self.get_node("description")
+            compsets = {}
+            descs = self.get_nodes("desc", root=rootnode)
+            for desc in descs:
+                attrib = desc.attrib["compset"]
+                text = desc.text
```
jedwards4b: why not
compsets[desc.attrib["compset"]] = desc.text

utils/python/CIME/XML/files.py
```diff
@@ -20,3 +20,12 @@ def __init__(self):
         """
         infile = os.path.join(get_cime_root(), "cime_config", get_model(), "config_files.xml")
         EntryID.__init__(self, infile)
+
+    def get_components(self, nodename, resolved=True):        
+        node = self.get_optional_node("entry", {"id":nodename})
+        valnodes = self.get_nodes("value", root=node)
+        values = list();
+        for valnode in valnodes:
+            value = valnode.attrib["component"]
```
jedwards4b: values.append(valnode.attrib["component"])

utils/python/CIME/XML/grids.py
```diff
@@ -0,0 +1,99 @@
...
+                domain_names = gridRE.split(lname.text)[1:]
+                domain_set = list(set(domain_names))
+                for domain in domain_set:
+                    print "   domain is ",domain
+                    domain_node = self.get_node(nodename="domain", attributes={"name":domain}) 
+                    for child in domain_node:
+                        print  "     ",child.tag,": ",child.text
+                    
+                # in verbose mode add grid maps 
+                # FIXME (mvertens, 3/2016) - how do I make the following more consise and less hard-wired
```
jedwards4b: You should get the components from driver_cpl/cime_config/config_component.xml
using  component.get_value("component")  then create dicts like grid["atm"], grid["lnd"] etc

jedwards4b:
Review complete.

scripts/manage_case
jgfouca: @mvertens , no that's way too much work. Just make additional commits to your branch and push.

jedwards4b: You can just update this PR by pushing to your fork.

scripts/manage_case
```diff
@@ -1,375 +1,172 @@
...
+def query_grids():  
+###############################################################################
+
+    files = Files()
+    config_file = files.get_value("GRIDS_SPEC_FILE", resolved=True)
+    expect(os.path.isfile(config_file),
+           "Cannot find config_file %s on disk" %config_file)
+
+    parsed_xml = Grids(config_file)
+    if verbose:
```
jgfouca: You should probably use the logger stuff that Jim set up. Also, where does verbose come from in this function? I don't see it as an argument or stack variable. You may want to run pylint on this code to find mistakes like this.

scripts/manage_case
```diff
@@ -1,375 +1,172 @@
...
+def query_machines():  
+###############################################################################
+
+    files = Files()
+    config_file = files.get_value("MACHINES_SPEC_FILE", resolved=True)
+    expect(os.path.isfile(config_file),
+           "Cannot find config_file %s on disk" %config_file)
+
+    parsed_xml = Machines(config_file)
+    if verbose:
```
jgfouca: See prior comment.

scripts/manage_case
```diff
@@ -1,375 +1,172 @@
...
+    comps_array = comps_object.get_value("component")
+    
+    # Loop through the elements for each component class (in config_files.xml) 
+    # and see if there is a match for the the target component in the component attribute 
+    match_found = False
+    valid_components = list()
+    for comp in comps_array:
+        string = "CONFIG_%s_FILE"%comp 
+
+        # determine all components in string
```
jgfouca: As Jim said, please get rid of all these resolve=True.

scripts/manage_case
```diff
@@ -1,375 +1,172 @@
...
+        parsed_xml.print_values()
+
+###############################################################################
+def query_compsets(name):  
+###############################################################################
+
+    # Determine valid component values by checking the value attributes for COMPSETS_SPEC_FILE
+    files = Files()
+    components = files.get_components("COMPSETS_SPEC_FILE", resolved=True)
+    match_found = "unset"
```
jgfouca: "unset" is a very non-pythonic choice to indicate not found. Use python's None instead.

scripts/manage_case
```diff
@@ -1,375 +1,172 @@
...
+        # determine if config_file is on disk
+        config_file = files.get_value(string, attribute={"component":name}, resolved=True)
+        if config_file is not None:
+            match_found = True
+            expect(os.path.isfile(config_file),
+                   "Cannot find config_file %s on disk" %config_file)
+            break
+
+    # If name is not a valid argument - exit with error
+    expect((match_found),
```
jgfouca: No need for parens around match_found.

scripts/manage_case
```diff
@@ -1,375 +1,172 @@
...
+def parse_command_line(args):
+###############################################################################
+
+    cime_model = CIME.utils.get_model()
+
+    parser = argparse.ArgumentParser()
+
+    CIME.utils.setup_standard_logging_options(parser)
+
+    parser.add_argument("--query_compsets_setby",
```
jgfouca: It is not POSIX compliant to have --long_options_like_this. You want --long-options-like-this.

scripts/manage_case
```diff
@@ -1,375 +1,172 @@
...
+
+    args = parser.parse_args()
+
+    CIME.utils.handle_standard_logging_options(args)
+
+    return args
+
+###############################################################################
+
+args = parse_command_line(sys.argv)
```
jgfouca: It is generally not a good idea to do work in the global namespace. Please follow the convention we've used in our other scripts:
`def _main_func():
    *do stuff*

if __name__ == "__main__":
    _main_func()`

utils/python/CIME/XML/component.py
```diff
@@ -22,7 +22,31 @@ def get_value(self, name, attribute={}, resolved=False):
             compnode = self.get_node("components")
```
jgfouca: This function confuses me and could use a docstring to explain. The return value can be a string, list, or dictionary depending on name? I'd rather have list, dictionary etc as a <type> than having special cases everywhere.

utils/python/CIME/XML/grids.py
```diff
@@ -0,0 +1,99 @@
...
+            gridmap_nodes = self.get_nodes(nodename="gridmap")
+            gridRE = re.compile(r"[_]{0,1}[a-z]{1,2}%")
+
+        # write out grid elements 
+        grid_nodes = self.get_nodes(nodename="grid")
+        for grid_node in grid_nodes:
+            lname   = grid_node.find("lname")
+            sname   = grid_node.find("sname")
+            support = grid_node.find("support")
+            alias   = grid_node.find("alias")
```
jgfouca: @jedwards4b can correct me if I'm wrong, but I thought we wanted the logger to handle output and debug outout?

utils/python/CIME/XML/grids.py
```diff
@@ -0,0 +1,99 @@
...
+                # in verbose mode add grid maps 
+                # FIXME (mvertens, 3/2016) - how do I make the following more consise and less hard-wired 
+                atm_grid  = domain_names[0]
+                lnd_grid  = domain_names[1]
+                ocn_grid  = domain_names[2]
+                rof_grid  = domain_names[3]
+                mask_grid = domain_names[4]
+                glc_grid  = domain_names[5]
+                wav_grid  = domain_names[6]
+
```
jgfouca: grids = [ ("atm_grid", atm_grid), "lnd_grid", lnd_grid), ...]
    for idx, grid in enumerate(grids):
        for other_grid in grids[idx+1:]:
            nodes = self.get_nodes(nodename="gridmap", attributes={grid[0]:grid[1], other_grid[0]:other_grid[1]})
            for gridmap_node in nodes:
                for child in gridmap_node:
                    print " ", child.tag, ": ", child.text

jgfouca:
Review complete.

utils/python/CIME/XML/component.py
mvertens: could you please clarify in the example above what you mean...

utils/python/CIME/XML/files.py
mvertens: could you also clarify in the example above what you mean - it would be really helpful

scripts/manage_case
mvertens: Done

scripts/manage_case
mvertens: Done

scripts/manage_case
mvertens: Done

scripts/manage_case
mvertens: Done

utils/python/CIME/XML/component.py
mvertens: I think my may is clearer - so I'll keep it that way for now. We agreed on this.
Done

jgfouca:
@mvertens , is this ready?

mvertens:
Yes - I pushed this to my branch that I had issued a pull request for - and
I thought this would automatically update you and Jim.

jgfouca:
@mvertens Looks much better! Looks like you addressed every comment.

PR #114, opened 2016-03-18
allow xmlchange to set one variable to another with --force option

jedwards4b:
You should be able to set an xml variable to the value of another using xmlchange
This PR implements that functionality by adding a --force option to the xmlchange command
and providing a code path that ignores the strong typing rules.

### DIFF
--- a/scripts/Tools/xmlchange
+++ b/scripts/Tools/xmlchange
@@ -21,7 +21,7 @@ import argparse, doctest, shutil, glob
 def parse_command_line(args, description):
 ###############################################################################
     parser = argparse.ArgumentParser(
-        usage="""\n%s [<changeargs>] [--verbose][--file file][--id id][--val value][--noecho][--append][--warn]
+        usage="""\n%s [<changeargs>] [--verbose][--file file][--id id][--val value][--noecho][--append][--warn][--force]
 OR
 %s --help
 OR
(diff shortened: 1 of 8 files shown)


scripts/Tools/xmlchange
```diff
@@ -89,13 +92,15 @@ def xmlchange(caseroot, listofsettings=None, xmlfile=None, xmlid=None, xmlval=No
             if(append is True):
                 value = case.get_value(xmlid, resolved=False, subgroup=subgroup)
                 xmlval = "%s %s" %(xmlval, value)
-            case.set_value(xmlid, convert_to_type(xmlval, type_str, xmlid), subgroup)
+            xmlval = convert_to_type(xmlval, type_str, xmlid, ok_to_fail=force)
```
jgfouca: If we're going to force it to use a string, why even call convert_to_type?

utils/python/CIME/utils.py
```diff
@@ -595,42 +595,57 @@ def convert_to_type(value, type_str, vid=""):
...
+                if not ok_to_fail:
+                    expect(False, "Entry %s was listed as type real but value '%s' is not valid real" % (vid, value))
 
         else:
             expect(False, "Unknown type '%s'" % type_str)
 
     return value
 
-def convert_to_string(value, type_str, vid=""):
+def convert_to_string(value, type_str=None, vid=""):
```
jgfouca: Is anyone using type_str=None? I don't see how that's needed for this change. Also, if we're just going to return value as-is if type_str is None, why even call the function in the first place?

jgfouca:
I don't think any changes are needed to convert_to_type or convert_to_string. Just bypass them if you need to --force.

jgfouca:
I see new commits, is this ready?

PR #116, opened 2016-03-22
read in soft keys in python the same way it is done in perl

sarich:
The current cime4 perl implementation of soft modules involve creating a temporary sh or csh script that sources the init_path file and then adds each soft key using the cmd_path value (`soft`). This script is then run and the new environment compared against the old environment. This is incompatible with the python implementation I initially wrote which runs successive cmd_path (`/soft/softenv/1.6.2/bin/soft-dec sh`) commands (soft-dec prints out the commands used to set the environment) and parses the output to set the environment variables.

The python scripts worked fine, but the perl scripts did not set the soft keys correctly would cause the user's default soft keys to be used.

Now both perl and python accept the `soft` command as the cmd_path value.


jgfouca:
Looks good to me.

PR #117, opened 2016-03-23
Fix #105 - Python version of xmlquerry

### LINKED ISSUE #105: Rewrite xmlquery in python

wilke:
Soliciting feedback


cime_config/config_headers.xml
```diff
@@ -28,6 +28,12 @@
     </header>
   </file>
 
+  <file name ="env_mach_specific.xml">
+    <header>
+    These variables should notbe changed.
```
jgfouca: Typo: "notbe" -> "not be"

scripts-python/xmlquerry
```diff
@@ -0,0 +1,182 @@
...
+import argparse, doctest, shutil, glob
+import xml.etree.ElementTree as ET
+import textwrap
+
+
+###############################################################################
+def parse_command_line(args, description):
+###############################################################################
+
+    # my %opts = (
```
jgfouca: Please delete.

scripts-python/xmlquerry
```diff
@@ -0,0 +1,182 @@
...
+    #     file =>0,
+    #     noexpandxml            =>0,
+    #     help        =>0,
+    #     listall        =>0,
+    #     loglevel        =>"INFO",
+    #     );
+
+
+    
+    epilog="""\
```
jgfouca: Please put this in a docstring at the top of the file. You can then refer to this string with __doc__. You can see create_test as an example.

scripts-python/xmlquerry
```diff
@@ -0,0 +1,182 @@
...
+    if( len(args.attributes) >= 1) :
+        args.attribute =  args.attribute + args.attributes   
+    
+    # remove duplicate names
+    args.attribute = list( set (args.attribute ))
+
+    return args
+
+
+# Same reads from same file as xmlchange
```
jgfouca: What is this a comment for?

scripts-python/xmlquerry
```diff
@@ -0,0 +1,182 @@
...
+# Use same methods
+
+
+
+def xmlquerry(caseroot, args, listofattributes=[] ,  subgroup=None,
```
jgfouca: query is misspelled.

scripts-python/xmlquerry
```diff
@@ -0,0 +1,182 @@
...
+  
+
+def _main_func(description):
+    if ("--test" in sys.argv):
+        test_results = doctest.testmod(verbose=True)
+        sys.exit(1 if test_results.failed > 0 else 0)
+
+
+    # Declare cli arguments - add to args
+    listofattributes = None
```
jgfouca: These variables don't look like they are used.

jgfouca:
Widespread misspelling of xmlquery as xmlquerry.

jgfouca:
I'm very confused as to why such extensive changes were needed to our core XML infrastructure (entry_id, case, etc)?

wilke:
Corrected spelling

wilke:
Not sure what you mean but perhaps I can elaborate:

- It is not the case that every attribute is unique, e.g. JOB_WALLCLOCK_TIME in env_batch.xml. For this it is beneficial if a function retrieving the value for the given attribute is returning a list of all occurrences. We did not want to change the behavior of get_value but instead agreed to implement get_values which can take a list of attributes and returns always a list of dictionaries - enforcing a consistent behavior (case.py)
- In addition the old xmlquery reported type, default values and description for the given attribute - hence the addition of those methods in entry_id.py
- Those additions have been made to not change the current implementations of the existing functions.

Is this helping you? Do you have any suggestions for changes?


On Mar 23, 2016, at 3:22 PM, James Foucar <notifications@github.com<mailto:notifications@github.com>> wrote:


I'm very confused as to why such extensive changes were needed to our core XML infrastructure (entry_id, case, etc)?

—
You are receiving this because you authored the thread.
Reply to this email directly or view it on GitHub<https://github.com/ESMCI/cime/pull/117#issuecomment-200528046>

Andreas Wilke
Argonne National Laboratory
Mathematics and Computer Science Division
Bldg. 240 , 4F8
9700 South Cass Avenue
Argonne, IL  60439
(630) 252-3190 (phone)
(630) 252-5986.(fax)

utils/python/CIME/XML/entry_id.py
```diff
@@ -56,6 +56,40 @@ def get_type_info(self, vid):
...
+    # Get description , expect child with tag "description" for parent node
+    def get_description (self, node):
+        type_node = self.get_optional_node("desc", root=node)
+        if type_node is not None:
+            return type_node.text
+        else:
+            return None
+
+    # Get group , expect child with tag "group" for parent node        
+    def get_group (self, node):
```
wilke: I  suggest to change the name to _get_group(self,node) . It seems that methods starting with "_get" are taking a xml node as an argument where methods starting with "get" are taking an attribute name as argument and invoking the underscore method.

jgfouca:
I must not be remembering that conversation. Did Jim Edwards agree to this? If so, I'm OK with it.

utils/python/CIME/XML/env_batch.py
```diff
@@ -32,23 +32,60 @@ def set_value(self, item, value, subgroup=None):
...
-            if subgroup is None:
-                nodes = self.get_nodes("job")
-            else:
-                nodes = self.get_nodes("job", {"name":subgroup})
+    # def get_value(self, item, attribute={}, resolved=True, subgroup=None):
+#         value = None
+#         logger.debug("Value=%s additional Attribute:%s" , item , attribute)
+#         nodes = self.get_nodes("entry",{"id":item})
+#         logger.debug("Nodes:%s" , nodes)
+#         if len(nodes) > 0:
```
jgfouca: There's no "get_value" method for EnvBatch anymore?

utils/python/CIME/XML/env_mach_specific.py
```diff
@@ -4,17 +4,92 @@
 from standard_module_setup import *
 
 from generic_xml import GenericXML
+from env_base import EnvBase
+
 
 logger = logging.getLogger(__name__)
 
-class EnvMachSpecific(GenericXML):
+#class EnvMachSpecific(GenericXML):
```
jgfouca: Remove comment,

cime_config/config_headers.xml
```diff
@@ -28,6 +28,12 @@
     </header>
   </file>
 
+  <file name ="env_mach_specific.xml">
+    <header>
+    These variables should not be changed.
```
jedwards4b: Let's change this message - How about "These variables control the machine dependent environment including the paths to compilers and libraries external to cime such as netcdf, environment variables for use in the running job should also be set here."

scripts-python/xmlquery
```diff
@@ -0,0 +1,168 @@
...
+
+
+
+from standard_script_setup import *
+
+from CIME.utils import expect, run_cmd
+from CIME.case import Case
+
+import argparse, doctest, shutil, glob
+import xml.etree.ElementTree as ET
```
jedwards4b: You should not need ElementTree here.

scripts-python/xmlquery
```diff
@@ -0,0 +1,168 @@
...
+
+    parser = argparse.ArgumentParser(   description="Querry the xml files for attributes and return their values." , 
+                                        formatter_class=argparse.RawDescriptionHelpFormatter, epilog=textwrap.dedent(__doc__) )
+    CIME.utils.setup_standard_logging_options(parser)
+
+    # Set command line options
+    parser.add_argument("attributes", nargs="*"  ,
+                        help="attribute value in env_*.xml file ( <element_name attribute='attribute_value'>value</element_name> )")
+    
+    parser.add_argument("--file",
```
jedwards4b: In order to be backward compatible these options should have both single dash and double dash, see file
https://github.com/ESMCI/cime/blob/master/scripts/Tools/xmlchange for an example.

utils/python/CIME/XML/entry_id.py
jedwards4b: The idea is that subroutines with an initial underscore are only called interior to this module or other modules within the XML directory.  node should never be referenced outside the XML directory so I think that you are correct, this should be _get_group

utils/python/CIME/XML/env_batch.py
jedwards4b: Maybe we should have get_value in env_batch - it would make the subgroup argument required.   Or maybe we don't need it?

jedwards4b:
Review complete.

jedwards4b:
Change the two logging.info statements in xmlquery to logging.debug, move to directory scripts/Tools.

scripts/Tools/xmlquery
```diff
@@ -1,251 +1,180 @@
...
-    my ($idlist) = @_; 
-
-    # First create the hash %xmlvars and %xmlfile in order to call
-    # expand_xml_var if needed
-    my %xmlvars;
-    my %xmlfile;
-    my @files;
-    if($opts{file}){
-	@files = ( $opts{file} );
+#!/usr/bin/env python
```
jedwards4b: I think we want python2 here.

utils/python/CIME/XML/entry_id.py
```diff
@@ -56,6 +56,40 @@ def get_type_info(self, vid):
...
+        if default is not None:
+            return default.text
+        else:
+            return None
+        
+    
+    # Get type description , expect child with tag "type" for node         
+    def _get_type (self, node):
+        type_node = self.get_optional_node("type", root=node)
+        if type_node is not None:
```
jedwards4b: I think this may need to be if type_node is not None and type_node.text is not None?

utils/python/CIME/XML/generic_xml.py
```diff
@@ -37,6 +37,7 @@ def __init__(self, infile=None):
         else:
             # if file does not exist create a root xml element
             # and set it's id to file
+            logger.warning("File %s does not exists." , infile)
```
jedwards4b: I don't think that we want a warning here - creating a new file will be a regular operation in this init.

utils/python/CIME/case.py
```diff
@@ -24,13 +24,19 @@
 class Case(object):
 
     def __init__(self, case_root=os.getcwd()):
+        
+        # Init first, if no valid case_root expect fails and tears down object, __del__ expects self._env_files_that_need_rewrite
+        self._env_files_that_need_rewrite = set()
+        
         expect(os.path.isdir(case_root),
```
jedwards4b: This will be removed in an upcoming commit - when we bring in create_newcase.   If __del__ is called before the case_root is created it should not call flush()

scripts/Testing/testscripts.py
```diff
@@ -0,0 +1,61 @@
+"""
```
jedwards4b: Is your plan to call this from scripts_regression_tests?

jedwards4b:
Review complete - I think that this is ready to merge but feel free to add more unit tests if you like.   
Have you run the scripts_regression_tests?

utils/python/CIME/XML/entry_id.py
wilke: This only applies if we want to return something else than None as default otherwise if a node has no text field node.text will return None.

rljacob:
What is up with merging this?

jedwards4b:
My understanding is that we are waiting for Andreas to integrate his unit testing with scripts_regression_tests and to run pre-commit testing.

wilke:
Misunderstanding, I thought the test integration will be part of a different branch and integrating with scripts_regression_tests will take more time. Have to resolve new conflicts first. Will rebase and fix conflicts first.

PR #121, opened 2016-03-27
Confopts

mvertens:
removal of confopts from create_newcase and incorporation of this logic in system_tests.py


utils/python/CIME/XML/env_batch.py
```diff
@@ -4,7 +4,7 @@
 from standard_module_setup import *
 
 from env_base import EnvBase
-
+from CIME.utils import convert_to_string
```
jgfouca: @mvertens , I don't see how these env_batch changes are related to the conf_opts move to system_test.py. Can you explain?

utils/python/CIME/system_test.py
```diff
@@ -361,15 +361,104 @@ def _create_newcase_phase(self, test):
...
-        logger.debug("Found drv_config_file %s" % drv_config_file)
-
         drv_comp = Component(drv_config_file)
+        component_classes = drv_comp.get_valid_model_components()
+
+        # Set up a case object
+        case = Case(self._get_test_dir(test)) 
+
+        # Determine the test_case from the test name
+        test_case = CIME.utils.parse_test_name(test)[0]
```
jgfouca: Minor improvement:

    test_case, case_opts = CIME.utils.parse_test_name(test)[:2]

utils/python/CIME/system_test.py
```diff
@@ -361,15 +361,104 @@ def _create_newcase_phase(self, test):
...
+        case_opts = CIME.utils.parse_test_name(test)[1] 
+
+        # Determine case_opts from the test_case
+        if case_opts is not None:
+            logger.info("case_opts are %s " %case_opts)
```
jgfouca: Please change all these logger.info to logger.debug. We want to keep the output of create_test nice and clean in the default case.

utils/python/CIME/system_test.py
```diff
@@ -361,15 +361,104 @@ def _create_newcase_phase(self, test):
...
+                    case.set_value("USE_ESMF_LIB", True)
+                    case.set_value("COMP_INTERFACE", "ESMF")
+                    logger.info (" USE_ESMF_LIB set to TRUE")
+                    logger.info (" COMP_INTERFACE set to ESMF")
+
+                if opt == 'CG':
+                    case.set_value("CALENDAR", "GREGORIAN")
+                    logger.info (" CALENDAR set to %s" %opt)
+
+                match =  re.search(r'^L([A-Za-z])([0-9]*)', opt, re.M)
```
jgfouca: re.M enables multiline matching. Is that really necessary here?

utils/python/CIME/system_test.py
```diff
@@ -361,15 +361,104 @@ def _create_newcase_phase(self, test):
...
+                if opt == 'E':
+                    case.set_value("USE_ESMF_LIB", True)
+                    case.set_value("COMP_INTERFACE", "ESMF")
+                    logger.info (" USE_ESMF_LIB set to TRUE")
+                    logger.info (" COMP_INTERFACE set to ESMF")
+
+                if opt == 'CG':
+                    case.set_value("CALENDAR", "GREGORIAN")
+                    logger.info (" CALENDAR set to %s" %opt)
+
```
jgfouca: All your re.searches start with a beginning of line symbol ('^'). You could replace all of these with:

    re.match(r'', opt)

... and you could get rid of the ('^') symbol.

utils/python/CIME/system_test.py
```diff
@@ -361,15 +361,104 @@ def _create_newcase_phase(self, test):
...
+                    stop_option = {"y":"nyears", "m":"nmonths", "d":"ndays", "h":"nhours", 
+                                   "s":"nseconds", "n":"nsteps"}
+                    opt = match.group(1)
+                    case.set_value("STOP_OPTION",stop_option[opt])
+                    opti = match.group(2)
+                    case.set_value("STOP_N", int(opti))
+                    logger.info (" STOP_OPTION set to %s" %stop_option[opt])
+                    logger.info (" STOP_N      set to %s" %opti)
+
+                match =  re.search(r'^M(.+)', opt, re.M)
```
jgfouca: re.search works here, but it way overkill. I would have coded this as follows:

    if opt.startswith("M"):
        mpilib = opt[1:]
        case.set_value("MPILIB", mpilib)

utils/python/CIME/system_test.py
```diff
@@ -361,15 +361,104 @@ def _create_newcase_phase(self, test):
...
+            logger.info("case_opts are %s " %case_opts) 
+            for opt in case_opts:
+
+                logger.debug("case_opt is %s" %opt)
+                if opt == 'D':
+                    case.set_value("DEBUG", True)
+                    logger.info (" DEBUG set to TRUE")
+
+                if opt == 'E':
+                    case.set_value("USE_ESMF_LIB", True)
```
jgfouca: Caution. Since you are using sequential if statements and not elifs, it will try to execute all of these. In the body of some of these, you change opt so opt no longer means what you expect it to. This also prevents you from throwing an error in the case of an unhandled case opt. I would strong recommend changing to:

    if opt == 'D':
        ...
    elif opt == 'E':
        ...
    elif opt == 'CG':
        ...
    elif opt.startswith('L'):
        ...
    elif opt.startswith('M'):
        ...
    elif opt.startswith('P'):
        ...
    else:
        expect(False, "Could not handle case_opt '%s' " % opt)

You should also have error handling within the elif statements. For example, if the opt starts with 'P' but the characters that follow are not what we expect for a P option, that needs to be an error.

utils/python/CIME/XML/env_batch.py
mvertens: Good question. I had one regression test fail - and when Jim provided me this fix to get the tests to pass.
I think he needs to comment. It was not tied directly to my changes - only tied to getting the tests to pass.

jgfouca:
Review complete.

utils/python/CIME/system_test.py
mvertens: I agree - this is much cleaner syntax. Thanks.

utils/python/CIME/XML/env_batch.py
jgfouca: OK, that's fine then. Probably not worth making another branch for this.

utils/python/CIME/system_test.py
mvertens: Done.

utils/python/CIME/system_test.py
mvertens: No - its not necessary. I'm getting rid of it now.

utils/python/CIME/system_test.py
mvertens: Good point. I'm making this replacement.

jgfouca: Just to reiterate, I think almost all of your regex operations could be replaced with simple string operations like "startswith" and slicing (varname[X:Y])

utils/python/CIME/XML/env_batch.py
jedwards4b: These were just to get tests working and when we merge to master we want to make sure that the master copy of env_batch and not this one is kept.

PR #124, opened 2016-04-04
Add default handling to batch.py.

jgfouca:
Prefer Element methods instead of direct access to element.attrib.

Fix very lame handling of queue and wall_time in machines.py.

### DIFF
--- a/scripts/Tools/cimeteststatus
+++ b/scripts/Tools/cimeteststatus
@@ -1,9 +1,11 @@
 #!/usr/bin/env python
-""" CIME test status reporting script. 
-Purpose:  Give basic and detailed summaries of CIME(CESM) tests, 
-and send the test results back to the test database. 
+"""
+CIME test status reporting script.
+Purpose:  Give basic and detailed summaries of CIME(CESM) tests,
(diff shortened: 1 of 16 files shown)


scripts/Tools/cimeteststatus
```diff
@@ -317,52 +318,52 @@ def findSpecs(args):
...
         suiteinfo['compiler'] = compiler.text
-    
+
     for mpilib in root.findall('mpilib'):
         suiteinfo['mpilib'] = mpilib.text
-    
+
     for t in root.findall('test'):
-        testlist.append(t.attrib['case'])
+        testlist.append(t.get('case'))
```
jedwards4b: You seem to favor .get over .attrib but both do the same thing - should we have an explicit policy to use one or the other?

jedwards4b:
Review complete.

scripts/Tools/cimeteststatus
jgfouca: Yes, the .get one should be preferred in most cases. The .attrib one can be used if you want to throw an exception if the attribute is missing.

PR #130, opened 2016-04-15
Jedwards/create newcase

jedwards4b:
Rewrite of create_newcase in python.


cime_config/acme/config_files.xml
```diff
@@ -48,11 +48,19 @@
     <desc>file containing specification of pio settings for target model possible machine, compiler, mpilib, compset and/or grid attributes (for documentation only - DO NOT EDIT)</desc>
   </entry>
 
+  <entry id="CONFIG_TESTS_FILE">
```
jedwards4b: Add a pointer to the config_tests.xml.

cime_config/acme/machines/config_batch.xml
```diff
@@ -210,5 +210,23 @@
      </directives>
    </batch_system>
 
+   <batch_jobs>
```
jedwards4b: These are the template codes and their corresponding case executable codes.

cime_config/cesm/config_files.xml
```diff
@@ -48,6 +48,14 @@
     <desc>file containing specification of pio settings for target model possible machine, compiler, mpilib, compset and/or grid attributes (for documentation only - DO NOT EDIT)</desc>
   </entry>
 
+  <entry id="CONFIG_TESTS_FILE">
```
jedwards4b: add a pointer to the config_tests.xml file.

cime_config/cesm/machines/config_batch.xml
```diff
@@ -67,7 +67,6 @@
       <directive                       > -R "span[ptile={{ ptile }}]"</directive>
       <directive                       > -N  </directive>
       <directive default="poe"         > -a {{ poe }} </directive>
-      <directive                       > -x {{ queue_exclusive }} </directive>
```
jedwards4b: Flag is implied by queue argument and not needed explicitly

cime_config/cesm/machines/template.st_archive
```diff
@@ -47,7 +47,7 @@ if(-e $testlog){
     close(TL);
 }
 $logger->info("st_archive starting");
-system("$config{CASEROOT}/st_archive >> stArchiveStatus 2>&1");
```
jedwards4b: move this file to be consistant with other xml

driver_cpl/cime_config/config_component.xml
```diff
@@ -608,8 +608,8 @@
   <entry id="BATCHREDIRECT">
     <type>char</type>
     <default_value></default_value>
-    <group>run_machine</group>
```
jedwards4b: move from env_case to env_batch.xml

driver_cpl/cime_config/config_component.xml
```diff
@@ -907,15 +907,6 @@
     <desc>Shared library root, (set automatically to $EXEROOT - DO NOT EDIT)</desc>
   </entry>
 
-  <entry id="SHAREDLIBROOTDIR">
```
jedwards4b: This variable was not used anyplace.

driver_cpl/cime_config/config_component.xml
```diff
@@ -1915,16 +1906,16 @@
   <entry id="BATCHQUERY">
     <type>char</type>
     <default_value>UNSET</default_value>
-    <group>run_machine</group>
```
jedwards4b: move to env_batch.xml

scripts/Tools/case.submit
```diff
@@ -74,9 +74,9 @@ sub main
 
     my $job;
     if(defined $config{TESTCASE}){
-	$job = "test";
```
jedwards4b: Using the full name instead of a subset just makes some of the parsing easier

scripts/create_clone
```diff
@@ -259,24 +261,25 @@ $sysmod = "chmod 755 $caseroot/case.build";
 system ($sysmod); if ($? == -1) {die "$sysmod failed: $!\n";}
 
 # Create case.clean_build
-$sysmod = "cp $cimeroot/scripts/Tools/clean_build $caseroot/case.clean_build"; 
+$sysmod = "cp $cimeroot/scripts/Tools/case.clean_build $caseroot/case.clean_build";
```
jedwards4b: Name changed to be consistent.

scripts/create_clone
```diff
@@ -259,24 +261,25 @@ $sysmod = "chmod 755 $caseroot/case.build";
...
 # Create case.submit
-$sysmod = "cp  $cimeroot/scripts/Tools/case.submit $caseroot/case.submit"; 
+$sysmod = "cp  $cimeroot/scripts/Tools/case.submit $caseroot/case.submit";
 system($sysmod) == 0 or die "ERROR: $sysmod failed: $?\n";
 
 
 #-----------------------------------------------------------------------------------------------
 # Create batch script
-$sysmod = "./case.setup"; 
+$ENV{PYTHONPATH} = "$cimeroot/scripts/Tools:$cimeroot/utils/python";
```
jedwards4b: Make sure we can call this from perl by setting the pythonpath

scripts/create_clone
```diff
@@ -259,24 +261,25 @@ $sysmod = "chmod 755 $caseroot/case.build";
...
+$sysmod = "./case.setup";
 system ($sysmod); if ($? == -1) {die "$sysmod failed: $!\n";}
 
 
 sub substitute()
 {
     my $filetosearch = $File::Find::name;
-    if(-f $filetosearch && ! -d $filetosearch)
+    if(-f $filetosearch && ! -d $filetosearch && -w $filetosearch)
```
jedwards4b: File needs to be writable.

utils/perl5lib/Batch/BatchMaker.pm
```diff
@@ -220,22 +220,24 @@ sub getBatchSystemTypeForMachine()
     my $configmachinesparser = $self->{'configmachinesparser'};
     my @batchtypes = $configmachinesparser->findnodes("/config_machines/machine[\@MACH=\'$mach\']/batch_system");
 
-    if(!@batchtypes)
+    if(@batchtypes)
```
jedwards4b: allow for batch_type == 'none'

utils/perl5lib/Batch/BatchMaker.pm
```diff
@@ -278,47 +280,47 @@ sub getField()
...
 
-    my @batch_directive = $batchparser->findnodes("/config_batch/batch_system[\@type=\'$self->{'batch_system'}\']/batch_directive");
+    # get the batch directive for this particular queueing system.
+
+    my @batch_directive = $batchparser->findnodes("/config_batch/batch_system[\@type=\'$self->{'batchtype'}\']/batch_directive");
 
     if(!@batch_directive)
     {
-	die "Cannot find batch directive for the batch system type $self->{'batch_system'}";
+	die "Cannot find batch directive for the batch system type $self->{'batchtype'}";
```
jedwards4b: xpath to batch_directive value different in config than in case.   this is only a perl problem.

utils/perl5lib/Config/SetupTools.pm
```diff
@@ -111,7 +111,7 @@ sub getAllResolved
     my %masterconfig;
 
     # Get all the env*.xml files into an array...
-    my @xmlfiles = qw( env_build.xml env_case.xml env_mach_pes.xml env_run.xml);
+    my @xmlfiles = qw( env_build.xml env_case.xml env_mach_pes.xml env_run.xml env_batch.xml);
```
jedwards4b: Need to add env_batch.xml to perl case config files.

utils/python/CIME/XML/batch.py
```diff
@@ -111,3 +111,18 @@ def get_batch_directives(self, batch_maker):
             result.append("%s %s" % (directive_prefix, directive))
 
         return result
+
```
jedwards4b: get the list of batch jobs from config_batch.xml

utils/python/CIME/XML/entry_id.py
```diff
@@ -15,32 +17,60 @@ def __init__(self, infile=None):
         GenericXML.__init__(self, infile)
         self.groups={}
 
-    def set_default_value(self, node, attributes=None):
```
jedwards4b: Decided that get makes more sense than set.

utils/python/CIME/XML/entry_id.py
```diff
@@ -15,32 +17,60 @@ def __init__(self, infile=None):
...
-            node.set("value", value.text)
-            return value.text
+
+        node = self.get_optional_node(vid)
+        value = None
+        if node is not None:
+            value = self._get_value_match(self,node,attributes)
+        return value
+
+    def _get_value_match(self, node, attributes={}):
```
jedwards4b: get value match finds the xml entry with the highest count of matching attributes.

utils/python/CIME/XML/entry_id.py
```diff
@@ -57,7 +87,12 @@ def get_type_info(self, vid):
         else:
             return self._get_type_info(node)
 
-    def _set_value(self, node, vid, value, subgroup=None, ignore_type=False):
+    def _set_value(self, node, value, vid=None, subgroup=None, ignore_type=False):
```
jedwards4b: vid should be option here, so i had to change the order.

utils/python/CIME/XML/entry_id.py
```diff
@@ -151,36 +182,71 @@ def get_elements_from_child_content(self, childname, childcontent):
         nodes = self.get_nodes("entry")
         elements = []
         for node in nodes:
-            childnode = self.get_node(childname,root=node)
+            childnode = self.get_optional_node(childname,root=node)
+            expect(childnode is not None,"No childname %s for id %s"%(childname,node.get("id")))
             content = childnode.text
             if content == childcontent:
-                elements.append(node)
+                elements.append(deepcopy(node))
```
jedwards4b: without deepcopy I was only getting the node and not the nodes children.

utils/python/CIME/XML/entry_id.py
```diff
@@ -151,36 +182,71 @@ def get_elements_from_child_content(self, childname, childcontent):
...
+        # For matchs found: Remove {<group>, <file>, <values>}
+        # children from each entry and set the default value for the
+        # new entries in self - putting the entries as children of
+        # group elements in file $file
+        for src_node in nodelist:
+	    node  = deepcopy(src_node)
+            gnode = src_node.find(".//group")
             gname = gnode.text
+            if gname is None:
+                gname = "group_not_set"
```
jedwards4b: A generic group to be used where the component didn't define one.

utils/python/CIME/XML/env_archive.py
```diff
@@ -3,14 +3,27 @@
 """
 from standard_module_setup import *
 
-from env_base import EnvBase
```
jedwards4b: env_archive does not follow the entry id format and so cannon inherit from EnvBase.

utils/python/CIME/XML/env_archive.py
```diff
@@ -3,14 +3,27 @@
...
         """
         initialize an object interface to file env_archive.xml in the case directory
         """
-        EnvBase.__init__(self, case_root, infile)
+        if os.path.abspath(infile):
+            fullpath = infile
+        else:
+            fullpath = os.path.join(case_root, infile)
+        GenericXML.__init__(self, fullpath)
+        if not os.path.isfile(infile):
```
jedwards4b: Currently this file is just a copy of cime_config/<model>/archive.xml with a header node added.

utils/python/CIME/XML/entry_id.py
```diff
@@ -151,36 +182,71 @@ def get_elements_from_child_content(self, childname, childcontent):
...
+                self.add_child(newgroup)
 
-            self.set_default_value(node, attlist)
+            # Remove {<group>, <file>, <values>} from the entry element
             node = self.cleanupnode(node)
+
+	    # Add the entry element to the group
             self.groups[gname].append(node)
+
+	    # Set the default value, it may be determined by a regular
```
jedwards4b: We need to get the default_value from the srcnode since it may be stripped from the case node.

utils/python/CIME/XML/env_batch.py
```diff
@@ -67,6 +69,34 @@ def get_type_info(self, vid):
...
         for node in self.get_nodes("job"):
-            result.append(node.get("name"))
-
+            name = node.get("name")
+            template = self.get_value("template", subgroup=name)
+            task_count = self.get_value("task_count", subgroup=name)
+            result.append((name, template, task_count))
         return result
+
+    def create_job_groups(self, bjobs):
```
jedwards4b: For each job listed in config_batch.xml create a case file from it's template.

utils/python/CIME/XML/env_batch.py
```diff
@@ -67,6 +69,34 @@ def get_type_info(self, vid):
...
+        # only expect one group in this file
+        group = self.get_node("group", {"id":"job_submission"})
+        # look to see if any jobs are already defined
+        cjobs = self.get_jobs()
+        childnodes = list()
+
+        expect(len(cjobs)==0," Looks like job groups have already been created")
```
jedwards4b: Not yet handling the case where you might add a job to an existing case.

utils/python/CIME/XML/env_mach_pes.py
```diff
@@ -14,3 +14,29 @@ def __init__(self, case_root=os.getcwd(), infile="env_mach_pes.xml"):
         initialize an object interface to file env_mach_pes.xml in the case directory
         """
         EnvBase.__init__(self, case_root, infile)
+
+    def get_value(self, vid, attribute={}, resolved=True, subgroup=None):
```
jedwards4b: negative values in these fields correspond to a number of nodes rather than number of tasks.

utils/python/CIME/XML/env_mach_pes.py
```diff
@@ -14,3 +14,29 @@ def __init__(self, case_root=os.getcwd(), infile="env_mach_pes.xml"):
...
+            value = -1*value*self.get_value("PES_PER_NODE")
+        return value
+#
+# We need a set value until we full transition from perl
+#
+
+    def _set_value(self, node, value, vid=None, subgroup=None, ignore_type=False):
+        if vid is None:
+            vid = node.get("id")
+
```
jedwards4b: But the perl code can't handle it so we need this correction until we eliminate the perl

utils/python/CIME/XML/generic_xml.py
```diff
@@ -14,17 +14,8 @@ def __init__(self, infile=None):
...
-        # Hold arbitary values. In create_newcase we may set values
-        # for xml files that haven't been created yet. We need a place
-        # to store them until we are ready to create the file. At file
-        # creation we get the values for those fields from this lookup
-        # table and then remove the entry. This was what I came up
-        # with in the perl anyway and I think that we still need it
-        # here.
-        self.lookups = {}
-        self.lookups['CIMEROOT'] = get_cime_root()
+        if not hasattr(self, 'tree'):
```
jedwards4b: This allows you to append to a previously opened file.   So for example we can append a $HOME/.cime/config_machines.xml to the cime_config/<model>/machines/config_machines.xml

utils/python/CIME/XML/generic_xml.py
```diff
@@ -63,55 +58,72 @@ def write(self, outfile=None):
             outfile = self.filename
 
         logger.debug("write: " + outfile)
-        xmlstr = ET.tostring(self.root)
-        doc = minidom.parseString(xmlstr)
-        with open(outfile,'w') as xmlout:
-            doc.writexml(xmlout,addindent='  ')
+        try:
```
jedwards4b: If the xml cannot be translated to a string then dump self.root - this usually points directly to the offending entry.

utils/python/CIME/XML/generic_xml.py
```diff
@@ -63,55 +58,72 @@ def write(self, outfile=None):
...
+            xmlstr = ET.tostring(self.root)
+        except:
+            ET.dump(self.root)
+            expect(False, "Could not write file %s, xml formatting error"%self.filename)
+
+        # xmllint provides a better format option for the output file
+        xmllint = which("xmllint")
```
jedwards4b: If xmllint exists in the path use it to format the output xml.   This makes the file indentation nicer.

utils/python/CIME/XML/generic_xml.py
```diff
@@ -63,55 +58,72 @@ def write(self, outfile=None):
...
                (nodename, attributes, self.filename))
         return nodes[0] if nodes else None
 
-    def get_nodes(self, nodename, attributes=None, root=None):
+    def get_nodes(self, nodename, attributes=None, root=None, xpath=None):
         if root is None:
             root = self.root
         nodes = []
-        xpath = ".//"+nodename
+        expect(attributes is None or xpath is None,
```
jedwards4b: Allow the user to provide more complicated xpath statements.

utils/python/CIME/XML/generic_xml.py
```diff
@@ -186,10 +187,19 @@ def get_resolved_value(self, raw_value):
             logger.debug("find: %s" % var)
             ref = self.get_value(var)
             if ref is not None:
-                logger.debug("resolve: %s" % ref)
-                item_data = item_data.replace(m.group(), str(self.get_resolved_value(ref)))
+                logger.debug("resolve: " + str(ref))
+                item_data = item_data.replace(m.group(), self.get_resolved_value(str(ref)))
+            elif var == "CIMEROOT":
```
jedwards4b: These two variables cause a lot of errors - handling them directly just makes life easier.

utils/python/CIME/batch_maker.py
```diff
@@ -108,33 +109,32 @@ def _set_task_info(self):
     def _set_queue(self):
         self.queue = self.env_batch.get_value("JOB_QUEUE", subgroup=self.job)
         self.ccsm_estcost = self.case.get_value("CCSM_ESTCOST")
-
         self.wall_time_max = None
         if self.queue:
             return
 
         # Make sure to check default queue first.
-        self.queue = self.config_machines_parser.get_default_queue()
```
jedwards4b: I found this code really confusing - sometimes self.queue was a string and other times it was a node.   Now it's always a string.

utils/python/CIME/batch_maker.py
```diff
@@ -285,3 +288,10 @@ class is returned.
         return batch_maker
 
 
+def find_project():
```
jedwards4b: Several possible locations of PROJECT information.

utils/python/CIME/case.py
```diff
@@ -6,49 +6,84 @@
...
+from CIME.XML.env_batch             import EnvBatch
+
+from CIME.XML.generic_xml           import GenericXML
 
 logger = logging.getLogger(__name__)
 
 class Case(object):
 
     def __init__(self, case_root=os.getcwd()):
-        expect(os.path.isdir(case_root),
```
jedwards4b: Allow for initializing a case that does not exist yet.

utils/python/CIME/case.py
```diff
@@ -6,49 +6,84 @@
...
-        for env_file in self._env_files_that_need_rewrite:
-            env_file.write()
+    def _get_env(self, short_name):
+          full_name = "env_%s.xml" % (short_name)
+          for env_file in self._files:
+              if os.path.basename(env_file.filename) == full_name:
+                  return env_file
+          expect(False, "Could not find object for %s in case"%full_name)
+
+    def flush(self, flushall=False):
```
jedwards4b: It's possible to update a file without setting the needs_rewrite flag.  Using flushall assures that all files are written.

utils/python/CIME/case.py
```diff
@@ -6,49 +6,84 @@
...
+        self._files = self._env_entryid_files + self._env_generic_files
 
-        self._case_root = case_root
+        # Hold arbitary values. In create_newcase we may set values
+        # for xml files that haven't been created yet. We need a place
+        # to store them until we are ready to create the file. At file
+        # creation we get the values for those fields from this lookup
+        # table and then remove the entry. This was what I came up
+        # with in the perl anyway and I think that we still need it here.
+        self.lookups = {}
```
jedwards4b: lookups is a place to save (and find) values that need to be written to env object that haven't yet been initialized.

utils/python/CIME/system_test.py
```diff
@@ -374,19 +369,48 @@ def _create_newcase_phase(self, test):
...
+        test_dir = self._get_test_dir(test)
+        envtest = EnvTest(test_dir)
 
         # Determine list of component classes that this coupler/driver knows how
         # to deal with. This list follows the same order as compset longnames follow.
         files = Files()
         drv_config_file = files.get_value("CONFIG_DRV_FILE")
         drv_comp = Component(drv_config_file)
         component_classes = drv_comp.get_valid_model_components()
+        envtest.add_elements_by_group(drv_comp, {}, "env_test.xml")
```
jedwards4b: Write these values to env_test.xml instead of directly to the other case files, this allows for the test to be reset and rerun easier.

utils/python/CIME/system_test.py
```diff
@@ -467,42 +491,24 @@ def _xml_phase(self, test):
...
-
+        lockedfiles = os.path.join(test_dir, "Lockedfiles")
+        try:
+            os.stat(lockedfiles)
+        except:
+            os.mkdir(lockedfiles)
+        shutil.copy(os.path.join(test_dir,"env_run.xml"),
+                    os.path.join(lockedfiles, "env_run.orig.xml"))
+        case = Case(test_dir)
+        case.set_value("SHAREDLIBROOT",
```
jedwards4b: Set the SHAREDLIBROOT for test build

utils/python/CIME/system_test.py
```diff
@@ -467,42 +491,24 @@ def _xml_phase(self, test):
...
+                       os.path.join(self._test_root,
+                                    "sharedlibroot.%s"%self._test_id))
+        envtest.set_initial_values(case)
```
jedwards4b: This initializes the case xml from env_test.xml

utils/python/CIME/utils.py
```diff
@@ -550,6 +543,14 @@ def handle_standard_logging_options(args):
     root_logger=logging.getLogger()
     # DEBUG trumps INFO trumps Silent (WARN)
     if (args.debug == True):
+        file1 = {
```
jedwards4b: This change fixes the problem of writing an empty scripts.log file when you run without --debug while still printing a complete debug log when the flag is set.

utils/python/CIME/utils.py
```diff
@@ -650,3 +653,20 @@ def convert_to_seconds(time_str):
         result += int(component) * pow(60, idx)
 
     return result
+
```
jedwards4b: Added to support use of xmllint if it's in the path

utils/python/tests/scripts_regression_tests
```diff
@@ -27,6 +27,7 @@ class RunUnitTests(unittest.TestCase):
 ###############################################################################
 
     def do_unit_tests(self, script,from_dir=SCRIPT_DIR):
+        os.environ["CIMEROOT"] = CIME.utils.get_cime_root()
```
jedwards4b: Needed for doctest

cime_config/acme/config_files.xml
jgfouca: Why? Are we going to do this for all non-env xml files?

cime_config/cesm/machines/taskmaker.pl
```diff
@@ -249,7 +249,7 @@
 my $taskgeom = "(0";
 my $thrdgeom = " $maxt[0]";
 my $taskcnt = 1;
-my $thrdcnt = $maxt[0];
+my $thrdcnt = $maxt[0] > 0 ? $maxt[0] : 1;
```
jgfouca: I hope you remembered to make this change in the python task maker.

driver_cpl/cime_config/config_component.xml
jgfouca: Good.

driver_cpl/cime_config/config_component.xml
jgfouca: Would it be possible to move batch_system too?

scripts/Tools/case.submit
jgfouca: I agree.

scripts/create_clone
jgfouca: Good.

scripts/create_clone
jgfouca: This concerns me. Shouldn't the scripts and scripts/Tools be able to run without having to set PYTHONPATH?

scripts/create_newcase
```diff
@@ -1,1003 +1,177 @@
...
+###############################################################################
+def parse_command_line(args):
+###############################################################################
+
+    cime_model = CIME.utils.get_model()
+
+    parser = argparse.ArgumentParser()
+
+    CIME.utils.setup_standard_logging_options(parser)
+
```
jgfouca: Can we get rid of the non compliant "-longoption"? I know this may require that you chase down all the old calls to create_newcase, but I don't think they should be too many of those. We could then use proper short options which helps usability.

scripts/create_newcase
```diff
@@ -1,1003 +1,177 @@
...
+    case, compset, grid, machine, compiler, \
+        mpilib, project, pecount,  \
+        machine_dir, usermods_dir, user_compset, pesfile, \
+        user_grid, gridfile, srcroot, test \
+        = parse_command_line(sys.argv)
+
+    cimeroot  = os.path.abspath(CIME.utils.get_cime_root())
+
+    if srcroot is None:
```
jgfouca: You could save yourself some work here by just defaulting srcroot to os.path.dirname(cimeroot) in the argparse setup above.

scripts/create_newcase
```diff
@@ -1,1003 +1,177 @@
...
+        srcroot  = os.path.dirname(cimeroot)
+
+    # Is case relative or full path?
+    if os.path.isabs(case):
```
jgfouca: This should work in both cases:

    caseroot = os.path.abspath(case)
    case = os.path.basename(case)

scripts/create_clone
jedwards4b: I think that this is a temporary measure until we get rid of the perl.   I do have some concern that the files created from templates such as case.run will need something like this since they are not linked from the source tree as the others are.

scripts/create_newcase
```diff
@@ -1,1003 +1,177 @@
...
+    # Set values for env_case.xml
+    caseobj.set_value("CASE", os.path.basename(case))
+    caseobj.set_value("CASEROOT", caseroot)
+    caseobj.set_value("SRCROOT", srcroot)
+    caseobj.set_value("MACH", machine)
+    if project:
+        caseobj.set_value("PROJECT",project)
+
+    # Configure the Case
+    caseobj.configure(compset, grid, machine,
```
jgfouca: Do you really want to merge the Case class with all the create_newcase stuff?

utils/perl5lib/Batch/BatchMaker.pm
jgfouca: Need similar change in python batch maker.

utils/perl5lib/Batch/BatchMaker.pm
```diff
@@ -887,11 +889,11 @@ sub overrideNodeCount()
 }
 
 #==============================================================================
-# Simple factory class to get the right BatchMaker class for each machine.  
-# The only downside to this strategy is that we have to have a BatchMaker_${machine} 
-# class for every machine we port. 
+# Simple factory class to get the right BatchMaker class for each machine.
```
jgfouca: I'm getting overwhelmed by the number of whitespace changes in this PR. I'm scrolling through them fast and potentially missing real changes. I like that we're removing trailing whitespaces, but maybe our next PR should just remove them globally so we stop seeing them in every PR.

utils/perl5lib/Module/ModuleLoader.pm
```diff
@@ -289,34 +290,34 @@ sub findModulesForCase()
     }
     my $parser = XML::LibXML->new(no_blanks => 1);
     my $casemoduleparser = $parser->parse_file($self->{machspecificfile});
-    my @allmodulenodes = $casemoduleparser->findnodes("/machine[\@MACH=\'$machine\']/module_system/modules");
+    my @allmodulenodes = $casemoduleparser->findnodes("//module_system/modules");
```
jgfouca: More perl changes in code that has been ported to python. This makes me nervous that one or the other is not getting the updates it needs.

utils/python/CIME/SystemTests/eri.py
```diff
@@ -9,12 +9,13 @@ def __init__(self, caseroot, case):
...
         """
-        SystemTestsCommon.__init__(self, caseroot, case)
+        SystemTestsCommon.__init__(self,  caseroot, case)
+        self._testname = "ERI"
 
     def run(self):
-        self._case.set_value("CONTINUE_RUN", False)
-        self._case.set_value("REST_OPTION","none")
-        self._case.set_value("HIST_OPTION","$STOP_OPTION")
-        self._case.set_value("HIST_N","$STOP_N")
```
jgfouca: Were these commented out for debugging? If we don't need them at all, please delete.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -9,16 +9,24 @@
...
+    def __init__(self, caseroot=os.getcwd(), case=None):
         """
         initialize a CIME system test object, if the file LockedFiles/env_run.orig.xml
         does not exist copy the current env_run.xml file.  If it does exist restore values
         changed in a previous run of the test.
         """
+        print caseroot
         self._caseroot = caseroot
-        self._case = case
+                # Needed for sh scripts
```
jgfouca: Strange indent

scripts/create_newcase
jedwards4b: Yes, I do think that's where it belongs.   Eventually I would like to see the leftovers from the non-OO like case_setup and batch_maker better integrated into the )) scheme as well.

utils/perl5lib/Batch/BatchMaker.pm
jedwards4b: Agreed

utils/python/CIME/XML/entry_id.py
```diff
@@ -15,32 +17,60 @@ def __init__(self, infile=None):
...
+        # Handle this case:
         # <entry id ...>
         #  <values>
         #   <value A="a1">X</value>
         #   <value A="a2">Y</value>
-        #   <value A="a3">Z</value>
+        #   <value A="a3" B="b1">Z</value>
         #  </values>
         # </entry>
-        valnodes = self.get_nodes("value", root=node)
```
jgfouca: Good cleanup

utils/python/CIME/XML/entry_id.py
jgfouca: I like this. Give precedence to the best match instead of the first/last one.

utils/python/CIME/XML/entry_id.py
jgfouca: How does it make sense to set an entry_id with no id?

jedwards4b: Because you already have the node associated with the id.

utils/python/CIME/XML/entry_id.py
jgfouca: This concerns me. I don't understand how deepcopy works with XML elements, but for standing python containers, I would not approve this change. My specific concern is that any modifications to the deepcopied nodes would not be reflected in the XML (either in memory or on file) because you are only modifying a copy. If you need children of nodes, I'd prefer some kind of recursion.

jedwards4b: But that is exactly the intent.  We need a copy that does not modify the original because we are in the process of creating a new object both in memory and on file.

utils/python/CIME/XML/env_batch.py
```diff
@@ -67,6 +69,34 @@ def get_type_info(self, vid):
...
+        # only the job_submission group is repeated
+        group = self.get_node("group", {"id":"job_submission"})
+        # look to see if any jobs are already defined
+        cjobs = self.get_jobs()
+        childnodes = list()
+
+        expect(len(cjobs)==0," Looks like job groups have already been created")
+
+        for child in reversed(group):
+            childnodes.append(deepcopy(child))
```
jgfouca: Why not re-use the get_elements_from_child_content?

utils/python/CIME/XML/generic_xml.py
jgfouca: Confused. This is the constructor, so there's no way this if statement will ever be false, right?

utils/python/CIME/XML/generic_xml.py
```diff
@@ -63,55 +58,72 @@ def write(self, outfile=None):
...
-        xpath = ".//"+nodename
+        expect(attributes is None or xpath is None,
+               " Arguments attributes and xpath are exclusive")
+        if xpath is None:
+            xpath = ".//"+nodename
         if attributes is not None:
             # xml.etree has limited support for xpath and does not allow more than
             # one attribute in an xpath query so we query seperately for each attribute
             # and create a result with the intersection of those lists
             for key, value in attributes.iteritems():
```
jgfouca: Just curious, when would an xml attribute have a None value?

utils/python/CIME/XML/generic_xml.py
jedwards4b: Look at __init__ in machines.py to see how this is used.   I could avoid calling the generic __init__ the second time and just call read - that probably makes more sense, thanks.

utils/python/CIME/XML/generic_xml.py
```diff
@@ -125,21 +137,12 @@ def add_child(self, node, root=None):
             root = self.root
         self.root.append(node)
 
-    def get_value(self, item, resolved=True):
+    def get_value(self, item, resolved=True, settype=True):
         """
         get_value is expected to be defined by the derived classes, if you get here it is an error.
         """
         logger.debug("Get Value for "+item)
-        result = None
```
jgfouca: If it is really an error if you get here, as the comment says, why not raise an error?

utils/python/CIME/XML/generic_xml.py
```diff
@@ -14,17 +14,8 @@ def __init__(self, infile=None):
...
-        self.tree = None
-
-        # Hold arbitary values. In create_newcase we may set values
-        # for xml files that haven't been created yet. We need a place
-        # to store them until we are ready to create the file. At file
-        # creation we get the values for those fields from this lookup
-        # table and then remove the entry. This was what I came up
-        # with in the perl anyway and I think that we still need it
-        # here.
-        self.lookups = {}
```
jgfouca: Thanks for getting rid of lookups.

utils/python/CIME/XML/generic_xml.py
```diff
@@ -186,10 +187,19 @@ def get_resolved_value(self, raw_value):
             logger.debug("find: %s" % var)
             ref = self.get_value(var)
             if ref is not None:
-                logger.debug("resolve: %s" % ref)
```
jgfouca: Why use "string" + str(ref)  instead of  "string %s" % ref?

utils/python/CIME/XML/generic_xml.py
jgfouca: Is this an indication of a bigger problem?

utils/python/CIME/XML/grids.py
```diff
@@ -5,25 +5,143 @@
...
     def _get_component_grids(self, name):
         gridRE = re.compile(r"[_]{0,1}[a-z]{1,2}%")
-        component_grids = gridRE.split(name.text)[1:]
+        component_grids = gridRE.split(name)[1:]
+
         return component_grids
 
+    def _get_domains(self, component_grids):
+        # use component_grids to create grids dictionary
+        grids = [("ATM", component_grids[0]), \
```
jgfouca: Wasn't there a more general way of getting this component names?

cime_config/cesm/machines/taskmaker.pl
jedwards4b: I believe that it's already there.

utils/python/CIME/XML/pes.py
```diff
@@ -0,0 +1,83 @@
...
+
+                               points = int(grid_match!="any")*4+int(mach_match!="any")*3+int(compset_match!="any")*2+int(pesize_match!="any")
+                               if points > max_points:
+                                   pe_select = pes_node
+                                   max_points = points
+                               elif points == max_points:
+                                   expect(False, "We dont expect to be here" )
+
+        pes_ntasks = {}
+
```
jgfouca: These three loops could be consolidated into one:

    pes_ntasks, pes_nthrds, pes_rootpe = {}, {}, {}
    for pes_data, xpath in [(pes_ntasks, "ntasks"), (pes_nthrds, "nthrds"), (pre_rootpe, "rootpe")]:
        nodes = pe_select.findall(".//%s/*" % xpath)
        for node in nodes:
            name = node.tag.upper()
            value = node.text
            pes_data[name] = value
            logger.debug("%s %s " % (name, value)

utils/python/CIME/XML/pio.py
```diff
@@ -0,0 +1,29 @@
...
+    def __init__(self, infile=None, files=None):
+        if infile is None:
+            if files is None:
+                files = Files()
+            infile = files.get_value("PIO_SPEC_FILE")
+
+        EntryID.__init__(self, infile)
+
+    def get_defaults(self, grid=None, compset=None, mach=None, compiler=None):
+# should we have a env_pio file
```
jgfouca: strange indent

utils/perl5lib/Module/ModuleLoader.pm
jedwards4b: already handled in the python

utils/python/CIME/batch_maker.py
jgfouca: I agree.

utils/python/CIME/XML/env_batch.py
jedwards4b: The function is different.   Here we are repeating the nodes in the job_submission group once for each job listed in config_batch.xml

utils/python/CIME/case.py
jgfouca: Then there might be a problem with our encapsulation of things. We should require that all updates go through Case.

utils/python/CIME/XML/generic_xml.py
jedwards4b: This is the input argument attributes which is used to create an xpath to search the xml object.  I found that in some cases the xpath was more complicated than that so I choose to optionally pass in the xpath argument directly instead of using the attributes argument.   I think that we could make this two different functions instead.

utils/python/CIME/system_test.py
```diff
@@ -343,19 +344,13 @@ def _create_newcase_phase(self, test):
...
+        create_newcase_cmd = "%s --case %s --res %s --mach %s --compiler %s --compset %s"\
+                               " --project %s --test"%\
                               (os.path.join(self._cime_root, "scripts", "create_newcase"),
-                               self._cime_model, test_dir, grid, machine, compiler, compset,
-                               test_case, self._project, sharedlibroot)
+                               test_dir, grid, machine, compiler, compset, self._project)
 
-        if test_case != 'PFS':
-            create_newcase_cmd += " -nosavetiming "
+#        if test_case != 'PFS':
```
jgfouca: Go ahead and delete.

utils/python/CIME/XML/generic_xml.py
jedwards4b: Changed comment to "get_value is expected to be defined by the derived classes, if you get here 
        the value was not found in the class."

utils/python/CIME/system_test.py
jgfouca: If we could encapsulate this in Case, then you wouldn't need that flushall flag.

utils/python/CIME/system_test.py
jgfouca: Didn't you delete SHAREDLIBROOT earlier because no one was using it?

jedwards4b: That was SHAREDLIBROOTDIR !

utils/python/CIME/utils.py
```diff
@@ -650,3 +653,20 @@ def convert_to_seconds(time_str):
...
+def which(program):
```
jgfouca: I think python supports this natively with distutils.spawn.find_executable

jgfouca:
Review complete, I did my best.

utils/python/CIME/XML/pes.py
jedwards4b: Incorporated - thanks,

utils/python/CIME/XML/generic_xml.py
jedwards4b: I don't think so.   But let's try not to let this list grow.

scripts/create_newcase
jedwards4b: done, thanks

utils/python/CIME/XML/grids.py
jedwards4b: Yes -  
I'm afraid that the quest for complete generality breaks down in config_grids.xml.  There is a lot of specific information here about interactions between components that currently is expressed in code instead of in XML.

PR #131, opened 2016-04-15
Compare with rebase

wilke:


scripts/Testing/testscripts.py
```diff
@@ -0,0 +1,61 @@
+"""
```
jedwards4b: I don't understand what this file is?

utils/python/CIME/XML/entry_id.py
```diff
@@ -106,28 +145,55 @@ def get_value(self, vid, attribute={}, resolved=True, subgroup=None):
         type_str = self._get_type_info(node)
         return convert_to_type(val, type_str, vid)
 
-    def get_values(self, vid, att, resolved=True):
+    def get_values(self, item, attribute={}, resolved=True, subgroup=None): # (self, vid, att, resolved=True , subgroup=None ):
```
jedwards4b: I think that your get_values is different enough from the get values in #130 that maybe they should be seperate functions?

utils/python/CIME/XML/generic_xml.py
```diff
@@ -132,7 +137,7 @@ def get_value(self, item, resolved=True):
         """
         get_value is expected to be defined by the derived classes, if you get here it is an error.
         """
-        logger.debug("Get Value for "+item)
+        logger.debug("Get Value for %s " , item)
```
jedwards4b: I think that the , here will give a error - should be %

jedwards4b:
I'm concerned about the merge of this change with that in #130.   I think that our definitions and expectations of get_values are quite different and I'm not sure how to resolve.

PR #132, opened 2016-04-16
Continue dev

jedwards4b:
This add a --reset option to case.setup which implies case.setup --clean ; case.setup
and converts tests to use a direct interface to case_setup instead of a shell (run_cmd) interface.
Builds on PR #130


utils/python/CIME/SystemTests/ncr.py
```diff
@@ -45,8 +46,7 @@ def build(self):
                     self._case.set_value("NTASKS_%s"%comp, ntasks*2)
             self._case.flush()
 
-            run_cmd("./case.setup --clean --test-mode")
-            run_cmd("./case.setup")
+            case_setup(self._caseroot, test_mode=True, reset=True)
```
jgfouca: Nice. This is much better.

PR #134, opened 2016-04-18
blues doesn't allow '-q' PBS batch option, this disables the pbs batc…

### LINKED ISSUE #128: Can't submit case for X case on blues
Got a bunch of error messages.
create_newcase -case Xcase2 -compset X -res f19_g16 -model acme -mach blues -compiler intel -project ACME
./case.setup
./case.build   all works.

But then ./case.submit ends with:

qsub: option requires an argument -- 'q'
Use of uninitialized value $output in scalar chomp at
	/lcrc/group/earthscience/jacob/cimeESMCI/utils/perl5lib/Batch/BatchUtils.pm line 220 (#1)
    (W uninitialized) An undefined value was used as if it were already
    defined.  It was interpreted as a "" or a 0, but maybe it was a mistake.
    To suppress this warning assign a defined value to your variables.
    
    To help you figure out what was undefined, perl will try to tell you the
    name of the variable (if any) that was undefined. In some cases it cannot
    do this, so it also tells you what operation you used the undefined value
    in.  Note, however, that perl optimizes your program and the operation
    displayed in the warning may not necessarily appear literally in your
    program.  For example, "that $foo" is usually optimized into "that "
    . $foo, and the warning will refer to the concatenation (.) operator,
    even though there is no . in your program.
    
Use of uninitialized value $jobstring in scalar chomp at
	/lcrc/group/earthscience/jacob/cimeESMCI/utils/perl5lib/Batch/BatchUtils.pm line 121 (#1)
Use of uninitialized value $jobstring in pattern match (m//) at
	/lcrc/group/earthscience/jacob/cimeESMCI/utils/perl5lib/Batch/BatchUtils.pm line 146 (#1)
 could not ascertain dependent job id... aborting
Uncaught exception from user code:
	 could not ascertain dependent job id... aborting at /lcrc/group/earthscience/jacob/cimeESMCI/utils/perl5lib/Batch/BatchUtils.pm line 152
 at /lcrc/group/earthscience/jacob/cimeESMCI/utils/perl5lib/Batch/BatchUtils.pm line 152
	Batch::BatchUtils::getJobID('Batch::BatchUtils=HASH(0x1518108)', undef) called at /lcrc/group/earthscience/jacob/cimeESMCI/utils/perl5lib/Batch/BatchUtils.pm line 221
	Batch::BatchUtils::submitSingleJob('Batch::BatchUtils=HASH(0x1518108)', 'case.run', undef, 1) called at /lcrc/group/earthscience/jacob/cimeESMCI/utils/perl5lib/Batch/BatchUtils.pm line 177
	Batch::BatchUtils::submitJobs('Batch::BatchUtils=HASH(0x1518108)') called at ./case.submit line 118
	main::main() called at ./case.submit line 156
Exit 255
blogin1[79]:

Full output of case.submit at https://gist.github.com/rljacob/d9509877446b35c9850cdd4eecf685cf

sarich:
Resolves #128 

Blues doesn't allow #PBS -q <queue> in case.run

This PR just checks to see if the directive line is ' -q ' and ignores it (-q itself will cause errors).

There may be a better fix for this, please feel free to opine.

### DIFF
--- a/utils/python/CIME/XML/batch.py
+++ b/utils/python/CIME/XML/batch.py
@@ -107,7 +107,6 @@ class Batch(GenericXML):
         for node in nodes:
             directive = self.get_resolved_value("" if node.text is None else node.text)
             directive = batch_maker.transform_vars(directive, default=node.get("default"))
-            if directive != ' -q ':
-                result.append("%s %s" % (directive_prefix, directive))
+            result.append("%s %s" % (directive_prefix, directive))
 
         return result
--- a/utils/python/CIME/XML/batch.py
+++ b/utils/python/CIME/XML/batch.py
@@ -107,7 +107,7 @@ class Batch(GenericXML):
         for node in nodes:
             directive = self.get_resolved_value("" if node.text is None else node.text)
             directive = batch_maker.transform_vars(directive, default=node.get("default"))
-
-            result.append("%s %s" % (directive_prefix, directive))
+            if directive != ' -q ':
+                result.append("%s %s" % (directive_prefix, directive))
 
         return result
--- a/utils/python/CIME/batch_maker.py
+++ b/utils/python/CIME/batch_maker.py
@@ -228,7 +228,6 @@ within model's Machines directory, and add a batch system type for this machine
             m = directive_re.search(text)
             variable = m.groups()[0]
             whole_match = m.group()
-
             if hasattr(self, variable.lower()) and getattr(self, variable.lower()) is not None:
                 repl = getattr(self, variable.lower())
                 text = text.replace(whole_match, str(repl))
@@ -238,8 +237,12 @@ within model's Machines directory, and add a batch system type for this machine
             elif default is not None:
                 text = text.replace(whole_match, default)
             else:
-                logger.warn("Could not replace variable '%s'" % variable)
-                text = text.replace(whole_match, "")
+                # If no queue exists, then the directive '-q' by itself will cause an error
+                if text.find('-q {{ queue }}')>0:
+                    text = ""
+                else:
+                    logger.warn("Could not replace variable '%s'" % variable)
+                    text = text.replace(whole_match, "")
 
         return text
 


jedwards4b:
I believe that it would be a better fix to say that if a directive_prefix occurs but the directive is not 
assigned then skip the append.

sarich:
The directive_prefix is '#PBS', the directive is ' -q <arg>', but I get the
idea. I'm just worried that there may be a case where someone might use a
directive with no argument.

jedwards4b:
The directive in config_batch.xml is -q {{ queue }} 
we need to find where {{ queue }} is substituted and if it returns
None then we drop the -q flag.  I think that's in batch_maker.py

rljacob:
The original perl version of the Batch utils also could not handle the Blues case.  It was fixed and here are the details from an email:

`In Batch/BatchMaker.pm starting at line 448, the setQueue function should be modified like so:

    #die "Cannot set queue for this machine! No default queue defined" if (! @defaultqueue);

    # set the default queue if we've found one.
    if(@defaultqueue)
    {
        my $defelement = $defaultqueue[0];
        $self->{'queue'} = $defelement->textContent();
    }


Then simply get rid of any and all <queues></queues> tags you have set for that machine in config_machines.xml.  I just tested this out on one machine, I have not tested it on any others.`

rljacob:
What is up with this PR?

jedwards4b:
I think we were waiting to hear from the originator.   But it'll be fixed in the near future without this PR.   Asking us to fix broken features in the perl that we are about to replace is a bit of a distraction IMHO.

sarich:
I think the perl reference was just adding flavor to the conversation, not an actual request.

> The directive in config_batch.xml is -q {{ queue }} 
> we need to find where {{ queue }} is substituted and if it returns
> None then we drop the -q flag. I think that's in batch_maker.py

I've moved this -q check to batch_maker.py

PR #135, opened 2016-04-18
Jedwards/check input issues

jedwards4b:
Builds on #132, this fixes some issues with check_input_data.


jgfouca:
Thanks for fixing those mistakes.

PR #138, opened 2016-04-21
pythonized create_clone

mvertens:


scripts/create_clone
```diff
@@ -1,318 +1,161 @@
...
+
+    # create caseroot        
+    caseobj.create_caseroot()
+    caseobj.flush(flushall=True, )
+
+    # copy user_nl_files
+    cmd = cloneroot + '/user_nl_*'
+    files = glob.glob(cmd)
+    for file in files:
+        cmd = "cp " + file + " " + caseroot + "/."
```
jedwards4b: Don't do shell copy commands - use shutil.copyfile instead.

scripts/create_clone
```diff
@@ -1,318 +1,161 @@
...
+
+    # copy SourceMod files
+    dirs = glob.glob(cloneroot + "/SourceMods/*")
+    for dir in dirs:
+        files = glob.glob(dir + "/*")
+        if files:
+            moddir = os.path.basename(dir)
+            for file in files:
+                cmd = "cp " + file + " " + caseroot + "/SourceMods/" + moddir
+                print " cmd is ",cmd
```
jedwards4b: replace with shutil.copyfile.   Also use os.path.join to build paths.

scripts/create_clone
```diff
@@ -1,318 +1,161 @@
...
+                run_cmd(cmd)
+
+    # copy env_case.xml to LockedFiles
+    cmd = "cp " + caseroot + "/env_case.xml " + caseroot + "/LockedFiles/." 
+    run_cmd(cmd)
```
jedwards4b: here too

jedwards4b:
I'm confused about the netcdf files in this commit.  Want to try to understand better how they were corrupted.

scripts/create_clone
```diff
@@ -1,318 +1,161 @@
...
+        expect(False, 
+               "Must specify -clone as an input argument")
+
+    return args.case, args.clone, args.keepexe, args.mach_dir, args.project
+
+##############################################################################
+def _main_func():
+###############################################################################
+
+    case, clone, keepexe, mach_dir, project = parse_command_line(sys.argv)
```
jgfouca: I have a fairly large request, can you package all this into a library so that we can access create_clone functionality without having to go to the shell? See check_input_data as an example.

scripts/create_clone
```diff
@@ -1,318 +1,161 @@
...
+
+    # copy user_nl_files
+    cmd = cloneroot + '/user_nl_*'
+    files = glob.glob(cmd)
+    for file in files:
+        cmd = "cp " + file + " " + caseroot + "/."
+        run_cmd(cmd)
+
+    # copy SourceMod files
+    dirs = glob.glob(cloneroot + "/SourceMods/*")
```
jgfouca: You should never use python builtins like "dir" as variable names.

scripts/create_clone
```diff
@@ -1,318 +1,161 @@
...
+    case_setup(caseroot, clean=False, test_mode=False)
+
+    logger.info("Successfully created new case %s from clone case %s " %(case, clone))
+
+
+###############################################################################
+
+if __name__ == "__main__":
+    _main_func()
+
```
jgfouca: you can delete all the lines after 155

utils/python/CIME/case.py
```diff
@@ -77,6 +78,24 @@ def _get_env(self, short_name):
                   return env_file
           expect(False, "Could not find object for %s in case"%full_name)
 
+    def copy(self, newcasename, newcaseroot, newcimeroot=None, newsrcroot=None):
+        newcase = deepcopy(self)
```
jgfouca: This deepcopy is highly suspect. Case is a custom type and python has no idea natively how to deep copy it. See this:
http://stackoverflow.com/questions/1500718/what-is-the-right-way-to-override-the-copy-deepcopy-operations-on-an-object-in-p

Is there any value in making a copy function instead of just creating a new Case object for the clone?

utils/python/CIME/case_setup.py
```diff
@@ -84,7 +84,7 @@ def case_setup(caseroot, clean=False, test_mode=False, reset=False):
 
     cimeroot = os.environ["CIMEROOT"]
 
-    case = Case()
+    case = Case(caseroot)
```
jgfouca: Due to the chdir above, this argument is not necessary. In fact, if caseroot happens to be relative, this change breaks the code.

jgfouca:
Review complete.

utils/python/CIME/case.py
mvertens: I think we should have a discussion about this - neither Jim nor I understand your point. The issue is that we need a copy of the object not a copy of the pointer to the object. Deepcopy should do that - and seems to be working correctly in all of our tests. Looking at the stack overflow reference above, neither Jim or I get it. Could you give us an example of where deepcopy would fail in our case?

jgfouca: Let's first establish that we need to do a copy. Why not just create a new Case object that points to the clone?

jedwards4b: Because you want all of the attributes of the original case.

jedwards4b: http://stackoverflow.com/questions/4005975/etree-clone-node

jgfouca: OK, maybe this does work then. I guess the default implementation of deepcopy is pretty smart.

utils/python/CIME/case_setup.py
mvertens: Actually, this did not work unless I passed caseroot into the Case object for create_clone. It still does not work when I put in the original change. I don't understand why. I have checked that you really are in the caseroot directory but the queries in case_setup all seem to return None.

jgfouca: Take out the caseroot argument and put this above the construction:

    import pdb
    pdb.set_trace()

And then step into the constructor and see what's going on. Pdb is very powerful and you can print any valid python expression as if it occurred in the current stack frame.

mvertens: I had already put in a pull request my work around. I am happy to do the set_trace commands and track this down - but I will not get a chance to do this before Monday.

jgfouca:
@mvertens , is this ready?

mvertens:
Just put in a pull request for this with the fixes put in that Jim
suggested yesterday

jgfouca:
@mvertens thanks. So this is ready? Btw, you didn't put in a new PR, you just added to this one, which is better.

jgfouca:
@mvertens , did you just reply to me by editing my comment? :)

I'm more likely to notice if you make a new post.

PR #143, opened 2016-04-25
Mvertens/cleanbuild

mvertens:
removal of case.clean_build and incorporation of logic into case.build.
New arguments to case.build:  --clean or case.build --cleanall that handles what was previously in case.clean_build


scripts/Tools/case.build
```diff
@@ -54,11 +54,31 @@ OR
     parser.add_argument("-m", "--model-only", action="store_true",
                         help="Assume shared libs already built")
 
+    comps = ("atm", "ice", "glc", "lnd", "ocn", "rof", "wav", "cpl")
```
jedwards4b: What if, instead of hardcoding the list of comps, you build it based on the subdirectories of EXEROOT?

scripts/Tools/check_case
```diff
@@ -1,65 +1,30 @@
...
 
-  checks that case is set up correctly
+logger = logging.getLogger(__name__)
 
-OPTIONS
+parser = argparse.ArgumentParser()
+CIME.utils.setup_standard_logging_options(parser)
 
-  -help [or -h]  Print usage to STDOUT (optional).
+caseroot = os.getcwd()
```
jedwards4b: both of these can be called directly from python.  No use of run_cmd is needed.

utils/python/CIME/XML/env_batch.py
```diff
@@ -11,10 +11,12 @@
 
 class EnvBatch(EnvBase):
 
-    def __init__(self, case_root=os.getcwd(), infile="env_batch.xml"):
```
jedwards4b: Why not just put the case_root initialization in EnvBase and leave it out of the others?

utils/python/CIME/build.py
```diff
@@ -563,3 +563,55 @@ def _build_model_thread(config_dir, caseroot, bldroot, compspec, file_build,
...
+
+    # reset following values in xml files
+    case.set_value("SMP_BUILD",str(0))
+    case.set_value("NINST_BUILD",str(0))
+    case.set_value("BUILD_STATUS",str(0))
+    case.set_value("BUILD_COMPLETE","FALSE")
+    case.flush()
+
+    # append call of to CaseStatus 
+    cmd = "date +\"%Y-%m-%d %H:%M:%S\""
```
jedwards4b: from datetime import datetime
date = datetime.now()

utils/python/CIME/case.py
```diff
@@ -455,16 +484,16 @@ def _create_caseroot_tools(self):
...
-    # print "DEBUG: infofiles are ",infofiles
-    #    try:
-    #        for infofile in infofiles:
-    #            print "DEBUG: infofile is %s, %s"  %(infofile, os.path.basename(infofile))
-    #            dst_file = caseroot + "/" + os.path.basename(infofile)
-    #            copyfile(infofile, dst_file)
-    #            os.chmod(dst_file, os.stat(dst_file).st_mode | stat.S_IXUSR | stat.S_IXGRP)
-    #    except Exception as e:
-    #        logger.warning("FAILED to set up infofiles: %s" % str(e))
+            #FIXME - the following does not work
```
jedwards4b: fixme now?

scripts/Testing/Testcases/NCR_build.csh
```diff
@@ -85,7 +85,7 @@ set NTASKS_CPL  = `./xmlquery NTASKS_CPL  --value`
 ./case.setup --clean --test-mode
 ./case.setup
 
-./case.clean_build
+./case.build --clean
```
jgfouca: these testcases_build files are no longer used, right?

scripts/Tools/case.build
```diff
@@ -54,11 +54,31 @@ OR
...
+    parser.add_argument("--clean", nargs="*", choices=allobjs,
+                        help="objects to clean"
+                        "if no arguments then clean only atm, ice, glc, lnd, ocn, rof, wav objects")
+
+    parser.add_argument("--cleanall", action="store_true",
+                        help="clean all objects ")
+
     args = parser.parse_args(args[1:])
 
+    cleanlist = []
```
jgfouca: could be slightly cleaner with a ternary:

    cleanlist = allobjs if args.cleanall else p[

scripts/Tools/case.build
```diff
@@ -54,11 +54,31 @@ OR
...
+    if args.cleanall:
+        cleanlist = allobjs
+    if args.clean is not None:
+        if args.clean:
+            cleanlist = args.clean
```
jgfouca: Same, use ternary

scripts/Tools/check_case
jgfouca: Strongly agree. I think most of the performance issues with the perl scripts was the tendency to use shell commands to run things which then typically required a ton of repeated work like reparsing all the XML.

scripts/Tools/check_case
```diff
@@ -1,65 +1,30 @@
...
 
-  check_case [options]
+from CIME.utils import expect, get_model, run_cmd
+from CIME.case  import Case
 
-  checks that case is set up correctly
+logger = logging.getLogger(__name__)
 
-OPTIONS
+parser = argparse.ArgumentParser()
```
jgfouca: I'd prefer if this tool followed the convention of our other tools, even if check_case is pretty simple. I will refer you to scripts/Tools/check_input_data as a simple example of what I think we want every tool to look like. The exe will have a parse_command_line function and a _main_func that will make a call to a library to do the actual work.

utils/python/CIME/XML/env_batch.py
jgfouca: Agree.

utils/python/CIME/build.py
jgfouca: Yes, or use the method out of utils.py that provides the standard CIME timestamp string.

scripts/Testing/Testcases/NCR_build.csh
jedwards4b: Yes these are removed in my branch - this should give a merge conflict.

scripts/Tools/case.build
```diff
@@ -54,11 +54,31 @@ OR
...
 
+    comps = ("atm", "ice", "glc", "lnd", "ocn", "rof", "wav", "cpl")
+    libs  = ("shr", "mct", "pio", "tim")
+    allobjs = comps + libs
+
+    parser.add_argument("--clean", nargs="*", choices=allobjs,
+                        help="objects to clean"
+                        "if no arguments then clean only atm, ice, glc, lnd, ocn, rof, wav objects")
+
+    parser.add_argument("--cleanall", action="store_true",
```
jgfouca: I'd prefer --clean-all to --cleanall

jgfouca:
@mvertens , review complete.

utils/python/CIME/XML/env_batch.py
mvertens: Done

scripts/Testing/Testcases/NCR_build.csh
mvertens: I have rebased to the head of Jim's branch - so these are going away. I will then rebase to master before issuing a new pull request.

scripts/Tools/case.build
mvertens: Done

scripts/Tools/case.build
mvertens: Done

scripts/Tools/case.build
mvertens: Done

scripts/Tools/check_case
mvertens: Done. This makes sense now. Thanks for the pointer to check_input_data.

scripts/Tools/check_case
mvertens: Done. Good point.

jgfouca:
@mvertens , is this ready? I see a lot of "Done" comments but no new commits.

jgfouca:
Looks great! Merging.

PR #146, opened 2016-04-27
Testcreatenewcase

jedwards4b:
Adds a test for create_newcase, orders tests in scripts_regression_tests and fixes some issues with PBS batch systems.

### DIFF
--- a/cime_config/acme/machines/python.case.test
+++ b/cime_config/acme/machines/python.case.test
@@ -4,6 +4,10 @@
  This is the system test submit script for CIME
 
 """
+import os
+workdir = os.environ.get("PBS_O_WORKDIR")
+if workdir is not None:
+    os.chdir(workdir)
 
 from standard_script_setup import *
 
--- a/cime_config/acme/machines/template.case.run
+++ b/cime_config/acme/machines/template.case.run
@@ -226,8 +226,6 @@ sub doPreRunChecks()
 # -------------------------------------------------------------------------
 sub runModel()
 {
-    chdir $config{'RUNDIR'};
-
     system("sleep 10");
     my $localtime = localtime;
     # Set OMP_NUM_THREADS
@@ -235,7 +233,13 @@ sub runModel()
 
     $logger->info("$localtime MODEL EXECUTION BEGINS HERE");
     # Run the model
-    {{ mpirun }}
+    my $cmd = "{{ mpirun }}";
+    $cmd = SetupTools::expand_xml_var($cmd, \%config);
+
+    chdir $config{'RUNDIR'};
+    $logger->info("Running '$cmd'\n");
+    system($cmd);
+    
     $localtime = localtime;
     $logger->info( "$localtime MODEL EXECUTION HAS FINISHED");
 }
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -32,7 +32,7 @@ def parse_command_line(args, cimeroot):
                         help="(required) Specify a model grid resolution. "
                         "To see list of current compsets, use the utility manage_case in this directory")
 
-    parser.add_argument("--mach", "-mach", required=True,
+    parser.add_argument("--mach", "-mach", 
                         help="(required) Specify a machine. "
                         "To see list of current compsets, use the utility manage_case in this directory")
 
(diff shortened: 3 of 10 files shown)


cime_config/acme/machines/python.case.test
```diff
@@ -4,6 +4,10 @@
  This is the system test submit script for CIME
 
 """
+import os
```
jedwards4b: PBS systems do not default to the cwd on batch submission so you need to change to that directory.

cime_config/acme/machines/template.case.run
```diff
@@ -226,16 +226,20 @@ sub doPreRunChecks()
...
-    chdir $config{'RUNDIR'};
-
     system("sleep 10");
     my $localtime = localtime;
     # Set OMP_NUM_THREADS
     $ENV{'OMP_NUM_THREADS'} = {{ thread_count }};
 
     $logger->info("$localtime MODEL EXECUTION BEGINS HERE");
     # Run the model
-    {{ mpirun }}
```
jedwards4b: Some of the variables in the {{ mpirun }} statement cannot be resolved until you are on the compute node, this change delays getting the resolved values of variables to solve that problem.

scripts/create_newcase
```diff
@@ -148,12 +148,9 @@ def _main_func():
     caseobj.set_value("CASE", os.path.basename(case))
     caseobj.set_value("CASEROOT", caseroot)
     caseobj.set_value("SRCROOT", srcroot)
-    caseobj.set_value("MACH", machine)
-    if project:
-        caseobj.set_value("PROJECT",project)
 
     # Configure the Case
-    caseobj.configure(compset, grid, machine,
+    caseobj.configure(compset, grid, machine_name=machine, project=project,
```
jedwards4b: machine argument does not need to be required.

utils/python/CIME/XML/env_archive.py
```diff
@@ -22,7 +22,7 @@ def __init__(self, case_root=None, infile="env_archive.xml"):
         else:
             fullpath = os.path.join(case_root, infile)
         GenericXML.__init__(self, fullpath)
-        if not os.path.isfile(infile):
+        if not os.path.isfile(fullpath):
```
jedwards4b: use of relative path infile instead of fullpath was causing the env_archive object to be appended each time init was called.

utils/python/CIME/XML/machines.py
```diff
@@ -354,9 +354,7 @@ def get_mpirun(self, attribs, batch_maker):
         if arg_node is not None:
             arg_nodes = self.get_nodes("arg", root=arg_node)
             for arg_node in arg_nodes:
-                arg_value = self.get_resolved_value(arg_node.text)
```
jedwards4b: Do not resolve yet,  need to wait until on compute node.

utils/python/CIME/case.py
```diff
@@ -297,8 +297,8 @@ def _get_component_config_data(self):
             if result is not None:
                 del self.lookups[key]
 
-    def configure(self, compset_name, grid_name, machine_name,
-                  pecount=None, compiler=None, mpilib=None,
+    def configure(self, compset_name, grid_name, machine_name=None,
+                  project=None, pecount=None, compiler=None, mpilib=None,
```
jedwards4b: Add project argument, make machine optional.

utils/python/tests/scripts_regression_tests
```diff
@@ -23,7 +23,7 @@ NO_BATCH    = False
 logging.basicConfig()
 
 ###############################################################################
-class RunUnitTests(unittest.TestCase):
+class A_RunUnitTests(unittest.TestCase):
```
jedwards4b: Tests are run in sort order of test names - I've added some letters to the beginning of classes to control order a little.  There is no dependence between tests but faster smaller ones are run first - if they fail you can probably abort and fix them before going on to the larger slower ones.

utils/python/tests/scripts_regression_tests
```diff
@@ -130,7 +130,33 @@ def setup_proxy():
     return False
 
 ###############################################################################
-class TestWaitForTests(unittest.TestCase):
+class B_TestCreateNewcase(unittest.TestCase):
```
jedwards4b: because create_test supplies a number of optional arguments to create_newcase we were missing several problems that this test will pick up.

jedwards4b:
Annotated.

utils/python/tests/scripts_regression_tests
jgfouca: Nice.

jgfouca:
Looks good.

PR #156, opened 2016-05-04
Case.submit

jedwards4b:
Redo case.submit in python


cime_config/acme/machines/config_batch.xml
```diff
@@ -210,23 +210,32 @@
...
-     <job name="case.st_archive">
-       <template>template.st_archive</template>
-       <task_count>1</task_count>
-     </job>
-     <job name="case.lt_archive">
-       <template>template.lt_archive</template>
-       <task_count>1</task_count>
-     </job>
-   </batch_jobs>
+  <batch_jobs>
```
jedwards4b: This puts all of the interdependancy information for template files in config_batch.xml and removes it from code.  If you want to add an additional template you should be able to it by only adding code here.

cime_config/acme/machines/template.case.run
```diff
@@ -5,12 +5,20 @@ template to create a case run script
 """
 
 import sys, os, calendar, shutil, time
+_CIMEROOT = os.environ.get("CIMEROOT")
```
jedwards4b: These lines need to be above the imports because they set the path to import from

cime_config/acme/machines/template.case.run
```diff
@@ -112,7 +113,8 @@ def preRunCheck(case):
     preview_namelists()
 
     # document process
-    appendCaseStatus(caseroot, "%s Run started \n" %(time.strftime("%Y-%m-%d %H:%M:%S")))
```
jedwards4b: I reworked this to allow appending any file and use it for all of CaseStatus, TestStatus and TestStatus.log
CaseStatus and TestStatus.log get a timestamp prepended to each entry.

cime_config/acme/machines/template.lt_archive
```diff
@@ -85,19 +79,7 @@ def _main_func():
                "Unable to perform long term archive...")
 
     # document completion
-    if os.path.isfile(testlog):
-        with open(testlog, "a") as fd:
-            fd.write("lg_archive completed %s" %(time.strftime("%Y-%m-%d %H:%M:%S")))
-
-    # resubmit case if appropriate
-    resubmit = int(case.get_value("RESUBMIT"))
```
jedwards4b: Resubmit from lt_archive doesn't make sense - it was a bug in the perl code

cime_config/cesm/machines/config_batch.xml
```diff
@@ -41,10 +41,10 @@
      <depend_string> --dependencies</depend_string>
      <submit_args>
        <arg flag="--cwd" name="cwd"/>
-       <arg flag="-A" name="project"/>
```
jedwards4b: Rename these to use the same name as the XML variable, much easier to parse.

driver_cpl/cime_config/config_component.xml
```diff
@@ -1849,13 +1849,13 @@
     </desc>
   </entry>
 
-  <entry id="TESTS">
-    <type>char</type>
-    <default_value>UNSET</default_value>
+  <entry id="TEST">
```
jedwards4b: This variable was not being used, i've repurposed it to indicate that the case is a TEST

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -40,45 +41,49 @@ def __init__(self, caseroot=os.getcwd(), case=None):
...
-        teststatusfile = result.join(li)
-        with open("TestStatus", 'w') as f:
-            f.write(teststatusfile)
+            self._runstatus = "FAIL"
+        if out:
+            appendStatus("case.run output is:\n %s\n"%out, sfile="TestStatus.log")
+        if err:
+            appendStatus("case.run error is:\n %s\n"%err, sfile="TestStatus.log")
+
+    def __del__(self):
```
jedwards4b: Only update the runstatus when the test object is destroyed

utils/python/CIME/env_module.py
```diff
@@ -23,7 +21,10 @@ def __init__(self, machine, compiler, cimeroot, caseroot, mpilib, debug=False):
         self._caseroot = caseroot
         self._mpilib   = mpilib
         self._debug    = debug
-
+        self._cshscript = list()
```
jedwards4b: Add write of .env_mach_specific.csh and .sh - this is needed for tests that are still in csh but it's also a really handy file to have for debugging.

utils/python/update_acme_tests.py
```diff
@@ -58,7 +58,8 @@
                              "SEQ_Ld3.f09_g16.X",
                              "ERS_Ld3.ne30_g16_rx1.A",
                              "ERS_N2_Ld3.f19_g16_rx1.A",
-                             "ERR_Ld3.f45_g37_rx1.A")
+                             "ERR_Ld3.f45_g37_rx1.A",
+                             "SMS_D_Ln9_Mmpi-serial.f19_g16_rx1.A")
```
jedwards4b: This test currently fails.

jedwards4b:
Annotation complete.

cime_config/acme/machines/template.case.run
jgfouca: Nice

utils/python/CIME/XML/env_batch.py
```diff
@@ -67,12 +69,15 @@ def get_type_info(self, vid):
         return type_info
 
     def get_jobs(self):
-        result = []
```
jgfouca: Any reason to prefer list() over []?

utils/python/CIME/SystemTests/system_tests_common.py
jgfouca: Seems like a good solution. There's a pretty wide timeframe of when this might be called, but it should be guaranteed to be called before the process exits.

utils/python/CIME/XML/env_batch.py
jedwards4b: none, I will revert this

jgfouca:
Looks good.

PR #168, opened 2016-05-09
Mvertens/starchive

mvertens:
Pythonized short term archiving utility. Slightly modified the archive.xml file to fix some problems for short term archiving of A compsets. Assuming still that template.st_archive and archive.xml are the same in acme and cesm. This could potentially be moved to a shared location in the near future.


cime_config/acme/machines/template.st_archive
```diff
@@ -16,53 +127,58 @@ from CIME.case_submit               import submit
...
+        description=description,
+        formatter_class=argparse.RawDescriptionHelpFormatter
+    )
 
     CIME.utils.setup_standard_logging_options(parser)
 
     parser.add_argument("--caseroot", default=os.getcwd(),
                         help="Case directory to build")
 
+    parser.add_argument("--input", "-in", action="store_true",
```
jedwards4b: There may be a problem here - it's not easy to pass command line arguments to a script submitted to the batch system - are these arguments intended only for using the script from the command line?

utils/python/CIME/st_archive.py
```diff
@@ -0,0 +1,190 @@
...
+        if datename in filename:
+            counter = counter + 1
+            if counter > 1:
+                expect(False, 'Multiple restart files found for suffix %s ' % suffix)
+            restfile = basename(filename)
+            if isfile(join(destdir, restfile)):
+                shutil.copy(join(destdir, restfile), join(restdir, restfile))
+
+
+def archiveProcess(case, archive, datename, runfiles):
```
jgfouca: I'd prefer "function_name_style" instead of functionNameStyle in order to be consistent with the rest of the codebase.

rljacob:
This will be redone.

PR #170, opened 2016-05-09
Cesm esp merge

jedwards4b:
scripts regression tests pass except ERS_N2_Ld3.f19_g16_rx1.A.yellowstone_intel
which is also currently failing on cesm-development.    I wanted to put this up for review bug fix the test before it's merged.


jgfouca:
@jedwards4b , this was way too much to review. GitHub wouldn't even should be the changes to the python scripts. I assume it was mostly whitespace?

rljacob:
Is this really just the mods for the ESP component?

PR #171, opened 2016-05-10
Single submit capability for create_test!

jgfouca:
Adds the --single-submit flag to create_test. Allows batch machines
to submit a single job to the queue to run all the tests. Should
open the possibility of reducing queue wait times.

Also, allows create_test to use pre-existing case directories if
test-id matches.

Change list:
1) Fix spelling errors 'dependancy'
2) Allow create_test to reuse case dirs. It will pick up where it left off.
3) Single submit capability
4) Users can now choose their processor pool size
5) BatchMaker is now optional when getting batch directives.
6) Some cleanup/code reduction in batch.py
7) Encapsulate queue selecion in machines.py
8) Remove unused vars in system_test.py
9) Added a seconds->HH:MM:SS converter function in utils
10) Added a time-estimator functions in utils for single-submit jobs

### DIFF
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -11,10 +11,12 @@ If this tool is missing any feature that you need, please notify jgfouca@sandia.
 from Tools.standard_script_setup import *
 
 import update_acme_tests
-from CIME.system_test import SystemTest
-from CIME.utils import expect
+from CIME.system_test import SystemTest, RUN_PHASE
+from CIME.utils import expect, convert_to_seconds, compute_total_time, convert_to_babylonian_time, run_cmd
+from CIME.XML.machines import Machines
+from CIME.XML.batch import Batch
 
-import argparse, doctest, tempfile
+import argparse, doctest, tempfile, math
 
 ###############################################################################
 def parse_command_line(args, description):
@@ -69,9 +71,16 @@ OR
     parser.add_argument("--no-build", action="store_true",
                         help="Do not build generated tests, implies --no-run")
 
+    parser.add_argument("-u", "--use-existing", action="store_true",
+                        help="Use pre-existing case directories. Requires test-id")
+
     parser.add_argument("--no-batch", action="store_true",
                         help="Do not submit jobs to batch system, run locally."
                         " If false, will default to machine setting.")
+
+    parser.add_argument("--single-submit", action="store_true",
+                        help="Use a single interactive allocation to run all the tests. "
+                        "Can drastically reduce queue waiting. Only makes sense on batch machines.")
 
     parser.add_argument("-r", "--test-root",
                         help="Where test cases will be created."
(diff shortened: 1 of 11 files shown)


scripts/create_test
```diff
@@ -174,19 +191,113 @@ OR
...
+    if no_arg_is_a_proc_pool_arg:
+        new_args.append("--proc-pool %d" % proc_pool)
+    if no_arg_is_a_machine_arg:
+        new_args.append("-m %s" % machine_name)
+
+    #
+    # Resolve batch directives manually. There is currently no other way
+    # to do this without making a Case object.
+    #
+    directives = "\n".join(batch.get_batch_directives())
```
jgfouca: All the batch stuff is so coupled to Case that I had to do this myself.

scripts/create_test
```diff
@@ -211,8 +323,38 @@ def create_test(testargs, compiler, machine_name, no_run, no_build, no_batch, te
...
+                time_needed = output.split('value="')[1].replace('">', "")
+                time_needed = convert_to_seconds(time_needed)
+            else:
+                time_needed = 60 # HACK
+
+            job_cost_map[test] = (procs_needed, time_needed)
+            if procs_needed > largest_case:
+                largest_case = procs_needed
+
+        if proc_pool is None:
```
jgfouca: Not sure if this is the right default proc_pool

scripts/create_test
```diff
@@ -211,8 +323,38 @@ def create_test(testargs, compiler, machine_name, no_run, no_build, no_batch, te
...
+                      xml_category=xml_category, xml_testlist=xml_testlist, walltime=walltime,
+                      proc_pool=proc_pool, use_existing=use_existing)
+
+    success = impl.system_test()
+
+    if single_submit:
+        job_cost_map = {}
+        largest_case = 0
+        for test in impl._tests:
+            test_dir = impl._get_test_dir(test)
```
jedwards4b: Why not just
case = Case(test_dir)
wallclock_time = case.get_value("JOB_WALLCLOCK_TIME",subgroup="case.test")

jedwards4b:
Review Complete, question:   You are creating a proc_pool which should have more tasks than the largest test - and then running all of the tests within that pool?   How do you control the number of mpi tasks used by each test?  I didn't see that change here.

jgfouca:
@jedwards4b the number of mpi tasks per case should be the same either way (with or without single submit). The one thing I'm concerning about is the build phase. create_test is going to think it has a huge proc pool and will try to build everything at once.

jedwards4b:
The number of mpi tasks is currently controlled in the batch directives, in order for this to work you need to move that control to the mpiexec command.

jgfouca:
Doesn't the mpiexec always have to have an -np N argument?

jgfouca:
I see this in case.run for one of the tests:
```
case.run:    cmd ="mpirun  -np 8  -npernode 8 ${EXEROOT}/acme.exe  >> acme.log.$LID 2>&1  " 
```

jedwards4b:
It depends on the mpi implementation.  I know that LSF, the system on yellowstone isn't going to be easy, but then we don't have to use this method for testing so maybe it doesn't matter.

jgfouca:
Yes, this is an experimental feature. As long as most MPI implementation support this, I'm fine with it.

PR #175, opened 2016-05-13
Update case.py - removed lookups

### LINKED ISSUE #173: Debug new xmlquery
./create_test cime_developer has issues.

wilke:
get_values was checking self.lookups in addition to xml 
but not checking for duplicates or converting string into dict 

removed statement since get_values and get_value should only operate on xml structure

Closes #173

### DIFF
--- a/utils/python/CIME/case.py
+++ b/utils/python/CIME/case.py
@@ -154,8 +154,6 @@ class Case(object):
         # Empty result list
         results = []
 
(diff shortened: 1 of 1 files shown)


wilke:
Closes #173

rljacob:
The "closes" keyword has to be in the body of the PR.   https://github.com/blog/1506-closing-issues-via-pull-requests

PR #177, opened 2016-05-16
Jgfouca/some fixups

jedwards4b:
Fixes issues with xmlquery and some batch system problems.   

Tested on Blues in interactive queue using no batch option - all tests pass.


utils/python/compare_namelists.py
```diff
@@ -215,6 +214,13 @@ def compare_values(namelist, name, gold_value, comp_value, case):
     """
     Compare values for a specific variable in a namelist.
     """
+# skip these for now
+    if name.startswith("pio") or name.startswith("esp") or name.startswith("prof") or\
```
jedwards4b: This is debug stuff - I will remove from the PR

scripts/Tools/xmlquery
```diff
@@ -1,148 +1,135 @@
...
+    parser.add_argument("-value", "--value", default=False, action="store_true",
                         help="Only print the value.")
-    parser.add_argument("-resolve", "--resolve", default=False, action="store_true", 
-                        help="Resolve variable names")                    
+    parser.add_argument("-resolve", "--resolve", default=True, action="store_true",
+                        help="Resolve variable names")
 
     args = parser.parse_args(args[1:])
-
+    args.debug = True
```
jgfouca: @wilke This seems like debug code that was accidentally left in.

jedwards4b: Thanks  - I was looking for that - I'll remove it.

PR #178, opened 2016-05-16
Gets ERR working on melvin.

jgfouca:

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes: [CIME Github issue #]

User interface changes?: 

Code review: 

Conversion of XML nodes to set and then back to list is not order
preserving. This commit removes this conversion.

Also, some minor refactor of batch_utils.

### DIFF
--- a/utils/python/CIME/batch_utils.py
+++ b/utils/python/CIME/batch_utils.py
@@ -12,7 +12,7 @@ logger = logging.getLogger(__name__)
 
 class BatchUtils(object):
 
-    def __init__(self, job, case=None, prereq_jobid=None):
+    def __init__(self, job, case=None, prereq_jobid=None, no_batch=False):
         """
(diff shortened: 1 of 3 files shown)


utils/python/CIME/batch_utils.py
```diff
@@ -12,29 +12,25 @@
...
         self.case = case if case is not None else Case()
         self.job = job
-        self.batchobj = None
         self.override_node_count = None
         self.batchmaker = None
         self.jobs = []
         self.prereq_jobid = prereq_jobid
 
-    def submit_jobs(self, no_batch=False):
-        if no_batch:
```
jgfouca: The old code was not safe. If no_batch is changed between calls, self.batchobj would not be updated.

jedwards4b:
Looks good, thanks.

PR #180, opened 2016-05-16
Cesm component testing3

jedwards4b:
Make sure that wallclock and queue information are written by create_newcase

Test suite: scripts regression on yellowstone cesm - all pass
Test baseline: 
Test namelist changes: 

Fixes: 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -261,9 +261,8 @@ def single_submit_impl(machine_name, test_id, proc_pool, project, args, job_cost
     else:
         wall_time_bab = wall_time
 
-    queue_node = mach.select_best_queue(proc_pool)
-    queue = queue_node.text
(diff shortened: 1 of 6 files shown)


scripts/create_test
```diff
@@ -261,9 +261,8 @@ def single_submit_impl(machine_name, test_id, proc_pool, project, args, job_cost
     else:
         wall_time_bab = wall_time
 
-    queue_node = mach.select_best_queue(proc_pool)
-    queue = queue_node.text
-    wall_time_max_bab = queue_node.get("walltimemax")
+    queue = mach.select_best_queue(proc_pool)
```
jedwards4b: I think that it's better to keep the xml in the xml modules.

jgfouca: I agree.

PR #183, opened 2016-05-17
Erstestrefactor

jgfouca:
Convert ERS and SMS scripts to python. Some other minor formatting changes.


utils/python/CIME/SystemTests/ers.py
```diff
@@ -12,12 +15,37 @@ def __init__(self, caseroot, case):
         SystemTestsCommon.__init__(self, caseroot, case)
 
     def run(self):
-        self._case.set_value("CONTINUE_RUN",False)
-        self._case.set_value("REST_OPTION","none")
-        self._case.set_value("HIST_OPTION","$STOP_OPTION")
-        self._case.set_value("HIST_N","$STOP_N")
+        stop_n = self._case.get_value("STOP_N")
```
jedwards4b: We may want to add a few sanity checks here - for example make sure that STOP_N and REST_N are > 0.

jedwards4b:
Looks good.

utils/python/CIME/SystemTests/ers.py
jgfouca: Done.

PR #184, opened 2016-05-18
Convert SEQ testcase script to python.

jgfouca:
Other changes:
1) Try to have return codes indicate success of case.test
2) Fixes to ERS
3) Fixes to system_tests_common.py that were hiding many errors


jgfouca:
Not ready to merge.

cime_config/acme/machines/template.case.test
```diff
@@ -77,14 +77,16 @@ def cimetest(caseroot, testname=None):
...
+
 def _main_func(description):
     if ("--test" in sys.argv):
         test_results = doctest.testmod(verbose=True)
         sys.exit(1 if test_results.failed > 0 else 0)
 
     caseroot, testname = parse_command_line(sys.argv, description)
 
-    cimetest(caseroot, testname)
+    sys.exit(0 if cimetest(caseroot, testname) else 1)
```
jgfouca: Trying to leverage error codes to indicate success.

utils/python/CIME/SystemTests/ers.py
```diff
@@ -15,37 +15,32 @@ def __init__(self, caseroot, case):
...
         logger.info("doing an %s %s initial test with restart file at %s %s"
                     %(str(stop_n), stop_option, str(rest_n), stop_option))
-        SystemTestsCommon.run(self)
+        success = SystemTestsCommon.run(self)
 
-        expect(self.coupler_log_indicates_run_complete(),
-               'ERROR: First run of ERS test Failed to complete')
-        self._case.set_value("STOP_N", stop_n - rest_n)
-        self._case.set_value("CONTINUE_RUN",True)
-        self._case.set_value("REST_OPTION","never")
```
jgfouca: Missing flush in old version.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -49,17 +49,28 @@ def build(self, sharedlib_only=False, model_only=False):
                          sharedlib_only=sharedlib_only, model_only=model_only)
 
     def run(self):
+        return self._run()
+
+    def _run(self, suffix="base"):
         rc, out, err = run_cmd("./case.run --caseroot %s" % self._caseroot, ok_to_fail=True)
-        if rc == 0 and self.coupler_log_indicates_run_complete():
```
jgfouca: In a multirun testcase, self._runstatus could have had a FAIL masked by a later PASS

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -82,16 +93,43 @@ def coupler_log_indicates_run_complete(self):
         return False
 
     def report(self):
-        newestcpllogfile = self._getlatestcpllog()
+        newestcpllogfile = self._get_latest_cpl_log()
         self._check_for_memleak(newestcpllogfile)
 
+    def _component_compare_move(self, suffix):
```
jgfouca: encapsulate component_compare_move.sh calls and component_compare_test.sh calls in these two methods. We can switch these to pure python later.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -82,16 +93,43 @@ def coupler_log_indicates_run_complete(self):
...
+
+        return True
+
     def _get_mem_usage(self, cpllog):
         """
         Examine memory usage as recorded in the cpl log file and look for unexpected
         increases.
         """
         memlist = []
-        meminfo = re.compile(".*model date =\s+(\w+).*memory =\s+(\d+\.?\d+).*highwater")
```
jgfouca: The old regex was not correct. You need to use a raw string to express a string with backslashes in it; otherwise the backslashses themselves must be escaped.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -140,24 +178,27 @@ def compare_env_run(self, expected=None):
...
-    def _getlatestcpllog(self):
+    def _get_latest_cpl_log(self):
         """
         find and return the latest cpl log file in the run directory
         """
         cpllog = None
         cpllogs = glob.glob(os.path.join(
                     self._case.get_value('RUNDIR'),'cpl.log.*'))
         if cpllogs:
-            cpllog = min(cpllogs, key=os.path.getctime)
```
jgfouca: This was a serious bug that hid errrors.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -49,17 +49,28 @@ def build(self, sharedlib_only=False, model_only=False):
...
         else:
+            success = False
             self._runstatus = "FAIL"
 
         if out:
             append_status("case.run output is:\n %s\n"%out, sfile="TestStatus.log")
         if err:
             append_status("case.run error is:\n %s\n"%err, sfile="TestStatus.log")
 
+        if success and suffix is not None:
```
jgfouca: Successful runs often need to change the name of the hist files to something so subsequent runs can be compared.

utils/python/CIME/SystemTests/ers.py
```diff
@@ -15,37 +15,32 @@ def __init__(self, caseroot, case):
...
     def run(self):
-        stop_n = self._case.get_value("STOP_N")
+        stop_n      = self._case.get_value("STOP_N")
         stop_option = self._case.get_value("STOP_OPTION")
+        expect(stop_n > 0, "Bad STOP_N: %d" % stop_n)
+
         rest_n = stop_n/2 + 1
         self._case.set_value("REST_N",rest_n)
         self._case.flush()
         expect(stop_n > 2, "ERROR: stop_n value %d too short"%stop_n)
```
jedwards4b: Here shouldn't you check rest_n not stop_n which you just checked above?

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -82,16 +93,43 @@ def coupler_log_indicates_run_complete(self):
...
+        return True
+
     def _get_mem_usage(self, cpllog):
         """
         Examine memory usage as recorded in the cpl log file and look for unexpected
         increases.
         """
         memlist = []
-        meminfo = re.compile(".*model date =\s+(\w+).*memory =\s+(\d+\.?\d+).*highwater")
+        meminfo = re.compile(r".*model date =\s+(\w+).*memory =\s+(\d+\.?\d+).*highwater")
```
jedwards4b: what is the r for?

utils/python/CIME/SystemTests/system_tests_common.py
jedwards4b: Really?  How'd I miss that?

utils/python/CIME/SystemTests/system_tests_common.py
jgfouca: "raw" string

utils/python/CIME/SystemTests/system_tests_common.py
jgfouca: Error hiding is easy to miss because things appear to work.

cime_config/acme/machines/template.case.test
jedwards4b: you missed doing this on the cesm side

scripts/Tools/normalize_cases
```diff
@@ -52,7 +52,7 @@ def normalize_cases(case1, case2):
         for log_dir in ["bld", "run"]:
             gzips = glob.glob(os.path.join(case_dir, log_dir, "*.gz"))
             if (gzips):
-                run_cmd("gunzip %s" % " ".join(gzips))
```
jedwards4b: We should use the python gzip instead of run_cmd here.

jedwards4b:
A good test to do when you are converting these tests is - can you run the same test case twice?  That is after a test completes you should be able to go to the case directory and run it again and get the same result.

PR #192, opened 2016-05-23
Cesm component tests4

jedwards4b:
Fix a few more issues found in cesm component tests.  
Test suite:  scripts_regression_tests, several cesm component tests
Test baseline: cesm1_5_alpha06i
Test namelist changes: 
Test status: bit for bit

Fixes:

User interface changes?: no

Code review:

### DIFF
--- a/utils/python/CIME/system_test.py
+++ b/utils/python/CIME/system_test.py
@@ -386,6 +386,10 @@ class SystemTest(object):
                     ninst = case_opt[1:]
                     create_newcase_cmd += " --ninst %s" %ninst
                     logger.debug (" NINST set to %s" % ninst)
+                pesize = re.match('P([SMLX][12]?)', case_opt)
+                if pesize:
+                    create_newcase_cmd += " --pecount %s"%pesize.group(1)
+
 
         logger.debug("Calling create_newcase: " + create_newcase_cmd)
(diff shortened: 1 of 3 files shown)


utils/python/CIME/system_test.py
```diff
@@ -489,24 +496,27 @@ def _xml_phase(self, test):
...
+                            envtest.set_test_parameter(string, opti_tasks)
+                            string = "NTHRDS_" + component_class
+                            envtest.set_test_parameter(string, opti_thrds)
+                            string = "ROOTPE_" + component_class
+                            envtest.set_test_parameter(string, str(0))
+                    elif match3:
+
+                        # handled by create_newcase
+                        continue
+                    if not match3:
```
jgfouca: I think you want an "else" here if you want to error-out if nothing matched.

jgfouca:
Minor fix needed before this can be merged.

utils/python/CIME/system_test.py
jedwards4b: I think that I have it correct - I want to run the expect code in any case except when match3 is true.

jgfouca: OK, I see what you're doing now.

PR #194, opened 2016-05-25
Seq refactor

jedwards4b:
Refactor the SEQ test so that it works regardless of initial PE layout.  

Test suite: cime_developer 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes: 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/Testing/Testcases/config_tests.xml
+++ b/scripts/Testing/Testcases/config_tests.xml
@@ -276,6 +276,7 @@
   <test NAME="SEQ">
     <DESC>sequencing bfb test (10 day seq,conc tests)</DESC>
     <INFO_DBUG>1</INFO_DBUG>
+    <BFBFLAG>TRUE</BFBFLAG>
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>10</STOP_N>
     <REST_OPTION>never</REST_OPTION>
--- a/utils/python/CIME/SystemTests/seq.py
+++ b/utils/python/CIME/SystemTests/seq.py
@@ -49,7 +49,7 @@ class SEQ(SystemTestsCommon):
                 # here we set the cpl to have the first 2 tasks
                 # and each component to have a different ROOTPE
                 if comp == "DRV":
-                    self._case.set_value("NTASKS_CPL"%comp, 2)
+                    self._case.set_value("NTASKS_CPL", 2)
                 else:
                     ntasks = self._case.get_value("NTASKS_%s"%comp)
                     if ntasks > 1:
--- a/utils/python/CIME/SystemTests/seq.py
+++ b/utils/python/CIME/SystemTests/seq.py
@@ -3,7 +3,7 @@ CIME smoke test  This class inherits from SystemTestsCommon
 """
 from CIME.XML.standard_module_setup import *
 from system_tests_common import SystemTestsCommon
-
+from CIME.case_setup import case_setup
 import shutil
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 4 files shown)


scripts/Testing/Testcases/config_tests.xml
```diff
@@ -276,6 +276,7 @@
   <test NAME="SEQ">
     <DESC>sequencing bfb test (10 day seq,conc tests)</DESC>
     <INFO_DBUG>1</INFO_DBUG>
+    <BFBFLAG>TRUE</BFBFLAG>
```
jgfouca: Who uses this flag?

utils/python/CIME/SystemTests/seq.py
```diff
@@ -14,7 +14,59 @@ def __init__(self, caseroot, case):
         """
         initialize an object interface to file env_test.xml in the case directory
         """
-        SystemTestsCommon.__init__(self, caseroot, case)
+        SystemTestsCommon.__init__(self, caseroot, case, expected=["TEST"])
```
jgfouca: What does this new expected argument do?

utils/python/CIME/SystemTests/seq.py
```diff
@@ -26,26 +78,27 @@ def run(self):
...
-            for comp in ['ATM','CPL','OCN','WAV','GLC','ICE','ROF','LND', 'ESP']:
-                any_changes |= self._case.get_value("ROOTPE_%s" % comp) != 0
-                self._case.set_value("ROOTPE_%s" % comp, 0)
-
-            expect(any_changes, "All ROOTPEs were already zero, we aren't testing anything")
+        shutil.copy("%s/%s.exe.SEQ1"%(exeroot,cime_model),
+                    "%s/%s.exe"%(exeroot,cime_model))
+        shutil.copy(os.path.join("LockedFiles", "env_mach_pes.SEQ1.xml"), "env_mach_pes.xml")
+        shutil.copy("env_mach_pes.xml", os.path.join("LockedFiles", "env_mach_pes.xml"))
+        success = SystemTestsCommon._run(self)
```
jgfouca: If success is false, the test should probably halt immediately. That's how I did ERS.

utils/python/update_acme_tests.py
```diff
@@ -64,8 +64,8 @@
 
     "cime_developer" : (None,
                             ("NCK_Ld3.f45_g37_rx1.A",
-                             "ERI_D.f45_g37_rx1.A",
-                             "SEQ_Ld3.f09_g16.X",
+                             "ERI_D.f45_g37.X",
```
jgfouca: Why did these need to change?

jgfouca:
Review complete. One code change requested and some questions.

scripts/Testing/Testcases/config_tests.xml
jedwards4b: The coupler - if the number of coupler tasks change between tests this needs to be set to true to get bfb results.

utils/python/CIME/SystemTests/seq.py
jedwards4b: See my comments in the slack general.   It causes a comparison of env_run.xml with the original to reset this field to the original value.

utils/python/CIME/SystemTests/seq.py
jedwards4b: I'll fix that.

utils/python/update_acme_tests.py
jedwards4b: The SEQ test makes more sense as an A case, but I wanted to make sure we were still testing an X case.  Also no need to run this test for 3 days, 9 steps works fine.

PR #199, opened 2016-05-27
Wait for rework

jedwards4b:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes: 

User interface changes?: 

Code review: Jim F

### DIFF
--- a/utils/python/CIME/SystemTests/system_tests_common.py
+++ b/utils/python/CIME/SystemTests/system_tests_common.py
@@ -115,8 +115,9 @@ class SystemTestsCommon(object):
         rc, out, err = run_cmd("%s -rundir %s -testcase %s -testcase_base %s -suffix1 %s -suffix2 %s"
                                %(cmd, self._case.get_value('RUNDIR'), self._case.get_value('CASE'),
                                  self._case.get_value('CASEBASEID'), suffix1, suffix2), ok_to_fail=True)
+        logger.debug("run %s results %d %s %s"%(cmd,rc,out,err))
         if rc == 0:
-            append_status(out, sfile="TestStatus")
+            append_status(out.replace("compare","compare functionality",1) + "\n", sfile="TestStatus")
         else:
             append_status("Component_compare_test.sh failed out: %s\n\nerr: %s\n"%(out,err)
                           ,sfile="TestStatus.log")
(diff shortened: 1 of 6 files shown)


utils/python/wait_for_tests.py
```diff
@@ -307,19 +307,21 @@ def parse_test_status(file_contents):
...
-                rv[phase] = reduce_stati({"%s_" % phase : status, phase : rv[phase]})
-            else:
-                rv[phase] = status
+        tokens = line.strip().split()
+        if (len(tokens)==0 or tokens[0] == COMMENT_STATUS):
+            pass # skip comments and blank lines
         else:
-            logging.warning("In TestStatus file for test '%s', line '%s' not in expected format" % (test_name, line))
+            if len(tokens) >= 3:
+                status, curr_test_name, phase = tokens[:3]
```
jgfouca: Due to the number of warnings I see, I'm still a little worried this change will cause this function to misinterpret lines that are not valid status lines as status lines. The prior condition that the number of tokens must exactly equal three was pretty weak, but now it's even weaker. Is there any way to look at content of these three tokens to see if they are sane? For example, status should be a member of the set of valid statuses, curr_test_name should match test_name, and phase should be valid.

jgfouca:
Review complete. One question needs answered.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -115,8 +115,9 @@ def _component_compare_test(self, suffix1, suffix2):
         rc, out, err = run_cmd("%s -rundir %s -testcase %s -testcase_base %s -suffix1 %s -suffix2 %s"
                                %(cmd, self._case.get_value('RUNDIR'), self._case.get_value('CASE'),
                                  self._case.get_value('CASEBASEID'), suffix1, suffix2), ok_to_fail=True)
+        logger.debug("run %s results %d %s %s"%(cmd,rc,out,err))
         if rc == 0:
-            append_status(out, sfile="TestStatus")
+            append_status(out.replace("compare","compare functionality",1) + "\n", sfile="TestStatus")
```
jgfouca: What is this replace for?

jgfouca: Is there any way we could get the suffixes added here?

jedwards4b: I don't understand the question

utils/python/wait_for_tests.py
```diff
@@ -307,19 +307,30 @@ def parse_test_status(file_contents):
...
+                if (test_name is None):
+                    test_name = curr_test_name
+                else:
+                    expect(test_name == curr_test_name, "inconsistant test name in parse_test_status %s %s"%(test_name, curr_test_name))
+                expect(status in [TEST_PENDING_STATUS ,TEST_PASS_STATUS, \
```
jgfouca: Why would a status ever start with "Test"?

utils/python/wait_for_tests.py
```diff
@@ -307,19 +307,30 @@ def parse_test_status(file_contents):
...
+                                      TEST_FAIL_STATUS, TEST_DIFF_STATUS, NAMELIST_FAIL_STATUS] or status.startswith("Test"),
+                       "Unexpected status in parse_test_status")
+                expect(phase in ["INIT","CREATE_NEWCASE","XML","SETUP","SHAREDLIB_BUILD",
```
jgfouca: Should we fail if these assumptions aren't met or just assume the line isn't intended to be a status line? (like if tokens are < 3).

jgfouca:
Replaced with new PR.

PR #202, opened 2016-05-31
Make sure user knows if exeroot already exists, abort if test

jedwards4b:
Check for existance of EXEROOT and RUNDIR, abort if this is a test and they exist,
otherwise give the user an option to use them or replace them or abort create_newcase. 

Test baseline: 
Test namelist changes: 
Test status:[bit for bit

Fixes:

User interface changes?: 

Code review:


jgfouca:
Looks good.

PR #203, opened 2016-05-31
update soft module loading to run on ANL workstations

sarich:
Fixes some assumptions about soft keys that weren't true for general workstations

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes: [CIME Github issue #]

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/env_module.py
+++ b/utils/python/CIME/env_module.py
@@ -7,6 +7,7 @@ import CIME.utils
 from CIME.utils import expect, run_cmd
 from CIME.XML.machines import Machines
 from CIME.XML.env_mach_specific import EnvMachSpecific
+import string
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


utils/python/CIME/env_module.py
```diff
@@ -119,37 +120,65 @@ def _load_module_modules(self, modules_to_load):
...
         output=run_cmd(cmd)
 
+        ###################################################
+        # Parse the output to set the os.environ dictionary
+        ###################################################
         newenv = {}
+        dolater = []
         for line in output.splitlines():
-            m=re.match(r'^(\S+)=(\S+)$',line)
+            if line.find('$')>0:
```
jedwards4b: Isn't this functionality already provided by case.get_resolved_value?

sarich: case.get_resolved_value() passes the string along to
GenericXML.get_resolved_value() which
can resolve strings using the current environment variables, but here I
wanted to resolve using
a separate dictionary that isn't fully defined until all the output has
been parsed.

PR #206, opened 2016-06-06
Sarich/convert x buildnml buildlib

sarich:
Convert X component buildlib and buildnml scripts to use python
individual scripts 
components/xcpl_comps/*/cime_config/buildnml
components/xcpl_comps/*/cime_config/buildlib

are almost identical, implementations are in 
utils/python/CIME/buildnml.py
utils/python/CIME/buildlib.py

data components should be able to follow this model as well, the implementation
classes will be a bit more involved though.

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes: [CIME Github issue #]

User interface changes?: 

Code review:

### DIFF
--- a/components/xcpl_comps/xwav/cime_config/buildnml
+++ b/components/xcpl_comps/xwav/cime_config/buildnml
@@ -1,41 +1,50 @@
-#! /bin/csh -f
+#!/usr/bin/env python2
+"""
+buildnml - create namelist
+"""
 
-set RUNDIR	= `./xmlquery  RUNDIR    -value`
-set NINST_WAV	= `./xmlquery  NINST_WAV -value`
-set WAV_NX	= `./xmlquery  WAV_NX	 -value`
-set WAV_NY	= `./xmlquery  WAV_NY	 -value`
+#Bootstrap python path
+import sys
+import os
+if "CIMEROOT" not in os.environ:
+    sys.stderr.write("ERROR: CIMEROOT must be defined\n")
+    sys.exit(1)
+sys.path.append(os.path.join(os.environ["CIMEROOT"],"scripts","Tools"))
 
-set NX = $WAV_NX
-set NY = $WAV_NY
+from standard_script_setup import *
+from CIME.case import Case
+import CIME.buildnml
 
-cd $RUNDIR
-set base_filename = "xwav_in"
+###############################################################################
+def parse_command_line(args, description):
+###############################################################################
+    parser = argparse.ArgumentParser(
+	    description=description,
+            formatter_class=argparse.ArgumentDefaultsHelpFormatter
+	)
 
-set inst_counter = 1
+    CIME.utils.setup_standard_logging_options(parser)
+    parser.add_argument("caseroot", nargs="?", default=os.getcwd(),
+                        help="Case directory for building namelist")
+    args = parser.parse_args(args[1:])
+    CIME.utils.handle_standard_logging_options(args)
+    return args.caseroot
 
-while ($inst_counter <= $NINST_WAV)
 
-    set inst_string = " "
-    if ($NINST_WAV > 1) then
-        set inst_string = $inst_counter
-        if ($inst_counter <= 999) set inst_string = 0$inst_string
-        if ($inst_counter <=  99) set inst_string = 0$inst_string
-        if ($inst_counter <=   9) set inst_string = 0$inst_string
-        set inst_string = _$inst_string
-    endif
+###############################################################################
+def _main_func(description):
+###############################################################################
+    if "--test" in sys.argv:
+        test_results = doctest.testmod(verbose=True)
+        sys.exit(1 if test_results.failed > 0 else 0)
 
-    set in_filename = ${base_filename}${inst_string}
+    caseroot = parse_command_line(sys.argv, description)
+    case = Case(caseroot)
+    builder = CIME.buildnml.buildxnml(case,'xwav','WAV')
+    builder.build_namelist()
 
-    cat >! ${in_filename} << EOF
-$NX                  !  i-direction global dimension
-$NY                  !  j-direction global dimension
-1                    !  decomp_type  1=1d-by-lat, 2=1d-by-lon, 3=2d, 4=2d evensquare, 11=segmented
-0                    !  num of pes for i (type 3 only)
-0                    !  length of segments (type 4 only)
-EOF
 
-    @ inst_counter = $inst_counter + 1
-
-end
+if __name__ == "__main__":
+    _main_func(__doc__)
 
 
--- a/components/xcpl_comps/xwav/cime_config/buildnml
+++ b/components/xcpl_comps/xwav/cime_config/buildnml
@@ -13,7 +13,7 @@ sys.path.append(os.path.join(os.environ["CIMEROOT"],"scripts","Tools"))
 
 from standard_script_setup import *
 from CIME.case import Case
-import CIME.buildnml
+import CIME.build_nml
 
 ###############################################################################
 def parse_command_line(args, description):
@@ -40,7 +40,7 @@ def _main_func(description):
 
     caseroot = parse_command_line(sys.argv, description)
     case = Case(caseroot)
-    builder = CIME.buildnml.buildxnml(case,'xwav','WAV')
+    builder = CIME.build_nml.BuildXNml(case,'xwav','WAV')
     builder.build_namelist()
 
 
--- a/components/xcpl_comps/xwav/cime_config/buildnml
+++ b/components/xcpl_comps/xwav/cime_config/buildnml
@@ -40,8 +40,7 @@ def _main_func(description):
 
     caseroot = parse_command_line(sys.argv, description)
     case = Case(caseroot)
-    builder = CIME.build_nml.BuildXNml(case,'xwav','WAV')
-    builder.build_namelist()
+    CIME.build_nml.build_x_nml(case,'xwav','ATM')
 
 
 if __name__ == "__main__":
--- a/components/xcpl_comps/xwav/cime_config/buildnml
+++ b/components/xcpl_comps/xwav/cime_config/buildnml
@@ -39,8 +39,7 @@ def _main_func(description):
         sys.exit(1 if test_results.failed > 0 else 0)
 
     caseroot = parse_command_line(sys.argv, description)
-    case = Case(caseroot)
-    CIME.build_nml.build_x_nml(case,'xwav','ATM')
+    CIME.build_nml.build_x_nml(caseroot,'xwav','WAV')
 
 
 if __name__ == "__main__":
--- a/utils/python/CIME/buildlib.py
+++ b/utils/python/CIME/buildlib.py
@@ -41,7 +41,12 @@ class buildxlib(buildlib):
         makefile = os.path.join(casetools, "Makefile")
         macfile = os.path.join(caseroot, "Macros.%s" % mach)
 
-        # build 
-        run_cmd("%s complib -j %d MODEL=%s COMPLIB=%s -f %s MACFILE=%s"
-                % (gmake, gmake_j, self.model, complib, makefile, macfile))
-
+        # build
+        cmd = "%s complib -j %d MODEL=%s COMPLIB=%s -f %s MACFILE=%s" % (gmake,
+                    gmake_j, self.model, complib, makefile, macfile))
+        rc, out, err = run_cmd(cmd)
+        expect(rc == 0, "Command %s failed rc=%d\nout=%s\nerr=%s" % (cmd, rc,
+                                                                     out, err))
+        logger.info("Command %s completed with output %s\nerr %s"% (cmd, out,
+                                                                    err))
+        
--- a/utils/python/CIME/buildlib.py
+++ b/utils/python/CIME/buildlib.py
@@ -43,7 +43,7 @@ class buildxlib(buildlib):
         macfile = os.path.join(self.caseroot, "Macros.%s" % self.mach)
 
         # build 
-        stat = run_cmd("%s complib -j %d MODEL=%s COMPLIB=%s -f %s MACFILE=%s"
-                       % (self.gmake, self.gmake_j, self.model, complib,
-                          makefile, macfile))
+        run_cmd("%s complib -j %d MODEL=%s COMPLIB=%s -f %s MACFILE=%s"
+                % (self.gmake, self.gmake_j, self.model, complib,
+                   makefile, macfile))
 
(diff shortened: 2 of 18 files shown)


utils/python/CIME/buildlib.py
```diff
@@ -0,0 +1,49 @@
+from CIME.XML.standard_module_setup import *
+from CIME.case import Case
+"""
+buildlibx - given the caseroot and component name(cmp), build the stub component
```
jedwards4b: comment doesnt match filename

utils/python/CIME/buildlib.py
```diff
@@ -0,0 +1,49 @@
...
+"""
+
+class buildlib(object):
+    def __init__(self, case, model, comp):
+        self.case = case
+        self.model = model
+        self.comp = comp
+        self.caseroot = case.get_value("CASEROOT")
```
jedwards4b: Why do all of these in init instead of just getting them in buildxlib?

utils/python/CIME/buildnml.py
```diff
@@ -0,0 +1,62 @@
+from CIME.XML.standard_module_setup import *
+from CIME.case import Case
+
+class buildnml(object):
+    def __init__(self, case, model, comp):
+        self.case = case
+        self.model = model
+        self.comp = comp
+        self.cimeroot = case.get_value("CIMEROOT")
```
jedwards4b: No need for get_value calls in init.

jedwards4b:
I think that the init method should just assign self.case = case and all of the get_value calls should be made where needed in the build methods.    Other than that this all looks good.

utils/python/CIME/buildlib.py
jgfouca: Pylint will flag that (it's good practice to initialize all your members in the constructor). I think this is probably the correct design if there are going to be multiple subclasses of buildlib.

utils/python/CIME/buildlib.py
```diff
@@ -0,0 +1,49 @@
+from CIME.XML.standard_module_setup import *
+from CIME.case import Case
+"""
+buildlibx - given the caseroot and component name(cmp), build the stub component
+"""
+
+class buildlib(object):
```
jgfouca: In standard python convention, class names should be capitalized in order to look different than functions. Please change this to BuildLib and make similar changes to the other three classes you added.

utils/python/CIME/buildlib.py
```diff
@@ -0,0 +1,49 @@
...
+        self.cimeroot = case.get_value("CIMEROOT")
+        self.mach = case.get_value("MACH")
+        self.objroot = case.get_value("OBJROOT")
+        self.libroot = case.get_value("LIBROOT")
+        self.gmake_j = case.get_value("GMAKE_J")
+        self.gmake = case.get_value("GMAKE")
+
+class buildxlib(buildlib):
+    def buildlib(self):
+        try:
```
jgfouca: Standard safe way to open file:

    with open('Filepath', 'w') as out:
        * do stuff to out

utils/python/CIME/buildlib.py
```diff
@@ -0,0 +1,49 @@
...
+            out = open('Filepath','w')
```
jgfouca: Is the file's name really "Filepath"? That's kinda odd

utils/python/CIME/buildlib.py
```diff
@@ -0,0 +1,49 @@
...
+                               self.comp, "cpl") + "\n")
+        out.close()
+
+        # generate macro values
+        complib = os.path.join(self.libroot, "lib%s.a" % self.model)
+        makefile = os.path.join(self.casetools, "Makefile")
+        macfile = os.path.join(self.caseroot, "Macros.%s" % self.mach)
+
+        # build 
+        stat = run_cmd("%s complib -j %d MODEL=%s COMPLIB=%s -f %s MACFILE=%s"
```
jgfouca: This is an error. Unless "ok_to_fail=True" is passed to run_cmd, the return value is just a string containing the output.

utils/python/CIME/buildnml.py
```diff
@@ -0,0 +1,62 @@
...
+            flood_mode = Case('XROF_FLOOD_MODE')
+            if flood_mode == "ACTIVE":
+                extras = [[".true.","flood flag"]]
+            else:
+                extras = [[".false.","flood flag"]]
+
+        for i in range(1, self.ninst + 1):
+            # If only 1 file, name is 'model_in'
+            # otherwise files are 'model_in0001', 'model_in0002', etc
+            if self.ninst == 1:
```
jgfouca: filename = base_filename if self.ninst == 1 else "%s%4.4d" % (base_filename, i)
    with open(filename, 'w') as infile:
         * do stuff *

components/xcpl_comps/xwav/cime_config/buildnml
```diff
@@ -1,41 +1,50 @@
...
+if "CIMEROOT" not in os.environ:
+    sys.stderr.write("ERROR: CIMEROOT must be defined\n")
+    sys.exit(1)
+sys.path.append(os.path.join(os.environ["CIMEROOT"],"scripts","Tools"))
+
+from standard_script_setup import *
+from CIME.case import Case
+import CIME.buildnml
+
+###############################################################################
```
jgfouca: All of these files are so similar, it seems a shame that we have so many copies of nearly identical code. Is there any way to leverage a bit of extra encapsulation so that these files are just a couple lines?

jgfouca:
@sarich looks pretty good for the most part. Please either push back on or complete the changes requested. Also, I'd really love to see encapsulation rather than copy/paste for all those buildlib executables, but I won't make that a requirement for merging this PR.

utils/python/CIME/buildlib.py
jedwards4b: pylint will only flag it if you use for example self.caseroot, instead what I propose is that that variable should never exist and you just locally do
 
caseroot = self.case.get_value("CASEROOT")

mvertens: I think Jim's suggestion makes sense - because different component sets (like the CESM prognostic components have more variables that are needed than the dead components. That brings up an interesting issue - where does having buildlib.py make sense. The buildlib for cam is not quite the same as the buildlib for clm and is different than the buildlib for pop. I'm really not sure it makes sense to introduce new methods for each of these in the buildlib.py - and am wondering if we shouldn't just have each prognostic component have its own separate buildlib?  In that case, buildlib.py would only be used for CIME components - is that the consensus?

utils/python/CIME/buildlib.py
mvertens: Yes it is.

utils/python/CIME/buildnml.py
mvertens: Jim - this is a great suggestion that should simplify the prognostic component buildnml files

utils/python/CIME/buildlib.py
mvertens: Comment "generate macros values" is incorrect - these are just file paths

mvertens: I am wondering if we should be using OBJROOT and LIBROOT as xml variables. Frankly, OBJROOT only makes sense if you are not using shared libraries (which is the case for all CESM components now other than clm4.5).
It seems that we want to pass buildroot to build.py and simply use that without querying case.get_value for OBJROOT or

mvertens: Actually, looking at this some more I think we need to change build.py to have build_model to remove the following statement:

bldroot = exeroot #What is this for?

and calculate it correctly in here (rather than inside the component's buildlib.

sarich: Thanks for the feedback everyone, I was travelling yesterday and trying to
catch up now.

I'll make the suggested changes, they all make sense

   -  documentation fixes
   - class names change buildlib, buildxlib, buildnml, buildxnml to
   BuildLib, BuildXLib, buildNML, buildXNML
   - open files using 'with open as'
   - don't do any work in __init__
   - fix run_cmd to ignore return value (it is not okay to fail)

I'm keeping these on the table for further discussion without changing
anything:

   - Looking at this some more I think we need to change build.py to have
   build_model to remove the following statement:   bldroot = exeroot #What is
   this for?
   - I also hate having identical or almost identical code in multiple file
   locations. Complicates development unnecessarily. Maybe we can fix this
   with symlinks or replace the python with a one-line bash script
   - I am wondering if we should be using OBJROOT and LIBROOT as xml
   variables. Frankly, OBJROOT only makes sense if you are not using shared
   libraries (which is the case for all CESM components now other than
   clm4.5). It seems that we want to pass buildroot to build.py and simply use
   that without querying case.get_value for OBJROOT orand calculate it
   correctly in here (rather than inside the component's buildlib.
   -
   - Is the file's name really "Filepath"? (Yes)
   - In that case, buildlib.py would only be used for CIME components - is
   that the consensus?

Jason

utils/python/CIME/buildlib.py
jgfouca: @jedwards4b : Ah, misunderstood you. Yes, that would be fine.

jedwards4b:
Your run_cmd lines should look something like this:
        rc, out, err = run_cmd(cmd, from_dir=os.path.join(caseroot,"Buildconf","clmconf"), ok_to_fail=True)
        expect(rc==0,"Command %s failed rc=%d\nout=%s\nerr=%s"%(cmd,rc,out,err))
        logger.info("Command %s completed with output %s\nerr %s"%(cmd,out,err))
Otherwise output from the cmd tends to get lost.

jgfouca:
@jedwards4b : What you wrote is functionally identical to using run_cmd without the ok_to_fail flag. I think the way he's using it is OK since he wants to program to abort if the command fails:

    output = run_cmd(something)

sarich:
ok, thanks, I'll use that.

sarich:
Yes, I was letting the expect() in run_cmd take care of the failure, but
that wasn't logging the failure message

jedwards4b:
okay - I think that at least some of the logger output is still going to stderr instead of stdout and is
lost when you do this.   I'll have another go at getting the logger to output to stdout.

sarich:
Yes, I see if I want to log a failure I have to do it before the expect().
How about
        rc, out, err = run_cmd("%s complib -j %d MODEL=%s COMPLIB=%s -f %s"
                               " MACFILE=%s"
                               % (gmake, gmake_j, self.model, complib,
                                  makefile, macfile))
        if (rc == 0):
            logger.info("Command %s completed with output %s\nerr
%s"%(cmd,out,err))
        else:
            logger.info("Command %s failed rc=%d\nout=%s\nerr=%s" % (cmd,
rc,
                                                                     out,
err))

        expect(rc == 0, "Command %s failed rc=%d\nout=%s\nerr=%s" % (cmd,
rc,
                                                                     out,
err))

This is starting to get ugly though

jedwards4b:
You don't need both the if/else and expect - just put the expect first as I have it above.

jgfouca:
@jedwards4b : wouldn't this work:

    output = run_cmd("something 2>&1")
    logger.info("Command 'something' completed with output: %s" % output)

But your point is well taken that stderr is thrown out by default if ok_to_fail is False.

jedwards4b:
@sarich In looking at this again it's not clear to me that there is any value in creating a buildnml and buildxnml class - a simple function will do the job.   Mariana has rewritten the data models with this approach and would like to discuss,  but she is in meetings all day today - can we set up a time tomorrow?   @jgfouca can you join us?

jgfouca:
@jedwards4b , we're both at the ACME all hands meeting, so it may be difficult. I was also wondering if using OO concepts was buying us much or just complicating the API. I generally prefer the simplest API, free functions, over over classes unless shared state must be maintained.

utils/python/CIME/buildlib.py
mvertens: Hi Jason,

I am going to post this on slack as well. I don't think we need a classes
structure for buildlib - but rather just a module with reused routines.
I also worked on build.py to address issues that I mentioned in the PR. Not
sure how to coordinate all of this. I am in a NSF site review meeting all
day today.

Mariana

sarich:
stand-alone functions are fine, where do they live? The same files?

Do we need the buildlib and buildnml scripts to exists or can this
functionality be called directly from the build.py?
instead of (in build.py)

>         my_file = os.path.join(os.path.dirname(machines_file),
> "buildlib.%s" % lib)
>         stat = run_cmd("%s %s %s >> %s 2>&1" %
>                        (my_file, sharedpath, caseroot, file_build),
>                        from_dir=exeroot,
>                        ok_to_fail=True, verbose=True)[0]
>         expect(stat == 0, "ERROR: buildlib.%s failed, cat %s" % (lib,
> file_build))

jedwards4b:
Yes I think that they can live in the same files.   The issue with calling them functionally:   Components outside of cime will interact with the cime build by providing their own buildlib and buildnml scripts and these may or may not be python scripts which are functionally compatible with cime - if you provide functional calls for the cime internal ones we may still need to provide for a shell cmd call for external ones.

sarich:
I think the consensus then is


for the executables, replace:
builder = CIME.buildnml.buildxnml(case,'xatm','ATM')
builder.build_namelist()

with:
CIME.buildnml.buildxnml(case,'xatm','ATM')


and then in CIME/buidnml.py:
use a stand-alone function

def buildxnml(case, model, comp):
   ...
-- or maybe better
def build_x_nml(case, model, comp)


and later add buil_stub_nml(), build_data_nml()


And likewise with buildlib?

sarich:
I've switched buildnml, buildlib from classes to standalone functions

jgfouca:
@sarich is this all ready?

sarich:
Yes

mvertens:
Hi,

I have implemented a stand-alone buildlib as well for the data models using
a lot of what Jason put in the classes. I hope that we have not stepped on
each other's toes. I was in the middle of testing this right now before
issuing a pull request. Part of what I did - was that I refactored parts of
buildlib to pass more consistent arguments to the components - this was
mentioned in my comments to the original PR from Jason.

At any rate - I'll issue a PR right now (even though the tests are still
running) - so that everyone can comment.

Mariana

mvertens:
Just as an additional note, the changes I made to build.py clarifies
confusion in creating the clm library (which is created as a shared library
for clm4.5 and clm5.0 but not for clm4.0).

jgfouca:
I'm going to hold off until master is working on melvin.

jgfouca:
Looks good.

mvertens:
I'd like to know how to proceed at this point. Can I fold in Jason's
changes for buildlib into my new buildlib? Is that a problem? I am happy to
also take on the stub components.

rljacob:
Just finish this PR to do data models.  After its in master, start a new branch to merge the two buildlib's.

PR #209, opened 2016-06-09
Buildnml

### LINKED ISSUE #211: convert component/.../buildlib and buildnml from perl/csh to python for data, stub models

mvertens:
Added new routine buildlib.py 

This routine reuse parsing the parsing of arguments, the running of gmake and also to create all data model builds. Also changed buildlib.py to pass new arguments to _build_model_thread to address confusing issues in levering the xml variables OBJROOT, EXEROOT  and LIBROOT that were noted as questions.

Test baseline: NA
Test namelist changes: none
Test status: bit-for-bit

Closes #211 

User interface changes?: 
None
Code review: 
Jim Edwards


rljacob:
PR #206 has 
utils/python/CIME/build_lib.py
while this PR has
utils/python/CIME/buildlib.py

I assume we only need one of those?

sarich:
Yes, they should be the same file. probably buildlib is the better name
I suggest pulling @mvertens branch into master and I'll merge in mine, but will probably need some help with how to do this without rebasing

mvertens:
Thanks. This sounds like a good plan. I'm wrapping up the testing - and
thought of a bit of a cleaner interface to run_gmake - i'll issue an
updated PR later this afternoon.

rljacob:
You don't need to make a new PR.  Just push more commits to this branch and comment that this is now a real PR.

rljacob:
"Buildlib" might be the better name but which is the better implementation?  Both branches should use that.

mvertens:
Yep - that's what I meant to do. Sorry for the confusing statement.

jgfouca:
@sarich : nothing wrong with rebasing.

sarich:
Rob,

these implement two different functions that will live in the file

rljacob:
Ah.  Well #206 has been merged to master so that combination can be made in this PR or a new PR.

mvertens:
Hi,

I'm happy to take on the merge of getting the dead components and stub
components into the new interface.

Mariana

components/data_comps/datm/cime_config/buildlib
```diff
@@ -1,34 +1,40 @@
...
+        test_results = doctest.testmod(verbose=True)
+        sys.exit(1 if test_results.failed > 0 else 0)
+
+    parser = argparse.ArgumentParser()
+
+    CIME.utils.setup_standard_logging_options(parser)
+
+    caseroot, bldroot, libroot = parse_command_line(sys.argv, parser)
+
+    build_data_lib('atm', 'datm', caseroot, bldroot, libroot)
```
jgfouca: Is the second argument redundant? The function is "build_**data**_lib" , so should it already know to do 'datm' if 'atm' is the first argument?

jgfouca:
@mvertens : I wonder if we can be even more clever and further reduce copy/pasting. I'm thinking that each buildlib and buildnml becomes a symlink to a generic version which uses the path of the symlink to determine what it's supposed to do. For example the components/data_comps/datm/cime_config/buildlib symlink would know to do a data build for 'atm' by seeing that it lives under data_comps/datm.

components/data_comps/datm/cime_config/buildlib
mvertens: Yes - you are right - this is redundant. I'll fix this and push my changes.

mvertens:
That's a good idea. Could we get this PR in first and them implement a more
generic functionality?
Thoughts?

quantheory:
@jgfouca I'm always nervous about checking `argv[0]` because that can be thrown off if you're calling the script from the "wrong" place, trying to call it through a symlink (to the symlink), or that sort of thing.

quantheory:
Maybe an alternative way of shortening the script; move `_main_func` to `CIME.buildlib` (changing the name to something else), and give it an argument that's just the component name (e.g. "atm"). Then the script does nothing except import that one function and call it with the component name.

components/data_comps/datm/cime_config/buildlib
mvertens: An alternative way of shortening the script; move `_main_func` to `CIME.buildlib` and give it an argument that's just the component name (e.g. "atm"). Then the script does nothing except import that one function and call it with the component name. I'm implementing this right now and will be updating this PR shortly.

jgfouca:
@mvertens , sure, just let me know when you've addressed all the other comments and I'll merge. @quantheory : I was thinking of using the place where the symlink lives (__file__), not argv[0].

quantheory:
@jgfouca I guess I'm confused. How would the script know whether/where it was being called through a symlink? The only way I know of to do that is to check `argv[0]` (or equivalent in whatever language you're in).

jgfouca:
The python internal variable

    __file__

Contains the current location on disk of the source file. It will display the symlink and not the true file if the current file is a symlink, thus giving us the info we need for buildnml and build_library.

quantheory:
Ah, right, but I think that `__file__` is always equal to `sys.argv[0]` inside a script. When a module is loaded, Python can figure out where it loaded that module from, but for a script, I think it cops out and takes whatever the OS gives it:

```python
#!/usr/bin/env python2

print __file__
```

then

```sh
> ./blah.py
./blah.py
> ln -s blah.py blah2.py
> ./blah2.py
./blah2.py
```

The fact that you can get a relative path back is what I see as the part that might make the system more fragile, in any case (well, that and the fact that adding a second layer of symlink would also break the script).

mvertens:
Actually, I think I've managed to have a very clean implementation at this
point without any code reuse that does not use symbolic links. I'm doing
the final testing now and will be issuing a PR soon.

jgfouca:
@quantheory : you can always convert a relative path to absolute via os.path.abspath

quantheory:
@jgfouca True. I guess this is basically the same as what a lot of our scripts have done in the past to try to find some root directory or other (e.g. `$CIMEROOT`). It's still theoretically vulnerable to problems if someone is adding their own layer of symbolic links, but I suppose that's pretty rare.

jgfouca:
@quantheory : That's true. I never thought of that. Python has a realpath function that resolves symlinks but that would obviously break the design I proposed for these buildnml files. @mvertens says she's done one better so maybe it's moot.

jgfouca:
@mvertens : I see a new push, is this PR ready for final review?

mvertens:
Yes.

components/data_comps/datm/bld/build-namelist.pl
```diff
@@ -0,0 +1,415 @@
+#!/usr/bin/env perl
+#-----------------------------------------------------------------------------------------------
+#
+# build-namelist
+#
+# This is the build-namelist script for the CIME datm (Data Atmosphere Model).
+#--------------------------------------------------------------------------------------------
+
```
jgfouca: What's the reason for all this new perl code?

components/data_comps/dwav/cime_config/buildnml
```diff
@@ -1,96 +1,43 @@
...
-}
-chdir "$CASEROOT/Buildconf/dwavconf";
+###############################################################################
+def _main_func():
+###############################################################################
 
-my $inst_string;
-my $inst_counter = 1;
-while ($inst_counter <= $NINST_WAV) {
+    if "--test" in sys.argv:
```
jgfouca: It looks like this file was not converted to the new extra small format?

jgfouca:
@mvertens : couple questions need to be addressed and then I'll merge.

components/data_comps/datm/bld/build-namelist.pl
mvertens: My mistake. It should not be there. I'll delete it. I'm trying to get all of the perl in the data model build-namelists put into one reusable routine that can then be pythonized.

components/data_comps/dwav/cime_config/buildnml
mvertens: sorry about that. Will fix that as well.

PR #210, opened 2016-06-09
Port changes made on ACME side to CIME for anlworkstations

jgfouca:
This exposed some important errors. First, pio.py was not providing
the mpilib in the attribute list when ask for PIO data. This meant that
the selection of mpiserial was not causing us to use netcdf instead of
pnetcdf as PIO_TYPENAME.

The second error was in entry_id. The algorithm for computing best match
was not picking up defaults in the case where the XML <value> tag has
no attributes.

Another error is that attributes were not being checked when loading
modules or setting env variables.

Test suite: cime_developer
Test baseline: 
Test namelist changes: 

User interface changes?: No

Code review: @jedwards4b

### DIFF
--- a/cime_config/acme/machines/config_machines.xml
+++ b/cime_config/acme/machines/config_machines.xml
@@ -409,13 +409,15 @@
       <cmd_path lang="sh">source /software/common/adm/packages/softenv-1.4.2/etc/softenv-aliases.sh ; soft</cmd_path>
       <modules compiler="gnu">
 	<command name="add">+gcc-5.1.0</command>
-	<command name="add">+mpich-3.2-gcc-5.1</command>
+	<command name="add" mpilib="!mpi-serial">+mpich-3.2-gcc-5.1</command>
 	<command name="add">+netcdf-4.3.3.1-parallel-gcc5.1-mpich3.2</command>
+	<command name="add" mpilib="!mpi-serial">+pnetcdf-1.6.1-gcc-5.1-mpich-3.2</command>
 	<command name="add">+cmake-2.8.12</command>
       </modules>
     </module_system>
     <environment_variables>
       <env name="NETCDFROOT">$(dirname $(dirname $(which ncdump)))</env>
+      <env name="PNETCDFROOT" mpilib="!mpi-serial">$(dirname $(dirname $(which pnetcdf_version)))</env>
     </environment_variables>
 </machine>
 
--- a/utils/python/CIME/XML/entry_id.py
+++ b/utils/python/CIME/XML/entry_id.py
@@ -68,28 +68,37 @@ class EntryID(GenericXML):
         Note that the component class has a specific version of this function
         '''
         match_value = None
-        match_max = 0
+        match_max = -1
         match_count = 0
         expect(node is not None," Empty node in _get_value_match")
+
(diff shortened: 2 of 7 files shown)


cime_config/acme/machines/config_machines.xml
```diff
@@ -409,13 +409,15 @@
       <cmd_path lang="sh">source /software/common/adm/packages/softenv-1.4.2/etc/softenv-aliases.sh ; soft</cmd_path>
       <modules compiler="gnu">
 	<command name="add">+gcc-5.1.0</command>
-	<command name="add">+mpich-3.2-gcc-5.1</command>
+	<command name="add" mpilib="!mpi-serial">+mpich-3.2-gcc-5.1</command>
```
jgfouca: These attribute checks were being lost until this PR.

utils/python/CIME/XML/entry_id.py
```diff
@@ -68,28 +68,37 @@ def _get_value_match(self, node, attributes={}):
...
         values = self.get_optional_node("values", root=node)
         if values is None:
             return
-        for valnode in self.get_nodes("value", root=node):
+
+        for valnode in self.get_nodes("value", root=values):
             # loop through all the keys in valnode (value nodes) attributes
-            for key,value in valnode.attrib.iteritems():
-                # determine if key is in attributes dictionary
+            if len(valnode.attrib) == 0:
```
jgfouca: Bug fix. Handles the case where value is not protected by any attributes.

utils/python/CIME/XML/env_mach_specific.py
```diff
@@ -116,7 +116,8 @@ def _compute_actions(self, nodes, child_tag):
             if (self._match_attribs(node.attrib)):
                 for child in node:
                     expect(child.tag == child_tag, "Expected %s element" % child_tag)
-                    result.append( (child.get("name"), child.text) )
+                    if (self._match_attribs(child.attrib)):
```
jgfouca: Bug fix: need to check child attributes. I left in the code that checks the parent's attributes, perhaps that can be removed.

utils/python/CIME/XML/pio.py
```diff
@@ -17,13 +17,21 @@ def __init__(self, infile=None, files=None):
...
         EntryID.__init__(self, infile)
 
-    def get_defaults(self, grid=None, compset=None, mach=None, compiler=None):
+    def get_defaults(self, grid=None, compset=None, mach=None, compiler=None, mpilib=None):
         # should we have a env_pio file
         defaults = {}
+
+        # Load args into attribute dict
+        attributes = {}
+        for attrib in ["grid", "compset", "mach", "compiler", "mpilib"]:
```
jgfouca: Bug fix: need to look at mpilib in order to pick up netcdf instead of pnetcdf if mpilib is mpi-serial.

jgfouca:
Annotations complete.

jedwards4b:
Looks good.

rljacob:
"made on ACME side" sounds like this was from testing in ACME-Climate/ACME/cime

jgfouca:
@rljacob the XML changes did orignate from ACME-climate.

PR #215, opened 2016-06-13
Remove `ADD_CPPFLAGS` from config_compilers.

quantheory:
The correct way of adding flags to `config_compilers` is via
`ADD_CPPDEFS`. There were some settings in this file that used
`ADD_CPPFLAGS` instead, but these appear to only take effect for:

 (a) obsolete AIX settings, and
 (b) the PIO build of GPTL (which never occurs when PIO is built by the
     CIME Makefile anyway).

Therefore I have removed these settings, since they are not useful and
interfere with other planned changes.

Testing:
    Not yet properly tested.
    This should of course be bit-for-bit.

### DIFF
--- a/cime_config/acme/machines/config_compilers.xml
+++ b/cime_config/acme/machines/config_compilers.xml
@@ -606,7 +606,6 @@ for mct, etc.
   <NETCDF_PATH>$(NETCDFROOT)</NETCDF_PATH>
   <PNETCDF_PATH>$(PNETCDFROOT)</PNETCDF_PATH>
   <ADD_SLIBS> $(shell $(NETCDF_PATH)/bin/nf-config --flibs) -lblas -llapack</ADD_SLIBS>
-  <ADD_CPPFLAGS> -DHAVE_COMM_F2C </ADD_CPPFLAGS>
   <CXX_LIBS>-lstdc++ -lmpi_cxx</CXX_LIBS>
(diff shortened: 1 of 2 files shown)


quantheory:
I'm 95% sure that this is fine, but it would help if someone could build on one of the affected machines (and/or we can wait until Jim has the chance to comment).

jgfouca:
@jedwards4b or me?

jgfouca:
It looks like anlworkstations and skybridge are both potentially affected. I can do a build on those if you want. If they work, I'm OK to merge this.

quantheory:
Ah, sorry, good point. I meant @jedwards4b because he knows the PIO build, but if you test on those machines, that should be good enough.

jayeshkrishna:
This change might also affect Cetus/Mira (ibm compilers)?

quantheory:
Neither of those uses AIX, since even IBM seems to have stopped using it for the last Blue Gene wave. I do not actually know of any extant machines that use AIX, and are actually being used to run CESM or ACME. Unless I'm missing something huge, we could probably just strip out the AIX settings? At least once we're sure it's dead. (In any case, I have no idea what `-DF2003` is supposed to do anyway, since AFAIK, we don't use that macro at all right now.)

jgfouca:
It looks like it works, merging.

PR #220, opened 2016-06-15
Things were badly broken on skybridge.

jgfouca:
Fix list:
1) {{ queue }} and {{ wall_time }} are apparently no longer supported, use job_queue and job_wallclock_time instead.
2) get_project needs to optionally look at config_machines.xml as its documentation advertises
3) Remove methods from batch.py that have been replaced by methods in env_batch.py (cleanup, not fix)
4) env_batch::get_value needs to use attributes when looking for optional node
5) Need to take job and case into account when getting batch directives
6) Fix env_batch::get_default_walltime
7) Fallback to default walltime if queue does not provide max walltime
8) Namespace errors in buildnml and buildlib
9) clone_case should use the PROJECT from the previous case
10) Make names line up better between env_batch and task_maker. This will allow
    easier substitution between the two when resolving {{ tokens }}.
11) Mpi args were not being handled correctly. Do not determine args until an
    mpirun has been selected.
12) empty-dict attributes was breaking get_nodes

Test suite: cime_developer
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

Fixes: [CIME Github issue #]

User interface changes?: no

Code review: @mvertens

### DIFF
--- a/utils/python/CIME/case.py
+++ b/utils/python/CIME/case.py
@@ -84,10 +84,6 @@ class Case(object):
         self._env_generic_files.append(EnvArchive(case_root))
         self._files = self._env_entryid_files + self._env_generic_files
 
-
-        # self._case_root = case_root
-
-
         # Hold arbitary values. In create_newcase we may set values
         # for xml files that haven't been created yet. We need a place
         # to store them until we are ready to create the file. At file
@@ -104,7 +100,6 @@ class Case(object):
         self._gridfile = None
         self._components = []
         self._component_config_files = []
-
 
     def __del__(self):
         self.flush()
@@ -549,11 +544,11 @@ class Case(object):
 
         # Set project id
         if project is None:
-            project = get_project()
+            project = get_project(machobj)
         if project is not None:
             self.set_value("PROJECT", project)
         elif machobj.get_value("PROJECT_REQUIRED"):
-            expect(project is not None, " PROJECT_REQUIRED is true but no project found")
+            expect(project is not None, "PROJECT_REQUIRED is true but no project found")
 
     def get_compset_var_settings(self):
         compset_obj = Compsets(infile=self.get_value("COMPSETS_SPEC_FILE"))
@@ -758,7 +753,7 @@ class Case(object):
         # user's case. However, note that, if a project is not given, the fallback will
         # be to copy it from the clone, just like other xml variables are copied.
         if project is None:
-            project = get_project()
+            project = self.get_value("PROJECT")
         if project is not None:
             newcase.set_value("PROJECT", project)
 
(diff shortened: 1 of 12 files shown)


rljacob:
Are these fixes only to help things on skybridge?   I see lots of general purpose infrastructure files being modified.

jgfouca:
Things were very broken, not just on skybridge.

jgfouca:
@mvertens hold off on this for a bit, seeing even more problems in broader testing

mvertens:
@foucar I had just started looking at it. I don't know a lot about the
batch implementation. I can review the rest fairly comfortably.

utils/python/CIME/case.py
```diff
@@ -758,7 +753,7 @@ def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None):
         # user's case. However, note that, if a project is not given, the fallback will
         # be to copy it from the clone, just like other xml variables are copied.
         if project is None:
-            project = get_project()
+            project = self.get_value("PROJECT")
```
quantheory: Is this change just to make sure that the code is doing what the comment actually says it's doing?

jgfouca: Yes.

quantheory:
This looks good to me for the most part. Regarding `ccsm_estcost`, the answer is yes, it should be taken into account (especially for regression tests where you don't want to have to monkey with each case by hand).

I think what should be happening is that, instead of using the max wall clock time, there should be a function that searches the list of walltimes and selects the one that has `ccsm_estcost` as small as possible while still at least as big as the current value of `CCSM_ESTCOST` in the case. Only if there's no such walltime listed should it fall back on the `walltimemax` setting. And if that's missing should we use the default, as you implemented.

rljacob:
I don't understand your second paragraph about ccsm_estcost.  Is it different from CCSM_ESTCOST?

We would like to set an individual walltime for each test because they could get through the queue faster if the walltime requested was what was needed and smaller then the max wall clock time.

quantheory:
Well, `CCSM_ESTCOST` is in the case (`env_mach_pes`), and specifies the estimated cost of that case, but there's also a `ccsm_estcost` attribute on the `walltime` elements in `config_machines`, and my understanding is that that's supposed to represent approximately how long a run with that estimated cost will take. It's a pretty crude way of matching things up, since AFAIK nothing in the cost calculations accounts for how many days you are running? Or the number of cores used? But it should at least be better than just using the max wall time every single time.

mvertens:
I can see still needed ccsm_estcost - but I think it should be renamed estcost - and get rid of the ccsm_ prefix. But this has nothing to do with this particular PR. However, do we still need CCSM_ESTCOST?

mvertens:
This looks fine to me. Could you please first fill in the commit template - and I'll then go ahead and merge the PR.

quantheory:
LGTM

@mvertens `CCSM_ESTCOST` is the actual estimated cost, isn't it? `ccsm_estcost` is just a way of translating it that to a walltime?

jgfouca:
OK, found several more problems and fixed them. Skybridge appears to be functional again.

jgfouca:
I'm going to go ahead and merge. I don't think any of my recent fixes are controversial.

jedwards4b:
I have been trying to phase out these estimated cost variables in favor of specifying the wallclock time for tests in testlist.xml.   Not sure where things stand at this point.

PR #224, opened 2016-06-17
Raise an error if batch_submit isn't found.

quantheory:
There seems to be a problem on yellowstone relating to lookup of batch_submit.

While I'm still working on tracking down that error, in the meantime, we should
improve the error message that results from this problem.


jgfouca:
I have a fix for this coming.

jgfouca:
I made an extra commit on my fixes branch that fixes this issue but I forgot to merge it.

quantheory:
OK, great!

jgfouca:
OK, I pushed the fixes directly to master.

PR #225, opened 2016-06-17
Go back to using python instead of python2 as the executable

jgfouca:
python2 was causing much grief to users of systems that did not
use this convention.

Test suite: None
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes: [CIME Github issue #]

User interface changes?: 

Code review: @mvertens


quantheory:
@mvertens delegated to me, and this looks good to me.

PR #226, opened 2016-06-17
Add `config_build` Macros generation.

### LINKED ISSUE #222: New schema for config_compilers

quantheory:
What this series of commits does is to add a new file format
(`config_build.xml`, replacing `config_compilers.xml`), along
with a schema (`config_build.xsd`), Python module
(`CIME.macros`, replacing `CIME.XML.compilers`), and
regression test suite (`test_config_build.py`).

Note that the old system still exists and is used by default.
This only adds an "experimental" option to `create_newcase`
and `create_test` to use the new file format. As a result
this should be a bit-for-bit command as long as the
`--config-build` option is *not* passed to `create_test`/
`create_newcase`.

The renaming is for two reasons:

1. It makes it possible to have both systems exist in the
   same codebase during the transition.
2. It's slightly more accurate, in that some of the information
   in the file is not compiler-specific.

The goals for this refactoring are as follows:

1. Create a new format that matches a well-defined XML
   schema.
2. Remove order-dependence from Macros file
   generation. The new code should produce the same
   output regardless of the order of `<compiler>` elements.
   This also makes it possible to raise an error when there
   are ambiguous or duplicated settings of an item,
   whereas the `CIME.XML.compilers` module relies on
   order within the file to resolve ambiguities.
3. Put CMake and Makefile output on the same footing.
   The `macros` module can output all of the same
   information to CMake as it does to a `Makefile`.
4. Improve testing and documentation (though the new
   implementation is also much more complex...).
5. Allow the compiler and MPI library to be changed
   without regenerating the Macros file (indeed, the Macros
   file is generated only once, at `create_newcase` time).
6. Fix minor warts, e.g. the `GPTL_CPPDEFS` variable,
   which should actually just be a GPTL-specific setting of
   `CPPDEFS`.
7. Do prep work for converting the `Depends.*` files from
   Makefile format to XML (with the ultimate goal of allowing
   them to be output as CMake as well).

The `config_build.xml` files added here are generated from
the existing `config_compilers.xml` using the Perl script
`compilers_translation_tool.pl`. I have not bothered to
convert this script to Python because this is more of a
transition tool rather than something intended to be a
permanent part of the scripts.

Note that the `config_build.xml` files are therefore not
complete until they have gone through a round of
editing to fix obsolete comments and fix any errors due
to assumptions made by the translation tool.

Closes #222


quantheory:
I should probably also write up a guide for the `config_build` format, but if you look at it side-by-side with `config_compilers`, it's not hugely different. The main change is that flags are now added within `<base>` and `<append>` elements, instead of using the `FFLAGS` vs. `ADD_FFLAGS` element naming scheme. The `<var>`, `<env>`, and `<shell>` elements are also new, and are used to get rid of the Makefile-specific syntax in `config_compilers`.

utils/perl5lib/compilers_translation_tool.pl
```diff
@@ -0,0 +1,575 @@
+#!/usr/bin/env perl
+
```
jgfouca: It would be nice to have a description here of what this tool does and the arguments it accepts.

utils/python/CIME/XML/machines.py
```diff
@@ -215,9 +214,7 @@ def is_valid_compiler(self,compiler):
         ...
         SystemExit: ERROR: COMPILERS value nag not supported for machine edison
         """
-        if self.get_field_from_list("COMPILERS", reqval=compiler) is not None:
-            return True
-        return False
+        return self.get_field_from_list("COMPILERS", reqval=compiler) is not None
```
jgfouca: Good changes.

scripts/create_newcase
```diff
@@ -95,6 +95,10 @@ def parse_command_line(args, cimeroot):
                         help="Used to indicate that create_newcase was called from create_test"
                         "- do not use otherwise")
 
+    parser.add_argument("--config-build", action="store_true",
```
jgfouca: I'm wondering if it wouldn't be better to use an environment variable to control this behavior.

utils/python/tests/test_config_build.py
```diff
@@ -0,0 +1,581 @@
+#!/usr/bin/env python
+"""Tests for the generation of Macros files from config_build.xml"""
+
+# The invalid-name check is not very useful, while line-too-long and
+# bad-continuation pick up on a lot of lines that are hard to split up
+# legibly because they contain long strings.
+# pylint: disable=invalid-name,line-too-long,bad-continuation
```
jgfouca: Please add these to the rest of the regression tests (scripts_regression_tests).

utils/python/CIME/macros.py
```diff
@@ -0,0 +1,1007 @@
+"""
+Classes used to build the CIME Macros file.
+
+The main "public" class here is MacroMaker. Other classes are mainly for
+internal use, though the various "MacroWriter" classes might be useful
+elsewhere one day.
```
jgfouca: This is a lot of new code and appears to be a fairly major change. Is there any way we could get a little more documentation here? What is the design? Why is it better than before?

utils/perl5lib/compilers_translation_tool.pl
quantheory: Will do.

utils/python/CIME/XML/machines.py
quantheory: Yep, but I just noticed that the doctest above will fail now. I'll rerun doctest and pylint next week and clean this up.

utils/python/CIME/macros.py
quantheory: I can definitely do more in the way of design documentation. I'm also considering putting the "Writer" classes in a different module, if that would help navigate through this module.

scripts/create_newcase
quantheory: Good idea. I had the same feeling about the API changes, but it didn't occur to me to just switch to an environment variable.

utils/python/tests/test_config_build.py
quantheory: OK. What's the preferred way to do this? Do you want me to copy it into the same file, or turn this into a module loaded by `scripts_regression_tests`?

jgfouca: Go ahead and copy it there.

quantheory:
@jgfouca I think that I've addressed the above concerns. Let me know what you think now.

jgfouca:
@quantheory this looks great! Well tested and I love that you used those pylint directives in-code to keep pylint clean. Even I don't do that, but maybe I we should start.

PR #229, opened 2016-06-21
Another large set of changes, trying to get to all pass on skybridge

jgfouca:
1) Add option to jenkins generic job to allow it to use a different root.
This allows us to regression test it without worrying about clashing
with any "real" jenkins jobs that may be running.

2) TESTRUNDIFF can now alternate between generating v1 and v2 of the
fake history file. Makes it much easier to test. This also fixes
the bless_test_results regression test that has been broken for a while.

3) An exception must not prevent the TestStatus file from
being marked FAIL.

4) Fix huge number issues flagged by pylint. Let's not use non-immutable
values as default values for function argument anymore, it's not safe.

5) CRITICAL FIX: scripts_regression_tests now removes all .pyc files before
starting. I think stale pyc's is why so many problems have been slipping through.

6) MAJOR CHANGE: the logging system has been refactored. The setup is much
cleaner and easier to understand than before. The new system is also enhanced:

    Guide to logging in CIME.

    logger.debug -> Verbose/detailed output, use for debugging, off by default. Goes to a .log file
    logger.info -> Goes to stdout (and log if --debug). Use for normal program output
    logger.warning -> Goes to stderr (and log if --debug). Use for minor problems
    logger.error -> Goes to stderr (and log if --debug)

With the logging change ALL terminal printing in CIME should be done
through the python logging system.

7) Walltime was not being propagated correctly by create_test if it was being supplied
by batch directives. The batch directives were being created with the case.test
and case.run scripts in the case-setup phase. The XML change for walltime wasn't
happening until run-phase and had no affect on the already-generated files.

8) Add CIME_GLOBAL_WALLTIME and set it in scripts_regression_tests to set a
10 minute walltime for all batch jobs created by the regression tests.

9) Improvements to bless_test_results regression test.

10) scripts_regression_tests now handles logging arguments.

11) Skybridge has 16 cores per node, not 8

12) SEQ test was broken, rootpe was being set too high by one

13) Remove before copy is more robust on HPCs

14) Re-add timing information at the end of all TestStatus files

15) CRITICAL FIX: An uncaught exception in a test should NEVER leave
TestStatus in a PASS state

16) Change ERI_D.f45_g37.X to ERI.f45_g37.X in cime_developer. The debug
version was taking way too long.

With these changes, skybridge should be back to passing for our nightly
testing.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes: [CIME Github issue #]

Code review: @quantheory


jgfouca:
@quantheory Please try to review this today. The nightlies will be broken until this goes in and I don't want to merge any other PRs until the nightlies are working.

utils/python/CIME/SystemTests/nck.py
```diff
@@ -107,6 +107,7 @@ def run(self):
         #======================================================================
 
         if success:
+            os.remove("%s/%s.exe" % (exeroot, cime_model))
```
quantheory: Can you explain what was happening that made this change necessary? Is it sometimes the case that the copy below fails, so you need to remove this file to ensure that the old executable isn't used?

utils/python/CIME/XML/env_batch.py
```diff
@@ -249,7 +249,7 @@ def set_job_defaults(self, bjobs, pesize=None):
                 task_count = int(task_count)
             queue = self.select_best_queue(task_count)
             self.set_value("JOB_QUEUE", queue, subgroup=job)
-            walltime = self.get_max_walltime(queue)
+            walltime = self.get_max_walltime(queue) if walltime is None else walltime
```
quantheory: Not clear to me when `walltime is None` will actually be true. Is this based on `os.environ['KEY']` being `None` when `KEY` is not in the environment?

quantheory: Or more specifically, `os.environ['CIME_GLOBAL_WALLTIME']` will be `None`?

quantheory: Sorry, `os.getenv('CIME_GLOBAL_WALLTIME')` is what you used above. I was wondering why you wouldn't just get a `KeyError`, but now I see that you used `getenv` rather than the dictionary-type interface.

quantheory:
The changes to `CIME.XML.grids` seem like bug fixes. Since I'm not as familiar with that file, can you explain? They don't seem to match the points you mentioned above.

utils/python/CIME/build.py
```diff
@@ -80,7 +79,7 @@ def build_model(case, build_threaded, exeroot, clm_config_opts, incroot, complis
     for model, comp, nthrds, ninst, config_dir in complist:
         if comp == "aquap":
             logger.debug("Now build aquap ocn component")
-            _build_model_thread(config_dir, caseroot, bldroot, libroot, incroot, file_build, 
+            _build_model_thread(config_dir, comp, caseroot, bldroot, libroot, incroot, file_build,
```
quantheory: Were aquaplanet builds simply failing before this change? I guess it's quite possible that no one has been testing them recently.

utils/python/CIME/buildnml.py
```diff
@@ -124,9 +126,9 @@ def build_data_nml(argv, compclass):
             # single-case restart for each instance
             rpointer = "rpointer." + compname
             if (os.path.isfile(os.path.join(rundir,rpointer)) and
-                (not os.path.isfile(os.path.join(rundir,rpointer+inst_string)))):
-                     shutil.copy(os.path.isfile(os.path.join(rundir,rpointer),
```
quantheory: Is removing the `isfile` here a bug fix for multi-instance runs?

utils/python/CIME/case.py
```diff
@@ -465,10 +467,10 @@ def configure(self, compset_name, grid_name, machine_name=None,
                     expect(response.startswith("u"), "Aborting by user request")
 
         # the following go into the env_mach_specific file
-        vars = ("module_system", "environment_variables", "mpirun")
+        items = ("module_system", "environment_variables", "mpirun")
```
quantheory: What's the reasoning behind this name change?

utils/python/CIME/case_run.py
```diff
@@ -131,18 +127,18 @@ def postRunCheck(case, lid):
...
         expect (False, msg)
     else:
-        if 'SUCCESSFUL TERMINATION' in open(cpl_logfile).read():
-            msg = "Run SUCCESSFUL"
-            append_status(msg, caseroot=caseroot, sfile="CaseStatus" )
-        else:
-            msg = "Model did not complete - see %s \n " %(cpl_logfile)
-            append_status(msg, caseroot=caseroot, sfile="CaseStatus")
-            expect (False, msg)
+        with open(cpl_logfile, 'r') as fd:
```
quantheory: :+1:

utils/python/CIME/SystemTests/nck.py
jgfouca: The copy below was failing on certain systems due to "file busy" type errors. I found that removing the file before the copy prevented these errors for reasons that are not entirely clear to me.

utils/python/CIME/XML/env_batch.py
jgfouca: So is it clear now or do you need me to comment?

utils/python/CIME/buildnml.py
jgfouca: The old code is broken. Someone was trying to copy the result of isfile, which is a bool.

utils/python/CIME/task_maker.py
```diff
@@ -41,6 +41,9 @@ def __init__(self, case, remove_dead_tasks=False):
         self._compute_values()
 
     def _compute_values(self):
+        """
+        JGF: This code really needs a refactor/rethink
+        """
```
quantheory: Yeah, I think that when you have a function named `_compute_values`, it's a bit of a red flag that it might have kind of a vaguely-defined purpose...

jgfouca:
@quantheory thanks for looking at this. If all your concerns are addressed, go ahead and click the merge button.

utils/python/CIME/XML/env_batch.py
quantheory: I understand now.

quantheory:
I still need to look at the logging changes, but I'll do that right after lunch.

utils/python/CIME/XML/grids.py
```diff
@@ -30,12 +30,12 @@ def get_grid_info(self, name, compset):
         atmnlev = None
         lndnlev = None
         #mechanism to specify atm levels
-        levmatch = re.match("([^_]+)z(\d+)(.*)$", name)
+        levmatch = re.match(r"([^_]+)z(\d+)(.*)$", name)
```
jgfouca: Bugs caught by pylint.

utils/python/CIME/XML/grids.py
```diff
@@ -150,9 +150,9 @@ def _get_gridmaps(self, component_grids, atmnlev=None, lndnlev=None):
                             logger.debug(" %s: %s" %gridmap)
 
         if atmnlev is not None:
-            grids["atm_grid"] += "z"+atmnlev
+            grids[0] += "z"+atmnlev
```
quantheory: This is the change that confused me in this file. I'm OK with this change, but the fact that we are using lists of tuples in this file at all seems to be a weird Perl-ism. I've created a separate issue for that, #232.

utils/python/CIME/utils.py
```diff
@@ -799,3 +797,20 @@ def transform_vars(text, case=None, subgroup=None, check_members=None, default=N
 def get_my_queued_jobs():
     # TODO
     return []
+
+def wait_for_unlocked(filepath):
```
quantheory: What is this for? I see that you've added this function, but I don't see it being called anywhere?

jgfouca: I was using this to try to get around the "busy file" errors I was getting when copying model.exe. I then stumbled upon the "remove it first" technique which seemed to work a lot better. I left this function in in case it would be useful in the future.

PR #233, opened 2016-06-29
Fix typo in `system_tests_common`.

quantheory:
As-is, there's a line that's trying to use `CASEBASEID` as a
relative path to a case directory, which makes no sense.
I think that it meant to use a `%` instead of `,`, so that
`CASEBASEID` is part of the error message, not an
argument to `append_status`.


quantheory:
Oops, I had a typo in the commit message (fittingly enough). Force-pushed a fix for that...

jgfouca:
Thanks for fixing.

PR #235, opened 2016-07-05
Change format of `compare_namelists` output.

quantheory:
The effect of this change is to sort `compare_namelists` output, and to tweak the output to make it easier to compare new and old namelists. Most namelist settings are "scalars" (not arrays or dictionary-like string specifications), so the output will look something like this:

```
Differences in namelist 'foo':
  BASE: param1 = 5.
  COMP: param1 = 6.
  BASE: bar_scheme = 'something'
  COMP: bar_scheme = 'other'
```

This should make it a little easier to see what the change to a namelist is, compared with the "inequivalent values" version now. (The current version is not too hard to parse either, but it's a little verbose, and I find it hard to remember with confidence whether the left or right side is supposed to be the baseline or "gold" setting.)

### DIFF
--- a/utils/python/compare_namelists.py
+++ b/utils/python/compare_namelists.py
@@ -211,14 +211,15 @@ def normalize_string_value(name, value, case):
         return value
 
 ###############################################################################
-def compare_values(namelist, name, gold_value, comp_value, case):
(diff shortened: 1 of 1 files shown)


jgfouca:
Looks like a nice update, thanks.

PR #238, opened 2016-07-05
Fix errors when running `cs.status.*`.

quantheory:
There were several errors related to baseline tests, including memory
and throughput baselines, when running the `cs.status.*` script. (This
is the script generated in the root directory for a test suite, for
checking the current status of all tests in the suite.)

Most of the necessary changes are in `system_tests_common`, where the
output produced using `append_status` needs to use a format and phase
identifiers that can be recognized by
`wait_for_tests.parse_test_status`. `parse_test_status` also needs to
recognize the memory and throughput comparisons as valid test phases, as
well as recognizing "BFAIL" (the status resulting from a *missing*
baseline directory) as a valid status.

Testing:

   There are no formal regression tests for this functionality, but the
   CAM tests on yellowstone have been run with these changes, and
   cs.status runs without throwing an exception.

### DIFF
--- a/utils/python/CIME/SystemTests/system_tests_common.py
+++ b/utils/python/CIME/SystemTests/system_tests_common.py
@@ -239,8 +239,8 @@ class SystemTestsCommon(object):
         basecmp_dir = os.path.join(baselineroot, self._case.get_value("BASECMP_CASE"))
         for bdir in (baselineroot, basecmp_dir):
             if not os.path.isdir(bdir):
-                append_status("GFAIL %s baseline\n"%self._case.get_value("CASEBASEID"),
-                             sfile="TestStatus")
+                append_status("BFAIL %s compare\n"%self._case.get_value("CASEBASEID"),
+                              sfile="TestStatus")
                 append_status("ERROR %s does not exist"%bdir, sfile="TestStatus.log")
                 return -1
         compgen = os.path.join(self._case.get_value("SCRIPTSROOT"),"Tools",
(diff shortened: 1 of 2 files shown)


utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -297,7 +303,7 @@ def generate_baseline(self):
         basegen_dir = os.path.join(baselineroot, self._case.get_value("BASEGEN_CASE"))
         for bdir in (baselineroot, basegen_dir):
             if not os.path.isdir(bdir):
-                append_status("GFAIL %s baseline\n" % self._case.get_value("CASEBASEID"),
+                append_status("FAIL %s generate\n" % self._case.get_value("CASEBASEID"),
```
jgfouca: Shouldn't this be BFAIL too?

jgfouca:
Looks good, just one question.

utils/python/CIME/SystemTests/system_tests_common.py
quantheory: When generating a baseline fails, we have typically called this a "GFAIL", but this is basically the same thing as failing the "generate" phase, so that's what I marked this as. The most likely reason to fail at this point would, I think, be if there's some kind of permissions issue that stops baseline generation, or if there's a scripts bug. To be honest, though, I have not thought about this too deeply, since during the testing I was doing I was only comparing rather than generating baselines. Getting rid of the "GFAIL" was an afterthought.

In contrast, we probably do want "BFAIL" to be a separate status because it doesn't necessarily signify an error. For instance, if you've just added a new test and there is no baseline for that configuration, a "BFAIL" is perfectly normal. I'm wondering if I should add it to `reduce_stati` to somehow emphasize the fact that (like "NLFAIL") it's not as severe as a "FAIL" or "DIFF" status.

jgfouca: Dealing with NLFAIL was a pain. I'd prefer fewer status types rather than more. The user can always get more info form TestStatus.log.

quantheory:
Does this work? Just using a FAIL of the compare phase instead of BFAIL.

jgfouca:
Yes, thank you.

PR #240, opened 2016-07-06
Single submit multiple compilers

jedwards4b:
Fix single-submit option on hobart, allow test suite to invoke multiple compilers.

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes:

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/cesm/machines/config_batch.xml
+++ b/cime_config/cesm/machines/config_batch.xml
@@ -57,7 +57,7 @@
 
    <batch_system type="cobalt" version="x.y">
      <batch_query>qstat</batch_query>
-     <batch_submit>bash</batch_submit>
+     <batch_submit>qsub</batch_submit>
      <batch_directive></batch_directive>
      <jobid_pattern>(\d+)</jobid_pattern>
      <depend_string> --dependencies</depend_string>
--- a/cime_config/cesm/machines/config_batch.xml
+++ b/cime_config/cesm/machines/config_batch.xml
@@ -91,8 +91,8 @@
       <directive                       > -R "span[ptile={{ ptile }}]"</directive>
       <directive                       > -N  </directive>
       <directive default="poe"         > -a {{ poe }} </directive>
-      <directive default="cesm.stdout" > -o {{ cesm_stdout }}.%J  </directive>
-      <directive default="cesm.stderr" > -e {{ cesm_stderr }}.%J  </directive>
+      <directive default="cesm.stdout" > -o {{ output_error_path }}.%J  </directive>
+      <directive default="cesm.stderr" > -e {{ output_error_path }}.%J  </directive>
       <directive                       > -J {{ job_id }} </directive>
     </directives>
   </batch_system>
--- a/cime_config/cesm/machines/config_batch.xml
+++ b/cime_config/cesm/machines/config_batch.xml
@@ -61,7 +61,7 @@
      <batch_directive></batch_directive>
      <jobid_pattern>(\d+)</jobid_pattern>
      <depend_string> --dependencies</depend_string>
-    <walltime_format>%H:%M:%S</walltime_format>
+    <walltime_format>%M</walltime_format>
      <submit_args>
        <arg flag="--cwd" name="$CASEROOT"/>
        <arg flag="-A" name="$PROJECT"/>
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -291,7 +291,7 @@ def single_submit_impl(machine_name, test_id, proc_pool, project, args, job_cost
     directives = directives.replace("{{ job_queue }}", queue)
     if project is not None:
         directives = directives.replace("{{ project }}", project)
-        
+
     expect("{{" not in directives, "Could not resolve all items in directives:\n%s" % directives)
 
     #
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -329,8 +329,8 @@ def create_test(testargs, compiler, machine_name, no_run, no_build, no_batch, te
                 if compiler is None:
                     compiler = testsplit[5]
                 else:
-                    logger.info("Switch compiler %s to %s"%(compiler, testsplit[5]))
-                    compiler = testsplit[5]
+                    expect(compiler == testsplit[5],
+                           "ambiguity in compiler, please use the --compiler option")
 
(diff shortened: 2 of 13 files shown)


cime_config/cesm/machines/config_batch.xml
```diff
@@ -91,8 +91,8 @@
       <directive                       > -R "span[ptile={{ ptile }}]"</directive>
       <directive                       > -N  </directive>
       <directive default="poe"         > -a {{ poe }} </directive>
-      <directive default="cesm.stdout" > -o {{ cesm_stdout }}.%J  </directive>
-      <directive default="cesm.stderr" > -e {{ cesm_stderr }}.%J  </directive>
+      <directive default="cesm.stdout" > -o {{ output_error_path }}.%J  </directive>
```
jgfouca: Should these both be "output_error_path"? I was expecting to see different values for output and errput.

utils/python/CIME/XML/env_batch.py
```diff
@@ -282,6 +282,8 @@ def get_batch_directives(self, case, job, raw=False):
                     default = node.get("default")
                     if not raw:
                         directive = transform_vars(directive, case=case, subgroup=job, default=default, check_members=self)
+                    elif default is not None:
+                        directive = transform_vars(directive, case=case, subgroup=job, default=default)
```
jgfouca: Case and subgroup should be None (or left unprovided) here. We want transform to only use the default.

scripts/create_test
```diff
@@ -319,13 +324,13 @@ def create_test(testargs, compiler, machine_name, no_run, no_build, no_batch, te
...
                     expect(machine_name == testsplit[4],
-                           "abiguity in machine, please use the --machine option")
+                           "ambiguity in machine, please use the --machine option")
             if testsplit[5] is not None:
                 if compiler is None:
                     compiler = testsplit[5]
                 else:
-                    expect(compiler == testsplit[5],
-                           "abiguity in compiler, please use the --compiler option")
+                    logger.info("Switch compiler %s to %s"%(compiler, testsplit[5]))
```
jgfouca: I don't know if this is safe. Not this particular line, but having multiple tests with varying compilers. If you look at system_test.py, there are some built-in assumptions that self._compiler is consistent across all tests. For example, self._baseline_cmp_name is derived using self._compiler among other things. If you allow compilers to vary across tests, the current code will use the wrong baselines for some of them.

jgfouca:
@jedwards4b : One minor fix needed for the single-submit upgrade. Unfortunately, the compiler change is not as simple as the changes you made. It may be fairly easy to fix this. Maybe use a placeholder like "<COMPILER>" when self._baseline_cmp_name and self._baseline_gen_name are created in the constructor. I think _xml_phase and _nlcomp_phase are the only phases that use the baseline compare member. These phases have access to the fully qualified test case name, which includes the compiler, so they should be able to resolve the real baseline dirs at the time they are run.

jedwards4b:
It turns out that the single submit code won't work on many hpc systems where you are not allowed to compile on compute nodes - yellowstone is like this, as are most cray systems.  I think that we need to disable this test in scripts_regression_tests.  Is there a way we can get cdash to report the number of tests that run and the number that pass instead of a single pass/fail?

jgfouca:
@jedwards4b , if that's a common restriction, then the maybe single-submit should never build on compute nodes which is easy to do. It's currently set to save the build and run phases for the single submission, it would be very easy to just save the run phase.

cime_config/cesm/machines/config_batch.xml
jedwards4b: typically we put them in the same file - not sure why they were separated just for yellowstone, don't need to be,

utils/python/CIME/XML/env_batch.py
jedwards4b: agreed - updated.

scripts/create_test
jedwards4b: I've changed it so that this feature is only used with the --xml options and so will not conflict with acme's style of compare and generate usage.

utils/python/CIME/XML/testlist.py
```diff
@@ -64,16 +64,16 @@ def _get_testsv2(self, machine=None, category=None, compiler=None):
...
         for tnode in testnodes:
             machnodes = self.get_nodes("machine",machatts,root=tnode)
-            thistest = {}
-
             if machnodes:
+                thistnode = {}
                 for key, value in tnode.items():
                     if key == "name":
-                        thistest["testname"] = value
+                        thistnode["testname"] = value
```
jgfouca: Should be thisnode instead of thistnode? Also, underscores are nice to seperate words in varnames. this_is_more_readable than thisislessreadable.

utils/python/tests/scripts_regression_tests
```diff
@@ -904,9 +904,11 @@ class TestSingleSubmit(TestCreateTestCommon):
         # Skip unless on a batch system and users did not select no-batch
         if (not self._hasbatch):
             self.skipTest("Skipping single submit. Not valid without batch")
+        if CIME.utils.get_model() != "acme":
+            self.skipTest("Skipping single submit. ACME experimental feature")
 
         # Keep small enough for now that we don't have to worry about load balancing
-        stat, output, errput = run_cmd("unset CIME_GLOBAL_WALLTIME && %s/create_test SMS_Ln9_P8.f45_g37_rx1.A SMS_Ln9_P8.f19_g16_rx1.A  -t %s --single-submit" % (SCRIPT_DIR, self._baseline_name), ok_to_fail=True)
+        stat, output, errput = run_cmd("unset CIME_GLOBAL_WALLTIME && %s/create_test SMS_Ln9_P4.f45_g37_rx1.A SMS_Ln9_P4.f19_g16.X  -t %s --single-submit" % (SCRIPT_DIR, self._baseline_name), ok_to_fail=True)
```
jgfouca: Why were these test cases changed?

jgfouca:
@jedwards4b , couple last comments, then I'll merge.

utils/python/tests/scripts_regression_tests
jedwards4b: Changed as part of trying to get it to work on yellowstone - i will revert this.

PR #243, opened 2016-07-08
Fixing batch parameters for blues

jayeshkrishna:
Fiix the batch parameters for blues.

Test status: BFB
User interface changes?: No

Code review: R. Jacob

Adding batch type and fixing the wallclock format for blues.
Without this fix case.setup will fail on blues.


rljacob:
Verified I can create, setup and build an X case on blues.

Issue #248, opened 2016-07-12
Misleading error message in case_setup.py 


wilke:
I suggest changing "%s NINST value %d greater than %s NTASKS %d" 
(e.g ERROR: ATM NINST value 1 greater than ATM NTASKS 0)

to 

"NINST_%s value %d greater than NTASKS_%s %d
(e.g. ERROR: NINST_ATM value 1 greater than NTASKS_ATM 0)

to reflect the real variable name which can be queried or changed with xmlquery/xmlchange

PR #249, opened 2016-07-12
Changed error message, variable name identical to entry in xml now

wilke:
Change in error message, prints real variable name now.

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes: #248 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/case_setup.py
+++ b/utils/python/CIME/case_setup.py
@@ -76,7 +76,7 @@ def _build_usernl_files(case, model, comp):
                     shutil.copy(model_nl, nlfile)
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


utils/python/CIME/case_setup.py
```diff
@@ -76,7 +76,7 @@ def _build_usernl_files(case, model, comp):
                     shutil.copy(model_nl, nlfile)
 
 ###############################################################################
-def case_setup(case, clean=False, test_mode=False, reset=False):
+def  (case, clean=False, test_mode=False, reset=False):
```
jedwards4b: What happened to the function name?

PR #252, opened 2016-07-13
Make `Case` a context manager.

quantheory:
This change guarantees that a `Case` is always flushed, even if an
exception is thrown, using this syntax:

```python
with Case(caseroot) as case:
    # do stuff with the case
```

I haven't tested this much yet, so it shouldn't be merged, but I wanted to get feedback on this change.

### DIFF
--- a/cime_config/acme/machines/template.case.test
+++ b/cime_config/acme/machines/template.case.test
@@ -66,7 +66,7 @@ def _main_func(description):
         sys.exit(1 if test_results.failed > 0 else 0)
 
     caseroot, testname = parse_command_line(sys.argv, description)
-    with Case(caseroot) as case:
+    with Case(caseroot, read_only=False) as case:
         success = case_test(case, testname)
 
     sys.exit(0 if success else 1)
--- a/cime_config/acme/machines/template.case.test
+++ b/cime_config/acme/machines/template.case.test
@@ -66,8 +66,8 @@ def _main_func(description):
         sys.exit(1 if test_results.failed > 0 else 0)
 
     caseroot, testname = parse_command_line(sys.argv, description)
-    case = Case(caseroot)
-    success = case_test(case, testname)
+    with Case(caseroot) as case:
+        success = case_test(case, testname)
 
     sys.exit(0 if success else 1)
 
--- a/utils/python/CIME/build.py
+++ b/utils/python/CIME/build.py
@@ -139,7 +139,7 @@ def post_build(case, logs):
     shutil.copy("env_build.xml", "LockedFiles")
 
 ###############################################################################
-def case_build(caseroot, case=None, sharedlib_only=False, model_only=False):
+def case_build(caseroot, case, sharedlib_only=False, model_only=False):
 ###############################################################################
 
     t1 = time.time()
@@ -156,7 +156,6 @@ def case_build(caseroot, case=None, sharedlib_only=False, model_only=False):
     expect(os.path.exists("case.run"),
            "ERROR: must invoke case.setup script before calling build script ")
 
-    case = Case() if case is None else case
     cimeroot = case.get_value("CIMEROOT")
 
     comp_classes = case.get_value("COMP_CLASSES").split(',')
--- a/cime_config/acme/machines/template.case.run
+++ b/cime_config/acme/machines/template.case.run
@@ -74,7 +74,7 @@ def _main_func(description):
         sys.exit(1 if test_results.failed > 0 else 0)
 
     caseroot = parse_command_line(sys.argv, description)
-    with Case(caseroot) as case:
+    with Case(caseroot, read_only=False) as case:
         success = case_run(case)
 
     sys.exit(0 if success else 1)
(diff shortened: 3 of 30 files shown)


cime_config/acme/machines/template.case.test
```diff
@@ -66,8 +66,8 @@ def _main_func(description):
         sys.exit(1 if test_results.failed > 0 else 0)
 
     caseroot, testname = parse_command_line(sys.argv, description)
-    case = Case(caseroot)
-    success = case_test(case, testname)
+    with Case(caseroot) as Case:
```
jedwards4b: Case should be case here.   Have you run the scripts_regression_tests on this PR?

jedwards4b:
It strikes me as a bit of an awkward construct but I can live with it if its useful, what do others think?

gold2718:
I haven't yet looked at the implementation but I have no issues with the usage syntax. I use it all the time for file operations and find it useful since the scope of the object is well defined.

utils/python/CIME/build.py
```diff
@@ -139,7 +139,7 @@ def post_build(case, logs):
     shutil.copy("env_build.xml", "LockedFiles")
 
 ###############################################################################
-def case_build(caseroot, case=None, sharedlib_only=False, model_only=False):
+def case_build(caseroot, case, sharedlib_only=False, model_only=False):
```
jgfouca: Did you confirm that no client was using this default argument?

jgfouca:
This looks really great! Is there any way to block users from using it without "with"?

jgfouca:
Maybe we can set a flag that __enter__ was run and then fail if user tries to do anything if that flag is false?

jgfouca:
Actually, we'd only need to block set_value calls if they did not use with.

cime_config/acme/machines/template.case.test
quantheory: No. I'm going to test today.

utils/python/CIME/build.py
quantheory: Yep!

quantheory:
Probably no one is using this, but just a heads up, I'm going to resolve the merge conflicts with a rebase.

@jgfouca I was also thinking that the `__enter__` could do some kind of "unlock" that would be necessary to edit the case. Besides `set_value` calls, we also need to prohibit edits to `env_batch` and `env_mach_specific`, which are not `EntryID`-formatted files. I think that instead of modifying `_env_files_that_need_rewrite` directly, we should mark files as needing a rewrite through a method on `Case`, and that method should check to make sure that we've unlocked the `Case` object.

A bonus of this approach is that if you want read-only access (e.g. in `xmlquery`), you can avoid using the `with` statement, and you can be sure that you aren't accidentally changing values in the file.

jgfouca:
@quantheory : I like that a lot.

gold2718:
If read-only is a desire, how about adding an optional read_only(=.false.) argument to the Case constructor? That way, we can standardize on one usage (with Case...) and still make it clear when the usage is read-only.
I argue that this approach is less obfuscatory than how the constructor is invoked.

jgfouca:
@gold2718 , I don't think that will work. We have to rely on python's "with" in order to ensure that the Case object is "destructed".

gold2718:
I'm not arguing against using Case in a with statement, quite the opposite. Sean was suggesting that creating a Case outside of with statement would automatically invoke read-only status as protection. I am arguing for consistent usage; If we move to having Case be a context manager, we always use it that way. Adding a read_only argument to the constructor is an explicit way to show intent for a particular block of code.
To reinterate my message from yesterday (see above), I am in favor of adding the context-manager methods to Case.

jgfouca:
@gold2718 , or so you're saying you want this:

    with Case(readonly=True) as case:
        ...

The problem with that is that it's unenforceable unless we get pretty fancy with the Case class definition. We can use it as a convention, but users will be able to construct Case objects without with.

jgfouca:
@gold2718 just to clarify, if we do what you suggest, it would open the opportunity for a user to do a non-managed construction of a writable Case object (if I'm understanding you correctly).

quantheory:
Hmm, I just committed something like Steve's suggestion (in 215987a):

1. A case constructed with `read_only=True` is permanently read-only, no matter what. This is also the default. The `with` syntax is used by convention, but it doesn't actually do anything different from `case = Case()`. It could do something in the future, though, if we decide we need some more complex locking mechanism for cases.
2. A case constructed with `read_only=False` can only be edited within a `with` block. You can construct such a case outside of a `with` block, but there's no reason to, since you need to start a `with` block anyway in order to call `set_value`.

The prior commit (d7ae79b) is an implementation of the earlier locking plan, i.e. there is no `read_only` flag for the constructor, but the data is still locked against writes outside of `with` blocks. So if we don't want the `read_only` flag after all, I can roll back to that commit.

gold2718:
@quantheory , this does look like what I was suggesting. Is another way of stating #2 that the read_only flag has no effect outside of a with block?
If people don't like the sense of the flag, we could writable which would be .false. by default.
Alternatively, we could use the same mode arguments as file opens (e.g., 'r', 'w','r+' -- 'e' would be equivalent to 'r+').

jgfouca:
Ready to merge?

quantheory:
@gold2718 Yep, pretty much what you said.

@jgfouca Yeah, if no one has any complaints, let's do it.

quantheory:
It occurs to me to ping @mvertens about this. This might cause a compatibility issue with the CICE buildnml if she has rewritten that in Python yet.

PR #259, opened 2016-07-14
Port performance archiving scripts to python.

jgfouca:
1) Add SAVE_TIMING_DIR run parameter
2) Define this paramter for melvin and skybrige
3) Add --save-timing to create_test to enable timing archiving via create_test
4) Fix exception hiding in system_test_common
5) Convert case_run.py to PEP-compliant naming conventions for functions
6) Add utility functions for easy zipping and touching of files
7) Add regression test that runs create_test with --save-timing

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes: [CIME Github issue #]

Code review: @jedwards4b @rljacob, @worleyph


utils/python/CIME/case_run.py
```diff
@@ -137,7 +137,162 @@ def postRunCheck(case, lid):
...
+        return os.environ("PBS_JOBID")
+    elif mach in ['edison', 'corip1']:
+        return os.environ("SLURM_JOB_ID")
+    elif mach == 'mira':
+        return os.environ("COBALT_JOBID")
+    else:
+        return None
+
+###############################################################################
+def save_timing_setup_acme(case, lid):
```
jgfouca: Key changes are here.

jedwards4b:
Did this change the behavior on the cesm side?   If not then it's fine with me.

jgfouca:
@jedwards4b , no it should not have.

worleyph:
@jgfouca , please indicate what you want me to do. I have not been involved in CIME testing (since CIME2 was integrated into ACME).

jgfouca:
@worleyph I'd like you to confirm that the new python which I tagged meets your needs.

cime_config/acme/machines/cesm_postbuild.perf_archive
```diff
@@ -1,78 +0,0 @@
...
-
-#--- save model provenance with the executable
-cd $cesmroot
-
-if (-d $cesmroot/.git) then
-  git describe > $exe_dir/GIT_DESCRIBE.$lid
-  chmod 664 $exe_dir/GIT_DESCRIBE.$lid
-  /bin/cp -p $exe_dir/GIT_DESCRIBE.$lid $exe_dir/GIT_DESCRIBE
-
-  if (-f $cesmroot/.git/logs/HEAD) then
```
jgfouca: I did not copy logs. This info can be derived from GIT_DESCRIBE if needed.

cime_config/acme/machines/cesm_postrun.perf_archive
```diff
@@ -1,153 +0,0 @@
...
-    if ($machine == 'titan') then
-
-      set string = `ls $app_jobid*OU`
-      cp -p $string $save_timing_dir/performance_archive/$user/$case/$lid/$string.$lid
-      chmod a+r $save_timing_dir/performance_archive/$user/$case/$lid/$string.$lid
-
-    else if ($machine == 'mira') then
-
-      set string = `ls $app_jobid*output`
-      cp -p $string $save_timing_dir/performance_archive/$user/$case/$lid/$string.$lid
```
jgfouca: I did not add $lid to any of the archived files. They are already under a directory with $lid as the name.

cime_config/acme/machines/cesm_prerun
```diff
@@ -1,128 +0,0 @@
...
-    set app_jobid = $SLURM_JOB_ID
-
-  else if ($machine == 'mira') then
-
-    set app_jobid = $COBALT_JOBID
-
-  endif  
-endif
-
-if ($machine == 'titan') then
```
jgfouca: I did not port any of the CaseStatus additions. Those don't belong in archiving code.

worleyph:
Since I do not know python, this may be difficult for me except by running the code and seeing what it does. I would then need a way to run it in a way that exercises this capability. 

If you are eliminating capturing the logs, then please educate me on the correct approach (directly). This was a bandaid that never got updated. (I did not like what I was told was adequate at the time, and but did not pursue it then.)

cime_config/acme/machines/cesm_prerun
worleyph: This was meant to provide augmentations that were not available in the 
original CESM code, to make it easier to identify what these 
augmentations were. Guess I'll have to see what is lost with the new 
implementation and then propose reintroducing what I found important.

Note that I like having redundant provenance (e.g. $lid suffix), and 
will likely argue to put these back in.

jgfouca: Will the lack of $lid suffix break anything? If not, we can have that discussion anytime.

jgfouca:
@worleyph , the logs should be captured. You can checkout this branch and run create_test with --save-timings. Do I have access to any of the post-processing scripts that use this data?

cime_config/acme/machines/cesm_prerun
worleyph: It will break any postprocessing scripts that look for this naming 
convention. I have a few, and so do some others. I also do not like 
relying on provenance by directory name, so the files whose names I've 
taken responsibility for I've tended to append $lid to. There are also 
some files that also live outside of the performance_archive, and 
require this to prevent being overwritten. I don't think that these are 
among the ones you are touching, but would have to check to be sure.

If possible, could we start by duplicating the current design? I use 
this stuff every day, and everything in there was put there for some 
reason (through perhaps not all is currently relevant).

The current design was always meant to be a prototype - just waiting for 
you to do this correctly :-) - but it is also code that is likely to 
continue to evolve (especially when workflow starts hijacking or 
reimplementing some of this). The current design was to make it easy to 
require for ACME internal use, but not to subject others (CESM partners, 
potential external users of ACME after release) to it since they are 
unlikely to care or use it. Making it easy to further develop it without 
affecting non-ACME users is probably likely to continue to be important. 
(I feel the same way about the timing library, fyi.)

worleyph:
>  the logs should be captured

Guess that I don't know what we are talking about ... one comment was that I was grabbing git hash-related information in an awkward way - no argument from me - but by logs you are referring to something else?

> Do I have access to any of the post-processing scripts that use this data?

Which data? At the moment, the post-processing scripts are one offs. I have some, as does Ben Mayer (for importing into a database) and Noel (I think). Writing them for automatic monitoring has not made it to the top of anyone's todo list, though a number of us want it (but ultimately a workflow task). It is pretty common  to grab files from the $lid subdirectories and move/copy them to somewhere else. Having $lid suffixes makes this straightfoward.

rljacob:
"Add --save-timing to create_test to enable timing archiving via create_test"
That's a new option?  Not the default?

worleyph:
By default (in ACME), save-timing is disable for create_test, so that the performance_archive is not filled with uninteresting data.

cime_config/acme/machines/cesm_prerun
jgfouca: OK, I can suffix everything with LID.

jgfouca:
It's actually even worse than that. ACME's create_test is hardcoded to not save performance data. The current create_test also defaults to not saving timing, but at least provides a flag to enable it now.

jgfouca:
All tests passed. LID suffixes restored. Merging.

rljacob:
I tried running an X-case on edison (where SAVE_TIMING is TRUE) and comparing the contents of CASEROOT and EXEROOT with ACME.   Some info that was part of the performance archiving is not being generated/saved.  See issues #319 and #325

PR #260, opened 2016-07-14
Remove esmf interface

mvertens:
Removed esmf interfaces in driver code

Also made fixes to cesm config_compsets.xml to have 2 digit reference to the cam version.

Test baseline: 
Test namelist changes: none
Test status: bit for bit

Fixes: None

User interface changes?: None 

Code review:


utils/python/CIME/preview_namelists.py
```diff
@@ -72,9 +72,12 @@ def preview_namelists(case, dryrun=False, casedir=None):
         cmd = os.path.join(config_dir, "buildnml")
         logger.info("Running %s"%cmd)
         if (logger.level == logging.DEBUG):
-            run_cmd("PREVIEW_NML=1 %s %s" % (cmd, caseroot))
+            rc, out, err = run_cmd("PREVIEW_NML=1 %s %s" % (cmd, caseroot), ok_to_fail=True)
```
jgfouca: What's the reason for this change? Before/after look semantically identical but the latter is double the code.

mvertens: I need to be able to get output to the TestStatus.log file - so in case a user wants to add any DEBUG print statements, they could easily do so. At this point, any print statements that are added simply do no appear anywhere.

PR #261, opened 2016-07-14
fix nck test and multiinstance support

jedwards4b:
Fixes the NCK test for any case in which the original pe layout has a component with only 1 task.

The build is done in two stages, a sharedlib_only stage and a model_only stage. This test has dependencies which require two different sharedlib builds and they both happen before either model build occurs. The env_build.xml and env_mach_pes.xml files must be staged correctly to handle this. There was also some special handling for multiinstance cases in testmods I moved to case_setup.py. 

Test suite:  NCK_Ld5.f19_g16.BC5L45BGCRG.yellowstone_intel.allactive-cism-test_coupling,
                   scripts_regression_tests
                   SMS_N8_P8.f19_g16.X.yellowstone_intel
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes: 

User interface changes?: 

Code review: mvertens, jgfouca


utils/python/CIME/SystemTests/nck.py
```diff
@@ -23,44 +23,58 @@ def __init__(self, case):
...
+            if not sharedlib_only:
+                shutil.move("%s/%s.exe"%(exeroot,cime_model),"%s/%s.exe.NCK%s"%(exeroot,cime_model,bld))
+                shutil.copy("env_build.xml",os.path.join("LockedFiles","env_build.NCK%s.xml"%bld))
 
         # Because mira/cetus interprets its run script differently than
         # other systems we need to copy the original env_mach_pes.xml back
-        shutil.copy(machpes1,"env_mach_pes.xml")
-        shutil.copy("env_mach_pes.xml",
-                    os.path.join("LockedFiles","env_mach_pes.xml"))
+#        shutil.copy(machpes1,"env_mach_pes.xml")
```
jgfouca: Please delete.

jgfouca:
Can you describe the bug?

utils/python/CIME/SystemTests/nck.py
jedwards4b: Still not sure if I need these on mira, will clean up in my mira port.

jedwards4b:
The build is done in two stages, a sharedlib_only stage and a model_only stage.  This test has dependances which require two different sharedlib builds and they both happen before either model build occurs.  The env_build.xml and env_mach_pes.xml files must be staged correctly to handle this.  There was also some special handling for multiinstance cases in testmods I moved to case_setup.py.

PR #263, opened 2016-07-15
Sarich/new acme config pes

### LINKED ISSUE #262: Get ACME config_pes ready

sarich:
Convert config_pes.xml file from ACME/cime/machines-acme/config_pes.xml to a format readable
from CIME5 in cime_config/acme/allactive/config_pes.xml

There is some information in the old schema that isn't in the new schema:
`<PIO_NUMTASKS>, <PIO_STRIDE>, <PIO_TYPENAME>, <PIO_ROOT>
<PES_LEVEL>`, 
`TEST`, `OS` attributes, and
`<NINST_XXX>` (which was always 1)

I've ignored these because I don't know what to do with them, but let me know if the information needs to go somewhere (even as a comment)

I've replaced the few instances of
`<NTASKS_ATM>$MAX_TASKS_PER_NODE</NTASKS_ATM> `
with 
`<ntasks_atm>-1</ntasks_atm>`
I'm not sure if that's totally correct.

Closes #262


jayeshkrishna:
The PIO* tags in old config_pes.xml should go to config_pio.xml, i think. Checkout out config_pio.xml in cesm/machines/config_pio.xml, this file has grid specific and mach specific settings for PIO.

rljacob:
Was this converted by hand or did you use a script?

@jedwards4b can we copy this to every cime_config directory and the logic for finding a pe-layout based on a compset will still work?

rljacob:
Rename the file to something like config_pes.xml.all so it doesn't overwrite the existing config_pes.xml that has only the B-cases.

Also add the tool used for conversion in scripts/Tools.

jedwards4b:
I can't really review this acme only change - I think that you will need to test against several configurations and make sure it's giving the result you expect, I am happy to consult if you find that it does not.

rljacob:
@sarich go ahead and make a push with the filename change and this can be closed.  We'll test and split up with different PRs

rljacob:
Name in "config_pesall.xml" so it will actually work.

jgfouca:
Should I merge this?

jedwards4b:
yes

rljacob:
No. Jason hasn't made the changes I requested.

sarich:
The requested changes are now completed

PR #274, opened 2016-07-20
Edison port fixes

jedwards4b:
Fix issues on edison - NODENAME_REGEX was not defined for cesm config_batch entry error for acme, do not call flush if error occurs before case directory is created. 

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Code review: jgfouca

### DIFF
--- a/scripts/Tools/code_checker
+++ b/scripts/Tools/code_checker
@@ -51,7 +51,6 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
 def run_pylint(on_file):
 ###############################################################################
     pylint = find_executable("pylint")
-    expect(pylint is not None,"pylint not found")
 
     cmd = "%s --disable I,C,R,logging-not-lazy,wildcard-import,unused-wildcard-import,fixme,broad-except,bare-except,eval-used,exec-used,global-statement %s" % (pylint, on_file)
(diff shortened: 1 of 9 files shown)


scripts/Tools/code_checker
```diff
@@ -51,7 +51,6 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
 def run_pylint(on_file):
 ###############################################################################
     pylint = find_executable("pylint")
-    expect(pylint is not None,"pylint not found")
```
jedwards4b: expect here was just causing a hang in scripts_regression_tests, moving it to main solved the problem

utils/python/CIME/case.py
```diff
@@ -159,6 +159,9 @@ def copy(self, newcasename, newcaseroot, newcimeroot=None, newsrcroot=None):
         return newcase
 
     def flush(self, flushall=False):
+        if not os.path.isdir(self._caseroot):
```
jedwards4b: If flush is called (from expect) before the caseroot directory is created the wrong error message is output.

jgfouca:
Looks good. Converting case_root to a member of Case is an improvement.

PR #287, opened 2016-07-22
Remove `-mismatch_all` from NAG options in CESM.

quantheory:
The `-mismatch_all` flag turns off useful checks for detecting cases
where wrong arguments have been passed to a procedure, and also turns
off bounds-checking across procedure calls. Unfortunately, it also
raises errors for some libraries with non-standard Fortran 77
interfaces, notably including the MPI Fortran 77 interface.

This commit adds a text file called `nag_mpi_argument.txt`, which
contains a sequence of (very long) command-line options that can deal
with problematic MPI functions. An include of this file replaces the
`-mismatch_all` argument.

These NAG checks also object to the PnetCDF Fortran 77 interface, but
other NAG checks object to using both the `netcdf` and `pnetcdf` modules
from the same file, due to name clashes between the two modules. In
order to work around this, when NAG is used to compile PIO 1, support
for PnetCDF is turned off entirely.

I've tested this with a couple of prealpha tests on hobart.


quantheory:
BTW, `nag_mismatch_argument.txt` is listing every single function in the MPI 3.1 standard such that the Fortran 77 interface is considered non-standard. That's why it's so long, but that also means that it shouldn't need to be updated anytime soon (at least for MPI functions). It also has to be three separate lines because `nagfor` has (approximately) a 1024 character limit on command-line arguments!

PR #296, opened 2016-07-26
Testing fixes

jedwards4b:
Fixes issues with LII and ERP tests.   Improves specification of _P modifier in test names.   Fixes an issue with shared build conflicts between clm and dlnd.  Fixes an issue with cime_bisect unit test. 

Test suite: scripts_regression_tests on yellowstone and hobart
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes: 

User interface changes?: 

Code review: jgfouca

### DIFF
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -52,8 +52,9 @@ def parse_command_line(args, cimeroot):
     parser.add_argument("--project", "-project",
                         help="Specify a project id")
 
-    parser.add_argument("--pecount", "-pecount", choices=('S','M','L','X1','X2'), default="M",
-                        help="Specify a target size description for the number of cores")
+    parser.add_argument("--pecount", "-pecount",  default="M",
+                        help="Specify a target size description for the number of cores"
+                         "Allowed options are   ('S','M','L','X1','X2','[0-9]x[0-9]','[0-9]')")
 
     parser.add_argument("--mach-dir", "-mach_dir",
                         help="Specify the locations of the Machines directory, other than the default"
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -26,7 +26,7 @@ def parse_command_line(args, description):
     parser = argparse.ArgumentParser(
         usage="""\n%s <TEST|SUITE> [<TEST|SUITE> ...] [--verbose]
 OR
-%s xml --category [CATEGORY] [--machine ...] [--compiler ...] [ --testlist ...]
+%s --xml-category [CATEGORY] [--xml-machine ...] [--xml-compiler ...] [ --xml-testlist ...]
 OR
 %s --help
 OR
@@ -161,6 +161,8 @@ OR
                         help="Use this category key in the lookup in testlist.xml")
     parser.add_argument("--xml-testlist",
                         help="Use this testlist to lookup tests")
+    parser.add_argument("--testfile", nargs=1, default=None,
+                        help="A file containing an ascii list of tests to run")
 
     args = parser.parse_args(args[1:])
 
@@ -205,6 +207,13 @@ OR
 
     if args.test_id is None:
         args.test_id = CIME.utils.get_utc_timestamp()
+
+
+    if args.testfile is not None:
+        with open(args.testfile[0], "r") as fd:
+                args.testargs.extend( fd.read().splitlines() )
+    # Remove empty items if any
+    args.testargs = filter(None, args.testargs)
 
     return args.testargs, args.compiler, args.machine, args.no_run, args.no_build, args.no_batch,\
         args.test_root, args.baseline_root, args.clean, args.compare, args.generate, \
@@ -327,12 +336,6 @@ def create_test(testargs, compiler, machine_name, no_run, no_build, no_batch, te
                 else:
                     expect(machine_name == testsplit[4],
                            "ambiguity in machine, please use the --machine option")
-            if testsplit[5] is not None:
-                if compiler is None:
-                    compiler = testsplit[5]
-                else:
-                    expect(compiler == testsplit[5],
-                           "ambiguity in compiler, please use the --compiler option")
 
     impl = SystemTest(testargs,
                       no_run=no_run, no_build=no_build, no_batch=no_batch,
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -161,7 +161,7 @@ OR
                         help="Use this category key in the lookup in testlist.xml")
     parser.add_argument("--xml-testlist",
                         help="Use this testlist to lookup tests")
-    parser.add_argument("--testfile", nargs=1, default=None,
+    parser.add_argument("--testfile",
                         help="A file containing an ascii list of tests to run")
 
     args = parser.parse_args(args[1:])
--- a/utils/python/CIME/SystemTests/erp.py
+++ b/utils/python/CIME/SystemTests/erp.py
@@ -95,7 +95,7 @@ class ERP(SystemTestsCommon):
     def run(self):
         # run will have values 1,2
         for run in range(1,3):
-            
+
             expect(os.path.isfile(os.path.join("LockedFiles","env_mach_pes.ERP%d.xml"%run)),
                    "ERROR: LockedFiles/env_mach_pes.ERP%d.xml does not exist, run case.build"%run )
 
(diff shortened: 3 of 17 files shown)


scripts/create_newcase
```diff
@@ -52,8 +52,9 @@ def parse_command_line(args, cimeroot):
     parser.add_argument("--project", "-project",
                         help="Specify a project id")
 
-    parser.add_argument("--pecount", "-pecount", choices=('S','M','L','X1','X2'), default="M",
-                        help="Specify a target size description for the number of cores")
+    parser.add_argument("--pecount", "-pecount",  default="M",
```
jedwards4b: Needed to add support for the TASKSxCOUNT option here.  It was previously handled in create_test after create_newcase was completed, but I ran into a problem where the default pelayout was too big for the test system which caused create_newcase to throw an error before the pelayout adjustment specified in the test could be applied.    The choices option to add_argument was a nice feature, but I couldn't figure out how to use it with these additional possibilities.

scripts/create_test
```diff
@@ -161,6 +161,8 @@ OR
                         help="Use this category key in the lookup in testlist.xml")
     parser.add_argument("--xml-testlist",
                         help="Use this testlist to lookup tests")
+    parser.add_argument("--testfile", nargs=1, default=None,
```
jedwards4b: This allows the user to specify multiple tests in a file rather than on the command line.

scripts/create_test
```diff
@@ -327,12 +336,6 @@ def create_test(testargs, compiler, machine_name, no_run, no_build, no_batch, te
                 else:
                     expect(machine_name == testsplit[4],
                            "ambiguity in machine, please use the --machine option")
-            if testsplit[5] is not None:
```
jedwards4b: We were unnecessarily restricting all tests to use the same compiler.   No such restriction should be required.

utils/python/CIME/SystemTests/erp.py
```diff
@@ -142,7 +142,11 @@ def run(self):
             if not success:
                 break
 
-        return success
+        if success:
```
jedwards4b: Fixes a bug in ERP - restart was not being compared to initial run.

utils/python/CIME/SystemTests/lii.py
```diff
@@ -43,49 +43,38 @@ def build(self, sharedlib_only=False, model_only=False):
             for filename in glob.glob(r'user_nl_clm*'):
```
jedwards4b: The lii test is very specific to the clm model and compares an initial run from internally interpolated initial data to one from externally interpolated initial data.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -280,7 +280,7 @@ def compare_baseline(self):
         if not os.path.isfile(baselog):
             # for backward compatibility
             baselog = os.path.join(basecmp_dir, "cpl.log")
-        if len(memlist) > 3:
+        if baselog is not None and len(memlist) > 3:
```
jedwards4b: Avoid an error here if baselog does not exist.

utils/python/CIME/XML/env_batch.py
```diff
@@ -277,6 +277,7 @@ def set_job_defaults(self, bjobs, pesize=None, walltime=None):
             self.set_value("JOB_QUEUE", queue, subgroup=job)
             walltime = self.get_max_walltime(queue) if walltime is None else walltime
             if walltime is None:
+                logger.warn("Could not find a queue matching task count %d, falling back to depreciated default walltime parameter"%task_count)
```
jedwards4b: Improve the error message here.

utils/python/CIME/build.py
```diff
@@ -511,7 +511,7 @@ def build_libraries(case, exeroot, caseroot, cimeroot, libroot, mpilib, lid, mac
         sharedpath = os.environ["SHAREDPATH"]
         bldroot = os.path.join(sharedpath, case.get_value("COMP_INTERFACE"), esmfdir, "clm","obj" )
         libroot = os.path.join(sharedpath, case.get_value("COMP_INTERFACE"), esmfdir, "lib")
-        incroot = os.path.join(sharedpath,"include")
+        incroot = os.path.join(sharedpath, case.get_value("COMP_INTERFACE"), esmfdir, "include")
```
jedwards4b: There was a conflict between the dlnd lnd_comp_mct.mod and the clm lnd_comp_mct.mod - moving the location of the clm include directory solves this problem.

utils/python/CIME/case.py
```diff
@@ -538,11 +538,36 @@ def configure(self, compset_name, grid_name, machine_name=None,
         #--------------------------------------------
         # pe payout
         #--------------------------------------------
-        pesobj = Pes(self._pesfile)
+        match1 = False
```
jedwards4b: Moved here from system_test.py

utils/python/CIME/utils.py
```diff
@@ -322,7 +322,7 @@ def get_full_test_name(partial_test, grid=None, compset=None, machine=None, comp
             expect(arg_val is not None,
                    "Could not fill-out test name, partial string '%s' had no %s information and you did not provide any" % (partial_test, name))
             result = "%s%s%s" % (result, "_" if name == "compiler" else ".", arg_val)
-        elif (arg_val is not None):
+        elif (arg_val is not None and partial_val != partial_compiler):
```
jedwards4b: Avoid error if tests use different compilers.

utils/python/tests/scripts_regression_tests.py
```diff
@@ -39,8 +39,8 @@ def do_unit_tests(self, script,from_dir=SCRIPT_DIR):
         stat, output, _ = run_cmd("./%s --test 2>&1" % script, from_dir=from_dir)
         self.assertEqual(stat, 0, msg=output)
 
-    def test_acme_bisect_unit_test(self):
-        self.do_unit_tests("acme_bisect",from_dir=TOOLS_DIR)
+    def test_cime_bisect_unit_test(self):
```
jedwards4b: update to match script name change

jedwards4b:
annotation complete

billsacks:
@jedwards4b : Regarding the LII test: From a read-through of the code, it looks like you fixed some problems but introduced a different problem: both runs call

```python
            success = SystemTestsCommon._run(self)
```

without giving the optional suffix argument. I'm just coming to understand the test scripts, but I'm pretty sure that's a problem.

Also: I found the previous style, with explicit methods to do the first and second runs, more understandable.

Since I already having fixing the LII test and doing some rigorous testing of it on my task list: Can you please simply back out your changes to the LII test here? Then I'll submit a separate PR with a fixed LII test.

scripts/create_test
jgfouca: nargs=1 and default=None are both default values for a string option, they don't need to be specified.

scripts/create_test
jgfouca: I thought we already fixed this. Did we just forget to change this particular block?

utils/python/CIME/SystemTests/system_tests_common.py
jgfouca: It's not possible for baselog to be None here due to line 279.

utils/python/CIME/case.py
```diff
@@ -538,11 +538,36 @@ def configure(self, compset_name, grid_name, machine_name=None,
...
+                string = "NTHRDS_" + component_class
+                pes_nthrds[string] = opti_thrds
+                string = "ROOTPE_" + component_class
+                pes_rootpe[string] = 0
+        else:
+            pesobj = Pes(self._pesfile)
 
-        #FIXME - add pesize_opts as optional argument below
-        pes_ntasks, pes_nthrds, pes_rootpe = pesobj.find_pes_layout(self._gridname, self._compsetname,
+            pes_ntasks, pes_nthrds, pes_rootpe = pesobj.find_pes_layout(self._gridname, self._compsetname,
```
jgfouca: Was this indent intentional? This means that find_pes_layout is only run in the else block.

jgfouca: Nevermind, I think I understand what you did here.

jgfouca:
@jedwards4b : A couple minor changes needed and then this is ready pending resolution of @billsacks comment.

scripts/create_test
jedwards4b: fixed

scripts/create_test
jedwards4b: I fixed it for xml tests but not for tests listed on the command line or in a testfile.

utils/python/CIME/SystemTests/system_tests_common.py
jedwards4b: good point, changed to os.path.isfile(baselog)

PR #317, opened 2016-08-02
Restore LII test to cime for now

billsacks:
I want to keep this in cime until the system test infrastructure settles down

Test suite: None
Test baseline: N/A
Test namelist changes: N/A
Test status: N/A

Fixes: None

User interface changes?: No

Code review: None

### DIFF
--- a/utils/python/CIME/SystemTests/lii.py
+++ b/utils/python/CIME/SystemTests/lii.py
@@ -1,0 +1,82 @@
+"""
+Implementation of the CIME LII test.  This class inherits from SystemTestsCommon
+
+This is a CLM specific test:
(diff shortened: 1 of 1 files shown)


rljacob:
Can you update the README to document this test?

billsacks:
@rljacob : How does this README file differ from the documentation at the top of scripts/Testing/Testcases/config_tests.xml ? It seems that we have near-duplicate documentation in both places. Is that intentional? If so, I can update the README accordingly on my branch where I'm reworking the test scripts.

rljacob:
Probably not intentional.  ASFIAK, the test scripts have been split in to two directories so the documentation should only describe the tests in its directory.  I'd move it out of .xml because its not intuitive to look there.

PR #341, opened 2016-08-04
Remove `GenericXML` check for env variables.

### LINKED ISSUE #326: CIMEROOT in your environment confuses scripts in CESM checkout
(I assume this is pertinent to ACME checkouts as well, hence the issue here instead of [CESM-Development/cime](https://github.com/CESM-Development/cime))

I have `$CIMEROOT` set in my environment, pointing at my local clone of the stand-alone CIME repository. This confuses `create_newcase` and the other scripts, because it looks for the active components in `$CIMEROOT/../components` instead of `$PWD/../../components`; if I unset the environment variable then the scripts appear to fall back to the desired `$PWD/../../components` because everything works as expected.

I think I set this variable so that `machines/configure` would run correctly; @jedwards4b sent out an email in April 2015 where said

```
I've just updated the machines/configure file to the new cime directory structure and added a couple of new options/ features:

-cimeroot :  you can provide the cime root directory with a command line argument, if
                    not provided it will look for CIMEROOT in the environment if nether is present                     it will fail.

​-output_format:   Can be make or cmake, defaults to make which produces the original                                       Makefile Macros file - if set to cmake it will produce a Macros file 
                            suitable for use with cmake



We should make sure that all auxiliary tools in the cesm distribution know how to use a
configure generated Macros file to build.   If you need any help to do this with a particular tool please ask.
```

There are several mapping tools (`tools/mapping`) that depend on the `Macros` file to build correctly.

---

So basically there's a conflict between needing to have `$CIMEROOT` set when building some tools and needing to have `$CIMEROOT` NOT set when creating CESM (and potentially ACME) cases.

One possible fix would be to add using the `-cimeroot` option when calling `configure` to the mapping tool install directions, but that seems like a band-aid rather than a fix given that there are probably plenty of other tools that also call `configure`. Another would be to change the behavior of the scripts (to not look for `$CIMEROOT` from the environment) and / or `configure` (to not "require" having `$CIMEROOT` set). I'm not terribly familiar with either the scripts or `configure` so I don't know if those options are palatable, but maybe someone else can come up with a fix that works for everything?

quantheory:
When a user set a variable in their environment, it could sometimes
override settings in XML files when resolving variables. This change
prevents that from happening.

Fixes #326

Tested using `scripts_regression_tests.py`.

### DIFF
--- a/utils/python/CIME/XML/generic_xml.py
+++ b/utils/python/CIME/XML/generic_xml.py
@@ -218,8 +218,5 @@ class GenericXML(object):
             elif var == "SRCROOT":
                 srcroot = os.path.join(get_cime_root(),"..")
                 item_data = item_data.replace(m.group(), srcroot)
-            elif var in os.environ:
-                logging.debug("resolve from env: " + var)
-                item_data = item_data.replace(m.group(), os.environ[var])
 
         return item_data


quantheory:
@douglasjacobsen Please check to see if this fixes your `PIO_VERSION` issue and any other environment variable clashes you've encountered.

jgfouca:
For what it's worth, this looks good to me. @jedwards4b may have something to say about this when he gets back.

douglasjacobsen:
@quantheory: Yes it does.

Just an idea, could it be a good idea to have a flag on case.setup that allows environment variables to override the XML files? To allow the previous usage, but maybe not make that the default?

I suspect, since @jgfouca said that was the design of it that there was some motivation behind that, I just found that aspect very annoying from a user perspective and I could see it causing lots of problems with users.

Another idea though, might be to have a script that sets the XML file values based on a users environment.

Anyway, I might wait until @jedwards4b gets back, just to make sure he's not against this type of change.

quantheory:
@douglasjacobsen The behavior you were seeing was definitely a bug; it was never intended to be the case that an environment variable would override the `PIO_VERSION` setting (and in fact, what you saw was entirely dependent on the arbitrary order in which files are searched). I don't really think that it's necessary to introduce an option to get the previous behavior back.

That said, this change goes farther than simply fixing the bug, by not searching environment variables at all for these variable formats, so I'm fine with waiting until @jedwards4b is back.

douglasjacobsen:
@quantheory OK, thanks for the clarification. I'll just make sure he doesn't have any changes and then merge it, but the result does fix my issue, which is great.

jedwards4b:
I know that early in the development process this was needed to get the CIMEROOT and/or something else I've forgotten and I am afraid that there may be some other unintended consequence of this change.  However I think that we should go ahead with this change and address the consequence when and if it appears.

PR #343, opened 2016-08-05
Wilke/scripts/xmlchange

### LINKED ISSUE #323: xmlchange doesn't report bad syntax
On edison, I tried to do: `./xmlchange SAVE_TIMING TRUE` and it returned quietly with no error message as if it was successful.  But this syntax does nothing.  The right syntax is: `./xmlchange SAVE_TIMING=TRUE`

wilke:
Limiting the number of positional command  line arguments to 1 or 0. 
Implicit provided argument is in the form of key=value. 

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes  #323

User interface changes?: 

Code review:

### DIFF
--- a/scripts/Tools/xmlchange
+++ b/scripts/Tools/xmlchange
@@ -78,10 +78,9 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
 
     CIME.utils.handle_standard_logging_options(args)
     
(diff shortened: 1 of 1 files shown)


rljacob:
This was reverted because of #364 and #359.  Replaced by #370

Issue #314, opened 2016-08-01
Baseline comparisons incorrectly give FAIL result when baseline directory is missing


billsacks:
First: Note that testing to demonstrate this issue was done with cime5.0.2, which I know isn't the latest. Apologies if it has been fixed since then.

When I run a test suite with baseline comparisons, the baseline comparison result is given as a BFAIL if the baseline directory exists but there are no files in it, but the result is given as a FAIL if the baseline directory does not exist at all. The latter should generate a BFAIL rather than a fail.

To be more specific: I compared against a baseline tag named 'baseline_test'. I then ran two tests with a compare tag of 'baseline_test':

SMS_D_Ly1.f09_g16_gl5.TG.yellowstone_intel
SMS_D_Ly1.f09_g16_gl20.TGIS2.yellowstone_intel

I setup the 'baseline_test' directory with an empty directory corresponding to one of these tests:

```
[yslogin4:/glade/scratch/sacks/cesm_baselines/baseline_test]$ ls
SMS_D_Ly1.f09_g16_gl20.TGIS2.yellowstone_intel
```

Then I got the following TestStatus results:

```
PASS SMS_D_Ly1.f09_g16_gl20.TGIS2.yellowstone_intel INIT
PASS SMS_D_Ly1.f09_g16_gl20.TGIS2.yellowstone_intel CREATE_NEWCASE
PASS SMS_D_Ly1.f09_g16_gl20.TGIS2.yellowstone_intel XML
PASS SMS_D_Ly1.f09_g16_gl20.TGIS2.yellowstone_intel SETUP
NLFAIL SMS_D_Ly1.f09_g16_gl20.TGIS2.yellowstone_intel nlcomp
PASS SMS_D_Ly1.f09_g16_gl20.TGIS2.yellowstone_intel SHAREDLIB_BUILD
PASS SMS_D_Ly1.f09_g16_gl20.TGIS2.yellowstone_intel MODEL_BUILD
PASS SMS_D_Ly1.f09_g16_gl20.TGIS2.yellowstone_intel RUN
COMMENT: insuffiencient data for memleak test
COMMENT for SMS_D_Ly1.f09_g16_gl20.TGIS2.yellowstone_intel COMPCOMPAREbaseline history file does not exist : cism.h.nc : baseline compare baseline cism.h base with baseline_test
[...]

PR #350, opened 2016-08-07
Large refactor: properly encapsulate TestStatus handling

### LINKED ISSUE #330: Want more easily parseable output in TestStatus and cs.status
I have noted some of this to @jedwards4b and others already, and it sounds like this is something that people are thinking about, but I wanted to open an issue to track this and to record some specific thoughts.

In cime4 (thanks to @mvertens, if I remember correctly), test status output in both TestStatus and from running cimeteststatus had the very helpful property that a single "FAIL" line gave you all of the information you needed to know about the failure on that one line. For example, the line comparing the .base and .rest files for an ERS test looked like this:

```
PASS ERS_Ld7.f10_f10.ICLM45.yellowstone_gnu.cpl.hi.nc : test compare cpl.hi (.base and .rest files)
```

and comparisons with baselines gave a lot of information, including the baseline tag being compared against... here's an example from a different test:

```
FAIL ERS_D_Ly3.f09_g16_gl5.TG.yellowstone_gnu.cism-noevolve.cism.h.nc : baseline compare cism.h (baseline: compare .base file with develop_update_2016_07_n06_cism2_1_18_alpha07d file)
FAIL ERS_D_Ly3.f09_g16_gl5.TG.yellowstone_gnu.cism-noevolve.cpl.hi.nc : baseline compare cpl.hi (baseline: compare .base file with develop_update_2016_07_n06_cism2_1_18_alpha07d file)
FAIL ERS_D_Ly3.f09_g16_gl5.TG.yellowstone_gnu.cism-noevolve : baseline compare summary (baseline: compare .base file with develop_update_2016_07_n06_cism2_1_18_alpha07d file)
```

This was great, because you could use grep or similar tools to filter out the PASSes and COMMENTs, and then could see at a glance all of the details of what failed in your test suite.

This is no longer the case in cime5.

The TestStatus file gives _almost_ enough information to do this kind of parsing:

```
COMMENT for ERS_Ld7.f10_f10.ICLM45.yellowstone_gnu COMPCOMPARE: cpl.hi.nc : test compare functionality cpl.hi (.base and .rest files)
PASS ERS_Ld7.f10_f10.ICLM45.yellowstone_gnu compare
```

However, once you filter out the COMMENT lines (again, using grep or some similar tool), notice that the PASS / FAIL lines for the comparison of .base and .rest files are identical to the PASS / FAIL lines for the baseline comparison:

```
COMMENT for ERS_Ld7.f10_f10.ICLM45.yellowstone_gnu COMPCOMPARE : cpl.hi.nc : baseline compare baseline cpl.hi base with cesm1_5_beta07
PASS ERS_Ld7.f10_f10.ICLM45.yellowstone_gnu compare
```

Furthermore, if I remember correctly, a failure in one of the comparisons generates a single 'FAIL... compare' line, which does not indicate which component's history files failed the comparison. The tag being compared against is also missing for the baseline comparison line. These two "furthermore" features are not critical, but would be nice to have back.

__Perhaps more importantly: The output from cs.status further abbreviates the test results, making it even harder to filter this output to see at a glance what has failed.__

So I'd like to request that both the TestStatus and cs.status output be beefed up to be more similar to what we had in cime4.

jgfouca:
Changes to TestStatus handling:
1) Remove as much as possible TestStatus handling from test_scheduler. Users can run phases by hand (case.build) and therefore it's better for the individual implementations of the phase to update TestStatus.
2) Make a new class TestStatus to encapsulate reading/writing of TestStatus and to provide the valid phases and phase statuses.
3) It is now required that each phase be completely unique. COMPARE phases are decorated with the thing that's being compared (e.g. COMPARE_baseline) to make it unique.
4) The COMMENT system is gone, instead, any tokens that come after the core (STATUS TEST PHASE) are assumed to be comments.

Changes to SystemTests:
1) The base class provides a common framework for the build and run phases. Sub classes may override build_phase or run_phase to define how their phases behave, but they may NOT override the framework (SystemTests.build, SystemTest.run). The idea is to ensure that proper exception handling and TestStatus management occurs no matter what.
2) SystemTestCommon now handles TestStatus throught the TestStatus class

Other:
1) In order to take advantage of automatic code-checking and unit-test discovery, move implementations of all key python tools into utils/python as libraries.
1.a) jenkins_generic_job.py is moved to utils/python because it's an ACME-specific thing, not part of CIME.
1.b) Everything else was moved to utils/python/CIME

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #330 , #314

User interface changes?: Most important are changes in how to define a system_test. @billsacks please note this for your lii.py test.


utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -8,6 +8,7 @@
 from CIME.case_setup import case_setup
 from CIME.case_run import case_run
 from CIME.case_st_archive import case_st_archive
+from CIME.test_status import *
```
jgfouca: Changes to this file are worth taking a close look at.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -48,70 +51,142 @@ def __init__(self, case, expected=None):
         self._case.set_value("TEST",True)
         self._case.flush()
 
-    def fail_test(self):
-        self._runstatus = "FAIL"
+    def build(self, sharedlib_only=False, model_only=False):
+        """
+        Do NOT override this method, this method is the framework that
```
jgfouca: build method is now a framework for the build phase that all subclasses must follow.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -48,70 +51,142 @@ def __init__(self, case, expected=None):
...
+        Perform an individual build
+        """
+        return build.case_build(self._caseroot, case=self._case,
+                                sharedlib_only=sharedlib_only, model_only=model_only)
 
     def clean_build(self, comps=None):
         build.clean(self._case, cleanlist=comps)
 
     def run(self):
-        return self._run()
```
jgfouca: Same idea for run as was implemented for build.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -48,70 +51,142 @@ def __init__(self, case, expected=None):
...
+                success = True
+
+            if success:
+                if self._case.get_value("GENERATE_BASELINE"):
+                    self._generate_baseline()
+
+                if self._case.get_value("COMPARE_BASELINE"):
+                    self._compare_baseline()
+
+                self._check_for_memleak()
```
jgfouca: Checking for memleak is now part of the core run framework and will always happen if the run_phase succeeds. It used to be in test.report.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -125,8 +200,10 @@ def _coupler_log_indicates_run_complete(self, coupler_log_path):
         return False
 
     def report(self):
-        newestcpllogfile = self._get_latest_cpl_log()
-        self._check_for_memleak(newestcpllogfile)
+        """
+        Please explain what kind of things happen in report
```
jgfouca: Nothing's happening in report

utils/python/CIME/case_setup.py
```diff
@@ -288,3 +288,17 @@ def case_setup(case, clean=False, test_mode=False, reset=False):
             f.write(env_module.list_modules())
         run_cmd_no_fail("echo -e '\n' >> software_environment.txt && \
                          env >> software_environment.txt")
+
+###############################################################################
+def case_setup(case, clean=False, test_mode=False, reset=False):
+###############################################################################
+    caseroot, casebaseid = case.get_value("CASEROOT"), case.get_value("CASEBASEID")
+    test_name = casebaseid if casebaseid is not None else case.get_value("CASE")
+    with TestStatus(test_dir=caseroot, test_name=test_name) as ts:
```
jgfouca: case_setup now handles the TestStatus update for the phase it represents.

utils/python/CIME/case_test.py
```diff
@@ -12,20 +13,7 @@ def case_test(case, testname=None):
     expect(testname is not None, "testname argument not resolved")
     logging.warn("Running test for %s" % testname)
 
-    try:
-        test = find_system_test(testname, case)(case)
-        success = test.run()
-
-        test.report()
```
jgfouca: baseline actions and reporting are now part of the core run framework in system_test_common.

utils/python/CIME/test_scheduler.py
```diff
@@ -731,14 +678,21 @@ def _consumer(self, test, test_phase, phase_method):
...
             self._update_test_status(test, test_phase, status)
-        self._handle_test_status_file(test, test_phase, success)
 
         status_str = "Finished %s for test %s in %f seconds (%s)" %\
                      (test_phase, test, elapsed_time, status)
         if not success:
             status_str += "    Case dir: %s" % self._get_test_dir(test)
         logger.info(status_str)
 
+        if test_phase in [CREATE_NEWCASE_PHASE, XML_PHASE, NAMELIST_PHASE]:
```
jgfouca: Unfortunately, test_scheduler still has to do some TestStatus managment.

utils/python/CIME/test_status.py
```diff
@@ -0,0 +1,227 @@
+"""
+Functions for managing the TestStatus file
+"""
+
+from CIME.XML.standard_module_setup import *
```
jgfouca: New class, please review closely.

utils/python/tests/scripts_regression_tests.py
```diff
@@ -34,32 +35,6 @@
 class A_RunUnitTests(unittest.TestCase):
 ###############################################################################
 
-    def do_unit_tests(self, script,from_dir=SCRIPT_DIR):
```
jgfouca: Just rely on unit test discovery.

utils/python/tests/scripts_regression_tests.py
```diff
@@ -68,14 +43,12 @@ def test_resolve_variable_name(self):
 
     def test_CIME_unit_tests(self):
         # Find and run all the unit tests in the CIME directory tree
-        os.environ["PYTHONPATH"] = LIB_DIR
```
jgfouca: This env chg had global impacts (caused pylint tests to fail on some systems). Not safe.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -48,70 +51,142 @@ def __init__(self, case, expected=None):
...
+                self._test_status.set_status(RUN_PHASE, TEST_PENDING_STATUS)
 
-            if success and self._coupler_log_indicates_run_complete(coupler_log_path):
-                self.pass_test()
-            else:
-                success = False
-                self.fail_test()
+            success = self.run_phase()
+            if not isinstance(success, bool):
+                logger.warning("build_phase did not return a bool, assuming PASS!")
```
billsacks: 'build_phase' should say 'run_phase' here.

It seems better to me to assume a FAIL if the return value doesn't match expectation; is there a reason you're assuming PASS here? (Is this perhaps to handle tests that don't return anything, so you get a None? If so, it would help to document that reason and maybe handle it specially... I still feel that, in general, it's best to assume a FAIL rather than PASS if you see something unexpected.)

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -48,70 +51,142 @@ def __init__(self, case, expected=None):
...
-                f.write(teststatusfile)
-                f.write("COMMENT TIME %d\n" % total_time)
+    def run_phase(self):
+        """
+        This is the default run phase implementation, it just does an individual run.
+        This is the subclass' extension point if they need to define a custom run phase.
+
+        PLEASE THROW AN EXCEPTION OR RETURN FALSE ON FAIL
+        """
+        return self.run_indv()
```
billsacks: I'm finding it a bit hard to understand the flow through run / run_phase / run_indv, because of the possibility that a failed run can _either_ throw an exception or return false on fail. Since run_indv appears to always throw an exception on failures, would it make sense to just make that the general rule? i.e., a failure is signaled by an exception; return value is irrelevant? This would make it easier to write tests that overrun run_phase (as most of them do). 

Actually, I see that at least some of the individual tests don't check the return value from run_indv, nor do they return a value themselves in their run_phase. So if run_indv were ever changed to return a False value rather than throwing an exception (which appears to be a possibility, given the way run_phase is written in this base class), then that return value wouldn't be handled properly by the individual tests. Again, I think the logic would be more clear and robust here if we had the rule that failures lead to exceptions, return value is irrelevant.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -136,8 +213,8 @@ def _component_compare_move(self, suffix):
         if rc == 0:
             append_status(out, sfile="TestStatus.log")
         else:
-            append_status("Component_compare_test.sh failed out: %s\n\nerr: %s\n"%(out,err)
-                          ,sfile="TestStatus.log")
+            append_status("Component_compare_test.sh failed out: %s\n\nerr: %s\n" % (out, err),
```
billsacks: 'component_compare_test' should say 'component_compare_move'

billsacks:
Thanks for your work on this!

Can you give an example of what the new TestStatus file looks like? I'm especially wondering what this means:

> 4) The COMMENT system is gone, instead, any tokens that come after the core (STATUS TEST PHASE) are assumed to be comments.

I typically use something akin to 'grep -v' to filter out COMMENT lines, and am trying to figure out if something like that will still be possible. To say that another way: I feel like each line should have some key token to facilitate filtering.

billsacks:
I have started looking at this, but want to do some more looking tomorrow. As you can see from my line comments, my main confusion is about how failures are signaled and handled now. I'm largely thinking about this in terms of what I'll need to do to my new general infrastructure for two-part tests.

Currently, this infrastructure basically looks like this (some irrelevant details excluded):

```python
        success = self._run(self._run_one_suffix)
        if success:
            self._status_run1 = "PASS"
        else:
            self._status_run1 = "FAIL"
            return False

        success = self._run(self._run_two_suffix)
        if success:
            self._status_run2 = "PASS"
        else:
            self._status_run2 = "FAIL"
            return False

        success = self._component_compare_test(self._run_one_suffix, self._run_two_suffix)
        if success:
            self._status_compare = "PASS"
        else:
            self._status_compare = "FAIL"
            return False

        return success
```

I'm thinking that the two run calls will now need to look like:

```python
    try:
        # This assumes that run_indv raises an exception on failure. More code
        # is needed if it's also possible for run_indv to return False on failure.
        self.run_indv()
    except:
        self._status_run1 = "FAIL"
        raise
    else:
        self._status_run1 = "PASS"
```

(where all of the exception handling is only needed in order to record the status of the individual runs separately). This would get more complicated if run_indv() can return a False value to indicate failure, in addition to the possibility that it raises an exception to indicate failure (part of my motivation for some questions in the line notes).

Then, I think the logic for component_compare_test can remain as is.

Can you please let me know if you agree with this assessment?

billsacks:
I've finished looking through the pieces of this that are relevant for what I'm working on – mainly related to how to define a new system test and the interaction of individual tests with SystemTestsCommon. Overall, this looks good; I'd just like to get my earlier comments resolved.

utils/python/CIME/SystemTests/system_tests_common.py
jgfouca: @billsacks : I wanted to give the test developer a choice of returning False or throwing an exception. Currently, I changed every test to just rely on exceptions since it's a little simpler. The vision of run_indv is that users would use try/except if they wanted to handle the error instead of letting it escape. That said, if you think that this is confusing, it would be easy to change.

utils/python/CIME/SystemTests/system_tests_common.py
jgfouca: Fixed.

utils/python/CIME/SystemTests/system_tests_common.py
billsacks: @jgfouca : I see your point about providing flexibility to the test developer. But I think that, if a test developer wants to handle the error themselves, they still could: They just would have to re-raise the exception after handling it, or raise some new exception (for example, see my proposed changes to the common two-run test infrastructure, in a separate comment).

So I think it's more straightforward to provide a single mechanism for signaling failure here, both from the test developer's perspective (putting myself in the shoes of the test developer, I'd be left wondering, "which should I do?"), and from a perspective of code robustness (or at the very least, we'd want some additional unit tests ensuring that things are handled correctly if run_indv / run_phase returns False without raising an exception).

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -146,10 +223,11 @@ def _component_compare_test(self, suffix1, suffix2):
...
-                          sfile="TestStatus")
-        else:
+        status = TEST_PASS_STATUS if rc == 0 else TEST_FAIL_STATUS
+        with self._test_status:
+            self._test_status.set_status("%s_%s_%s" % (COMPARE_PHASE, suffix1, suffix2), status)
+
+        if rc != 0:
             append_status("Component_compare_test.sh failed out: %s\n\nerr: %s\n"%(out,err),
                           sfile="TestStatus.log")
             return False
```
jgfouca: That's a great point. I didn't notice this because the TestStatus gets set correctly. I will change it so that it only returns success if there are no FAILs in the TestStatus file.

jgfouca:
@billsacks instead of grepping for comment, you'd just grep for the phase name that you were interested in.

jgfouca:
@billsacks Each test only has one RUN status, so you don't need to catch any exceptions.

utils/python/CIME/SystemTests/system_tests_common.py
jgfouca: Good point, will check for None explicitly.

utils/python/CIME/SystemTests/system_tests_common.py
jgfouca: Ok, we'll just go with exceptions then.

PR #357, opened 2016-08-09
Fixpionml

ekluzek:
Fix so driver_cpl modelio namelist definition is separate from

Test namelist changes: Gets pnetcdf into base pio_typename in drv_in namelist by
keeping modelio namelist_definition seperate from everything else.

Also get driver_cpl/bld/testdir standalone namelist test working with cime5

Also fix issue where drv_flds_in wasn't being copied over as should be in buildnml.

Fixes: CESM-Development/cime#473

User interface changes?: no

Code review: @jedwards4b


ekluzek:
The exciting ending to the sentence is....
 
Fix so driver_cpl modelio namelist definition is separate from the other drv namelist definitions (drv_in and drv_flds_in).

PR #363, opened 2016-08-09
Fix several issues in acme machines

### LINKED ISSUE #354: case.submit broken on edison with CIME_MODEL=acme
case.submit is broken on edison with CIME_MODEL=acme.

```
Submitting job script sbatch   case.run 
ERROR: Command: 'sbatch   case.run ' failed with error 'sbatch: option requires an argument -- 'A'
sbatch: error: Batch job submission failed: Invalid user id'
```

Works fine with CIME_MODEL=cesm.

rljacob:
Fix issues found in acme machines files while testing the new CIME with ACME.

blues:  increase default walltime from 15 mins to 1 hour
edison:  remove -A directive;  define SAVE_TIMING_DIR

Tested with an X-case on blues and edison.
Test namelist changes:  none
Test status: bit for bit

Fixes #354

### DIFF
--- a/cime_config/acme/machines/config_batch.xml
+++ b/cime_config/acme/machines/config_batch.xml
@@ -145,7 +145,7 @@
 	<queue>shared</queue>
       </queues>
       <walltimes>
-	<walltime default="true">00:15:00</walltime>
+	<walltime default="true">01:00:00</walltime>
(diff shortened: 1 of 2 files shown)


jgfouca:
Looks good.

PR #373, opened 2016-08-10
cime_bisect: Add better support for modifying create_test cmd

### LINKED ISSUE #360: cime_bisect suggestions
cime_bisect is a great tool but it could use more functionality.  

1. Be able to pass options through to create_test
2. Be able to specify the failing phase so that if for example there was a build phase failure 
     there is no need to do the run phase.

jgfouca:
Test suite: tested by hand
Test baseline: 
Test namelist changes: 

Fixes #360

User interface changes?: Minor changes to cime_bisect

Code review: @jedwards4b


jedwards4b:
Looks good, thanks

PR #377, opened 2016-08-11
Data model build-namelist refactoring

quantheory:
This commit encompasses two major changes to the data model `build-namelist` scripts:

1. The Perl `build-namelist` is refactored by adding `perl5lib/Streams/BuildNamelistUtils.pm`. As part of this change, the namelist files are consolidated so that each data component only has a single namelist file for its options (plus the `modelio` file).

2. The main `build-namelist` functionality is translated from Perl to Python. As an example, the datm `build-namelist` has been translated to Python. Since we have a more full-featured Python namelist parser, `compare_namelists.py` has also been refactored as part of this change.

Additionally, some of the remaining ESMF interfaces are removed.

Some minor changes that were helpful for the above:

- `Case` objects now have a `get_case_root` method.
- `scripts_regression_tests` now have a `--machine` option so that you can ask the tests to "pretend" that they are running on a different platform (useful mainly for unit tests and other simple tests).

This commit passes `scripts_regression_tests`.

This PR makes progress on #221 (but does not fix it until the other components have `build-namelist` rewritten, as datm has here).
Connects to #221


rljacob:
Can you please do the esmf interface removal in a separate PR?

components/data_comps/datm/bld/namelist_files/namelist_definition_datm.xml
```diff
@@ -4,32 +4,93 @@
...
+List of streams used for the given datm_mode.
 </entry>
 
-<entry id="DATM_PRESAERO"
-type="char*256"
-category="datm_setting"
-group="datm_env_settings"
-valid_values="none,pt1_pt1,clim_2000,clim_1850,trans_1850-2000,rcp2.6,rcp4.5,rcp6.0,rcp8.5">
-Prescribed aerosol mode
+<entry id="strm_domdir"
```
jedwards4b: I have a philosophical issue with the format of this xml file.  The issue is in the distinction between attributes and entries:  this file incorrectly handles everything as attributes.   It would be more correct and consistent with other xml in cime to have:
```
    <entry id="strm_domdir">
        <type>char*4096</type>
        <category>streams</category>
        <group>streams_file</group>
        <desc>Stream domain file directory</desc>
    </entry>

File: utils/python/CIME/XML/namelist_defaults.py [hunk 2]
```diff
@@ -0,0 +1,113 @@
+"""Interface to `namelist_defaults.xml`.
```
jedwards4b: Following up on my previous comment, this file should inherit from EntryID and the xml should follow the entryid format.

utils/python/CIME/XML/namelist_definition.py
```diff
@@ -0,0 +1,309 @@
...
+# pylint:disable=wildcard-import,unused-wildcard-import
+
+import re
+
+from CIME.namelist import fortran_namelist_base_value, \
+    is_valid_fortran_namelist_literal, character_literal_to_string, \
+    expand_literal_list, Namelist
+
+from CIME.XML.standard_module_setup import *
+from CIME.XML.generic_xml import GenericXML
```
jedwards4b: Should follow the entryid standard

utils/python/CIME/namelist.py
```diff
@@ -0,0 +1,1903 @@
...
+    except (_NamelistEOF, _NamelistParseError) as error:
+        # Deal with unexpected EOF or other parsing errors.
+        expect(False, str(error))
+    if groupless:
+        return namelist_dict
+    else:
+        return Namelist(namelist_dict)
+
+
+class Namelist(object):
```
jedwards4b: I would prefer to have only one class per file.

utils/python/CIME/namelist.py
```diff
@@ -0,0 +1,1903 @@
...
+                        lines[-1] += ",\n"
+                        lines.append("      " + value)
+                lines[-1] += "\n"
+                for line in lines:
+                    out_file.write(line)
+            if format_ == 'nml':
+                out_file.write("/\n")
+
+
+class _NamelistEOF(Exception):
```
jedwards4b: new file

jedwards4b:
There are several Classes in each file in the new python code - I would like to suggest instead a Namelists directory under CIME with each class in a separate file.

components/data_comps/datm/bld/namelist_files/namelist_definition_datm.xml
mvertens: The xml schema was purposely not addressed in this implementation. I agree
with your suggestion - but I think an even more general schema that
combines both the namelist_definition.xml and namelist_defaults.xml file
needs to be created. That will be done in the next phase the project.

utils/python/CIME/XML/namelist_defaults.py
mvertens: I think what you are suggesting is outside the scope of what we had asked
Sean to do. No xml schema changes were allowed in this implementation.

jedwards4b: Okay I see where you are coming from but it seems like a lot of python code is being written to support this format that we plan to get rid of and that a lot of reusable code in entryid is not being taken advantage of.

quantheory:
@rljacob OK.

@jedwards4b Every class except `Namelist` is intended to be private (hence the leading underscores). I'm OK with splitting the parser into another file, but that file wouldn't have anything public except the `parse` function in it.

utils/python/CIME/XML/namelist_defaults.py
quantheory: I don't think that `entry_id` supports any kind of ranking system for selecting among multiple matches (?), so the search code here is not really duplicating any existing `entry_id` functionality. Even after this is redesigned, we still need some way for specific matches to override non-specific matches, which will be doing something similar to this module.

components/data_comps/datm/bld/namelist_files/namelist_definition_datm.xml
quantheory: @jedwards4b The `namelist_definition` code wouldn't really need to change all that much for what you're suggesting. I think that pretty much all of the differences would be in lines 92-122 of the file. However, changing the XML files themselves would take a bit longer; you'd presumably want a conversion utility for that.

jedwards4b:
I guess same file is okay if they are private.

jedwards4b:
I think this branch is ready to merge - any objections?

mvertens:
I'd like for Sean to document what tests were run as part of making this
pull request. In particular - where any of the aux_clm45 tests run for CLM?

quantheory:
It passes `scripts_regression_tests` on yellowstone, and I also ran namelist compares for all of the CLM tests. I did not run the full `aux_clm45`, but since this is all about namelist generation, I was hoping that the namelist compares would pick any differences up. The only `NLFAIL` I saw were due to differences in trailing whitespace in the `streams` variable.

quantheory:
I should clarify that I compared namelists between the version with my changes, and one with just Mariana's changes. There's no easy way for me to compare to master, other than to run the full `aux_clm45` tests.

mvertens:
Great. I'd feel more comfortable with just running the aux_clm45 tests with
your branch and then I'd okay the PR. I'll take that on.

quantheory:
I just did a merge to deal with some conflicts, but we probably should re-test this. Please hold off on any cleanup of namelist-related scripts until this is merged; most of the merge conflicts were due to cleanup of code that this commit deletes anyway.

jedwards4b:
Superseded by PR#640

PR #391, opened 2016-08-12
Report an incomplete test as a failure

### LINKED ISSUE #390: Errors in system test's __init__ method not reported correctly
I noticed that, if I raise an exception in the __init__ method of a system test, the test gets reported as a PASS.

Example:

```diff
--- a/utils/python/CIME/SystemTests/sms.py
+++ b/utils/python/CIME/SystemTests/sms.py
@@ -15,3 +15,4 @@ class SMS(SystemTestsCommon):
         initialize an object interface to the SMS system test
         """
         SystemTestsCommon.__init__(self, case)
+        raise RuntimeError
```

Then

`./create_test --test-root . SMS.f10_f10.ICLM45.yellowstone_gnu`

Results in:

```
Finished SHAREDLIB_BUILD for test SMS.f10_f10.ICLM45.yellowstone_gnu in 22.090373 seconds (FAIL)    Case dir: /glade/scratch/sacks/cesm_code/cesm2_0_alpha01_v5/cime/scripts/SMS.f10_f10.ICLM45.yellowstone_gnu.20160811_214325
FAIL SMS.f10_f10.ICLM45.yellowstone_gnu (phase SHAREDLIB_BUILD)
    Case dir: /glade/scratch/sacks/cesm_code/cesm2_0_alpha01_v5/cime/scripts/SMS.f10_f10.ICLM45.yellowstone_gnu.20160811_214325
test-scheduler took 45.6888129711 seconds
```

But TestStatus has:

```
PASS SMS.f10_f10.ICLM45.yellowstone_gnu CREATE_NEWCASE
PASS SMS.f10_f10.ICLM45.yellowstone_gnu XML
PASS SMS.f10_f10.ICLM45.yellowstone_gnu SETUP
```

and running cs.status gives:

```
SMS.f10_f10.ICLM45.yellowstone_gnu (Overall: PASS), details:
  PASS CREATE_NEWCASE
  PASS XML
  PASS SETUP
```

I am particularly thinking about this with respect to the clone-based test infrastructure: I have set this up to do all test-specific setup from the constructor. If there's no way to address this issue (and thus errors in the constructor will lead to tests looking like they have passed), then I should probably move this test-specific setup into the build_phase method instead. But it seems like it's worth addressing this issue for more general reasons, if possible.

billsacks:
This is a proposed solution to issue #390 - reporting tests as
INCOMPLETE if they don't have a status for the RUN_PHASE. This does all
of the handing of this condition in the TestStatus class itself. The key
assumption is that RUN_PHASE is a necessary phase for any test to be
considered complete.

This addresses the issue of failures in test initialization. In
addition, and perhaps more importantly, it also correctly reports tests
that have been created but either haven't been built or haven't been
submitted. This could be an issue, for example, if a test is created
with --no-build, and then the user forgets to either build or submit the
test: Previously, the test would be reported as a PASS by cs.status, and
there would be no FAIL lines in the TestStatus file.

I'm unclear on what other tools interact with TestStatus / cs.status, so
I'm not sure if other things would need to be changed for this. Note,
though, that I have maintained the PEND overall status if
wait_for_tests=True in the call to get_overall_test_status, because it
appears that that is important. (Though as I note in the comment, maybe
eventually we could have callers respond to the TEST_INCOMPLETE status
and remove the wait_for_tests argument.)

I have tested this with a few manual tests: I have verified that passing
tests are reported correctly after they are complete (I ran one test
with --no-build and build & submitted manually, and another test without
--no-build).

In addition, I verified that, with this diff:

```diff
--- a/utils/python/CIME/SystemTests/sms.py
+++ b/utils/python/CIME/SystemTests/sms.py
@@ -15,3 +15,4 @@ class SMS(SystemTestsCommon):
         initialize an object interface to the SMS system test
         """
         SystemTestsCommon.__init__(self, case)
+        raise RuntimeError
```

This test: `SMS.f10_f10.ICLM45.yellowstone_gnu` gives the following
TestStatus file:

```
PASS SMS.f10_f10.ICLM45.yellowstone_gnu CREATE_NEWCASE
PASS SMS.f10_f10.ICLM45.yellowstone_gnu XML
PASS SMS.f10_f10.ICLM45.yellowstone_gnu SETUP
FAIL SMS.f10_f10.ICLM45.yellowstone_gnu TEST_COMPLETE
```

and this cs.status output:

```
SMS.f10_f10.ICLM45.yellowstone_gnu (Overall: INCOMPLETE), details:
  PASS CREATE_NEWCASE
  PASS XML
  PASS SETUP
```

An ERS test created with --no-build had essentially the same output
after the test was created but before running ./case.build, as desired.

Note that, for an incomplete test, cs_status reports an overall
INCOMPLETE status, but does not output the FAIL status for
TEST_COMPLETE. If we wanted it to output that, I think we could add a
line at the end of TestStatus.\_\_iter\_\_ like:

    if not self._has_run_phase():
        yield INCOMPLETE_PHASE, TEST_INCOMPLETE_STATUS

Test suite: Only ran: python -m doctest -v CIME/test_status.py
   Can run full test suite if others like this idea
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

Fixes #390 

User interface changes?: none

Code review: none yet

Again: More testing is needed on this before it should come to master, but I wanted to put it out there for consideration.

### DIFF
--- a/utils/python/CIME/test_status.py
+++ b/utils/python/CIME/test_status.py
@@ -13,6 +13,11 @@ TEST_PENDING_STATUS  = "PEND"
 TEST_PASS_STATUS     = "PASS"
 TEST_FAIL_STATUS     = "FAIL"
 
+# An extra status possible for the overall test result
+TEST_INCOMPLETE_STATUS = "INCOMPLETE"
+
+# TEST_INCOMPLETE_STATUS is not in the ALL_PHASE_STATUSES list because it is not
+# a valid status for an individual phase
 ALL_PHASE_STATUSES = [TEST_PENDING_STATUS, TEST_PASS_STATUS, TEST_FAIL_STATUS]
 
 # Special statuses that the overall test can be in
@@ -34,6 +39,10 @@ MEMLEAK_PHASE         = "MEMLEAK"
 COMPARE_PHASE         = "COMPARE" # This is one special, real phase will be COMPARE_$WHAT
 GENERATE_PHASE        = "GENERATE"
 
+# A pseudo-phase used to indicate if a test is incomplete
+INCOMPLETE_PHASE      = "TEST_COMPLETE"
+
+# INCOMPLETE_PHASE is not in the ALL_PHASES list because it is handled specially
 ALL_PHASES = [INITIAL_PHASE,
               CREATE_NEWCASE_PHASE,
               XML_PHASE,
@@ -125,6 +134,9 @@ class TestStatus(object):
                     else:
                         fd.write("%s %s %s %s\n" % (status, self._test_name, phase, comments))
 
+                if not self._has_run_phase():
+                    fd.write("%s %s %s\n" % (TEST_FAIL_STATUS, self._test_name, INCOMPLETE_PHASE))
+
     def _parse_test_status(self, file_contents):
         """
         >>> contents = '''
@@ -133,6 +145,7 @@ class TestStatus(object):
         ... FAIL ERS.foo.A SETUP
         ... PASS ERS.foo.A COMPARE_baseline
         ... PASS ERS.foo.A SHAREDLIB_BUILD Time=42
+        ... FAIL ERS.foo.A TEST_COMPLETE
         ... '''
         >>> _test_helper1(contents)
         OrderedDict([('CREATE_NEWCASE', ('PASS', '')), ('XML', ('PASS', '')), ('SETUP', ('FAIL', '')), ('COMPARE_baseline', ('PASS', '')), ('SHAREDLIB_BUILD', ('PASS', 'Time=42'))])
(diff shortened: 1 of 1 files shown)


utils/python/CIME/test_status.py
```diff
@@ -205,11 +225,8 @@ def get_overall_test_status(self, wait_for_run=False, check_throughput=False, ch
         'PASS'
         """
         rv = TEST_PASS_STATUS
-        run_phase_found = False
         for phase, data in self._phase_statuses.iteritems():
             status = data[0]
-            if phase == RUN_PHASE:
```
billsacks: Since this "run_phase_found" logic is now needed in two places, I refactored it to a _has_run_phase method

jgfouca:
@billsacks I think I have any idea. I'll submit a PR to see what you think.

jgfouca:
@billsacks take a look at #392. I prefer that approach over the one in this PR which increases the complexity of the system.

gold2718:
No objections.

jgfouca:
Closed, replaced with #399

PR #392, opened 2016-08-12
Exceptions in SystemTest constructors should leave TestStatus in decent state

### LINKED ISSUE #390: Errors in system test's __init__ method not reported correctly

jgfouca:
Also,
1) Added regression tests to test the above
2) case_setup should not put stuff in TestStatus unless it's a test case
3) Rename some acme_* things to cime_*

Test suite: scripts_regresssion_test
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #390

Code review: @billsacks @jedwards4b

### DIFF
--- a/utils/python/tests/scripts_regression_tests.py
+++ b/utils/python/tests/scripts_regression_tests.py
@@ -425,7 +425,7 @@ class C_TestCreateTest(TestCreateTestCommon):
     ###########################################################################
         if NO_BATCH:
             extra_args += " --no-batch"
-        cmd = "%s/create_test acme_test_only_pass %s" % (SCRIPT_DIR, extra_args)
+        cmd = "%s/create_test cime_test_only_pass %s" % (SCRIPT_DIR, extra_args)
         stat, output, errput = run_cmd(cmd)
         if (expect_works):
             self.assertEqual(stat, 0, msg="COMMAND '%s' SHOULD HAVE WORKED\ncreate_test output:\n%s\n\nerrput:\n%s\n\ncode: %d" % (cmd, output, errput, stat))
@@ -479,8 +479,11 @@ class E_TestTestScheduler(TestCreateTestCommon):
     def test_a_phases(self):
     ###########################################################################
         # exclude the MEMLEAK tests here.
-        tests = update_acme_tests.get_full_test_names(["acme_test_only",
-                                                       "^TESTMEMLEAKFAIL_Mmpi-serial.f19_g16.X", "^TESTMEMLEAKPASS_Mmpi-serial.f19_g16.X"],
+        tests = update_acme_tests.get_full_test_names(["cime_test_only",
+                                                       "^TESTMEMLEAKFAIL_Mmpi-serial.f19_g16.X",
+                                                       "^TESTMEMLEAKPASS_Mmpi-serial.f19_g16.X",
+                                                       "^TESTBUILDFAILEXC.f19_g16_rx1.A",
+                                                       "^TESTRUNFAILEXC_Mmpi-serial.f19_g16_rx1.A"],
                                                       self._machine, self._compiler)
         self.assertEqual(len(tests), 3)
         ct = TestScheduler(tests)
@@ -551,15 +554,17 @@ class E_TestTestScheduler(TestCreateTestCommon):
     ###########################################################################
     def test_b_full(self):
     ###########################################################################
-        tests = update_acme_tests.get_full_test_names(["acme_test_only"], self._machine, self._compiler)
+        tests = update_acme_tests.get_full_test_names(["cime_test_only"], self._machine, self._compiler)
         test_id="%s-%s" % (self._baseline_name, CIME.utils.get_utc_timestamp())
         ct = TestScheduler(tests, test_id=test_id, no_batch=NO_BATCH)
 
-        build_fail_test = [item for item in tests if "TESTBUILDFAIL" in item][0]
-        run_fail_test   = [item for item in tests if "TESTRUNFAIL" in item][0]
-        pass_test       = [item for item in tests if "TESTRUNPASS" in item][0]
-        mem_fail_test   = [item for item in tests if "TESTMEMLEAKFAIL" in item][0]
-        mem_pass_test   = [item for item in tests if "TESTMEMLEAKPASS" in item][0]
+        build_fail_test     = [item for item in tests if "TESTBUILDFAIL." in item][0]
+        build_fail_exc_test = [item for item in tests if "TESTBUILDFAILEXC" in item][0]
+        run_fail_test       = [item for item in tests if "TESTRUNFAIL_" in item][0]
+        run_fail_exc_test   = [item for item in tests if "TESTRUNFAILEXC" in item][0]
+        pass_test           = [item for item in tests if "TESTRUNPASS" in item][0]
+        mem_fail_test       = [item for item in tests if "TESTMEMLEAKFAIL" in item][0]
+        mem_pass_test       = [item for item in tests if "TESTMEMLEAKPASS" in item][0]
 
         log_lvl = logging.getLogger().getEffectiveLevel()
         logging.disable(logging.CRITICAL)
@@ -584,8 +589,16 @@ class E_TestTestScheduler(TestCreateTestCommon):
                 self.assertEqual(ts.get_status(CIME.test_scheduler.MODEL_BUILD_PHASE), TEST_FAIL_STATUS)
                 self.assertTrue("Intentional fail for testing infrastructure" in open(log_file, "r").read(),
                                 "Broken test did not report build error")
+            elif (test_name == build_fail_exc_test):
+                self.assertEqual(ts.get_status(CIME.test_scheduler.SHAREDLIB_BUILD_PHASE), TEST_FAIL_STATUS)
+                self.assertTrue("Exception from init" in open(log_file, "r").read(),
+                                "Broken test did not report build error")
             elif (test_name == run_fail_test):
                 self.assertEqual(ts.get_status(CIME.test_scheduler.RUN_PHASE), TEST_FAIL_STATUS)
+            elif (test_name == run_fail_exc_test):
+                self.assertEqual(ts.get_status(CIME.test_scheduler.RUN_PHASE), TEST_FAIL_STATUS)
+                self.assertTrue("Exception from run_phase" in open(log_file, "r").read(),
+                                "Broken test did not report build error")
             elif (test_name == mem_fail_test):
                 self.assertEqual(ts.get_status(MEMLEAK_PHASE), TEST_FAIL_STATUS)
                 self.assertEqual(ts.get_status(CIME.test_scheduler.RUN_PHASE), TEST_PASS_STATUS)
@@ -644,7 +657,7 @@ class TestJenkinsGenericJob(TestCreateTestCommon):
         # the testroot (bld/run dump area) and jenkins root
         if (test_id is None):
             test_id = self._baseline_name
-        num_tests_in_tiny = len(update_acme_tests.get_test_suite("acme_test_only_pass"))
+        num_tests_in_tiny = len(update_acme_tests.get_test_suite("cime_test_only_pass"))
 
         jenkins_dirs = glob.glob("%s/*%s*/" % (self._jenkins_root, test_id)) # case dirs
         # scratch_dirs = glob.glob("%s/*%s*/" % (self._testroot, test_id)) # blr/run dirs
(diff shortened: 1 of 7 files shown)


utils/python/tests/scripts_regression_tests.py
```diff
@@ -479,8 +479,11 @@ class E_TestTestScheduler(TestCreateTestCommon):
     def test_a_phases(self):
     ###########################################################################
         # exclude the MEMLEAK tests here.
-        tests = update_acme_tests.get_full_test_names(["acme_test_only",
-                                                       "^TESTMEMLEAKFAIL_Mmpi-serial.f19_g16.X", "^TESTMEMLEAKPASS_Mmpi-serial.f19_g16.X"],
+        tests = update_acme_tests.get_full_test_names(["cime_test_only",
```
jedwards4b: These test were all mpi serial before, should we be concerned about adding a parallel test?

jedwards4b:
One minor issue, otherwise I am fine with this.

utils/python/tests/scripts_regression_tests.py
jgfouca: These are negations due to the leading "^" symbol.

billsacks:
I feel like this adequately addresses the original issue raised in #390 . However, it leaves open some other scenarios where all lines that appear in TestStatus are reported as PASS - i.e., there are no lines indicating that there are any problems. My desire would be to have the following requirement:

__If there is a TestStatus file in the test directory, and if the test has not completed successfully, then there must be at least one line in the TestStatus file with a status other than PASS__

As far as I can tell, this requirement was met with cime4 and earlier, but there are some scenarios where it is not currently met. For example:

1. A test created with --no-build shows only PASS lines in its TestStatus file before it is built. I suspect the same may be true of a test created with --no-submit (before it is submitted), but haven't tested that.

2. If I kill a test during `SETUP` with ctrl-C,  TestStatus shows:

```
PASS SMS.f10_f10.ICLM45.yellowstone_intel CREATE_NEWCASE
PASS SMS.f10_f10.ICLM45.yellowstone_intel XML
```

I'm not tied to my proposed solution in #391 . I agree that had some complexity, but it isolated this complexity to the TestStatus class, rather than requiring all creators of the test object to do appropriate error-handling; I'd argue that the latter actually adds more system-level complexity, even though it involves fewer lines of code. In addition, I believe that my proposed solution met the above (possible) requirement - ensuring that there is a FAIL or INCOMPLETE status reported if the test hasn't gotten through the run phase, for any reason.

As a way to proceed, I'd propose that we have a broader discussion about the requirements of the test system and test reporting - specifically addressing whether others feel that my proposed requirement above should truly be a requirement. I'd be interested in hearing from @mvertens @rljacob @jedwards4b and @fischer-ncar , and possibly others, about whether this should be a requirement or whether I'm simply being overly paranoid.

If others agree that this should be a requirement, then again: I'm not tied to my proposal, but would be happy with any solution that addresses this requirement without adding too much extra complexity. If others feel this should not be a requirement, then I'm fine with accepting this PR.

jgfouca:
@billsacks : My opinion is that in your case 1, the TestStatus file has exactly what a user would want. It should be obvious to the user that the test was never run because the run phase isn't listed. If the user was just testing builds, they may never intend to run. 

Your case 2 opens up the much bigger issue of signal handling; I don't think it's reasonable for users to expect things to be left in a good state if they're hitting our tools with signals. I do do a little signal handling in our Jenkins scripts because we have to handle the SIGTERM that's generated when the job times out, but this is a special case.

I agree that others should weigh-in here if they care.

jgfouca:
One thing I noticed in #391, the TestStatus file will look the same if you run 'create_test X --no-build' verses if you run 'create_test X' and there's an exception in the X __init__.

billsacks:
@jgfouca : I provided those two examples because they were things that I could think of easily. They suggested to me that there might be other failure modes where only PASSes appeared in the TestStatus file, and that it might be hard to ever be confident that we have addressed all failure modes without a solution like the one in #391 . But if you feel that my examples are the only cases where we'd get all PASSes, then I'm willing to let it go. Also, point taken that my case 1 is ambiguous as to what a user may want: I was going under the assumption that the user had forgotten to manually run build & submit, which suggests that a FAIL / INCOMPLETE status would be appropriate.

In general, I'm coming at this from my typical use of the test system, which is to quickly parse the TestStatus or cs.status output with essentially a `grep -v PASS`, so I want to be confident that something will appear from that if a test has failed for any reason or hasn't been completely run.

> One thing I noticed in #391, the TestStatus file will look the same if you run 'create_test X --no-build' verses if you run 'create_test X' and there's an exception in the X init.

This is a good point, and probably undesirable. I guess that fixing this would require combining my solution with yours - or something like that. i.e., your solution would explicitly catch failures in test init, whereas mine catches anything that still falls through the cracks.

> I agree that others should weigh-in here if they care.

Yes; I should probably be quiet now and let others weigh in if they have opinions :-)

billsacks:
@apcraig : I should have added you to the list of cc's here: I feel like you have had valuable thoughts in the past about how the test system should behave, so feel free to weigh in here if you have any thoughts. Particularly see my comment above: I'm curious what you and others think about this:

> My desire would be to have the following requirement:

> If there is a TestStatus file in the test directory, and if the test has not completed successfully, then there must be at least one line in the TestStatus file with a status other than PASS

-- whether this has, indeed, been a requirement in the past, or whether I'm making up a requirement that doesn't and shouldn't really exist.

jgfouca:
@billsacks , test objects should only ever be created by case.build and case.test, so I'm pretty confident that the only way that a failure could not be captured in the TestStatus file is if our tools got hit with a signal.

Your point about cs.status is well taken though. I typically use wait_for_tests which is why my experience is different than yours. Maybe we need to make a change to cs.status to note incomplete tests?

billsacks:
> Your point about cs.status is well taken though. I typically use wait_for_tests which is why my experience is different than yours. Maybe we need to make a change to cs.status to note incomplete tests?

Yes, actually that's probably the most important piece of my proposed change.

billsacks:
@jgfouca : As I think about this more, I think that I'd be happy with your solution in this PR, together with a change to the cs.status reporting so that an incomplete test is guaranteed to result in at least one line with a non-PASS status code.

I'd still be happy to hear if others have thoughts on this, because I don't have a complete understanding of the past requirements of the test system myself.

gold2718:
I'm still playing catch-up looking through the code so I hope my $0.02 below is not too far off base. I agree with Bill that it is important for test phases (and therefore the overall test) to have an INCOMPLETE status. When a test is created, it should know all the phases that are part of the test. If the test is created with --no-submit then the test does not expect a run phase and if it is created with --no-build, it does not expect a build phase. At the end of the init, I would like to see the self.phase_statuses be filled out with each phase marked INCOMPLETE (with INCOMPLETE added to the list of valid test statuses.

Doing this would cause the TestStatus file to be atomic in that every time it is written out (flush), it is overwritten with the complete set of phases. I think this meets Bill's criterion. If we have flush create a backup before writing (move filename to filename~), it would even be fairly robust against getting killed. Hopefully, it is also not hard to implement in that init methods would just have to fill out phases.

Reasonable? Ridiculous? Cool idea but impossible to implement?

billsacks:
@gold2718 : I like your suggestion. (Personally, the phases I care about are build, run, and then anything that's expected after the run - including any comparisons done as part of the test (e.g., base vs. rest), comparisons with baselines, and baseline generation.) Given the amount of care that seems needed to ensure that all phases are reported in all cases (e.g., being very careful with exception handling), and the number of times I've seen bugs creep into this in just the last few weeks (resulting in the test comparisons not being done), it seems like your suggestion would add an extra level of robustness & confidence to the test reporting - in addition to making it easier to parse test results of a test-in-progress.

That said, I also don't have a great sense of how hard it would be to implement.

I'm not sure I agree with this, though:

> If the test is created with --no-submit then the test does not expect a run phase and if it is created with --no-build, it does not expect a build phase.

I sometimes use a workflow with --no-build, but rarely (if ever) has this been because I don't plan on building the test at all. Instead, my workflow in these cases looks like this:

1. I find a bug; I'm not sure what caused it but I have a few hypotheses

2. I set up ~ 4 tests at once with --no-build (actually, in cime4 and earlier, this was the default behavior when creating a single test... I like that it's no longer the default behavior, by the way)

3. In each of them, I make some change in sourceMods to test one of my hypotheses, then run case.build && case.submit

4. I see which passed & which failed via grep -v PASS TestStatus

Incidentally: This workflow is what led me to initially request at least one line that indicates a test is incomplete: Because there's a lot of manual intervention in this process, it's easy for me to forget to build & submit one of the cases. (This actually happens to me more often than I'd like... maybe I'm too absent-minded, or try to multi-task too much.)

gold2718:
Not that I don't want to accommodate your workflow but if we always pre-populate with build and run phases, then we can never have a test that does not include these. Would it be possible for you to use a little script to do your test? Something like (typing off the top of my head):

#!/bin/bash
if [ -n "`grep FAIL $1`" ]; then
  echo "$1 shows FAIL"
elif [ -n "`grep PASS $1`" ]; then
  if [ -z "`grep MODEL_BUILD`" ]; then
    echo "$1 is not BUILT"
  elif [ -z "`grep RUN`" ]; then
    echo "$1 has not been RUN"
  elif [ -n "`grep -v PASS`" ]; then
    echo "$1 is incomplete" # This test could change if INCOMPLETE becomes a status
  else
    echo "$1 appears to PASS"
  fi
else
  echo "$1 appears to be incomplete"
fi

slap that baby in your bin directory and adjust the logic as needed.

If you do build the model after a --no-build, the MODEL_BUILD phase would be added (same with --no-submit).

billsacks:
@gold2718 : Sorry, I didn't mean to imply that we should be designing things to match my workflow (okay, fine, maybe I would like that, actually ;-)  ). What I meant was: I have never viewed `--no-build` / `--no-submit` as being a mechanism for doing testing where you don't want to build / submit a test: rather, I have always viewed those as being for the case where you don't want tests _automatically_ built & submit, but rather plan to do it yourself: maybe `--no-auto-build` / `--no-auto-submit` would be more precise names. In particular, I can't imagine how these options could be used to accommodate specific tests that do not include these phases, since this test would presumably be in a test suite with other tests that _do_ have build & run phases. I liked your idea of having the tests themselves dictate whether they have certain phases, and I could imagine that being extended to say, "I don't have a build phase", for example. My point was just that, if I run a test that _typically_ has build & run phases (let's say ERS), then it should be marked as incomplete in some way if it hasn't been built / run, even if I specified `--no-build` to `create_test`.

That said, thanks for the useful little script :-)

gold2718:
@billsacks Aha, that explanation helps. Therefore, I also like --no-auto-build over --no-build (and s/build/submit/g) for clarity.

jgfouca:
@billsacks OK, it sounds like we have agreement. The changes on this PR are still needed, so you can go ahead and merge and then we can revisit your PR.

billsacks:
I'm moving the broader discussion of this possible requirement:

> If there is a TestStatus file in the test directory, and if the test has not completed successfully, then there must be at least one line in the TestStatus file with a status other than PASS

along with @gold2718 's suggestion to https://github.com/ESMCI/cime/issues/399

Please make any further comments there.

PR #394, opened 2016-08-12
work on support for external systems

jedwards4b:
Allows for a machine to be defined in xml in $HOME/.cime/config_*.xml (machines, compiler, batch)   This is useful for vendors and other users who do not want to have their machine defined in our distributed xml files.  

Test suite: scripts_regression_tests on yellowstone after removing all reference to yellowstone from the cime_config/cesm/machines directory.    
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Code review:


jgfouca:
Looks good.

PR #395, opened 2016-08-13
Fix vars resolved too early

jedwards4b:
Some variables in env_*.xml files were being resolved too early - this resulted in a number of problems including one found by John Truesdale:  If the variable CASE was defined in the environent when you ran create_newcase then the EXEROOT and RUNDIR would be defined using the CASE from the environment instead of the one in the create_newcase argument. 

Another possible test for this would be for user b to try to clone a case created by user a - if this results in failures because user a paths are still in the case cloned by user b than there are likely other variables being resolved too early.  
Test suite: cime_regression_tests
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Code review: mvertens, jgfouca

### DIFF
--- a/driver_cpl/bld/build-namelist
+++ b/driver_cpl/bld/build-namelist
@@ -276,24 +276,12 @@ if (defined $opts{'infile'}) {
     $nl->merge_nl($nl_infile_valid);
 }
 
+unshift @INC, "${CIMEROOT}/utils/perl5lib";
+require Config::SetupTools;
+
 #-----------------------------------------------------------------------------------------------
 # Determine xml variables
 #-----------------------------------------------------------------------------------------------
-
-my %xmlvars = ();
-my @files = <${CASEROOT}/*xml>;
-foreach my $file (@files) {
-    my $xml = XML::Lite->new( "$file" );
-    my @e = $xml->elements_by_name('entry');
-    while ( my $e = shift @e ) {
-	my %a = $e->get_attributes();
-	my $val = $a{value};
-	$xmlvars{$a{'id'}} = $a{'value'};
-    }
-}
-
-unshift @INC, "${CIMEROOT}/utils/perl5lib";
-require Config::SetupTools;
 
 my %xmlvars = ();
 SetupTools::getxmlvars($CASEROOT, \%xmlvars);
--- a/driver_cpl/bld/build-namelist
+++ b/driver_cpl/bld/build-namelist
@@ -47,7 +47,7 @@ OPTIONS
 
 NOTE: The precedence for setting the values of namelist variables is (highest to lowest):
       1. values read from the file specified by -infile,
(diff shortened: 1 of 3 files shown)


driver_cpl/bld/build-namelist
```diff
@@ -287,14 +287,11 @@ foreach my $file (@files) {
     my @e = $xml->elements_by_name('entry');
     while ( my $e = shift @e ) {
 	my %a = $e->get_attributes();
+	my $val = $a{value};
```
mvertens: $val is a local variable that does not appear to be used in this while loop - so why is it introduced?

mvertens:
I am not seeing how these changes would deal with the CASE value being obtained from the environment rather than the xml file. Could you please clarify a bit.

driver_cpl/bld/build-namelist
jedwards4b: You're right I started to do this one way and then noticed that I just needed to move the DIN_LOC_ROOT line down below the next xmlvars initialization - is there a reason to do it twice or can I just remove the first one?

driver_cpl/cime_config/config_component.xml
```diff
@@ -154,7 +154,7 @@
 
   <entry id="USER">
     <type>char</type>
-    <default_value>UNSET</default_value>
```
jedwards4b: This now sets the correct value for USER in the xml file.

utils/python/CIME/case.py
```diff
@@ -491,9 +491,10 @@ def configure(self, compset_name, grid_name, machine_name=None,
                       'COMPILER' not in x and 'MPILIB' not in x]
 
         for nodename in nodenames:
-            value = machobj.get_value(nodename)
+            value = machobj.get_value(nodename, resolved=False)
```
jedwards4b: This resolved=False causes the EXEROOT and other variables in the file to be $CESMSCRATHROOT/$CASE/bld
instead of /glade/scratch/jedwards/whatever/bld  without the resolved=False this was trying to resolve case before the case variable is defined in xml and so was getting it from the environment.

driver_cpl/bld/build-namelist
jedwards4b: This code was changed because after I fixed the resolve issue in the env_run.xml this was causing the ATM2ICE_SMAPNAME and others to get unresolved values which they then tried to resolve incorrectly.
Moving DIN_LOC_ROOT to after the expand_xml_var solved this problem.

jedwards4b:
It seems like the xml files were being read twice in driver build-namelist.  I have removed the redundant read.

mvertens:
I am fine with the implementation.

jgfouca:
Looks good to me.

PR #398, opened 2016-08-15
More early resolve issues

jedwards4b:
Remove more early resolve issues - these were found by cloning a case from one user to another.

Test suite: scripts_regression_test & create_clone from one user to another
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes [CIME Github issue #]

User interface changes?: 

Code review:

### DIFF
--- a/driver_cpl/cime_config/config_component.xml
+++ b/driver_cpl/cime_config/config_component.xml
@@ -1766,14 +1766,6 @@
     <desc>Machine name</desc>
   </entry>
 
-  <entry id="MACHINES_FILE">
-    <type>char</type>
-    <default_value></default_value>
-    <group>case_def</group>
-    <file>env_case.xml</file>
-    <desc>full pathname of file specifying supported machines location</desc>
-  </entry>
-
   <entry id="MACHDIR">
     <type>char</type>
     <default_value></default_value>
--- a/utils/python/CIME/XML/machines.py
+++ b/utils/python/CIME/XML/machines.py
@@ -27,9 +27,9 @@ class Machines(GenericXML):
         if infile is None:
             if files is None:
                 files = Files()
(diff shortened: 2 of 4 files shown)


driver_cpl/cime_config/config_component.xml
```diff
@@ -1766,14 +1766,6 @@
     <desc>Machine name</desc>
   </entry>
 
-  <entry id="MACHINES_FILE">
```
jedwards4b: Not used anyplace as far as i can tell.

utils/python/CIME/utils.py
```diff
@@ -21,8 +21,8 @@ def expect(condition, error_msg, exc_type=SystemExit):
     """
     if (not condition):
         # Uncomment these to bring up a debugger when an expect fails
-        #import pdb
-        #pdb.set_trace()
+        import pdb
```
jgfouca: You probably didn't intend to push this.

jedwards4b:
I've removed the debug stuff and updated the PR

utils/python/CIME/XML/machines.py
```diff
@@ -27,9 +27,9 @@ def __init__(self, infile=None, files=None, machine=None):
         if infile is None:
             if files is None:
                 files = Files()
-            infile = files.get_value("MACHINES_SPEC_FILE")
+            infile = files.get_value("MACHINES_SPEC_FILE", resolved=False)
             self.machines_dir = os.path.dirname(infile)
-
+            infile = files.get_resolved_value(infile)
```
jgfouca: I don't understand this change. Why is calling get_resolved_value by hand better than resolved=True?

jedwards4b: I want self.machines_dir to have the unresolved value

PR #400, opened 2016-08-15
Mfdeakin sandia/refactor/class cleanup

mfdeakin-sandia:
Test suite: cime_developer, scripts_regression_tests
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes nothing
Passes the same tests as master on my machine

User interface changes?: Replaces classes EnvBase, EnvBuild, EnvCase, EnvRun which introduce no new functionality with class method constructors to EntryID. Remaining classes that did inherit from EnvBase should be constructed with CLASSNAME.constructCLASSNAME(caseroot) as they were previously.  EnvBuild, EnvCase, and EnvRun were deleted and should be constructed as EntryID.constructEnvBuild(caseroot), EntryID.constructEnvCase(caseroot), and EntryID.constructEnvRun(caseroot), respectively.

Code review:


jedwards4b:
What was the motivation for this change?   It provides no new functionality, fixes nothing and in my opinion obfuscates code.

mfdeakin-sandia:
My motivation was to make the class hierarchy consistent with provided functionality. The removed classes only existed for their default parameters, when really they should be treated as EntryID types.

jedwards4b:
It should have been discussed with the group before you put the effort into it.

mfdeakin-sandia:
Jim gave me the assignment of improving the code in CIME (specifically the case class, but this looked like an easier place to start) as a way to really learn it, so that's why I was looking into doing this in the first place.
Could you clarify on how I could make the code more readable? Using classes when another constructor is what was really wanted seems wrong to me for reasons including readability.

jedwards4b:
The intent was that each xml file have a corresponding class in python in a file in the CIME/XML directory.   I think that relationship is important to maintain.

mfdeakin-sandia:
Shouldn't we be aiming for less code for readability and understanding though?
I don't understand why we want a correspondence between classes and configuration files. Is this a best practice I'm unaware of, and if so could you explain it? If not, I guess it's okay for a small number of classes and files, though I worry about scalability.

jedwards4b:
The number of classes and configuration files is pretty limited, I don't think scalability is a concern.  
The point is that if you find that you need to change something in the format of the env_build.xml that will now distinguish it from other entryid files there is a clear single place to do that in XML/env_build.py  I also think that distinguishing the interface to each of these files by class is helpful in debugging.

mfdeakin-sandia:
Fair enough. I'm not certain now is the right time to be worried about changes like that, but feel free to close this pull request.
Thanks for trying to help me understand.

jedwards4b:
Thank you, I look forward to your next contribution - please feel free to discuss your ideas on our slack channel at https://cime.slack.com/

Issue #290, opened 2016-07-25
Test cases should themselves be tested to ensure they pick up failures


billsacks:
This issue is prompted by my finding two bugs in utils/python/CIME/SystemTests/lii.py, which both acted to make this test always pass, even if there were problems with the system under test that should make the test fail.

In talking with @mvertens , it sounds like the new python-based tests have generally been tested to ensure that they pass when they should, but not all of them have been tested to ensure that they fail when they should - i.e., that they actually pick up the sorts of problems they are designed to catch. @mvertens feels that the core tests (ERS, ERP and a few others) have been extensively vetted, but that other tests like LII, NCK and others have undergone less careful scrutiny.

Ideally we would have automated tests that test the test scripts, making sure that they fail when they should. But it could be hard to construct a test like this. So, lacking that, I'd be more comfortable if I knew that there had at least been manual testing done on each non-trivial test type to ensure that it indeed fails when it should.

Since this may be a gradual process, I'd suggest using this issue to keep track of what testing has been done. We can close this issue once all non-trivial tests have been tested in this way.

jedwards4b:
To the extent that this is done it is done in the scripts_regression_tests script.  It is important to distinguish tests that can (and should) be added as extensions to this script and what tests rely on the existence of external components that may or may not be present.   These tests might go into a CIME_MODEL specific extension of the tests.

billsacks:
I have done manual testing of the ERS and ERP tests from cime5.0.2. The ERS test fails when it should, but the ERP test does not: see issue #295.

billsacks:
Brainstorming a bit about how we could automate the testing of these high-level system test scripts.... From a quick look, I don't see an easy way to put unit tests around them as is done in some other places in scripts_regression_tests. It feels like, to be confident that they're working right, you need to force a failure in the system under test (i.e., CESM or ACME), and verify that the test correctly reports the failure.

[...]

PR #404, opened 2016-08-16
Infrastructure for clone-based two-part system tests

### LINKED ISSUE #214: Add reproducibility test
This was suggested at today's CSEG meeting; it doesn't seem critical, but could be a good addition to the set of possible system test types (SMS, ERS, etc.): A test that two startup runs with the exact same configuration are bit-for-bit identical.

The specific issue this would have caught was a reproducibility problem that just appeared in the first timestep of a startup run with WaveWatch. Because this problem only appeared in startup runs, exact restart tests didn't catch it. Regression tests against baselines would have caught it, but there were expected answer changes that masked the existence of this problem.

For more heavily-developed components, bit-for-bit refactoring is common enough that this test may not be needed, since reproducibility problems are usually caught in the course of running the component's system test suite on a set of changes that are expected to be bit-for-bit. But it could be a useful, quick sanity check: e.g., if an ERS test is failing, it can be helpful to confirm that two startup cases give bit-for-bit identical results before looking into problems that are specific to restart.

### LINKED ISSUE #291: LII test does not fail when it should
There are two bugs in the LII test (utils/python/CIME/SystemTests/lii.py):

1. the 'interp' version sets use_init_interp to .false., when it should in fact be set to .true.

2. The comparison between the 'base' and 'init_interp_on' versions was never implemented

### LINKED ISSUE #292: Refactor system test scripts to make it easier to add new tests that compare two runs
Most of our system tests do two runs that differ in some way, then compare them to make sure they are identical. There is a desire - from myself and others - to add many more tests along these lines that tweak small things and ensure that the simulation isn't affected (e.g., turn carbon isotopes on in CLM and ensure that this doesn't change the evolution of the bulk carbon pools).

The new python-based system tests are a big step forward from the old shell scripts - thank you to everyone who worked on that! However, to make the test scripts more maintainable, and to make it easier to add new system tests, I'd suggest even more aggressive removal of duplication between these tests. Here I'm using the LII test as an example. 

There are three main things I'd like to see:

1. Introduce a base class for doing two runs

2. Introduce a pre-build stage, so I can move the namelist setup out of the build stage

   The main point would be to avoid needing to check if things have already been done and avoid doing them again. This would also make the whole test flow logic more clear.

3. Move this to some common place:

```python
        self._case.set_value("CONTINUE_RUN",False)
        self._case.set_value("REST_OPTION","none")
        self._case.set_value("HIST_OPTION","$STOP_OPTION")
        self._case.set_value("HIST_N","$STOP_N")
```

I envision something like the following prototype (which is probably buggy, but hopefully conveys what I envision well enough):

```python
class SystemTestsCompareTwo(SystemTestsCommon):

    def __init__(self, case):
        self._test_suffix = 'test'  # will typically be overridden in derived classes

    def run(self):
        self._setup_first_phase()
        success = SystemTestsCommon._run(self)

        if success:
            self._setup_second_phase()
            success = SystemTestsCommon._run(self, self._test_suffix)

        if success:
            success = self._component_compare_test("base", self._test_suffix)

        return success

    def _setup_first_phase(self):
        pass

    def _setup_second_phase(self):
        pass


class LII(SystemTestsCompareTwo):
    def __init__(self, case):
        SystemTestsCompareTwo.__init__(self, case)
        self._test_suffix = 'init_interp_on'

    # This is a new method that does not currently exist in
    # SystemTestsCommon. For many tests it won't do anything, so the default
    # implementation in SystemTestsCommon will just be 'pass'. However, it can
    # be overridden to do something useful in tests where something is needed
    # prior to the build. The advantages of having this in a separate stage
    # rather than as part of the build are:
    #
    # (a) This is just called once, right after test creation, even if the build
    # is called multiple times - so you don't need logic in here ensuring that
    # it's only done once
    #
    # (b) For tests that need this setup but don't need to override the standard
    # build method, the method signature can remain simpler, and you don't need
    # to explicitly call SystemTestsCommon.build(...)
    #
    # (c) Separating things into when they actually need to be done yields
    # better communication about the actual process of setting up the test
    #
    # I haven't looked through the higher-level scripts to determine where this
    # will be called from.
    def initial_setup(self):
        # Note: If copying user_nl files and making some small tweaks to them
        # for the two different runs is a common operation, then this operation
        # could be moved into the SystemTestsCompareTwo class, and called from
        # each derived class that needs it.

        os.makedirs("user_nl_nointerp")
        for filename in glob.glob(r'user_nl_clm*'):
            shutil.copy(filename, os.path.join("user_nl_nointerp",filename))
            with open(os.path.join("user_nl_nointerp",filename), "a") as newfile:
                newfile.write("use_init_interp = .false.")

        os.makedirs("user_nl_interp")
        for filename in glob.glob(r'user_nl_clm*'):
            shutil.copy(filename, os.path.join("user_nl_interp",filename))
            with open(os.path.join("user_nl_interp",filename), "a") as newfile:
                newfile.write("use_init_interp = .false.")


    def _setup_first_phase(self):
        caseroot = self._case.get_value("CASEROOT")
        for filename in glob.glob(r'user_nl_nointerp/*'):
            shutil.copy(filename, os.path.join(caseroot,filename))

    def _setup_second_phase(self):
        caseroot = self._case.get_value("CASEROOT")
        for filename in glob.glob(r'user_nl_interp/*'):
            shutil.copy(filename, os.path.join(caseroot,filename))

```

The intent here is to make the specific test class (LII above) contain only what is unique to that test. Once that's the case, it will become much easier to create and maintain a wide variety of tests.

billsacks:
This adds infrastructure for doing two-part system tests (i.e., our
typical system tests, where you do two runs and compare them) via
clones. From discussion on a recent CIME telecon, there was general
agreement that we want to migrate our system tests to this style where
possible: It's easier to write, understand and maintain tests set up
like this, where there are two separate cases - rather than trying to do
two different runs from the same case, which requires careful saving and
restoring of the appropriate files.

One slight oddity: A TestStatus.log file appears in the clone directory
(rather than the test's main case directory). At least for passing
tests, this only includes a date/time stamp, so this didn't seem to be a
problem. But it's possible that this could include more information if a
test failed at a certain point. Presumably this is because I chdir to
the case2 case directory before running case2 (I tried not doing this
chdir, but this caused the run to fail) - and presumably there is some
code somewhere that opens TestStatus.log in the current directory. I
haven't tried to dig into this, because it didn't seem like a big deal
to me, but can come back to it if others feel it's a problem.

This also adds a new test based on this infrastructure (REP - a basic
reproducibility test), and rewrites one existing test based on this
infrastructure (LII).

This also adds some unit tests written using the unittest module, and
adds the capability for scripts_regression_tests to discover other
unittests that may be distributed throughout the code base.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

   All passed except for the same 4 failures that I got in testing master:

```
test_check_code (__main__.CheckCode) ... FAIL
test_b_full (__main__.E_TestTestScheduler) ... FAIL
test_save_timings (__main__.TestSaveTimings) ... FAIL
test_full_system (__main__.Z_FullSystemTest) ... FAIL
```

Fixes #214
Fixes #291
Fixes #292

Connects to #290

User interface changes?: none

Code review: @jedwards4b @jgfouca @mvertens @rljacob


utils/python/CIME/SystemTests/rep.py
```diff
@@ -0,0 +1,22 @@
+"""
+Implementation of the CIME REP test
+
+This test verifies that two identical runs give bit-for-bit results
+"""
+
+from CIME.SystemTests.system_tests_compare_two import SystemTestsCompareTwo
+
+class REP(SystemTestsCompareTwo):
```
billsacks: Is REP a bad name? I realized that it's easy to mistype ERP as REP and vice versa. Is it time to start allowing longer test names? ("REPRODUCE"???)

utils/python/CIME/SystemTests/tests/test_system_tests_compare_two.py
```diff
@@ -0,0 +1,384 @@
+#!/usr/bin/env python
```
billsacks: Note that I have unit tests living in a tests directory that sits along-side the code being tested. I find this to be an easy-to-work-with organization of test code, but am open to reorganizing this if others prefer some different organization.

One thing I'd like to avoid is having test code living in the same module as the production code: I feel like that makes it harder to understand what's test code and what's production code.

utils/python/CIME/SystemTests/tests/case_fake.py
```diff
@@ -0,0 +1,99 @@
+"""
+This module contains a fake implementation of the Case class that can be used
+for testing the tests.
+"""
+
+import os
+from copy import deepcopy
+
+class CaseFake(object):
```
billsacks: If this CaseFake class would have more general utility, it could be moved to some higher level

utils/python/tests/scripts_regression_tests.py
```diff
@@ -41,13 +41,41 @@ def test_resolve_variable_name(self):
...
+            pattern = 'test*.py')
+
+        testrunner = unittest.TextTestRunner(buffer=False)
+
+        # Disable logging; otherwise log messages written by code under test
+        # clutter the unit test output
+        log_lvl = logging.getLogger().getEffectiveLevel()
+        logging.disable(logging.CRITICAL)
+        try:
+            results = testrunner.run(testsuite)
```
billsacks: Note that this produces output like this:

```
test_unittests (__main__.A_RunUnitTests) ... .............
Ran 13 tests in 0.021s

OK
ok
```

According to earlier discussions with @jgfouca , it sounded like this bit of extra output is okay.

utils/python/tests/scripts_regression_tests.py
```diff
@@ -1635,7 +1663,7 @@ def _main_func():
 
     CIME.utils.handle_standard_logging_options(args)
 
-    unittest.main(verbosity=2)
+    unittest.main(verbosity=2, catchbreak=True)
```
billsacks: catchbreak allows you to abort the unit test run with ctrl-C, and then get a full report of the tests that have failed so far. I found this useful, if I got some failures in the very first tests - so I could see a full report without having to wait for the entire scripts_regression_tests to finish.

utils/python/tests/scripts_regression_tests.py
```diff
@@ -41,13 +41,41 @@ def test_resolve_variable_name(self):
...
+        log_lvl = logging.getLogger().getEffectiveLevel()
+        logging.disable(logging.CRITICAL)
+        try:
+            results = testrunner.run(testsuite)
+        finally:
+            logging.getLogger().setLevel(log_lvl)
+
+        self.assertTrue(results.wasSuccessful())
+
+    def test_CIME_doctests(self):
```
billsacks: Renamed from unit_tests to doctests to be more accurate, and to distinguish this from the new test that runs all true unittests

utils/python/CIME/SystemTests/README
```diff
@@ -1,11 +1,13 @@
 The following are the test functionality categories:
   1) smoke tests
-  2) restart tests
-  3) threading/pe-count modification tests
-  4) sequencing (layout) modification tests
-  5) multi-instance tests
-  6) performance tests
-  7) spinup tests (TODO)
+  2) basic reproducibility tests
```
billsacks: I have updated the README in addition to the header documentation in config_tests.xml. We should probably pick one or the other to delete, to avoid redundancy.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -30,9 +30,23 @@ def __init__(self, case, expected=None):
         self._casebaseid = self._case.get_value("CASEBASEID")
         self._test_status = TestStatus(test_dir=caseroot, test_name=self._casebaseid)
 
+        self._init_environment(caseroot)
+        self._init_locked_files(caseroot, expected)
+        self._init_case_setup()
```
billsacks: I needed to break out these interactions with the environment and case object into separate methods so that I could override them for unit tests (avoiding their operation). I could have just split out one method, but I figured I'd give helpful names to the different things done here while I was at it.

billsacks:
Done adding line notes; ready for review

utils/python/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -0,0 +1,357 @@
...
+        self._common_setup()
+        self._case_two_setup()
+        # Flush the case so that, if errors occur later, then at least case2 is
+        # in a correct, post-setup state
+        self._case.flush()
+
+        # Go back to case 1 to ensure that's where we are for any following code
+        self._activate_case1()
+
+    def _force_case2_settings(self):
```
jedwards4b: Why not use case.submit for case2?   Are you running case2 in the same submit as case 1 and thus using the same environment and pe-layout?  That seems rather over restrictive since environment and pe-layout might be among the things that we want to test.

jedwards4b:
I've reviewed this code.   I don't know if I agree with the assertion that the two case approach is better than the current one, but I am willing to try it.   I think that you should consider going ahead and implementing the two build code.    I will repeat my request that if you are not going to follow the current directory structure and location of tests that you change the existing tests to follow your prefered structure and location.   Too many systems of organization is no organization at all.

rljacob:
So scripts_regression_tests will now run all the Fortran unit tests in the code?  Tagging @adubey64

jedwards4b:
No, I think Bill was just distinguishing between python doctests and python unittests.

utils/python/CIME/SystemTests/README
rljacob: Isn't the scripts/Testing/Testcases directory (which contains config_tests.xml) going away?  I'd remove it from config_tests.xml since there is a README in that directory with a pointer to the README here.

adubey64:
If you are referring to tests that should run with pFunit, I am not sure they are running. If someone knows otherwise please let me know. I have run into issues with pFunit for which I am scheduled to have a debugging session with Tom Clune tomorrow. So if someone else already has them figured out it would be good to know.

rljacob:
I was but that was a mistake.  That comment was about python unit tests.

billsacks:
@adubey64 : As @rljacob says, this PR isn't actually referring to pFUnit tests. However, I just successfully ran the pFUnit tests out of something close to cime master, using the instructions in README.unit_testing, as well as on my local machine, following the instructions here:

https://github.com/NCAR/cesm_unit_test_tutorial

utils/python/CIME/SystemTests/README
billsacks: @rljacob : @jedwards4b says config_tests.xml is being moved. I get the sense that there is some disagreement about where the documentation of the tests should live, so I'd rather not make a unilateral decision to remove the documentation from config_tests.xml in this PR.

utils/python/CIME/SystemTests/rep.py
jgfouca: Based on the parsers I've written, I don't think there's any limit on test name length. I'm fine with REP though.

utils/python/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -0,0 +1,357 @@
...
+
+In the __init__ method for your test, you MUST call
+    SystemTestsCompareTwo.__init__
+See the documentation of that method for details.
+
+Classes that inherit from this are REQUIRED to implement the following methods:
+
+(1) _case_one_setup
+    This method will be called to set up case 1, the "base" case
+
```
jgfouca: Thanks for leaving good documentation.

utils/python/CIME/tests/SystemTests/test_system_tests_compare_two_link_to_case2_output.py
```diff
@@ -0,0 +1,144 @@
...
+import tempfile
+from CIME.SystemTests.system_tests_compare_two import SystemTestsCompareTwo
+from CIME.tests.case_fake import CaseFake
+
+# ========================================================================
+# Fake version of SystemTestsCompareTwo that overrides some functionality for
+# the sake of unit testing
+# ========================================================================
+
+class SystemTestsCompareTwoFake(SystemTestsCompareTwo):
```
jgfouca: Same classname appears twice?

utils/python/CIME/SystemTests/test_utils/user_nl_utils.py
```diff
@@ -0,0 +1,52 @@
+"""
+This module contains functions for working with user_nl files in system tests.
```
jgfouca: These methods look like they might be generally useful. Why did you decide to make a new directory instead of just adding a test_utils.py file to the common CIME area?

utils/python/tests/scripts_regression_tests.py
jgfouca: Yeah, that's a nice change. I wasn't aware of this option.

utils/python/CIME/tests/case_fake.py
```diff
@@ -0,0 +1,99 @@
+"""
+This module contains a fake implementation of the Case class that can be used
+for testing the tests.
```
jgfouca: Another new directory? What's the conceptual distinction of files here?

utils/python/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -0,0 +1,357 @@
...
+        location, it is maintained as is. However, an exception will be raised
+        if the destination link is not exactly as it should be: we avoid
+        overwriting some existing file or link.
+        """
+
+        casename1 = self._case1.get_value("CASE")
+        casename2 = self._case2.get_value("CASE")
+        rundir1 = self._case1.get_value("RUNDIR")
+        rundir2 = self._case2.get_value("RUNDIR")
+        run2suffix = self._run_two_suffix
```
jgfouca: I won't be able to have an informed opinion until I try to use this framework to set up a test. That said, I'm a little concerned at the size and complexity (14 methods!) of this class when the goal was to reduce code size. It does appear that your lii.py test was made very easy under this framework, so that's good. If you can get similar benefits on several of the other tests then I guess I'd be comfortable with this.

jgfouca:
I'll echo some of @jedwards4b concerns regarding directory organization and add that I was quite surprised at how much code was added in this PR. A fair bit of it was test code, so most of my concern is centered on system_tests_compare_two.py

billsacks:
I have addressed some of @jedwards4b and @jgfouca 's comments via commit 4560c5a3850e245ad1f529d8caab022e3245edee - specifically, addressing the organization of the new unit tests.

In consultation with @jedwards4b , I have set up a new directory structure under utils/python/CIME/tests, which mirrors the production directory structure. So, for example, unit tests for CIME/foo.py should live in CIME/tests/test_foo.py, and unit tests for CIME/SystemTests/bar.py should live in CIME/tests/SystemTests/test_bar.py. We felt this was better than adding yet more unit tests directly to scripts_regression_tests.py.

(The original suggestion @jedwards4b made was to put these under utils/python/tests/CIME. However, doing so felt a bit wrong to me, because we don't currently have any loadable python code under utils/python/tests. So, for example, I would have needed to add an __init__.py file in utils/python/tests. It felt best to me to put these unit tests under utils/python/CIME, since that's the root of the main production python modules. But I'm open to switching this if others prefer a different organization.)

Along with this reorganization, I moved case_fake.py to a higher level - now under utils/python/CIME/tests. @jedwards4b and I liked this better than putting CaseFake in the production code.

utils/python/CIME/tests/SystemTests/test_system_tests_compare_two_link_to_case2_output.py
billsacks: I define a Fake version in both test_system_tests_compare_two.py and test_system_tests_compare_two_link_to_case2_output.py, but these fakes differ slightly: different things need to be stubbed out for those two sets of tests. If you prefer, I can give the Fakes more meaningful names to distinguish between them.

utils/python/CIME/SystemTests/test_utils/user_nl_utils.py
billsacks: I figured that these would just be useful for system tests - I couldn't see when else you might want to call a function to add something to a user_nl file, since typically that's an operation that a user does.

I have no objection to moving these to a higher level, though. Note, however, that there is already a CIME/test_utils.py that seems to be very different:

```
Utility functions used in test_scheduler.py
```

I could imagine moving this to something like CIME/system_test_utils.py. Would you prefer that?

utils/python/CIME/tests/case_fake.py
billsacks: Sorry, I don't understand what you're asking here.

utils/python/CIME/SystemTests/system_tests_compare_two.py
billsacks: I talked with @jedwards4b about this in person, but for the record: I believe that trying to do a separate job submission would add a lot of complexity. If it becomes needed at some point, we can revisit this possibility.

utils/python/CIME/SystemTests/system_tests_compare_two.py
billsacks: > I'm a little concerned at the size and complexity (14 methods!) of this class when the goal was to reduce code size.

The goal here was to reduce code size and complexity of the individual tests. As you note from the LII test, I believe this goal was accomplished. Since you and @jedwards4b both seem a bit skeptical on this, I just rewrote the PET test to use this new infrastructure; I'll push that in a moment. I hope this will demonstrate that the LII test is not unique in this respect. (I chose the PET test because it doesn't require two builds and doesn't require setting CONTINUE_RUN. As we have discussed, tests that set CONTINUE_RUN for one of the runs will be more challenging to fit into this infrastructure. Tests that require two builds will require a bit of extension to the infrastructure, but I believe that the individual tests themselves will be no more complex than the current examples.)

@mvertens has suggested that I should bring this to master before doing a more extensive rewrite of our other tests, but if you and @jedwards4b feel uncomfortable bringing this to master before this is further demonstrated, then we can revisit this.

I find it is often the case that enabling "client" code (in this case the individual tests) to be simpler requires some additional complexity in the shared infrastructure code. So I don't think it should be surprising that SystemTestsCompareTwo has some complexity. Writers of individual tests do not need to know anything of this complexity. Furthermore, the complex aspects are wrapped in unit tests, providing a level of confidence that was hard to achieve when complexity was duplicated in multiple individual tests.

Finally, it appears that you have some discomfort about the number of methods in this class. I tend to be aggressive in my application of the "Extract method" refactoring, creating new methods whenever I see duplicated code or just some code that I want to abstract away. But I'd be fine inlining some of the methods if you feel that would make this class easier to understand. If so, I'd be happy to hear your thoughts on which methods should be inlined.

billsacks:
I just pushed three additional commits, which rewrite the PET test to use this new infrastructure. This was in response to some concern that I sensed from @jedwards4b and @jgfouca about how well this would generalize to other tests. I hope you'll agree with me that this has allowed some nice simplification of the PET test as well.

I have run this test to demonstrate that the new PET test passes:

PET_D.f10_f10.I1850CLM45BGC.yellowstone_gnu.clm-interp_f19_noncrop

I have also verified the number of threads in the cesm log files for each case.

I have tried to demonstrate a failure, but am having trouble at the moment coming up with a case that causes a failure in the comparison without causing the model to crash.

billsacks:
One oddity I noticed when testing the new PET test: SETUP appears in TestStatus in the clone case rather than the main test case. However, I also noticed that, on master, SETUP doesn't seem to appear in TestStatus at all (I checked an ERS and a PET test). So I suspect that, if that problem were fixed, then we'd see SETUP in the TestStatus file in both the main case and the clone case... which seems a bit odd, perhaps, but probably wouldn't cause any problems?

billsacks:
Okay, I just forced a failure in the new PET test, by introducing a threading bug in CLM, leading to a FAIL in COMPARE_base_single_thread. So I'm pretty confident it's working right.

utils/python/CIME/SystemTests/pet.py
```diff
@@ -6,63 +6,41 @@
...
 
-        self.run_indv(suffix="single_thread")
-        self._component_compare_test("base", "single_thread")
-
-    def run_phase(self):
-        self._pet_first_phase()
-        self._pet_second_phase()
+        # TODO(wjs, 2016-08-16) Is this needed? I'm thinking it may be because
+        # we changed the number of threads, but I'm not sure
+        case_setup(self._case, reset=True)
```
mvertens: I definitely think that the new implementation for the PET test is much cleaner.

jedwards4b: yes this is needed.  Changing the number of threads can change the arguments to the submit command and potentially changes the blocksize in pop and cice

utils/python/CIME/SystemTests/test_utils/user_nl_utils.py
jgfouca: It's fine for now, we can always move it later.

jgfouca:
I'm OK with merging this too. I think the risk of this breaking the merge efforts is low since this just impacts testing.

jgfouca:
@billsacks , I see that there are conflicts, can you do an upstream merge?

billsacks:
> @billsacks , I see that there are conflicts, can you do an upstream merge?

@jgfouca : Okay, I have done the merge, so it sounds to me like this is ready to come to master, I think.

PR #411, opened 2016-08-16
Batch fix reorder scripts regression tests

### LINKED ISSUE #401: Directories created by scripts_regression_tests should have unique names
I noticed that I was getting inconsistent results in my runs of scripts_regression_tests. I specifically noticed this in `test_createnewcase (__main__.B_TestCreateNewcase)`, which seemed to pass sometimes and fail sometimes. I began to suspect (but have not confirmed) that this was because I was running scripts_regression_tests out of two different cime clones at once (I was running this for my own branch at the same time as I was running it for master).

In looking a little closer, I found that, at least for this test, a directory is created with a hard-coded path, meaning that tests can stomp on each other. I'm guessing that was what happened in my case... but even if not, this opens the door to problems running multiple instances of the test suite at once, so it seems this should be fixed:

```python
    def setUp(self):
        self._testroot = MACHINE.get_value("CESMSCRATCHROOT")

    def test_createnewcase(self):
        testdir = os.path.join(self._testroot, 'scripts_regression_tests.testcreatenewcase')
        if os.path.exists(testdir):
            shutil.rmtree(testdir)
```

Would it be possible to create a directory with a unique ID in this case? Using https://docs.python.org/2/library/tempfile.html seems like it could be ideal, but maybe not necessary if there's some reason to do something different.

jedwards4b:
Reorder tests in scripts_regression_tests (fast first slower later)
Test suite: scripts regression tests
Test baseline: 
Test namelist changes: 
Test status:  1 known failure in TestTestScheduler
Fixes #401 

User interface changes?: 

Code review:


cime_config/cesm/machines/config_batch.xml
```diff
@@ -424,7 +424,8 @@
 
   <batch_system MACH="yellowstone" type="lsf" version="9.1">
     <queues>
-      <queue walltimemax="24:00" jobmin="1" jobmax="8">caldera</queue>
+      <queue walltimemax="24:00" jobmin="1" jobmax="1" jobname="case.lt_archive">hpss</queue>
```
jedwards4b: adds support for jobname allowing a special queue for a certain job

utils/python/CIME/XML/env_batch.py
```diff
@@ -330,6 +330,8 @@ def get_submit_args(self, case, job):
             if name is None:
                 submitargs+=" %s"%flag
             else:
+                if name.startswith("$"):
```
jedwards4b: correct parsing of name field in queue

jgfouca:
Looks good.

Issue #295, opened 2016-07-26
ERP test does not fail when it should


billsacks:
From cime5.0.2, the ERP test does not fail despite my deliberately introducing a restart bug in CLM.

I see that the critical 'self._component_compare_test("base", "rest")' call was removed in the July 12 commit dea5064a4ea64605ea35d0b0664d8b9e5ef667eb, where the ERP test was substantially rewritten. That removal seems sufficient to explain this bug.

I tested this by introducing these diffs in CLM (in components/clm/src/biogeophys/WaterStateType.F90):

```
Index: biogeophys/WaterStateType.F90
===================================================================
--- biogeophys/WaterStateType.F90    (revision 79738)
+++ biogeophys/WaterStateType.F90    (working copy)
@@ -807,7 +807,7 @@
     real(r8)         , intent(in)    :: watsat_col (bounds%begc:, 1:)  ! volumetric soil water at saturation (porosity)
     !
     ! !LOCAL VARIABLES:
-    integer  :: c,l,j,nlevs
+    integer  :: p,c,l,j,nlevs
     logical  :: readvar
     real(r8) :: maxwatsat    ! maximum porosity
     real(r8) :: excess       ! excess volumetric soil water
@@ -851,6 +851,13 @@
          dim1name='pft', &
          long_name='canopy water', units='kg/m2', &
          interpinic_flag='interp', readvar=readvar, data=this%h2ocan_patch)
+    if (flag == 'read') then
+       do p = bounds%begp, bounds%endp
+          if (this%h2ocan_patch(p) < 1._r8) then
+             this%h2ocan_patch(p) = this%h2ocan_patch(p) * 2._r8
+          end if
+       end do
+    end if

     call restartvar(ncid=ncid, flag=flag, varname='SNOCAN', xtype=ncd_double,  &
          dim1name='pft', &
```

I then ran ERS_Ld7.f10_f10.ICLM45.yellowstone_intel and ERP_Ld7.f10_f10.ICLM45.yellowstone_intel. 

The ERS test fails as it should:

```
[yslogin6:/glade/scratch/sacks/cesm_code/cesm2_0_alpha01_v2/cime/scripts/ERS_Ld7.f10_f10.ICLM45.yellowstone_intel.0725_a]$ cat TestStatus
PASS ERS_Ld7.f10_f10.ICLM45.yellowstone_intel INIT
PASS ERS_Ld7.f10_f10.ICLM45.yellowstone_intel CREATE_NEWCASE
[...]

PR #413, opened 2016-08-16
Upgrade history tools to python

### LINKED ISSUE #332: CIME baseline comparison infrastructure is clunky and outdated
This stuff remains mostly in csh and is causing problems for scripts like bless_test_results.

jgfouca:
Full change list:
1) Remove old shell-based tools and update calls to use python versions
2) Move most functionality in old shell tools into hist_utils.py
3) Make thin python wrapper programs to access hist_utils from the command line
4) Do st_archive as LAST step in run_indv so that coupler_log_path is not needed
5) Fix ERR test
6) Update fake tests to create a fake hist file
7) Large refactor of bless_test_results
8) Add new compare_test_results, counterpart to bless_test_results.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #332

User interface changes?: Significant changes to compare_* scripts

Code review: @jedwards4b @mvertens @billsacks @gold2718


utils/python/CIME/SystemTests/err.py
```diff
@@ -28,7 +28,7 @@ def run_phase(self):
             dout_s_root = self._case.get_value("DOUT_S_ROOT")
             rundir = self._case.get_value("RUNDIR")
             logger.info("staging files from archive %s" % dout_s_root)
-            for item in glob.glob(os.path.join(dout_s_root, "rest", "*", "*")):
+            for item in glob.glob(os.path.join(dout_s_root, "*", "hist", "*base")):
```
jgfouca: Important change: I think ERR was broken before this PR, we just didn't notice it. The *base file did not exist in the "rest" subdirectory before or after this PR, so the "rest" vs. "base" comparison in ers_second_phase should not have worked.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -174,17 +175,18 @@ def run_indv(self, suffix="base", coupler_log_path=None, st_archive=False):
         logger.info(infostr)
 
         case_run(self._case)
-        if st_archive:
```
jgfouca: Important change: Do archiving as last step in this function. Removes the need for coupler_log_path.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,172 @@
+"""
+Functions for actions pertaining to history files.
+"""
+
+from CIME.XML.standard_module_setup import *
```
jgfouca: This file contains the key changes, worth review.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,172 @@
...
+    drv_comp = Component()
+    models = drv_comp.get_valid_model_components()
+
+    for model in models:
+        model = "cpl" if model == "DRV" else model.lower()
+        yield model
+
+def _get_all_hist_files(testcase, model, suffix="", from_dir=os.getcwd()):
+    suffix = (".%s" % suffix) if suffix else ""
+    test_hists = glob.glob("%s/%s.%s*.h?.nc%s" % (from_dir, testcase, model, suffix))
```
jgfouca: This function is how I get away with not having to worry about extensions.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,172 @@
...
+            comments += "    Copying '%s' to '%s'\n" % (test_hist, new_file)
+            shutil.copy(test_hist, new_file)
+
+    expect(num_moved > 0, "move failed: no hist files found in rundir '%s'" % rundir)
+
+    return comments
+
+def _compare_hists(case, suffix1="", suffix2="", from_dir1=os.getcwd(), from_dir2=os.getcwd()):
+
+    testcase = case.get_value("CASE")
```
jgfouca: This is the most important function in the file. It encapsulates comparing a sets of hists files which is used by compare_test and compare_baseline.

utils/python/CIME/SystemTests/err.py
jedwards4b: thanks

billsacks:
I'd like to take a close look at this, but don't have time this week. If this can wait, I'd like to review it early next week.

billsacks:
Have you done some system tests where you force a difference in the history files, to confirm that this is picked up and reported correctly? I'd like to make sure that has been tested both for the in-test comparisons and for baseline comparisons.

jedwards4b:
I am ready to accept this PR.   It removes and replaces the some of the last functionality still in sh.

mvertens:
Before this PR is accepted - I think that Bill's question should be
answered. I feel that these type of tests should be run to confirm the
robustness of the PR - if they have not been run.

jgfouca:
@billsacks , yes we have our TESTRUNDIFF fake test case that generates a DIFF and confirms that it was caught. I hope the new python-based infrastructure is more robust than the old; there is some evidence this it is because I caught some longstanding problems when I made the switch.

scripts/Tools/component_compare_baseline
```diff
@@ -0,0 +1,55 @@
...
+                        help="Case directory")
+
+    parser.add_argument("-b", "--baseline-dir",
+                        help="Use custom baseline dir")
+
+    args = parser.parse_args(args[1:])
+
+    CIME.utils.handle_standard_logging_options(args)
+
+    return args.caseroot
```
billsacks: You only return args.caseroot, yet the caller expects caseroot and baseline_dir

scripts/Tools/baseline_gen_comp
```diff
@@ -1,265 +0,0 @@
-#!/bin/bash
```
billsacks: The baseline_gen_comp script provided critical functionality. I don't see where this is provided now. In particular, this allowed you to run all baseline comparisons or generations on a given test suite after the fact. Some typical workflows were:

1. Running a test suite without the -generate option, then later (after you determine things look good), run baseline_gen_comp to generate baselines in the appropriately-named baseline directory

2. Running a test suite with comparison against some baseline tag 'foo', but then later also comparing that test suite with some different baseline tag 'bar'.

3. Running a test suite without any baseline comparisons, and doing the comparisons in a separate step later. This is particularly useful when running two test suites at once from different code bases, then comparing them after they both finish.

Is it possible to achieve this functionality with the new scripts? If not, I would ask that this functionality be restored before this PR comes to master.

scripts/Tools/component_compare_baseline
jgfouca: Thanks

scripts/Tools/baseline_gen_comp
jgfouca: 1. This can be done with bless_test_results
2, 3: Good point, I don't think that can be trivially done with the scripts in this PR. I'll address this.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,172 @@
...
+    """
+    rundir   = case.get_value("RUNDIR")
+
+    return _compare_hists(case, suffix1, suffix2, rundir, rundir)
+
+def cprnc(file1, file2, case, rundir):
+    cprnc_exe = case.get_value("CCSM_CPRNC")
+    basename = os.path.basename(file1)
+    stat, out, _ = run_cmd("%s %s %s 2>&1 | tee %s/%s.cprnc.out" % (cprnc_exe, file1, file2, rundir, basename))
+    return (stat == 0 and "IDENTICAL" in out, out)
```
billsacks: This is not a new issue (it looks like this is just replicating the old behavior), so it doesn't necessarily need to be addressed in this PR, but if it isn't, I'd like to open a separate issue to have this fixed: It looks like the comparison passes as long as the string "IDENTICAL" is found _anywhere_ in the cprnc output. In the event that there is a history field with IDENTICAL in its name, the comparison would always look successful. It would be more robust to check against the exact string - i.e., looking for "diff_test: the two files seem to be IDENTICAL".

utils/python/CIME/hist_utils.py
billsacks: I believe that the main point of explicitly dealing with extensions in the previous version was so that we could just find the _last_ history file for each extension. If I understand the new code, it will compare _all_ history files. So, for example, if we run a 5-day test with daily output, then the new version will do 5x as many history file comparisons.

I'm not sure which is better. Certainly the new version is more robust, in that it will catch problems with diagnostic fields that could have slipped through the cracks in the old version. However, it also means greater cost of the cprnc step - I'm going to guess something like 2x - 10x typically, but as high as 100x in a few cases for the CLM test suite, for some long, coarse-resolution tests.

If we do stick with this new version, I think that will argue for:

1. Keeping cprnc in Fortran as opposed to python, because its performance will now be noticeable in more applications; and/or:

2. Decreasing the amount of history output produced by some of the long tests in the test suite.

@mvertens : Do you have any thoughts on this?

billsacks:
> @billsacks , yes we have our TESTRUNDIFF fake test case that generates a DIFF and confirms that it was caught. I hope the new python-based infrastructure is more robust than the old; there is some evidence this it is because I caught some longstanding problems when I made the switch.

I would definitely believe that the new infrastructure is more robust than the old, and I like the way you have changed things to be robust to some additional cases. (I was going to make a request yesterday to add some more robustness in this respect, and then saw that you just added it - so thank you!) I also think that it's great to have this TESTRUNDIFF fake test case.

However, I still feel like these unit tests need to be accompanied by some manual system tests for changes of this magnitude, and @mvertens also agreed in a conversation I just had with her. It sounds like @mvertens is in the process of doing some manual tests.

Brainstorming what I think should be verified, at the request of @mvertens (it's possible that enough confidence can be gained from unit tests for some of these, but it's not clear to me how many of these are covered by unit tests right now... and it would be good to supplement unit tests with manual comparisons for at least the most critical of these like (2)).

1. Passing in-test comparison
a. Ensure that all history files were compared in a fully-coupled case; it's probably sufficient to ensure that all of the expected cprnc files were created.

2. Failing in-test comparison: can be verified with an ERS test using the diffs I gave here: #295 

3. Baselines generated correctly: ensure that all expected files show up in the baseline directory

4. Results reported correctly when a file exists in the baseline directory but not in the test case, for some / all of the history files
a. This can be tested by adding some fake file to the baseline directory
b. I can't remember how this was reported in the past

5. Results reported correctly when a file exists in the test case but not in the baseline directory, for some / all of the history files
a. i.e., baseline directory exists, but some / all history files are missing from it

6. Results reported correctly when the baseline directory is missing entirely
a. e.g., if you're comparing some test ERS.f09_g16.X.yellowstone_intel in baseline tag cesm2_0_alpha01e, I'm imagining that the cesm2_0_alpha01e directory exists, but there is no ERS.f09_g16.X.yellowstone_intel subdirectory

7. Some history files identical but some differ
a. From examining the code, I'm pretty confident this works correctly, and it's probably hard to force with a manual test, so I'm okay skipping it. (But it would be a good target for a unit test if there isn't already one like that.)

8. What about the various _N2, NCK, etc. cases that were handled specially in the old component_compare_test.sh? I can't say I understand the old logic well, so it's hard for me to tell what is needed in the new code to reproduce the necessary functionality.

jedwards4b:
In hist_utils.py _iter_model_file_substrs() is using drv_comp.get_valid_model_components
which is a list of component class names (atm, lnd, etc) but it should use the list 
of specific component names (cam, clm, etc)

jedwards4b:
I think I have a fix, testing now.

utils/python/CIME/hist_utils.py
jgfouca: Yeah, I'd rather check return code, but I don't think I can currently rely on cprnc to have a sane return code. I'll look for "seem to be IDENTICAL"

utils/python/CIME/hist_utils.py
jgfouca: @billsacks , that's a good point. Let's not forgot that we are running on modern machines and we should be able to do at least 4-8 cprnc's in parallel.

jedwards4b: Right now it's comparing all of the history files - I would suggest that it would be better to solve this by reducing the history files written in the test instead of hard coding something here.

billsacks: > Right now it's comparing all of the history files - I would suggest that it would be better to solve this by reducing the history files written in the test instead of hard coding something here.

I agree, in principle. However, off-hand, I'm having trouble seeing an easy way to do that. Two options I can think of are:

1. Giving components the option to listen to the coupler's HIST_OPTION in determining their history frequency. I have done this for CISM, but I don't think any other CESM components have this ability. I discussed this with @mvertens (and maybe you?) a while ago, and we felt this would be non-trivial.

2. Using testmods. However, most tests in the CLM test suite already have testmods to control their options, so adding separate testmods to control history frequency would lead to an explosion of testmods - e.g., in addition to the existing crop and crop_monthly testmods that we have, we'd need crop_annual, crop_5years, crop_20years, etc. The best way I could see to handle this would be to allow a test to specify two different testmods that are both applied... if that's feasible, that could be a good solution.

Thoughts?

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,175 @@
...
+    test_hists.extend(glob.glob("%s/%s.%s*.h.nc%s" % (from_dir, testcase, model, suffix)))
+    test_hists.extend(glob.glob("%s/%s.%s*.h?.*.nc%s" % (from_dir, testcase, model, suffix)))
+    test_hists.extend(glob.glob("%s/%s.%s*.h.*.nc%s" % (from_dir, testcase, model, suffix)))
+    if suffix == "":
+        test_hists.extend(glob.glob("%s/%s.h.nc" % (from_dir, model)))
+        test_hists.extend(glob.glob("%s/%s.h?.nc" % (from_dir, model)))
+    test_hists.sort()
+    return test_hists
+
+def move(case, suffix):
```
billsacks: For this and other functions in this file - at least for public functions: Can you please add documentation? At the least, I'd like to have documentation on each of the arguments to each function.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,175 @@
...
+import logging, glob, os, shutil, re
+
+def _iter_model_file_substrs(case):
+
+    models = case.get_compset_components()
+    models.append('cpl')
+    for model in models:
+        yield model
+
+def _get_all_hist_files(testcase, model, suffix="", from_dir=os.getcwd()):
```
billsacks: Here and elsewhere: Please change the way the directory defaults are done so that the default is None, and then, within the function, if the argument is None it is set to os.getcwd(). (In order to address the same problem I noted in my PR a few days ago: os.getcwd() is evaluated when the module is loaded rather than when the function is called.)

Or, better: avoid having a default for these directory arguments, and require them to be passed explicitly. I think that requiring these arguments to be explicit will reduce a possible source of errors moving forward - particularly since I suspect that the current working directory is often _not_ what you want here.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,175 @@
...
+    suffix = (".%s" % suffix) if suffix else ""
+    test_hists = glob.glob("%s/%s.%s*.h?.nc%s" % (from_dir, testcase, model, suffix))
+    test_hists.extend(glob.glob("%s/%s.%s*.h.nc%s" % (from_dir, testcase, model, suffix)))
```
billsacks: Can you add comments describing the types of files that each of these globs are matching? For example, "This glob matches multi-instance files". Without such comments, I'm worried that it will be hard to keep this function up to date as we change / add / delete certain history file naming patterns.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,175 @@
...
+        baselineroot = case.get_value("BASELINE_ROOT")
+        basecmp_dir = os.path.join(baselineroot, case.get_value("BASECMP_CASE"))
+        dirs_to_check = (baselineroot, basecmp_dir)
+    else:
+        basecmp_dir = baseline_dir
+        dirs_to_check = (basecmp_dir,)
+
+    for bdir in dirs_to_check:
+        if not os.path.isdir(bdir):
+            return False, "ERROR %s does not exist" % bdir
```
billsacks: Can this message (which I assume is added to the TestStatus file) be changed to something more uniquely greppable, like: "ERROR %s baseline directory does not exist"? Then I can do a grep -v 'baseline directory does not exist', which I imagine is less likely to occur elsewhere than simply 'does not exist'.

jgfouca:
@billsacks : I've added compare_test_results and included a simple test of it in the test for bless_test_results.

utils/python/CIME/compare_test_results.py
```diff
@@ -0,0 +1,86 @@
+import CIME.compare_namelists, CIME.simple_compare
+from CIME.utils import expect
+from CIME.XML.machines import Machines
```
jgfouca: @billsacks  New functionality here.

utils/python/CIME/hist_utils.py
jgfouca: Good point. Fix incoming.

utils/python/CIME/hist_utils.py
jgfouca: Yes, good idea

utils/python/CIME/hist_utils.py
jgfouca: Pfffft, documentation... hah.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,175 @@
...
+
+    comments = "Generating baselines into '%s'\n" % basegen_dir
+    num_gen = 0
+    for model in _iter_model_file_substrs(case):
+        comments += "  generating for model '%s'\n" % model
+        hists = _get_all_hist_files(testcase, model, from_dir=rundir)
+        num_gen += len(hists)
+        for hist in hists:
+            ext = get_extension(model, hist)
+            basename = "%s.%s.nc" % (model, ext)
```
billsacks: So it looks like this generates baselines for every file, but if you have multiple files (from multiple times) with the same extension, then they just overwrite each other, so you end up with one in the end. Maybe a bit inefficient, but shouldn't be a big deal in practice.

However, this appears to be problematic in the comparison of baselines: If you have, say:

somecase.cpl.hi.0001-01-01.nc
somecase.cpl.hi.0001-01-02.nc

Then the baseline directory will just have:
cpl.hi.nc

and the comparison will fail (I think) because the number of files differs between the test case and the baseline directory.

@jedwards4b noticed something similar, and is working on a fix for this, along with possibly changing the behavior for the in-test comparisons so we just look at the last history file, as has been done in the past.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,175 @@
...
+        dirs_to_check = (baselineroot, basecmp_dir)
+    else:
+        basecmp_dir = baseline_dir
+        dirs_to_check = (basecmp_dir,)
+
+    for bdir in dirs_to_check:
+        if not os.path.isdir(bdir):
+            return False, "ERROR %s does not exist" % bdir
+
+    return _compare_hists(case, from_dir1=rundir, from_dir2=basecmp_dir)
```
billsacks: It looks like you're not using the '.base' file for the baseline comparisons or generation, but rather using the no-suffix files. In the past I think we have used the files with a .base suffix for the generation and comparison. For a test that passed, this will typically be the same as using the file with no suffix. I'm not clear on why we used the .base file as opposed to the no-suffix file; I believe this is something that @mvertens put in; maybe she can comment on whether it's important for the old behavior to be maintained.

billsacks: Typically we _do_ expect a difference between the .base file and the no-suffix file for a test that failed the in-test comparison (e.g. if the restart test failed). However, I'm not sure if we expect the baseline comparison to behave in a certain way for this case.

utils/python/CIME/hist_utils.py
jgfouca: Done.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,217 @@
...
+                os.remove(new_file)
+
+            comments += "    Copying '%s' to '%s'\n" % (test_hist, new_file)
+            shutil.copy(test_hist, new_file)
+
+    expect(num_moved > 0, "move failed: no hist files found in rundir '%s'" % rundir)
+
+    return comments
+
+def _compare_hists(case, suffix1="", suffix2="", from_dir1=os.getcwd(), from_dir2=os.getcwd()):
```
billsacks: @jgfouca : thanks for removing the other os.getcwd() calls from defaults, but there are still these two.

jgfouca: Sorry, must have been something wrong with my grep. They're gone now.

utils/python/CIME/hist_utils.py
jgfouca: Good idea, commit incoming.

billsacks:
The shell script-based component_compare_test had logic specific to the multi-instance tests (NCK and its variants) so that it would compare both the _0001 and the _0002 file from the multi-instance run with the standard file from the non-multi-instance run. I believe that was a critical piece of these tests.

I don't see that functionality replicated in this new version.

I'll be the first to admit that the old way of doing it wasn't ideal - i.e., embedding this test-specific logic in the general component_compare_test. Ideally, we'd have a mechanism for an individual test to specify some custom code for doing in-test comparisons. (And, as an aside, that would address a desire that I and others have expressed for a while: we'd like to be able to make component-specific tests that do something more than just doing a blanket comparison of history files for equality.) But I realize that that might require some significant rework (or maybe not... I'm not sure).

In any case, I believe we should have some mechanism for maintaining the old NCK (etc.) behavior, even if it's via a short-term kludge that we will improve later.

Tagging @jedwards4b and @mvertens on this, as they might have further thoughts.

jgfouca:
@billsacks what if each occurrence of a multi-instance run was copied into the baselines?

So, on the baseline side, we'd have something like:
cpl.hi.0001.nc

billsacks:
> @billsacks what if each occurrence of a multi-instance run was copied into the baselines?
> 
> So, on the baseline side, we'd have something like:
> cpl.hi.0001.nc

That would be good (although note that the cpl hist file is the one for which there are _not_ multiple instances). But I don't think it addresses the in-test comparison that was done for NCK. I'd want an NCK test to pick up whether there are differences between the instances without needing to rely on baseline comparisons for this assurance that you haven't broken something.

billsacks:
@jedwards4b and @mvertens : It also seems like the IOP-specific logic is gone from what used to be component_compare_test.sh. I don't understand what this was accomplishing before, and whether it's important to maintain, so I'll let you comment on whether this is important.

jgfouca:
@billsacks : we don't have any IOP testing in CIME5 yet, there's an issue to address this.

jedwards4b:
The IOP is intended to compare files written using the various output methods supported:
netcdf, pnetcdf, netcdf4p, netcdf4c   I think that it's better to develop a new test that explicitly does this rather than reuse the existing logic.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,219 @@
...
+        if len_hist1 == 0:
+            comments += " no hist files found for model %s\n"%model
+            continue
+        if len_hist1 != len_hist2:
+            comments += "    num hists does not match %d != %d\n" % (len_hist1, len_hist2)
+            all_success = False
+            continue
+
+        num_compared += len(hists1)
+        for hist1, hist2 in zip(hists1, hists2):
```
billsacks: The assumption here seems to be that the sorted list of hists1 lines up with the sorted list of hists2. This seems to be a safe assumption for compare_test, but not always for compare_baseline. For example, if the new version of the code outputs some additional history file, it looks like this function will just report that the number of history files differs.  Previously, I believe the baseline comparison would report on differences for all history files that match up, which I feel is useful behavior, although maybe not critical.

@mvertens and @jedwards4b : do you have feelings on how important it is to maintain the old behavior in this respect?

(There could be further problems if the new version of the code has one new history file and at the same time drops out some other history file - so the total number is equal to the baseline, but the correspondence is messed up. But I'm fine neglecting this corner case for now.)

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,219 @@
...
+            ext = get_extension(model, hist)
+            basename = "%s.%s.nc" % (model, ext)
+            baseline = os.path.join(basegen_dir, basename)
+            if os.path.exists(baseline):
+                os.remove(baseline)
+
+            shutil.copy(hist, baseline)
+            comments += "    generating baseline '%s'\n" % baseline
+
+    return True, comments
```
billsacks: Should this return False if num_gen = 0? I think the old version of the scripts passed even if no hist files were generated, but with this clean rewrite it becomes easy to check how many hist files were generated, and I believe generating 0 files should be flagged as an error.

(If you don't flag the generation of 0 files as an error, then I believe an error will be generated in the later compare step, when it hits `expect(num_compared > 0, ...)` .)

scripts/Tools/component_generate_baseline
```diff
@@ -0,0 +1,55 @@
...
+    args = parser.parse_args(args[1:])
+
+    CIME.utils.handle_standard_logging_options(args)
+
+    return args.caseroot
+
+###############################################################################
+def _main_func(description):
+###############################################################################
+    caseroot, baseline_dir = parse_command_line(sys.argv, description)
```
billsacks: Tries to get caseroot and baseline_dir, but only caseroot is returned from the parse_command_line function

jedwards4b: fixed

utils/python/CIME/hist_utils.py
jedwards4b: Added a new function _get_latest_hist_files which finds only the most recent history file for each model and extension type.   This really reduces the time required for the test and fixes the issue with matching to the baseline files.

scripts/Tools/compare_test_results
```diff
@@ -0,0 +1,89 @@
+#!/usr/bin/env python
+
+"""
+Analyze results from a test root area, comparing non-BFB
+changes. Purpose is, instead of re-running tests
+in generate mode, which is very slow, allow for very fast analsis of diffs.
```
billsacks: "generate mode" should say "compare mode"

utils/python/CIME/hist_utils.py
jgfouca: Actually, to be consistent with compare, it should probably throw an exception given that something must be seriously wrong if there are no hist files present. Thanks for catching this.

scripts/Tools/compare_test_results
```diff
@@ -0,0 +1,89 @@
...
+    default_baseline_name = CIME.utils.get_current_branch(repo=CIME.utils.get_cime_root())
+    default_compiler      = _MACHINE.get_default_compiler()
+    scratch_root          = _MACHINE.get_value("CESMSCRATCHROOT")
+    default_testroot      = os.path.join(scratch_root)
+
+    CIME.utils.setup_standard_logging_options(parser)
+
+    parser.add_argument("-b", "--baseline-name", default=default_baseline_name,
+                        help="Name of baselines to use, corresponds to branch used.")
+
```
billsacks: Also need a way to specify baseline root: At least in my workflow, I often used baseline_gen_comp with a baseline_root that differed from the machine default (because I don't want to clutter up the standard baseline area with all of the one-off baselines that I generate).

scripts/Tools/compare_test_results
jgfouca: Fixed.

scripts/Tools/compare_test_results
```diff
@@ -0,0 +1,89 @@
...
+    \033[1;32m# From most recent run of jenkins, compare history changes for next \033[0m
+    > %s -r /home/jenkins/acme/scratch/jenkins -b next
+""" % ((os.path.basename(args[0]), ) * 6),
+
+description=description,
+
+formatter_class=argparse.ArgumentDefaultsHelpFormatter
+)
+
+    default_baseline_name = CIME.utils.get_current_branch(repo=CIME.utils.get_cime_root())
```
billsacks: For the sake of comparison, this doesn't make sense as a default. I suggest making --baseline-name be a required argument.

scripts/Tools/compare_test_results
```diff
@@ -0,0 +1,89 @@
...
+    default_compiler      = _MACHINE.get_default_compiler()
+    scratch_root          = _MACHINE.get_value("CESMSCRATCHROOT")
+    default_testroot      = os.path.join(scratch_root)
+
+    CIME.utils.setup_standard_logging_options(parser)
+
+    parser.add_argument("-b", "--baseline-name", default=default_baseline_name,
+                        help="Name of baselines to use, corresponds to branch used.")
+
+    parser.add_argument("-c", "--compiler", default=default_compiler,
```
billsacks: I don't understand why the compiler is needed here. Isn't the compiler part of the test name? And can't you run a test suite that has multiple compilers?

If a compiler truly is needed, then I'd suggest making this a required argument: Typically we run with multiple compilers on a given machine, and making this an optional argument could easily make someone think that it doesn't need to be specified and that it goes through ALL compilers for the given test id... at least, that's what I assumed at first.

utils/python/CIME/compare_test_results.py
```diff
@@ -0,0 +1,86 @@
...
+from CIME.test_status import *
+from CIME.hist_utils import compare_baseline
+from CIME.case import Case
+
+import os, glob
+
+_MACHINE = Machines()
```
billsacks: Why is this needed?

utils/python/CIME/compare_test_results.py
```diff
@@ -0,0 +1,86 @@
...
+        if result:
+            return True, None
+        else:
+            logging.info(comments)
+            return False, "Diff'd"
+
+###############################################################################
+def compare_test_results(baseline_name, test_root, compiler, test_id=None, compare_tests=None):
+###############################################################################
+    test_id_glob = "*%s*%s*" % (compiler, baseline_name) if test_id is None else "*%s" % test_id
```
billsacks: I don't understand what you're doing here if test_id is None. In practice, though, it seems that test_id will always be provided; maybe just make this a required argument?

scripts/Tools/compare_test_results
```diff
@@ -0,0 +1,89 @@
...
+                        help="Compiler of run you want to compare")
+
+    parser.add_argument("-r", "--test-root", default=default_testroot,
+                        help="Path to test results that are being compared")
+
+    parser.add_argument("-t", "--test-id",
+                        help="Limit processes to case dirs matching this test-id. Can be useful if mutiple runs dumped into the same dir.")
+
+    parser.add_argument("compare_tests", nargs="*",
```
billsacks: This is a nice feature - thanks

utils/python/CIME/compare_test_results.py
```diff
@@ -0,0 +1,86 @@
...
+
+    broken_compares = []
+    for test_status_file in test_status_files:
+        ts = TestStatus(test_dir=os.path.dirname(test_status_file))
+        test_name = ts.get_name()
+        if (compare_tests in [[], None] or CIME.utils.match_any(test_name, compare_tests)):
+            overall_result = ts.get_overall_test_status()
+
+            # Compute hist status, False implies it diffed
+            run_result = ts.get_status(RUN_PHASE)
```
billsacks: As an aside: This is another place where having the TestStatus class responsible for determining if a run is incomplete (as evidenced by not having a RUN phase) would help simplify some client code.

scripts/Tools/compare_test_results
jgfouca: OK, i'll add that to both compare and bless test results.

utils/python/CIME/hist_utils.py
```diff
@@ -0,0 +1,254 @@
...
+        comments += "  comparing model '%s'\n" % model
+        hists1 = _get_latest_hist_files(testcase, model, from_dir1, suffix1)
+        hists2 = _get_latest_hist_files(testcase, model, from_dir2, suffix2)
+        len_hist1 = len(hists1)
+        len_hist2 = len(hists2)
+        if len_hist1 == 0:
+            comments += " no hist files found for model %s\n"%model
+            continue
+        if len_hist1 != len_hist2:
+            comments += "    num hists does not match %d != %d\n" % (len_hist1, len_hist2)
```
gold2718: I also am uncomfortable with this logic. In my tree compares, I explicitly look for omissions from either side. Thus, both len(set(hists1) - set(hists2)) and len(set(hists2) - set(hists1)) must be zero. If either is not zero then, the difference value (e.g., set(hists1) - set(hists2)) is the set of missing files (in one direction).

utils/python/CIME/compare_test_results.py
```diff
@@ -0,0 +1,86 @@
...
+                # Compare hist files
+                if (not hist_no_compare):
+                    success, reason = compare_history(testcase_dir_for_test, baseline_name)
+                    if (not success):
+                        broken_compares.append((test_name, reason))
+
+    # Make sure user knows that some tests were not compareed
+    success = True
+    for broken_compare, reason in broken_compares:
+        logging.warning("COMPARE FAILED FOR TEST: %s, reason %s" % (broken_compare, reason))
```
billsacks: Ideally I would want the output from this tool to look basically the same as the output put in the TestStatus file - although just for the COMPARE_baseline phase. That is, I'd want to see something like:

PASS   SOMETEST  BASELINE_COMPARE

or

FAIL   SOMETEST  BASELINE_COMPARE

That's approximately what happened with the earlier version of these tools.

This comment applies both to the main output from the comparison, and also to various other extraneous information that is printed (or not). So, for example, I'd prefer not to have this output for every test:

```python
                logging.info("Comparing results for test: %s, most recent result: %s" % (test_name, overall_result))
```

- because I'd like to be able to parse this output in the same way that I parse the normal TestStatus or cs.status output. 

However, I agree that it does make sense to report on failed tests here. One possibility would be to print out the testStatus output for each test that failed.

Actually, maybe you could accomplish all of this cleanly by making more use of the TestStatus class here: Rather than having this function be responsible for output, it could instead just update the ts object with the results of the BASELINE_COMPARE phase (I know, there isn't a separate BASELINE_COMPARE phase yet... but the idea still applies). Then you can output the updated TestStatus results.

One complication with this is that I don't think you want to change the TestStatus file in the test itself (it seems like a Bad Idea to have some external script change the TestStatus file). In this usage, we instead would want the TestStatus to be written for each test via the logger. I think this could be accomplished by adding a __str__ method to TestStatus. This would contain the bulk of what is currently TestStatus.flush, and then TestStatus.flush could simply write out the results of str(self).

utils/python/CIME/compare_test_results.py
```diff
@@ -0,0 +1,86 @@
...
+            else:
+
+                logging.info("###############################################################################")
+                logging.info("Comparing results for test: %s, most recent result: %s" % (test_name, overall_result))
+                logging.info("###############################################################################")
+
+                # Get testcase dir for this test
+                if (test_id is None):
+                    # The full name already contains the compiler, so we just need to glob for the branch name
+                    globs = glob.glob("%s/%s*%s*" % (test_root, test_name, baseline_name))
```
billsacks: As with my above comment (on line 27), I don't understand what you're doing here. Is this possibly a carry-over from the baseline generation logic (for which I could more easily convince myself that this is right)? Again, I wonder if it's really needed to handle test_id None?

utils/python/CIME/compare_test_results.py
```diff
@@ -0,0 +1,86 @@
...
+            # Now, do the compare
+            if hist_no_compare:
+                logging.info("Cannot compare test: %s, overall status: %s" % (test_name, overall_result))
+            else:
+
+                logging.info("###############################################################################")
+                logging.info("Comparing results for test: %s, most recent result: %s" % (test_name, overall_result))
+                logging.info("###############################################################################")
+
+                # Get testcase dir for this test
```
billsacks: I don't understand why you need to re-do a similar glob to the one you do near the beginning of this function. It seems like you get an initial list of tests via this glob:

```python
    test_id_glob = "*%s*%s*" % (compiler, baseline_name) if test_id is None else "*%s" % test_id
    test_status_files = glob.glob("%s/%s/%s" % (test_root, test_id_glob, TEST_STATUS_FILENAME))
```

then later you find the test_dir with:

```python
        ts = TestStatus(test_dir=os.path.dirname(test_status_file))
```

Couldn't you simply save that test_dir - i.e., changing the above line to:

```python
        test_dir = os.path.dirname(test_status_file)
        ts = TestStatus(test_dir=test_dir)
```

and then use test_dir in place of testcase_dir_for_test?

scripts/Tools/compare_test_results
jgfouca: Done.

scripts/Tools/compare_test_results
jgfouca: It's not used for much. It allows the user to specify a baseline_name + compiler as a way of selecting tests instead of the test_id.

scripts/Tools/compare_test_results
jgfouca: The default baseline dir is: baseline_root/compiler/branch_name, so it's not unreasonable to use the current branch as the default.

utils/python/CIME/compare_test_results.py
jgfouca: Not needed. copy/paste error.

utils/python/CIME/compare_test_results.py
jgfouca: This had to do with identifying a set of tests via test_id or baseline_name + compiler that I mentioned earlier. It is convenient for some scripts not to have to know the test_id.

utils/python/CIME/hist_utils.py
jgfouca: Can you elaborate? The sets of files will not match in many cases. For example, if suffix1 != suffx2, then it would be impossible for the sets of hist files to be identical unless you made some effort to normalize the names.

utils/python/CIME/compare_test_results.py
jgfouca: There actually is a BASELINE_COMPARE phase, it would be called "COMPARE_baseline" though.

I agree we do not want this script to change TestStatus files. Leveraging the TestStatus class for output is an interesting idea but I'm starting to get a bit overwhelmed by the number of changes being requested. Is there anyway we could punt this thought to a later ticket/PR?

utils/python/CIME/hist_utils.py
billsacks: @jedwards4b : Sounds great! However, I don't think this addresses my original comment on this line.... I realize I wasn't very clear in that original comment: I was really referring to the fact that the new version of the code could output a file with some new history extension - e.g., now we have a .h2 file where we had none before. In the old scripts, I believe the .h0 and .h1 files would still be compared properly; in this new code, you'll just get a message saying that the number of history files differs.

utils/python/CIME/compare_test_results.py
jgfouca: Yeah, this could be much simpler. Pushing fix now, thanks

scripts/Tools/compare_test_results
billsacks: Replying to a few related comments here:

> It's not used for much. It allows the user to specify a baseline_name + compiler as a way of selecting tests instead of the test_id.

> The default baseline dir is: baseline_root/compiler/branch_name, so it's not unreasonable to use the current branch as the default.

> This had to do with identifying a set of tests via test_id or baseline_name + compiler that I mentioned earlier. It is convenient for some scripts not to have to know the test_id.

Ah, perhaps this is a difference in the ACME vs. CESM workflows??? For CESM, the baseline directory is typically a frozen tag, not the name of an evolving branch. If these defaults fit in with the ACME workflow, I'm okay with them being kept as is. Maybe the real need then is for explicit documentation of the expected usage with the CESM workflow... I _think_ that, for the CESM workflow, you would specify -b and -t but not -c; do you agree? Also, it would help to document that the --compiler option is only used if --test-id is not specified (I think).

utils/python/CIME/compare_test_results.py
```diff
@@ -0,0 +1,86 @@
...
+
+import os, glob
+
+_MACHINE = Machines()
+
+###############################################################################
+def compare_history(testcase_dir_for_test, baseline_name):
+###############################################################################
+    with Case(testcase_dir_for_test) as case:
+        baseline_full_dir = os.path.join(case.get_value("BASELINE_ROOT"), case.get_value("COMPILER"), baseline_name, case.get_value("CASEBASEID"))
```
billsacks: In the CESM workflow, the COMPILER doesn't enter in to the baseline path. At least, it didn't with cime4, and still doesn't seem to in the latest baselines (cesm2_0_alpha01e, which uses some version of cime5).

utils/python/CIME/compare_test_results.py
billsacks: Yes, I'm okay with that. Do you mean punting on the whole reworking of the output format? (I'm okay with that; I just want to make sure I open an appropriate ticket.)

scripts/Tools/compare_test_results
jgfouca: Yes, that's exactly it. ACME maintains baselines for branches, not tags.

You're correct, to do things the CESM way, -b and -t would be adequate.

utils/python/CIME/compare_test_results.py
jgfouca: Would it be OK if the CESM workflow was modified in this way? It does make it much easier to maintain multiple sets of baselines for multiple compilers.

utils/python/CIME/compare_test_results.py
jgfouca: Yes, reworking the output of compare/bless test results and changing TestStatus class to be able to write to logging.

utils/python/CIME/compare_test_results.py
billsacks: I can't answer that myself. It would affect many people in CSEG, so I imagine needs some broader buy-in. Personally, I can't think of any major objections, but I also don't see the advantage, since our test names have the compiler in them. So it seems to work just fine to have a baseline directory that contains, say, SMS.f10_f10.ICLM45.yellowstone_intel, SMS.f10_f10.ICLM45.yellowstone_pgi and SMS.f10_f10.ICLM45.yellowstone_gnu.

I'll defer to @mvertens , @jedwards4b and other CSEG members on this question.

I do see that there are other places in the scripts that assume the ACME organization in this respect: at least bless_test_results.py and maybe test_scheduler.py and scripts_regression_tests.py (I just did a simple grep for grep -i 'path.join.*compiler').

scripts/Tools/compare_test_results
```diff
@@ -0,0 +1,93 @@
...
+
+import argparse, sys, os, doctest
+
+_MACHINE = Machines()
+
+###############################################################################
+def parse_command_line(args, description):
+###############################################################################
+    parser = argparse.ArgumentParser(
+usage="""\n%s [-n] [-r <TESTROOT>] [-b <BRANCH>] [-c <COMPILER>] [<TEST> <TEST> ...] [--verbose]
```
billsacks: This usage message is wrong: it's missing -t, and I don't see any documentation / meaning of the -n option

scripts/Tools/compare_test_results
```diff
@@ -0,0 +1,93 @@
...
+OR
+%s --help
+OR
+%s --test
+
+\033[1mEXAMPLES:\033[0m
+    \033[1;32m# From most recent run, compare all changes \033[0m
+    > %s
+    \033[1;32m# From most recent run, compare only changes for test foo and bar only \033[0m
+    > %s -n foo bar
```
billsacks: Again, it looks like the -n option doesn't exist

scripts/Tools/compare_test_results
billsacks: Okay, thanks for the clarification. I have pushed a commit where I added a usage example for the typical CESM workflow.

utils/python/CIME/hist_utils.py
gold2718: Then what is the purpose of this check? I don't see how just having a length match tests anything except to avoid a zip truncate. Since you know the suffix for each set, you can create the sets without the suffixes and compare them. That way we will know if the filenames don't match up.

utils/python/CIME/compare_test_results.py
jgfouca: Looking at my current baselines, I see directory names like "ERS.f09_g16.I1850CLM45CN.melvin_gnu", so the compiler is clearly included in the name and therefore runs with different compilers would not conflict; so maybe adding compiler to the path is not necessary. This was a change we made in ACME a long time ago back when we were working with an ancient version of CIME, so it's possible it's not necessary anymore. For now, I can do an if/else based on model here and in bless_test_results. I don't think anywhere else is impacted.

jgfouca: Should have read your post more closely. Since this path system is built into test_scheduler, then CESM must be using it already. I guess they like it since I haven't seen any complaints ;)

scripts/Tools/compare_test_results
jgfouca: You're right, fix incoming.

scripts/Tools/compare_test_results
jgfouca: You're right, fix incoming.

utils/python/CIME/hist_utils.py
jgfouca: @gold2718 , that's a good point, it might be more robust to do it that way.

jgfouca: @gold2718 Fix incoming

jgfouca:
I believe all concerns have been addressed. I'm going to merge.

gold2718:
Are we merging our own pull requests now?

jgfouca:
@gold2718 Once the OK has been given by the reviewers, does it matter who clicks the button?

jgfouca:
@gold2718 To clarify, there's no such thing as an integrator role in CIME like there is in ACME.

gold2718:
Understood, it's just that there was no time to review your significant change to the compare code before the merge.

jgfouca:
Ah, sorry about that. You can click here to see the changes: https://github.com/ESMCI/cime/pull/413/commits/164cfd912e49fb133e2cbb35579c262e1edf4bf5

I'll be sure to implement whatever changes you'd like in another PR.

PR #415, opened 2016-08-17
Introducing quotes around timing event names

worleyph:
As it is legal for white spaces to appear in timing event names, but
parsing the timing data is complicated by this, modified t_startf and
t_stopf in perf_mod.F90 to prepend and append double quotes to each
timing event name.

The getTiming script looks for particular timing events in order to
calculate the performance summary. These event names now are quoted,
and the script was modified to look for the modified names.

Test suite: none yet
Test baseline: none yet
Test namelist changes: none
Test status: bit for bit (but changes format of timing files)

Connects to #2

User interface changes?: 

Code review:

### DIFF
--- a/scripts/Tools/getTiming
+++ b/scripts/Tools/getTiming
@@ -212,7 +212,7 @@ class TimingParser:
         else:
             logger.warning("Unknown NCPL_BASE_PERIOD=%s" % ncpl_base_period)
 
-        nprocs, ncount = self.gettime2('CPL:CLOCK_ADVANCE ')
+        nprocs, ncount = self.gettime2('"CPL:CLOCK_ADVANCE" ')
         nsteps = ncount / nprocs
         adays = nsteps*tlen/atm_ncpl
         odays = adays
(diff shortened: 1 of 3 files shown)


worleyph:
The "idea" was tested in ACME. The timing source in CIME and ACME differ by quite a lot - primarily white space cleanup in the CIME version - so I had to reimplement the changes in this branch instead of just copying over the new version of perf_mod.F90. The number of changes is small and localized however.

SImilarly, I tried to implement the same changes that I made to the ACME version of getTiming in the CIME version. I made the "same" changes, but getTiming is enough different in CIME that I do not know if I made all of the changes that are necessary. In any case, none of this is tested in CIME.

@quantheory may also have comments about whether this PR does what he wants well enough to bring it into CIME now. Consider this a prototype as the basis of additional discussion.

quantheory:
Between being out sick and being on my way out of NCAR again, I haven't had the chance to think much about this, but I think that this change is OK. One minor change I would suggest is to have the `gettime` and `gettime2` functions add the quotation marks themselves, rather than adding quotes to every string used as an argument to them.

worleyph:
> rather than adding quotes to every string used as an argument to them.

SInce I did not even do that correctly, sounds like an excellent solution. It is also one that I do not have a clue as how to implement. If someone wants to hijack this PR, or recreate a new one that incorporates the bits that I did correctly, please do so. There is no particular value to this PR. As @rljacob suggested, this was just for discussion purposes.

quantheory:
Closing in favor of #467, which is basically the same thing except for the change I suggested above.

PR #430, opened 2016-08-19
Pea test fix

### LINKED ISSUE #423: PEA_P1 fails on yellowstone
This is due to differences in the contents of the Macros file between a default mpi library build and a mpi-serial build.   Explicitly removing the file in the test allows case.setup to update the file for each build.    I believe that issue 421 may also fix this problem.

jedwards4b:
Rewrite of the pea test to use the system_tests_compare_two framework.  This includes extending that framework to do two build. 
Test suite: scripts_regression_tests , PEA_P1_Ld3.f45_g37_rx1.A.yellowstone_intel
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #423 

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/config_tests.xml
+++ b/cime_config/config_tests.xml
@@ -348,9 +348,12 @@ LII    CLM initial condition interpolation test
   <test NAME="PEA">
     <DESC>single pe bfb test (default length)</DESC>
     <INFO_DBUG>1</INFO_DBUG>
-    <REST_OPTION>never</REST_OPTION>
     <CCSM_TCOST>1</CCSM_TCOST>
     <DOUT_S>FALSE</DOUT_S>
+    <CONTINUE_RUN>FALSE</CONTINUE_RUN>
+    <REST_OPTION>never</REST_OPTION>
+    <HIST_OPTION>$STOP_OPTION</HIST_OPTION>
+    <HIST_N>$STOP_N</HIST_N>
   </test>
 
   <test NAME="PEM">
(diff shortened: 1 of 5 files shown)


billsacks:
Thanks a lot for taking this next step with the clone-based test infrastructure - I really appreciate your wrapping this up.

This basically looks great to me. I have a few minor comments. If you'd like, I can make these changes myself, though I'm not sure what the preferred way is to do this, since it's on your fork.

1. We need to set HIST_OPTION="$STOP_OPTION", HIST_N="$STOP_N" and CONTINUE_RUN="FALSE" in config_tests.xml. (In principle you could set these in _common_setup, but I've been trying to move these to config_tests.xml since that seemed to be the direction others were moving.)

2. I'm having trouble seeing a reason why an individual test would want to override _case_one_build or _case_two_build: It seems like all test-specific setup can be done in _case_one_setup and _case_two_setup (just as there are no hooks for _case_one_run or _case_two_run in this infrastructure). If you agree, then I'd suggest that we:
a. Inline these methods: Rather than calling _case_one_build or _case_two_build, simply call `self.build_indv(sharedlib_only=sharedlib_only, model_only=model_only)` in those two places
b. Remove them from the top-level API documentation

Again, I'm happy to make these changes myself if you'd prefer; if so, please let me know the easiest way I can do this.

jedwards4b:
Updated based on reviewer feedback

PR #450, opened 2016-08-24
refuse to overwrite existing baseline directory in cesm workflow

jedwards4b:
Cowardly refuse to overwrite existing baseline directory in cesm workflow.

Test suite: scripts_regression_tests, Tests using --generate with and without existing baseline.
Test baseline: 
Test namelist changes: 

Fixes 

User interface changes?: 

Code review:


jgfouca:
ACME typically uses bless_for_tests to update, so this shouldn't be a problem for us. Can you add "-o" as the short version of that flag?

PR #458, opened 2016-08-25
Return to the original directory after unit tests

billsacks:
This is needed because some of the SystemTestsCompareTwo tests trigger
code that includes an os.chdir call. Because these chdirs happen into a
temporary directory that is later removed, this was calling os.getcwd()
calls in later tests to fail.

Test suite: scripts_regression_tests.py A_RunUnitTests
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

In doing these tests, I also added a print of os.getcwd() in the
tearDown method. This showed that, prior to this fix, os.getcwd() failed
after running test_unittests; after this fix, os.getcwd() shows that we
are back in the original directory after running test_unittests.

Fixes: None

User interface changes?: No

Code review: @quantheory

### DIFF
--- a/utils/python/CIME/tests/SystemTests/test_system_tests_compare_two.py
+++ b/utils/python/CIME/tests/SystemTests/test_system_tests_compare_two.py
@@ -227,10 +227,15 @@ class SystemTestsCompareTwoFake(SystemTestsCompareTwo):
 class TestSystemTestsCompareTwo(unittest.TestCase):
 
     def setUp(self):
+        self.original_wd = os.getcwd()
         # create a sandbox in which case directories can be created
         self.tempdir = tempfile.mkdtemp()
 
     def tearDown(self):
+        # Some tests trigger a chdir call in the SUT; make sure we return to the
+        # original directory at the end of the test
+        os.chdir(self.original_wd)
+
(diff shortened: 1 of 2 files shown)


quantheory:
Looks good to me.

jgfouca:
@billsacks is there any way to not have to do a chdir in the first place? That will affect the global state of the process and could lead to really difficult bugs like the ones I was struggling with today.

billsacks:
@jgfouca : I tried to avoid doing a chdir in SystemTestsCompareTwo, but this caused the clone test to fail. If I remember correctly, the case2 timing file was put in the case1 case directory, causing a failure when the test tried to examine the timing file. This suggested to me that there were some uses of os.getcwd() that were assuming that the current case directory was given by the current working directory. I never dug into this deeply.

I agree with you that it would be good to avoid doing chdir calls, but I think that for that to work, we also need to remove most / all getcwd calls.

(The unit tests only end up doing a chdir indirectly - i.e., because the code under test does a chdir. In principle we could avoid doing the chdir in the unit tests through appropriate mocking, but that feels like it could add unnecessary complexity, and I'm not sure it addresses your concern.)

jgfouca:
@billsacks lets talk on slack

PR #459, opened 2016-08-25
Jgfouca/remove compiler in baseline dir

### LINKED ISSUE #451: bless_test_results and compare_test_results assume ACME's baseline directory organization
bless_test_results and compare_test_results assume that you're using ACME's organization of the baseline directories, whereby there is a COMPILER subdirectory of the given baseline directory. CESM does not use this directory organization. This assumption appears on this line in bless_test_results.py and compare_test_results.py:

```python
        baseline_full_dir = os.path.join(baseline_root, case.get_value("COMPILER"), baseline_name, case.get_value("CASEBASEID"))
```

@jgfouca recently commented by email that the need for this extra COMPILER level in the directory structure may no longer exist. If so, it seems like it could simplify the scripts if this were removed. But if there is a reason to maintain it in the ACME workflow, then bless_test_results and compare_test_results need to be generalized to handle a directory structure that doesn't contain the compiler.

jgfouca:
Also:
1) Pull interface handling out of test_scheduler and put into create_test
2) acme/cime models now have different interfaces to create_test

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes  #451

User interface changes?: create_test, minor

Code review: jedwards billsacks

### DIFF
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -306,6 +306,8 @@ OR
                 baseline_gen_name = args.generate
         else:
             baseline_name = args.baseline_name if args.baseline_name else CIME.utils.get_current_branch(repo=CIME.utils.get_cime_root())
+            if not baseline_name.startswith("%s/" % args.compiler): # pylint: disable=maybe-no-member
+                baseline_name = os.path.join(args.compiler, baseline_name)
             expect(baseline_name is not None,
                    "Could not determine baseline name from branch, please use -b option")
             if args.compare:
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -23,10 +23,26 @@ logger = logging.getLogger(__name__)
 ###############################################################################
 def parse_command_line(args, description):
 ###############################################################################
-    parser = argparse.ArgumentParser(
-        usage="""\n%s <TEST|SUITE> [<TEST|SUITE> ...] [--verbose]
-OR
+    model = CIME.utils.get_model()
+
+    if model == "cesm":
+        help_str = \
+"""
 %s --xml-category [CATEGORY] [--xml-machine ...] [--xml-compiler ...] [ --xml-testlist ...]
+OR
+%s --help
+OR
+%s --test
+
+\033[1mEXAMPLES:\033[0m
+    \033[1;32m# Run all tests in the xml prealpha category and yellowstone machine \033[0m
+    > %s --xml-machine yellowstone --xml-category prealpha
+""" % ((os.path.basename(args[0]), ) * 4)
+
+    else:
+        help_str = \
+"""
+%s <TEST|SUITE> [<TEST|SUITE> ...] [--verbose]
 OR
 %s --help
 OR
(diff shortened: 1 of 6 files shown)


scripts/create_test
```diff
@@ -159,44 +193,43 @@ OR
...
 
     CIME.utils.handle_standard_logging_options(args)
 
     # generate and compare flags may not point to the same directory
-    if args.generate is not None:
-        expect(not (args.generate == args.compare),
-               "Cannot generate and compare baselines at the same time")
-    expect(not (args.baseline_name is not None and (not args.compare and not args.generate)),
-           "Provided baseline name but did not specify compare or generate")
+    if model == "cesm":
```
jgfouca: CESM/ACME differences are handled here in order to cut down complexity in TestScheduler.

utils/python/CIME/utils.py
```diff
@@ -542,7 +542,6 @@ def get_project(machobj=None):
 
 def setup_standard_logging_options(parser):
     helpfile = "%s.log"%sys.argv[0]
-    helpfile = os.path.join(os.getcwd(),os.path.basename(helpfile))
```
jgfouca: For reasons that are utterly baffling to me, this os.getcwd() was causing an OSError (both on master and this branch).

rljacob: See PR #458

scripts/create_test
```diff
@@ -104,32 +111,59 @@ OR
...
+                            "Default is to match the name of the machine in the test name or "
+                            "the name of the machine this script is run on to the "
+                            "NODENAME_REGEX field in config_machines.xml")
+
+    else:
 
-    parser.add_argument("-g", "--generate", const=True, nargs="?",
-                        help="While testing, generate baselines,"
-                        " optionally provide a generate directory")
+        parser.add_argument("testargs", nargs="+",
```
jedwards4b: oh - wait - we use the testargs form extensively in cesm, we don't want to loose that.

scripts/create_test
```diff
@@ -104,32 +111,59 @@ OR
...
+                            help="Tests or test suites to run."
+                            " Testnames expect in form CASE.GRID.COMPSET[.MACHINE_COMPILER]")
 
-    parser.add_argument("-b", "--baseline-name",
-                        help="If comparing or generating baselines with default paths,"
-                        " use this directory under baseline root. "
-                        "Default will be current branch name. Do NOT add the compiler to"
-                        " this argument that will be done for you.  If you provide directories"
-                        " in the generate and or compare argument do not use this option.")
+        parser.add_argument("-b", "--baseline-name",
```
jedwards4b: we also use the --baseline-name option

scripts/create_test
```diff
@@ -159,44 +193,43 @@ OR
...
-                        help="Use this compiler key in the lookup in testlist.xml, default is all if any --xml- argument is used")
-    parser.add_argument("--xml-category",
-                        help="Use this category key in the lookup in testlist.xml, default is all if any --xml- argument is used")
-    parser.add_argument("--xml-testlist",
-                        help="Use this testlist to lookup tests, default specified in config_files.xml")
-    parser.add_argument("--testfile",
+    parser.add_argument("-f", "--testfile",
                         help="A file containing an ascii list of tests to run")
 
     parser.add_argument("-o", "--allow-baseline-overwrite", action="store_true",
```
jedwards4b: I thought that you were moving this flag to the cesm only side?

scripts/create_test
jgfouca: You guys still have it, the difference is that it's a "*" for you guys (optional)

scripts/create_test
jgfouca: Why would it be needed if -c and -g have to provide the name?

scripts/create_test
jgfouca: I think ACME can benefit from it as well.

utils/python/CIME/utils.py
jedwards4b: I think that it was fixed with PR 458

scripts/create_test
jedwards4b: got it

scripts/create_test
jedwards4b: yes, you're right - nevermind

utils/python/CIME/utils.py
jgfouca: OK, I'll add this back.

jedwards4b:
Review complete - do you want to remove the change in utils.py before we merge?  It's actually incorrect as is because the log is written to the current working directory and not the location of sys.argv[0]

jgfouca:
Done, merging.

billsacks:
@jgfouca : I just got a chance to look at this, quickly. It looks good to me. I like the cleanup you were able to do by "normalizing" the inputs up front. It does feel kind of funny to me to have different usages depending on the model, although I don't see a good way around that right now. (Though I can't help but notice that you use the cesm workflow only if the model matches 'cesm', and the acme workflow in all other cases, implicitly assuming that all other models will use the acme workflow ;-)  .)

jgfouca:
@billsacks only CESM would be crazy enough to use that workflow :)

PR #460, opened 2016-08-25
Convert `configure` to Python.

### LINKED ISSUE #299: Problem with cime5 configure when building other tools.
configure in cime/tools/configure fails in cime5 due to a missing perl module?

Can't locate Module/ModuleLoader.pm in @INC (@INC contains: /glade/scratch/erik/cesm2_0_alpha01/cime/utils/perl5lib /glade/scratch/erik/cesm2_0_alpha01/cime/utils/perl5lib/Config /usr/local/lib64/perl5 /usr/local/share/perl5 /usr/lib64/perl5/vendor_perl /usr/share/perl5/vendor_perl /usr/lib64/perl5 /usr/share/perl5 .) at /glade/scratch/erik/cesm2_0_alpha01/cime/tools/configure line 146.

quantheory:
This change set passes `scripts_regression_tests`, and I was able to use `configure` with CESM's `config_build.xml` to build `cprnc`.

Fixes #299.

### DIFF
--- a/tools/configure
+++ b/tools/configure
@@ -32,9 +32,8 @@ The pieces of information that will be written include:
     parser = argparse.ArgumentParser(description=description)
     CIME.utils.setup_standard_logging_options(parser)
 
-    parser.add_argument("--machine", required=True,
-                        help="(required) The machine to create build "
-                        "information for.")
+    parser.add_argument("--machine",
+                        help="The machine to create build information for.")
     parser.add_argument("--machines-dir",
                         help="The machines directory to take build information "
                         "from. Overrides the CIME_MODEL environment variable, "
--- a/tools/configure
+++ b/tools/configure
@@ -1,291 +1,147 @@
-#!/usr/bin/env perl
-#-----------------------------------------------------------------------------------------------
-#
-# configure
-#
-# This utility allows the CIME utilities user to specify configuration
-# options via a commandline interface.
-#
-#-----------------------------------------------------------------------------------------------
+#!/usr/bin/env python
 
-use strict;
-#use warnings;
-#use diagnostics;
+import os
+import shutil
+import sys
 
-use Cwd qw( getcwd abs_path chdir);
-use English;
-use Getopt::Long;
-use IO::File;
-use IO::Handle;
-use File::Copy;
+_CIMEROOT = os.environ.get("CIMEROOT")
+if _CIMEROOT is None:
+    raise SystemExit("ERROR: must set CIMEROOT environment variable")
 
+_LIBDIR = os.path.join(_CIMEROOT, "scripts", "Tools")
+sys.path.append(_LIBDIR)
 
-# Check for manditory case input if not just listing valid values
-my $model;
-my $mach;
-my $compiler;
-my $mpilib;
-my $output_format='make';
-my $cimeroot;
-my $machdir;
-my $eol="\n";
-my $layout = '%m%n';
+from standard_script_setup import *
+from CIME.macros import MacroMaker
+from CIME.utils import expect
+from CIME.XML.env_mach_specific import EnvMachSpecific
+from CIME.XML.machines import Machines
 
+logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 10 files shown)


tools/configure
```diff
@@ -1,291 +1,167 @@
...
-SetupTools::set_compiler($config->get('OS'),"$machdir/$compiler_file",$compiler,
-			 $mach, $mpilib, "$output_dir/Macros.$output_format", $output_format);
-
-$logger->info("Successfully created auxilary build files for $mach $eol");
-
-exit 0;
-
-
+#!/usr/bin/env python
+
```
jgfouca: Please put description here.

tools/configure
```diff
@@ -1,291 +1,167 @@
...
+from CIME.XML.env_mach_specific import EnvMachSpecific
+from CIME.XML.machines import Machines
+
+logger = logging.getLogger(__name__)
+
+def parse_command_line(args):
+    """Command line argument parser for configure."""
+    description = """This script writes CIME build information to a directory.
+
+The pieces of information that will be written include:
```
jgfouca: Use __doc__ to refer to description above.

tools/configure
```diff
@@ -1,291 +1,167 @@
...
+        debug = os.environ["DEBUG"].lower() == "true"
+    else:
+        debug = False
+        os.environ["DEBUG"] = "FALSE"
+    opts['debug'] = debug
+
+    return opts
+
+def _main():
+    opts = parse_command_line(sys.argv)
```
jgfouca: Consider making this functionality a library so that others can access without having to issue shell command.

utils/python/CIME/XML/build.py
```diff
@@ -1,9 +1,9 @@
 """
```
jgfouca: Why did this file get moved to the XML directory?

quantheory: This was requested by @jedwards4b, based on the fact that this is the file responsible for reading `config_build.xml`.

jgfouca: Oh, I didn't see that. Can you point me to the line where that's happening?

rljacob:
Does this fix issue #299 ?

utils/python/CIME/XML/build.py
quantheory: The file is the `xml_file` argument to `write_macros` (line 695 forward). On lines 719-723 we take each `compiler` element and deal with it. Most of the processing of the XML file is in the `CompilerBlock` class (lines 486-648). None of the rest of the file uses `ElementTree` or does any XML processing at all.

quantheory:
@rljacob Yes.

@jgfouca Better?

tools/configure
```diff
@@ -1,291 +1,176 @@
...
+from CIME.utils import expect
+from CIME.XML.build import Build
+from CIME.XML.env_mach_specific import EnvMachSpecific
+from CIME.XML.machines import Machines
+
+logger = logging.getLogger(__name__)
+
+def parse_command_line(args):
+    """Command line argument parser for configure."""
+    description = __doc__
```
jgfouca: Did you test this? I think that __doc__ refers to the docstring of the current context, which would be the parse_command_line method in this case. If you look at some of the other tools, you'll see that I typically get a handle to __doc__ in the global context and then pass it into parse_command_line.

quantheory: I did test this. The `__doc__` here does indeed refer to the global definition. The docstring for `parse_command_line` is stored as an attribute of the function, i.e. as `parse_command_line.__doc__`, so it is *not* a local variable. TLDR; there's no name clash here, everything is fine. :smiley:

quantheory: Might be worth noting, however, that `argparse` throws out the newlines in the docstring and does its own word-wrapping. The resulting help string is still legible, but kind of weird looking.

jgfouca: OK, great. That's good to know.

jgfouca:
I'm OK to merge if everyone else is.

PR #461, opened 2016-08-25
Clean up some help text

billsacks:
This cleans up the help text for --allow-baseline-overwrite, as I discussed via email with @jedwards4b . I also moved the text added by @jgfouca regarding bless_test_results into the help for the --generate option, where it seems more appropriate.

The current help text for --allow-baseline-overwrite assumes that this is needed in both the acme and cesm workflows, which I believe is now the case. @jgfouca , is this correct?

Test suite: just ran create_test -h with CIME_MODEL=cesm and acme
Test baseline: N/A
Test namelist changes: N/A
Test status: N/A

Fixes: None

User interface changes?: No

Code review: @jedwards4b

### DIFF
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -117,8 +117,9 @@ OR
                             "  against the given compare directory ")
 
         parser.add_argument("-g", "--generate",
-                            help="While testing, generate baselines,"
-                            " to the given generate directory")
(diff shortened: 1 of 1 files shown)


jgfouca:
Yes, that's correct.

PR #465, opened 2016-08-26
Fix multiinstance issues

### LINKED ISSUE #309: Some desired hist file comparisons are not done for _N2 tests; also, confusing results in .base.cprnc.out
This issue does NOT appear to be new to the ESMCI version of cime, but I just noticed it.

It looks like some desired hist file comparisons are not done for the .base vs. .rest file comparison for ERS_N2 tests (and I suspect the same is true for other two-run tests).

With cime5.0.3 (in the CESM alpha01 sandbox), I ran ERS_D_N2.f10_f10.ICLM45.yellowstone_gnu.clm-default; I also ran the same test out of clm4_5_8_r182 (with cime4).

In both, it appears that the only history file comparisons that were done (based on the *.rest.cprnc.out files that were produced) were the clm2_0002 hist files. In particular, I see no evidence that comparisons were done for clm2_0001 hist files (a relatively minor issue, but I believe that should be fixed), or for cpl hist files (a more significant issue, I believe).

I was also confused by the presence of *.base.cprnc.out files in these tests, despite the fact that I didn't ask for baseline comparisons. It looks like these compared the clm2_0001 .base history files with the clm2_0002 .base history files. Perhaps this is an intentional feature of _N2 tests. If so, that's fine, but it's confusing for these results to appear in the .base.cprnc.out file, which typically are used for comparisons with baselines. Again, this behavior seems to be the same in cime5 and cime4.

### LINKED ISSUE #310: Multi-instance log files not zipped or copied to $CASEROOT/logs
This issue is NOT new to cime5, but I just noticed it. (I see the same issue with cime5.0.3 and with cime4.) I have observed this with ERS_D_N2.f10_f10.ICLM45.yellowstone_gnu.clm-default.

The multi-instance log files (files like lnd_0001.log.*, lnd_0002.log.*, etc.) are not gzipped, and are not copied to the $CASEROOT/logs directory.

### LINKED ISSUE #449: hist_utils.py: _get_latest_hist_files not general enough
I'm continuing my review of PR #413 .... Since that has already been brought to master, I'm opening issues for additional things I notice in this continued review.

hist_utils.py: _get_latest_hist_files (which I believe was written by @jedwards4b ) has this logic:

```python
    date_match = re.compile(r"(\d\d\d\d)-(\d\d)-(\d\d)-(\d\d\d\d\d).nc")
# Some code omitted
            (yr,month,day,time) = s1.group(1,2,3,4)
            s2 = date_match.search(latest_files[ext])
            (pyr,pmonth,pday,ptime) = s2.group(1,2,3,4)
            if yr > pyr or (yr == pyr and month > pmonth) or \
                    (yr == pyr and month == pmonth and day > pday) or \
                    (yr == pyr and month == pmonth and day == pday and time > ptime):
                latest_files[ext] = hist
```

This date_match regular expression isn't general enough to match all history files. For example, monthly history files have dates of the form yyyy-mm.

However, my sense is that this logic for date_match and the sort that's done based on checking year, month, day and time is unnecessary here: As long as _get_all_hist_files returns a sorted list, which is currently the case, then I'm pretty sure that the sort will already have taken care of ordering files correctly by date.

So I'd suggest getting rid of the date_match logic, which would make the core loop in this function simply look like:

```python
    for hist in test_hists:
        ext = get_extension(model, hist)
        latest_files[ext] = hist
```

If I'm missing something, and in fact this date match is important, then it should be generalized to work with monthly history files and possibly others.

### LINKED ISSUE #452: All multi-instance history files should be compared in NCK test
Until PR #413 the shell script-based component_compare_test had logic specific to the multi-instance tests (NCK and its variants) so that it would compare both the _0001 and the _0002 file from the multi-instance run with the standard file from the non-multi-instance run. I believe that was a critical piece of these tests. @mvertens wrote this.

That functionality is now gone.

I'll be the first to admit that the old way of doing it wasn't ideal - i.e., embedding this test-specific logic in the general component_compare_test. Ideally, we'd have a mechanism for an individual test to specify some custom code for doing in-test comparisons. (And, as an aside, that would address a desire that I and others have expressed for a while: we'd like to be able to make component-specific tests that do something more than just doing a blanket comparison of history files for equality.) But I realize that that might require some significant rework (or maybe not... I'm not sure).

In any case, I believe we should have some mechanism for maintaining the old NCK (etc.) behavior, even if it's via a short-term kludge that we will improve later.

### LINKED ISSUE #454: Incorrect error reported in test baseline comparison: file had no counterpart
I ran a test with a baseline comparison:

```
[hobart:~/cesm_code/cesm2_0_alpha01/cime/scripts]$ CIME_MODEL=cesm ./create_test NCK_P24x1_Ld3.f10_f10.ICLM45.hobart_nag.clm-default --compare testtag --baseline-root $scratch/cesm_baselines --test-root .
```

and the baseline comparisons failed with this message:

```
2016-08-24 15:45:31: Comparing hists for case 'NCK_P24x1_Ld3.f10_f10.ICLM45.hobart_nag.clm-default.C.20160824_213601' dir1='/scratch/cluster/sacks/NCK_P24x1_Ld3.f10_f10.ICLM45.hobart_nag.clm-default.C.20160824_213601/run', suffix1='',  dir2='/scratch/cluster/sacks/cesm_baselines/testtag/NCK_P24x1_Ld3.f10_f10.ICLM45.hobart_nag.clm-default' suffix2=''
  comparing model 'datm'
    no hist files found for model datm
  comparing model 'clm'
    File '/scratch/cluster/sacks/NCK_P24x1_Ld3.f10_f10.ICLM45.hobart_nag.clm-default.C.20160824_213601/run/NCK_P24x1_Ld3.f10_f10.ICLM45.hobart_nag.clm-default.C.20160824_213601.clm2.h0.0001-01-04-00000.nc' had no counterpart in '/scratch/cluster/sacks/cesm_baselines/testtag/NCK_P24x1_Ld3.f10_f10.ICLM45.hobart_nag.clm-default' with suffix ''
    File '/scratch/cluster/sacks/NCK_P24x1_Ld3.f10_f10.ICLM45.hobart_nag.clm-default.C.20160824_213601/run/NCK_P24x1_Ld3.f10_f10.ICLM45.hobart_nag.clm-default.C.20160824_213601.clm2.h1.0001-01-04-00000.nc' had no counterpart in '/scratch/cluster/sacks/cesm_baselines/testtag/NCK_P24x1_Ld3.f10_f10.ICLM45.hobart_nag.clm-default' with suffix ''
```

This is despite the fact that the baseline directory contains the appropriate files:

```
[hobart:/scratch/cluster/sacks/cesm_baselines/testtag/NCK_P24x1_Ld3.f10_f10.ICLM45.hobart_nag.clm-default]$ ls -l
total 72012
drwxrwxr-x 2 sacks cgdtss     4096 Aug 24 14:12 CaseDocs
-rw------- 1 sacks cgdtss  1290812 Aug 24 14:24 clm2.h0.0001-01-04-00000.nc
-rw------- 1 sacks cgdtss  4810332 Aug 24 14:24 clm2.h1.0001-01-04-00000.nc
-rw------- 1 sacks cgdtss 67602892 Aug 24 14:24 cpl.hi.0001-01-04-00000.nc
-rw-r--r-- 1 sacks cgdtss     5863 Aug 24 14:24 cpl.log.gz
-rw-rw---- 1 sacks cgdtss     1218 Aug 24 14:12 user_nl_clm
-rw-rw---- 1 sacks cgdtss     2102 Aug 24 14:12 user_nl_cpl
-rw-rw---- 1 sacks cgdtss      787 Aug 24 14:12 user_nl_datm
-rw-rw---- 1 sacks cgdtss      573 Aug 24 14:12 user_nl_rtm
```

I can't figure out why the comparisons aren't finding the appropriate baseline hist file. It's possible that I did something wrong here, though I can't see anything amiss. At first I thought this was specific to the multi-instance test, although it doesn't seem like it's trying to compare multi-instance (_0001 or _0002) files as far as I can tell.

### LINKED ISSUE #468: Cannot create multiple tests that write baselines to the same directory
It appears that the new logic to prevent you from overwriting existing baselines is too aggressive.

The way it's supposed to work is: If you have a baseline for (e.g.) test SMS.f45_g37_rx1.A.yellowstone_intel in baseline directory cesm2_0_alpha01e, then you shouldn't be able to re-generate baselines for that same test in the same baseline directory.

However, the way it's currently operating is that it prevents you from writing ANY new baselines (even for other tests) to an existing baseline directory.

Steps to reproduce:

```
./create_test --no-build SMS.f45_g37_rx1.A.yellowstone_intel --generate foo --baseline-root /glade/scratch/$USER/cesm_baselines

./create_test --no-build SMS.f45_g37_rx1.A.yellowstone_pgi --generate foo --baseline-root /glade/scratch/$USER/cesm_baselines
```

The first command succeeds (as expected), but the second fails with:

```
ERROR: Refusing to overwrite existing baseline directory, use -o to force overwrite /glade/scratch/sacks/cesm_baselines/foo
```

This error should only have been given if I tried to rerun the exact same test - i.e., the intel test.

### LINKED ISSUE #472: This line of code looks wrong in hist_utils.py: _hists_match
In this code snippet:

```python
    for hists, suffix, normalized in [(hists1, suffix1, normalized1), (hists2, suffix2, normalized2)]:
        for hist in hists:
            normalized_name = hist[hist.rfind(model):]
            if suffix1 != "":
                expect(normalized_name.endswith(suffix), "How did '%s' not have suffix '%s'" % (hist, suffix))
                normalized_name = normalized_name[:len(normalized_name) - len(suffix)]
```

The use of suffix1 in the conditional:

```python
            if suffix1 != "":
```

looks wrong to me: Intuitively, it looks like this should be simply `suffix`. I haven't taken the time to fully understand this function, so my intuition may be wrong... but if it is, an explanatory comment would help here.

jedwards4b:
Correct several issues related to multiinstance component testing

Test suite: 
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #309
Fixes #310 
Fixes #449 
Fixes #452 
Fixes #454
Fixes #468 
Fixes #472 
User interface changes: 

Code review: billsacks, jgfouca

### DIFF
--- a/utils/python/CIME/case_run.py
+++ b/utils/python/CIME/case_run.py
@@ -1,7 +1,6 @@
 
 from CIME.XML.standard_module_setup import *
 from CIME.case_submit               import submit
-from CIME.XML.files                 import Files
 from CIME.XML.machines              import Machines
 from CIME.utils                     import append_status, touch, gzip_existing_file
 from CIME.check_lockedfiles         import check_lockedfiles
--- a/utils/python/CIME/case_run.py
+++ b/utils/python/CIME/case_run.py
@@ -2,7 +2,6 @@
 from CIME.XML.standard_module_setup import *
 from CIME.case_submit               import submit
 from CIME.XML.files                 import Files
-from CIME.XML.component             import Component
 from CIME.XML.machines              import Machines
 from CIME.utils                     import append_status, touch, gzip_existing_file
 from CIME.check_lockedfiles         import check_lockedfiles
@@ -329,19 +328,8 @@ def save_logs(case, lid):
 
         caseroot = case.get_value("CASEROOT")
         rundir = case.get_value("RUNDIR")
-
-        # get components
-        files = Files()
-        config_file = files.get_value("CONFIG_DRV_FILE")
-        component = Component(config_file)
-        comps = [x.lower() for x in component.get_valid_model_components()]
-        comps = [x.replace('drv', 'cpl') for x in comps]
-        model = [case.get_value("MODEL")]
-        comps = comps + model
-
-        # for each component, compress log files and copy to logdir
-        for comp in comps:
-            logfile = os.path.join(rundir, comp + '.log.' + lid)
+        logfiles = glob.glob(os.path.join(rundir,"*.log.%s"%(lid)))
+        for logfile in logfiles:
             if os.path.isfile(logfile):
                 logfile_gz = gzip_existing_file(logfile)
                 shutil.copy(logfile_gz,
--- a/utils/python/CIME/hist_utils.py
+++ b/utils/python/CIME/hist_utils.py
@@ -249,12 +249,14 @@ def get_extension(model, filepath):
     expect(m is not None, "Failed to get extension for file '%s'" % filepath)
     return m.groups()[0]
 
-def generate_baseline(case, baseline_dir=None):
+def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
     """
     copy the current test output to baseline result
 
     case - The case containing the hist files to be copied into baselines
     baseline_dir - Optionally, specify a specific baseline dir, otherwise it will be computed from case config
+    allow_baseline_overwrite is only used in the CESM model workflow and must be true
+    to generate baselines to an existing directory.
 
     returns (SUCCESS, comments)
     """
@@ -268,6 +270,11 @@ def generate_baseline(case, baseline_dir=None):
 
     if not os.path.isdir(basegen_dir):
         os.makedirs(basegen_dir)
+
+    if (os.path.isdir(os.path.join(basegen_dir,testcase)) and
+        case.get_value("MODEL") == "CESM"  and not allow_baseline_overwrite):
+        expect(False, " Cowardly refusing to overwrite existing baseline directory")
+
 
     comments = "Generating baselines into '%s'\n" % basegen_dir
     num_gen = 0
--- a/utils/python/CIME/hist_utils.py
+++ b/utils/python/CIME/hist_utils.py
@@ -93,21 +93,27 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
     >>> hists1 = ['cam.h0.1850-01-08-00000.nc']
     >>> hists2 = ['cam_0001.h0.1850-01-08-00000.nc','cam_0002.h0.1850-01-08-00000.nc']
     >>> _hists_match('cam', hists1, hists2, '', '')
-    ([], [], [('cam.h0.1850-01-08-00000.nc', 'cam_0001.h0.1850-01-08-00000.nc')])
+    ([], [], [('cam.h0.1850-01-08-00000.nc', 'cam_0001.h0.1850-01-08-00000.nc'), ('cam.h0.1850-01-08-00000.nc', 'cam_0002.h0.1850-01-08-00000.nc')])
+    >>> hists1 = ['cam_0001.h0.1850-01-08-00000.nc.base','cam_0002.h0.1850-01-08-00000.nc.base']
+    >>> hists2 = ['cam_0001.h0.1850-01-08-00000.nc.rest','cam_0002.h0.1850-01-08-00000.nc.rest']
+    >>> _hists_match('cam', hists1, hists2, 'base', 'rest')
+    ([], [], [('cam_0001.h0.1850-01-08-00000.nc.base', 'cam_0001.h0.1850-01-08-00000.nc.rest'), ('cam_0002.h0.1850-01-08-00000.nc.base', 'cam_0002.h0.1850-01-08-00000.nc.rest')])
     """
     normalized1, normalized2 = [], []
+    multi_normalized1, multi_normalized2 = [], []
     multiinst = False
-    for hists, suffix, normalized in [(hists1, suffix1, normalized1), (hists2, suffix2, normalized2)]:
+
+    for hists, suffix, normalized, multi_normalized in [(hists1, suffix1, normalized1, multi_normalized1), (hists2, suffix2, normalized2, multi_normalized2)]:
         for hist in hists:
             normalized_name = hist[hist.rfind(model):]
-            if suffix1 != "":
+            if suffix != "":
                 expect(normalized_name.endswith(suffix), "How did '%s' not have suffix '%s'" % (hist, suffix))
                 normalized_name = normalized_name[:len(normalized_name) - len(suffix)]
 
             m = re.search(r"(.*%s.*)_[0-9]{4}(.h.*)"%model, normalized_name)
             if m is not None:
                 multiinst = True
-                normalized_name = m.group(1)+m.group(2)
+                multi_normalized.append(m.group(1)+m.group(2))
 
             normalized.append(normalized_name)
 
@@ -118,8 +124,31 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
     two_not_one = sorted([hists2[normalized2.index(item)] for item in set_of_2_not_1])
 
     both = set(normalized1) & set(normalized2)
+
     match_ups = sorted([ (hists1[normalized1.index(item)], hists2[normalized2.index(item)]) for item in both])
-    if not multiinst:
+    if multiinst:
+        new_two_not_one = []
+        new_one_not_two = []
+        for normalized_name in two_not_one:
+            m = re.search(r"(.*%s.*)_[0-9]{4}(.h.*)"%model, normalized_name)
+            if m is not None:
+                multi_normalized = m.group(1)+m.group(2)
+                if multi_normalized in one_not_two:
+                    match_ups.append((multi_normalized, normalized_name) )
+                else:
+                    new_two_not_one.append(normalized_name)
+        two_not_one = new_two_not_one
+        for normalized_name in one_not_two:
+            m = re.search(r"(.*%s.*)_[0-9]{4}(.h.*)"%model, normalized_name)
+            if m is not None:
+                multi_normalized = m.group(1)+m.group(2)
+                if multi_normalized in two_not_one:
+                    match_ups.append((multi_normalized, normalized_name) )
+                else:
+                    new_one_not_two.append(normalized_name)
+        one_not_two = new_one_not_two
+
+    else:
         expect(len(match_ups) + len(set_of_1_not_2) == len(hists1), "Programming error1")
         expect(len(match_ups) + len(set_of_2_not_1) == len(hists2), "Programming error2")
 
@@ -155,6 +184,7 @@ def _compare_hists(case, from_dir1, from_dir2, suffix1="", suffix2=""):
             all_success = False
 
         num_compared += len(match_ups)
+
         for hist1, hist2 in match_ups:
             success, cprnc_comments = cprnc(hist1, hist2, case, from_dir1, multiinst_cpl_compare)
             if success:
@@ -251,7 +281,7 @@ def get_extension(model, filepath):
     '0002.h0'
     """
     basename = os.path.basename(filepath)
-    ext_regex = re.compile(r'.*%s[^_]*_?(\d{4})?[.](h.?)([.][^.]+)?[.]nc' % model)
+    ext_regex = re.compile(r'.*%s[^_]*_?([0-9]{4})?[.](h.?)([.][^.]+)?[.]nc' % model)
 
     m = ext_regex.match(basename)
     expect(m is not None, "Failed to get extension for file '%s'" % filepath)
@@ -268,8 +298,7 @@ def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
 
     case - The case containing the hist files to be copied into baselines
     baseline_dir - Optionally, specify a specific baseline dir, otherwise it will be computed from case config
-    allow_baseline_overwrite is only used in the CESM model workflow and must be true
-    to generate baselines to an existing directory.
+    allow_baseline_overwrite must be true to generate baselines to an existing directory.
 
     returns (SUCCESS, comments)
     """
@@ -285,7 +314,7 @@ def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
         os.makedirs(basegen_dir)
 
     if (os.path.isdir(os.path.join(basegen_dir,testcase)) and
-        case.get_value("MODEL") == "CESM"  and not allow_baseline_overwrite):
+        not allow_baseline_overwrite):
         expect(False, " Cowardly refusing to overwrite existing baseline directory")
 
     comments = "Generating baselines into '%s'\n" % basegen_dir
@@ -315,8 +344,15 @@ def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
             comments += "    generating baseline '%s' from file %s\n" % (baseline, hist)
 
     # Copy namelist files from CaseDocs
-    shutil.copytree(os.path.join(case.get_value("CASEROOT"),"CaseDocs"),
-                    os.path.join(basegen_dir,"CaseDocs"))
+    base_case_docs = os.path.join(basegen_dir,"CaseDocs")
+    caseroot = case.get_value("CASEROOT")
+    if os.path.isdir(base_case_docs):
+        files = os.listdir(os.path.join(caseroot,"CaseDocs"))
+        for f in files:
+            shutil.copy2(os.path.join(caseroot, "CaseDocs", f), base_case_docs)
+    else:
+        shutil.copytree(os.path.join(case.get_value("CASEROOT"),"CaseDocs"),
+                        os.path.join(basegen_dir,"CaseDocs"))
 
     expect(num_gen > 0, "Could not generate any hist files for case '%s', something is seriously wrong" % testcase)
 
--- a/utils/python/CIME/hist_utils.py
+++ b/utils/python/CIME/hist_utils.py
@@ -288,7 +288,6 @@ def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
         case.get_value("MODEL") == "CESM"  and not allow_baseline_overwrite):
         expect(False, " Cowardly refusing to overwrite existing baseline directory")
 
-
     comments = "Generating baselines into '%s'\n" % basegen_dir
     num_gen = 0
     for model in _iter_model_file_substrs(case):
@@ -314,6 +313,10 @@ def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
                 logger.debug("Found multiinstance hist file %s"%hist)
             shutil.copy(hist, baseline)
             comments += "    generating baseline '%s' from file %s\n" % (baseline, hist)
+
+    # Copy namelist files from CaseDocs
+    shutil.copytree(os.path.join(case.get_value("CASEROOT"),"CaseDocs"),
+                    os.path.join(basegen_dir,"CaseDocs"))
 
     expect(num_gen > 0, "Could not generate any hist files for case '%s', something is seriously wrong" % testcase)
 
--- a/utils/python/CIME/hist_utils.py
+++ b/utils/python/CIME/hist_utils.py
@@ -41,21 +41,7 @@ def _get_latest_hist_files(testcase, model, from_dir, suffix=""):
     date_match = re.compile(r"(\d\d\d\d)-(\d\d)-(\d\d)-(\d\d\d\d\d).nc")
     for hist in test_hists:
         ext = get_extension(model, hist)
-        if ext in latest_files:
-            s1 = date_match.search(hist)
-            if s1 is None:
-                latest_files[ext] = hist
-                continue
-            (yr,month,day,time) = s1.group(1,2,3,4)
-            s2 = date_match.search(latest_files[ext])
-            (pyr,pmonth,pday,ptime) = s2.group(1,2,3,4)
-            if yr > pyr or (yr == pyr and month > pmonth) or \
-                    (yr == pyr and month == pmonth and day > pday) or \
-                    (yr == pyr and month == pmonth and day == pday and time > ptime):
-                latest_files[ext] = hist
-                logger.debug("ext %s hist %s %s"%(ext,hist,latest_files))
-        else:
-            latest_files[ext] = hist
+        latest_files[ext] = hist
 
     for key in latest_files.keys():
         histlist.append(latest_files[key])
@@ -105,14 +91,24 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
     >>> hists2 = ['cpl.h2.nc.SUF2', 'cpl.h3.nc.SUF2', 'cpl.h4.nc.SUF2']
     >>> _hists_match('cpl', hists1, hists2, 'SUF1', 'SUF2')
     (['FOO.G.cpl.h1.nc.SUF1'], ['cpl.h4.nc.SUF2'], [('FOO.G.cpl.h2.nc.SUF1', 'cpl.h2.nc.SUF2'), ('FOO.G.cpl.h3.nc.SUF1', 'cpl.h3.nc.SUF2')])
+    >>> hists1 = ['cam.h0.1850-01-08-00000.nc']
+    >>> hists2 = ['cam_0001.h0.1850-01-08-00000.nc']
+    >>> _hists_match('cam', hists1, hists2, '', '')
+    ([], [], [('cam.h0.1850-01-08-00000.nc', 'cam_0001.h0.1850-01-08-00000.nc')])
     """
     normalized1, normalized2 = [], []
+    multiinst = False
     for hists, suffix, normalized in [(hists1, suffix1, normalized1), (hists2, suffix2, normalized2)]:
         for hist in hists:
             normalized_name = hist[hist.rfind(model):]
             if suffix1 != "":
-                expect(normalized_name.endswith(suffix), "How did '%s' hot have suffix '%s'" % (hist, suffix))
+                expect(normalized_name.endswith(suffix), "How did '%s' not have suffix '%s'" % (hist, suffix))
                 normalized_name = normalized_name[:len(normalized_name) - len(suffix)]
+
+            m = re.search("(.*%s.*)_(\d\d\d\d)(.h.*)"%model, normalized_name)
+            if m is not None:
+                multiinst = True
+                normalized_name = m.group(1)+m.group(3)
 
             normalized.append(normalized_name)
 
@@ -124,8 +120,9 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
 
     both = set(normalized1) & set(normalized2)
     match_ups = sorted([ (hists1[normalized1.index(item)], hists2[normalized2.index(item)]) for item in both])
-    expect(len(match_ups) + len(set_of_1_not_2) == len(hists1), "Programming error1")
-    expect(len(match_ups) + len(set_of_2_not_1) == len(hists2), "Programming error2")
+    if not multiinst:
+        expect(len(match_ups) + len(set_of_1_not_2) == len(hists1), "Programming error1")
+        expect(len(match_ups) + len(set_of_2_not_1) == len(hists2), "Programming error2")
 
     return one_not_two, two_not_one, match_ups
 
@@ -242,12 +239,20 @@ def get_extension(model, filepath):
     'hi'
     >>> get_extension("cpl", "TESTRUNDIFF_Mmpi-serial.f19_g16_rx1.A.melvin_gnu.C.fake_testing_only_20160816_164150-20160816_164240.cpl.h.nc")
     'h'
+    >>> get_extension("clm","clm2_0002.h0.1850-01-06-00000.nc")
+    '0002.h0'
     """
     basename = os.path.basename(filepath)
-    ext_regex = re.compile(r'.*%s.*[.](h.?)([.][^.]+)?[.]nc' % model)
+    ext_regex = re.compile(r'.*%s[^_]*_?(\d{4})?[.](h.?)([.][^.]+)?[.]nc' % model)
+
     m = ext_regex.match(basename)
     expect(m is not None, "Failed to get extension for file '%s'" % filepath)
-    return m.groups()[0]
+    if m.group(1) is not None:
+        result = m.group(1)+'.'+m.group(2)
+    else:
+        result = m.group(2)
+
+    return result
 
 def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
     """
@@ -288,7 +293,18 @@ def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
             baseline = os.path.join(basegen_dir, basename)
             if os.path.exists(baseline):
                 os.remove(baseline)
+            '''
+            special care for multi-instance cases, only keep first instance and
+            remove instance string from filename
+            '''
+            m = re.search("(.*%s.*)_(\d\d\d\d)(.h.*)"%model, baseline)
+            if m is not None:
+                inst = m.group(2)
+                if inst != '0001':
+                    continue
+                baseline = m.group(1)+m.group(3)
 
+                logger.debug("Found multiinstance hist file %s"%hist)
             shutil.copy(hist, baseline)
             comments += "    generating baseline '%s' from file %s\n" % (baseline, hist)
 
--- a/utils/python/CIME/hist_utils.py
+++ b/utils/python/CIME/hist_utils.py
@@ -38,7 +38,6 @@ def _get_latest_hist_files(testcase, model, from_dir, suffix=""):
     test_hists = _get_all_hist_files(testcase, model, from_dir, suffix)
     latest_files = {}
     histlist = []
-    date_match = re.compile(r"(\d\d\d\d)-(\d\d)-(\d\d)-(\d\d\d\d\d).nc")
     for hist in test_hists:
         ext = get_extension(model, hist)
         latest_files[ext] = hist
@@ -105,10 +104,10 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
                 expect(normalized_name.endswith(suffix), "How did '%s' not have suffix '%s'" % (hist, suffix))
                 normalized_name = normalized_name[:len(normalized_name) - len(suffix)]
 
-            m = re.search("(.*%s.*)_(\d\d\d\d)(.h.*)"%model, normalized_name)
+            m = re.search(r"(.*%s.*)_[0-9]{4}(.h.*)"%model, normalized_name)
             if m is not None:
                 multiinst = True
-                normalized_name = m.group(1)+m.group(3)
+                normalized_name = m.group(1)+m.group(2)
 
             normalized.append(normalized_name)
 
@@ -293,14 +292,13 @@ def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
             baseline = os.path.join(basegen_dir, basename)
             if os.path.exists(baseline):
                 os.remove(baseline)
-            '''
-            special care for multi-instance cases, only keep first instance and
-            remove instance string from filename
-            '''
-            m = re.search("(.*%s.*)_(\d\d\d\d)(.h.*)"%model, baseline)
+
+            #special care for multi-instance cases,
+            #only keep first instance and
+            #remove instance string from filename
+            m = re.search("(.*%s.*)_([0-9]{4})(.h.*)"%model, baseline)
             if m is not None:
-                inst = m.group(2)
-                if inst != '0001':
+                if m.group(2) != '0001':
                     continue
                 baseline = m.group(1)+m.group(3)
 
--- a/utils/python/CIME/hist_utils.py
+++ b/utils/python/CIME/hist_utils.py
@@ -91,7 +91,7 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
     >>> _hists_match('cpl', hists1, hists2, 'SUF1', 'SUF2')
     (['FOO.G.cpl.h1.nc.SUF1'], ['cpl.h4.nc.SUF2'], [('FOO.G.cpl.h2.nc.SUF1', 'cpl.h2.nc.SUF2'), ('FOO.G.cpl.h3.nc.SUF1', 'cpl.h3.nc.SUF2')])
     >>> hists1 = ['cam.h0.1850-01-08-00000.nc']
-    >>> hists2 = ['cam_0001.h0.1850-01-08-00000.nc']
+    >>> hists2 = ['cam_0001.h0.1850-01-08-00000.nc','cam_0002.h0.1850-01-08-00000.nc']
     >>> _hists_match('cam', hists1, hists2, '', '')
     ([], [], [('cam.h0.1850-01-08-00000.nc', 'cam_0001.h0.1850-01-08-00000.nc')])
     """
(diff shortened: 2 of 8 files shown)


utils/python/CIME/case_run.py
```diff
@@ -329,19 +327,8 @@ def save_logs(case, lid):
...
-        component = Component(config_file)
-        comps = [x.lower() for x in component.get_valid_model_components()]
-        comps = [x.replace('drv', 'cpl') for x in comps]
-        model = [case.get_value("MODEL")]
-        comps = comps + model
-
-        # for each component, compress log files and copy to logdir
-        for comp in comps:
-            logfile = os.path.join(rundir, comp + '.log.' + lid)
+        logfiles = glob.glob(os.path.join(rundir,"*.log.%s"%(lid)))
```
jedwards4b: This is the fix for #310

utils/python/CIME/hist_utils.py
```diff
@@ -38,24 +38,9 @@ def _get_latest_hist_files(testcase, model, from_dir, suffix=""):
     test_hists = _get_all_hist_files(testcase, model, from_dir, suffix)
     latest_files = {}
     histlist = []
-    date_match = re.compile(r"(\d\d\d\d)-(\d\d)-(\d\d)-(\d\d\d\d\d).nc")
```
jedwards4b: Fix for #449

jgfouca:
Looks good to me. I'll let @billsacks take a look before I merge.

utils/python/CIME/hist_utils.py
```diff
@@ -105,15 +90,25 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
     >>> hists2 = ['cpl.h2.nc.SUF2', 'cpl.h3.nc.SUF2', 'cpl.h4.nc.SUF2']
     >>> _hists_match('cpl', hists1, hists2, 'SUF1', 'SUF2')
     (['FOO.G.cpl.h1.nc.SUF1'], ['cpl.h4.nc.SUF2'], [('FOO.G.cpl.h2.nc.SUF1', 'cpl.h2.nc.SUF2'), ('FOO.G.cpl.h3.nc.SUF1', 'cpl.h3.nc.SUF2')])
+    >>> hists1 = ['cam.h0.1850-01-08-00000.nc']
```
jedwards4b: match any multiinstance case to the non-multiinstance of the same model

utils/python/CIME/hist_utils.py
```diff
@@ -124,8 +119,9 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
 
     both = set(normalized1) & set(normalized2)
     match_ups = sorted([ (hists1[normalized1.index(item)], hists2[normalized2.index(item)]) for item in both])
-    expect(len(match_ups) + len(set_of_1_not_2) == len(hists1), "Programming error1")
-    expect(len(match_ups) + len(set_of_2_not_1) == len(hists2), "Programming error2")
+    if not multiinst:
```
jedwards4b: need to override for multiinstance

utils/python/CIME/hist_utils.py
```diff
@@ -242,19 +247,29 @@ def get_extension(model, filepath):
     'hi'
     >>> get_extension("cpl", "TESTRUNDIFF_Mmpi-serial.f19_g16_rx1.A.melvin_gnu.C.fake_testing_only_20160816_164150-20160816_164240.cpl.h.nc")
     'h'
+    >>> get_extension("clm","clm2_0002.h0.1850-01-06-00000.nc")
```
jedwards4b: make each instance of multi instance model a different extension

utils/python/CIME/hist_utils.py
```diff
@@ -282,6 +302,16 @@ def generate_baseline(case, baseline_dir=None):
             if os.path.exists(baseline):
                 os.remove(baseline)
 
+            #special care for multi-instance cases,
```
jedwards4b: For multiinstance only keep the first instance in the baseline and rename it to remove the _0001

utils/python/CIME/hist_utils.py
```diff
@@ -198,7 +196,14 @@ def cprnc(file1, file2, case, rundir):
     cprnc_exe = case.get_value("CCSM_CPRNC")
     basename = os.path.basename(file1)
     stat, out, _ = run_cmd("%s -m %s %s 2>&1 | tee %s/%s.cprnc.out" % (cprnc_exe, file1, file2, rundir, basename))
-    return (stat == 0 and "files seem to be IDENTICAL" in out, out)
+    if multiinst_cpl_compare:
+        #  In a multiinstance test the cpl hist file will have a different number of
+        # dimensions and so cprnc will indicate that the files seem to be DIFFERENT
+        # in this case we only want to check that the fields we are able to compare
+        # have no differences.
+        return (stat == 0 and " 0 had non-zero differences" in out, out)
```
billsacks: I like that you still check what you're able to check in this case

utils/python/CIME/hist_utils.py
```diff
@@ -269,6 +284,10 @@ def generate_baseline(case, baseline_dir=None):
     if not os.path.isdir(basegen_dir):
         os.makedirs(basegen_dir)
 
+    if (os.path.isdir(os.path.join(basegen_dir,testcase)) and
+        case.get_value("MODEL") == "CESM"  and not allow_baseline_overwrite):
```
billsacks: My understanding from talking with @jgfouca was that allow_baseline_overwrite is now expected for the ACME workflow, too. Is that not true? I don't love the sprinkling of model-specific logic throughout the scripts, because this will start to get unmaintainable when we add other models, but if it's unavoidable, I can live with it for now.

However, isn't MODEL 'cesm' (lowercase), not 'CESM' (uppercase)?

utils/python/CIME/hist_utils.py
```diff
@@ -242,19 +247,29 @@ def get_extension(model, filepath):
     'hi'
     >>> get_extension("cpl", "TESTRUNDIFF_Mmpi-serial.f19_g16_rx1.A.melvin_gnu.C.fake_testing_only_20160816_164150-20160816_164240.cpl.h.nc")
     'h'
+    >>> get_extension("clm","clm2_0002.h0.1850-01-06-00000.nc")
+    '0002.h0'
     """
     basename = os.path.basename(filepath)
-    ext_regex = re.compile(r'.*%s.*[.](h.?)([.][^.]+)?[.]nc' % model)
+    ext_regex = re.compile(r'.*%s[^_]*_?(\d{4})?[.](h.?)([.][^.]+)?[.]nc' % model)
```
billsacks: Here you use \d{4} for the multi-instance suffix, whereas in two other places you use [0-9]{4}. I think those mean the same thing, but it would be helpful for future understanding and maintenance if you were consistent in all three places. (Even better if you could somehow store that in a shared variable, but I'm not sure how easy that is, so I won't push on that.)

utils/python/CIME/hist_utils.py
```diff
@@ -282,9 +301,23 @@ def generate_baseline(case, baseline_dir=None):
             if os.path.exists(baseline):
                 os.remove(baseline)
 
+            #special care for multi-instance cases,
+            #only keep first instance and
```
billsacks: Why do you go to the trouble of only keeping the first instance? I know that, in general use currently, the _0001 and _0002 instances are identical. But it would be good if we started having multi-instance tests where the instances actually differed in some way. So it seems like it would be good to just go ahead and save (and then compare) all instances - unless there's some problem with doing that that I'm not seeing.

billsacks:
I tried running this test from a cesm2_0_alpha01 sandbox: `./create_test ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default --test-root . --walltime 0:15 --generate tests_0826 --baseline-root $scratch/cesm_baselines`

This led to two problems:

1. Despite the fact that .base and .rest files were correctly created for both the _0001 and _0002 instances, it appears that comparisons (i.e., base vs rest comparisons) were only done for the _0001 instance. These comparisons should be done for both the _0001 and _0002 instances.

2. This gave a result of:

```
FAIL ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default RUN time=77
PASS ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default COMPARE_base_rest
```

As far as I can tell, the model run itself was successful, but there was a failure in generation:

```
[Errno 17] File exists: '/glade/scratch/sacks/cesm_baselines/tests_0826/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default/CaseDocs'
  File "/glade/scratch/sacks/cesm_code/cesm2_0_alpha01_v5/cime/utils/python/CIME/SystemTests/system_tests_common.py", line 142, in run
    self._generate_baseline()
  File "/glade/scratch/sacks/cesm_code/cesm2_0_alpha01_v5/cime/utils/python/CIME/SystemTests/system_tests_common.py", line 386, in _generate_baseline
    success, comments = generate_baseline(self._case)
  File "/glade/scratch/sacks/cesm_code/cesm2_0_alpha01_v5/cime/utils/python/CIME/hist_utils.py", line 319, in generate_baseline
    os.path.join(basegen_dir,"CaseDocs"))
  File "/glade/apps/opt/python/2.7.7/gnu-westmere/4.8.2/lib/python2.7/shutil.py", line 177, in copytree
    os.makedirs(dst)
  File "/glade/apps/opt/python/2.7.7/gnu-westmere/4.8.2/lib/python2.7/os.py", line 157, in makedirs
    mkdir(name, mode)
OSError: [Errno 17] File exists: '/glade/scratch/sacks/cesm_baselines/tests_0826/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default/CaseDocs'
```

This points to both (a) a problem with the generation of CaseDocs and (b) (possibly) a problem with the reporting of errors in the generate phase - since it seems like this should have resulted in a FAIL for baseline generation, not for the RUN phase. (I suspect that (b) is long-standing, though, and not due to this PR.)

utils/python/CIME/hist_utils.py
jedwards4b: fixed

utils/python/CIME/hist_utils.py
jedwards4b: I think that a multi-instance test with different results in each instance is a fundamentally different test than what we have now.   In the current tests we now allow for any number of ensemble members and expect them all to give the same result.   Putting all of them in the baseline dir is a waste of space.    This is also the logic by which the NCK test can compare the single instance output to the multi-instance output.

billsacks: > I think that a multi-instance test with different results in each instance is a fundamentally different test than what we have now. In the current tests we now allow for any number of ensemble members and expect them all to give the same result. Putting all of them in the baseline dir is a waste of space. This is also the logic by which the NCK test can compare the single instance output to the multi-instance output.

With respect to NCK: Yes, I agree that, for this particular test, the point is that the two instances are identical so there's no need to save baselines for both. However, we discussed quite a while ago (at a CSEG meeting, I believe) that really the NCK test is a pretty weak test of the multi-instance capabilities, and we really want a stronger test, where the instances differ in some way (e.g., via some namelist parameter). For example, I could imagine a test where you do 3 runs: (a) single instance with parameter foo = 17; (b) single instance with parameter foo = 42; (c) multi-instance with instance 1 having foo=17 and instance 2 having foo=42. Then do the appropriate comparisons. Nobody ever implemented something like this, even though there was general agreement that we should. I'd like to think that it's more likely to be implemented now that it's easier to write new tests.

But even now, outside the NCK test and its variants (NC*), there is no assumption (as far as I know) that the two instances of a multi-instance test are identical: If I run an SMS_N2 test, I can point to a testmods directory that has different test mods for inst 1 vs inst 2. Again, this would provide a much stronger test of multi-instance than the current typical practice. And for this to be a good test, I'd want baselines for both instances.

All that said: I realize that there are no tests like these right now, so if you'd rather have code to reduce disk usage now, knowing that we'll likely want to delete this code later, then I'm okay with that.

billsacks:
I ran an NCK test with this new code: `NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default`. I ran into a few issues, some not related directly to this PR, but some related to this PR.

1. History files currently aren't being produced for components: see issue #469 . So for my testing, I manually copied user_nl_clm to user_nl_clm_0001 and user_nl_clm_0002.

2. If I run the NCK test directly from the branch in this PR, then the component history files without a multi-instance suffix (i.e., no _0001 or _0002) are copied to the .multiinst suffix:

```
2016-08-28 06:51:25: Moving hist files to suffix 'multiinst'
  Moving hist files for model 'datm'
  Moving hist files for model 'clm'
    Copying '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2_0002.h1.0001-01-0
4-00000.nc' to '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2_0002.h1.0001-0
1-04-00000.nc.multiinst'
    Copying '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2_0002.h0.0001-01-0
4-00000.nc' to '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2_0002.h0.0001-0
1-04-00000.nc.multiinst'
    Copying '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2.h0.0001-01-04-000
00.nc' to '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2.h0.0001-01-04-00000
.nc.multiinst'
    Copying '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2.h1.0001-01-04-000
00.nc' to '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2.h1.0001-01-04-00000
.nc.multiinst'
    Copying '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2_0001.h0.0001-01-0
4-00000.nc' to '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2_0001.h0.0001-0
1-04-00000.nc.multiinst'
    Copying '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2_0001.h1.0001-01-0
4-00000.nc' to '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2_0001.h1.0001-0
1-04-00000.nc.multiinst'
```

If we accept PR #471 then this problem goes away: when you have a separate case for the multi-instance run, then it becomes acceptable to use the aggressive wildcard match that's used here (rather than explicitly selecting files with _0001 or _0002 for the multi-instance run, as was done in cime4).

3. **The remaining problem that needs to be addressed, which is specific to this PR:** Regardless of whether PR #471 is included, only one set of multi-instance files are compared between the multi-instance run and the non-multi-instance run. Oddly, which instance (_0001 or _0002) differed in my NCK tests that included and didn't include the changes in PR #471 :

Here was the log without including #471 

```
  comparing model 'clm'
    /glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2.h0.0001-01-04-00000.nc.base matched /glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2_0002.h0.0001-01-04-00000.nc.multiinst
    /glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2.h1.0001-01-04-00000.nc.base matched /glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_124124.clm2_0002.h1.0001-01-04-00000.nc.multiinst
```

Here was the log when I included #471 

```
  comparing model 'clm'
    /glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_140416/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_140416.clm2.h0.0001-01-04-00000.nc.base matched /glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_140416/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_140416.clm2_0001.h0.0001-01-04-00000.nc.multiinst
    /glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_140416/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_140416.clm2.h1.0001-01-04-00000.nc.base matched /glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_140416/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160828_140416.clm2_0001.h1.0001-01-04-00000.nc.multiinst
```

**The desired behavior is to compare both the _0001 and _0002 files from the multi-instance run against the base file from the non-multi-instance run.** I believe that if this were done, then you could remove the `if not multiinst` conditional from this check of the invariant:

```python
    if not multiinst:
        expect(len(match_ups) + len(set_of_1_not_2) == len(hists1), "Programming error1")
        expect(len(match_ups) + len(set_of_2_not_1) == len(hists2), "Programming error2")
```

billsacks:
Baseline comparisons appear not to be working right for multi-instance tests in this PR: For both ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default and NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default (latter with #471 merged in, and with manually copying user_nl_clm to _0001 and _0002 versions), I get:

```
FAIL ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default COMPARE_baseline

  comparing model 'clm'
    File '/glade/scratch/sacks/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160828_192130/run/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160828_192130.clm2_0001.h0.0001-01-04-00000.nc' had no counterpart in '/glade/scratch/sacks/cesm_baselines/tests_0826/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default' with suffix ''
    File '/glade/scratch/sacks/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160828_192130/run/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160828_192130.clm2_0001.h1.0001-01-04-00000.nc' had no counterpart in '/glade/scratch/sacks/cesm_baselines/tests_0826/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default' with suffix ''
```

```
FAIL NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default COMPARE_baseline

  comparing model 'clm'
    File '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160828_192228/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160828_192228.clm2.h0.0001-01-04-00000.nc' had no counterpart in '/glade/scratch/sacks/cesm_baselines/tests_0828b/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default' with suffix ''
    File '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160828_192228/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160828_192228.clm2.h1.0001-01-04-00000.nc' had no counterpart in '/glade/scratch/sacks/cesm_baselines/tests_0828b/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default' with suffix ''
```

Despite having generated baselines from the same tests from this branch:

```
[yslogin4:/glade/scratch/sacks/cesm_baselines/tests_0826/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default]$ ls -1
CaseDocs
clm2.h0.0001-01-04-00000.nc
clm2.h1.0001-01-04-00000.nc
cpl.hi.0001-01-04-00000.nc
user_nl_clm_0001
user_nl_clm_0002
user_nl_cpl
user_nl_datm_0001
user_nl_datm_0002
user_nl_rtm_0001
user_nl_rtm_0002
```

```
[yslogin4:/glade/scratch/sacks/cesm_baselines/tests_0828b/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default]$ ls -1
CaseDocs
clm2.h0.0001-01-04-00000.nc
clm2.h1.0001-01-04-00000.nc
cpl.hi.0001-01-04-00000.nc
user_nl_clm
user_nl_cpl
user_nl_datm
user_nl_rtm
```

billsacks:
I'm done with my review.

utils/python/CIME/hist_utils.py
jedwards4b: Now if hist1 has single instance and hist2 has multiinstance or visa-versa the single will be compared to all of the multis.  If both have multiinstance each instance is compared

jedwards4b:
I believe that I've addressed all of the issues raised in this PR with the exception of the user_nl issue raised which I will address in #471

utils/python/CIME/hist_utils.py
```diff
@@ -123,9 +124,33 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
     two_not_one = sorted([hists2[normalized2.index(item)] for item in set_of_2_not_1])
 
     both = set(normalized1) & set(normalized2)
+
     match_ups = sorted([ (hists1[normalized1.index(item)], hists2[normalized2.index(item)]) for item in both])
-    expect(len(match_ups) + len(set_of_1_not_2) == len(hists1), "Programming error1")
-    expect(len(match_ups) + len(set_of_2_not_1) == len(hists2), "Programming error2")
+    if multiinst:
```
billsacks: Can you please add some high-level comments about what the code in this conditional is doing? I *think* what this is saying is: If there's a file with a multi-instance in one set, but this doesn't exist in the other set, and the same file without a multi-instance suffix exists in the other set, then consider these two to be a match. But I was only able to figure that out based on knowledge of the issue you were fixing by adding this code.

In general, this routine has gotten complicated enough, including some complex regex matches, that it's hard to understand the big picture of what each piece of code is doing. I'd really love to see some more comments throughout this routine, and/or have it broken up into evocatively-named lower-level functions. For example, it would help if you gave an example of what normalized_name will look like, what multi_normalized will look like, and the purposes of the various pieces of this routine.

utils/python/CIME/hist_utils.py
```diff
@@ -123,9 +124,33 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
...
+            if m is not None:
+                multi_normalized = m.group(1)+m.group(2)
+                if multi_normalized in two_not_one:
+                    match_ups.append((multi_normalized, normalized_name) )
+                else:
+                    new_one_not_two.append(normalized_name)
+        one_not_two = new_one_not_two
+
+    else:
+        expect(len(match_ups) + len(set_of_1_not_2) == len(hists1), "Programming error1")
```
billsacks: Is it possible to have similar 'expect' statements for the multi-instance case? Checking these invariants seems important given the complexity of what's going on here, especially in the multi-instance case.

billsacks:
Other than the two line comments I just made, the new code looks good to me. Before I sign off on it, I'd like to rerun the tests I documented in earlier comments. I'll do that tonight or tomorrow, and will let you know when these are done.

billsacks:
I have tested the new code with the two tests I mentioned above:

ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default
NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default

I ran both in --generate mode, then both in --compare mode.

The ERS_N2 test appears to be working correctly now - thanks for these fixes.

However, the NCK test still isn't working:

```
  comparing model 'clm'
    File '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926.clm2_0001.h0.0001-01-04-00000.nc.multiinst' had no counterpart in '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run' with suffix 'base'
    File '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926.clm2_0001.h1.0001-01-04-00000.nc.multiinst' had no counterpart in '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run' with suffix 'base'
    File '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926.clm2_0002.h0.0001-01-04-00000.nc.multiinst' had no counterpart in '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run' with suffix 'base'
    File '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926.clm2_0002.h1.0001-01-04-00000.nc.multiinst' had no counterpart in '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run' with suffix 'base'
```

Despite

```
$ ls /glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run/*.base
/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926.clm2.h0.0001-01-04-00000.nc.base
/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926.clm2.h1.0001-01-04-00000.nc.base
/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.G.20160829_191926.cpl.hi.0001-01-04-00000.nc.base
```

Baseline comparisons fail similarly for the ERS_N2 test:

```
    File '/glade/scratch/sacks/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193817/run/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193817.clm2_0001.h0.0001-01-04-00000.nc' had no counterpart in '/glade/scratch/sacks/cesm_baselines/tests_0829/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default' with suffix ''
    File '/glade/scratch/sacks/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193817/run/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193817.clm2_0001.h1.0001-01-04-00000.nc' had no counterpart in '/glade/scratch/sacks/cesm_baselines/tests_0829/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default' with suffix ''
    File '/glade/scratch/sacks/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193817/run/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193817.clm2_0002.h0.0001-01-04-00000.nc' had no counterpart in '/glade/scratch/sacks/cesm_baselines/tests_0829/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default' with suffix ''
    File '/glade/scratch/sacks/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193817/run/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193817.clm2_0002.h1.0001-01-04-00000.nc' had no counterpart in '/glade/scratch/sacks/cesm_baselines/tests_0829/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default' with suffix ''
```

and for the NCK test:

```
  comparing model 'clm'
    File '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193949/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193949.clm2.h0.0001-01-04-00000.nc' had no counterpart in '/glade/scratch/sacks/cesm_baselines/tests_0829/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default' with suffix ''
    File '/glade/scratch/sacks/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193949/run/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default.C.20160829_193949.clm2.h1.0001-01-04-00000.nc' had no counterpart in '/glade/scratch/sacks/cesm_baselines/tests_0829/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default' with suffix ''
```

Despite:

```
[yslogin4:/glade/scratch/sacks/cesm_baselines]$ ls -1 tests_0829/*
tests_0829/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default:
CaseDocs
clm2.h0.0001-01-04-00000.nc
clm2.h1.0001-01-04-00000.nc
cpl.hi.0001-01-04-00000.nc
cpl.log.gz
user_nl_clm_0001
user_nl_clm_0002
user_nl_cpl
user_nl_datm_0001
user_nl_datm_0002
user_nl_rtm_0001
user_nl_rtm_0002

tests_0829/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default:
CaseDocs
clm2.h0.0001-01-04-00000.nc
clm2.h1.0001-01-04-00000.nc
cpl.hi.0001-01-04-00000.nc
cpl.log.gz
user_nl_clm
user_nl_cpl
user_nl_datm
user_nl_rtm
```

Note that I did my testing with the head of this branch merged with the head of the branch from PR #471 . And for the NCK test, I did the manual copying of user_nl_clm files documented above.

utils/python/CIME/test_scheduler.py
```diff
@@ -114,11 +114,8 @@ def __init__(self, test_names, test_data=None,
                        "Missing baseline comparison directory %s" % full_baseline_dir)
 
             # the following is to assure that the existing generate directory is not overwritten
-            if self._baseline_gen_name and not allow_baseline_overwrite:
+            if self._baseline_gen_name:
                 full_baseline_dir = os.path.join(self._baseline_root, self._baseline_gen_name)
-                expect(not os.path.isdir(full_baseline_dir),
```
jgfouca: Why can this check not be done here?

jgfouca:
I have one additional question in test_scheduler.

utils/python/CIME/test_scheduler.py
jedwards4b: Because that is checking for the baseline root which can have a number of individual tests, we want to overwrite or not based on the existence of individual tests not baseline groups.
I do this in xml_phase

jedwards4b:
In order to do further testing I have merged this with PR #471 and will open a new PR with 
the combination + additional fixes.

utils/python/CIME/test_scheduler.py
jgfouca: Ah, that makes sense. Can we loop over test dirs here and do the check? I'd like to avoid adding another class member, but if it's hard to do it here then I'm OK with this change.

jedwards4b: If I put it there aren't I going to fail the entire create_test if one baseline exists?   I suppose this behavior is fine.

jgfouca: You could just mark the create_newcase phase as failed, but maybe it's easier to just do it the way you're doing it.

jedwards4b: Should have put this here instead of directly on slack:
Here is what I have now:
```Testnames: ['ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default', 'NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default']
Using project from env ACCOUNT: P93300606
ERROR: Baseline directories already exists ['/glade/p/cesmdata/cseg/ccsm_baselines/cesm2jpetests/ERS_D_N2_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default', '/glade/p/cesmdata/cseg/ccsm_baselines/cesm2jpetests/NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default']
Use --allow_baseline_overwrite to avoid this error```

jgfouca: That looks fine to me

Issue #469, opened 2016-08-27
Multi-instance user_nl files not created correctly for NCK tests


billsacks:
When I create an NCK test like `NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default`, there is a user_nl_clm file that has lots of important stuff. However, the user_nl_clm_0001 and user_nl_clm_0002 files just contain the blank templates that are initially created for user_nl_clm files:

```
!----------------------------------------------------------------------------------
! Users should add all user specific namelist changes below in the form of
! namelist_var = new_namelist_value
!
! EXCEPTIONS:
! Set use_cndv           by the compset you use and the CLM_BLDNML_OPTS -dynamic_vegetation setting
! Set use_vichydro       by the compset you use and the CLM_BLDNML_OPTS -vichydro           setting
! Set use_cn             by the compset you use and CLM_BLDNML_OPTS -bgc  setting
! Set use_crop           by the compset you use and CLM_BLDNML_OPTS -crop setting
! Set spinup_state       by the CLM_BLDNML_OPTS -bgc_spinup      setting
! Set irrigate           by the CLM_BLDNML_OPTS -irrig           setting
! Set co2_ppmv           with CCSM_CO2_PPMV                      option
! Set dtime              with L_NCPL                             option
! Set fatmlndfrc         with LND_DOMAIN_PATH/LND_DOMAIN_FILE    options
! Set finidat            with RUN_REFCASE/RUN_REFDATE/RUN_REFTOD options for hybrid or branch cases
!                        (includes $inst_string for multi-ensemble cases)
! Set glc_smb            with GLC_SMB                            option
! Set maxpatch_glcmec    with GLC_NEC                            option
! Set glc_do_dynglacier  with GLC_TWO_WAY_COUPLING               env variable
!----------------------------------------------------------------------------------
```

It appears the logic for copying user_nl_* to the multi-instance user_nl files is correct for _N2 tests, but not for NCK tests.

billsacks:
@jedwards4b can you please look at this?

billsacks:
At first I thought the problem could be fixed by creating an nck test with the new SystemTestsCompareTwo infrastructure, but that turns out to have this same problem. Nevertheless, I'll submit a PR with the new NCK test, because it seems more maintainable.

billsacks:
[...]

PR #471, opened 2016-08-28
Rewrite NCK test to use SystemTestsCompareTwo infrastructure

### LINKED ISSUE #443: Convert NCK test to SystemTestsCompareTwo infrastructure

billsacks:
I was seeing problems in the NCK test - particularly, issue #469. I was
hopeful that rewriting it using the SystemTestsCompareTwo infrastructure
would fix these problems. It didn't. But we wanted to do this rewrite
anyway, and it does make the behavior of #465 more correct for NCK tests
-- I'll add a comment there describing how.

Test suite: NCK_D_Ld3.f10_f10.ICLM45.yellowstone_gnu.clm-default
   Ran out of a branch where I merged this branch with the branch from
   PR #465. For the sake of this test, I manually copied user_nl_clm to
   user_nl_clm_0001 and user_nl_clm_0002 as a workaround for #469.
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

   Also, for this test as well as
   NCK_D_Ld3_P15x1.f10_f10.ICLM45.yellowstone_gnu.clm-default and
   NCK_D.f09_g16_gl5.TG.yellowstone_gnu, I compared the PE layouts with
   the cime4 NCK test to ensure we're getting the same PE layouts for
   both runs.

Fixes #443

User interface changes?: No

Code review: @jedwards4b

### DIFF
--- a/utils/python/CIME/SystemTests/nck.py
+++ b/utils/python/CIME/SystemTests/nck.py
@@ -32,19 +32,19 @@ class NCK(SystemTestsCompareTwo):
         # think that the halving was unnecessary; but it's needed in case the
         # original NTASKS was odd. (e.g., for NTASKS originally 15, we want to
         # use NTASKS = int(15/2) * 2 = 14 tasks for case two.)
-        for comp in _COMPONENT_LIST:
+        for comp in self._COMPONENT_LIST:
             ntasks = self._case.get_value("NTASKS_%s"%comp)
             if ( ntasks > 1 ):
                 self._case.set_value("NTASKS_%s"%comp, int(ntasks/2))
 
     def _case_one_setup(self):
-        for comp in _COMPONENT_LIST:
+        for comp in self._COMPONENT_LIST:
             self._case.set_value("NINST_%s"%comp, 1)
 
         case_setup(self._case, test_mode=True, reset=True)
 
     def _case_two_setup(self):
-        for comp in _COMPONENT_LIST:
+        for comp in self._COMPONENT_LIST:
             self._case.set_value("NINST_%s"%comp, 2)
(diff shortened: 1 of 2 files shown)


utils/python/CIME/SystemTests/nck.py
```diff
@@ -1,126 +1,52 @@
...
-        # Because mira/cetus interprets its run script differently than
-        # other systems we need to copy the original env_mach_pes.xml back
-#        shutil.copy(machpes1,"env_mach_pes.xml")
-#        shutil.copy("env_mach_pes.xml",
-#                    os.path.join("LockedFiles","env_mach_pes.xml"))
+class NCK(SystemTestsCompareTwo):
 
-    def run_phase(self):
-        os.chdir(self._caseroot)
+    _COMPONENT_LIST = ('ATM','OCN','WAV','GLC','ICE','ROF','LND')
```
jedwards4b: Why hardcode this list instead of getting it from the case?

billsacks: Good question: I wondered the same thing. I just copied what was already being done for the NCK test. I noticed that ESP was absent from this list, and wasn't sure if that was intentional. CPL / DRV is also absent, and I'm pretty sure that _is_ intentional. If you are able to tell me what this is supposed to look like (e.g., whether ESP is supposed to be excluded), then I can change this.

jedwards4b: Like this:
```
for comp in self._case.get_value("COMP_CLASSES").split(','):
            if comp == "DRV":
                continue
            ntasks = self._case.get_value("NTASKS_%s"%comp)
       
```

jedwards4b:
In order to do additional testing I have merged this with #465 and will open a new PR

PR #476, opened 2016-08-29
Needed suffix, not suffix1

### LINKED ISSUE #472: This line of code looks wrong in hist_utils.py: _hists_match

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #472

User interface changes?: No

Code review: bill sacks

### DIFF
--- a/utils/python/CIME/hist_utils.py
+++ b/utils/python/CIME/hist_utils.py
@@ -110,7 +110,7 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
     for hists, suffix, normalized in [(hists1, suffix1, normalized1), (hists2, suffix2, normalized2)]:
         for hist in hists:
             normalized_name = hist[hist.rfind(model):]
-            if suffix1 != "":
+            if suffix != "":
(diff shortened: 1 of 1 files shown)


billsacks:
It looks like @jedwards4b also fixed this in PR #465 . Do you want to just close this one and let the fix come in with #465?

jgfouca:
Yeah, that's fine.

Issue #480, opened 2016-08-30
-mach no longer a required argument to create_newcase


ekluzek:
-mach is still labeled as a required argument to create_newcase, but it no longer is. Here's a sample change that clarifies it...

Index: create_newcase
===================================================================
--- create_newcase	(revision 10787)
+++ create_newcase	(working copy)
@@ -33,7 +33,7 @@
                         "To see list of current compsets, use the utility manage_case in this directory")
 
     parser.add_argument("--mach", "-mach",
-                        help="(required) Specify a machine. "
+                        help="Specify a machine (if not set, base it off of the hostname). "
                         "To see list of current  machines, use the utility manage_case in this directory")
 
     parser.add_argument("--compiler", "-compiler",

jedwards4b:
Erik - can you please submit a PR with this change.

jedwards4b:
Fixed by #484

ndkeen:
that's nice

PR #482, opened 2016-08-30
Fix query-machines and remove required from doc of create_newcase

### LINKED ISSUE #479: "./manage_case --query-machines" doesn't work on a non defined machine
manage_case is a simple query script that should be able to work on any *nix machine. However, when the --query-machines option is chosen it sets up a machine object that queries the hostname  and when it doesn't find it, will fail as follows...

./manage_case --query-machines
Could not probe machine for hostname 'yongi'
ERROR: Could not initialize machine object

By setting a machine, you can get beyond this problem and it will properly do the query. It doesn't really matter what machine you choose, as long as it's in the config_machine.xml list, since it won't really do anything with the machine information for manage_case which is merely a query tool. You would want to pick a machine that will be in both the cesm and acme machine lists, and you want to pick something that will be in both for the duration. Possibly, we should add an option that is specific for this case?

Here's a sample change that gets it to work:

Index: manage_case
--- manage_case	(revision 10787)
+++ manage_case	(working copy)
@@ -35,7 +35,7 @@
     expect(os.path.isfile(config_file),
            "Cannot find config_file %s on disk" %config_file)
 
-    machines = Machines(config_file)
+    machines = Machines(config_file,machine="mira")
     machines.print_values()
 
ekluzek:
Add a "Query" machine option to creation of machine in manage_case, so that 

./manage_case --query-machines

will work on an undefined machine.

Remove documentation that "-mach" is required for create_newcase, add note that
hostname will be queried if it isn't set.

Add missing OS to acme machine ocip5

Test suite: none
Test baseline: none
Test namelist changes: none
Test status: bit for bit
manage_case --query-machines works for both acme and cesm on an undefined machine
./create_newcase --help works

Fixes #479 #480

User interface changes?: Just documentation of --help for create_newcase

Code review: self,@rljacob, @jedwards4b


jedwards4b:
Hi Erik,  This should be a one or two line change - I think you've pushed something not in sync with master.   I'm closing this.

PR #484, opened 2016-08-31
Fix query machines

### LINKED ISSUE #479: "./manage_case --query-machines" doesn't work on a non defined machine

ekluzek:
Add a "Query" machine option to creation of machine in manage_case, so that

./manage_case --query-machines

will work on an undefined machine.

Remove documentation that "-mach" is required for create_newcase, add note that
hostname will be queried if it isn't set.

Add missing OS to acme machine ocip5

Test suite: none
Test baseline: none
Test namelist changes: none
Test status: bit for bit
manage_case --query-machines works for both acme and cesm on an undefined machine
./create_newcase --help works

Fixes #479 #480

User interface changes?: Just documentation of --help for create_newcase

Code review: self,@rljacob, @jedwards4b

OK, I'm not sure what happened before, but this one is correct.


scripts/create_newcase
```diff
@@ -33,7 +33,7 @@ def parse_command_line(args, cimeroot):
                         "To see list of current compsets, use the utility manage_case in this directory")
 
     parser.add_argument("--mach", "-mach",
-                        help="(required) Specify a machine. "
+                        help="Specify a machine (if not set, base it off of the hostname). "
```
jedwards4b: How about "default: match NODENAME_REGEX in config_machines.xml"

scripts/manage_case
```diff
@@ -34,10 +34,7 @@ def query_machines():
     config_file = files.get_value("MACHINES_SPEC_FILE")
     expect(os.path.isfile(config_file),
            "Cannot find config_file %s on disk" %config_file)
-    # provide a machine name that is defined for both cesm and acme
-    # to avoid an error here on machines not in config_machines.xml
-    machines = Machines(config_file, machine="edison")
-
+    machines = Machines(config_file,machine="Query")
```
jedwards4b: # Provide a special machine name indicating no need for a machine name

ekluzek:
@jedwards4b ok made those changes.

PR #487, opened 2016-08-31
Config pes fix

### LINKED ISSUE #486: cime5.0.9 does not correctly use PES_PER_NODE when set in config_pes.xml
Although the value is set correctly in the case, the translation of NTASKS and ROOTPE from negative values (number of nodes) is using the default value instead of the one in config_pes.xml

Note that this is only a problem when you attempt to modify PES_PER_NODE in config_pes.xml

jedwards4b:
The PES_PER_NODE value should be modifiable in the config_pes.xml file but in cime5.0.9 that value was being ignored when computing the NTASK_ values.   

Test suite: scripts_regression_tests, PFS_PX2.f09_g16.B1850.yellowstone_intel.allactive-default
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #486 
User interface changes?: 

Code review:

### DIFF
--- a/cime_config/cesm/allactive/config_pes.xml
+++ b/cime_config/cesm/allactive/config_pes.xml
@@ -1407,7 +1407,7 @@
   </grid>
   <grid name="a%0.9x1.25.+l%0.9x1.25.+oi%gx1" >
     <mach name="yellowstone">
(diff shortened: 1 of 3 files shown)


jgfouca:
Looks good.

PR #492, opened 2016-09-01
Make builds thread safe

### LINKED ISSUE #491: cime5.0.9 SMP flag not thread safe in build
Whether a component uses openmp to build is determined by the setting of variable SMP or BUILD_THREADED to the Makefile, SMP is set depending on the value of NTHRDS_* for each component and is currently happening in a non-threadsafe way so if any component may set 
SMP=FALSE then that setting may be interfering with the build of other components.

A workaround is to set the BUILD_THREADED flag in your case to TRUE which will override the SMP setting and just build all components with openmp.

jgfouca:
... by moving environment changes to shell command.

Test suite: create_test --no-run cime_developer
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #491

User interface changes?: No

Code review: jedwards

### DIFF
--- a/utils/python/CIME/build.py
+++ b/utils/python/CIME/build.py
@@ -20,7 +20,6 @@ def build_model(build_threaded, exeroot, clm_config_opts, incroot, complist,
 ###############################################################################
 
     logs = []
-    overall_smp = os.environ["SMP"]
 
     thread_bad_results = []
     for model, comp, nthrds, _, config_dir in complist:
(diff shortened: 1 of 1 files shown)


utils/python/CIME/build.py
```diff
@@ -523,17 +520,17 @@ def build_libraries(case, exeroot, caseroot, cimeroot, libroot, mpilib, lid, mac
             if (not os.path.isdir(ndir)):
                 os.makedirs(ndir)
 
-        _build_model_thread(config_lnd_dir, "lnd", caseroot, bldroot, libroot, incroot, file_build, logs)
+        _build_model_thread(config_lnd_dir, "lnd", caseroot, bldroot, libroot, incroot, file_build, logs, False)
```
jgfouca: How does lnd know if it's threaded or not?

jedwards4b: Actually - how do any of the shared libraries know?

jgfouca: The global os.environ[SMP] I assume

jgfouca: The thing we were saving as "overall_smp"

jedwards4b: But the variable THREADDIR in Makefile is only set if openmp is on and this appears to be working for shared_libraries.

PR #495, opened 2016-09-01
lnd build should do SMP build if overall case is SMP

jgfouca:
Test suite: None
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: None

Code review: jedwards


jgfouca:
I think this is a better solution that just turning off SMP for lnd.

jgfouca:
And this should exactly match the previous behavior.

PR #499, opened 2016-09-02
workaround for problem resolving vars in perl

jedwards4b:
A better workaround for problem with perl resolving xml correctly

Test suite: scripts_regression_tests (coripi, cesm) 
Test baseline: 
Test namelist changes: 
Test status: bit for bi

Fixes 

User interface changes?: 

Code review:


jgfouca:
Not familiar with that code, but looks like minor change.

PR #501, opened 2016-09-02
User mods and pe layouts

### LINKED ISSUE #496: -user_mods_dir should check if the given path exists
If I do:

```
./create_newcase -case test_user_mods_0901 -compset A -res f45_g37_rx1 -user_mods_dir /some/path/that/does/not/exist
```

I get a nice message saying:

```
Adding user mods directory /some/path/that/does/not/exist
```

No errors are thrown at create_newcase or case.setup time. What I really want in this case is an error message saying that the given directory doesn't exist.

### LINKED ISSUE #497: -user_mods_dir does not handle relative paths correctly
If I do this:

```
./create_newcase -case test_user_mods_0901a -compset A -res f45_g37_rx1 -user_mods_dir ../driver_cpl/cime_config/testdefs/testmods_dirs/drv/5steps
```

I see:

```
Adding user mods directory /glade/p/work/sacks/cime/scripts/Tools/../../driver_cpl/cime_config/usermods_dirs/../driver_cpl/cime_config/testdefs/testmods_dirs/drv/5steps
```

This portion of the generated path: `driver_cpl/cime_config/usermods_dirs/../` is incorrect and causes the directory not to be found (which causes this to silently not work due to #496 ).

jedwards4b:
Fix issues with user_nl_* files and user_mods files not being found
fix pelayouts for several cesm cases
fix issues with compare fail causing a run FAIL in teststatus

Test suite: scripts_regression_tests, tests suggested by issues resolved
Test baseline: 
Test namelist changes: 

Fixes #496 
Fixes #497

User interface changes?: 

Code review:

### DIFF
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -170,7 +170,7 @@ def _main_func():
 
         # Write out the case files
         case.flush(flushall=True)
-        if user_mods_dir.startswith(".."):
+        if user_mods_dir is not None and user_mods_dir.startswith(".."):
             user_mods_dir = os.path.abspath(user_mods_dir)
 
         case.apply_user_mods(user_mods_dir)
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -170,10 +170,10 @@ def _main_func():
 
         # Write out the case files
         case.flush(flushall=True)
-        if user_mods_dir is not None and user_mods_dir.startswith(".."):
-            user_mods_dir = os.path.abspath(user_mods_dir)
-
-        case.apply_user_mods(user_mods_dir)
+        if user_mods_dir is not None:
+            if os.path.isdir(user_mods_dir):
+                user_mods_dir = os.path.abspath(user_mods_dir)
+            case.apply_user_mods(user_mods_dir)
 
     # Copy env_case.xml into LockedFiles
     copyfile(os.path.join(caseroot,"env_case.xml"),
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -170,6 +170,8 @@ def _main_func():
 
         # Write out the case files
         case.flush(flushall=True)
+        if user_mods_dir.startswith(".."):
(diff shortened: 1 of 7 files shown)


scripts/create_newcase
```diff
@@ -170,6 +170,8 @@ def _main_func():
 
         # Write out the case files
         case.flush(flushall=True)
+        if user_mods_dir is not None and user_mods_dir.startswith(".."):
```
jgfouca: This is probably not what you want. Just always call os.path.abspath. If it's already an abspath, no hard is done.

jedwards4b: I tried that but there are three possiblities:   1. The path is already absolute, 2. the path is relative to the current directory.   The path is incomplete and the root directory is found in config_paths.xml    Calling abspath everytime breaks #3 but I cant think of a case of 2 where the above wouldn't work.

utils/python/CIME/hist_utils.py
```diff
@@ -75,9 +75,12 @@ def move(case, suffix):
             comments += "    Copying '%s' to '%s'\n" % (test_hist, new_file)
             shutil.copy(test_hist, new_file)
 
-    expect(num_moved > 0, "move failed: no hist files found in rundir '%s'" % rundir)
+    all_success = True
```
jgfouca: I don't think a success return is needed here.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -223,8 +223,12 @@ def _coupler_log_indicates_run_complete(self):
         return False
 
     def _component_compare_move(self, suffix):
-        comments = move(self._case, suffix)
+        success, comments = move(self._case, suffix)
         append_status(comments, sfile="TestStatus.log")
+        status = TEST_PASS_STATUS if success else TEST_FAIL_STATUS
```
jgfouca: I don't think this code is needed.

PR #507, opened 2016-09-05
Fix indent error in hist_utils

jgfouca:
Test suite: None
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:  None

Code review:  None


jgfouca:
@jedwards4b This problem broke all testing over the weekend. I'm going to self-merge so we can get good tests tonight.

PR #508, opened 2016-09-05
New command-line access to provenance capabilities

jgfouca:
1) New save_provenance tool. Allows use to execute provenance
capabilities from command-line and independent from running the case.
 1.a) Allows users to decide to save timing info after a run without
      having to rerun.
2) Move getTiming implementation to a library so it does not have to
   be invoked via shell command line.
3) Encapsulate provenance-related code in it's own library: provenance.py
4) Improve quality of save-timing regression test
5) Add regression test for save_provenance
6) Simplify get_timing call from case_run

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: Yes, new save_provenance tool

Code review: @jedwards4b


scripts/Tools/save_provenance
```diff
@@ -0,0 +1,79 @@
+#!/usr/bin/env python
+
+"""
+This tool provide command-line access to provenance-saving functionality
```
jgfouca: New tool. Worth closer review.

utils/python/CIME/case_run.py
```diff
@@ -393,17 +209,21 @@ def case_run(case):
...
-    save_timing_setup_acme(case, lid)
+    save_prerun_provenance(case)
 
     for _ in range(data_assimilation_cycles):
         pre_run_check(case)
         run_model(case)
         post_run_check(case, lid)
         save_logs(case, lid)       # Copy log files back to caseroot
-        get_timings(case, lid)     # Run the getTiming script
+        if case.get_value("CHECK_TIMING") or case.get_value("SAVE_TIMING"):
```
jgfouca: Significant change. Things that were being done in this file's get_timings method (directory creation, zipping) have been moved to the core get_timing implementation. Also, notice that we run get_timing if CHECK_TIMING or SAVE_TIMING is true; it used to only run if CHECK_TIMING was true.

utils/python/CIME/case_run.py
```diff
@@ -393,17 +209,21 @@ def case_run(case):
     lid = time.strftime("%y%m%d-%H%M%S")
     os.environ["LID"] = lid
 
-    save_timing_setup_acme(case, lid)
+    save_prerun_provenance(case)
 
     for _ in range(data_assimilation_cycles):
```
jgfouca: One question I had here, if data_assimilation_cycles is more than 1, doesn't that break all our timing savings? The LID is the same regardless of how many times this loop runs, so wouldn't the timing data get overwritten?

utils/python/CIME/get_timing.py
```diff
@@ -0,0 +1,600 @@
...
+
+"""
+Library for implementing getTiming tool which gets timing
+information from a run.
+"""
+
+from CIME.XML.standard_module_setup import *
+
+import datetime, shutil, re, gzip
+
```
jgfouca: Mostly just copied from old version of getTiming.

utils/python/CIME/provenance.py
```diff
@@ -0,0 +1,207 @@
...
+"""
+
+from CIME.XML.standard_module_setup import *
+from CIME.utils import touch, gzip_existing_file
+
+import tarfile, getpass, signal, glob, shutil
+
+logger = logging.getLogger(__name__)
+
+def _get_batch_job_id(case):
```
jgfouca: Mostly just copied out of case_run

jgfouca:
Annotations complete. This PR should greatly enhance my ability to debug my other open PR on edison.

utils/python/CIME/case_run.py
jedwards4b: Good point - I don't think that we've considered interaction between the data assimilation cycle and timing code.

jedwards4b: Would it work just to update the lid inside the loop - thus each da cycle would get a new lid?

gold2718:
Just a note that the timing issue for data assimilation runs is temporary. As data assimilation moves to the pause-resume model, there will only be one model run.

utils/python/CIME/case_run.py
jgfouca: Yes, I think that would work.

PR #513, opened 2016-09-06
Mask grid fix

### LINKED ISSUE #510: MASK_GRID unset in cime5?
The variable MASK_GRID in env_build.xml used to be set to good values in cime2.  But in cime5 it always says "UNSET" in all the cases I've tried.   ACME needs this in its CLMBuildNamelist to set the mask.

### LINKED ISSUE #511: Short term and long-term archiving issues in cim5.0.10
case.st_archive and case.lt_archive setup for the number of tasks as in case.run rather than 1 as they should. This is on yellowstone.

Also DOUT_S is default FALSE, and we probably want it default TRUE? 

For example, the testcase SMS_D.f10_f10.A.yellowstone_intel. But, you can probably see it with other cases as well.

jedwards4b:
It appeared that the MASK_GRID xml variable was not being used and so it was removed in an earlier commit, this commit restores it and sets it's value so that it can be used be the clm buildnml

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #510 
Fixes #511 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/tests/scripts_regression_tests.py
+++ b/utils/python/tests/scripts_regression_tests.py
@@ -1010,21 +1010,18 @@ class L_TestSaveTimings(TestCreateTestCommon):
         if manual_timing:
             run_cmd_assert_result(self, "cd %s && %s/save_provenance postrun" % (casedir, TOOLS_DIR))
 
-        provenance_dir = os.path.join(timing_dir, "performance_archive", getpass.getuser(), casename, lids[0])
-        self.assertTrue(os.path.isdir(provenance_dir), msg="'%s' was missing" % provenance_dir)
+        if CIME.utils.get_model() != "acme":
+            provenance_dir = os.path.join(timing_dir, "performance_archive", getpass.getuser(), casename, lids[0])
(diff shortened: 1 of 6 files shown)


utils/python/tests/scripts_regression_tests.py
```diff
@@ -1016,11 +1016,15 @@ def simple_test(self, manual_timing=False):
     ###########################################################################
     def test_save_timings(self):
     ###########################################################################
+        if CIME.utils.get_model() != "acme":
```
jgfouca: You probably don't want to skip entire test, just skip the provenance_dir check at 1013.

jgfouca:
Looks good, one minor change needed.

utils/python/tests/scripts_regression_tests.py
jedwards4b: noted and changed

rljacob:
Which one of these commits actually addresses the MASK_GRID problem?  The older version of CIME5 I'm using still has the MASK_GRID entry in config_components and the extra lines in grids.py.

rljacob:
Nevermind.  I'll just upgrade my branch to the latest version of cime.

PR #515, opened 2016-09-07
fix mechanism to read project from file

### LINKED ISSUE #505: .cesm_proj causes screwy things to happen when it has multiple lines
Danica noticed this. She had a list of project numbers in her ~/.cesm_proj and in cime4 -- it just took the first and runs happily. In cime5.0.9 -- however, it reads in the whole list, and then hangs when it tries to submit. The utility in having a list is to list all your project numbers so you can flip between the one you want to be active.

So "get_project" should do more checking to only read a single line and maybe make sure the input isn't garbage.

jedwards4b:
The mechanism to read the .cesm_proj file for a project number was reading the entire file,
this change causes it to read until the first line not beginning with # and use that for the project id
Also fix a minor issue in scripts_regression_tests

Test suite: tested several possible project files on yellowstone  
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #505 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/utils.py
+++ b/utils/python/CIME/utils.py
@@ -530,7 +530,10 @@ def get_project(machobj=None):
     projectfile = os.path.abspath(os.path.join(os.path.expanduser("~"), ".cesm_proj"))
     if (os.path.isfile(projectfile)):
         with open(projectfile,'r') as myfile:
-            project = myfile.read().rstrip()
+            for line in myfile:
+                project = line.rstrip()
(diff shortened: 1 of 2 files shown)


rljacob:
So is this a valid .cesm_proj?

```
myproj1
myproj2
#myproj3 
```

jedwards4b:
Yes in this case it would use myproj1

Issue #518, opened 2016-09-07
TOTALPES value computed incorrectly on hyperthread capable platforms


jedwards4b:
Many HPC systems are hyperthread capable and can overload the number of tasks x threads per
node by factors of 2x or 4x or more.   The current setting of TOTALPES in cime fails to take this into account which results in an incorrect Model Cost estimate.   For example on yellowstone for a particular F case I use 960 tasks and 2 threads per task.  I have set the ptile to use 15 tasks per node so the number of tasks to be used for the cost estimate should be 1024.   Currently the 
TOTALPES value in cime for this run is 1920 and COST_PES=0.    They should be 960 and 1024.

jgfouca:
This is going to involve that TaskMaker code that none of us want to touch.

jedwards4b:
I think most of the functonality of task_maker should be moved to env_mach_pes.py and be broken into smaller subroutines - but this is a task beyond the cime5.1 milestone, for 5.1 I think I have a workaround for this problem, it involves making a special case for yellowstone where by default we use 15 of the 16 tasks per node.

PR #519, opened 2016-09-08
fix issues with timing cost calculation on SMT machines

jedwards4b:
The timing cost estimates were not taking into account the use of hyperthreads, this addresses that issue.   

Test suite: scripts_regression_tests , PFS.f19_g16.F1850.yellowstone_intel
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Connects to #518 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/case_setup.py
+++ b/utils/python/CIME/case_setup.py
@@ -189,7 +189,17 @@ def _case_setup_impl(case, caseroot, casebaseid, clean=False, test_mode=False, r
             check_lockedfiles()
 
             tm = TaskMaker(case)
-            pestot = tm.fullsum
+            mtpn = case.get_value("MAX_TASKS_PER_NODE")
+            pespn = case.get_value("PES_PER_NODE")
+            # This is hardcoded because on yellowstone by default we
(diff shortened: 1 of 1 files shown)


utils/python/CIME/case_setup.py
```diff
@@ -189,7 +189,17 @@ def _case_setup_impl(case, caseroot, casebaseid, clean=False, test_mode=False, r
...
+            mtpn = case.get_value("MAX_TASKS_PER_NODE")
+            pespn = case.get_value("PES_PER_NODE")
+            # This is hardcoded because on yellowstone by default we
+            # run with 15 pes per node
+            # but pay for 16 pes per node.  See github issue #518
+            if case.get_value("MACH") == "yellowstone":
+                pespn = 16
+            pestot = tm.totaltasks
+            if mtpn > pespn:
+                pestot = pestot * (mtpn // pespn)
```
jgfouca: Did you mean "/" instead of "//"?

jedwards4b: No I mean // its a floor division and assures an integer output

PR #520, opened 2016-09-08
Minor changes to get us back to 100% pass

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: None

Code review: jedwards

### DIFF
--- a/scripts/Tools/save_provenance
+++ b/scripts/Tools/save_provenance
@@ -71,6 +71,7 @@ def _main_func(description):
                 expected_timing_file = os.path.join(caseroot, "timing", "%s_timing.%s.%s.gz" % (model, caseid, lid))
                 if (not os.path.exists(expected_timing_file)):
                     get_timing(case, lid)
+                save_prerun_provenance(case, lid=lid)
                 save_postrun_provenance(case, lid=lid)
         else:
             expect(False, "Unhandled mode '%s'" % mode)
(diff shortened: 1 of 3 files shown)


scripts/Tools/save_provenance
```diff
@@ -71,6 +71,7 @@ def _main_func(description):
                 expected_timing_file = os.path.join(caseroot, "timing", "%s_timing.%s.%s.gz" % (model, caseid, lid))
                 if (not os.path.exists(expected_timing_file)):
                     get_timing(case, lid)
+                save_prerun_provenance(case, lid=lid)
```
jgfouca: When the user decides post-run that they want to save timings, we'll need to save the pre-run stuff in addition to post-run since none of it would have been saved originally.

utils/python/CIME/build.py
```diff
@@ -139,7 +139,9 @@ def post_build(case, logs):
 
     shutil.copy("env_build.xml", "LockedFiles")
 
-    save_build_provenance(case)
+    # must ensure there's an lid
+    lid = os.environ["LID"] if "LID" in os.environ else run_cmd_no_fail("date +%y%m%d-%H%M%S")
```
jgfouca: Some of the fake tests were not setting LID.

PR #534, opened 2016-09-13
St archive fix

### LINKED ISSUE #518: TOTALPES value computed incorrectly on hyperthread capable platforms

### LINKED ISSUE #523: create_test --test-id [arg] needs to error if a component name is included in the arg string
If the create_test --test-id argument contains the name of a model component that is also listed in the config_archive.xml, then the st_archive will incorrectly move/copy files from the rundir to the DOUT_S_ROOT location because of the regex pattern matching. 

For example,
execca ./create_test --test-id b01_mosart_nomods ERR.f19_g16.B1850.yellowstone_intel.allactive-defaultio

Creates $CASE = b01_mosart_nomods ERR.f19_g16.B1850.yellowstone_intel.allactive-defaultio.b01_mosart_nomods

When the st_archive runs, it matches on the XML comp_archive_spec compname="mosart" for all
case files and incorrectly copies them all to $DOUT_S_ROOT/archive/$CASE/rof/hist. 

This issue maybe addressed by #500 which will include better pattern matching based on component output filename settings.

### LINKED ISSUE #532: ./create_test --test errors out
```
% ./create_test --test
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 162, in _run_module_as_main
    "__main__", fname, loader, pkg_name)
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/runpy.py", line 72, in _run_code
    exec code in run_globals
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/doctest.py", line 2817, in <module>
    sys.exit(_test())
  File "/opt/local/Library/Frameworks/Python.framework/Versions/2.7/lib/python2.7/doctest.py", line 2806, in _test
    m = __import__(filename[:-3])
  File "/Users/goldy/Coding/cime/scripts/Tools/../../utils/python/CIME/test_scheduler.py", line 12, in <module>
ImportError: No module named CIME.XML.standard_module_setup
ERROR: Command: 'python -m doctest /Users/goldy/Coding/cime/scripts/Tools/../../utils/python/CIME/test_scheduler.py -v' failed with error 'None'

```
Seems to be machine independent

jedwards4b:
Fix issues with st_archive regular expression matching.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #532 
Fixes #523 
Fixes #518 

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/acme/config_archive.xml
+++ b/cime_config/acme/config_archive.xml
@@ -1,8 +1,9 @@
 <components>
 
   <comp_archive_spec compname="cam" compclass="atm">
-    <rest_file_extension>\.[ri].*</rest_file_extension>
-    <hist_file_extension>\.h.*</hist_file_extension>
+    <rest_file_extension>\.[ri]\..*</rest_file_extension>
+    <rest_file_extension>\.r[sh]\.*</rest_file_extension>
+    <hist_file_extension>\.h.*.nc$</hist_file_extension>
     <rest_history_varname>nhfil</rest_history_varname>
     <rpointer>
       <rpointer_file>rpointer$NINST_STRING.atm</rpointer_file>
@@ -11,7 +12,7 @@
   </comp_archive_spec>
 
   <comp_archive_spec compname="datm" compclass="atm">
-    <rest_file_extension>\.r.*</rest_file_extension>
+    <rest_file_extension>\.r\..*</rest_file_extension>
     <rest_history_varname>unset</rest_history_varname>
     <rpointer>
       <rpointer_file>rpointer$NINST_STRING.atm</rpointer_file>
@@ -20,8 +21,9 @@
   </comp_archive_spec>
 
   <comp_archive_spec compname="clm" compclass="lnd">
-    <rest_file_extension>\.[ri].*</rest_file_extension>
-    <hist_file_extension>\.h.*</hist_file_extension>
+    <rest_file_extension>\.[ri]\..*</rest_file_extension>
+    <rest_file_extension>\.rh.*</rest_file_extension>
+    <hist_file_extension>\.h.*.nc$</hist_file_extension>
     <rest_history_varname>locfnh</rest_history_varname>
     <rpointer>
       <rpointer_file>rpointer$NINST_STRING.lnd</rpointer_file>
@@ -40,7 +42,7 @@
 
   <comp_archive_spec compname="rtm" compclass="rof">
     <rest_file_extension>\.r.*</rest_file_extension>
-    <hist_file_extension>\.h.*</hist_file_extension>
+    <hist_file_extension>\.h.*.nc$</hist_file_extension>
     <rest_history_varname>unset</rest_history_varname>
     <rpointer>
       <rpointer_file>rpointer$NINST_STRING.rof</rpointer_file>
@@ -50,7 +52,7 @@
 
   <comp_archive_spec compname="mosart" compclass="rof">
     <rest_file_extension>\.r.*</rest_file_extension>
-    <hist_file_extension>\.h.*</hist_file_extension>
+    <hist_file_extension>\.h.*.nc$</hist_file_extension>
     <rest_history_varname>unset</rest_history_varname>
     <rpointer>
       <rpointer_file>rpointer$NINST_STRING.rof</rpointer_file>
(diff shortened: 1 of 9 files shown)


cime_config/cesm/config_archive.xml
```diff
@@ -3,7 +3,7 @@
   <comp_archive_spec compname="cam" compclass="atm">
     <rest_file_extension>\.[ri]\..*</rest_file_extension>
     <rest_file_extension>\.r[sh]\.*</rest_file_extension>
-    <hist_file_extension>\.h.*</hist_file_extension>
+    <hist_file_extension>\.h.*.nc$</hist_file_extension>
```
jedwards4b: This prevents the files copied into suffixes for comparison from being archived

scripts/create_test
```diff
@@ -471,8 +471,10 @@ def create_test(test_names, test_data, compiler, machine_name, no_run, no_build,
 def _main_func(description):
 ###############################################################################
     if "--test" in sys.argv:
-        CIME.utils.run_cmd_no_fail("python -m doctest %s/CIME/test_scheduler.py -v" %
```
jedwards4b: fix for 532 although no tests are actually run with this flag

utils/python/CIME/case_st_archive.py
```diff
@@ -132,6 +132,7 @@ def _archive_history_files(case, archive, archive_entry,
 
     # determine history archive directory (create if it does not exist)
     dout_s_root = case.get_value("DOUT_S_ROOT")
+    casename = case.get_value("CASE")
```
jedwards4b: by explicitly including the casename we prevent suffixes matching something in that part of the string, this fixes #523

utils/python/CIME/user_mod_support.py
```diff
@@ -31,7 +31,7 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
                 continue
             case_user_nl = user_nl.replace(include_dir, caseroot)
             comp = case_user_nl.split('_')[-1]
-            if ninst is not None and comp in ninst.keys():
+            if ninst is not None and comp in ninst.keys() and ninst[comp] > 1:
```
jedwards4b: Should not enter this code unless at least one component is multi instance.

cime_config/cesm/config_archive.xml
jgfouca: Are these changes needed on the ACME side too?

jedwards4b: Yes, if you are using short term archiving.

jgfouca:
Looks good, thanks. Just one question.

jedwards4b:
Do you want me to go ahead and change the acme/config_archive.xml file?

rljacob:
Yes, thanks.

cime_config/cesm/config_archive.xml
jgfouca: Yeah,  we do use it. Would you mind mirroring these changes on the ACME side? I know we don't like to touche each others' config files. But for bug fixes like this I think it's appropriate.

cime_config/acme/config_archive.xml
```diff
@@ -132,33 +134,33 @@
   <comp_archive_spec compname="dwav" compclass="wav">
     <rest_file_extension>\.r.*</rest_file_extension>
     <rest_history_varname>unset</rest_history_varname>
-    <rpointer_file>rpointer$NINST_STRING.wav</rpointer_file>
```
jgfouca: It looks like these are changes that were not on the CESM side with this PR and were probably brought over when you copied the cesm file. Is this an error? Do we want these changes on ACME?

jedwards4b: I didn't see anything that would adversely affect acme - I think that these are previous updates made on the cesm side that were not made on the acme side.   When acme customizes the config_archive.xml for external acme components we will need to take care not to overwrite them, but afaict that hasn't happened yet.

rljacob:
Does this really fix #518 ?

jedwards4b:
No, sorry it does not - I've removed that from the list.

PR #537, opened 2016-09-13
Rasm fix

jedwards4b:
Update coupler budget calculation to correct an error when using 
RASM_OPTION1 and RASM_OPTION2. Does not change coupling or
science, just budget diagnostics. The old implementation was using
an averaged valued for ocean fluxes on ocean coupling timesteps instead
of the instantaneous value. This was corrected by rearranging the call
to the ocean budget diagnostic.

Test suite: B1850 
Test baseline: cesm1_5_beta07
Test namelist changes: 

Fixes:

User interface changes?: 

Code review:

### DIFF
--- a/driver_cpl/driver/cesm_comp_mod.F90
+++ b/driver_cpl/driver/cesm_comp_mod.F90
@@ -2277,6 +2277,19 @@ end subroutine cesm_init
          enddo
          call t_drvstopf  ('CPL:atmocnp_ocnalb',hashint=hashint(5))
 
+         !----------------------------------------------------------
+         !| ocn budget (rasm_option1)
+         !----------------------------------------------------------
+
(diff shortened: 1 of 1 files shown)


douglasjacobsen:
I don't know much about how the budget calculations work, but @jonbob said it looked fine to him. I'll give it a test and merge it.

rljacob:
@jonbob can you look at this?

jonbob:
@rljacob sure.  How pressing is this?  Is Monday OK?

rljacob:
That's fine.

jonbob:
I did look, and I think it makes sense.  Should I run any tests, or do you just need something as simple as this?

rljacob:
Just want to make sure you were aware of it.

PR #538, opened 2016-09-13
Allow PET test to work on all batch machines

### LINKED ISSUE #539: PET tests do not work on skybridge
Skybridge insta-fails the single-threaded case because it tries to use 16 procs-per-node and the sbatch only requested 8 ppn.

jgfouca:
The need for this is subtle. On batch systems, the entire PET test runs
under a single submission and that submission is configured based on
the case settings for case 1, IE 2 threads for all components. This causes
the procs-per-node to be half of what it would be for single thread. On some
machines, if the mpiexec tries to exceed the procs-per-node that were given
to the batch submission, things break. Setting MAX_TASKS_PER_NODE to half of
it original value prevents this from happening in the single-thread case.

Test suite: PET.f19_g16.X on both batch and non-batch machine
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #539

User interface changes?: 

Code review: jedwards , sacks

### DIFF
--- a/utils/python/CIME/SystemTests/pet.py
+++ b/utils/python/CIME/SystemTests/pet.py
@@ -40,5 +40,14 @@ class PET(SystemTestsCompareTwo):
         for comp in self._COMPONENT_LIST:
             self._case.set_value("NTHRDS_%s"%comp, 1)
 
+        # The need for this is subtle. On batch systems, the entire PET test runs
+        # under a single submission and that submission is configured based on
+        # the case settings for case 1, IE 2 threads for all components. This causes
+        # the procs-per-node to be half of what it would be for single thread. On some
+        # machines, if the mpiexec tries to exceed the procs-per-node that were given
+        # to the batch submission, things break. Setting MAX_TASKS_PER_NODE to half of
+        # it original value prevents this.
+        self._case.set_value("MAX_TASKS_PER_NODE", self._case.get_value("MAX_TASKS_PER_NODE") / 2)
+
         # Need to redo case_setup because we may have changed the number of threads
         case_setup(self._case, reset=True)


billsacks:
First off: I really appreciate the detailed comment you made here. That really helps with these non-intuitive changes.

Personally, I'm having trouble feeling confident that this solution will work correctly on all machines. I'm really outside my comfort zone here, and am hoping that @jedwards4b will give this a critical review. My original thought was that you'd want to use the same value for MAX_TASKS_PER_NODE in case2 as in case1 (which I guess you could get via a call to self._case1.get_value from _case_two_setup, or you could save this value in _case_one_setup). This is because it's not obvious to me that MAX_TASKS_PER_NODE/2 in _case_two_setup is guaranteed to be the same as MAX_TASKS_PER_NODE in _case_one_setup (though I'm happy to believe anyone who confidently tells me that they are guaranteed to be the same). However, even with that idea, I'm not personally confident that that would do the right thing.

I'm curious if you tried the other idea we discussed on slack:
1. Renaming _case_one_setup to _common_setup
2. Adding a _case_one_setup that simply has 'pass'
3. Removing the case_setup call from _case_two_setup
4. (Then my hypothesis is that you wouldn't need this MAX_TASKS_PER_NODE setting in _case_two_setup.)

I don't know which of these is better or worse, but based on your analysis of how this worked in cime4, it sounds like that other idea more closely matches what we used to do.

In the end, I'm fine with any of these solutions, as long as @jedwards4b is.

utils/python/CIME/SystemTests/pet.py
```diff
@@ -40,5 +40,14 @@ def _case_two_setup(self):
...
+        # under a single submission and that submission is configured based on
+        # the case settings for case 1, IE 2 threads for all components. This causes
+        # the procs-per-node to be half of what it would be for single thread. On some
+        # machines, if the mpiexec tries to exceed the procs-per-node that were given
+        # to the batch submission, things break. Setting MAX_TASKS_PER_NODE to half of
+        # it original value prevents this.
+        self._case.set_value("MAX_TASKS_PER_NODE", self._case.get_value("MAX_TASKS_PER_NODE") / 2)
+
         # Need to redo case_setup because we may have changed the number of threads
         case_setup(self._case, reset=True)
```
jedwards4b: Sorry to weigh in late on this - I think that the intended solution here is that the case_setup call at line 53 should include the test_mode=True argument.   Do you mind testing again with that change instead of this one?  Thanks

jedwards4b:
I'm not finding a system where this test fails (without the proposed change) I've tried both PBS and slurm systems and it seems to work fine as is.

rljacob:
Could this be a particular problem with how skybridge configured slurm?

utils/python/CIME/SystemTests/pet.py
jgfouca: Trying now.

jgfouca:
@jedwards4b @rljacob it's possible that our admins here have made the system more strict than is normal.

@billsacks I did not try that yet

utils/python/CIME/SystemTests/pet.py
jgfouca: No luck.

rljacob:
@jgfouca edison is now slurm so you could try the test there.

jgfouca:
@rljacob yes, it's possible it would work there if their slurm is not as strict. That doesn't really help us though because we need to work on skybridge, right?

rljacob:
right, just thought it might be useful for you to see it work on some slurm system.  Maybe the case.run or case.test is a little different?

utils/python/CIME/SystemTests/pet.py
amametjanov: Since there is now a new variable `MAX_MPITASKS_PER_NODE`, halving probably needs to be done for `MAX_MPITASKS_PER_NODE` too. For example, if the case-dir starts out with MAX_TASKS_PER_NODE equal to MAX_MPITASKS_PER_NODE, then this division makes MAX_TASKS_PER_NODE (ranks and threads) less than MAX_MPITASKS_PER_NODE (only ranks).

jedwards4b: @amametjanov Is correct that MAX_MPITASKS_PER_NODE should always be <= MAX_TASKS_PER_NODE.   Note though that this is not a new variable, in the past it was called PES_PER_NODE, it's name was recently changed to clarify its intent.

Issue #408, opened 2016-08-16
Include xmlchange in tests


wilke:
xmlchange has no test in scripts_regression_tests.

gold2718:
@rljacob , @mvertens 
Is xmlchange supposed to error on invalid inputs? For instance,
`./xmlchange STOP_OPTION=foo`
seems to run just fine (and a subsequent xmlquery call verifies that foo is the new value). In CIME4, this gave an error.

jedwards4b:
It should check against valid values if defined.  I have a fix, testing now, pr soon.

rljacob:
What to test for is being discussed in #550

rljacob:
There are now 2 uses of xmlchange in scripts_regression_tests.

rljacob:
Also above example from gold2718 now gives an error.

PR #542, opened 2016-09-14
Check against valid values

jedwards4b:
This is a response to @gold2718 comment in issue #408 xmlchange should check against the list of valid values.    The fix is in entry_id.py routine _set_value and so checks against valid values for more than just xmlchange.    Note that although ```xmlchange STOP_OPTION=foo``` now traps and reports an error  ```xmlchange STOP_OPTION=\$foo``` is still allowed.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Connects to #550 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/XML/entry_id.py
+++ b/utils/python/CIME/XML/entry_id.py
@@ -157,15 +157,17 @@ class EntryID(GenericXML):
         subgroup is ignored in the general routine and applied in specific methods
         """
         expect(subgroup is None, "Subgroup not supported")
+        valid_values = self.get_optional_node("valid_values", root=node)
         if ignore_type:
             expect(type(value) is str, "Value must be type string if ignore_type is true")
-            node.set("value",value)
+            str_value = value
         else:
             type_str = self._get_type_info(node)
-            valid_values = self.get_optional_node("valid_values", root=node)
-            if valid_values.text is not None:
-                expect(value in valid_values.text.split(','), "Did not find %s in valid values:%s"%(value, valid_values.text))
-            node.set("value", convert_to_string(value, type_str, vid))
+            str_value = convert_to_string(value, type_str, vid)
+        if valid_values is not None and valid_values.text is not None and not str_value.startswith('$'):
+            vvlist = map(str.lstrip, valid_values.text.split(','))
+            expect(str_value in vvlist, "Did not find %s in valid values:%s"%(value, vvlist))
+        node.set("value", str_value)
 
         return value
 
--- a/utils/python/CIME/XML/entry_id.py
+++ b/utils/python/CIME/XML/entry_id.py
@@ -162,7 +162,11 @@ class EntryID(GenericXML):
             node.set("value",value)
         else:
             type_str = self._get_type_info(node)
+            valid_values = self.get_optional_node("valid_values", root=node)
+            if valid_values.text is not None:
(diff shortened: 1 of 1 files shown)


utils/python/CIME/XML/entry_id.py
```diff
@@ -157,12 +157,18 @@ def _set_value(self, node, value, vid=None, subgroup=None, ignore_type=False):
...
+        valid_values = self.get_optional_node("valid_values", root=node)
         if ignore_type:
             expect(type(value) is str, "Value must be type string if ignore_type is true")
-            node.set("value",value)
+            str_value = value
         else:
             type_str = self._get_type_info(node)
-            node.set("value", convert_to_string(value, type_str, vid))
+            str_value = convert_to_string(value, type_str, vid)
+        if valid_values is not None and valid_values.text is not None and not str_value.startswith('$'):
```
gold2718: Things look good and the check functionality does work. However, I have a question about our policy for resolving values. If I do something like `./xmlchange REST_OPTION=\$foo`, I get a warning in `./case.setup`:
> Not able to fully resolve item '$foo'

However, if `foo` has a valid value, `./case.setup` completes (as does `./case.build`).
What is our policy on when and whether we use environment values when resolving XML entries? Should foo be a valid XML key? If so, you should add a check for this above. If not, when does that environment variable have to be valid? Can we document the expected behavior of setting an XML variable to a variable-to-be-resolved-later?

jedwards4b: The concern you raise is valid however I don't think that I can address it in this PR - I am aware of at least one case in which changing that warning to an error would break things - that is on mira/cetus we pass $COBALT_PARTNAME to runjob, this variable is not available on the submit node but is resolved on the run node when the job starts, we also pass a $LOCARGS which may or may not be defined in the environment for a given job.   I don't know how to apply the strictness that your comment implies without breaking this functionality.

gold2718: I wasn't trying to suggest that the warning should be an error or even that the warning should somehow be removed. I really want to understand the policy on setting XML values to `$var`.
@rljacob , @mvertens , what is the policy on XML values?

1. Anything goes, if the variable is not valid at runtime, the user will get an error then.
2. Only valid XML keys can be used with the `$var` syntax (e.g., `$STOP_OPTION` is okay, `$foo` is not)
3. Environment variables (i.e., not XML keys) are handled with special syntax (e.g., `env(foo)`)
4. Some combination of 2 and 3

I need to understand this policy to correctly evaluate this pull request. Also, we should document the policy.

wilke: I think we should close this PR and create an issue for the points @gold2718 raised. The code is ok and does what is required.

gold2718: That depends on 'what is required'. Is env(foo) a valid entry? If so, the new code does not accept it.

rljacob: Some other discussion

rljacob: Yet another discussion

rljacob:
From telecon:   this PR is an improvement but doesn't completely fix the problem of checking xmlchange input.  See #550 for more discussion.

gold2718:
Connects to #408 .
Also, see #550 for future work.

PR #575, opened 2016-09-22
Cesm xml args fix

### LINKED ISSUE #294: Need an "all" option to --xml-category or preferably a regular expression check to create_test
In cime4 the xml options for create_test had the ability for an "all" option or to enter a regular expression to match more than one category or compiler. I used this to match several CLM test lists (with -xml_category aux_clm.+), as well as match all compilers (with "-xml_compiler .+"). This was also useful to do a namelist generate/compare test for ALL tests for any machine, compiler, or category in the xml test list. 

The cime5 version doesn't require all of the -xml settings, and if not set will match everything -- but you still have to provide at least one "--xml-*" option. So if the "--xml-category" option could have the ability to do a regular expression pattern match for categories that would get the same functionality back. If the regular expression match seems too strange, the ability to set an option like "all" to "--xml-category" would at least allow the namelist generation functionality to work. But, since regular expression matching is a standard module of Python, I think the regular expression matching is a reasonable request.

### LINKED ISSUE #574: Leaving --xml-compiler off runs all cases with the default compiler...
This is with cime5.1.1. If I leave --xml-compiler off the list of --xml-* settings, what I get is all the tests being run on the default compiler. What I expected was for them to run with the compiler the test is setup for.

For example on hobart...

./create_test --xml-machine hobart --xml-category aux_clm45

has all the tests run with the intel compiler. Since, "intel" is the first compiler listed in config_machines for hobart, this would be the default. But, all of the aux_clm45 tests in the aux_clm45 testlist are for the nag compiler, none are for intel.

jedwards4b:
The compiler option for cesm xml based tests is always retreved from the file and should not be forced to the default compiler. 

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #574 
Fixes #294 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -281,6 +281,7 @@ OR
 
         mach_obj = Machines(machine=machine_name)
         if args.testargs:
+            args.compiler = mach_obj.get_default_compiler() if args.compiler is None else args.compiler
             test_names = update_acme_tests.get_full_test_names(args.testargs,
                                                                mach_obj.get_machine_name(), args.compiler)
         else:
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -280,8 +280,6 @@ OR
                                "ambiguity in machine, please use the --machine option")
 
         mach_obj = Machines(machine=machine_name)
-        args.compiler = args.xml_compiler if args.compiler is None else args.compiler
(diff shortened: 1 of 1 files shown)


jgfouca:
I had to think about it a while but it looks right.

Issue #573, opened 2016-09-22
Remove new duplication in test_status.py


billsacks:
This is basically a review of PR #570 - but I didn't see that PR until it was closed, so I'm doing my review as a new issue.

First off: Thanks a lot, @fischer-ncar , for addressing issue #564 .

However, I now see a lot of duplication between these two routines, which could cause them to get out of sync in the future:

```python
    def flush(self):
        if self._phase_statuses:
            with open(self._filename, "w") as fd:
                for phase, data in self._phase_statuses.iteritems():
                    status, comments = data
                    if not comments:
                        fd.write("%s %s %s\n" % (status, self._test_name, phase))
                    else:
                        fd.write("%s %s %s %s\n" % (status, self._test_name, phase, comments))
```

```python
    def phase_statuses_dump(self):
        if self._phase_statuses:
            for phase, data in self._phase_statuses.iteritems():
                status, comments = data
                if not comments:
                    logging.info("%s %s %s" % (status, self._test_name, phase))
                else:
                    logging.info("%s %s %s %s" % (status, self._test_name, phase, comments))
```

In addition, it seems asymmetric to me that cs.status uses a print for the overall status, but a logging call for the individual pieces.

I'm thinking that this duplication could be removed as follows (untested!):

```python
    def flush(self):
        with open(self._filename, "w") as fd:
            self.phase_statuses_dump(fd)

    def phase_statuses_dump(self, fd):
        """
        Args:
            fd: file open for writing
        """
        if self._phase_statuses:
            for phase, data in self._phase_statuses.iteritems():
                status, comments = data
                if not comments:
                    fd.write("%s %s %s\n" % (status, self._test_name, phase))
                else:
                    fd.write("%s %s %s %s\n" % (status, self._test_name, phase, comments))

    
def cs_status(test_paths, model):
    ...
        print "%s (Overall: %s), details:" % (test_name, summary)
        ts.phase_statuses_dump(sys.stdout)
```

This change also ensures that the phase_statuses_dump output goes to the same place as the overall status - since writing to sys.stdout is effectively the same as doing a 'print'.

Issue #396, opened 2016-08-14
TestStatus: Still would like clearer distinction between test comparisons and baseline comparisons


billsacks:
Inspired by @gold2718 's comment here:

https://github.com/ESMCI/cime/pull/392#issuecomment-239626725
- where he suggests knowing ahead of time which test phases are expected: I realized that it would still be good to have a clearer distinction made between history file comparisons done as an integral part of a test (e.g., comparing .base with .rest files for an ERS test) vs. baseline comparisons. Even though these are both about comparing history files, they are fundamentally different in their meaning. 

cime4 test reporting made a clear distinction between these, as:

```
--- Test Functionality  ---:
PASS ERS_D_Ld10.T31_g37_gl5.IGHISTCLM45.yellowstone_intel.clm-glcMEC_decrease.clm2.h0.nc : test compare clm2.h0 (.base and .rest files)
<snip>
--- Baseline Comparison ---:
PASS ERS_D_Ld10.T31_g37_gl5.IGHISTCLM45.yellowstone_intel.clm-glcMEC_decrease.cism.h.nc : baseline compare cism.h (baseline: compare .base file with clm4_5_9_r185 file)
```

and I feel it would be good to maintain that clear distinction in cime5.

@jgfouca made some steps towards this, by adding suffixes to the single 'COMPARE' phase. But it seems like it would be good to explicitly split this into two phases - a TEST_COMPARE phase and a BASELINE_COMPARE phase - both in the code and in the test reporting. In the test reporting, this will make it easier to do greps to include / exclude particular aspects of the tests. In the code, this will make it easier to add new functionality, such as the functionality proposed by @gold2718 .

jgfouca:
@billsacks the handling of COMPARE phases in CIME is already awkward and having another phase like COMPARE is not something I'm excited about. If you can describe a specific need/capability that would require this change, would you add a comment hear describing it? I'm going to tentatively close this for now. Feel free to reopen if you think we should reconsider.

rljacob:
We do need to distinguish between "compare with baseline fail" and "compare within a test fail".  The latter is much more serious.  In fact, the test should just stop and not bother with baseline compare if the primary test compare failed.

jgfouca:
@rljacob I have a different PR going in soon that will mark a test with an "internal" diff (non-baseline) as FAIL instead of DIFF.

billsacks:
[...]

PR #577, opened 2016-09-23
Cs status output updates

### LINKED ISSUE #563: Report baseline fails differently if just because baseline directory DNE (both Overall and detailed)
In cime4 this was called a "BFAIL" if the baseline directory didn't exist. I don't know that's the best name, but we need a way to easily figure out that the baseline comparison failed -- just because it didn't have a baseline directory to compare to. I'd like to see this in the Overall test status, as well as the detailed test status.

@jedwards4b @mvertens @billsacks @bandre-ucar

fischer-ncar:
Refactor code is test_status.py to remove duplicate code in phase_statuses_dump.
Add BFAIL the TestStatus comment when baselines are missing.
Set Overall status to BFAIL when test runs, but baselines are missing.

Test suite: Tested with cesm2_0_beta01 testresults
Test baseline: 
Test namelist changes: 
Test status:

Fixes #563 #573 

User interface changes?: 

Code review:


utils/python/CIME/test_status.py
```diff
@@ -121,24 +123,22 @@ def get_status(self, phase):
...
     def flush(self):
-        if self._phase_statuses:
-            with open(self._filename, "w") as fd:
-                for phase, data in self._phase_statuses.iteritems():
-                    status, comments = data
-                    if not comments:
-                        fd.write("%s %s %s\n" % (status, self._test_name, phase))
-                    else:
-                        fd.write("%s %s %s %s\n" % (status, self._test_name, phase, comments))
+        with open(self._filename, "w") as fd:
```
jgfouca: Good cleanup

utils/python/CIME/hist_utils.py
```diff
@@ -277,7 +277,7 @@ def compare_baseline(case, baseline_dir=None):
 
     for bdir in dirs_to_check:
         if not os.path.isdir(bdir):
-            return False, "ERROR baseline directory '%s' does not exist" % bdir
+            return False, "ERROR BFAIL baseline directory '%s' does not exist" % bdir
```
jgfouca: I'm fine with this.

utils/python/CIME/test_status.py
```diff
@@ -241,7 +242,10 @@ def get_overall_test_status(self, wait_for_run=False, check_throughput=False, ch
                         rv = NAMELIST_FAIL_STATUS
 
                 elif (rv in [NAMELIST_FAIL_STATUS, TEST_PASS_STATUS] and phase.startswith(COMPARE_PHASE)):
-                    rv = TEST_DIFF_STATUS
+                    if (phase.startswith(COMPARE_BASELINE_PHASE) and TEST_BFAIL_STATUS in comment):
```
jgfouca: I'm not sure about this one. I thought we agreed not to add a new state to the system?

utils/python/CIME/hist_utils.py
billsacks: I think this should use the constant defined in test_status.py (currently named TEST_BFAIL_STATUS) rather than hard-coding the string 'BFAIL'.

utils/python/CIME/test_status.py
billsacks: Good point @jgfouca . (It's possible that I was the one who suggested something like this.) So I think your point is that this should just be changed back to giving a DIFF overall status if the baselines do not exist? I'd support that. (But then again, I personally care more about the status of the individual pieces than the overall status in my parsing, so maybe someone else like @ekluzek - who I think relies more on the overall status - should weigh in on this.)

utils/python/CIME/test_status.py
```diff
@@ -18,21 +18,23 @@
...
+SETUP_PHASE            = "SETUP"
+NAMELIST_PHASE         = "NLCOMP"
+SHAREDLIB_BUILD_PHASE  = "SHAREDLIB_BUILD"
+MODEL_BUILD_PHASE      = "MODEL_BUILD"
+RUN_PHASE              = "RUN"
+THROUGHPUT_PHASE       = "TPUTCOMP"
+MEMCOMP_PHASE          = "MEMCOMP"
+MEMLEAK_PHASE          = "MEMLEAK"
+COMPARE_PHASE          = "COMPARE" # This is one special, real phase will be COMPARE_$WHAT
+COMPARE_BASELINE_PHASE = "COMPARE_baseline"
```
billsacks: I like that you distinguish this as a separate phase, addressing the request in issue #396 . However, if you make this change, I'd request:

1. Changing this code in system_tests_common.py:

```python
            self._test_status.set_status("%s_baseline" % COMPARE_PHASE, status, comments=ts_comments)
```

Now this should use COMPARE_BASELINE_PHASE in place of "%s_baseline" % COMPARE_PHASE.

2. To make this phase more distinct from the in-test comparison, I'd suggest renaming this to something like BASELINE_COMPARE - i.e., so the name looks more different from the COMPARE_whatever used for the in-test comparison.

utils/python/CIME/test_status.py
```diff
@@ -18,21 +18,23 @@
 # Special statuses that the overall test can be in
 TEST_DIFF_STATUS     = "DIFF"   # Implies a failure in one of the COMPARE phases
 NAMELIST_FAIL_STATUS = "NLFAIL" # Implies a failure in the NLCOMP phase
+TEST_BFAIL_STATUS    = "BFAIL"  # Implies baselines are missing in the COMPARE_baseline phase
```
billsacks: Based on @jgfouca 's comment: It seems like there shouldn't be a separate overall status for BFAIL. However, I'd still support having "BFAIL" be defined somewhere in this file as a special string that can appear in comments, to clearly signal that this string has some special meaning.

Maybe you can introduce a new section of constants here like this:

```python
# Special strings that can appear in comments, indicating particular types of failures
TEST_NO_BASELINES_COMMENT = "BFAIL" # Implies baseline directory is missing in the baseline comparison phase
```

utils/python/CIME/hist_utils.py
billsacks: Or, according to my comment below, this would use the TEST_NO_BASELINES_COMMENT

utils/python/CIME/test_status.py
fischer-ncar: I have no issue with changing it back to diff.  But I thought it was no new states to the TestStatus file.  This just changes the output from cs_status.

utils/python/CIME/test_status.py
fischer-ncar: When I remove the BFAIL from the overall status, the COMPARE_BASELINE_PHASE is no longer needed.

utils/python/CIME/test_status.py
fischer-ncar: Good idea.

utils/python/CIME/test_status.py
billsacks: @fischer-ncar : @jgfouca can chime in more definitively, but I think the issue is that other tools (like test_scheduler.py) use this get_overall_test_status function, too.

utils/python/CIME/test_status.py
billsacks: Right, you could just revert that. However, I personally feel that having a separate COMPARE_BASELINE_PHASE is a step in the right direction. Though feel free to do whatever you prefer here - if you revert it, we can always reintroduce it later to address #396

fischer-ncar:
@jgfouca I've removed the code that had the BFAIL code for the overall status.
@billsacks I've added the new section of constants that can appear in comments.

jgfouca:
Waiting for @billsacks to approve and then I'll merge.

PR #578, opened 2016-09-23
Task maker cleanup

### LINKED ISSUE #518: TOTALPES value computed incorrectly on hyperthread capable platforms

jedwards4b:
Remove the task_maker.py and replace its functionality with something closer to the intended CIME coding style

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #518 

User interface changes?: 

Code review:


jedwards4b:
Putting this up for review and discussion but I would like to test it on PBS, COBALT and SLURM systems before merging.  Deserves careful scrutiny, tested on LSF.

jedwards4b:
Now tested on PBS and SLURM systems

PR #590, opened 2016-09-27
Eri hybrid fix

### LINKED ISSUE #580: ERI test with REFCASE prestages file into wrong run directory
In @ekluzek 's CLM testing with cime5.1.1, this test failed:

```
ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default
```

It looks like the problem is that this uses a CLM initial conditions file from a REFCASE, and the ERI test prestaged that initial conditions file into the main case's run directory, while it _should have_ prestaged it into the .ref1 case's run directory.

I also noticed that, in cime4, this test uses finidat set to '/glade/p/cesmdata/cseg/inputdata/ccsm4_init/b40.1850.track1.2deg.003/year_401/b40.1850.track1.2deg.003.clm2.r.0401-01-01-00000.nc' (which comes from namelist_defaults). But Erik thinks the new version is correct - i.e., it's correct for the REFCASE to take precedence over the value from namelist_defaults.

Tagging @ekluzek @jedwards4b @mvertens

### LINKED ISSUE #582: create_test should prevent running two test suites in the same directory with the same testid
Inspired by #581 : It would be nice if create_test attempted to detect when you were running multiple create_test instances in the same directory with the same testid. Doing so can create subtle problems with your test suites.

This is likely hard (or impossible) to catch 100% of the time, but I think we could come up with a solution that works 99.something% of the time, which is better than nothing.

For example, a quick and dirty solution could be for create_test to create some empty file whose name contains the given testid... this would be done right at the beginning of create_test. But, before creating this file, it would check to see if a file with this name already appears, and if so, abort. This could still be prone to race conditions, but they should be significantly less common than how often you accidentally get the same testid for two runs of create_test.

Tagging @ekluzek

jedwards4b:
Fixes bug when ERI is run on hybrid compset, adds random number to autogenerated testids

Test suite: scripts_regression_tests, ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #582 
Fixes #580

User interface changes?: 

Code review:

### DIFF
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -16,7 +16,7 @@ from CIME.utils import expect, convert_to_seconds, compute_total_time, convert_t
 from CIME.XML.machines import Machines
 from CIME.case import Case
 
-import argparse, math, glob
+import argparse, math, glob, random
 
 logger = logging.getLogger(__name__)
 
@@ -257,7 +257,7 @@ OR
         args.no_batch = True
 
     if args.test_id is None:
-        args.test_id = CIME.utils.get_timestamp()
+        args.test_id = "%s.%s"%(CIME.utils.get_timestamp(), random.randrange(0,99,1))
 
     if args.testfile is not None:
         with open(args.testfile, "r") as fd:
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -178,8 +178,8 @@ OR
     parser.add_argument("-t", "--test-id",
                         help="Specify an 'id' for the test. This is simply a"
                         "string that is appended to the end of a test name."
-                        "If no testid is specified, then a time stamp will be"
-                        "used.")
+                        "If no testid is specified, then a time stamp plus random number"
+                        "will be used.")
 
     parser.add_argument("-j", "--parallel-jobs", type=int, default=None,
                         help="Number of tasks create_test should perform simultaneously. Default "
@@ -257,7 +257,7 @@ OR
         args.no_batch = True
 
     if args.test_id is None:
-        args.test_id = "%s.%s"%(CIME.utils.get_timestamp(), random.randrange(0,99,1))
+        args.test_id = "%s.%s"%(CIME.utils.get_timestamp(), random.randrange(0,9999,1))
 
     if args.testfile is not None:
         with open(args.testfile, "r") as fd:
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -257,7 +257,7 @@ OR
         args.no_batch = True
 
     if args.test_id is None:
-        args.test_id = "%s.%s"%(CIME.utils.get_timestamp(), random.randrange(0,9999,1))
+        args.test_id = "%s.%s"%(CIME.utils.get_timestamp(), CIME.utils.id_generator())
 
     if args.testfile is not None:
         with open(args.testfile, "r") as fd:
--- a/utils/python/CIME/SystemTests/eri.py
+++ b/utils/python/CIME/SystemTests/eri.py
@@ -42,7 +42,7 @@ class ERI(SystemTestsCommon):
         clone1, clone2 = [self._case.create_clone(clone_path, keepexe=True) for clone_path in [clone1_path, clone2_path]]
         orig_case = self._case
         orig_casevar = orig_case.get_value("CASE")
-
+        orig_runtype = orig_case.get_value("RUN_TYPE")
         #
         # determine run lengths needed below
         #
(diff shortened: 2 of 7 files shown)


scripts/create_test
```diff
@@ -257,7 +257,7 @@ OR
         args.no_batch = True
 
     if args.test_id is None:
-        args.test_id = CIME.utils.get_timestamp()
+        args.test_id = "%s.%s"%(CIME.utils.get_timestamp(), random.randrange(0,99,1))
```
jgfouca: I like this idea but I'm worried that 0-99 is not enough entropy. I found this via google:
```
>>> import uuid
>>> print str(uuid.uuid4())[:8]
```
We could use :4 here. That would give us entropy of 16^4

scripts/create_test
```diff
@@ -257,7 +257,7 @@ OR
         args.no_batch = True
```
jgfouca: Around line 180, please update the help field for --test-id to explain the added random component.

utils/python/CIME/test_scheduler.py
```diff
@@ -80,7 +80,7 @@ def __init__(self, test_names, test_data=None,
             self._test_root = self._test_root.replace("$PROJECT", self._project)
 
         self._test_root = os.path.abspath(self._test_root)
-        self._test_id   = test_id if test_id is not None else CIME.utils.get_timestamp()
+        self._test_id   = test_id
```
jgfouca: I'm not sure about this change. The default value for test_id is None. Don't we need to handle the case where it's None by assigning it to something? We don't need to add the random component here.

scripts/create_test
jedwards4b: Changing 99 to 9999 would do the same thing?

utils/python/CIME/test_scheduler.py
jedwards4b: I don't see how it could ever be None - but I can change this back

jgfouca: I think I might be using test_scheduler via library call from some of the scripts_regression_tests. I probably provide a test_id, but as long as the API is what is it, there's the chance of test_id being None.

scripts/create_test
jgfouca: Yeah. I'm thinking that maybe the hex characters might be more obvious to users (makes it clear it's not a date), but I don't particularly care.

billsacks: @jgfouca beat me to it. I was going to suggest this:

```python
def id_generator(size=6, chars=string.ascii_lowercase + string.digits):
    return ''.join(random.choice(chars) for _ in range(size))
```

(maybe we can put that in utils.py).

Then

id_generator()

generates a 6-character ID that has > 1 billion possibilities. To me that seems safe enough.

Credit: http://stackoverflow.com/questions/2257441/random-string-generation-with-upper-case-letters-and-digits-in-python

jgfouca: @billsacks that is probably a little better than uuid since it makes it much more likely that there will be at least on letter and therefore increase the chance that the user will recognize it for what it is.

utils/python/CIME/SystemTests/eri.py
```diff
@@ -90,6 +90,9 @@ def run_phase(self):
         clone1.set_value("REST_N", rest_n1)
         clone1.set_value("HIST_OPTION", "never")
         clone1.flush()
+        # if the initial case is hybrid this will put the reference data in the correct location
+        if orig_runtype == "hybrid":
```
billsacks: Why do you only do this for hybrid? Wouldn't it also be needed for branch?

But more generally: If I'm understanding the check_all_input_data call correctly, then it looks like it would work to just call that all the time - i.e., removing the check of this orig_runtype. Then it already does its own check for whether any work is needed:

```python
    if get_refcase and run_type != "startup" and not continue_run:
```

jedwards4b: there is no out of the box branch run is there?

billsacks: I have no idea; you may be right. My bigger point was that it seems unnecessary to have this conditional at all, and since it may be overly restrictive, I was thinking it could be deleted.

scripts/create_test
jedwards4b: done

jgfouca:
I approve. I will wait for @billsacks to approve and then merge.

jedwards4b:
@billsacks I think I've addressed all of your concerns

PR #602, opened 2016-09-28
Add walltimemax to skybridge and redsky

jgfouca:
Make sure num_nodes is always int.

Test suite: 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: none

Code review: jedwards

### DIFF
--- a/cime_config/acme/machines/config_batch.xml
+++ b/cime_config/acme/machines/config_batch.xml
@@ -201,21 +201,21 @@
 
   <batch_system MACH="skybridge" type="slurm" version="x.y">
     <queues>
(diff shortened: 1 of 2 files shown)


rljacob:
Do we need to check this for all the ACME machines?

jgfouca:
@rljacob yes, i think so

rljacob:
Ok.  Also can this same branch be pushed to the maint-cime5.1 branch?

jedwards4b:
No need - this problem was only introduced after maint-cime5.1

PR #607, opened 2016-09-28
fix issues with gen_domain build process

### LINKED ISSUE #606: Some problems with gen_domain using the cime configure
This is with cime5.1.1. There are several problems with using the new python configure for creating gen_domain. The most notable of which is that "-DLINUX" isn't included in the CPPDEFS and as a result when you build and run it as follows it dies with an error trying to evaluate an environment variable (USER).

The error is ...

 create domain.ocn.domain.ocn.gx3v7_test.nc.160928.nc
 write domain.ocn.domain.ocn.gx3v7_test.nc.160928.nc
(shr_sys_getenv) ERROR: no implementation of getenv for this architecture
(shr_sys_getenv) no implementation of getenv for this machine

This is for building with the bash shell as follows...

$CIMEROOT/tools/configure --machine yellowstone --macros-format Makefile
. .env_mach_specific.sh
./gen_domain -m /glade/p/cesm/cseg/inputdata/cpl/cpl6/map_gx3v7_to_T31_aave_da_090903.nc -o domain.ocn.gx3v7_test.nc -l domain.lnd.10x15_gx3v7.test.nc

jedwards4b:
Fix issues with gen_domain build process and documentation 

Test suite: none (scripts_regression_tests does not cover this) hand inspected and built get_domain
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #606 

User interface changes?: 

Code review:

### DIFF
--- a/tools/mapping/gen_domain_files/src/Makefile
+++ b/tools/mapping/gen_domain_files/src/Makefile
@@ -70,7 +70,7 @@ ifeq ($(SMP),$(null))
   SMP := FALSE
 endif
 
-CPPDEF += $(USER_CPPDEFS)
+CPPDEFS += $(USER_CPPDEFS) -D$(OS)
 
 # Set optimization on by default
 ifeq ($(OPT),$(null))
@@ -78,7 +78,7 @@ ifeq ($(OPT),$(null))
 endif
 
 ifeq ($(OPT),TRUE)
-  CPPDEF := -DOPT
+  CPPDEFS := -DOPT
 endif
 
 # Load dependency search path.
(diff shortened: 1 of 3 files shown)


tools/mapping/gen_domain_files/src/Makefile
```diff
@@ -70,15 +70,15 @@ ifeq ($(SMP),$(null))
...
+CPPDEFS += $(USER_CPPDEFS) -D$(OS)
 
 # Set optimization on by default
 ifeq ($(OPT),$(null))
   OPT := TRUE
 endif
 
 ifeq ($(OPT),TRUE)
-  CPPDEF := -DOPT
+  CPPDEFS := -DOPT
```
ekluzek: This needs to be += so it doesn't throw everything away...
So...

CPPDEFS += -DOPT

jedwards4b: fixed

PR #615, opened 2016-09-30
Comment code preventing comparison of month / year intervals

douglasjacobsen:
This merge comments out the code that prevents comparing time intervals
with years and months, allowing these time intervals to be compared
using days and seconds instead.

These changes shouldn't affect intervals that previously worked, and
should only allow year and month intervals to be compared.

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Code review:


douglasjacobsen:
See discussion on an older commit here: https://github.com/ESMCI/cime/commit/63772d874cdfe7618a27805b82aab9bbdd6fbd00#commitcomment-19233787

PR #626, opened 2016-09-30
Memleak tolerance

### LINKED ISSUE #588: More memleak false positives
This test:

```
ERS_Ly5.T31_g37_gl5.IGCLM45.yellowstone_intel.clm-glcMEC_long
```

generates a memleak false positive. In the 2-year restart portion of the run (which is the portion tested for memleak), there is a true, small increase of about 0.1 MB per day for the first few months (but I don't think that alone would have triggered a failure). Then there is a big jump at the start of the second year:

```
FAIL ERS_Ly5.T31_g37_gl5.IGCLM45.yellowstone_intel.clm-glcMEC_long MEMLEAK memleak detected, memory went from 533.950000 to 620.100000 in 19999 days
```

```
 tStamp_write: model date =    41230       0 wall clock = 2016-09-26 13:49:09 avg dt =     1.10 dt =     1.05
 memory_write: model date =    41230       0 memory =     549.67 MB (highwater)       2426.76 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC WAV ESP)
 tStamp_write: model date =    41231       0 wall clock = 2016-09-26 13:49:10 avg dt =     1.10 dt =     1.02
 memory_write: model date =    41231       0 memory =     549.67 MB (highwater)       2426.76 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC WAV ESP)
 tStamp_write: model date =    50101       0 wall clock = 2016-09-26 13:49:19 avg dt =     1.12 dt =     8.77
 memory_write: model date =    50101       0 memory =     619.98 MB (highwater)       2446.15 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC WAV ESP)
 tStamp_write: model date =    50102       0 wall clock = 2016-09-26 13:49:20 avg dt =     1.12 dt =     1.02
 memory_write: model date =    50102       0 memory =     619.98 MB (highwater)       2446.15 MB (usage)  (pe=    0 comps= cpl ATM LND ICE OCN GLC WAV ESP)
```

The full, 5-year run also has the increase at the start of year 2, but then memory use stays roughly constant after that (other than an increase when the restart files are written).

This test uses CISM, which only runs once per year. So this increase in memory use at the start of the second year (really, the very end of the first year) is expected.

I'm not quite sure what the ideal solution is here. Off the top of my head, I can see two:

1. Quick & dirty: Allow tests to specify that they expect a (false) memleak failure, via some xml entry, and thus don't run the memory check on those tests.

2. Improve the heuristics used in these memleak tests. I'm not quite sure what would be the best way to code this, though. I know that you should ignore a jump at the end of the first year / beginning of the second year in a run that uses CISM, but how should we specify that in a general way? e.g., Should each component somehow specify the time frequency at which it does its work, and then ignore memory jumps at the first instance of each of those frequencies? Note that, for CISM, this differs from GLC_NCPL (generally 1 day), which complicates things.

### LINKED ISSUE #620: Tweak testlist xml format to facilitate adding test metadata
The current testlist xml format looks like this:

```xml
  <test name="ERR_N3" grid="f19_g16" compset="B1850" testmods="allactive/defaultio">
    <machines>
      <machine name="edison" compiler="intel" category="prebeta"/>
      <machine name="yellowstone" compiler="intel" category="prealpha"/>
      <options>
	<option name="wallclock"> 00:20 </option>
      </options>
    </machines>
  </test>
```

(First a word of appreciation: This is a lot nicer than the older testlist xml format, which was heavily nested.)

Currently, this supports the option of wallclock. We have also talked about adding options for expected memleak amounts (in #588 ) and expected failures (in some older discussions). But these latter options - and to some extent the wallclock option - are machine and compiler-specific, so it seems like these need to be nested under machine, not in a separate options block that applies to all machines/compilers.

Also, somewhat related: Can someone document how to do the following with the testlist xml files (or point me to this documentation if it's already documented somewhere):

1. Add a comment that appears when using the tool that queries testlists (i.e., not just a comment in the xml, but something with a special comment tag). I don't see any examples of this in the new xml formatted files. Incidentally, this would be another tag that is good to have nested under a specific machine / compiler / category, because my comments are often specific to that level.

2. Have the exact same test appear in multiple categories: Is the correct way to do this to have two <machine...> lines that are identical except for their category, or to have something like <machine name="yellowstone" compiler="intel" category="aux_clm45|prealpha">?

3. Similar question to (2), but how do you have the same test run with multiple compilers? Again, is the expectation that there should be multiple <machine...> lines or one machine line with a '|' in the match?

jedwards4b:
Adds TEST_MEMLEAK_TOLERANCE variable to xml and memleak_tolerance option to testlist.xml files (version 2) 
Test suite: scripts_regression_tests, ERS_Ly5.T31_g37_gl5.IGCLM45.yellowstone_intel.clm-glcMEC_long
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #588 
Fixes #620 

User interface changes?: 

Code review:


driver_cpl/cime_config/config_component.xml
```diff
@@ -2891,6 +2891,14 @@
     <desc>supplied or computed test id</desc>
   </entry>
 
+  <entry id="TEST_MEMLEAK_TOLERANCE">
+    <type>real</type>
+    <default_value>0.10</default_value>
+    <group>test</group>
+    <file>env_test.xml</file>
+    <desc>Expected memory usage growth for test</desc>
```
billsacks: Should note that this is _relative_ growth - e.g., 0.1 = 10% memory growth. (Right?)

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -289,10 +289,13 @@ def _check_for_memleak(self):
                 memdiff = -1
                 if originalmem > 0:
                     memdiff = (finalmem - originalmem)/originalmem
-
+                tolerance = self._case.get_value("TEST_MEMLEAK_TOLERANCE")
+                if tolerance is None:
+                    tolerance = 0.1
+                expect(tolerance > 0.0 and tolerance < 1.0, "Bad value for memleak tolerance in test")
```
billsacks: I don't think this should necessarily be less than 1. What if we expect a 3x growth in memory, from 100MB to 300MB? Wouldn't that be given as a tolerance of 2.0?

jedwards4b:
Thanks - I've updated based on your comments.

utils/python/CIME/XML/testlist.py
```diff
@@ -80,9 +86,14 @@ def _get_testsv2(self, machine=None, category=None, compiler=None):
                         else:
                             this_test[key] = value
 
+                    optionnodes = self.get_nodes("option", root=tnode)
```
billsacks: It looks to me like you're trying to say: Get all options attached to this specific machine/compiler (with the pre-existing code block below), and get all options that are at a higher level (with this new code block). However, this doesn't seem to work as intended: This new code block seems to pick up _all_ options, even those nested under a specific machine/compiler.

I tested this with this testlist:

```xml
<?xml version="1.0"?>
<testlist version="2.0">
  <test name="SMS" grid="ne30_g16" compset="X">
    <machines>
      <machine name="yellowstone" compiler="intel" category="bill">
      </machine>
      <machine name="yellowstone" compiler="pgi" category="bill">
        <options>
          <option name="wallclock">00:30:00</option>
        </options>
      </machine>
    </machines>
  </test>
</testlist>
```

And then:

```
./create_test --no-build --xml-category bill --test-root $scratch/tests_machopts_1004 --xml-testlist /glade/p/work/sacks/testing/testlist_machopts.xml
```

The result was that both the intel and pgi tests had their wallclock time set to 0:30.

I think that, for this new block of code, you need to somehow only find immediate children of tnode, rather than picking up any nodes rooted at tnode. Actually, that raises a question: Are these non-machine-specific options meant to be nested directly under the <test> node, so siblings of <machines>? Or are they meant to appear within the <machines> block? The first option would be most intuitive to me, and would agree with my suggestion of finding immediate children of tnode. But testlist_allactive uses the second.

However this is done: I'd suggest backing out this change here, and submitting a separate PR with it. That would keep this current PR clean and simple (and ready to go). My main rationale is that this change will potentially change behavior for many of our tests - because the options defined in testlist_allactive will begin to be applied, whereas they weren't before. So if this causes any problems, it would be good to have this change isolated to its own PR.

jedwards4b: thanks - fixed

billsacks:
This looks good now, but it looks like you'll need to resolve conflicts before merging to master.

jgfouca:
@jedwards4b this will need to be remerged.

Issue #624, opened 2016-09-30
manage_testlists doesn't work with the new test xml format


billsacks:
The manage_testlists tool in scripts doesn't work with the new test xml format. This tool isn't quite as critical for the new format, because it's easier to hand edit, but it still provides some nice querying capabilities that would be good to maintain.

So, updating this tool isn't critical, but would be good to do at some point.

billsacks:
The most important functionality is the query functionality... I think this will be pretty straightforward to recreate, so I'll take a stab at that.

PR #634, opened 2016-10-04
Utility to list tests from testlist xml files

billsacks:
This PR adds a new script, query_testlists, in cime/scripts/. This is
similar to the older manage_testlists, but only providing the query
functionality (along with related functionality --list and --count). The
motivation for creating this new tool is that the old manage_testlists
doesn't work with the new testlist xml format; this new tool works with
both (implicitly, because it simply calls other new code that works with
both formats).

Test baseline: n/a
Test namelist changes: n/a
Test status: bit for bit

Fixes: #624 
(Addresses that issue sufficiently.)

User interface changes?: none

Code review:

### DIFF
--- a/cime_config/cesm/allactive/testlist_allactive.xml
+++ b/cime_config/cesm/allactive/testlist_allactive.xml
@@ -303,8 +303,8 @@
   </test>
   <test name="PFS" grid="ne30_g16" compset="B1850" testmods="allactive/default">
     <machines>
-      <machine name="yellowstone" compiler="intel " category="prealpha"/>
-      <machine name="yellowstone" compiler="intel " category="prebeta"/>
+      <machine name="yellowstone" compiler="intel" category="prealpha"/>
+      <machine name="yellowstone" compiler="intel" category="prebeta"/>
       <options>
 	<option name="wallclock"> 00:20 </option>
       </options>
--- a/cime_config/cesm/allactive/testlist_allactive.xml
+++ b/cime_config/cesm/allactive/testlist_allactive.xml
@@ -285,7 +285,7 @@
   </test>
   <test name="PFS" grid="f09_g16" compset="B1850" testmods="allactive/default">
     <machines>
-      <machine name="bluewaters" compiler="pgi " category="prebeta"/>
+      <machine name="bluewaters" compiler="pgi" category="prebeta"/>
       <machine name="yellowstone" compiler="intel" category="prealpha"/>
       <machine name="yellowstone" compiler="intel" category="prebeta"/>
       <options>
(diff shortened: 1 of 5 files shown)


cime_config/cesm/allactive/testlist_allactive.xml
```diff
@@ -285,7 +285,7 @@
   </test>
   <test name="PFS" grid="f09_g16" compset="B1850" testmods="allactive/default">
     <machines>
-      <machine name="bluewaters" compiler="pgi " category="prebeta"/>
+      <machine name="bluewaters" compiler="pgi" category="prebeta"/>
```
billsacks: This and some other similar changes to the testlists are just removing whitespace from the end of compiler names, which I noticed when testing the new utility. I'm not sure if these changes will affect testing: it's _possible_ that some tests will start being run that weren't being run before.

utils/python/CIME/test_utils.py
```diff
@@ -47,6 +47,6 @@ def get_tests_from_xml(xml_machine=None,xml_category=None,xml_compiler=None, xml
                                                          testmod=None if "testmods" not in test else test["testmods"])
             logger.debug("Adding test %s with compiler %s"%(test["name"], test["compiler"]))
         listoftests += newtests
-        logger.info("Found %d tests"% len(listoftests))
+        logger.debug("Found %d tests"% len(listoftests))
```
billsacks: I changed this logging level from info to debug to avoid cluttering the output. Personally, I didn't find this output very useful without the other debug output in this function, so it seemed safe to suppress. But if you feel it should be kept in, then I can figure out some other way to suppress the output when this is called from query_testlists.

utils/python/tests/scripts_regression_tests.py
```diff
@@ -1705,6 +1705,34 @@ def xml_to_tester(self, xml_string):
...
+                              (SCRIPT_DIR, testlist_allactive, extra_args))
+
+    def test_query_testlists_runs(self):
+        """Make sure that query_testlists runs successfully
+
+        This simply makes sure that query_testlists doesn't generate any errors
+        when it runs. This helps ensure that changes in other utilities don't
+        break query_testlists.
+        """
+        self._run_and_assert_query_testlist()
```
billsacks: I generally like to do more than simple "smoke" tests like this, but these simple smoke tests seemed better than nothing, and I was having trouble justifying spending the time writing more detailed tests for this less-critical code.

cime_config/cesm/allactive/testlist_allactive.xml
fischer-ncar: The tests with the whitespace at the end of the compiler name were not running.

PR #637, opened 2016-10-05
Show comments and options in query_testlists

### LINKED ISSUE #627: Want better documentation for new testlist xml format
Moving some comments from #620 

Can someone document how to do the following with the testlist xml files:

1. Add a comment that appears when using the tool that queries testlists (i.e., not just a comment in the xml, but something with a special comment tag). I don't see any examples of this in the new xml formatted files. Incidentally, this would be another tag that is good to have nested under a specific machine / compiler / category, because my comments are often specific to that level.

2. Have the exact same test appear in multiple categories: Is the correct way to do this to have two lines that are identical except for their category, or to have something like ?

3. Similar question to (2), but how do you have the same test run with multiple compilers? Again, is the expectation that there should be multiple lines or one machine line with a '|' in the match?

4. Document the different ways you can specify options - either nested under a machine tag or not.

I'm not sure what the best place would be for this documentation, but testlist.py seems like a reasonable option.

cc @jedwards4b

billsacks:
Adds functionality to query_testlists to:

(1) Show any 'comments' entries (always)

(2) Show other test options, if --show-options is given

Output will look like this:

```
prebeta : ERI.f09_g16.B1850Ws.bluewaters_pgi.allactive-defaultio                       # wallclock:  00:20
prebeta : ERI.f09_g16.B1850Ws.edison_intel.allactive-default                           # wallclock:  00:20
prebeta : ERR_N3.f19_g16.B1850.edison_intel.allactive-defaultio                        # wallclock:  00:20
prebeta : ERS.f19_g16.B1850C4L45BGCRBPRPWs.bluewaters_pgi.allactive-defaultio
```

Each separate option is separated by a '#' character. If there is a comment, this will appear first.

Test baseline: n/a
Test namelist changes: n/a
Test status: bit for bit

Fixes #627 
(I have addressed points 1 and 4 in that issue; punting on points 2 and 3)

User interface changes?: none, except new option in query_testlists

Code review:

### DIFF
--- a/utils/python/CIME/XML/testlist.py
+++ b/utils/python/CIME/XML/testlist.py
@@ -2,10 +2,36 @@
 Interface to the config_files.xml file.  This class inherits from generic_xml.py
 It supports versions 1.0 and 2.0 of the testlist.xml file
 
-In version 2 of the file options can be specified to further refine a test or set of tests
-they can be specified at the <machines> level or at the level of a particular machine.
-currently supported options are walltime which sets the wallclock time for a given test
-in the queing system, and memleak_tolerance which specifies the relative memory growth expected for a given test.
+In version 2 of the file options can be specified to further refine a test or
+set of tests. They can be specified either at the top level, in which case they
+apply to all machines/compilers for this test:
+
+<test ...>
+  <options>
+    <option name="wallclock">00:20</option>
+  </options>
+  ...
+</test>
+
+or at the level of a particular machine/compiler:
+
+<test ...>
+  <machines>
+    <machine ...>
+      <options>
+        <option name="wallclock">00:20</option>
+      </options>
+    </machine>
+  </machines>
+</test>
+
+Currently supported options are:
+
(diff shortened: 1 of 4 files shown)


utils/python/CIME/XML/testlist.py
```diff
@@ -2,10 +2,36 @@
 Interface to the config_files.xml file.  This class inherits from generic_xml.py
 It supports versions 1.0 and 2.0 of the testlist.xml file
 
-In version 2 of the file options can be specified to further refine a test or set of tests
-they can be specified at the <machines> level or at the level of a particular machine.
-currently supported options are walltime which sets the wallclock time for a given test
-in the queing system, and memleak_tolerance which specifies the relative memory growth expected for a given test.
+In version 2 of the file options can be specified to further refine a test or
```
billsacks: Changes to this file are just in the docstring, comments and whitespace

billsacks:
This change was made at the suggestion of @jedwards4b

PR #641, opened 2016-10-06
Add xsd and test xml

jedwards4b:
Defines and implements a schema definition for entry_id style xml files.   Tests all config_component.xml files against this schema.  This is the first step in a plan to add schema testing for all xml files used by cime.    Some changes will be required to component config_component.xml files before they can be updated to the new cime however these reformatted files should be backward compatible with older cime versions.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes

User interface changes?: 

Code review:


jgfouca:
Looks good, this will be nice to have.

Issue #527, opened 2016-09-09
Remove PEA from cime_developer


rljacob:
cime_developer should run everywhere.
Not everywhere has mpi-serial
so remove PEA.

jedwards4b:
I don't think that we should remove this test from the suite because a few platforms don't support it, perhaps we can skip the test on those platforms?

rljacob:
"cime_developer should run everywhere"  or its not a good developer test suite.  Maybe PEA can be modified to use the default mpi library if mpi-serial is not available.

jedwards4b:
PEA is explicitly a comparison of mpi-serial to mpi build it doesn't make any sense for it to use the default mpi library.

rljacob:
Ok.  Then it should be removed.  For porting, cime_developer is also the suite that MUST pass to verify you've ported CIME correctly.  But we don't want to force people to get mpi-serial working.

jgfouca:
@rljacob as best as I can tell, there was never a PEA test in cime_developer. Did you mean acme_developer?

rljacob:
Sorry it wasn't PEA. It was SMS_D_Ln9_Mmpi-serial.f19_g16_rx1.A.

jgfouca:
Looks like @jedwards4b already did this.

PR #647, opened 2016-10-06
PEA should not always fail on machines without mpi-serial

jgfouca:
Test suite: Pylint (PEA not currently tested)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: No

Code review: jedwards

### DIFF
--- a/utils/python/CIME/SystemTests/pea.py
+++ b/utils/python/CIME/SystemTests/pea.py
@@ -9,6 +9,7 @@ Builds runs and compares a single processor mpi model to a model built using mpi
 from CIME.SystemTests.system_tests_compare_two import SystemTestsCompareTwo
 from CIME.XML.standard_module_setup import *
 from CIME.case_setup import case_setup
+from CIME.XML.machines import Machines
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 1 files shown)


jgfouca:
@jedwards4b , thoughts?

billsacks:
I don't like the idea of forcing a test to do nothing under some circumstances: It will be hard to determine what's actually being tested. ("Our PEA tests have all been passing.... Oh, right, they aren't actually testing anything.") It would be better to (a) not run the PEA test on such machines, or (b) let it fail and mark it as an expected failure.

jedwards4b:
Please resolve conflicts and merge

rljacob:
Is this related to issue #527 ?    What we really need is a rule: "don't put PEA in a test suite that must run on every machine".  Because not every machine can run mpi-serial.  If we can do that, we don't need this PR.

jgfouca:
Yes, this PR was initially related to 527 which turned out to be something else. I don't feel strongly either way, @rljacob you can be the tie-breaker. Do you think PEA should fail if the system doesn't support mpi-serial?

jedwards4b:
What about having the test in test suites but add code to create_test to skip it if the machine doesn't support mpi-serial?  I think that might be better than having the test do nothing and it's certainly better that not having it in suites.

rljacob:
I'm not sure I like that idea either because its sets a bad precedent.  create_test should just run tests and not know anything about them.

If we want PEA to fail when mpi-serial doesn't exist, that's a different kind of FAIL then mpi-serial gives different answers.  Can we introduce an INVALID outcome?

jgfouca:
@rljacob we could but that's a major change. If most machines that we test on support mpi-serial, then I'd prefer the changes in my PR. This will make the test show PASS on the few machines without mpi-serial so developers there won't have to dig into errors that are not really errors.

billsacks:
If people feel this is the best solution, I'm willing to let this go as long as it doesn't become a slippery slope.

billsacks:
As I think about this more: It strikes me that the PEA test is really more about testing the mpi-serial library than it is about testing cime/CESM/ACME: It's saying, "Is the mpi-serial library a valid substitute for a real mpi library?" In that light, these machine-specific failures are real signals that the answer is, in some cases, "no": mpi-serial is NOT always a valid substitute. In that sense, it seems like the correct solution would be to (eventually) fix the mpi-serial library. Although I could be misunderstanding these failures entirely, in which case feel free to ignore me :-)

I'm still okay with the current solution, I guess, acknowledging that it isn't ideal but is expedient.

rljacob:
The fails that prompted this were the mpi-serial library failed to even build.

jgfouca:
Merge?

jedwards4b:
I'm okay with it.

billsacks:
Me too

PR #648, opened 2016-10-06
Separate baseline and non-baseline compare phases

### LINKED ISSUE #396: TestStatus: Still would like clearer distinction between test comparisons and baseline comparisons

jgfouca:
The new BASELINE phase does not need decorators because there is
at most one per test.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #396 

User interface changes?: Yes, new BASELINE phase will appear in TestStatus

Code review: jedwards, billsacks

### DIFF
--- a/scripts/Testing/testreporter.pl
+++ b/scripts/Testing/testreporter.pl
@@ -305,7 +305,7 @@ sub getTestStatus
         $teststatushash{$testcase}{'comment'} .= " @testsummary_arr[3..$#testsummary_arr]";
 
         #Compare to baseline
-        my @testsummarylines = grep { /COMPARE_baseline/} @statuslines;
+        my @testsummarylines = grep { /BASELINE/} @statuslines;
         my @testsummary_arr = (split(/\s+/, $testsummarylines[0]));
         $teststatushash{$testcase}{'compare'} = ($#testsummarylines > -1) ? @testsummary_arr[0] : "----";
         $teststatushash{$testcase}{'comment'} .= " @testsummary_arr[3..$#testsummary_arr]";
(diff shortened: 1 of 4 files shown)


jgfouca:
Testing is ongoing. Just review for now.

PR #660, opened 2016-10-10
Mfdeakin sandia/system tests compare two/pem

mfdeakin-sandia:
This makes the PEM test class a subclass of SystemTestsCompareTwo and utilizes it's framework to implement the test 

Test suite: Most of cime_developer
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 446

User interface changes?: No

Code review:

### DIFF
--- a/utils/python/CIME/SystemTests/pem.py
+++ b/utils/python/CIME/SystemTests/pem.py
@@ -37,25 +37,11 @@ class PEM(SystemTestsCompareTwo):
             ntasks = self._case.get_value("NTASKS_%s"%comp)
             if ( ntasks > 1 ):
                 self._case.set_value("NTASKS_%s"%comp, ntasks/2)
-
-    def _pem_first_phase(self):
-        stop_n = self._case.get_value("STOP_N")
-        stop_option = self._case.get_value("STOP_OPTION")
-
-        self._case.flush()
-
-        self.run_indv()
-
-    def _pem_second_phase(self):
-        stop_n = self._case.get_value("STOP_N")
-        stop_option = self._case.get_value("STOP_OPTION")
-
-        self._case.flush()
-
-        self.run_indv(suffix="modpes")
-        self._component_compare_test("base", "modpes")
+        case_setup(self._case, clean=False, test_mode=False, reset=True)
 
     def _config_run(self):
+        # These options must be set in the run phase,
+        # as their environment variables don't exist before then
         for f in (self._activate_case1, self._activate_case2):
             f()
             self._case.set_value("HIST_OPTION","$STOP_OPTION")
@@ -63,5 +49,4 @@ class PEM(SystemTestsCompareTwo):
 
     def run_phase(self):
         self._config_run()
-        self._pem_first_phase()
-        self._pem_second_phase()
+        SystemTestsCompareTwo.run_phase(self)
--- a/utils/python/CIME/SystemTests/pem.py
+++ b/utils/python/CIME/SystemTests/pem.py
@@ -36,7 +36,7 @@ class PEM(SystemTestsCompareTwo):
         for comp in ['ATM','CPL','OCN','WAV','GLC','ICE','ROF','LND']:
             ntasks = self._case.get_value("NTASKS_%s"%comp)
             if ( ntasks > 1 ):
-                self._case.set_value("NTASKS_%s"%comp, ntasks/2)
+                self._case.set_value("NTASKS_%s"%comp, int(ntasks/2))
         case_setup(self._case, clean=False, test_mode=False, reset=True)
 
     def _config_run(self):
--- a/utils/python/CIME/SystemTests/pem.py
+++ b/utils/python/CIME/SystemTests/pem.py
@@ -28,8 +28,6 @@ class PEM(SystemTestsCompareTwo):
         self._case.set_value("SMP_BUILD","0")
         self._case.set_value("CONTINUE_RUN",False)
         self._case.set_value("REST_OPTION","none")
-        self._case.set_value("HIST_OPTION","$STOP_OPTION")
-        self._case.set_value("HIST_N","$STOP_N")
 
     def _case_one_setup(self):
         pass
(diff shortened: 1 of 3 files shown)


utils/python/CIME/SystemTests/pem.py
```diff
@@ -12,119 +12,41 @@
...
+    def _case_one_setup(self):
+        pass
 
-        self._case.flush()
-        logger.info("doing an %d %s initial test, no restarts written" % (stop_n, stop_option))
+    def _case_two_setup(self):
+        for comp in ['ATM','CPL','OCN','WAV','GLC','ICE','ROF','LND']:
+            ntasks = self._case.get_value("NTASKS_%s"%comp)
+            if ( ntasks > 1 ):
+                self._case.set_value("NTASKS_%s"%comp, ntasks/2)
```
jedwards4b: What happens here if ntasks is odd? I think you need int(ntasks/2)

utils/python/CIME/SystemTests/pem.py
```diff
@@ -12,119 +12,41 @@
...
+        case_setup(self._case, clean=False, test_mode=False, reset=True)
 
-        self.run_indv(suffix="modpes")
-        self._component_compare_test("base", "modpes")
+    def _config_run(self):
+        # These options must be set in the run phase,
+        # as their environment variables don't exist before then
+        for f in (self._activate_case1, self._activate_case2):
+            f()
+            self._case.set_value("HIST_OPTION","$STOP_OPTION")
```
jedwards4b: The STOP_OPTION and STOP_N are defined in XML and should not be retreved from the environment, perhaps what you need here is ("HIST_OPTION","\$STOP_OPTION")

mfdeakin-sandia: I guess my comment is misleading - I assumed they were environment variables based on the $ sign. I don't know if  case will look for them elsewhere

utils/python/CIME/SystemTests/pem.py
mfdeakin-sandia: Good point, and a potential bug fix from the original

utils/python/CIME/SystemTests/pem.py
jedwards4b: Actually I believe that the correct place for these settings is in cime_config/config_tests.xml  - that would be more consistent with other tests.

jedwards4b:
Please move the HIST_N and HIST_OPTION settings to config_tests.xml and run the full scripts_regression_tests as documented in https://github.com/ESMCI/cime/wiki/Testing

utils/python/CIME/SystemTests/pem.py
mfdeakin-sandia: How do I move them to config_tests.xml? Is there an interface I can use to do so?

jedwards4b: No interface, you can edit the file with emacs or vi and look for 
```
  <test NAME="PEM">                                                             
    <DESC>pes counts mpi bfb test (seq tests; default length)</DESC>            
    <INFO_DBUG>1</INFO_DBUG>                                                    
    <BFBFLAG>TRUE</BFBFLAG>                                                     
    <REST_OPTION>never</REST_OPTION>                                            
    <DOUT_S>FALSE</DOUT_S>                                                      
  </test>       
``` 
just add in that test section the following:
```
    <HIST_OPTION>$STOP_OPTION</HIST_OPTION>                                     
    <HIST_N>$STOP_N</HIST_N>     
```

mfdeakin-sandia: Alright, thanks.
Though when the test case is run, will it automatically substitute $STOP_OPTION and $STOP_N?

jedwards4b: yes

utils/python/CIME/SystemTests/pem.py
```diff
@@ -12,119 +12,29 @@
...
-        machpes1 = os.path.join("LockedFiles","env_mach_pes.PEM1.xml")
-        if ( os.path.isfile(machpes1) ):
-            shutil.copy(machpes1,"env_mach_pes.xml")
-        else:
-            logging.warn("Copying env_mach_pes.xml to %s"%(machpes1))
-            shutil.copy("env_mach_pes.xml", machpes1)
+        SystemTestsCompareTwo.__init__(self, case, True)
 
+    def _common_setup(self):
         self._case.set_value("SMP_BUILD","0")
```
billsacks: I strongly suspect that the setting of SMP_BUILD to 0 is no longer needed if you use this clone-based infrastructure. Though I can't be 100% sure since I don't actually understand what this setting does :-)

utils/python/CIME/SystemTests/pem.py
```diff
@@ -12,119 +12,29 @@
...
-
-        exeroot = self._case.get_value("EXEROOT")
-        cime_model = CIME.utils.get_model()
-        exefile  = "%s/%s.exe"%(exeroot,cime_model)
-        exefile1 = "%s/%s.exe.PEM1"%(exeroot,cime_model)
-        if (os.path.isfile(exefile)):
-            os.remove(exefile)
-        shutil.copy(exefile1, exefile)
-
         self._case.set_value("CONTINUE_RUN",False)
```
billsacks: To be consistent with other (newer) tests, you should move the settings of CONTINUE_RUN and REST_OPTION to config_tests.xml, too. Then - together with my previous comment - you could remove _common_setup entirely.

mfdeakin-sandia:
As Jim noted in issue 659, I'm actually not able to run all of scripts_regression_tests on Skybridge right now, but as far as I can tell, everything is currently working. Running a PEM test produces the same netcdf file as master.

jedwards4b:
How about pylint (B_CheckCode) and the unit tests?   I can accept if those have been run.  
Also single_submit is an option, not a required feature and should not be used in testing.

mfdeakin-sandia:
A_RunUnitTests is successful. I'm working on getting a version of pylint on my machine which works with the tests; though this is looking harder than I initially expected.

mfdeakin-sandia:
B_CheckCode is passing now

jgfouca:
Sorry for the late review. This looks like a good improvement.

PR #668, opened 2016-10-12
Remove support for user setting CIMEROOT.

### LINKED ISSUE #657: CIMEROOT environment variable used too aggressively
If you have CIMEROOT defined in your environment (or, presumably, in your .cime_config), it appears that any scripts you run from cime load at least some things from the directory referenced from CIMEROOT, rather than the current directory.

I tested this by setting:

```bash
export CIMEROOT=/path/does/not/exist
```

and then trying to run create_test from cime/scripts. This gave:

```
  File "./create_test", line 11, in <module>
    from Tools.standard_script_setup import *
  File "/glade/p/work/sacks/cime/scripts/Tools/standard_script_setup.py", line 15, in <module>
    import CIME.utils
ImportError: No module named CIME.utils
```

(similar things happen for create_newcase, etc.)

I'm not really concerned about this exact case (where CIMEROOT points to a non-existent directory). I was more using this as a measure of whether standard scripts like create_newcase and create_test try to load anything out of CIMEROOT - and apparently they do.

It's unclear to me, then, what comes from CIMEROOT and what comes from the current directory. 

This could be very confusing and cause incorrect results in situations like this: Someone has CIMEROOT set to point to some cime clone, and they also have a CESM checkout with a different version of cime. They run create_newcase from their CESM checkout. Would this actually be using some / all things from CIMEROOT rather than from the current directory? This, in my opinion, would be a Bad Thing.

I forget the reason why CIMEROOT was introduced originally. But I'm having trouble seeing any reason why you would ever want to run a script out of cime directory A, and have it use pieces from cime directory B. It feels like this creates potential for significant confusion and unexpected behavior.

jgfouca:
CIMEROOT will always be the home repo of where the tool lives.

Also, as I was scanning for usages of CIMEROOT, I found several files
that I strongly suspected are obsolete and I removed them.

Test suite: scripts_regression_tests (both on batch system and desktop)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #657

User interface changes?: CIMEROOT is no longer exposed to user

Code review: jedwards billsacks


jgfouca:
Skybridge is still finishing testing but I put this up so we could begin review. My main concern is making sure that the files I deleted are not being used anymore.

billsacks:
I don't feel a need to review this if Jim E is good with it. My one question is whether anything in any of the CESM (or ACME) components will need to be modified with this change. @jedwards4b , do you know? For example, I see various references to the CIMEROOT environment variable in CLM (components/clm), which @ekluzek should be aware of if he isn't already:

Accessing CIMEROOT:

```
./cime_config/buildlib:8:_CIMEROOT = os.environ.get("CIMEROOT")
./cime_config/buildlib:9:if _CIMEROOT is None:
./cime_config/buildlib:10:    raise SystemExit("ERROR: must set CIMEROOT environment variable")
```

(There are other references via xmlquery, but I'm thinking that's still safe?)

As well as setting the CIMEROOT environment variable:

```
./test/tools/TCBCFGtools.sh:79:env       CIMEROOT=$CLM_ROOT/cime COMPILER=$CESM_COMP $config_string $CLM_ROOT/cime/tools/configure --macros-format Makefile --machine $CESM_MACH $TOOLS_CONF_STRING >> test.log 2>&1
./test/tools/TCBtools.sh:88:         env CIMEROOT=${CLM_ROOT}/cime ${CLM_ROOT}/cime/tools/configure -mach bluefire >> test.log 2>&1
```

```
./tools/shared/PTCLM/buildtools:107:      export CIMEROOT=$CESM_ROOT/cime
```

and various places in the tools documentation, in tools/README and tools/shared/gen_domain/INSTALL

jgfouca:
@billsacks Yes, tools getting CIMEROOT from the environment will need to change. xmlquery is a safe way to get it.

jedwards4b:
That's a problem since there isn't always an existing case when using tools.

jedwards4b:
I think that CIMEROOT in the enviroment should be allowed for a limited number of situations, 
using case independent tools is one of those and unless someone is going to fix all of the external components before this PR is accepted, that's another one.

jgfouca:
There's nothing preventing it, it just isn't supported by anything in CIME.

rljacob:
case-independent tools require an existing case?

jedwards4b:
No they shouldn't.

jedwards4b:
But looking closely at Jim's change I see that it will work correctly independent of a case. 
I'm happy with this PR.

jgfouca:
So, the key changes in this script were the removal of os.environ["CIMEROOT"] = ... from case.setup and case.submit. The one in case.build I had to keep due to the nature of our build system. I could restore these in order to support buildnml/buildlib programs from components outside of CIME. In general, any tool within CIME should know where it is in the repo and can set it's own _CIMEROOT. Any tool outside of CIME can locate CIME any way it wants including putting CIMEROOT in the env.

jgfouca:
For what it's worth, I assume CIME is snapshotted into CESM in a known location, so any CESM tool should also be able to find CIME without having to use CIMEROOT.

jedwards4b:
No, I don't want to use that method - I often have two or three cime directories checked out into
cesm at the same time for testing and don't want to have the components expect a location.

jgfouca:
OK, so you do want to use an environment variable. So I should restore CIMEROOT in case.setup and case.submit? This won't impact case.get_value("CIMEROOT") will it?

jedwards4b:
Yes case.get_value should always get what is in the case xml.   
I would say put it back in case.setup and case.submit but I didn't open the issue.

jgfouca:
OK, then I think we all agree that CIMEROOT (the env var) should have no impact within CIME but can be used as a means for outside tools to link up with CIME.

jedwards4b:
👍

billsacks:
I'm not completely following this discussion, but I think I'm happy as long as the following is true: If I have the environment variable CIMEROOT set to point to some non-existent location (or, say, a completely broken cime clone), that should have 0 impact on the running of anything out of cime, or on the operation of any case.

jgfouca:
@billsacks that will be the case once this is merged.

PR #674, opened 2016-10-13
Improve test status documentation

### LINKED ISSUE #663: Want better documentation for usage of test_status.py
In trying to understand what ensures that TestStatus flushes its results if any error occurs, I realized that the key seems to be the use of all TestStatus objects via a 'with' context manager. This required usage (if correct) should be documented in test_status.py.

@jgfouca can you add appropriate usage documentation at some point?

jgfouca:
Test suite: code_checker test_status.py
Test baseline: 
Test namelist changes: 

Fixes #663

User interface changes?: None

Code review: billsacks


billsacks:
Looks good - thanks!

PR #681, opened 2016-10-17
Map irrigation flux normalized by volr

billsacks:
Irrigation in LND withdraws water from the rivers in ROF. Up until now,
the irrigation flux (a negative flux from LND to ROF) has been mapped
like all other fluxes. However, Sean Swenson realized that this leads to
problems, because ROF cells with very little water volume are drawn down
just as much as ROF cells with high water volume, increasing the
likelihood that irrigation will attempt to draw ROF volumes negative.

With the changes here, a custom mapping routine is used to remap
the irrigation flux from LND to ROF: A normalization factor of VOLR (the
river volume) is used in performing this mapping.

In order for this to work as intended, changes are needed in both LND
(CLM) and ROF (RTM / MOSART). However, there is no harm in using old
versions of the component models with this new cime code: There simply
will not be any special treatment of irrigation in that case.

At the suggestion of @mvertens, I have put the addition of the special
irrigation field inside a 'cesm' conditional in seq_flds_mod. While not
strictly necessary, this avoids confusion, particularly in the budgets
output, and eliminates the performance overhead of the new irrigation
mapping.

Also, some changes to the Fortran unit tests.

Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

By itself, this change is bit-for-bit. However, when combined with the
relevant component model changes, this changes answers for runs with
irrigation turned on.

Also ran many tests out of the aux_clm40, aux_clm45, rtm and mosart test
suites in CLM, with the new versions of the CLM, RTM and MOSART
component models.

Fixes: none

User interface changes?: none

Code review: @mvertens did a partial review of a slightly earlier version

@bishtgautam - you may want to be aware of this for the sake of ACME land

also cc @ekluzek


billsacks:
I ran scripts_regression_tests Z_FullSystemTest on yellowstone/caldera, with CIME_MODEL=acme. While some tests failed due to #684 , the tests that passed on master:

```
PASS ERS_Ld3.ne30_g16_rx1.A.caldera_intel RUN
PASS ERI.f45_g37.X.caldera_intel RUN
PASS ERR_Ld3.f45_g37_rx1.A.caldera_intel RUN
```

passed here as well. Interestingly (but maybe with no real meaning), this one also passed here:

```
PASS NCK_Ld3.f45_g37_rx1.A.caldera_intel RUN
```

So, to my best determination, this commit doesn't break things for acme.

driver_cpl/driver/map_lnd2rof_irrig_mod.F90
```diff
@@ -0,0 +1,287 @@
+module map_lnd2rof_irrig_mod
```
jedwards4b: Why did this code go into the driver directory instead of driver_cpl/shr?

driver_cpl/driver/seq_diag_mct.F90
```diff
@@ -880,6 +883,9 @@ subroutine seq_diag_lnd_mct( lnd, frac_l, infodata, do_l2x, do_x2l)
                                                                     - dl*l2x_l%rAttr(index_l2x_Flrl_rofgwl,n) &
                                                                     - dl*l2x_l%rAttr(index_l2x_Flrl_rofsub,n) &
                                                                     - dl*l2x_l%rAttr(index_l2x_Flrl_rofdto,n)
+         if (index_l2x_Flrl_irrig /= 0) then
+            if = f_wroff ; budg_dataL(if,ic,ip) = budg_dataL(if,ic,ip) - dl*l2x_l%rAttr(index_l2x_Flrl_irrig,n)
```
jedwards4b: I really don't like the use of a variable 'if'   It took me a while to figure this code out - I know it's not your doing but can we change that?

driver_cpl/driver/map_lnd2rof_irrig_mod.F90
billsacks: My understanding is that driver_cpl/shr is for code needed by other components (as it is included in the share library, right?), whereas driver_cpl/driver is for code that is specific to the driver/coupler. This code is specific to the driver/coupler. In addition, since it has dependencies on seq_map_mod and seq_map_type_mod, which are in driver_cpl/driver, my impression is that it won't work to have this module in driver_cpl/shr. (My impression is that driver_cpl/shr can only depend on stuff in the share code library.)

But let me know if you feel differently on this.

driver_cpl/driver/seq_diag_mct.F90
billsacks: I agree completely. I'd suggest that this refactoring be done separately, though: I don't like mixing refactoring with behavior/answer-changing feature additions if I can help it.

billsacks:
@rljacob - do you want to review this before we bring it to master? No rush, just wondering if we should wait.

rljacob:
I'd like @bbye to look at this.

billsacks:
Sounds good. @bbye : I'm happy to chat if you have questions.

rljacob:
The "cesm" conditional means these changes won't affect ACME?  We'd also need the changes to CLM/RTM/MOSART?

billsacks:
> The "cesm" conditional means these changes won't affect ACME? 

Yes, that's right. Actually, even without that conditional, answers should be bit-for-bit identical in ACME, even with irrigation enabled, because the new, specially-mapped irrigation flux would be 0 (irrigation would all go in other runoff fluxes, as it always has). The conditional just helps:

1. Make it more clear that this doesn't apply to ACME

2. Avoid unused terms in the budgets

3. Avoid the performance penalty of this special mapping.

> We'd also need the changes to CLM/RTM/MOSART?

Yes. The CLM changes would need to be reimplemented because irrigation has been completely reworked in CESM's CLM.

rljacob:
Thanks.  We're ok with this.

bbye:
I took a look and I am OK with this merge, although I admit I haven't worked with the irrigation code yet.

PR #682, opened 2016-10-17
In addition to printing the commit, print the current cime_model

billsacks:
Today I had to run a bunch of tests both with CIME_MODEL=cesm and with
CIME_MODEL=acme. I had a hard time remembering which test suite was run
with each. This change helps with that.

However, I'm not sure if this breaks anything that parses the tested
commit (do we have anything like that?).

Test suite: start of scripts_regression_tests.py, both with cesm and acme
   aborted after the first few tests, because it didn't seem important to run it further
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

Fixes: None

User interface changes?: No

Code review:

### DIFF
--- a/utils/python/tests/scripts_regression_tests.py
+++ b/utils/python/tests/scripts_regression_tests.py
@@ -52,12 +52,17 @@ ERRPUT: %s
     return output
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


jgfouca:
Nothing parses that output.

PR #683, opened 2016-10-17
Always print provenance in scripts_regression_tests

billsacks:
Print provenance info from main function, rather than in a 'test'

The main point of this is so that we still get provenance info even if
just running a subset of the unit tests

Test suite: ./scripts_regression_tests.py A_RunUnitTests
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

Fixes: None

User interface changes?: No

Code review:

### DIFF
--- a/utils/python/tests/scripts_regression_tests.py
+++ b/utils/python/tests/scripts_regression_tests.py
@@ -52,19 +52,6 @@ ERRPUT: %s
     return output
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


billsacks:
I'm not at all tied to this if there's some reason it was done as a "test".

jgfouca:
There was a reason for it to be a test (so that people could click on the dashboard and see which commit was tested) but that didn't work anyway, so I'm fine with this.

PR #686, opened 2016-10-18
Bugfix for failing C compset DEBUG tests

### LINKED ISSUE #685: water isotope code breaks intel DEBUG for C compset on yellowstone
The way water isotopes are implemented, when running without water isotopes (`flds_wiso = .false.` in drv_in) there are two calls to `seq_map_map()` with non-existent fields in the `fldslist` argument (calls occur in `prep_ocn_mod`). This causes some `_D` tests to fail on `yellowstone_intel` although the non-DEBUG tests run successfully. Pull request with a fix coming momentarily.

mnlevy1981:
Added a couple of new fields to track which r2o fluxes are needed

Test suite: no full test suite run, just verified previously failing ERS_D test now passes (and previously working ERS_D test continues to work)
Test baseline: n/a
Test namelist changes: none 
Test status: bit for bit

Fixes #685 

User interface changes?: none

Code review: 

When water isotopes were added to the coupler, the runoff -> ocean fields
changed -- this is fine in coupled and ocean / ice cases, but the new isotope
fields should not be used in ocean-only compsets. To get around this, I
introduced two new module variables (seq_flds_r2o_liq_fluxes and _ice_fluxes)
in seq_flds_mod.F90 that always contain 'Forr_rofl' and 'Forr_rofi',
respectively, and can also include the isotope fluxes if flds_wiso is .true.

The call to seq_map_map uses these new variables as the fldlist arguments in
calls from prep_ocn_calc_r2x_ox(), so that we don't run into the issue of
trying to map non-existent fields when water isotopes are not enabled.

### DIFF
--- a/driver_cpl/driver/prep_ocn_mod.F90
+++ b/driver_cpl/driver/prep_ocn_mod.F90
@@ -1200,10 +1200,10 @@ contains
     do eri = 1,num_inst_rof
        r2x_rx => component_get_c2x_cx(rof(eri))
        call seq_map_map(mapper_Rr2o_liq, r2x_rx, r2x_ox(eri), &
-            fldlist='Forr_rofl:Forr_rofl_16O:Forr_rofl_18O:Forr_rofl_HDO', norm=.false.)
+            fldlist=seq_flds_r2o_liq_fluxes, norm=.false.)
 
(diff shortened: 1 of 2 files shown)


mnlevy1981:
071f341 accounts for @billsacks comment

billsacks:
Okay, this looks good to me now.

Issue #643, opened 2016-10-06
Add DEBUGGER option to env_run.xml


ekluzek:
I'd like to add a DEBUGGER option to env_run.xml so that the mpirun command can be setup to run with a debugger easier than it is now. The abstraction of the run scripts, also makes it difficult for a user to even figure out how to go about connecting a debugger to a cime model. And the complexity of the cime models certainly makes it make sense to use a debugger as a tool to help with debugging the system.

So for example in the config_machine.xml for yellowstone you currently have...

```
<mpirun mpilib="mpi-serial">
  <executable></executable>
</mpirun>
<mpirun mpilib="default" threaded="false">
  <executable>TARGET_PROCESSOR_LIST=AUTO_SELECT mpirun.lsf $ENV{CESMDATAROOT}/tools/bin/launch </executable>
</mpirun>
<mpirun mpilib="default" threaded="true">
  <executable>unset MP_PE_AFFINITY; unset MP_TASK_AFFINITY; unset MP_CPU_BIND_LIST; mpirun.lsf $ENV{CESMDATAROOT}/tools/bin/hybrid_launch </executable>
</mpirun>
<mpirun compiler="gnu">
  <executable>mpirun.lsf </executable>
</mpirun>
```

We would add something like the following....

```
<mpirun mpilib="mpi-serial" debugger="ddt">
  <executable>ddt --connect</executable>
</mpirun>
<mpirun mpilib="default" debugger="ddt">
  <executable>ddt --connect mpirun.lsf </executable>
</mpirun>
```

It probably has to be setup individually for each machine, and for each debugger available (and wanted to be used on that machine). Having something available out of the box makes it easier to setup though, and all debuggers take some setup to work with CESM.

You could also add a DEBUGGERS list of available debuggers for a given machine. But, that's probably only important if we can show that having this option is useful to begin with.

@cacraigucar @fvitt @billsacks

ekluzek:
OK, here's some updates on this for yellowstone...

Index: cime_config/cesm/machines/config_machines.xml
===================================================================
--- cime_config/cesm/machines/config_machines.xml	(revision 11641)
+++ cime_config/cesm/machines/config_machines.xml	(working copy)
@@ -1261,9 +1261,25 @@
     <PES_PER_NODE>15</PES_PER_NODE>
     <PROJECT_REQUIRED>TRUE</PROJECT_REQUIRED>
     <mpirun mpilib="mpi-serial" debugger="ddt">
-      <executable></executable>
[...]

PR #689, opened 2016-10-18
Major refactor to handling of namelists

### LINKED ISSUE #650: Don't rebuild namelists if you don't have to
If namelists, XML, and sourcemods have not changed, do not rebuild namelists because it's somewhat expensive and produces lots of output.

jgfouca:
1) Removes namelist creation from setup and build phases!
 1.a) This reduces case.setup time on this machine from 12s to 1.5s
 1.b) Reduces the number of expensive namelist creations from 4 to 2 (case.submit, case.run) for an average test case
2) Split out namelist phase from test_scheduler.
 2.a) Allows user to invoke this phase from command line like all the other major phases
   2.a.i) New executable is case.cmpgen_namelists
 2.b) Namelist phase is now done during case.submit
3) Split preview_namelists into two pieces, one for directory setup (inexpensive), one to create namelists (very expensive)
 3.a) dir setup needs to be done early in case.setup phase
4) Remove --dryrun option from preview_namelists; it was broken and clearly not being used
5) Make it easier to load environment for case by adding API to case
 5.a) Also, no need to do anything if env is already loaded
6) Signficant refactor to compare_namelists implementation using OrderedDicts to preserve order
7) Don't use hardcoded model list in create_namelists
8) Remove namelist-related complexity from test_scheduler.py
9) Add some simple case.do_namelists tests to N_TestCreateTest

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #650

User interface changes?: Yes, new case.cmpgen_namelists tool

Code review: jedwards, billsacks, gold2718


jgfouca:
Good news: this PR appears to dramatically cut the time needed to run scripts_regression_tests. It went from ~2700 seconds to ~1700 seconds on melvin!

jgfouca:
One thing I was disappointed about what that I could not remove the namelist creation from case_run even though it generally adds very little to the namelist creation that happens in case_submit.

utils/python/CIME/preview_namelists.py
```diff
@@ -7,80 +7,59 @@
...
-                except OSError as e:
-                    expect(False, "Could not make directory '%s', error: %s" % (dir_to_make, e))
+    # Load modules
+    case.load_env()
 
     # Create namelists
+    models = case.get_values("COMP_CLASSES")
     for model in models:
-        model_str = "drv" if model == "cpl" else model
+        model_str = model.lower()
```
jedwards4b: I'm not sure how you could remove line 67?

jgfouca: Because "DRV" is in COMP_CLASSES

billsacks:
Thank you for taking this on! It's certainly nice to get this speedup. It also looks like you were able to do some nice cleanup of the tests, with respect to the namelist phase.

However: I am uncomfortable with namelist generation being removed from the case.setup phase without further discussion, including participation from @mvertens and other CSEG members. My impression is that users rely on having the initial namelists generated during this phase, to use as a starting point for modifications. I know there's a trade-off here (user convenience vs. testing time), but I think this warrants broader input. Personally, I feel that the removal from the build phase is less likely to mess users up... though even that may warrant some broader input.

Some additional comments, based on a high-level review (I'm not going to have time to do a detailed code review, but am happy to defer to others on that):

1. If you do keep the current behavior, then this output should be removed from case.setup:

    ```
    See ./CaseDoc for component namelists
    ```

2. I tried running preview_namelists in a case after running case.setup and got this error:

 ```
   File "./preview_namelists", line 11, in <module>
     from CIME.preview_namelists import preview_namelists
 ImportError: cannot import name preview_namelists    
 ```

 It looks like the preview_namelists user script has not been updated for the splitting of the underlying python functionality. What is your intention here in terms of the user interface? I'm guessing that this script should just call the namelist creation part, not the directory setup part?

3. I get a broken link in my case directory to case.do_namelists

4. I don't love the name 'do_namelists'. Based on what this is doing, it looks like something like 'case.compare_and_generate_namelists' would be more appropriate.

5. Regarding case.do_namelists: I feel that we shouldn't put a script like this in the case directory which only works for tests. This just clutters the case directory and adds confusion for users. Maybe only copy/link this in if you're doing a test?

Finally, a note regarding this:

> One thing I was disappointed about what that I could not remove the namelist creation from case_run even though it generally adds very little to the namelist creation that happens in case_submit.

Even if you _were_ able to remove this: I feel it's important to keep namelist generation in the run phase: I think users expect that any changes they make up until the time when the model starts running will be reflected in the run.

jedwards4b:
Overall this PR looks very good, I am running some additional tests with external components.

billsacks:
> 3. That's odd, I never saw this myself. Where is the link pointing to?

This was with both:

```
./create_newcase -case test_baseline_namelists_1019b -compset A -res f45_g37_rx1
```

and:

```
./create_test --no-build --test-root . SMS.f45_g37_rx1.A.yellowstone_intel
```

This points to:

```
lrwxrwxrwx  1 sacks ncar    56 Oct 19 05:48 case.do_namelists -> /glade/p/work/sacks/cime/scripts/Tools/case.do_namelists
```

> 4. Yeah, I didn't like the name either. How about case.cmpgen_namelists?

I personally prefer names that are spelled out - and with tab completion, that typically doesn't mean any more typing is needed. But I'm okay with your suggestion if you like it better.

But actually: I'm wondering how necessary this is. You said:

> 2) Split out namephase phase from test_scheduler.
> 2.a) Allows user to invoke this phase from command line like all the other major phases

But really, this namelist comparison & generation is analogous to the baseline comparison and generation. There's no script put in the case directory for those. That said, I guess it doesn't hurt to have this script, so maybe you should ignore what I just said :-)

> Users must go through case.submit to get to case.run, so there shouldn't be any opportunity for them to make changes in between except maybe in custom run_phase definitions for system test cases.

A typical usage that I see is: A user submits a run, then (before it actually starts) realizes that they forgot to add something to user_nl_*. So they add it just in time, and rely on the namelist regeneration in the run phase to pick it up.

I'm not saying that we necessarily _need_ to support that, but it may take some retraining if we stop supporting it.

> Due to the high cost of creating namelists, I'd prefer that we retrain users not to expect namelists after case.setup. If they want them, there's always preview_namelists. If I get a lot of push back on this, I may push for this to be customize-able via env or .cimeconfig

I certainly see your point. I'm on the fence about this myself. I just think this warrants broader input.

worleyph:
>> Due to the high cost of creating namelists, I'd prefer that we retrain users not to expect namelists after case.setup. If they want them, there's always preview_namelists. If I get a lot of push back on this, I may push for this to be customize-able via env or .cimeconfig

> I certainly see your point. I'm on the fence about this myself. I just think this warrants broader input.

Had to jump in - as a user who sees jobs sit in the queue for up to one week, not being able to modify namelists after submission would be a real downer. Having a different behavior for a test suite seems reasonable. but most users will not notice the cost of the namelist generation, since it is minimal compared to the cost of compliation, sitting in the queue, and job execution. Looks like you are designing for the software engineers, not the users. Having a "fast" option for testing is fine, but that should not be the only mechanism in my opinion.

Just curious - is user_nl_XXX support part of this? Would changes to these also be ignored after the submit?

As an aside, in the current setup, I can recompile, to fix bugs, between job submission and run. Not being able to do the same thing for namelists would seem peculiar. I have not tracked this discussion however - if there is a separate "regenerate namelists" command, similar to invoking a recompile, that would be sufficient for me - we would just need to educate people.

billsacks:
@worleyph - thanks for adding your experience here. There is a command to regenerate namelists (`preview_namelists` - as I'm typing this, I'm thinking that perhaps it should be renamed, because it's doing more than just 'previewing' the namelists). However, the question to me is whether we should require users to run that command at various points where, in the past, it has been run automatically. There is enough to learn and remember already in order to run CESM/ACME, and I worry that the more steps we require, the more (a) there will be mistakes (expecting changes to be picked up automatically, when in fact they aren't), and (b) people will get turned off and just give up on CESM entirely.

But I'm a worrier, so I don't want to unilaterally dictate anything :-)

jgfouca:
@worleyph I was not able to get rid of the namelist re-creation in case.run, so that usecase is still supported. The changes were removing the namelist creation in case.setup and case.build.

jgfouca:
@billsacks just to be clear, there is no requirement for users to run preview_namelists. They only need to run it if they want to inspect the namelists without running case.submit.

jgfouca:
@billsacks that symlink for case.do_namelists looks correct. Is the target missing for some reason?

billsacks:
> @billsacks just to be clear, there is no requirement for users to run preview_namelists. They only need to run it if they want to inspect the namelists without running case.submit.

Sorry, yes, I confused the issue by replying to Pat's thoughts. My bad.

billsacks:
> @billsacks that symlink for case.do_namelists looks correct. Is the target missing for some reason?

@jgfouca - yes that target is missing. Maybe you missed adding it to your commit?

rljacob:
I'd like to see us move to 2 modes of usage for the scripts: normal and expert.  At the script level, there might be a "cime.confandbuild" which, under the hood, calls "preview_namelist" and "case.build".  Normal users would call the single script, expert users could call the pieces individually when they know they need to be.

jgfouca:
@billsacks haha, yep

worleyph:
@jgfouca , thanks. I was reacting to

> I'd prefer that we retrain users not to expect namelists after case.setup. 

@billsacks , looks like I need to reread the whole thread to comment intelligently.  Again, "as a user", I'd also prefer not to have to run preview_namelist after editing user_nl_XXX, but personally could live with that. I would vote for changing the name though (generate_namelists?)

jgfouca:
I've resolved all of @billsacks requests except for 5 since he retracted that one later.

billsacks:
> I've resolved all of @billsacks requests except for 5 since he retracted that one later.

Thank you; looks good. But did I really retract 5? I thought that still stands: Ideally, case.cmpgen_namelists should only appear in tests, so as to avoid confusion.

Other than that, my only outstanding issue is the big one: I still feel that this should get some broader discussion. Unfortunately, that probably needs to wait a few weeks until @mvertens is back in the office; are you comfortable with this remaining open until then?

jgfouca:
@billsacks OK I will make case.cmpgen_namelists be like the baseline tools; they will live in scripts/Tools but won't get symlinked.

As far as sitting on this for weeks, I kinda hate the sound of that. I'd rather move forward and we can add create namelists back into case.setup (I think that was the only controversial change) if that's what people want.

jedwards4b:
Although I agree that further discussion and testing is warranted I don't think that we need to wait for @mvertens.   I've found several problems in testing but have not had a chance to look into them yet.  Please standby.

billsacks:
@jgfouca and @jedwards4b - I'm okay with this, I guess, but would like to propose a compromise to help keep things smoother for CLM development over the next few weeks: We need to update CLM to point to a newer version of CIME, but have been waiting to do so until CESM's beta03 testing is done, so we can point to a CIME version that has been through beta testing. Furthermore, I'd like to include #681 in the updated CLM, because this is needed for further science development; I am still waiting to hear anything from Beth on that one. At the same time, I'd like to avoid subjecting CLM scientists to this user interface change until we're sure that we're sticking with it.

So my proposed compromise is to bring this to master once:

1. CESM beta03 testing is complete
2. #681 comes to master

(I don't care much which order those two things happen in.)

Does that seem reasonable?

jedwards4b:
@billsacks works for me.

ekluzek:
I like Bill's proposal, on waiting a bit for this. I think I'm OK with the changes, but I do think it will impact users. One subtle shift that is going on is that the prestage of data is being moved to the case.submit time. So that means the first time you run submit it will be slow if it needs to prestage files. I also wonder if that change will slow down the submit in general?

jgfouca:
@billsacks fine with me too.

jedwards4b:
First problem: in config_machines.xml I have   <env name="MP_MPILIB">$MPILIB</env>
this is not being resolved from the setting in env_build.xml for MPILIB

jedwards4b:
.env_mach_specific files are not being updated when env_mach_specific.xml is changed

rljacob:
After reading this thread, I don't like all the heavy work being moved to case.submit.  case.submit is when everything is ready to go and its time to start a run.  Generating namelists, editing them, moving files in to place, that should all happen before case.submit.   Also, I still want the freedom to skip case.submit and just qsub the run script (I was going to make an issue for that).

bishtgautam:
What happens when a developer uses an interactive job to debug the code in DDT/TotalView without running the case.submit?

jgfouca:
@jedwards4b can you give me some more context? Are these errors caused by this PR? They look unrelated to anything I did.

jgfouca:
@rljacob case.submit was calling preview_namelists before this PR and so was already a slow command. The addition of the cmpgen_namelists to case.submit is almost free (takes less than a second). The move of check_input_data to case.submit will have no impact in the vast majority of use cases because you usually don't need to download input files. Also, putting the check_input_data in case.submit is more robust than doing in it case.build because it ensures that the input data is present when you're actually running the case.

As far as submitting case.run, maybe @jedwards4b can chime in here, but I don't think that's going to happen. The current system is designed for case.submit to be the entry point for running the case. I was even thinking about making case.run and case.test hidden files to make it very clear to users that they shouldn't run them by hand. Out of curiosity, why would you not want to run case.submit?

jgfouca:
@bishtgautam I'm not familiar with that usecase. Are you talking about running mpi directly? That shouldn't be impacted by this PR one way or the other.

rljacob:
I totally agree check_input_data shouldn't be in case.build.   By the way, can you update the PR description to reflect the new names?  I keep jumping back to that but not sure its accurate now.  What is the difference between cmpgen_namelists and preview_namelists?

jedwards4b:
@bishtgautam: Note that issue #643 will address your concern. 

@rljacob: Case.submit is the entry point and allows for running jobs without submission.   A clever user could submit case.run directly however I think that this usage should be discouraged.

@jgfouca:  You are correct that this issue with .env_mach_specific getting an unresolved value is present on master and not related to this PR.   However in master when I submit the job that variable is resolved and the run completes but on the branch it is not resolved and the run fails.

bishtgautam:
> @bishtgautam: Note that issue #643 will address your concern.

Sounds good to me. Thanks.

jgfouca:
@rljacob I have updated the names. case.cmpgen_namelists allows for blessing/comparison of current namelists against the namelists in the baselines. preview_namelists just creates namelists for the current case.

jgfouca:
@jedwards4b can you add tests that expose these failures to this PR?

rljacob:
When you say "Split preview_namelists into two pieces, one for directory setup (inexpensive), one to create namelists (very expensive)" is there also 2 new user commands available?

jgfouca:
@rljacob no, the command line interface just does both.

jedwards4b:
@jgfouca I have a fix for that issue, it's in 
To git@github.com:jedwards4b/cime.git
 * [new branch]      jgfouca/optimize_preview_namelists -> jgfouca/optimize_preview_namelists

jedwards4b:
Some of the external components in CESM rely on the buildnml script to create CPPDEFS for the build (both pop and cice do this) - this functionality is broken in this PR.   I see this behavior in test
SMS.T62_g16.CECO

jedwards4b:
I've added a solution to this problem to my fork of this branch, it only invokes create_namelist in the build phase if the component sets the flag %s_PE_CHANGE_REQUIRES_REBUILD" % comp
which is used in pop and cice to indicate that changing the pe layout requires a recompile, it also happens to indicate that buildnml will need to be rerun prior to the build.  Now case.build only runs build namelist for these external components that require it.

jgfouca:
Thanks all for the comments and @jedwards4b for testing and fixing. I didn't get as many create namelist calls removed as I want, but this is a good start.

PR #696, opened 2016-10-21
my mods to Jgfouca/optimize preview namelists

jedwards4b:

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/build.py
+++ b/utils/python/CIME/build.py
@@ -275,12 +275,9 @@ def case_build(caseroot, case, sharedlib_only=False, model_only=False):
     # Load modules
     case.load_env()
 
-    run_create_namelist = build_checks(case, build_threaded, comp_interface, use_esmf_lib,
+    build_checks(case, build_threaded, comp_interface, use_esmf_lib,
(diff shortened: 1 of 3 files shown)


jgfouca:
All ready?

rljacob:
Please fix the PR description and title.

utils/python/CIME/build.py
```diff
@@ -396,6 +397,10 @@ def build_checks(case, build_threaded, comp_interface, use_esmf_lib, debug, comp
     case.set_value("BUILD_COMPLETE", False)
 
     case.flush()
+    if not model_only:
```
jgfouca: This looks different than the call that was there before, is that on purpose?

jedwards4b: Yes, I realized that this needs to happen before the sharedlib build not after

jgfouca: OK, thanks

jedwards4b:
ready

PR #697, opened 2016-10-21
Sharedlib build installs to case

### LINKED ISSUE #690: Occasional build failures with large test suites
I have been running into occasional build failures when running the large aux_clm test suite on yellowstone. Here, "occasional" means: happens for every test suite, though just for a small fraction of the tests. The failed tests work when I rebuild them. I'd be tempted to chalk this up to a system problem, except that it was quite rare to encounter problems like this prior to cime5 - whereas I am commonly running into this with cime5: I have seen this in all 3 of my last 3 runs of the test suite (which I think have been my only runs of the full aux_clm45 test suite with cime5). In addition, I just did 5 separate runs of the full aux_clm test suite out of a CLM version with cime4, and there was not a single build failure, giving me more confidence that this is _not_ just due to recent machine flakiness. So this makes me wonder if the parallel builds are occasionally stomping on each other somehow.

I have been running the aux_clm test suite as 5 separate jobs (with 5 different test ids): one for clm40-intel, one for clm40-pgi, one for clm45-intel, one for clm45-pgi, and one for clm45-gnu.

This is with a slightly old version of cime - something basically equivalent to d321e4b . So, apologies if this is something that has been fixed since then - though I don't remember seeing any relevant PRs.

Here are the relevant build failures:

**From my most recent run of the test suite; the sharedlib directory was in /glade/scratch/sacks**

For `ERP_D_Ld5.f10_f10.ICLM45.yellowstone_pgi.clm-reduceOutput.GC.1018-1324.45.p`: Error in building the executable:

```
/glade/scratch/sacks/tests_1018-1324/sharedlibroot.1018-1324.45.p/pgi/mpich2/debug/threads/lib/libmpeu.a: file not recognized: File truncated
```
(even though the modification time on that file is about an hour before the time stamp on the relevant build log file).

**From a run a few weeks ago; the sharedlib directory was under /glade/p/work/sacks**

For `ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default.GC.0929-2129.40.p`: Error in building the lnd library:

```
PGF90-F-0004-Corrupt or Old Module file pio_types.mod (/glade/p/work/sacks/cesm_code/clm_limitirrig/components/clm/src_clm40/main/accumulMod.F90: 577)
```

For `ERP_D_Ld5.f19_g16.ICLM45BGC.yellowstone_intel.clm-default.GC.0929-2129.45.i`: Error in building the executable:

```
ld: BFD version 2.20.51.0.2-5.36.el6 20100205 internal error, aborting at merge.c line 872 in _bfd_merged_section_offset

ld: Please report this bug.

gmake: *** [/glade/scratch/sacks/ERP_D_Ld5.f19_g16.ICLM45BGC.yellowstone_intel.clm-default.GC.0929-2129.45.i/bld/cesm.exe] Error 1
```

For `ERI_D_Ld9.f10_f10.ICLM45BGC.yellowstone_pgi.clm-reduceOutput.GC.0929-2129.45.p`: Error in building the executable:

```
PGF90-F-0004-Corrupt or Old Module file m_mpif.mod (/glade/p/work/sacks/cesm_code/clm_limitirrig/cime/driver_cpl/driver/vertical_gradient_calculator_2nd_order.F90: 208)
PGF90/x86-64 Linux 15.10-0: compilation aborted
```

**And from another run a few weeks ago; the sharedlib directory was under /glade/p/work/sacks**

For `ERP_D_Ld5.f19_g16.ICLM45BGC.yellowstone_intel.clm-default.GC.0929-0852.45.i`: Error in building the executable:

```
/glade/p/work/sacks/cesm_code/clm_limitirrig/cime/scripts/tests_0929-0852/sharedlibroot.0929-0852.45.i/intel/mpich2/debug/threads/lib/libmct.a: file not recognized: File truncated
```

For `ERP_Ld5.f10_f10.I1850CLM45BGC.yellowstone_pgi.clm-default.GC.0929-0852.45.p`: Simultaneous failures in building stub glc, esp and ice: like this:

```
PGF90-F-0004-Corrupt or Old Module file pio_types.mod (/glade/p/work/sacks/cesm_code/clm_limitirrig/cime/components/stub_comps/sglc/cpl/glc_comp_mct.F90: 11)
```

### LINKED ISSUE #695: mpi-serial builds rebuild a lot even when reusing a sharedlib build
When you run a test suite, tests with the same build configuration reuse the sharedlib build space for some libraries. The intent is that these libraries shouldn't need to be rebuilt each time, but instead the later tests can reuse the builds from the earlier tests.

But it seems this isn't working as intended for mpi-serial builds. (See #690 for the original reporting of this issue; but I'm copying the relevant information below.)

I looked at how long the SHAREDLIB_BUILD takes for tests that are sharing an already-built sharedlib build, for CLM (yellowstone_intel, with cime at d321e4b).

For mpich2 cases, an initial SHAREDLIB_BUILD took 185 sec (or maybe longer - it's possible that the case I looked at wasn't really the first), and a subsequent build (for a later test) took 37 sec. So the sharedlib build is saving substantial time. (Though I'm surprised that it still takes 37 sec, when it should have nothing to do. Some of this time must be spent in generating namelists, etc. - but 37 sec still seems like a lot.)

However: For mpi-serial cases, an initial SHAREDLIB_BUILD took 286 sec, and subsequent builds (for later tests) took 175 sec. So mpi-serial cases seem to be doing a lot of rebuilds.

I also looked at the modification times on various files in the sharedlib build area. For the mpi-serial builds, quite a few .mod and .a files are recreated in each build. It looks to me like the root cause of the problem is that there are some mpi header files (.h) copied into the include/ directory. It appears that these mpi-serial header files are re-copied with each build; I'm thinking they should be copied with a makefile rule, so that they are only re-copied if the version in the source tree has changed. It looks like this re-copying of the .h files has a cascading effect, presumably because make sees updated prerequisites which force other things to be rebuilt: lib/libmpi-serial.a is rebuilt, as are about half of the .mod files in csm_share, csm_share.a, and essentially all CLM files. I think that this constant re-creation of libraries causes problems when tests in the model_build phase are trying to use them: they're trying to read files that are changing out from under them.

**Brief summary of what I think the issue is: It appears that these mpi-serial header files are re-copied with each build; I'm thinking they should be copied with a makefile rule, so that they are only re-copied if the version in the source tree has changed.**

jedwards4b:
Update to install shared libs and includes in each case directory to avoid potential conflicts during the build.

Test suite: scripts_regression_tests with CIME_MODEL acme and cesm, various specific builds
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #690 
Fixes #695 

User interface changes?: 

Code review:


utils/python/CIME/build.py
```diff
@@ -411,36 +411,35 @@ def build_checks(case, build_threaded, comp_interface, use_esmf_lib, debug, comp
     case.set_value("BUILD_COMPLETE", False)
 
     case.flush()
+    return sharedpath
```
billsacks: It seems unintuitive to have build_checks return the sharedpath. At the very least, there should be some function-level documentation of this return value.

cime_config/cesm/machines/Makefile
```diff
@@ -291,14 +288,7 @@ endif
...
 #===============================================================================
-INCLDIR += -I$(SHAREDPATH)/include -I$(SHAREDPATH)/$(COMP_INTERFACE)/$(ESMFDIR)/$(NINST_VALUE)/csm_share
-
-#
-#  Tell gmake to look in the csm_share directory for shr_*.mod files
-#
-
-vpath  shr%.mod $(SHAREDPATH)/$(COMP_INTERFACE)/$(ESMFDIR)/$(NINST_VALUE)/csm_share
-
+INCLDIR += -I$(INSTALL_SHAREDPATH)/include -I$(INSTALL_SHAREDPATH)/$(COMP_INTERFACE)/$(ESMFDIR)/$(NINST_VALUE)/include
```
billsacks: If I'm understanding this change (and other similar changes) correctly, it means that, even for the shared library builds, the include directory is pointing to the "install" location - i.e., some case-specific location. That implies that (e.g.) the sharedlib csm_share build is done against a case-specific installation of the mct build. I'm not sure if this will cause problems in practice, although it feels weird.

I had imagined that anything built in the sharedlib area would be built against module and .h files in the sharedlib area, in order to avoid having things from sharedlib point into specific cases. Then it would just be the things built in the case-specific location (i.e., non-shared-lib) that would point to this new INSTALL_SHAREDPATH.

I don't know how important this is, though, beyond a general sense of aesthetics. Is it important to avoid problems if the case-specific install directory is blown away???

cime_config/buildlib.csm_share
```diff
@@ -6,12 +6,12 @@ use File::Path;
 # Purpose: Build csm_share library
 #===============================================================================
 
-if ($#ARGV == -1) {
-    die " ERROR buildlib_csm_share: must specify a caseroot input argument";
+if ($#ARGV < 2) {
```
billsacks: This check and the associated 'die' message look wrong: it looks like 3 arguments are actually expected

utils/python/CIME/build.py
```diff
@@ -452,10 +451,9 @@ def build_libraries(case, exeroot, caseroot, cimeroot, libroot, mpilib, lid, mac
     if comp_lnd == "clm" and not "clm4_0" in clm_config_opts:
         logging.info("         - Building clm4_5/clm5_0 Library ")
         esmfdir = "esmf" if case.get_value("USE_ESMF_LIB") else "noesmf"
-        sharedpath = os.environ["SHAREDPATH"]
-        bldroot = os.path.join(sharedpath, case.get_value("COMP_INTERFACE"), esmfdir, "clm","obj" )
-        libroot = os.path.join(sharedpath, case.get_value("COMP_INTERFACE"), esmfdir, "lib")
-        incroot = os.path.join(sharedpath, case.get_value("COMP_INTERFACE"), esmfdir, "include")
+        bldroot = os.path.join(sharedlibroot, sharedpath, case.get_value("COMP_INTERFACE"), esmfdir, "clm","obj" )
```
billsacks: I don't see code to copy the CLM build into the install location; should there be? Maybe this needs to go in CLM itself? (If so, I'm happy to help make the necessary CLM tag.)

jedwards4b: This is handled in the Makefile

cime_config/cesm/machines/Makefile
jedwards4b: There is a chain of dependancies.  The install location is dependent on the files in the build location.  Blowing away a case specific install directory only affects the given case, no other case using that shared build would be affected.

cime_config/buildlib.csm_share
jedwards4b: perl $#ARGV is the index of the last array element

billsacks: Ah, thanks. I guess it's just the 'die' message that's wrong.

cime_config/cesm/machines/Makefile
billsacks: In talking with Jim E, it sounds like things in the sharedlib area do get built against things in the case-specific INSTALL_SHAREDPATH, but this shouldn't cause a problem. (In particular, I wondered if the include path has any effect at runtime, and it sounds like the answer is No.) So I guess I'm okay with this.

utils/python/CIME/build.py
billsacks: It doesn't look like it's handled in the Makefile, but is apparently handled somewhere. I'll test it.

utils/python/CIME/build.py
jedwards4b: I've added a comment, it should be pointed out that this is an improvement over the prior behavior which was to set and read in an environment variable.

billsacks: Thanks. Yes, very good point.

jedwards4b:
rebased to resolve conflicts with master

billsacks:
I ran a bunch of testing on this. I'm getting failures in all mpi-serial builds in the latest version due to #700 so I'm just looking at the non-mpi-serial builds for now.

The good news is that this appears to fix the build errors.

However, build times are significantly longer than before - it seems that we're no longer getting the full benefit of the sharedlib build area. Poking around, here's what I see:

1. First, a difference from before... as far as I can tell, this isn't a problem, but it leads to an asymmetry: previously, the csm_share build was put in a path like:

 `intel/mpich2/debug/nothreads/MCT/noesmf/c1a1l1i1o1r1g1w1e1/csm_share`

 Now, the csm_share build is put in a path like:

 `intel/mpich2/debug/nothreads/csm_share/MCT/noesmf/c1a1l1i1o1r1g1w1e1/csm_share`

 Note the extra csm_share directory between nothreads and MCT. This differs from other libraries, where the library name (csm_share in this case) does not appear in the path until after all of the different build options. For example, the CLM library was - and still is - in a path like:

 `intel/mpich2/debug/nothreads/MCT/noesmf/clm`

 So, previously, the clm build sat alongside the csm_share build (because they shared options), but now that is no longer the case.

 Furthermore, note that there is _not_ that extra csm_share in the path in the "install" location, further adding to the asymmetry.

 Again, this doesn't seem like a huge problem, but I wanted to point it out in case it was accidental, because it seems best to keep things consistent, if possible.

2. The real problem, I think: Most (but not quite all) of the CLM library seems to be getting rebuilt for every case. My best guess is that this is because the time stamps of the PIO library and module files are updated when they are copied to the install location. I'm thinking that fixing this might be as easy as changing:

 ```python
 shutil.copy(item, "%s/include"%installpath)
 ```

 to

 ```python
 shutil.copy2(item, "%s/include"%installpath)
 ```

 on line 87 of buildlib.pio.


Once these are fixed - or at least (2) - and once #700 is fixed and merged into this branch, then I'll redo my testing.

jedwards4b:
Fixes #700 
Addresses @billsacks comments, thank you for testing.

billsacks:
@jedwards4b - with your latest version, all builds are successful, but a bunch of stuff is still getting rebuilt which should be reusing existing sharedlib builds.

**For non-mpi-serial**

For non-mpi-serial: Much of the CLM build is getting redone, starting with landunit_varcon. My best guess (not investigated carefully) is that it's this line that's triggering the rebuild:

```
#include "shr_assert.h"
```

As far as I can tell, that header file itself is not updated (unless I'm missing something): it seems like it's getting the header file from `/glade/scratch/sacks/cesm_code/clm4_5_12_r198_cime_je/cime/scripts/Tools/../../share/csm_share/include/shr_assert.h`. But is it possible that things are getting rebuilt simply because there are _directories_ in the include path with updated time stamps???

Some other things that I see have updated time stamps in the CLM build are Filepath, Deppath and Depends. Those seem to be the first files updated. Could that be the source of the problem?

In addition, all of the genf90 stuff seems to be getting redone unnecessarily - although that happens after landunit_varcon is rebuilt, so that seems to be independent of the landunit_varcon rebuild.

**For mpi-serial**

For the mpi-serial build: Much of the csm_share build is still getting redone – starting with esmf_stubs. Then the whole CLM build is redone.

I think the problem here is this include directory: `/glade/scratch/sacks/SMS_Ly3_Mmpi-serial.1x1_numaIA.ICRUCLM50BGCCROP.yellowstone_intel.clm-clm50dynroots.1022-0926e.45.i/bld/lib/include`:

```
-rw-r--r-- 1 sacks ncar   1529 Oct 22 13:48 list.h
-rw-r--r-- 1 sacks ncar    352 Oct 22 13:48 listP.h
-rw-r--r-- 1 sacks ncar    296 Oct 22 13:48 listops.h
-rw-r--r-- 1 sacks ncar   2867 Oct 22 13:48 type.h
-rw-r--r-- 1 sacks ncar   6890 Oct 22 13:48 mpif.h
-rw-r--r-- 1 sacks ncar   2924 Oct 22 13:48 mpiP.h
-rw-r--r-- 1 sacks ncar  15384 Oct 22 13:48 mpi.h
-rw-r--r-- 1 sacks ncar  72844 Oct 22 13:52 esp_comp_mct.mod
-rw-r--r-- 1 sacks ncar  93647 Oct 22 13:52 ocn_comp_mct.mod
-rw-r--r-- 1 sacks ncar  93647 Oct 22 13:52 glc_comp_mct.mod
-rw-r--r-- 1 sacks ncar  93647 Oct 22 13:52 wav_comp_mct.mod
-rw-r--r-- 1 sacks ncar  93647 Oct 22 13:52 ice_comp_mct.mod
-rw-r--r-- 1 sacks ncar 189913 Oct 22 13:52 datm_comp_mod.mod
-rw-r--r-- 1 sacks ncar  99791 Oct 22 13:52 atm_comp_mct.mod
-rw-r--r-- 1 sacks ncar 215334 Oct 22 13:52 rof_comp_mct.mod
-rw-r--r-- 1 sacks ncar    299 Oct 22 13:52 rof_comp_esmf.mod
```

The updated time stamps on the .h files seem to be the problem.

billsacks:
@jedwards4b - would you be able to take over testing of continued iterations of this branch? (I'm happy to re-do some final testing once it looks to you like the problems are resolved.) The procedure I'm using to see what's getting rebuilt is:

1. Check out latest CLM tag: `https://svn-ccsm-models.cgd.ucar.edu/clm2/trunk_tags/clm4_5_12_r198`

2. Switch cime to your branch

3. Run the following commands:

 ```
export testid=`date +%m%d-%H%M`
export testroot=/glade/scratch/$USER/tests_${testid}
mkdir -p $testroot
ln -s $testroot
export category=aux_clm45
export compiler=intel
execca ./create_test --xml-category $category --xml-machine yellowstone --xml-compiler ${compiler} --test-id ${testid}.${category:7:2}.${compiler:0:1} --test-root ${testroot} --walltime 0:01 --no-run
 ```

4. Check the SHAREDLIB_BUILD times for the various builds: these should go drastically down for subsequent builds of a given configuration.

5. Check time stamps on files throughout the shared lib area and the new "install" area.

jedwards4b:
The bld/include directory should not have mpi-serial files in it, I'm not sure where that is happening - do you know?

billsacks:
No, I haven't been able to figure out anything beyond what I posted.

jedwards4b:
I've pushed a fix for the mpi-serial issue.   I'd like you to accept this PR and we will address improvements in subsequent commits.

billsacks:
@jedwards4b - you said that the necessary CLM stuff is getting copied into the install location. I see libclm.a there, but don't see any .mod files there.

An example is this: `/glade/scratch/sacks/SMS_D_Ld5.f45_f45.ICLM45ED.yellowstone_intel.clm-edTest.1022-0926e.45.i/bld`. I see:

```
[yslogin6:/glade/scratch/sacks/SMS_D_Ld5.f45_f45.ICLM45ED.yellowstone_intel.clm-edTest.1022-0926e.45.i/bld]$ find . -name '*comp_mct*.mod'
./atm/obj/atm_comp_mct.mod
./ocn/obj/ocn_comp_mct.mod
./esp/obj/esp_comp_mct.mod
./glc/obj/glc_comp_mct.mod
./rof/obj/rof_comp_mct.mod
./lib/include/wav_comp_mct.mod
./lib/include/glc_comp_mct.mod
./lib/include/ocn_comp_mct.mod
./lib/include/rof_comp_mct.mod
./lib/include/ice_comp_mct.mod
./lib/include/atm_comp_mct.mod
./lib/include/esp_comp_mct.mod
./wav/obj/wav_comp_mct.mod
./ice/obj/ice_comp_mct.mod
```

But I don't see lnd_comp_mct.mod.

One file that may be serving this purpose is: `/glade/scratch/sacks/SMS_D_Ld5.f45_f45.ICLM45ED.yellowstone_intel.clm-edTest.1022-0926e.45.i/bld/intel/mpich2/debug/nothreads/MCT/noesmf/include`

This is a binary file (not a directory). Is it correct to have this binary include file? (I've never seen anything like that before.)

The reasons I raise this question are:

1. If the necessary CLM stuff isn't getting installed into the "install" location, then there's still at least a theoretical possibility that builds will interfere with each other (if a file in the sharedlib area is getting rewritten by one test while another test is building the executable).

2. Just in case this is somehow related to the CLM rebuild problem I've been discussing.

jedwards4b:
I have fixed the issue with the include directory, there are still a few problems with files being rebuilt, but if you have no objection I would like to merge this pr and fix the remaining (and present prior to this pr) issues in future commits.

jedwards4b:
Merged with okay from @billsacks

PR #701, opened 2016-10-23
Changes to handling of environment_variables

jgfouca:
Incorporates jedwards' fix on another PR but in such a way
as to not break melvin. Specifically, we need to be able to use
environment variables set by the modules in environment_variables. Also,
if we're going to call get_resolved_value, then we cannot support bash
in this field

Test baseline: 
Test namelist changes: 
Test status: bit for bit

User interface changes?: Minor. environment_variables setting in config_machines.xml works a bit differently now

Code review: jedwards

### DIFF
--- a/utils/python/CIME/XML/env_mach_specific.py
+++ b/utils/python/CIME/XML/env_mach_specific.py
@@ -51,7 +51,6 @@ class EnvMachSpecific(EnvBase):
                 for node in nodes:
                     self.add_child(node)
 
-
     def get_full_records(self, item, attribute=None, resolved=True, subgroup=None):
         """Returns the value as a string of the first xml element with item as attribute value.
         <element_name attribute='attribute_value>value</element_name>"""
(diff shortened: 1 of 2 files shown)


utils/python/CIME/XML/env_mach_specific.py
```diff
@@ -189,7 +194,16 @@ def _compute_actions(self, nodes, child_tag, compiler, debug, mpilib):
...
                     if (self._match_attribs(child.attrib, compiler, debug, mpilib)):
-                        result.append( (child.get("name"), child.text) )
+                        val = child.text
+                        if val is not None:
+                            # We allow a couple special substitutions for these fields
+                            for repl_this, repl_with in [("$COMPILER", compiler), ("$MPILIB", mpilib)]:
+                                val = val.replace(repl_this, repl_with)
+
+                            val = self.get_resolved_value(val)
+                            expect("$" not in val, "Not safe to leave unresolved items in env var value: '%s'" % val)
```
jedwards4b: There is at least one case of a place where you need to leave environment variables unresolved - on mira there are variables which must only be resolved after the job has booted on the compute node, I believe that that system would fail with this expect.

jgfouca: Wouldn't your changes to the preview_namelists PR have the same problem?

jgfouca: I just looked at the mira environment variables settings:
```
    <module_system type="soft">
      <init_path lang="csh">/etc/profile.d/00softenv.csh</init_path>
      <cmd_path lang="csh">soft</cmd_path>
      <init_path lang="sh">/etc/profile.d/00softenv.sh</init_path>
      <cmd_path lang="sh">soft</cmd_path>
      <modules>
        <command name="add">+mpiwrapper-xl</command>
        <command name="add">@ibm-compilers-2015-02</command>
        <command name="add">+cmake</command>
        <command name="add">+python</command>
      </modules>
    </module_system>
    <environment_variables>
      <env name="MPI_TYPE_MAX">10000</env>
      <env name="OMP_DYNAMIC">FALSE</env>
      <env name="OMP_STACKSIZE">64M</env>
      <env name="HDF5">/soft/libraries/hdf5/1.8.14/cnk-xl/current</env>
    </environment_variables>
```
... and it doesn't look like it will be a problem.

jedwards4b: The issue is in the arguments to mpirun, maybe they aren't affected by this change?  
The variables COBALT_PARTNAME and LOCARGS need to be unresolved until the compute node is booted.

jgfouca: This should only impact the fields in env_mach_specific (module_system and environment_variables).

jedwards4b:
I'm okay with it, please resolve conflict and merge to master.

jgfouca:
OK, thanks

PR #707, opened 2016-10-25
fix indentation error

jedwards4b:
Fix Indentation error in env_mach_specific causing module to load incorrectly

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Code review:


jgfouca:
Nevermind, you're right. None would be the argument, not the action.

PR #708, opened 2016-10-25
Pylint cleanup tools

jedwards4b:
Add scripts/Tools to the list of directories to check with pylint, clean up code to pass pylint tests

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Code review:


scripts/Tools/code_checker
```diff
@@ -57,10 +57,16 @@ def run_pylint(on_file):
 ###############################################################################
     pylint = find_executable("pylint")
 
-    cmd = "%s --disable=I,C,R,logging-not-lazy,wildcard-import,unused-wildcard-import,fixme,broad-except,bare-except,eval-used,exec-used,global-statement %s" % (pylint, on_file)
+    cmd_options = " --disable=I,C,R,logging-not-lazy,wildcard-import,unused-wildcard-import,fixme,broad-except,bare-except,eval-used,exec-used,global-statement"
+
+    if "scripts/Tools" in on_file:
```
jgfouca: Why remove relative import warning? It's actually useful because it helps reduce the cases where the tools can't find imports because you ran them from a different location.

cime_config/buildlib.pio
```diff
@@ -1,6 +1,6 @@
 #!/usr/bin/env python
 import shutil, glob, re
-from Tools.standard_script_setup import *
+from standard_script_setup import *
```
jgfouca: Why change to relative import?

scripts/Tools/code_checker
```diff
@@ -57,10 +57,16 @@ def run_pylint(on_file):
...
+        cmd_options +=",relative-import"
+    cmd = "%s %s %s"%(pylint, cmd_options, on_file)
+    print "HERE %s"%cmd
```
jgfouca: Don't think we want this print command.

scripts/Tools/code_checker
```diff
@@ -57,10 +57,16 @@ def run_pylint(on_file):
...
     stat, out, err = run_cmd(cmd, verbose=False)
     if stat != 0:
         logger.info("File %s has pylint problems, please fix\n    Use command: %s" % (on_file, cmd))
+        logger.info(out + "\n" + err)
```
jgfouca: I've been using --debug when I want to see the full output. In any case, we definitely don't want the two lines below if we keep this.

scripts/Tools/code_checker
```diff
@@ -85,12 +91,6 @@ def check_code(dir_to_check, num_procs, files):
 
     Returns True if all files had no problems
     """
-    # Pylint won't work right if the imports within the checked files fails
```
jgfouca: How were you able to get rid of this?

scripts/Tools/jenkins_generic_job
```diff
@@ -14,8 +14,6 @@ from CIME.utils import expect
 from CIME.XML.machines import Machines
 from jenkins_generic_job import jenkins_generic_job
 
-import argparse, sys, os, shutil, glob, doctest, signal, time
```
jgfouca: This doesn't look right. I see some of these being used below

utils/python/CIME/XML/generic_xml.py
```diff
@@ -227,7 +227,9 @@ def get_resolved_value(self, raw_value):
                 srcroot = os.path.join(get_cime_root(),"..")
                 item_data = item_data.replace(m.group(), srcroot)
             elif var in os.environ:
-                logging.warn("Resolved from env: " + var)
+                # this is a list of suppressed warnings (things normally expected to be resolved in env)
+                if var not in ("USER"):
```
jgfouca: This is definitely wrong. You want 
```
("USER",)
```
... if you want that to be interpreted as a tuple.

scripts/Tools/code_checker
jedwards4b: Because the import directory is the directory this file is in.   I tried several different things and this is the one that works.

scripts/Tools/code_checker
jedwards4b: This is much lower impact now that we are doing one file at a time, removing lines below.

scripts/Tools/code_checker
jedwards4b: Because I provide the python path in scripts_regression_tests and the path should be to 
scripts/Tools and utils/python (and not to .) to reproduce the behavior we expect outside of pylint.

cime_config/buildlib.pio
jedwards4b: Because the directory that is in pythonpath is Tools and not scripts.

scripts/Tools/jenkins_generic_job
jedwards4b: they are provided by standard_scripts_setup.py

utils/python/CIME/XML/generic_xml.py
jedwards4b: thanks

scripts/Tools/code_checker
jedwards4b: removed

cime_config/buildlib.pio
jgfouca: Ah, OK

scripts/Tools/code_checker
jgfouca: Which file was causing this problem? You could just disable that warning there

jgfouca: oh, looks like you only turn it off for that directory, that's fine

PR #709, opened 2016-10-25
Change env settings from use of bash to use of environment

### LINKED ISSUE #703: PR #689 has caused case.setup script to fail on ANL workstations
In cime/scripts directory:
./create_newcase -mach anlworkstation -compiler gnu -case test_X_compute001 -compset X -res f19_g16
cd test_X_compute001
./case.setup

ERROR: Could not resolve value for val=$(dirname $(dirname $(which ncdump)))

It seems that this error is introduced by PR #689 (git checkout f6e3e14 has no such error)

jgfouca:
Test suite: None
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #703 

User interface changes?: No

Code review: @jayeshkrishna

### DIFF
--- a/cime_config/acme/machines/config_machines.xml
+++ b/cime_config/acme/machines/config_machines.xml
@@ -490,8 +490,8 @@
       </modules>
     </module_system>
     <environment_variables>
-      <env name="NETCDFROOT">$(dirname $(dirname $(which ncdump)))</env>
-      <env name="PNETCDFROOT" mpilib="!mpi-serial">$(dirname $(dirname $(which pnetcdf_version)))</env>
+      <env name="NETCDFROOT">$ENV{NETCDFROOT}</env>
+      <env name="PNETCDFROOT" mpilib="!mpi-serial">$ENV{PNETCDFROOT}</env>
     </environment_variables>
 </machine>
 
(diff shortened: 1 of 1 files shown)


jgfouca:
@jayeshkrishna in order for this to work, there must be some way to get the PNETCDFROOT from the environment. Would it be possible to get the admins to add this to the soft definition for pnetcdf? If that's really impossible, we can just hardcode it I guess.

jayeshkrishna:
Why doesn't the old way (which ncdump) work any more?

jgfouca:
Because @jedwards4b wanted better error checking on that field which made having bash in there impossible. Also, bash is not generally supported, so it was kinda odd to just support in this one field.

jayeshkrishna:
Ok, we will have to hardcode the paths for now and I will start a conversation with systems to add NETCDFROOT and PNETCDFROOT when loading the appropriate soft keys.
Also, should the names be NETCDF_DIR or NETCDF_PATH (instead of NETCDFROOT)?

jedwards4b:
NETCDF_PATH and PNETCDF_PATH are what is expected by make, using these will allow you to avoid having to translate them in the scripts.

jgfouca:
Ok, the hardcoded pnetcdf path is added.

PR #713, opened 2016-10-26
fix issue with recursive include of user mods

### LINKED ISSUE #712: Recursion in testmods_dirs (include_user_mods file)
In POP testing, we are having an issue where namelist changes (via `user_nl_pop`) are not being applied, and I'm not sure how to trace the problem. I'm also not sure of the best way to provide all the info needed in a readable manner, but we have (among others) the following directories in `components/pop/cime_config/testdefs/testmods_dirs/pop`:

```
cice_ecosys_dust_flux_driver/
cice/
ecosys_dust_flux_driver/
ecosys/
default/
```

And all but `default/` contain `include_user_mods` files:

```
$ cat cice_ecosys_dust_flux_driver/include_user_mods 
../cice
../ecosys_dust_flux_driver

$ cat cice/include_user_mods 
../default

$ cat ecosys_dust_flux_driver/include_user_mods 
../ecosys

$ cat ecosys/include_user_mods 
../default
```

Some of them also include `user_nl_pop` or `user_nl_cice` files:

```
$ find . -name user_nl_*
./cice/user_nl_cice
./ecosys_dust_flux_driver/user_nl_pop
./ecosys/user_nl_pop
./default/user_nl_pop
```

When we run a `pop-cice_ecosys_dust_flux_driver` test, we get the expected `user_nl_cice` but `user_nl_pop` matches the version in `default/` rather than containing the additional lines from `ecosys_dust_flux_driver/` and `ecosys/`

My current theory is that having two separate chains leading back to `default` is messing something up, because running just a `pop-ecosys_dust_flux_driver` test brings in all the expected `user_nl_pop` changes (but omits the `user_nl_cice` changes we would like to have).

---

A similar issue occurs with the `pop-ciso_abio_dic_dic14` test, which has the following chain of includes

```
$ cat ciso_abio_dic_dic14/include_user_mods 
../abio_dic_dic14
../ciso

$ cat abio_dic_dic14/include_user_mods 
../default

$ cat ciso/include_user_mods 
../ecosys

$ cat ecosys/include_user_mods 
../default
```

In this second case, only `default` and `ecosys` have `user_nl_pop` files, but the file from `ecosys/` is ignored. All of the `shell_commands` files (in `abio_dic_dic14/`, `ciso/`, and `default/`) are applied... it seems like a `user_nl` issue only.

jedwards4b:
Indentation errors caused incorrect inclusion of test mods user_nl_ files when multiple files should be appended 

Test suite: scripts_regression_tests, SMS.T62_g16.C.yellowstone_intel.pop-cice_ecosys_dust_flux_driver
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #712 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/user_mod_support.py
+++ b/utils/python/CIME/user_mod_support.py
@@ -39,10 +39,10 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
                     if os.path.isfile(case_user_nl_inst):
                         with open(case_user_nl_inst, "r") as fd:
                             old_contents = fd.read()
-                            if old_contents.find(contents) == -1:
-                                contents = contents + old_contents
-                                with open(case_user_nl_inst, "w") as fd:
-                                    fd.write(contents)
+                        if old_contents.find(contents) == -1:
+                            contents = contents + old_contents
+                            with open(case_user_nl_inst, "w") as fd:
+                                fd.write(contents)
             else:
                 contents = newcontents
                 logger.info("Pre-pending file %s from %s"%(case_user_nl, include_dir))
--- a/utils/python/CIME/user_mod_support.py
+++ b/utils/python/CIME/user_mod_support.py
@@ -35,24 +35,9 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
                 for comp_inst in xrange(1, ninst[comp]+1):
                     contents = newcontents
                     case_user_nl_inst = case_user_nl + "_%4.4d"%comp_inst
-                    logger.info("Pre-pending file %s"%(case_user_nl_inst))
-                    if os.path.isfile(case_user_nl_inst):
-                        with open(case_user_nl_inst, "r") as fd:
-                            old_contents = fd.read()
-                        if old_contents.find(contents) == -1:
-                            contents = contents + old_contents
-                            with open(case_user_nl_inst, "w") as fd:
-                                fd.write(contents)
+                    update_user_nl_file(case_user_nl_inst, contents)
             else:
-                contents = newcontents
-                logger.info("Pre-pending file %s from %s"%(case_user_nl, include_dir))
-                if os.path.isfile(case_user_nl):
-                    with open(case_user_nl, "r") as fd:
-                        old_contents = fd.read()
-                    if old_contents.find(contents) == -1:
-                        contents = contents + old_contents
-                        with open(case_user_nl, "w") as fd:
-                            fd.write(contents)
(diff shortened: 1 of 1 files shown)


utils/python/CIME/user_mod_support.py
```diff
@@ -35,24 +35,24 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
...
-                logger.info("Pre-pending file %s"%case_user_nl)
+                logger.info("Pre-pending file %s from %s"%(case_user_nl, include_dir))
                 if os.path.isfile(case_user_nl):
                     with open(case_user_nl, "r") as fd:
                         old_contents = fd.read()
-                        if old_contents.find(contents) == -1:
-                            contents = contents + old_contents
-                with open(case_user_nl, "w") as fd:
-                    fd.write(contents)
+                    if old_contents.find(contents) == -1:
```
billsacks: You changed the indentation of this conditional here, so that it is no longer nested under the 'with open...' - but you did not change it in the above block that applies to multi-instance. It looks to me like either one would work, though the version here (with this code _not_ nested under the 'with open...') seems better. Unless this difference was intentional, it would probably make sense to change the above block, too, so that they're consistent.

billsacks:
Looks good now - and I like the restructuring - thanks!

mnlevy1981:
This looks good to me. One difference I noticed between CIME 4 and CIME 5 is the order that `user_nl` files are updated... here's the CIME 4 `user_nl_pop` for one test:

```
& tavg_nml
    ltavg_ignore_extra_streams = .true.
    n_tavg_streams = 1
    tavg_freq_opt = 'nday'
    tavg_file_freq_opt = 'nday'
/
& tavg_nml
    ltavg_ignore_extra_streams = .true.
    n_tavg_streams = 1
    tavg_freq_opt = 'nday'
    tavg_file_freq_opt = 'nday'
/
lecosys_debug = .true.
dust_flux_source = 'driver'
iron_flux_source = 'driver-derived'
```

and the same test with this pull request

```
lecosys_debug = .true.
dust_flux_source = 'driver'
iron_flux_source = 'driver-derived'
& tavg_nml
    ltavg_ignore_extra_streams = .true.
    n_tavg_streams = 1
    tavg_freq_opt = 'nday'
    tavg_file_freq_opt = 'nday'
/
```

But I think the new order is preferable, because now the `default` is at the bottom of the file... so if I have a test that changes something in `default` then the change will occur in the top of the file and I believe that's the value build-namelist will use.

PR #721, opened 2016-10-26
change lt_archive queue to caldera, add sleep in scripts_regression_t…

### LINKED ISSUE #720: case.lt_archive is broken on yellowstone
The hpss queue that is supposed to be used to submit hpss job runs on a virtual machine that does not have modules installed and so case.lt_archive gives an error because of too old a python version.  Workaround is to submit to the caldera queue

jedwards4b:
The hpss queue on yellowstone does not support modules and has too old a python version installed.  
Test suite:  case.submit --job case.lt_archive 
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #720 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/tests/scripts_regression_tests.py
+++ b/utils/python/tests/scripts_regression_tests.py
@@ -463,6 +463,7 @@ class TestCreateTestCommon(unittest.TestCase):
         else:
             for file_to_clean in files_to_clean:
                 if (os.path.isdir(file_to_clean)):
+                    time.sleep(5) # Kinda hacky
                     shutil.rmtree(file_to_clean)
                 else:
                     os.remove(file_to_clean)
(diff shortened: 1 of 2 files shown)


utils/python/tests/scripts_regression_tests.py
```diff
@@ -463,6 +463,7 @@ def tearDown(self):
         else:
             for file_to_clean in files_to_clean:
                 if (os.path.isdir(file_to_clean)):
+                    time.sleep(5) # Kinda hacky
```
jgfouca: I don't think I can accept this, it will slow down testing dramatically. What are we trying to solve here? Can we limit this to certain machines? Can you put the sleep outside of the loop?

PR #725, opened 2016-10-26
cpl restart read should only happen on cpl tasks

### LINKED ISSUE #724: driver_cpl: With some PE layouts, on a restart, fractions_ax is accessed without being allocated
Testing the CLM trunk with cime5.2.0-alpha.5 (but I don't see any driver_cpl updates to cime since that tag), I'm getting failures in all ERP debug tests with hobart-nag: e.g., from `ERP_D_Ld5_P24x1.f10_f10.I1850CLM45BGC.hobart_nag.clm-ciso.GC.1025-0639.45.n`

```
Runtime Error: /home/sacks/cesm_code/cime_update_cesm2_0_beta03_ish/cime/driver_cpl/driver/cesm_comp_mod.F90, line 1955: Reference to undefined POINTER FRACTIONS_AX
```

I think there might actually be two problems contributing to this error:

(1) Due to #723 ESP is running on 24 tasks while the rest of the system is running on 12 tasks

(2) This issue at hand: It appears that fractions_ax is only allocated on the coupler PEs, but is accessed on all PEs when doing a restart: In cesm_comp_mod.F90:

Line 1744:

``` fortran
   if (iamin_CPLID) then
      call t_startf ('CPL:init_fracs')
      call t_adj_detailf(+2)

      allocate(fractions_ax(num_inst_frc))
      allocate(fractions_lx(num_inst_frc))
      allocate(fractions_ox(num_inst_frc))
      allocate(fractions_ix(num_inst_frc))
      allocate(fractions_gx(num_inst_frc))
      allocate(fractions_rx(num_inst_frc))
      allocate(fractions_wx(num_inst_frc))
```

Line 1954:

``` fortran
   if (read_restart) then
      call seq_rest_read(rest_file, infodata, &
           atm, lnd, ice, ocn, rof, glc, wav, esp, &
           fractions_ax, fractions_lx, fractions_ix, fractions_ox, &
           fractions_rx, fractions_gx, fractions_wx)
   endif
```

It looks to me like this has been a long-standing issue: the relevant code looks the same back to at least March, 2015 (pre-cime1). It seems like this would be a problem whenever you do a restart with a PE layout where the coupler is on a subset of the PEs. I'm not sure why this hasn't triggered an error before - maybe it's only picked up by nag in debug mode, and we don't typically run with these PE layouts there?

cc @fischer-ncar @jedwards4b @apcraig

jedwards4b:
Test suite: ERP_D_Ld5_P24x1.f10_f10.I1850CLM45BGC.hobart_nag.clm-ciso
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #724

User interface changes?: 

Code review:

### DIFF
--- a/driver_cpl/driver/cesm_comp_mod.F90
+++ b/driver_cpl/driver/cesm_comp_mod.F90
@@ -1951,7 +1951,7 @@ subroutine cesm_init()
    call t_adj_detailf(+2)
 
    call seq_diag_zero_mct(mode='all')
-   if (read_restart) then
+   if (read_restart .and. iamin_CPLID) then
       call seq_rest_read(rest_file, infodata, &
            atm, lnd, ice, ocn, rof, glc, wav, esp, &
            fractions_ax, fractions_lx, fractions_ix, fractions_ox, &


billsacks:
Looks good - thanks! I'm thinking you should at least run `./create_test cime_developer` to make sure this works in some other configurations. Once you have done this, I think this is good to come to master.

Issue #692, opened 2016-10-20
Unexpected repetition of case.setup in tests


billsacks:
In doing some investigation for #690 - but also possibly relevant for #689 - I noticed that various pieces of case setup are done over and over again for tests. I picked a random SMS test (which was NOT rerun), and saw (in CaseStatus):

```
[yslogin2:/glade/p/work/sacks/cesm_code/clm_limitirrig/cime/scripts/tests_1018-1324/SMS_Lm1_D.f10_f10.ICLM45BGCCROP.yellowstone_intel.clm-snowlayers_3_monthly.GC.1018-1324.45.i]$ cat CaseStatus
2016-10-18 13:32:57: <command> ./xmlchange --force CLM_BLDNML_OPTS=-fire_emis --append  </command>
2016-10-18 13:33:00: case.setup starting
2016-10-18 13:33:04: <command> ./xmlchange --force CLM_BLDNML_OPTS=-fire_emis --append  </command>
2016-10-18 13:34:59: case.setup complete
2016-10-18 15:12:20: case.setup starting
2016-10-18 15:12:20: case.setup clean complete
2016-10-18 15:12:22: <command> ./xmlchange --force CLM_BLDNML_OPTS=-fire_emis --append  </command>
2016-10-18 15:12:46: case.setup complete
2016-10-18 15:12:46: case.testbuild starting
2016-10-18 15:13:04: case.testbuild complete
2016-10-18 15:13:54: case.setup starting
2016-10-18 15:13:54: case.setup clean complete
2016-10-18 15:13:55: <command> ./xmlchange --force CLM_BLDNML_OPTS=-fire_emis --append  </command>
2016-10-18 15:14:22: case.setup complete
2016-10-18 15:14:22: case.testbuild starting
2016-10-18 15:15:30: case.testbuild complete
2016-10-18 15:35:31: case.setup starting
2016-10-18 15:35:32: case.setup clean complete
2016-10-18 15:35:34: <command> ./xmlchange --force CLM_BLDNML_OPTS=-fire_emis --append  </command>
2016-10-18 15:35:58: case.setup complete
2016-10-18 15:36:19: Run started
2016-10-18 15:43:07: Run SUCCESSFUL
```

I was initially surprised by the two calls to case.testbuild... but maybe that's expected? (One for sharedlib, one for model??). But even setting that aside, I was surprised to see case.setup and the same xmlchange command repeated so many times for a simple SMS test.

[...]

PR #727, opened 2016-10-26
trim include directories list so that each is only added once

### LINKED ISSUE #719: With testmods, default directory sometimes occurs after others
@mnlevy1981 and I were just discussing this, connected with his recent comment in #713 

In some cases where you have an include_user_mods with two or more includes, it's possible for the various directories to be parsed in a way such that the 'default' directory (i.e., the directory at the bottom of the includes tree) appears in the middle of the user_nl file, rather than at the start. We both feel it should appear at the start, so that any other testmods will override settings set by default.

Here's an example; there are a number of examples like this in the POP test suite:

```
          ciso_abio
          /       \
        abio     ciso
        /          \
    default     default
```

Here, what we'd want is for the 'default' to appear at the top of user_nl_pop. Instead, it appears that the parsing happens like this:
1. ciso_abio: user_nl_pop is written
2. abio: user_nl_pop is pre-pended
3. default: user_nl_pop is pre-pended
4. ciso: user_nl_pop is pre-pended
5. (default is ignored)

So the resulting user_nl_pop would look like:

```
ciso
default
abio
ciso_abio
```

whereas we want

```
default
ciso
abio
ciso_abio
```

@mnlevy1981 's suggestion for how to accomplish this is to first build up a list of include directories. The key point is: whenever you encounter a duplicate, remove the existing instance, just keeping it at the end.

So in this case, the list would get built up like this:
1. abio
2. abio, default
3. abio, default, ciso
4. abio, ~~default,~~ ciso, default

Then you run through the list using code like we already have, doing the necessary pre-pends, resulting in:
1. ciso_abio: user_nl_pop is written
2. abio: user_nl_pop is pre-pended
3. ciso: user_nl_pop is pre-pended
4. default: user_nl_pop is pre-pended

resulting in the desired user_nl_pop.

Note also that the check for existing contents could then be removed.

jedwards4b:
Include directories are only listed once if repeated in chain.

Test suite: scripts_regression_tests, SMS.T62_g16.C.yellowstone_intel.pop-cice_ecosys_dust_flux_driver
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #719 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/case_setup.py
+++ b/utils/python/CIME/case_setup.py
@@ -8,8 +8,7 @@ from CIME.check_lockedfiles import check_lockedfiles
 from CIME.preview_namelists import create_dirs, create_namelists
 from CIME.XML.env_mach_pes  import EnvMachPes
 from CIME.XML.compilers     import Compilers
-from CIME.utils             import append_status, parse_test_name, get_cime_root
-from CIME.user_mod_support  import apply_user_mods
+from CIME.utils             import get_cime_root, append_status
 from CIME.test_status       import *
 
 import shutil, time, glob
@@ -77,7 +76,7 @@ def _build_usernl_files(case, model, comp):
                     shutil.copy(model_nl, nlfile)
 
 ###############################################################################
-def _case_setup_impl(case, caseroot, casebaseid, clean=False, test_mode=False, reset=False):
+def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False):
 ###############################################################################
     os.chdir(caseroot)
     msg = "case.setup starting"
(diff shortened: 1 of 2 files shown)


billsacks:
I don't think this works right. I added some testmods directories on a branch from this branch:

https://github.com/billsacks/cime/tree/user_mod_fix2_withTestDirs

following the example given in #719 

Then I created this test:

```
./create_test --test-root . --no-build SMS.f45_g37.A.yellowstone_intel.drv-ciso_abio
```

user_nl_cpl contained

```
ciso
default
ciso_abio
```

when it should have contained

```
default
ciso
abio
ciso_abio
```

billsacks:
Also, another change that should be made here is removing the code that checks whether the contents are already present in the file. With @mnlevy1981 's suggestion in #719 , this content check is unnecessary - and could actually do the wrong thing in some edge cases.

I think the diffs to remove this content-checking code are:

```diff
--- a/utils/python/CIME/user_mod_support.py
+++ b/utils/python/CIME/user_mod_support.py
@@ -72,19 +72,13 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
             run_cmd_no_fail(shell_command_file)

 def update_user_nl_file(case_user_nl, contents):
-    update_file = True
     if os.path.isfile(case_user_nl):
         with open(case_user_nl, "r") as fd:
             old_contents = fd.read()
-        if old_contents.find(contents) == -1:
-            contents = contents + old_contents
-            update_file = True
-        else:
-            update_file = False
-    if update_file:
-        logger.info("Pre-pending file %s"%(case_user_nl))
-        with open(case_user_nl, "w") as fd:
-            fd.write(contents)
+        contents = contents + old_contents
+    logger.info("Pre-pending file %s"%(case_user_nl))
+    with open(case_user_nl, "w") as fd:
+        fd.write(contents)



```

jedwards4b:
The content check is actually required.  If the case is rerun the user mods directories are traversed again and without the content check the file grows with duplicate settings.   You can't just remove the file before building it again because the user may make changes outside those applied by user mods.

jedwards4b:
Now does the right thing on example in your branch

billsacks:
I'll look at this more closely later this morning. For now I just want to comment on this:

> The content check is actually required. If the case is rerun the user mods directories are traversed again and without the content check the file grows with duplicate settings.

I'm wondering why this is done. In cime4, the testmods / user_mods were only applied at create_newcase time. What is the reason why this now needs to be done in case.setup, too? See also #692 . I'm not asking that this behavior be changed as part of this PR, but I'd like to understand if it can be changed later - i.e., removing this repeated application of testmods, and just applying them once at create_newcase time.

jedwards4b:
I don't recall why this was done in case.setup (it did make it much easier to test) I've removed it
and removed the check for duplicated code in the user_nl files.

billsacks:
I just realized why you put it in case.setup : https://github.com/ESMCI/cime/issues/734#issuecomment-256936910

Later this morning, I can take a stab at making the necessary changes in the NCK test to make sure it still works.

jedwards4b:
My last commit to user_mod_fix2 broke things because create_newcase isn't creating the user_nl_* files before trying to apply user mods.

mnlevy1981:
This looks good to me; @billsacks as we talked about earlier, can you hit the green button if you're happy with it as well?

PR #730, opened 2016-10-27
Initial port to ALCF Theta machine

amametjanov:
This is a port of CIME to the new Theta machine with KNL nodes.

Test suite: `cime_developer`, all 7 tests with status `Overall: PASS`
Test baseline: none
Test namelist changes: none
Test status: bit for bit
User interface changes?: none
Code review:

### DIFF
--- a/cime_config/acme/machines/config_batch.xml
+++ b/cime_config/acme/machines/config_batch.xml
@@ -51,6 +51,21 @@
      </submit_args>
    </batch_system>
 
+  <batch_system type="cobalt_theta" version="x.y">
+    <batch_query>qstat</batch_query>
+    <batch_submit>qsub</batch_submit>
+    <batch_directive>#COBALT</batch_directive>
+    <jobid_pattern>(\d+)</jobid_pattern>
+    <depend_string> --dependencies</depend_string>
+    <submit_args>
+      <arg flag="-A" name="PROJECT"/>
+      <arg flag="-t" name="JOB_WALLCLOCK_TIME"/>
+      <arg flag="-n" name=" $TOTALPES/$PES_PER_NODE"/>
+      <arg flag="-q" name="JOB_QUEUE"/>
+      <arg flag="--mode script"/>
+    </submit_args>
+  </batch_system>
+
   <batch_system type="lsf" version="9.1">
     <batch_query args=" -w" >bjobs</batch_query>
     <batch_submit>bsub</batch_submit>
@@ -193,6 +208,12 @@
     <batch_system MACH="cetus" type="cobalt">
       <queues>
         <queue walltimemax="00:59:00" jobmin="1" jobmax="4096" default="true">default</queue>
+      </queues>
+    </batch_system>
+
+    <batch_system MACH="theta" type="cobalt_theta">
+      <queues>
+        <queue walltimemax="00:60:00" jobmin="1" jobmax="3200" default="true">default</queue>
       </queues>
     </batch_system>
 
--- a/cime_config/acme/machines/config_compilers.xml
+++ b/cime_config/acme/machines/config_compilers.xml
@@ -590,6 +590,24 @@ for mct, etc.
   <MPICXX> CC </MPICXX>
 </compiler>
 
+<compiler COMPILER="intel" MACH="theta">
+  <MPIFC> ftn </MPIFC>
+  <MPICC> cc </MPICC>
+  <MPICXX> CC </MPICXX>
+  <ADD_FFLAGS DEBUG="FALSE"> -O2 -xMIC-AVX512 </ADD_FFLAGS>
+  <ADD_CFLAGS DEBUG="FALSE"> -O2 -xMIC-AVX512 </ADD_CFLAGS>
+  <ADD_FFLAGS DEBUG="TRUE"> -O0 -g -xMIC-AVX512 </ADD_FFLAGS>
+  <ADD_CFLAGS DEBUG="TRUE"> -O0 -g -xMIC-AVX512 </ADD_CFLAGS>
+  <ADD_CFLAGS compile_threaded="true"> -qopenmp </ADD_CFLAGS>
+  <ADD_FFLAGS compile_threaded="true"> -qopenmp </ADD_FFLAGS>
+  <ADD_FFLAGS_NOOPT compile_threaded="true"> -qopenmp </ADD_FFLAGS_NOOPT>
+  <ADD_LDFLAGS compile_threaded="true"> -qopenmp </ADD_LDFLAGS>
+  <CONFIG_ARGS> --host=Linux </CONFIG_ARGS>
+  <ADD_SLIBS> -L$(NETCDF_DIR)/lib -lnetcdff -L$(NETCDF_DIR)/lib -lnetcdf -Wl,-rpath -Wl,$(NETCDF_DIR)/lib </ADD_SLIBS>
+  <ADD_SLIBS> -mkl -lpthread -lm </ADD_SLIBS>
+  <ADD_CPPFLAGS> -DHAVE_COMM_F2C </ADD_CPPFLAGS>
+</compiler>
+
 <compiler COMPILER="gnu" MACH="mac">
   <NETCDF_PATH> $(NETCDF_PATH)</NETCDF_PATH>
   <ADD_LDFLAGS>-framework Accelerate</ADD_LDFLAGS>
--- a/utils/python/CIME/XML/env_mach_pes.py
+++ b/utils/python/CIME/XML/env_mach_pes.py
@@ -95,6 +95,8 @@ class EnvMachPes(EnvBase):
             pstrid = self.get_value("PSTRID_%s"%comp)
             tt = rootpe + (ntasks - 1) * pstrid + 1
             total_tasks = max(tt, total_tasks)
+        if total_tasks < self.get_value("PES_PER_NODE"): # all PEs fit in 1 node
+            self.set_value("PES_PER_NODE", total_tasks)
         return total_tasks
 
     def get_tasks_per_node(self, total_tasks, max_thread_count):
--- a/utils/python/CIME/XML/env_mach_pes.py
+++ b/utils/python/CIME/XML/env_mach_pes.py
@@ -95,8 +95,6 @@ class EnvMachPes(EnvBase):
             pstrid = self.get_value("PSTRID_%s"%comp)
             tt = rootpe + (ntasks - 1) * pstrid + 1
             total_tasks = max(tt, total_tasks)
-        if total_tasks < self.get_value("PES_PER_NODE"): # all PEs fit in 1 node
-            self.set_value("PES_PER_NODE", total_tasks)
         return total_tasks
 
     def get_tasks_per_node(self, total_tasks, max_thread_count):
--- a/utils/python/CIME/case.py
+++ b/utils/python/CIME/case.py
@@ -983,6 +983,12 @@ class Case(object):
             }
 
         executable, args = env_mach_specific.get_mpirun(self, mpi_attribs, job=job)
+        # special case for aprun if using < 1 full node
+        if executable == "aprun":
+            totalpes = self.get_value("TOTALPES")
+            pes_per_node = self.get_value("PES_PER_NODE")
+            if totalpes < pes_per_node:
+                args["tasks_per_node"] = "-N "+str(totalpes)
 
         mpi_arg_string = " ".join(args.values())
 
--- a/cime_config/acme/machines/config_machines.xml
+++ b/cime_config/acme/machines/config_machines.xml
@@ -272,6 +272,100 @@
 </machine>
 
 
+<machine MACH="theta">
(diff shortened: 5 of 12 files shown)


cime_config/acme/machines/config_batch.xml
```diff
@@ -51,6 +51,21 @@
      </submit_args>
    </batch_system>
 
+  <batch_system type="cobalt_theta" version="x.y">
```
jedwards4b: Why a new cobalt_theta section here instead of the cobalt definition already provided?

cime_config/acme/machines/config_compilers.xml
```diff
@@ -590,6 +590,24 @@ for mct, etc.
   <MPICXX> CC </MPICXX>
 </compiler>
 
+<compiler COMPILER="intel" MACH="theta">
```
jedwards4b: It seems that the os="CNL" section above will provide what you need here.

utils/python/CIME/XML/env_mach_pes.py
```diff
@@ -95,6 +95,8 @@ def get_total_tasks(self, comp_classes):
             pstrid = self.get_value("PSTRID_%s"%comp)
             tt = rootpe + (ntasks - 1) * pstrid + 1
             total_tasks = max(tt, total_tasks)
+        if total_tasks < self.get_value("PES_PER_NODE"): # all PEs fit in 1 node
```
jedwards4b: This is not required for any other system and I don't think that it's needed here.

cime_config/acme/machines/config_batch.xml
amametjanov: Cobalt on Mira uses `runjob <runjob_args> : <executable>`.
Cobalt on Theta uses `aprun <aprun_args> <executable>`.
The colon-separator is added elsewhere, outside of XML, and causes a job submission error.

There is also now support for `#COBALT ` directives similar to `#PBS ` directives and having a new batch system may be useful

cime_config/acme/machines/config_compilers.xml
amametjanov: The use of `-xMIC-AVX512` in compile flags is new and is needed for KNLs. Also, `-openmp` will cause lots of `deprecated` warnings in builds logs: this flag is now `-qopenmp` starting with Intel v17 compiler.

utils/python/CIME/XML/env_mach_pes.py
amametjanov: If a case has total number of tasks that is fewer than PES_PER_NODE, then it may cause job launch errors: e.g. `aprun -n 8 -N 64`.

Also, rather than adjusting upwards to `-n 64 -N 64`, it's probably better to run with expected total number of MPI tasks with `-n 8 -N 8`.

jedwards4b: Unless the behavior of aprun has changed this is not an error.

rljacob:
We shouldn't expect Theta to be like a normal cluster.

utils/python/CIME/XML/env_mach_pes.py
amametjanov: The error message is `aprun: -N cannot exceed -n`.
```
> aprun --version
aprun (ALPS) 6.1.6
Revision 5fa2c76
Network type Aries
```

jedwards4b: Okay but I think that we need another way to fix this - here's why: If a case is created with a default total_tasks < PES_PER_NODE the PES_PER_NODE value is changed for the job.   If the user then changes the total_tasks for the case so that it now exceeds the cases PES_PER_NODE value they will be using less than the full capacity of the node and will be unaware that the situation exists.

amametjanov: I'm somewhat unclear on the difference between MAX_TASKS_PER_NODE and PES_PER_NODE. I think MAX_TASKS_PER_NODE is actually max PEs per node. PE is either an MPI task (rank) or an OMP thread. And PES_PER_NODE is a user-editable value: i.e. a case can be run with different number of PEs per node.

In the above scenario, if a user changes number of ranks or threads for any of the components so that total_tasks becomes greater than PES_PER_NODE, then the user knows what they are doing and should also change PES_PER_NODE. Also, there are very few cases where total_tasks is less than PES_PER_NODE: most cases require more than 1 node to run on.

jedwards4b: I think that we can add code in env_mach_specific to assure that the value applied to -n is >= that applied to -N  .     I can help you figure that out if you find it a suitable alternative.

jedwards4b: It's not being changed in the code.   The default pelayout defined in driver_cpl/cime_config/config_pes.xml
is to use one full node (value -1 where negative means node count rather than task count) .  Acme gets its pe layouts from a different file and has a value of 8.   How did you ask for 8 MPI ranks?

jedwards4b:
I've pushed an update to your branch, please try it on theta.   Also it would be good to run the full scripts_regression_tests on that machine before we accept this PR.   See https://github.com/ESMCI/cime/wiki/Testing

jedwards4b:
Have you had a chance to test this yet?

utils/python/CIME/case.py
```diff
@@ -987,6 +987,12 @@ def get_mpirun_cmd(self, job="case.run"):
             }
 
         executable, args = env_mach_specific.get_mpirun(self, mpi_attribs, job=job)
+        # special case for aprun if using < 1 full node
+        if executable == "aprun":
+            totalpes = self.get_value("TOTALPES")
+            pes_per_node = self.get_value("PES_PER_NODE")
+            if totalpes < pes_per_node:
+                args["tasks_per_node"] = "-N "+str(totalpes)
```
amametjanov: The `-n ` argument of qsub also needs a condition to ensure that `$TOTALPES/$PES_PER_NODE` is not 0.

jedwards4b: Is this happening?   totalpes = 0 should have been captured as an error before we got to this point.

jedwards4b:
I did a rebase to the current trunk and made some changes on the cesm side, unfortunately I apparently do not have a valid account to run on the machine, so I cannot complete testing until I get that resolved.

utils/python/CIME/case.py
amametjanov: For example, if TOTALPES=8 and PES_PER_NODE=64, then $TOTALPES/$PES_PER_NODE=0 and the constructed job submission line is `qsub -n 0`, which is invalid.

jedwards4b: Are you actually testing anything?  There is no division happening here.  What are you looking at?

amametjanov: This is for qsub here:
https://github.com/ESMCI/cime/blob/90f1b9aaf80eccf742023f8d224354ed754e2f51/cime_config/acme/machines/config_batch.xml#L63

jedwards4b: Thanks for the clarification,  I've pushed a fix.

amametjanov:
Yes, `cime_developer` with `CIME_MODEL=acme` is a `PASS` for 7 out of 7.

jedwards4b:
Thanks - what about running scripts_regression_tests?

amametjanov:
Out of 95, 3 skipped, 2 failed. Failed tests were:
```
FAIL: test_bless_test_results (__main__.Q_TestBlessTestResults)
  File "./scripts_regression_tests.py", line 885, in test_bless_test_results
    self.simple_test(False, "%s -t %s" % (comparg, test_id))
  File "./scripts_regression_tests.py", line 862, in simple_test
    from_dir=self._testroot, expected_stat=(0 if expect_works else CIME.utils.TESTS_FAILED_ERR_CODE))
  File "./scripts_regression_tests.py", line 50, in run_cmd_assert_result
    test_obj.assertEqual(stat, expected_stat, msg=msg)
AssertionError:
COMMAND: /gpfs/theta-fs1/home/azamatm/repos/cime-master/scripts/Tools/wait_for_tests *fake_testing_only_20161102_021909-20161102_022147/TestStatus
FROM_DIR: /projects/EarlyPerf_theta/acme/scratch/azamatm
EXPECTED STAT 100, INSTEAD GOT STAT 0
OUTPUT: Test 'TESTRUNDIFF_P1.f19_g16_rx1.A.theta_intel' finished with status 'PASS'
    Path: TESTRUNDIFF_P1.f19_g16_rx1.A.theta_intel.C.fake_testing_only_20161102_021909-20161102_022147/TestStatus
ERRPUT:

FAIL: test_single_submit (__main__.X_TestSingleSubmit)
  File "./scripts_regression_tests.py", line 1073, in test_single_submit
    "unset CIME_GLOBAL_WALLTIME && %s/create_test SMS_Ln9_P8.f45_g37_rx1.A SMS_Ln9_P8.f19_g16_rx1.A  -t %s --single-submit" % (SCRIPT_DIR, self._baseline_name))
  File "./scripts_regression_tests.py", line 50, in run_cmd_assert_result
    test_obj.assertEqual(stat, expected_stat, msg=msg)
AssertionError:
COMMAND: unset CIME_GLOBAL_WALLTIME && /gpfs/theta-fs1/home/azamatm/repos/cime-master/scripts/create_test SMS_Ln9_P8.f45_g37_rx1.A SMS_Ln9_P8.f19_g16_rx1.A  -t fake_testing_only_20161102_022325 --single-submit
FROM_DIR: /gpfs/theta-fs1/home/azamatm/repos/cime-master/utils/python/tests
SHOULD HAVE WORKED, INSTEAD GOT STAT 1
OUTPUT: Using project from env PROJECT: Theta_ESP
RUNNING TESTS:
  SMS_Ln9_P8.f45_g37_rx1.A.theta_intel
  SMS_Ln9_P8.f19_g16_rx1.A.theta_intel
Starting CREATE_NEWCASE for test SMS_Ln9_P8.f45_g37_rx1.A.theta_intel with 1 procs
Starting CREATE_NEWCASE for test SMS_Ln9_P8.f19_g16_rx1.A.theta_intel with 1 procs
Finished CREATE_NEWCASE for test SMS_Ln9_P8.f45_g37_rx1.A.theta_intel in 0.807811 seconds (PASS)
Finished CREATE_NEWCASE for test SMS_Ln9_P8.f19_g16_rx1.A.theta_intel in 0.815436 seconds (PASS)
Starting XML for test SMS_Ln9_P8.f45_g37_rx1.A.theta_intel with 1 procs
Starting XML for test SMS_Ln9_P8.f19_g16_rx1.A.theta_intel with 1 procs
Finished XML for test SMS_Ln9_P8.f45_g37_rx1.A.theta_intel in 0.393628 seconds (PASS)
Finished XML for test SMS_Ln9_P8.f19_g16_rx1.A.theta_intel in 0.396442 seconds (PASS)
Starting SETUP for test SMS_Ln9_P8.f45_g37_rx1.A.theta_intel with 1 procs
Starting SETUP for test SMS_Ln9_P8.f19_g16_rx1.A.theta_intel with 1 procs
Finished SETUP for test SMS_Ln9_P8.f19_g16_rx1.A.theta_intel in 1.935636 seconds (PASS)
Finished SETUP for test SMS_Ln9_P8.f45_g37_rx1.A.theta_intel in 1.960016 seconds (PASS)
PASS SMS_Ln9_P8.f45_g37_rx1.A.theta_intel SETUP 
    Case dir: /projects/EarlyPerf_theta/acme/scratch/azamatm/SMS_Ln9_P8.f45_g37_rx1.A.theta_intel.fake_testing_only_20161102_022325
PASS SMS_Ln9_P8.f19_g16_rx1.A.theta_intel SETUP 
    Case dir: /projects/EarlyPerf_theta/acme/scratch/azamatm/SMS_Ln9_P8.f19_g16_rx1.A.theta_intel.fake_testing_only_20161102_022325
test-scheduler took 3.42278504372 seconds
Using project from env PROJECT: Theta_ESP
Script:
#! /bin/bash


cd /gpfs/theta-fs1/home/azamatm/repos/cime-master/utils/python/tests
/gpfs/theta-fs1/home/azamatm/repos/cime-master/scripts/create_test SMS_Ln9_P8.f45_g37_rx1.A SMS_Ln9_P8.f19_g16_rx1.A -t fake_testing_only_20161102_022325 --no-batch --use-existing --proc-pool 16 -m theta

RUN: qsub   -A Theta_ESP -t 00:60:00 -n 1 -q default --mode script
Usage: qsub.py --help
Usage: qsub.py [options] <executable> [<excutable options>]

Refer to man pages for JOBID EXPANSION and SCRIPT JOB DIRECTIVES.


  stat: 1
ERRPUT: No executable or script specified

ERROR: Command: 'qsub   -A Theta_ESP -t 00:60:00 -n 1 -q default --mode script' failed with error 'None'
```

jgfouca:
Don't worry about single submit. The other error is a bit concerning though.

jedwards4b:
I think that's because we haven't installed cprnc

jgfouca:
Ah, that would explain it.

Issue #369, opened 2016-08-10
Feature request: README file pointing from bld/run back to case directory


billsacks:
This is low priority; I'm mainly opening this issue to start a discussion of whether others would find this useful, and if so, what it should look like.

In debugging someone else's case, I often know where their run directory is, but not their case directory or source code directory. The only way I know to find these is to find a Filepath file in one of the bld directories; this works, but it's awkward.

It would be helpful if there were a README-paths file (or something else) in the bld and/or run directories pointing to the $CASEROOT and $SRCROOT for this case. Maybe there are some other paths that would be useful while we're at it, like $DOUT_S_ROOT (of course, you can find this out if you know $CASEROOT, but it could be convenient to have this directly available from the run directory).

At first I was thinking that this file should be a sibling of bld and run, so we'd have:

/glade/scratch/sacks/mycase
   README
   bld/
   run/

But I don't know if that will work, since bld and run can theoretically be in different paths. So maybe it's best for this to just go in both the bld and run directories? Or just the run directory? In that case, it might also be useful to have a pointer to $EXEROOT in the README in the run directory.

billsacks:
fixed by #739

PR #739, opened 2016-10-31
rename CESMSCRATCHROOT to CIME_OUTPUT_ROOT, add cli

### LINKED ISSUE #485: When specifying test root with ./create_test -r test_root, the run/ and bld/ directories should also change
`./create_test cime_developer -r $SCRATCH/acme_scratch/foo -t foo`

Will create `/global/cscratch1/sd/ndk/acme_scratch/foo` with the case directories as expected, however I also get the same case directory names in `/global/cscratch1/sd/ndk/acme_scratch` which contain only bld/ and run/ dirs.

In xml, we have:

``` <RUNDIR>$CESMSCRATCHROOT/$CASE/run</RUNDIR>
  <EXEROOT>$CESMSCRATCHROOT/$CASE/bld</EXEROOT>
```

Could we add $TESTROOT?  That is:
`<RUNDIR>$CESMSCRATCHROOT/$TESTROOT/$CASE/run</RUNDIR>`
Where it would normally be empty unless specified.

@jgfouca edit: Add a flag[s] to create_test to allow user to change rundir and exeroot potentially with a short cut option to have them go under the case directory.

### LINKED ISSUE #569: Evaluate $ENV{USER} before writing env_case.xml?
`env_case.xml` contains the following entry for `USER`:

```
<entry id="USER" value="$ENV{USER}">
```

`$USER` is then used in `$CESMSCRATCHROOT`, which appears in several other environment variables:

```
$ grep \$USER *
env_build.xml:    <entry id="CESMSCRATCHROOT" value="/glade/scratch/$USER">
$ grep \$CESMSCRATCHROOT *
env_build.xml:    <entry id="EXEROOT" value="$CESMSCRATCHROOT/$CASE/bld">
env_run.xml:    <entry id="RUNDIR" value="$CESMSCRATCHROOT/$CASE/run">
env_run.xml:    <entry id="DOUT_S_ROOT" value="$CESMSCRATCHROOT/archive/$CASE">
```

This means that when I do an `xmlquery` on any of those variables in a case created by someone else (@jedwards4b, for example), I see

```
$ cd /glade/scratch/jedwards/SMS_Ld1.f09_f09.FW2000.yellowstone_intel.cam-outfrq1d.GC.20160918_075947
$ ./xmlquery RUNDIR
run_desc:RUNDIR /glade/scratch/mlevy/SMS_Ld1.f09_f09.FW2000.yellowstone_intel.cam-outfrq1d.GC.20160918_075947/run type:char
                       The directory where the executable will be run.       By
                default this is set to EXEROOT/../run.       RUNDIR allows you to
                keep the run directory separate from the build directory
```

which is incorrect (note the `mlevy` instead of `jedwards`). I think the proper behavior is to populate `env_case.xml` with the actual user name rather counting on the xml tools to parse `$USER` from the environment.

This was pointed out by @klindsay28, who often evaluates other peoples' runs... so a common task for him is to start in a case directory that he did not create and pull things like `SRCROOT` and `RUNDIR` into his environment so that he can easily find files related to the run.

### LINKED ISSUE #737: create_clone with --keep-exe uses wrong EXEROOT
I did the following from `cime5.2.0-alpha.7`:

```
./create_clone -clone /glade/p/cesmdata/cseg/runs/cesm2_0/b.e20.B1850.f09_g16.pi_control.all.123 -case $scratch/cesm_tests/b.e20.B1850.f09_g16.pi_control.all.123 -keepexe
```

This let me submit without building, but it died at runtime: 

```
 991:numactl: execution of `/glade/scratch/sacks/b.e20.B1850.f09_g16.pi_control.all.123/bld/cesm.exe': No such file or directory
```

It looks like EXEROOT is set to point to my new case, instead of the bld directory in the cloned case.

### LINKED ISSUE #738: xmlquery EXEROOT in someone else's case gives the wrong path
I went into Cecile's case: `/glade/p/cesmdata/cseg/runs/cesm2_0/b.e20.B1850.f09_g16.pi_control.all.123` and did

```
./xmlquery EXEROOT
```

I get:

```
EXEROOT /glade/scratch/sacks/b.e20.B1850.f09_g16.pi_control.all.123/bld type:char
```

This suggests that Cecile's case is using an EXEROOT in my scratch space, which I'm sure isn't the case.

I think this is a problem with when environment variables are resolved. In particular, I'm guessing that the USER variable is resolved too late here, so that when I do an xmlquery from someone else's case, it treats me as the USER.

I think the solution there is to resolve $USER at create_newcase time and store it in an xml variable... then use that xml user variable in all other places. If I remember correctly, this is consistent with the suggestion made at a recent cseg meeting.

jedwards4b:
Add command line option for --output-root 

Test suite: scripts_regression_tests, create_test cime_developer --output-root /something/else
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #738 
Fixes #737 
Fixes #569 
Fixes #485 

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/acme/machines/config_machines.xml
+++ b/cime_config/acme/machines/config_machines.xml
@@ -60,8 +60,6 @@
   <TESTS>acme_developer</TESTS>
   <COMPILERS>intel,gnu,cray</COMPILERS>
   <MPILIBS>mpt,mpi-serial</MPILIBS>
-  <RUNDIR>$CIME_OUTPUT_ROOT/$CASE/run</RUNDIR>
-  <EXEROOT>$CIME_OUTPUT_ROOT/$CASE/bld</EXEROOT>
   <DOUT_S_ROOT>$CIME_OUTPUT_ROOT/archive/$CASE</DOUT_S_ROOT>
   <DOUT_L_MSROOT>csm/$CASE</DOUT_L_MSROOT>
   <OS>CNL</OS>
@@ -174,8 +172,6 @@
     <COMPILERS>intel,gnu,cray</COMPILERS>
     <MPILIBS>mpt,mpi-serial</MPILIBS>
     <CIME_OUTPUT_ROOT>$ENV{SCRATCH}/acme_scratch</CIME_OUTPUT_ROOT>
-    <RUNDIR>$CIME_OUTPUT_ROOT/$CASE/run</RUNDIR>
-    <EXEROOT>$CIME_OUTPUT_ROOT/$CASE/bld</EXEROOT>
     <DIN_LOC_ROOT>/project/projectdirs/acme/inputdata</DIN_LOC_ROOT>
     <DIN_LOC_ROOT_CLMFORC>/project/projectdirs/acme/inputdata/atm/datm7</DIN_LOC_ROOT_CLMFORC>
     <DOUT_S_ROOT>$CIME_OUTPUT_ROOT/archive/$CASE</DOUT_S_ROOT>
@@ -283,8 +279,6 @@
     <OS>Darwin</OS>
     <COMPILERS>gnu</COMPILERS>
     <MPILIBS>openmpi,mpich,mpi-serial</MPILIBS>
-    <RUNDIR>$ENV{HOME}/projects/acme/scratch/$CASE/run</RUNDIR>
-    <EXEROOT>$ENV{HOME}/projects/acme/scratch/$CASE/bld</EXEROOT>
     <DIN_LOC_ROOT>$ENV{HOME}/projects/acme/cesm-inputdata</DIN_LOC_ROOT>    <DIN_LOC_ROOT_CLMFORC>$ENV{HOME}/projects/acme/ptclm-data</DIN_LOC_ROOT_CLMFORC>
     <DOUT_S_ROOT>$ENV{HOME}/projects/acme/scratch/archive/$CASE</DOUT_S_ROOT>
     <DOUT_L_MSROOT>csm/$CASE</DOUT_L_MSROOT>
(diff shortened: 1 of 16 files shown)


jedwards4b:
Testing is still in progress.

jedwards4b:
Fixes #369

billsacks:
It looks like #569 is not actually fixed by this: I think it would still be good to resolve the $USER environment variable at create_newcase time rather than leaving it unresolved in the xml files. But if you'd like to defer changing that, we should at least remove the note saying that #569 is fixed by this PR.

jedwards4b:
The intent of #569 is fixed by this PR and I would argue that $USER should be removed from the XML.

billsacks:
> The intent of #569 is fixed by this PR and I would argue that $USER should be removed from the XML.

Okay, that makes sense - I agree.

This PR looks good to me - thanks for addressing all of this!

mnlevy1981:
> The intent of #569 is fixed by this PR and I would argue that $USER should be removed from the XML. 

Can you point to a case directory where this applies? I picked a recent dir from your scratch directory and it doesn't seem to address the issue in #569 at all:

```
$ cd /glade/scratch/jedwards/PEA_P1_Ld3.f45_g37_rx1.A.yellowstone_intel.20161031_102044_v0jl5s
$ ./xmlquery EXEROOT
EXEROOT /glade/scratch/mlevy/PEA_P1_Ld3.f45_g37_rx1.A.yellowstone_intel.20161031_102044_v0jl5s/bld        type:char
```

Not sure if I picked a bad example when checking up on this, though...

jedwards4b:
All tests pass except N_TestCreateTest currently failing on master

jedwards4b:
@mnlevy1981 yes you picked a bad example.   How about 
/glade/scratch/jedwards/cesmtests/ERS_N2_Ld3.f19_g16_rx1.A.yellowstone_intel.20161031_083157_ui7e1s

mnlevy1981:
> How about
/glade/scratch/jedwards/cesmtests/ERS_N2_Ld3.f19_g16_rx1.A.yellowstone_intel.20161031_083157_ui7e1s

Yup, that one looks good. Thanks!

jgfouca:
So are EXEROOT and RUNDIR now derived from CIME_OUTPUT_ROOT?

jedwards4b:
They are still defined in config_component.xml and still appear in the case, but yes they are now
derived from CIME_OUTPUT_ROOT.

jgfouca:
And certain machines can still override EXEROOT in their XML if they want?

jedwards4b:
They would change CIME_OUTPUT_ROOT, they could override EXEROOT if they wanted to but I'm not sure why they would.

jgfouca:
OK, I like this change.

Issue #741, opened 2016-10-31
Hard to find reason for NLCOMP failure in TestStatus.log


mnlevy1981:
Apologies for not giving concrete examples, I cleaned up a lot of old testing directories before realizing I should keep the results to include in this ticket. But basically, I had two different causes for `NLCOMP` failures in various tags -- one was a change in `user_nl_pop` and the other a change in `pop_in`. The `TestStatus` file just reads `FAIL ... NLCOMP`, and then the `TestStatus.log` file is several hundred lines long with no clear way to search for these namelist-generated fails. Could we include the `NLCOMP` keyword along with a message about which file(s) is/are to blame for the fail somewhere in the log?

jgfouca:
@mnlevy1981 sorry for taking so long to get around to this. The modifications were simple.

PR #743, opened 2016-11-01
Fix regression namelist regression test

jgfouca:
1) case.cmpgen_namelists: make better use of return code
to differentiate between fatal errors and namelist diffs.
1.a) reoganize expcetion handling to allow fatal errors to escape
but catch all non-fatal errors

2) Restore some smarts to test_scheduler so that it will report
NLFAIL tests as NLFAIL even though it is no longer directly responsible for
doing the namelist comparisons

Test baseline: 
Test namelist changes: 
Test status: bit for bit

User interface changes?: Return code changes for case.cmpgen_namelists. create_test will report test status more accurately.

Code review: jedwards

### DIFF
--- a/utils/python/CIME/test_scheduler.py
+++ b/utils/python/CIME/test_scheduler.py
@@ -12,7 +12,7 @@ import shutil, traceback, stat, threading, time
 from CIME.XML.standard_module_setup import *
 import CIME.compare_namelists
 import CIME.utils
-from CIME.utils import append_status
+from CIME.utils import append_status, TESTS_FAILED_ERR_CODE
 from CIME.test_status import *
 from CIME.XML.machines import Machines
 from CIME.XML.env_test import EnvTest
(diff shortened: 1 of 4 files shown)


utils/python/CIME/test_scheduler.py
```diff
@@ -721,12 +721,28 @@ def run_tests(self):
...
+                # self._test_states does not include comparison fail information,
+                # so we need to parse test status.
+                ts = TestStatus(self._get_test_dir(test))
+                nlfail = ts.get_status(NAMELIST_PHASE) == TEST_FAIL_STATUS
+                ts_status = ts.get_overall_test_status(ignore_namelists=True)
+
+                if ts_status not in [TEST_PASS_STATUS, TEST_PEND_STATUS]:
+                    logger.info( "%s %s (phase %s)" % (status, test, phase))
+                    rv = False
+                elif nlfail:
```
jedwards4b: Is Issue #741 addressed here?  Or can it be?

jgfouca: Unfortunately, no

jgfouca: I will address it in case_cmpgen_namelists in a future PR.

Issue #630, opened 2016-10-03
Question on use of logging vs. print


billsacks:
This is a question for @jedwards4b and @jgfouca . I'm asking it as an issue in case it's of interest to others.

I'm writing a little script to query the testlist xml files. Thus, the fundamental result of this script is output to stdout. I know we are generally using the logging infrastructure for output, and I see how that works well for tracking events that occur as various operations are performed. But for this use case, where output to stdout is the fundamental result, and it would make no sense to suppress the output, it feels to me like `print` is the right tool for the job.

This feeling is confirmed here

https://docs.python.org/2/howto/logging.html

For "Task you want to perform: Display console output for ordinary usage of a command line script or program" it lists "The best tool for the task: print()".

So, do you agree that I should use `print` in this case?

jedwards4b:
yes print is what you should use here.

billsacks:
Okay, thanks!

PR #748, opened 2016-11-02
Rework output format for compare_test_results

### LINKED ISSUE #429: Rework output of compare/bless_test_results to look more like TestStatus output
Moving a comment from PR #413 because it isn't critical to address this for that PR:

It would help if the output of compare_test_results (as well as bless_test_results, though that's less important) looked basically the same as the output put in the TestStatus file - although possibly just for the COMPARE_baseline phase. That is, I'd want to see something like:

PASS   SOMETEST  BASELINE_COMPARE

or

FAIL   SOMETEST  BASELINE_COMPARE

That's approximately what happened with the earlier version of these tools.

This comment applies both to the main output from the comparison, and also to various other extraneous information that is printed (or not). So, for example, I'd prefer not to have this output for every test:

``` python
                logging.info("Comparing results for test: %s, most recent result: %s" % (test_name, overall_result))
```

-- because I'd like to be able to parse this output in the same way that I parse the normal TestStatus or cs.status output. 

However, I agree that it does make sense to report on failed tests here. One possibility would be to print out the full testStatus output for each test that failed.

billsacks:
Simplifies output sent to stdout from compare_test_results, for easier
parsing. Now each test outputs a single line to stdout, with a status
code of PASS, FAIL or SKIP. SKIP is given for tests that either didn't
have a RUN phase, or whose RUN phase didn't pass. The output format
essentially matches the BASELINE line in TestStatus - e.g.

FAIL SMS.f10_f10.ICLM45.yellowstone_intel.clm-default BASELINE [comment]

In addition, in each test compared, a new file is put in the test (i.e.,
case) directory: compare.log.BASELINE_NAME.TIMESTAMP. This file contains
extra details, similar to what is put in TestStatus.log during the main
run of the test suite. Also, in each run directory, cprnc output files
are created with names like *.nc.cprnc.out.BASELINE_NAME.TIMESTAMP. Note
that the suffix is added to prevent files from stomping on existing
files.

I have changed the return code from compare_test_results so that a
non-zero return code is only generated if the script exits
abnormally. In particular, it exits with a 0 return code if the script
succeeded but differences were found. Personally, I prefer if non-zero
return codes are reserved for cases where there was some real
failure. I can change this back if others prefer, although I don't love
the old behavior where (I think) a 1 return code was generated either
when differences were found or when the script exited abnormally: I feel
it's important to distinguish abnormal exits.

   Also some manual testing
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes #429

User interface changes?:
- Significant change in output format and return code from
  compare_test_results
- New -output_suffix argument to summarize_cprnc_diffs

Code review:

### DIFF
--- a/scripts/Tools/compare_test_results
+++ b/scripts/Tools/compare_test_results
@@ -3,14 +3,19 @@
 """Analyze results from a test root area, comparing non-BFB changes.
 
 Purpose is, instead of re-running tests in compare mode, which is very slow,
-allow for very fast analsis of diffs.
+allow for very fast analysis of diffs.
 
 Outputs results for each test to stdout (one line per test); possible status
-codes are: PASS, FAIL, SKIP.
+codes are: PASS, FAIL, SKIP. (A SKIP denotes a test that did not make it to the
+run phase or a test for which the run phase did not pass: we skip baseline
+comparisons in this case.)
 
 In addition, creates files named compare.log.BASELINE_NAME.TIMESTAMP in each
 test directory, which contain more detailed output. Also creates
 *.cprnc.out.BASELINE_NAME.TIMESTAMP files in each run directory.
+
+Returns a 0 exit status if all tests are bit-for-bit, and a non-zero exit status
+(TESTS_FAILED_ERR_CODE) if any tests differed from the baseline.
 
 You may need to load modules for cprnc to work.
 
@@ -95,8 +100,8 @@ def _main_func(description):
     baseline_name, baseline_root, test_root, compiler, test_id, compare_tests = \
         parse_command_line(sys.argv, description)
 
-    compare_test_results(baseline_name, baseline_root, test_root, compiler, test_id, compare_tests)
-    sys.exit(0)
+    success = compare_test_results(baseline_name, baseline_root, test_root, compiler, test_id, compare_tests)
+    sys.exit(0 if success else CIME.utils.TESTS_FAILED_ERR_CODE)
 
 ###############################################################################
 
--- a/scripts/Tools/compare_test_results
+++ b/scripts/Tools/compare_test_results
@@ -1,11 +1,19 @@
 #!/usr/bin/env python
 
-"""
-Analyze results from a test root area, comparing non-BFB
-changes. Purpose is, instead of re-running tests
-in compare mode, which is very slow, allow for very fast analsis of diffs.
+"""Analyze results from a test root area, comparing non-BFB changes.
+
+Purpose is, instead of re-running tests in compare mode, which is very slow,
+allow for very fast analsis of diffs.
+
+Outputs results for each test to stdout (one line per test); possible status
+codes are: PASS, FAIL, SKIP.
+
+In addition, creates files named compare.log.BASELINE_NAME.TIMESTAMP in each
+test directory, which contain more detailed output. Also creates
+*.cprnc.out.BASELINE_NAME.TIMESTAMP files in each run directory.
 
 You may need to load modules for cprnc to work.
+
 """
 
 from standard_script_setup import *
--- a/scripts/Tools/compare_test_results
+++ b/scripts/Tools/compare_test_results
@@ -87,8 +87,8 @@ def _main_func(description):
     baseline_name, baseline_root, test_root, compiler, test_id, compare_tests = \
         parse_command_line(sys.argv, description)
 
-    success = compare_test_results(baseline_name, baseline_root, test_root, compiler, test_id, compare_tests)
-    sys.exit(0 if success else 1)
+    compare_test_results(baseline_name, baseline_root, test_root, compiler, test_id, compare_tests)
+    sys.exit(0)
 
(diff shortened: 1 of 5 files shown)


billsacks:
Note that I have _not_ changed the output format of bless_test_results, because that is less critical for my workflow and I didn't want to mess with the workflow of @jgfouca and others. (To be honest, I haven't even looked at the output format from that script.)

scripts/Tools/compare_test_results
```diff
@@ -1,11 +1,19 @@
 #!/usr/bin/env python
 
-"""
-Analyze results from a test root area, comparing non-BFB
-changes. Purpose is, instead of re-running tests
-in compare mode, which is very slow, allow for very fast analsis of diffs.
+"""Analyze results from a test root area, comparing non-BFB changes.
+
+Purpose is, instead of re-running tests in compare mode, which is very slow,
+allow for very fast analsis of diffs.
```
jgfouca: You inherited my spelling mistake "analsis"

scripts/Tools/compare_test_results
```diff
@@ -87,8 +95,8 @@ def _main_func(description):
     baseline_name, baseline_root, test_root, compiler, test_id, compare_tests = \
         parse_command_line(sys.argv, description)
 
-    success = compare_test_results(baseline_name, baseline_root, test_root, compiler, test_id, compare_tests)
-    sys.exit(0 if success else 1)
+    compare_test_results(baseline_name, baseline_root, test_root, compiler, test_id, compare_tests)
```
jgfouca: I would like to preserve the behavior of a non-zero error code for diffs. One thing I've done in many places in CIME (which I probably should have done here too) is to return CIME.utils.TESTS_FAILED_ERR_CODE when everything works but there are diffs. See create_test as an example.

scripts/Tools/compare_test_results
```diff
@@ -1,11 +1,19 @@
...
-Analyze results from a test root area, comparing non-BFB
-changes. Purpose is, instead of re-running tests
-in compare mode, which is very slow, allow for very fast analsis of diffs.
+"""Analyze results from a test root area, comparing non-BFB changes.
+
+Purpose is, instead of re-running tests in compare mode, which is very slow,
+allow for very fast analsis of diffs.
+
+Outputs results for each test to stdout (one line per test); possible status
+codes are: PASS, FAIL, SKIP.
```
jgfouca: In what cases would we see SKIP? Missing baseline?

utils/python/CIME/compare_test_results.py
```diff
@@ -7,64 +9,86 @@
...
+                success, detailed_comments = compare_history(test_dir, baseline_name, baseline_root, log_id)
+                if success:
+                    compare_result = TEST_PASS_STATUS
+                else:
+                    compare_result = TEST_FAIL_STATUS
+
+            brief_result = "%s %s %s"%(compare_result, test_name, BASELINE_PHASE)
+            if compare_comment:
+                brief_result += " %s"%(compare_comment)
+            print(brief_result)
```
jgfouca: you probably want logging.info here instead of print

utils/python/tests/scripts_regression_tests.py
```diff
@@ -884,12 +884,12 @@ def test_bless_test_results(self):
         test_id = "%s-%s" % (self._baseline_name, CIME.utils.get_timestamp())
         self.simple_test(False, "%s -t %s" % (comparg, test_id))
 
-        # compare_test_results should detect the fail
+        # compare_test_results should run to completion
         cpr_cmd = "%s/compare_test_results -b %s -t %s 2>&1" % (TOOLS_DIR, self._baseline_name, test_id)
-        output = run_cmd_assert_result(self, cpr_cmd, expected_stat=1)
```
jgfouca: run cmds in scripts_regression_tests should go through run_cmd_assert_result even if they are expected to pass. just remove the expected_stat argument so it defaults to 0.

fischer-ncar:
I have no issues with the output format changes.

scripts/Tools/compare_test_results
billsacks: A SKIP denotes a test that did not make it to the run phase or a test for which the run phase did not pass: we skip baseline comparisons in this case. I'll add this to the documentation.

scripts/Tools/compare_test_results
billsacks: Okay, thanks - I'll use TESTS_FAILED_ERR_CODE. I'll return that error code if there were any FAIL results, though will still return an exit code of 0 if there were some SKIP results. (Rationale: I figure it's not the responsibility of this script to determine if tests didn't pass - its responsibility is just to determine if there are any baseline comparison differences among tests that passed.)

utils/python/CIME/compare_test_results.py
billsacks: I chose print based on the discussion here: #630

PR #755, opened 2016-11-02
Better handling of timeouts for case.test

### LINKED ISSUE #610: Tests that are running or have hung are reported as PEND
It appears that the status for the run phase is listed as PEND until the run completes. This is a departure from cime4, in which a run was given the status of RUN once it started running. I prefer the cime4 behavior: It's helpful to see which tests are truly pending in the queue and which are running. This is particularly helpful for tests that have exited due to hanging and running out of wallclock time: such tests currently have a final status of PEND (rather than RUN in cime4).

I have a nagging feeling that this was discussed at some point, but I can't remember the details.... There may have been an argument about not introducing more status codes, but I'd personally prefer to have one more status code that prevents this misleading PEND status.

An example test that currently hangs in CESM is `SMS_D_Ld1_P24x1.f10_f10.ICRUCLM45.hobart_nag.clm-af_bias_v5`

(See also #383 for some initial discussion of this issue.)

jgfouca:
Add signal handlers for SIGTERM and SIGINT to ensure that a python
exception is generated when the case.test process is hit with these
signals. slurm documentation states that it will use a SIGTERM before
issuing a SIGKILL.

These changes will ensure that the RUN phase will be FAIL for timed-out
tests. Also, the last line in TestStatus.log will be 'test timed out'

Test suite: by hand + code checker
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #610

User interface changes?: Better output in TestStatus and TestStatus.log when a problem times out

Code review: billsacks

### DIFF
--- a/utils/python/CIME/case_test.py
+++ b/utils/python/CIME/case_test.py
@@ -8,12 +8,28 @@ from CIME.SystemTests.system_tests_common import *
 
 import sys, signal
 
-def _signal_handler(*_):
-    expect(False, "Problem timed out!")
+def _iter_signal_names():
+    for signame in [item for item in dir(signal) if item.startswith("SIG") and not item.startswith("SIG_")]:
+        yield signame
+
+def _signal_handler(signum, _):
+    name = "Unknown"
+    for signame in _iter_signal_names():
+        if signum == getattr(signal, signame):
+            name = signame
+
+    expect(False, "Job killed due to receiving signal %d (%s)" % (signum, name))
 
 def _set_up_signal_handlers():
-    signal.signal(signal.SIGTERM, _signal_handler)
-    signal.signal(signal.SIGINT, _signal_handler)
+    """
+    Add handles for all signals that might be used to abort a test
+
+    We need to handle a wide variety due to different implementations of the
+    timeout mechanism for different batch systems.
+    """
+    for signame in ["SIGINT", "SIGTERM", "SIGXCPU", "SIGUSR1", "SIGUSR2"]:
+        signum = getattr(signal, signame)
+        signal.signal(signum, _signal_handler)
 
 def case_test(case, testname=None):
     if testname is None:
--- a/utils/python/CIME/case_test.py
+++ b/utils/python/CIME/case_test.py
@@ -3,7 +3,7 @@ Run a testcase.
 """
 
 from CIME.XML.standard_module_setup import *
-from CIME.utils import expect, find_system_test, append_status
+from CIME.utils import expect, find_system_test, append_status, find_proc_id
 from CIME.SystemTests.system_tests_common import *
 
 import sys, signal
@@ -18,6 +18,17 @@ def _signal_handler(signum, _):
         if signum == getattr(signal, signame):
             name = signame
 
+    # Terminate children
+    proc_ids = find_proc_id(children_only=True)
+    for proc_id in proc_ids:
+        try:
+            os.kill(proc_id, signal.SIGKILL)
+        except OSError:
+            # If the batch system killed the entire process group, these
+            # processes might already be dying
+            pass
+
+    # Throw an exception so SystemTest infrastructure can handle this error
     expect(False, "Job killed due to receiving signal %d (%s)" % (signum, name))
 
 def _set_up_signal_handlers():
--- a/utils/python/CIME/case_test.py
+++ b/utils/python/CIME/case_test.py
@@ -6,7 +6,14 @@ from CIME.XML.standard_module_setup import *
 from CIME.utils import expect, find_system_test, append_status
 from CIME.SystemTests.system_tests_common import *
 
-import sys
+import sys, signal
+
+def _signal_handler(*_):
+    expect(False, "Problem timed out!")
+
+def _set_up_signal_handlers():
+    signal.signal(signal.SIGTERM, _signal_handler)
+    signal.signal(signal.SIGINT, _signal_handler)
 
 def case_test(case, testname=None):
     if testname is None:
@@ -14,6 +21,8 @@ def case_test(case, testname=None):
 
     expect(testname is not None, "testname argument not resolved")
     logging.warn("Running test for %s" % testname)
+
+    _set_up_signal_handlers()
 
     try:
         # The following line can throw exceptions if the testname is


billsacks:
This doesn't appear to work on yellowstone: I set up a test that timed out (with --walltime 0:01). This still had:

```
PEND SMS_D_Ly1.f10_f10.ICLM45BGC.yellowstone_intel RUN
```

and did not have the new message in TestStatus.log.

I also tried killing a job with bkill (the equivalent of qdel), but with the same result (it's still listed as PEND, and does not have the new message).

utils/python/CIME/case_test.py
```diff
@@ -6,7 +6,14 @@
 from CIME.utils import expect, find_system_test, append_status
 from CIME.SystemTests.system_tests_common import *
 
-import sys
+import sys, signal
+
+def _signal_handler(*_):
+    expect(False, "Problem timed out!")
```
billsacks: Rather than 'Problem timed out', I'm thinking this should be more general, like, "Job killed due to signal" - because other things can send signals like the user deleting the job, right?

billsacks:
This doesn't appear to work on hobart, either, which uses PBS.

I also tried adding this on yellowstone:

```diff
--- a/utils/python/CIME/case_test.py
+++ b/utils/python/CIME/case_test.py
@@ -14,6 +14,8 @@ def _signal_handler(*_):
 def _set_up_signal_handlers():
     signal.signal(signal.SIGTERM, _signal_handler)
     signal.signal(signal.SIGINT, _signal_handler)
+    # should probably wrap the following in try/except, to catch an AttributeError, in case some systems don't have SIGXCPU
+    signal.signal(signal.SIGXCPU, _signal_handler)

 def case_test(case, testname=None):
     if testname is None:
```

because apparently LSF sends SIGXCPU when a job exceeds wallclock time. This still didn't have the expected result: I still get:

```
PEND SMS_D_Ly1.f10_f10.ICLM45BGC.yellowstone_intel RUN
```

but interestingly it led to the job sitting in the queue (in the running state) for about 10 min longer than the allowed wallclock time.

utils/python/CIME/case_test.py
jgfouca: True. I will do that.

jgfouca:
That's odd. I tested this on skybridge with this:

./create_test SEQ_Ln9.f19_g16_rx1.A --walltime="00:00:01"

... and it worked perfectly.

jgfouca:
I did have to put a sleep in SEQ to get it to last more than a minute. I can't test any real problems because ACME isn't using CIME5 yet. I wonder if that's why it worked.

jgfouca:
I'll try changing seq to run a long shell command and see if it still works.

jgfouca:
@billsacks I removed the Ln9 from my SEQ case and that was good enough to get it to run for over a minute, so I was able to take out the sleep and therefore get a more realistic test. Unfortunately, this PR still seems to work perfectly on skybridge in its current implementation.

Does yellowstone use slurm?

billsacks:
Yellowstone uses IBM's LSF. Cheyenne (the replacement for yellowstone) will use PBS, which didn't work for me on hobart.

If you'd like, I'm fine with your bringing this to master as is, with the understanding that it doesn't fully address #610 and we'd still like to come up with a more general solution for that issue.

jgfouca:
@billsacks on the systems that aren't working, could you please submit this python program:

```
#!/usr/bin/env python
# Batch system directives

#SBATCH  --job-name=foo
#SBATCH  --nodes=1
#SBATCH  --ntasks-per-node=1
#SBATCH  --output=foo.out
#SBATCH  --exclusive
#SBATCH  --time=00:00:02
#SBATCH  --partition=ec
#SBATCH  --account=fy150001

import signal, sys, time

def sighandler(signum, stack):
    print 'Caught signal %s, exiting.' % (str(signum))
    sys.exit()

for i in [x for x in dir(signal) if x.startswith("SIG")]:
    try:
        signum = getattr(signal,i)
        signal.signal(signum, sighandler)
    except:
        print "Skipping %s"%i

time.sleep(999)
def sighandler(signum, stack):
    print 'Caught signal %s, exiting.' % (str(signum))
    sys.exit()

for i in [x for x in dir(signal) if x.startswith("SIG")]:
    try:
        signum = getattr(signal,i)
        signal.signal(signum, sighandler)
    except:
        print "Skipping %s"%i

time.sleep(999)
```

... and report back to me on the output? You'll need to change the directives to match the batch system of the machines. An easy way to get the directives is to copy them from the top of  case.test from a case.

jgfouca:
@billsacks , just a moment, let me give you a better version of that program.

billsacks:
Sorry for the delay... I was waiting for a notification that you had updated this, then realized that you must have just edited your original comment - darn github's lack of notifications for edits!

On yellowstone (lsf), the output is:

```
Skipping SIGKILL
Skipping SIGSTOP
Skipping SIG_DFL
Caught signal 12, exiting.
```

and my job message (i.e., email) showed:

```
TERM_RUNLIMIT: job killed after reaching LSF run time limit.
Exited with signal termination: Stopped (signal).
```

On hobart (pbs), I'm having trouble getting a python job to run on the batch system, but running the script from the login node, I get the same "skipping" lines as above. If you want to know the behavior of the batch job on hobart, I'll need to enlist Jim E.

jgfouca:
@billsacks or @jedwards4b please submit this python program on yellowstone and/or hobart.

```
import signal, sys, time

def sighandler(signum, stack):
    for i in [x for x in dir(signal) if x.startswith("SIG")]:
        if signum == getattr(signal, i):
            name = i

    print 'Caught signal %s, (%s).' % (str(signum), name)
    sys.stdout.flush()

def set_all_signals():
    for i in [x for x in dir(signal) if x.startswith("SIG")]:
        try:
            signum = getattr(signal,i)
            signal.signal(signum, sighandler)
        except:
            print "Skipping %s"%i
            sys.stdout.flush()

def _main_func():
    set_all_signals()
    time.sleep(999)

if __name__ == "__main__":
    _main_func()
```

I left off the batch directives since I don't know what they're supposed to be on those machines. Just make sure the timeout is set very short.

billsacks:
I ran it on yellowstone and got:

```
Skipping SIGKILL
Skipping SIGSTOP
Skipping SIG_DFL
Caught signal 12, (SIGUSR2).
```

I guess that means, "user-defined signal 2". Does that imply that you'd need to set up a signal handler for all possible signals for this to work robustly on all systems?

I'll have to leave running on hobart to @jedwards4b if it's desired, because I'm not sure how to get a python program off the ground there.

jgfouca:
Ah! That's why this PR isn't working on yellowstone (I wasn't handling SIGUSR2).

jgfouca:
@billsacks OK, plz do a pull and try it again.

jgfouca:
@billsacks I forgot to answer your question. I initially thought as you did that I'd have to handle every possible signal; however, there are many signals that are routine and harmless (SIGCHD) that should not cause the test to be aborted.

jedwards4b:
@jgfouca have you tested in slurm?  I've noticed that wait_for_tests hangs on slurm when sbatch fails, but this is on master and not with this PR.

jgfouca:
@jedwards4b yes, I'm testing this on skybridge which is a slurm machine. This PR should ensure that wait-for-tests does not hang when the jobs timeout.

billsacks:
@jgfouca - Thanks for all of your work on this.

I have run your latest version on yellowstone, and it now behaves as it should, in terms of reporting in TestStatus and TestStatus.log. Thank you! However, one oddity is that the job still remained in the RUN state in the queue for about 10 minutes longer than the allotted wallclock time. That is: When I gave the job a 1 minute wallclock limit, the signal was caught after about 1 minute, and reported properly in TestStatus and TestStatus.log, but according to bjobs (and according to when I received the job email) it took another 10 minutes for the job to actually die. It seems that something about catching that signal prevents a timely exit. I can live with this, though.

On hobart, I realized that I could run your test script by sticking it into a case.test script. Doing that, I get this output:

```
Skipping SIGKILL
Skipping SIGSTOP
Skipping SIG_DFL
=>> PBS: job killed: walltime 73 exceeded limit 60
-------------------- Post Job Clean Up --------------------

Running cleanipcs as sacks on h020 Killed by signal 15.
```

Adapting your script, I determined that signal 15 is SIGTERM. It's not clear to me why that isn't being trapped, then.

utils/python/CIME/case_test.py
```diff
@@ -6,7 +6,30 @@
...
+    expect(False, "Job killed due to receiving signal %d (%s)" % (signum, name))
+
+def _set_up_signal_handlers():
+    """
+    Add handles for all signals that might be used to abort a test
+
+    We need to handle a wide variety due to different implementations of the
+    timeout mechanism for different batch systems.
+    """
+    for signame in ["SIGINT", "SIGTERM", "SIGXCPU", "SIGUSR1", "SIGUSR2"]:
```
jedwards4b: Just a thought - perhaps the singles trapped for each batch system type are different and should be defined in config_batch.xml?

billsacks:
Other than the line comment made by @jedwards4b and myself, I'm happy with this PR. It doesn't work on all systems (specifically hobart), but at least it's a step in the right direction.

jedwards4b:
Here is output from Hobart:
stderr:
```
=>> PBS: job killed: walltime 26 exceeded limit 5
Killed by signal 15.
```
stdout:
```
Skipping SIGKILL
Skipping SIGSTOP
Skipping SIG_DFL
-------------------- Post Job Clean Up --------------------

Running cleanipcs as jedwards on h020 

--------------------------- End ---------------------------
```

jedwards4b:
this is cori
```
Skipping SIGKILL
Skipping SIGSTOP
Skipping SIG_DFL
slurmstepd: error: *** JOB 3050848 ON nid00119 CANCELLED AT 2016-11-07T08:21:08 DUE TO TIME LIMIT ***
Caught signal 18, (SIGCONT).
```

utils/python/CIME/case_test.py
jgfouca: Not a bad idea longerm. I think for now this list of signals is safe to handle on any batch system.

jedwards4b:
It looks like SIGCONT isn't in your list yet.

jgfouca:
@billsacks we are now in the fun world of signal handling. I assume the reason for the 10 minute delay is that the batch system on yellowstone is configured to hit the parent process (not the process group) with a polite signal (SIGUSR2) to ask it nicely to please stop. It then waits 10 minutes for all processes to clear and then hits everything with SIGKILL. Since it hit the parent and not the process group, all the children of the parent (the mpirun processes for example) continue to run, hence why it took an additional 10 minutes to clear.

I'm going to modify the code to force the signal to propagate to all children, that should help.

jgfouca:
SIGCONT is used for job control so I don't think it's safe to catch that one. Slurm should follow up SIGCONT with SIGTERM.

jgfouca:
@billsacks I just pushed a commit, plz see if the 10 minute wait on yellowstone is gone.

billsacks:
Nope, looks like there's still a 10-minute delay before the job is killed. Maybe this isn't a big deal, though.

jedwards4b:
Can we merge this?

billsacks:
> Can we merge this?

I'm okay with merging this, though I wasn't sure what we want to do about the fact that @jgfouca 's hopeful fix for the extra 10 minute delay on yellowstone doesn't work: Does that indicate a bug in that fix, or just that there's something else going on on yellowstone and we should give up on this problem? I'm partly wondering if we should take that last change out, since it seems like it adds complexity without accomplishing anything, at least on yellowstone. (Incidentally: I just ran a one-node job (`SMS_D_Ly1_P15x1.f10_f10.ICLM45BGC.yellowstone_intel`) as opposed to the two-node jobs I was running before, thinking that the problem may be lack of communication across nodes, but it still hung for the extra 10 minutes.)

And should we then open an issue ("enhancement") to eventually move this machine-specific signal into env_batch, or elsewhere?

jgfouca:
@billsacks I'm not sure what else I can do to eliminate the 10 minute delay on yellowstone, so I'm fine with merging for now.

PR #757, opened 2016-11-02
user_mods ran through xmlchange in the wrong order

### LINKED ISSUE #756: More issues with user_mods and test dependencies
Similar to #719 (and the issue it raised with how #712 was fixed), if a testmod_dir depends on another testmod_dir and they want to change the same XML variable to different values, the test higher up the chain should take precendent. Using the example from #719, if our test is

```
          ciso_abio
          /       \
        abio     ciso
        /          \
    default     default
```

Then `xmlchange` commands are being called in the following order:

```
ciso_abio
abio
ciso
default
```

whereas we want

```
default
ciso
abio
ciso_abio
```

Note that this is an issue even if we are only dependent on a single test. If our test is

```
ecosys_r4
    |
 ecosys
    |
 default
```

Then we expect `xmlchange` calls from `ecosys_r4` to take precedence over those in `default` but in practice the order is reversed. I have a branch that fixes this (pull request coming shortly) but it needs a critical set of eyes on it.

mnlevy1981:
Reverse the order that `xmlchange` commands are called

Test suite: none, just tested by hand
Test baseline: n/a
Test namelist changes: none
Test status: should be bit-for-bit unless there is a test dependency that is currently being mishandled

Fixes #756

User interface changes?: None

Code review: (assigning to Jim and Bill because they seem more familiar than me with python in general and the cime scripts in particular)

If test A depends on test B, and both run xmlchange on the same variable, then
the expected behavior is that the value set by test A would overwrite the value
set by test B. Unfortunately, that was not the case... the value from test B
would overwrite test A. This commit fixes that.

### DIFF
--- a/utils/python/CIME/user_mod_support.py
+++ b/utils/python/CIME/user_mod_support.py
@@ -63,8 +63,17 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
             case_shell_commands = shell_commands_file.replace(include_dir, caseroot)
             with open(shell_commands_file,"r") as fd:
                 new_shell_commands = fd.read().replace("xmlchange","xmlchange --force")
-            with open(case_shell_commands, "a") as fd:
+            # prepend new_shell_commands to case_shell_commands file
+            old_shell_commands = None
+            if os.path.isfile(case_shell_commands):
+                with open(case_shell_commands, "r") as fd:
+                    old_shell_commands = fd.read()
+                    fd.close()
+            with open(case_shell_commands, "w") as fd:
                 fd.write(new_shell_commands)
+                if old_shell_commands is not None:
+                    fd.write(old_shell_commands)
+                fd.close()
 
     for shell_command_file in case_shell_command_files:
         if os.path.isfile(shell_command_file):


utils/python/CIME/user_mod_support.py
```diff
@@ -63,8 +63,17 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
             case_shell_commands = shell_commands_file.replace(include_dir, caseroot)
             with open(shell_commands_file,"r") as fd:
                 new_shell_commands = fd.read().replace("xmlchange","xmlchange --force")
-            with open(case_shell_commands, "a") as fd:
+            # prepend new_shell_commands to case_shell_commands file
+            old_shell_commands = None
+            if os.path.isfile(case_shell_commands):
```
jedwards4b: the fd.close is implied, you don't need it.   It seems like you should build up the shell commands in memory and only write them once instead of repeatedly opening and closing the file:
shell_commands = new_shell_commands + old_shell_commands

jedwards4b:
Prefer #758

mnlevy1981:
closing in favor of #758

PR #758, opened 2016-11-02
Another possible fix to #756

### LINKED ISSUE #756: More issues with user_mods and test dependencies

mnlevy1981:
Reverse the order that xmlchange commands are called (pick this PR or #757 not both)

Test suite: none, just tested by hand
Test baseline: n/a
Test namelist changes: none
Test status: should be bit-for-bit unless there is a test dependency that is currently being mishandled

Fixes #756

User interface changes?: None

Code review: (assigning to Jim and Bill because they seem more familiar than me with python in general and the cime scripts in particular)

If test A depends on test B, and both change the same XML variable, expected
behavior is for test A's value to take precedence.

### DIFF
--- a/utils/python/CIME/user_mod_support.py
+++ b/utils/python/CIME/user_mod_support.py
@@ -66,6 +66,11 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
                               glob.glob(os.path.join(include_dir,"xmlchange_cmnds"))
         for shell_commands_file in shell_command_files:
             case_shell_commands = shell_commands_file.replace(include_dir, caseroot)
+            # add commands from both shell_commands and xmlchange_cmnds to
+            # the same file (caseroot/shell_commands)
+            # Note that use of xmlchange_cmnds has been deprecated and will soon
+            # be removed altogether, so new tests should rely on shell_commands
+            case_shell_commands = case_shell_commands.replace("xmlchange_cmnds","shell_commands")
             with open(shell_commands_file,"r") as fd:
                 new_shell_commands = fd.read().replace("xmlchange","xmlchange --force")
             with open(case_shell_commands, "a") as fd:
--- a/utils/python/CIME/user_mod_support.py
+++ b/utils/python/CIME/user_mod_support.py
@@ -56,6 +56,8 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
                             expect(False, "Could not write file %s in caseroot %s"
                                    %(case_source_mods,caseroot))
 
+    include_dirs.reverse()
+    for include_dir in include_dirs:
         # create xmlchange_cmnds and shell_commands in caseroot
         shell_command_files = glob.glob(os.path.join(include_dir,"shell_commands")) +\
                               glob.glob(os.path.join(include_dir,"xmlchange_cmnds"))
--- a/utils/python/CIME/user_mod_support.py
+++ b/utils/python/CIME/user_mod_support.py
@@ -68,9 +68,13 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
             case_shell_commands = shell_commands_file.replace(include_dir, caseroot)
             # add commands from both shell_commands and xmlchange_cmnds to
             # the same file (caseroot/shell_commands)
+            case_shell_commands = case_shell_commands.replace("xmlchange_cmnds","shell_commands")
             # Note that use of xmlchange_cmnds has been deprecated and will soon
             # be removed altogether, so new tests should rely on shell_commands
-            case_shell_commands = case_shell_commands.replace("xmlchange_cmnds","shell_commands")
+            if shell_commands_file.endswith("xmlchange_cmnds"):
+                logger.warn("xmlchange_cmnds is deprecated and will be removed " +\
+                            "in a future release; please rename this file " +\
+                            "shell_commands")
             with open(shell_commands_file,"r") as fd:
                 new_shell_commands = fd.read().replace("xmlchange","xmlchange --force")
             with open(case_shell_commands, "a") as fd:
--- a/utils/python/CIME/user_mod_support.py
+++ b/utils/python/CIME/user_mod_support.py
@@ -56,6 +56,9 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
                             expect(False, "Could not write file %s in caseroot %s"
                                    %(case_source_mods,caseroot))
 
+    # Reverse include_dirs to make sure xmlchange commands are called in the
+    # correct order; it may be desireable to reverse include_dirs above the
+    # previous loop and then append user_nl changes rather than prepend them.
     include_dirs.reverse()
     for include_dir in include_dirs:
         # create xmlchange_cmnds and shell_commands in caseroot
--- a/utils/python/CIME/user_mod_support.py
+++ b/utils/python/CIME/user_mod_support.py
@@ -73,8 +73,8 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
             # be removed altogether, so new tests should rely on shell_commands
             if shell_commands_file.endswith("xmlchange_cmnds"):
                 logger.warn("xmlchange_cmnds is deprecated and will be removed " +\
-                            "in a future release; please rename this file " +\
-                            "shell_commands")
+                            "in a future release; please rename %s shell_commands" %\
+                            shell_commands_file)
             with open(shell_commands_file,"r") as fd:
                 new_shell_commands = fd.read().replace("xmlchange","xmlchange --force")
             with open(case_shell_commands, "a") as fd:


utils/python/CIME/user_mod_support.py
```diff
@@ -56,6 +56,8 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
                             expect(False, "Could not write file %s in caseroot %s"
                                    %(case_source_mods,caseroot))
 
+    include_dirs.reverse()
```
jedwards4b: Did you test this?   I think that without an assignment in line 59 it doesn't do anything.   You could also write line 60 as :
for include_dir in include_dirs.reverse()

I prefer this change to the one in 757

mnlevy1981: I did test it, that's the proper use for `.reverse()`; to do it inline it would be `for include_dir in include_dirs[::-1]:`

which I didn't like as much.

jedwards4b: Okay - let's put this one in then.

billsacks:
I find it confusing that part of this function now traverses the include_dirs in one order, and then the second part traverses it in the reverse order. Would it be better if the include_dirs list were reversed at the beginning? I guess that would mean that contents would have to be appended rather than pre-pended to the user_nl files.

You don't need to solve that problem as part of this PR, but can you at least:

1. Add a comment saying why you need to do the 'reverse'

2. Optionally, if you agree with my above comment: Add a TODO note suggesting that we reverse the include_dirs at the beginning and then append rather than pre-pend contents in the user_nl files.

I'm thinking that will help the readers of this code (like myself, more than a week from now).

billsacks:
Actually, reading through this code more carefully, I think I see another related issue - this is a pre-existing issue, not a new one in this PR: It looks like the cime5 code potentially builds up two files in the case directory: xmlchange_cmnds and shell_commands. It then executes each of them in order. This means that, for example: if you have a testmod 'mymod' which includes 'default', where 'mymod' has shell_commands but default has xmlchange_cmnds, it looks like the xmlchange_cmnds from default will still override whatever was done in mymod's shell_commands.

The implementation in cime4 was to append all commands into a single, master file in the case directory (named shell_commands). I believe that's what should be done to fix this robustly. If you don't want to add on to this PR, then we should open a new issue to that effect.

jedwards4b:
I think that you are wrong Bill, both the xmlchange_cmnds and shell_commands from one directory are handled prior to either from another directory.

billsacks:
I'm basically fine with this PR, but see my two comments. The most essential piece before this should come to master is a comment describing the need for the reverse here.

mnlevy1981:
@jedwards4b Both files are handled in one directory, but @billsacks points out that the way they are handled is that they are copied into separate files in `$CASEROOT` which are then processed in series. In a diagram:

```
testA
  |
testB
```

testA contains
```
shell_commands
---
xmlchange VAR1=VAL1

xmlchange_cmnds
---
xmlchange VAR2=VAL2
```

testB contains
```
shell_commands
---
xmlchange VAR2=VAL3

xmlchange_cmnds
---
xmlchange VAR1=VAL4
```

The current implementation will process `testB` first, so we get

```
shell_commands
---
xmlchange VAR2=VAL3

xmlchange_cmnds
---
xmlchange VAR1=VAL4
```

Then process `testA`:

```
shell_commands
---
xmlchange VAR2=VAL3
xmlchange VAR1=VAL1

xmlchange_cmnds
---
xmlchange VAR1=VAL4
xmlchange VAR2=VAL2
```

Then run `shell_commands` followed by `xmlchange_cmnds` => we end up with

```
VAR1 = VAL4
VAR2 = VAL2
```

When we want the results of `testA`, namely

```
VAR1 = VAL1
VAR2 = VAL2
```

billsacks:
> I think that you are wrong Bill, both the xmlchange_cmnds and shell_commands from one directory are handled prior to either from another directory.

No. I just tested this and my hypothesis is right: With Mike's branch, I set up this:

```
[yslogin6:/glade/p/work/sacks/cime/driver_cpl/cime_config/testdefs/testmods_dirs/drv]$ tail default/* bill/*
==> default/xmlchange_cmnds <==
./xmlchange RUN_STARTDATE=0002-01-01

==> bill/include_user_mods <==
../default


==> bill/shell_commands <==
./xmlchange RUN_STARTDATE=0003-01-01
```

and ran `SMS.f45_g37.A.yellowstone_intel.drv-bill`

We should get RUN_STARTDATE of 0003, but actually get 0002. It works right if the 'bill' testmod has xmlchange_cmnds rather than shell_commands.

mnlevy1981:
I think the proper fix for that is to not allow `xmlchange` calls in `shell_commands`: reserve that file for things like copies that are needed for prestaging data. (This would require updating POP's tests and probably the tests in other components as well because I know POP defaults to only using `shell_commands`)

billsacks:
> I think the proper fix for that is to not allow xmlchange calls in shell_commands: reserve that file for things like copies that are needed for prestaging data. (This would require updating POP's tests and probably the tests in other components as well because I know POP defaults to only using shell_commands)

That would depart from the earlier decision. When I reworked the handling of testmods a while ago (pre-cime5), the decision was to just use the file 'shell_commands' moving forward, and that this would include everything - xmlchange commands and others. However, for backwards compatibility, we continued to support xmlchange_cmnds. I think the intent was to eventually remove xmlchange_cmnds and only support shell_commands. But the desire to remove this support has never been strong enough to make it worth breaking backwards compatibility in this respect, so we have continued to support these two files that do exactly the same thing.

jedwards4b:
Why do we need two files?   My vote is to phase out one of them as quickly as possible and just use the other.

mnlevy1981:
A simpler fix would be to concatenate `shell_commands` and `xmlchange_cmnds` into a single file in caseroot, rather than keeping them split. In my example, this single file would contain

```
xmlchange VAR2=VAL3
xmlchange VAR1=VAL4
xmlchange VAR1=VAL1
xmlchange VAR2=VAL2
```

and the output would be as expected.

@billsacks I didn't realize this had been previously discussed, if `xmlchange_cmnds` has been deprecated then maybe as @jedwards4b says it's time to finally remove it.

billsacks:
> A simpler fix would be to concatenate shell_commands and xmlchange_cmnds into a single file in caseroot, rather than keeping them split. In my example, this single file would contain

Right, that was the cime4 behavior (sorry if I didn't explain it clearly)

> @billsacks I didn't realize this had been previously discussed, if xmlchange_cmnds has been deprecated then maybe as @jedwards4b says it's time to finally remove it.

It was probably a small, informal discussion (I remember Mariana being there, I don't know who else was), which we probably failed to communicate widely.

I'm fine with removing one. I'd vote for keeping shell_commands, but don't have strong feelings.

billsacks:
It looks like the only uses of xmlchange_cmnds in CESM are in cime, clm and cism, so I can remove those if we want to go with shell_commands. I don't know about ACME.

jedwards4b:
Go ahead and remove them, acme won't be on 5.2 for a while.

mnlevy1981:
I was about to push a commit that dumps commands in both `shell_commands` and `xmlchange_cmnds` into `$CASEROOT/shell_commands`, but it seems like maybe that won't be necessary?

jedwards4b:
Go ahead and push it, and add a message that xmlchange_cmnds is depreciated and will be removed.

mnlevy1981:
I think this is ready to go if one of you two wants to do one last round of testing and then bring it in.

mnlevy1981:
@billsacks I believe this latest update should produce the expected result in the [example test](https://github.com/ESMCI/cime/pull/758#issuecomment-258177094) you provided 30 minutes ago

billsacks:
I just tested this with my example case, and it works correctly - thanks! So I'm happy with this PR if @jedwards4b is.

utils/python/CIME/user_mod_support.py
```diff
@@ -56,11 +56,21 @@ def apply_user_mods(caseroot, user_mods_path, ninst=None):
...
+    for include_dir in include_dirs:
         # create xmlchange_cmnds and shell_commands in caseroot
         shell_command_files = glob.glob(os.path.join(include_dir,"shell_commands")) +\
                               glob.glob(os.path.join(include_dir,"xmlchange_cmnds"))
         for shell_commands_file in shell_command_files:
             case_shell_commands = shell_commands_file.replace(include_dir, caseroot)
+            # add commands from both shell_commands and xmlchange_cmnds to
+            # the same file (caseroot/shell_commands)
+            # Note that use of xmlchange_cmnds has been deprecated and will soon
```
jedwards4b: if shell_commands_file.endswith("xmlchange_cmnds"):
    logger.warn("xmlchange_cmnds is depreciated and will be removed in a future release, please rename this file shell_commands")

mnlevy1981:
Message is

```
if shell_commands_file.endswith("xmlchange_cmnds"):
logger.warn("xmlchange_cmnds is depreciated and will be removed in a future release, please rename this file shell_commands")
```

mnlevy1981:
Message is

```
Errput: xmlchange_cmnds is deprecated and will be removed in a future release; please rename /glade/p/work/mlevy/codes/CESM/cesm2_0_alpha03a+/components/pop/cime_config/testdefs/testmods_dirs/pop/default/xmlchange_cmnds shell_commands
```

PR #775, opened 2016-11-07
Cori port2a

jedwards4b:
Port to corip1, make it easier to understand what fields are available for batch and mpirun commands.  Fix issue with erio test. 

Test suite:  scripts_regression_tests
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Code review:


jedwards4b:
Because of the nersc policy: http://www.nersc.gov/users/computational-systems/cori/running-jobs/queues-and-policies/    only one job is allowed to run in debug at a time and only 5 are allowed to be in the queue.   I was seeing a lot of failures because scripts_regression_tests submits more than 5 jobs at a time.

Issue #776, opened 2016-11-07
st_archive does not delete the restart files correctly in the rundir


bertinia:
If DOUT_S_SAVE_INTERIM_RESTART_FILES is set to FALSE, then the st_archive 
should delete the interim restart files from the rundir. There is a comment in the
case_st_archive.py script that says to do this but the remove file command isn't 
actually in place.

lines 269 - 288:
`                    # Only archive intermediate restarts if requested - otherwise remove them                                               
                    if case.get_value('DOUT_S_SAVE_INTERIM_RESTART_FILES'):

....

`                    else:
                        srcfile = os.path.join(rundir, restfile)
                        logger.info("removing interim restart file %s" %srcfile)

This line should be added in the else clause after the srcfile = ...

`os.remove(srcfile)`

PR #777, opened 2016-11-07
Fix case_st_archive.py to delete restart files correctly

bertinia:
If DOUT_S_SAVE_INTERIM_RESTART_FILES=FALSE, then the
st_archive should automatically delete any restart files
in the rundir whose filename date strings don't match the current
cpl log filename date string associated with the end-of-run.

Test suite: 
1. A compset, f19_f16 on hobart
2. 5 day run STOP_N=5, REST_N=1 
3. DOUT_S_SAVE_INTERIM_RESTART_FILES=FALSE 
4. run case.st_archive and all interim restarts are deleted as expected
5. CONTINUE_RUN=TRUE 
6. DOUT_S_SAVE_INTERIM_RESTART_FILES=TRUE
7. rerun case.st_archive and all interim restarts are saved for days 6-11 as expected

Fixes [Github issue #]: #776

User interface changes: None

Input data changes: None

Code review:


utils/python/CIME/case_st_archive.py
```diff
@@ -285,7 +285,12 @@ def _archive_restarts(case, archive, archive_entry,
                             logger.debug("copying %s to %s" %(srcfile, destfile))
                     else:
                         srcfile = os.path.join(rundir, restfile)
-                        logger.info("removing interim restart file %s" %srcfile)
+                        if (os.path.isfile(srcfile)):
+                            try:
+                                os.remove(srcfile)
+                                logger.info("removing interim restart file %s" %srcfile)
+                            except OSError as e:
+                                logger.warn("interim restart file %s does not exist " %srcfile)
```
jedwards4b: Shouldn't this warning be that the file could not be removed?  It won't even be in this code block if the file didn;t exist (per the if statement at 288)

utils/python/CIME/case_st_archive.py
```diff
@@ -286,6 +286,10 @@ def _archive_restarts(case, archive, archive_entry,
                     else:
                         srcfile = os.path.join(rundir, restfile)
                         logger.info("removing interim restart file %s" %srcfile)
+                        try:
+                            os.remove(srcfile)
+                        except OSError as e:
+                            logger.warn("interim restart file %s does not exist " %srcfile)
```
bertinia: No, because I removed the "if (os.path.isfile(srcfile)):" outside of the try block because it will 
throw an exception if the file doesn't exist and so the "if" check was redundant.

I added the images into this branch so I can test the wiki doc links. Do you want a separate branch for these?

PR #778, opened 2016-11-07
fix erio test,and a minor format change for cray compiler

jedwards4b:
Fix issues with erio test.  Minor formating change for cray compiler

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Code review:

### DIFF
--- a/share/csm_share/shr/shr_strdata_mod.F90
+++ b/share/csm_share/shr/shr_strdata_mod.F90
@@ -1184,7 +1184,7 @@ subroutine shr_strdata_readnml(SDAT,file,rc,mpicom)
          read (nUnit, nml=shr_strdata_nml, iostat=rCode)
          if (rCode /= 0) then
             write(logunit,F01) 'ERROR: reading input namelist shr_strdata_input from file, &
-                 '//trim(file)//' iostat=',rCode
+                 &'//trim(file)//' iostat=',rCode
             call shr_sys_abort(subName//": namelist read error "//trim(file))
          end if
       end if
--- a/utils/python/CIME/SystemTests/erio.py
+++ b/utils/python/CIME/SystemTests/erio.py
@@ -21,47 +21,15 @@ class ERIO(SystemTestsCommon):
         self._pio_types = self._case.get_env("run").get_valid_values("PIO_TYPENAME")
         self._stop_n = self._case.get_value("STOP_N")
 
-    def build_phase(self, sharedlib_only=False, model_only=False):
-        """
-        Build one executable per PIO method
-        """
-        exeroot = self._case.get_value("EXEROOT")
-        cime_model = self._case.get_value("MODEL")
-
-        for pio_type in self._pio_types:
-            if pio_type == "default":
-                continue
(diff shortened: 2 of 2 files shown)


share/csm_share/shr/shr_strdata_mod.F90
```diff
@@ -1184,7 +1184,7 @@ subroutine shr_strdata_readnml(SDAT,file,rc,mpicom)
          read (nUnit, nml=shr_strdata_nml, iostat=rCode)
          if (rCode /= 0) then
             write(logunit,F01) 'ERROR: reading input namelist shr_strdata_input from file, &
-                 '//trim(file)//' iostat=',rCode
+                 &'//trim(file)//' iostat=',rCode
```
jedwards4b: unrelated to the erio fix - the cray compiler didn't like this line.

utils/python/CIME/SystemTests/erio.py
```diff
@@ -21,47 +21,15 @@ def __init__(self, case):
         self._pio_types = self._case.get_env("run").get_valid_values("PIO_TYPENAME")
         self._stop_n = self._case.get_value("STOP_N")
 
-    def build_phase(self, sharedlib_only=False, model_only=False):
```
jedwards4b: PIO options are all runtime, there is no need for this build_phase

utils/python/CIME/SystemTests/erio.py
```diff
@@ -21,47 +21,15 @@ def __init__(self, case):
...
-        pio_exe_path  = "%s/%s.exe.%s" % (exeroot, cime_model, pio_type)
-
-        if os.path.exists(base_exe_path):
-            os.remove(base_exe_path)
-
-        shutil.copy(pio_exe_path, base_exe_path)
-
         stop_option = self._case.get_value("STOP_OPTION")
-        expect(stop_n > 0, "Bad STOP_N: %d" % stop_n)
+        expect(self._stop_n > 0, "Bad STOP_N: %d" % self._stop_n)
```
jedwards4b: Need to get and keep the original stop_n value, otherwise the stop_n value gets smaller with each repeat of the test.

utils/python/CIME/SystemTests/erio.py
```diff
@@ -109,7 +68,9 @@ def run_phase(self):
 
         for idx, pio_type1 in enumerate(self._pio_types):
             if pio_type1 != "default":
+                self._case.set_value("PIO_TYPENAME", pio_type1)
```
jedwards4b: Here is where you set the various pio_types.   One concern I have - the netcdf4c format is a lossy compressed format, I would expect that output not to compare to the others with cprnc - It seems to pass for X and A cases though - I am trying some tests with external components.

utils/python/CIME/SystemTests/erio.py
jgfouca: Thanks, I'm not sure how this ever passed on melvin.

PR #779, opened 2016-11-07
Adds support for LANL IC machines

vanroekel:
Test suite: cime_developer, All PASS on Grizzly and Wolf
Test baseline: none
Test namelist changes: none 
Test status: [bit for bit]

User interface changes?: None

Code review: @jonbob 

This adds support for Grizzly (new linux cluster), updates support
for wolf, and removes support for mustang.

There was a non machine file change as moab was not listed as a valid batch system.


vanroekel:
I'm not sure if this is where to add this, but while working on these changes, I had difficulty with the PIO_REARRANGER variable.  Looking at the code, it seems this is dilineating between PIO v1 and v2.  When I would create a case, the script would fail with "1.7.2 is not a valid int for PIO_REARRANGER".  This value is set via PIO_VERSION, which for LANL is 1.7.2.  Hard-coding this value to 1 allowed all tests to pass.  

Am I understanding the PIO_REARRANGER variable correctly?  If so, should the PIO_VERSION system variable be set to 1 for LANL machines?

jedwards4b:
PIO_REARRANGER has two values 1 and 2 and only 1 is supported by PIO_VERSION 1 
PIO_VERSION also has only two values also 1 and 2.   It is not the repository version number of the code.

vanroekel:
Thank you @jedwards4b.  I will make the change in the LANL PIO module file.

rljacob:
Luke, note that the head of maint-cime5.1 is what will be merged to ACME shortly.  So you may want to redo these changes for that branch.

vanroekel:
Rob and Jon,  I think I have cherry-picked the appropriate changes onto maint-cime5.1 and changed the target branch for the merge.  Let me know if there are any other changes you would like

rljacob:
Looks good.

PR #803, opened 2016-11-11
Atmnlev fix

### LINKED ISSUE #786: Bug when atmnlev NOT NONE
When running test ERP_D_Ln9.T42z30_T42.FDABIP04.hobart_nag.cam-outfrq9s, getting the following error message.

  File "/home/fischer/code/cesm2_0_alpha03c/cime/scripts/Tools/../../utils/python/CIME/XML/grids.py", line 150, in _get_gridmaps
    grids[0] += "z"+atmnlev
TypeError: can only concatenate tuple (not "str") to tuple

It looks like the only problem is a missing comma at the end of the line since it's appending to a tuple.  Also line 152 has the same problem.

           grids[1] += "z"+lndnlev

I'm not sure if this is the fix, since I'm not familiar with this section of code.

This is a bug that @cacraigucar passed back to me.

fischer-ncar:
Fix bug where T42z30_T42 grid was causing an error when concatenating tuples.

Test suite:   scripts_regression_tests.py
Test baseline: 
Test namelist changes: 
Test status: 

Fixes #786 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/XML/grids.py
+++ b/utils/python/CIME/XML/grids.py
@@ -147,9 +147,9 @@ class Grids(GenericXML):
                             logger.debug(" %s: %s" %gridmap)
 
         if atmnlev is not None:
-            grids[0] += "z"+atmnlev
+            grids[0] += "z"+atmnlev,
         if lndnlev is not None:
-            grids[1] += "z"+lndnlev
+            grids[1] += "z"+lndnlev,
 
         return gridmaps
 


jgfouca:
We probably don't want to see upstream merges in your PR. Consider rebasing instead of merging when updating your branch.

utils/python/CIME/XML/grids.py
```diff
@@ -147,9 +147,9 @@ def _get_gridmaps(self, component_grids, atmnlev=None, lndnlev=None):
                             logger.debug(" %s: %s" %gridmap)
 
         if atmnlev is not None:
-            grids[0] += "z"+atmnlev
+            grids[0] += "z"+atmnlev,
```
jgfouca: I think this would be more clear:

    grids[0][0] += "z"+atmnlev

jgfouca: So if these tuples need to be mutable, then they should be lists, not tuples.

fischer-ncar:
I'm closing this pull request since this fix no longer works with new the config_grids.  I'll need to revisit this with an updated fix.

PR #805, opened 2016-11-12
Allow modification of vmag with gustiness

rljacob:
Use an input gust_fac and other functions to adjust vmag in the ocean/atmosphere surface flux
calculation to account for gustiness in storms.

gust_fac is input in drv_in and is default 0.0 which turns off the effect.

Originially implemented in ACME by Balwinder Singh.


rljacob:
cime-developer passed on anvil.

PR #813, opened 2016-11-16
Add tool to find circular dependencies

gold2718:
Pass a Depends file (created by mkDepends) to this script and it will print out any circular dependencies.

Test suite: NA (new tool)
Test baseline: NA
Test namelist changes: None
Test status: bit for bit

Fixes [CIME Github issue #] NA

User interface changes?: New tool, script takes a Depends file

Code review:

### DIFF
--- a/tools/findCircularDepend.py
+++ b/tools/findCircularDepend.py
@@ -1,0 +1,92 @@
+#!/usr/bin/python
+
+import sys
+import os
+import re
+import inspect
+
+## Important paths
+thisFile = os.path.realpath(inspect.getfile(inspect.currentframe()))
+currDir = os.path.dirname(thisFile)
+
+## Regular expression for object file line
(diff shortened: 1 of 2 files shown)


tools/findCircularDepend.py
```diff
@@ -0,0 +1,92 @@
+#!/usr/bin/python
```
jedwards4b: Hi Steve, 

Can you add the standard python script interface from the Tools directory.    Also should we have a new sub directory instead of putting it directly in tools?

Adding code similar to this:
```def main(arg1, arg2, etc):
    # do whatever the script does


if __name__ == "__main__":
    main(sys.argv[1], sys.argv[2], sys.argv[3])
```
will allow this script to be called as a subroutine from other scripts, we could then pass an option to case.build to run it from a case.

jedwards4b: okay

gold2718:
Okay, how's this? I tested it both standalone and imported and called.

PR #814, opened 2016-11-16
Config grids2

jedwards4b:
Reformat of the config_grids.xml file.  This change is backward compatible but the new version includes a schema check of the config_grids file. 

Test suite: scripts_regression_tests, several cesm specific tests 
Test baseline: cesm2_0_alpha04b
Test namelist changes: 
Test status: bit for bit with namelist format changes

Fixes 

User interface changes?: 

Code review:


utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -199,6 +199,10 @@ def run_indv(self, suffix="base", st_archive=False):
             infostr += ", with restarts every %d %s"%(rest_n, rest_option)
         logger.info(infostr)
 
+
```
jedwards4b: Adds a run phase in test status

utils/python/CIME/XML/component.py
```diff
@@ -16,14 +16,13 @@ def __init__(self, infile):
         initialize an object
         """
         files = Files()
-        schema = files.get_schema("CONFIG_DRV_FILE")
-        if schema is not None:
-            # not checking schema on external components yet
-            cimeroot = get_cime_root()
-            if  cimeroot in os.path.abspath(infile):
-                self.validate_xml_file(infile, schema)
+        schema = None
```
jedwards4b: schema check moved to generic_xml.py

utils/python/CIME/XML/entry_id.py
```diff
@@ -24,9 +24,7 @@ def get_default_value(self, node, attributes=None):
         value = self._get_value_match(node, attributes)
         if value is None:
             # Fall back to default value
-            val_node = self.get_optional_node("default_value", root=node)
```
jedwards4b: This is done all over the place so I made a generic get_element_text function.

utils/python/CIME/case_st_archive.py
```diff
@@ -249,8 +249,11 @@ def _archive_restarts(case, archive, archive_entry,
                 # need to do this before archiving restart files
                 histfiles_for_restart = get_histfiles_for_restarts(case, archive,
                                                                    archive_entry, restfile)
+
                 if datename_is_last and histfiles_for_restart:
-                    histfiles_savein_rundir = histfiles_for_restart
+                    for histfile in histfiles_for_restart:
```
jedwards4b: st_archiver was dropping files in multi-instance cases, this is a fix for that problem.

utils/python/tests/scripts_regression_tests.py
```diff
@@ -1115,11 +1115,15 @@ def simple_test(self, manual_timing=False):
     ###########################################################################
     def test_save_timings(self):
     ###########################################################################
+        if CIME.utils.get_model() != "acme":
```
jedwards4b: skip these tests for cesm

jedwards4b:
I must have stripped some tabs someplace where I shouldn't have, all of the make regression tests are failing.   I'll push an update with a fix.

jedwards4b:
Problem solved, ready to merge.

utils/python/tests/scripts_regression_tests.py
jgfouca: I think both of us need to be able to save timings. Was this test causing problems?

utils/python/CIME/test_status.py
```diff
@@ -31,10 +31,11 @@
 
 # The statuses that a phase can be in
 TEST_PEND_STATUS = "PEND"
+TEST_RUN_STATUS = "RUN"
 TEST_PASS_STATUS = "PASS"
 TEST_FAIL_STATUS = "FAIL"
 
-ALL_PHASE_STATUSES = [TEST_PEND_STATUS, TEST_PASS_STATUS, TEST_FAIL_STATUS]
+ALL_PHASE_STATUSES = [TEST_PEND_STATUS, TEST_RUN_STATUS, TEST_PASS_STATUS, TEST_FAIL_STATUS]
```
jgfouca: There's a lot of code that's highly dependent on the states in TestStatus. I'm pretty sure most of the code is not prepared to handle this additional state.

utils/python/CIME/SystemTests/system_tests_common.py
jgfouca: What's the idea behind TEST_RUN_STATUS? It looks like it's only used here.

jedwards4b: There was a request that test jobs waiting in the queue have a PEND status while test jobs currently running have a RUN status.   I think that this is the only place it's needed.

jedwards4b:
I'll review the issue with testsavetimings now

utils/python/CIME/SystemTests/system_tests_common.py
jgfouca: Yeah, I remember that request. I think @billsacks got most of what he really wanted with the signal handling changes that I put in last week. I'm not adamantly against this TEST_RUN_STATUS concept, but it will certainly require more work and probably deserves its own PR. Adding a new state will impact things like wait_for_tests and test_status.get_overall_test_status and probably others. I'm actually a little shocked and concerned that this change made it through testing without causing any fails.

billsacks: @jgfouca is right that much of my desire was satisfied with his signal handling changes. I'd still support differentiating between PEND and RUN if it's feasible.

jedwards4b:
I've removed the run_status and replaced and repaired the testsavetimings.

jgfouca:
Thanks.

PR #824, opened 2016-11-17
mfdeakin-sandia/systemtestscomparetwo/erp

### LINKED ISSUE #448: Convert ERP test to SystemTestsCompareTwo infrastructure?

The ERP test would be less straightforward to convert to this infrastructure, since currently it makes use of setting CONTINUE_RUN which doesn't work if we're using a clone-based test. On the other hand, the criticality and complexity of the ERP test argue for translating it to this new infrastructure, which will make it more maintainable and less error-prone.

@mvertens suggests changing its behavior so that, rather than doing a CONTINUE_RUN, it instead uses a branch run for the second case. For this to work, we'd need to introduce a new hook into the SystemTestsCompareTwo infrastructure: a _pre_run_two hook that is executed between the case1 run and the case2 run. For the ERP test, this would need to copy (or link) the relevant restart files from the case1 run directory to the case2 run directory.

mfdeakin-sandia:
Reimplements the ERP test with SystemTestsCompareTwo

Test baseline: 
Test namelist changes: 
Test status: ok

Fixes #448 

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/config_tests.xml
+++ b/cime_config/config_tests.xml
@@ -194,6 +194,7 @@ LII    CLM initial condition interpolation test
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>11</STOP_N>
     <DOUT_S>FALSE</DOUT_S>
+    <SMP_BUILD>0</SMP_BUILD>
   </test>
 
   <test NAME="ERS">
--- a/cime_config/config_tests.xml
+++ b/cime_config/config_tests.xml
@@ -197,7 +197,6 @@ LII    CLM initial condition interpolation test
     <HIST_N>$STOP_N</HIST_N>
     <CONTINUE_RUN>FALSE</CONTINUE_RUN>
     <DOUT_S>FALSE</DOUT_S>
-    <SMP_BUILD>0</SMP_BUILD>
   </test>
 
   <test NAME="ERS">
--- a/cime_config/config_tests.xml
+++ b/cime_config/config_tests.xml
@@ -193,6 +193,9 @@ LII    CLM initial condition interpolation test
     <BFBFLAG>TRUE</BFBFLAG>
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>11</STOP_N>
+    <HIST_OPTION>$STOP_OPTION</HIST_OPTION>
+    <HIST_N>$STOP_N</HIST_N>
+    <CONTINUE_RUN>FALSE</CONTINUE_RUN>
     <DOUT_S>FALSE</DOUT_S>
     <SMP_BUILD>0</SMP_BUILD>
   </test>
--- a/utils/python/CIME/SystemTests/erp.py
+++ b/utils/python/CIME/SystemTests/erp.py
@@ -15,6 +15,9 @@ from CIME.SystemTests.system_tests_compare_two import SystemTestsCompareTwo
 
 logger = logging.getLogger(__name__)
 
+def _compute_rest_n(stop_n):
+    return stop_n // 2 + 1
+
 class ERP(SystemTestsCompareTwo):
 
     def __init__(self, case):
@@ -33,7 +36,7 @@ class ERP(SystemTestsCompareTwo):
 
     def _case_one_setup(self):
         stop_n = self._case.get_value("STOP_N")
-        rest_n = stop_n // 2 + 1
+        rest_n = _compute_rest_n(stop_n)
         expect(stop_n > rest_n , "STOP_N value too small for test")
         self._case.set_value("REST_N", rest_n)
         self._case.set_value("REST_OPTION", self._case.get_value("STOP_OPTION"))
@@ -43,7 +46,7 @@ class ERP(SystemTestsCompareTwo):
         self._case.set_value("REST_OPTION", "never")
 
         stop_n_1 = self._case1.get_value("STOP_N")
-        stop_n = stop_n_1 - stop_n_1 // 2 - 1
+        stop_n = stop_n_1 - _compute_rest_n(stop_n_1)
         expect(stop_n > 0, "STOP_N value too small for test")
         self._case.set_value("STOP_N", stop_n)
 
@@ -87,3 +90,4 @@ class ERP(SystemTestsCompareTwo):
         date = date_tail[:date_tail.rfind('-')]
         logger.info("RUN_REFDATE found as %s" % date)
         self._case2.set_value("RUN_REFDATE", date)
+
--- a/utils/python/CIME/SystemTests/erp.py
+++ b/utils/python/CIME/SystemTests/erp.py
@@ -12,133 +12,36 @@ import shutil
 from CIME.XML.standard_module_setup import *
 from CIME.case_setup import case_setup
 import CIME.utils
-from CIME.SystemTests.system_tests_common import SystemTestsCommon
+from CIME.SystemTests.system_tests_compare_two import SystemTestsCompareTwo
 
 logger = logging.getLogger(__name__)
 
-class ERP(SystemTestsCommon):
+class ERP(SystemTestsCompareTwo):
 
     def __init__(self, case):
         """
         initialize a test object
         """
-        SystemTestsCommon.__init__(self, case)
+        SystemTestsCompareTwo.__init__(self, case, run_one_st_archive = True)
 
-    def build_phase(self, sharedlib_only=False, model_only=False):
-        """
-        Build two cases.  Case one uses defaults, case2 uses half the number of threads
-        and tasks. This test will fail for components (e.g. pop) that do not reproduce exactly
-        with different numbers of mpi tasks.
-        """
-        if sharedlib_only:
-            return self.build_indv(sharedlib_only=sharedlib_only, model_only=model_only)
+    def _case_one_setup(self):
+        pass
 
-        exeroot = self._case.get_value("EXEROOT")
-        cime_model = CIME.utils.get_model()
+    def _case_two_setup(self):
+        """ Case two uses half the number of threads and tasks as the defaults and case one. """
+        for comp in self._case.get_values("COMP_CLASSES"):
+            if comp == "DRV":
+                comp = "CPL"
+            ntasks = self._case.get_value("NTASKS_%s"%comp)
+            if ntasks > 1:
+                rootpe = self._case.get_value("ROOTPE_%s"%comp)
+                self._case.set_value("NTASKS_%s"%comp, ntasks // 2)
+                self._case.set_value("NTASKS_%s"%comp, rootpe // 2)
+            nthreads = self._case.get_value("NTHRDS_%s"%comp)
+            if nthreads > 1:
+                self._case.set_value("BUILD_THREADED", True)
+                self._case.set_value("NTHRDS_%s"%comp, nthreads // 2)
+        case_setup(self._case, test_mode=True, reset=True)
 
-        # Make backup copies of the ORIGINAL env_mach_pes.xml and
-        # env_build.xml in LockedFiles if they are not there. If there
-        # are already copies there then simply copy them back to
-        # have the starting env_mach_pes.xml and env_build.xml
-        machpes1 = os.path.join("LockedFiles","env_mach_pes.ERP1.xml")
-        envbuild1 = os.path.join("LockedFiles","env_build.ERP1.xml")
-        if ( os.path.isfile(machpes1) ):
-            shutil.copy(machpes1,"env_mach_pes.xml")
-        else:
-            shutil.copy("env_mach_pes.xml","env_mach_pes.ERP1.xml")
-
-        if ( os.path.isfile(envbuild1) ):
-            shutil.copy(envbuild1,"env_build.xml")
-
-        # Build two executables, one using the original tasks and threads (ERP1) and
-        # one using the modified tasks and threads (ERP2)
-        # The reason we currently need two executables that CESM-CICE has a compile time decomposition
-        # For cases where ERP works, changing this decomposition will not affect answers, but it will
-        # affect the executable that is used
-        self._case.set_value("SMP_BUILD","0")
-        for bld in range(1,3):
-            logging.warn("Starting bld %s"%bld)
-
-            if (bld == 2):
-                # halve the number of tasks and threads
-                for comp in self._case.get_values("COMP_CLASSES"):
-                    if comp == "DRV":
-                        comp = "CPL"
-                    ntasks    = self._case.get_value("NTASKS_%s"%comp)
-                    nthreads  = self._case.get_value("NTHRDS_%s"%comp)
-                    rootpe    = self._case.get_value("ROOTPE_%s"%comp)
-                    if ( nthreads > 1 ):
-                        self._case.set_value("BUILD_THREADED", True)
-                        self._case.set_value("NTHRDS_%s"%comp, nthreads/2)
-                    if ( ntasks > 1 ):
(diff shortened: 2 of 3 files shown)


utils/python/CIME/SystemTests/erp.py
```diff
@@ -12,133 +12,50 @@
...
+
+    def _case_two_setup(self):
+        """ Case two uses half the number of threads and tasks as the defaults and case one. """
+        for comp in self._case.get_values("COMP_CLASSES"):
+            if comp == "DRV":
+                comp = "CPL"
+            ntasks = self._case.get_value("NTASKS_%s"%comp)
+            if ntasks > 1:
+                rootpe = self._case.get_value("ROOTPE_%s"%comp)
+                self._case.set_value("NTASKS_%s"%comp, ntasks // 2)
```
jedwards4b: I think that this may be a problem if the original value of ntasks is odd?

utils/python/CIME/SystemTests/erp.py
```diff
@@ -12,133 +12,50 @@
...
+                self._case.set_value("ROOTPE_%s"%comp, rootpe // 2)
+            nthreads = self._case.get_value("NTHRDS_%s"%comp)
+            if nthreads > 1:
+                self._case.set_value("BUILD_THREADED", True)
+                self._case.set_value("NTHRDS_%s"%comp, nthreads // 2)
+
+        rest_n = self._case.get_value("STOP_N") // 2 + 1
+        self._case.set_value("REST_N", rest_n)
```
jedwards4b: You might want to add an expect(rest_n > 0 , "STOP_N value too small for test")

utils/python/CIME/SystemTests/erp.py
mfdeakin-sandia: This is what was done in the original. Would you prefer it to use nearest integer rounding?

utils/python/CIME/SystemTests/erp.py
mfdeakin-sandia: Is STOP_N ever negative? That condition will never trigger otherwise.
Though it might be a good sanity check on future changes

utils/python/CIME/SystemTests/erp.py
jedwards4b: oh nevermind - I think that because its // instead of / it works correctly.

utils/python/CIME/SystemTests/erp.py
jedwards4b: yes it's just  a sanity check - that's what expect statements do, you never expect to trigger them.

billsacks:
Please don't bring this to master yet: I'm in the midst of reviewing it, but won't be able to finish until later this morning.

utils/python/CIME/SystemTests/erp.py
```diff
@@ -12,133 +12,51 @@
...
+                rootpe = self._case.get_value("ROOTPE_%s"%comp)
+                self._case.set_value("NTASKS_%s"%comp, ntasks // 2)
+                self._case.set_value("ROOTPE_%s"%comp, rootpe // 2)
+            nthreads = self._case.get_value("NTHRDS_%s"%comp)
+            if nthreads > 1:
+                self._case.set_value("BUILD_THREADED", True)
+                self._case.set_value("NTHRDS_%s"%comp, nthreads // 2)
+
+        rest_n = self._case.get_value("STOP_N") // 2 + 1
+        expect(rest_n > 0 , "STOP_N value too small for test")
```
billsacks: Is this really the right assertion? I can't see how rest_n will ever be <= 0. I think the right assertion is something like: stop_n of the second case should be > 0. But the logic for rest_n and stop_n is confused for this rewritten erp test so it's hard for me to say (see my overall review comment).

cime_config/config_tests.xml
```diff
@@ -193,7 +193,11 @@ LII    CLM initial condition interpolation test
     <BFBFLAG>TRUE</BFBFLAG>
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>11</STOP_N>
+    <HIST_OPTION>$STOP_OPTION</HIST_OPTION>
+    <HIST_N>$STOP_N</HIST_N>
+    <CONTINUE_RUN>FALSE</CONTINUE_RUN>
     <DOUT_S>FALSE</DOUT_S>
+    <SMP_BUILD>0</SMP_BUILD>
```
jedwards4b: Actually if the initial case has threads=2 then case 2 needs to be built with OMP enabled or it will fail.

billsacks: I just talked with Jim about this, and we agree that you **can** remove SMP_BUILD here, because the necessary logic is in the setting of BUILD_THREADED in erp.py.

utils/python/CIME/SystemTests/erp.py
mfdeakin-sandia: I also expected a test for stop_n > 0 at a somewhat more appropriate location.
Perhaps a more appropriate test would be case1.rest_n < case1.stop_n? This would still check for negatives like Jim E wanted, and also prevent the mistake I made with what the rest point should be.

utils/python/CIME/SystemTests/erp.py
```diff
@@ -12,133 +12,50 @@
...
+                self._case.set_value("BUILD_THREADED", True)
+                self._case.set_value("NTHRDS_%s"%comp, nthreads // 2)
+
+        rest_n = self._case.get_value("STOP_N") // 2 + 1
+        self._case.set_value("REST_N", rest_n)
+        self._case.set_value("REST_OPTION", self._case.get_value("STOP_OPTION"))
+
+        case_setup(self._case, test_mode=True, reset=True)
+
+    def _pre_run_one_hook(self):
```
billsacks: My point here is that these aren't symmetrical: There are legitimate cases (like this one) where you need to do something in between run 1 and run 2. However, I think that anything you need to do before run1 can safely be done earlier in the process, in one of the existing hooks. Thus, adding that extra hook just adds an unnecessary decision for test developers.

utils/python/CIME/SystemTests/erp.py
billsacks: Yes, that would be a reasonable assertion.

utils/python/CIME/SystemTests/erp.py
mfdeakin-sandia: In that case I'd like to change the hook name to something that makes it clear it's run after case 1 was run and before case 2. Something like but less verbose than _inter_baseline_test_run_hook.

billsacks: Good point - I agree. Maybe `_between_run_setup`? I don't love that name, so am open to alternatives.

billsacks: > would it make sense to rename case1 and case2 to something like baseline and test? I believe they were supposed to be interchangeable; but with this test they can't be.

We're not totally consistent about making case1 the 'baseline' - e.g., the PET test turns on threading for case1 and turns it off for case2 - so I'd argue that case2 is more of the 'baseline' here. I feel that the current number-based names illustrate exactly the asymmetry that exists - namely, that case1 is run first, and then case2 is run second. So for tests like this new ERP test that require a certain run to happen first, it seems more clear that that run must be done as case1. I feel that changing the names to 'baseline' and 'test' would actually make this asymmetry _less_ clear. So I'd prefer keeping the names as they are.

In which case, `_inter_baseline_test_run` doesn't make a lot of sense, because we aren't calling the runs 'baseline' and 'test'. I also find this name confusing, because 'baseline' is overloaded here - I think of the BASELINE phase of the test when I see it.

As I think about it more, I still personally find something like `_pre_run_one` or `post_run_two` acceptable and clear.

mfdeakin-sandia: Alright, if the numbers are explicitly referring to the order in which they are setup, built, and run, I can live with the names of the case variables.
I still prefer naming the hook with something that doesn't imply a symmetric hook exists. Such as _inter_run_hook? Sorry if I'm being too picky based on ideas I haven't fully understood

billsacks: I'm happy with _inter_run_hook.

mfdeakin-sandia:
@billsacks Completing the 3 changes you mentioned doesn't seem to result in a working test, and it's not immediately clear to me what other changes I need to have a branching run.
In it's current state on this branch, the test fails to pass, which suggests that either it's not picking up the restart files (my current suspicion), or I'm calculating or setting one of the case values incorrectly.

utils/python/CIME/SystemTests/erp.py
```diff
@@ -11,134 +11,82 @@
...
+        rest_n = _compute_rest_n(stop_n)
+        expect(stop_n > rest_n , "STOP_N value too small for test")
+        self._case.set_value("REST_N", rest_n)
+        self._case.set_value("REST_OPTION", self._case.get_value("STOP_OPTION"))
+
+    def _case_two_setup(self):
+        """ Case two uses half the number of threads and tasks as the defaults and case one. """
+        self._case.set_value("REST_OPTION", "never")
+
+        stop_n_1 = self._case1.get_value("STOP_N")
```
billsacks: It shouldn't hurt to get this from _case1, but it isn't necessary: you can get the old STOP_N value from the current case. I've been trying to write the other tests that way - i.e., only referring to the current self._case unless that's truly impossible. The SystemTestsCompareTwo infrastructure makes that possible by ensuring that case setup only happens once for each case - so (for example) you don't run the risk of re-halving the STOP_N value.

utils/python/CIME/SystemTests/erp.py
```diff
@@ -11,134 +11,82 @@
...
+        stop_n = stop_n_1 - _compute_rest_n(stop_n_1)
+        expect(stop_n > 0, "STOP_N value too small for test")
+        self._case.set_value("STOP_N", stop_n)
+
+        self._case.set_value("CONTINUE_RUN", True)
```
billsacks: CONTINUE_RUN should _not_ be set to True here.

utils/python/CIME/SystemTests/erp.py
```diff
@@ -11,134 +11,82 @@
...
+        restdir = os.path.join(st_archive_dir_one, restdir_name)
+        rundir = self._case2.get_value("RUNDIR")
+        path_1 = None
+        for root, _, files in os.walk(restdir):
+            for f in files:
+                if path_1 == None:
+                    path_1 = root
+                fpath_in = os.path.join(root, f)
+                fpath_out = os.path.join(rundir, f)
+                os.symlink(fpath_in, fpath_out)
```
billsacks: os.symlink will fail if fpath_out already exists - e.g., if rerunning the test

utils/python/CIME/SystemTests/erp.py
```diff
@@ -11,134 +11,82 @@
...
+
+    def _inter_run_hook(self):
+        st_archive_dir_one = self._case1.get_value("DOUT_S_ROOT")
+        restdir_name = "rest"
+        restdir = os.path.join(st_archive_dir_one, restdir_name)
+        rundir = self._case2.get_value("RUNDIR")
+        path_1 = None
+        for root, _, files in os.walk(restdir):
+            for f in files:
+                if path_1 == None:
```
billsacks: Check for None with `is None` (see https://www.python.org/dev/peps/pep-0008/#programming-recommendations)

More importantly: Can you describe what you're doing with the logic here for setting path_1? This is unclear to me.

utils/python/CIME/SystemTests/erp.py
```diff
@@ -11,134 +11,82 @@
...
+                    path_1 = root
+                fpath_in = os.path.join(root, f)
+                fpath_out = os.path.join(rundir, f)
+                os.symlink(fpath_in, fpath_out)
+
+        ref_case = self._case1.get_value("CASE") + ".ref1"
```
billsacks: You do not want the '.ref1' here - that's specific to the eri test

utils/python/CIME/SystemTests/erp.py
```diff
@@ -11,134 +11,82 @@
...
+        self._case2.set_value("REF_CASE", ref_case)
```
billsacks: This should be RUN_REFCASE, not REF_CASE.

In addition, you need to set RUN_TYPE to branch.

I'd prefer if the setting of RUN_REFCASE and RUN_TYPE were done in _case_two_setup, because I prefer if as much as possible is done there rather than in this new _inter_run_hook. It does seem like you need to wait until this hook to set RUN_REFDATE, though. So I don't feel strongly about this.

mfdeakin-sandia: Setting RUN_TYPE did solve my issue, thanks!

mfdeakin-sandia:
I haven't actually completed the testing for failures you asked for, and need to run scripts_regression_tests again, so please don't merge this yet

billsacks:
@mfdeakin-sandia : Yes, I was just in the midst of saying the following:

Your latest version of the code looks good to me. The testing I'd like to see is:

1. (From my previous review) I would like to see
evidence that the new test fails when it should fail before this PR is brought
to master. At the very least I'd like to see evidence that the exact restart
functionality of this test correctly picks up problems. See #295 for an example
of how to test this. This may need to be adapted for the ACME code base.

 i.e., First as a control run, run ERS and the new ERP on some unmodified code:
they should pass. Then make some code changes that break exact restart. Verify
that ERS fails (to make sure you have truly broken exact restart). Then verify
that this new ERP test fails. (Or you can use the old ERP test rather than the
ERS test for comparison purposes.)

2. If you haven't already done so, the one other thing that should probably be verified is that the task & thread counts are correct for both pieces of the test. I forget where you can see this - maybe in the cpl.log file?

mfdeakin-sandia:
@billsacks Would you have suggestions on how to break a data or stub component? I had hoped to do so by changing the input step number and datetime, but that doesn't seem to have worked.

billsacks:
No idea.

mfdeakin-sandia:
I looked at the logs, and the number of pes, threads, and the pio_strides match the numbers in the original ERP test case logs, so I believe that the task and thread counts are correct.
I've also finished running scripts_regression_tests with all passes.
I'm still not certain how to break restarts, as just deleting the code in the seq_rest_read subroutine did not break the ERP or ERS test cases with the data compset. I think using a different compset might fix this issue, though.

jgfouca:
@mfdeakin-sandia ready to merge?

mfdeakin-sandia:
Not quite yet; I've finally got CIME5 working on Edison, so now I can try and run a test with a non data or stub component so I can break the tests.
I think I'll need some help doing so, @jedwards4b was willing to last week, so hopefully it'll be ready by tomorrow.

mfdeakin-sandia:
It looks like I won't be able to break the reset until ACME catches up to using this version of CIME, unfortunately; so either we wait to merge this, or we remember to test with broken resets when we can

jedwards4b:
I'll try to do this on a cesm machine...

jedwards4b:
It turns out that changing this test to conform to the systemtestscomparetwo framework ran into some problems that we didn't consider.  Namely the mechanism for handling data accumulation in history files is different in a branch run than in a restart run.  In a restart run the history accumulation continues after restart from where it left off in the original run while a branch run resets the accumulation counter to 0. 

In order to get exact restart conditions in a branch run the history frequency must be set such that hist_n=rest_n but that turns out to be a difficult thing for the test to force since the components are allowed to modify their hist_n through the testmods mechanism (eg cams outfrq9s).   So we have decided to leave the ERP test as is and close this issue and PR without change.

PR #828, opened 2016-11-22
Config build

jedwards4b:
Several different issues are contained in this PR:

This PR brings the file currently called config_build.xml into config_compilers.xml.
There is no longer a config_build.xml file.   The new build tools as written were not compatible with the basic structure of cime, this PR tries to reconcile the discontinuities.   In both acme/machines and cesm/machines:
- removed config_compilers.xml
- renamed config_build.xml to config_compilers.xml

A new CIME directory has been created, CIME/BuildTools that contain build utilities. 
- The file XML/build.py has been removed and the classes contained in build.py have been split into separate new files in CIME/BuildTools.
- A new file, configure.py, has been added that contains the functionality in tools/configure (tools/configure now calls BuildTools/configure)
`  
macro_maker = Compilers(machobj)
`
      before in tools/configure
`    
macro_maker = Build(machobj)
`
As a result the refactoring of configure needs to accompany this PR

- both case.setup and tools/configure now use the same tools to create the Macros file
      now in  configure.py (called by tools/configure)

Will carry out extensive prebeta and prealpha testing on all machines for CESM before the PR is accepted

Test suite: scripts_regression_tests on yellowstone, cori and theta.   
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Code review:


jedwards4b:
I'm still testing and cleaning stuff up, but wanted to get some feedback.

scripts/Tools/code_checker
```diff
@@ -107,6 +107,8 @@ def check_code(dir_to_check, num_procs, files):
         files_to_check = [item for item in files_to_check if matches(item, files)]
     else:
         files_to_check = [item for item in files_to_check if item.endswith(".py")]
+    if len(files_to_check) == 0:
```
jedwards4b: If you pass in a file on the command line then it should be checked even if it's not in the repo.

tools/configure
```diff
@@ -117,59 +115,6 @@ def parse_command_line(args):
 
     return opts
 
-def configure(machobj, output_dir, macros_format, compiler, mpilib, debug, sysos):
```
jedwards4b: create_newcase and configure now both follow the same path to create the Macros and Depends files.

utils/python/CIME/XML/compilerblock.py
```diff
@@ -0,0 +1,232 @@
+"""
```
jedwards4b: This really should be a part of the compilers class to best fit with the rest of cime

utils/python/CIME/XML/compilers.py
```diff
@@ -5,27 +5,40 @@
 from CIME.XML.standard_module_setup import *
 from CIME.XML.generic_xml import GenericXML
 from CIME.XML.files import Files
+from CIME.XML.compilerblock import CompilerBlock
```
jedwards4b: Supports both the old config_compilers.xml and the new one which was renamed from config_build.xml

utils/python/CIME/case.py
```diff
@@ -811,42 +810,6 @@ def _create_caseroot_tools(self):
             except Exception as e:
                 logger.warning("FAILED to set up toolfiles: %s %s %s" % (str(e), toolfile, destfile))
 
-        # Create Macros file.
```
jedwards4b: done in case.setup

mvertens:
@jedwards4b I'd like to see a detailed test plan for this PR as part of the code review. Since this effects a wide variety of machines and compilers, it would be important to know that we have tested this thoroughly before bringing it into CIME.

driver_cpl/cime_config/config_component.xml
```diff
@@ -2974,14 +2974,6 @@
     <desc>Whether to clean the test after it is built/run</desc>
   </entry>
 
-  <entry id="CCSM_BASELINE">
```
mvertens: Why has CCSM_BASELINE been removed for config_component.xml in the driver. How does a user change this via changing an xml file? Or is the idea that this will only be set on the commandline for create_test?

utils/python/CIME/BuildTools/configure.py
```diff
@@ -0,0 +1,87 @@
...
+    macros_format - Container containing the string 'Makefile' to produce
+                    Makefile Macros output, and/or 'CMake' for CMake output.
+    compiler - String containing the compiler vendor to configure for.
+    mpilib - String containing the MPI implementation to configure for.
+    debug - Boolean specifying whether debugging options are enabled.
+    """
+    # Macros generation.
+    suffixes = {'Makefile': 'make', 'CMake': 'cmake'}
+    macro_maker = Compilers(machobj)
+    for form in macros_format:
```
mvertens: Why not just use macro_format directly rather than an for statement and simply check if macro_format is not None?

utils/python/CIME/BuildTools/configure.py
```diff
@@ -0,0 +1,87 @@
...
+        logger.warn("%s already exists, delete to replace"%ems_path)
+        return
+    ems_file = EnvMachSpecific(output_dir)
+    ems_file.populate(machobj)
+    ems_file.write()
+    for shell in ('sh', 'csh'):
+        ems_file.make_env_mach_specific_file(compiler, debug, mpilib, shell)
+        shell_path = os.path.join(output_dir, ".env_mach_specific." + shell)
+        with open(shell_path, 'a') as shell_file:
+            if shell == 'sh':
```
mvertens: In the following, why not create a single string to write all of the following - rather than have repeated calls to disk?

driver_cpl/cime_config/config_component.xml
jedwards4b: CCSM_BASELINE has been replaced in the code by BASELINE_ROOT - I'll finish this cleanup by replacing the rest of the cases of CCSM_BASELINE in cime.

utils/python/CIME/BuildTools/configure.py
jedwards4b: This is inherited code, allows you to create both cmake and make macros files in one call.

driver_cpl/cime_config/config_component.xml
mvertens: so will BASELINE_ROOT be added as a new xml variable in config_component.xml?
I am wondering if it was necessary to do this as part of this PR? could we have separated this into another PR?

mvertens:
There seems to be a lot of different things being bundled together into this PR.

cime_config/cesm/machines/config_machines.xml
```diff
@@ -218,18 +218,18 @@
     <mpirun mpilib="mvapich2">
       <executable>srun</executable>
       <arguments>
-        <arg name="mpi">--mpi=none</arg>
```
jgfouca: It looks like the rest of the file is using 2-space indenting. Why change that here?

scripts/Tools/code_checker
jgfouca: I like the sentiment here. I would change it so that if files contains paths (relative or absolute) they should be checked as-is.

jedwards4b: I'm not sure what you mean - can you elaborate?

jgfouca: Pattern: foo.py -> Goes through the matches check
Path: ./foo.py, /home/jgfouca/foo.py -> Goes directly into files_to_check

jedwards4b: okay, will change

jgfouca:
@jedwards4b is this ready to merge?

jedwards4b:
We are ready to merge this now.

cime_config/cesm/machines/config_machines.xml
jgfouca: Just wanted to repeat the comment that the indenting changes in this file are confusing to me.

scripts/Tools/code_checker
jgfouca: Looks like this still needs to be changed.

jgfouca:
@jedwards4b just a couple final questions in the above review

jedwards4b:
The code_checker issue has been fixed, I can only apologize for the white space change.

PR #886, opened 2016-12-05
Make bless_test_results more robust by providing full path to create_test run_cmd

### LINKED ISSUE #882: bless_test_results - can not find create_test command
I am using tag:
 https://github.com/CESM-Development/cime/tags/cime5.2.0-alpha.10

When I issue the command:
 /glade/u/home/cacraig/cam5_4_94/cime/scripts/Tools/bless_test_results  -t '' -c '' -r /glade/scratch/cacraig/aux_cam_20161123172411 -b cam5_4_94-copy6 -f

I get the following error:
FAILED TO BLESS TEST: SMS_Lm13.f09_f09.F2000_DEV.yellowstone_intel.cam-outfrq1m_clm5, reason Namelist regen failed: '/bin/sh: create_test: command not found'

The baseline directory created by this script has the netCDF files in it, but no namelists are copied.  I've tried running this command from various locations including the cime/scripts directory where create_test resides, and can not eliminate this error.

jgfouca:
compare_test_result output for a case with both NL and baseline diffs:
FAIL TESTRUNPASS_P1.f19_g16_rx1.A.sandia-srn-sems_gnu NLCOMP See /home/jgfouca/acme/scratch/TESTRUNPASS_P1.f19_g16_rx1.A.sandia-srn-sems_gnu.C.20161206_130940_yfq5ct/compare.log.foucar.20161206_131733
FAIL TESTRUNPASS_P1.f19_g16_rx1.A.sandia-srn-sems_gnu BASELINE

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #882

User interface changes?: No

Code review: jedwards

### DIFF
--- a/scripts/Tools/compare_test_results
+++ b/scripts/Tools/compare_test_results
@@ -66,6 +66,12 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
 
     CIME.utils.setup_standard_logging_options(parser)
 
+    parser.add_argument("-n", "--namelists-only", action="store_true",
+                        help="Only analyze namelists.")
+
+    parser.add_argument("--hist-only", action="store_true",
+                        help="Only analyze history files.")
+
     parser.add_argument("-b", "--baseline-name", default=default_baseline_name,
                         help="Name of baselines to use, corresponds to branch used.")
 
@@ -88,7 +94,7 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
 
     CIME.utils.handle_standard_logging_options(args)
 
-    return args.baseline_name, args.baseline_root, args.test_root, args.compiler, args.test_id, args.compare_tests
+    return args.baseline_name, args.baseline_root, args.test_root, args.compiler, args.test_id, args.compare_tests, args.namelists_only, args.hist_only
 
 ###############################################################################
 def _main_func(description):
@@ -97,10 +103,10 @@ def _main_func(description):
         test_results = doctest.testmod(verbose=True)
         sys.exit(1 if test_results.failed > 0 else 0)
 
-    baseline_name, baseline_root, test_root, compiler, test_id, compare_tests = \
+    baseline_name, baseline_root, test_root, compiler, test_id, compare_tests, namelists_only, hist_only = \
         parse_command_line(sys.argv, description)
 
-    success = compare_test_results(baseline_name, baseline_root, test_root, compiler, test_id, compare_tests)
+    success = compare_test_results(baseline_name, baseline_root, test_root, compiler, test_id, compare_tests, namelists_only, hist_only)
     sys.exit(0 if success else CIME.utils.TESTS_FAILED_ERR_CODE)
 
 ###############################################################################
--- a/utils/python/CIME/bless_test_results.py
+++ b/utils/python/CIME/bless_test_results.py
@@ -1,6 +1,6 @@
 import CIME.compare_namelists, CIME.simple_compare
 from CIME.test_scheduler import NAMELIST_PHASE
-from CIME.utils import run_cmd, expect
+from CIME.utils import run_cmd, expect, get_scripts_root
 from CIME.test_status import *
 from CIME.hist_utils import generate_baseline, compare_baseline
 from CIME.case import Case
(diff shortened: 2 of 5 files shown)


jgfouca:
@cacraigucar , once this is merged, namelist blessing should work.

jgfouca:
@jedwards4b hold up on merge.

jgfouca:
@jedwards4b OK, this is ready for review. @billsacks you may care about this too.

utils/python/CIME/bless_test_results.py
```diff
@@ -18,7 +18,7 @@ def bless_namelists(test_name, report_only, force, baseline_name, baseline_root)
     print "Test '%s' had a namelist diff" % test_name
     if (not report_only and
         (force or raw_input("Update namelists (y/n)? ").upper() in ["Y", "YES"])):
-        stat, _, err = run_cmd("create_test -n -g %s -b %s --baseline-root %s" % (test_name, baseline_name, baseline_root))
+        stat, _, err = run_cmd("%s/create_test -n -g %s -b %s --baseline-root %s -o" % (get_scripts_root(), test_name, baseline_name, baseline_root))
```
jedwards4b: Doesn't create_test always need to run from the scripts_root directory?   In that case this can simply be ./create_test

jgfouca: This is the bug fix.

scripts/Tools/compare_test_results
```diff
@@ -66,6 +66,12 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
 
     CIME.utils.setup_standard_logging_options(parser)
 
+    parser.add_argument("-n", "--namelists-only", action="store_true",
```
jgfouca: Add these in order to be more similar to it's counterpart bless_test_results.

utils/python/CIME/bless_test_results.py
jgfouca: Create_test can be run from anywhere and so can bless_test_results, so we have to use an absolute path.

utils/python/CIME/compare_test_results.py
```diff
@@ -60,39 +67,74 @@ def compare_test_results(baseline_name, baseline_root, test_root, compiler, test
...
+            if not hist_only:
+                brief_result += "%s %s %s %s\n" % (nl_compare_result, test_name, NAMELIST_PHASE, nl_compare_comment)
+
+            if not namelists_only:
+                brief_result += "%s %s %s" % (compare_result, test_name, BASELINE_PHASE)
+                if compare_comment:
+                    brief_result += " %s" % compare_comment
+                brief_result += "\n"
+
+            print brief_result,
```
billsacks: Why did you change this from print(...) to this, getting rid of the `from __future__ import print_function`? My understanding is that the former is preferred for the sake of python3-compatibility. Is there some disadvantage to the old way? (I ask largely so I can understand for the future.)

scripts/Tools/compare_test_results
```diff
@@ -66,6 +66,12 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
...
+                        help="Only analyze namelists.")
+
+    parser.add_argument("--hist-only", action="store_true",
```
billsacks: Would it make sense to also add a short option here, '-h', analogous to the '-n' option for --namelists-only?

utils/python/CIME/compare_test_results.py
```diff
@@ -60,39 +67,74 @@ def compare_test_results(baseline_name, baseline_root, test_root, compiler, test
...
-            detailed_comments = ""
-            if (run_result is None):
-                compare_result = "SKIP"
-                compare_comment = "Test did not make it to run phase"
-                do_compare = False
-            elif (run_result != TEST_PASS_STATUS):
-                compare_result = "SKIP"
-                compare_comment = "Run phase did not pass"
-                do_compare = False
+            if (not hist_only):
```
billsacks: No need to rework this code now, but a general comment for the sake of future code: I find this `if (not XXX_only)` logic (which appears in many places) to be confusing. It would be especially confusing and error-prone if there were more than two pieces. I'd find this sort of code easier to read, write and maintain if this were flipped around, so this line (for example) became `if (do_namelist_compare)`. Again, no need to rework this now, but I'd like us to consider that for the future.

scripts/Tools/compare_test_results
jgfouca: Yes, unfortunately -h always maps to --help.

billsacks: Ah right, of course. I'd still vote for a one-character equivalent like '-i' (the second letter of 'hist'), but don't feel strongly about that.

utils/python/CIME/compare_test_results.py
billsacks: Okay, that's fine... this probably warrants a broader discussion at some point: Personally, I feel that we should try to avoid writing new code that will obviously break under py3, in order to make our lives easier later... because presumably we will want this to work with py3 at some point in the near future(?). But I don't have the energy to push this point for now :-)

scripts/Tools/compare_test_results
jgfouca: Yeah, we can save that for another day.

utils/python/CIME/compare_test_results.py
jgfouca: Print statements are the easiest thing to port to py3 and there are automated scripts to do this, so we shouldn't sweat this issue too much.

jgfouca:
I think this is ready.

PR #895, opened 2016-12-07
Replace strncpy and snprintf where applied to non-null-term. strings

### LINKED ISSUE #851: Uninitialised variable in t_drvstartf/t_drvstopf
While running ALM via valgrind, I got the following message
```
==58860== Conditional jump or move depends on uninitialised value(s)
==58860==    at 0x10AA9A45E: stpncpy (in /usr/lib/system/libsystem_c.dylib)
==58860==    by 0x10AB0DB74: __strncpy_chk (in /usr/lib/system/libsystem_c.dylib)
==58860==    by 0x1007410B1: GPTLstartf (in ../bld/cesm.exe)
==58860==    by 0x10074472A: gptlstart_ (in ../bld/cesm.exe)
==58860==    by 0x100745230: __perf_mod_MOD_t_startf (in ../bld/cesm.exe)
==58860==    by 0x10007C6BB: __t_drv_timers_mod_MOD_t_drvstartf (in ../bld/cesm.exe)
==58860==    by 0x1000092B2: __cesm_comp_mod_MOD_cesm_run (in ../bld/cesm.exe)
==58860==    by 0x100B51C0F: main (in ../bld/cesm.exe)
```

Since `hashint` is an optional argument of [`t_drvstartf()`](https://github.com/ESMCI/cime/blob/master/driver_cpl/driver/t_driver_timers_mod.F90#L62) and [`t_drvstopf `](https://github.com/ESMCI/cime/blob/master/driver_cpl/driver/t_driver_timers_mod.F90#L69), shouldn't the subroutines be modified ([t_driver_timers_mod.F90#L62](https://github.com/ESMCI/cime/blob/master/driver_cpl/driver/t_driver_timers_mod.F90#L62)) to check the if the optional argument is present? 

```
    if (present(hashint)) then
       call t_startf   (trim(string),hashint)
       call t_adj_detailf(+1)
    endif
```

worleyph:

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes #851 

User interface changes?: no

Code review: 

Valgrind is complaining when strncpy or snprintf are used with
non-null-terminated strings. 

To eliminate the issue, strncpy and snprintf are replaced with
equivalent character-by-character copy loops in the
instances in which the input string is non-null-terminated.

While this resolves #851 , the clean-up to t_drvstartf/t_drvstops suggested by @bishtgautam should still be implemented.

This is BFB in limited testing using an ACME B case, and should be BFB if not buggy.

### DIFF
--- a/share/timing/ChangeLog
+++ b/share/timing/ChangeLog
@@ -1,3 +1,8 @@
+timing_161207: Replaced strncpy and snprintf where applied to 
+               non-null-terminated strings, to avoid memory issues
+               when strncpy and snprintf implementations use strlen
+               to check validity of passed in string length parameter.
+               [Patrick Worley and Gautam Bisht]
 timing_160320: Added routines t_set_prefixf and t_unset_prefixf.
                Setting the prefix adds this to the beginning of all subsequent 
                timer event names (defined in t_startf/t_stopf).
--- a/share/timing/ChangeLog
+++ b/share/timing/ChangeLog
@@ -1,7 +1,8 @@
 timing_161207: Replaced strncpy and snprintf where applied to 
                non-null-terminated strings, to avoid memory issues
-               when strncpy and snprintf implementations use strlen
+               when strncpy and snprintf implementations use strnlen
                to check validity of passed in string length parameter.
+               (Why this causes problems is still a mystery.)
                [Patrick Worley and Gautam Bisht]
 timing_160320: Added routines t_set_prefixf and t_unset_prefixf.
                Setting the prefix adds this to the beginning of all subsequent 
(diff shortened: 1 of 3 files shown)


worleyph:
Deleted the discussion in the original version of the PR as to the reason for the problem:

     Apparently some versions of these
     routines call strlen on the input string as an sanity check on the
     passed-in length parameter, and this is leading to the access of
     random memory. (Note that this perhaps is a performance issue, but
     it has not affected correct functioning - this code has been in
     there for a long while.)

The error message indicates that strnlen is being called to determine the length of the string, which should be safe even with non-null-terminated strings, as long as the string length is correct. The string length is coming from the string length passed in from the Fortran-to-C convention. I have no idea why this would fail. The replacement of strncpy and snprintf by copy loops does eliminate the error from valgrind according to @bishtgautam , and this uses the same string length parameter. I consider this to be a valgrind issue, but the workaround does no harm I suppose.

PR #898, opened 2016-12-08
force compilers version 2 for regression testing

jedwards4b:
Fix for scripts_regression_tests when config_compilers.xml is in version 1
The tests must be forced to version 2. 

Test suite: scripts_regression_tests for both acme and cesm
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes current red lights on master cdash 

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/acme/machines/config_compilers.xml
+++ b/cime_config/acme/machines/config_compilers.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0" encoding="UTF-8"?>
-<config_build>
+<config_compilers version="2.0">
 <!--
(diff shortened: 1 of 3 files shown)


jgfouca:
Does this fix the problem or sweep it under the rug?

jedwards4b:
The unit tests in scripts_regression_tests assume version 2 of config_compilers.xml and don't work with version 1.   The tests involve providing snipets of xml code to the python tools and getting an expected result, but the code snipets were not used in determining the xml version so I had to force it.

PR #467, opened 2016-08-26
Add quotation marks to timing entry names.

quantheory:
This is a (currently untested) variation on #415.

PR #906, opened 2016-12-09
Add quotation marks to timing entry names - Resolved conflicts

jayeshkrishna:
Adds quotation marks to timing entry names.

Test suite: X case
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Code review: @jedwards4b

### DIFF
--- a/cime_config/acme/machines/config_batch.xml
+++ b/cime_config/acme/machines/config_batch.xml
@@ -150,6 +150,17 @@
       </walltimes>
     </batch_system>
 
+    <!-- anvil is PBS -->
+    <batch_system MACH="anvil" type="pbs" version="x.y">
+      <directives>
+        <directive>-A {{ PROJECT }}</directive>
+        <directive>-l nodes={{ num_nodes }}:ppn={{ tasks_per_node }}</directive>
+      </directives>
+      <queues>
+	<queue walltimemax="01:00:00" jobmin="1" jobmax="4320" default="true">acme</queue>
+      </queues>
+    </batch_system>
+
     <!-- edison is SLURM as of Jan-4-2016 -->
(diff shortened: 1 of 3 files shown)


jayeshkrishna:
This is a conflict resolved version of PR #467

jayeshkrishna:
I could not run script_regression_tests.py on anlworkstation (compute001) or blues (on this branch or CIME master)
Got an X case working and verified that timer names had quotes as required after building with a local merge to cime-maint5.1 (I could build the X case after merging this branch to cime-maint5.1)
Will see if I can get script_regression_tests.py to run on anlworkstation+cime-maint5.1_local_merge_with_this_branch

jayeshkrishna:
The script_regression_tests.py passes most of the tests with anlworkstation+cime-maint5.1_local_merge_with_this_branch . The failures are however not related to this change,
* Fail due to issues with mpi-serial (Some MPI functions missing)
* Fail due to manual reduction (from 64 to 16) of pes available on compute001 (By default it uses all cores but compute001 is a shared machine. Some tests fail if you run it with < 64 procs)

jayeshkrishna:
This change should be ready for review+merge

worleyph:
Note that this approach will not affect direct calls to GPTL (e.g. GPTLstart/GPLTstop), only calls that go through perf_mod.F90. So, some of the timers in pio2 will not be quoted, and they will not pick up the prefixes added at the perf_mod level. I don't know if there are other instances of this. I don't have a suggestion about how to address this. Have something similar to perf_mod on the C side to add model-specific customizations? or have C call the fortran entry points? Both seem awkward. However scripts that expect timer names to be quoted will break if they are missing.

PR #910, opened 2016-12-09
Starchive fixes

mvertens:
Fixed short term archiving problems

The new short term archiving utility looks at the component restart file and examines the variable set by the config_archive.xml variable `rest_history_varname` (in cam's case this is `nhfil`) to determine what history files needed to remain in the short term archiving directory when the run is restarted. In some cases (like cam) this includes cam initial files. It ends up that the last cam.i file appears on this list - and its not always relevant and might have been archived in a previous step. 
Previously, the short term archiving script exited if any of the files were not there. And this caused the problem. So now, the archiver only archives those files that are present in `rundir` and ignores the file if it is not there.

Also - removed doc/ directory which is no longer needed

Test suite: manual testing to verify that problem was resolved 
Test baseline: NA
Test namelist changes: NA
Test status: bit-for-bit

Fixes [CIME Github issue #] #594 

User interface changes?: None

Code review:


utils/python/CIME/case_st_archive.py
```diff
@@ -180,7 +181,7 @@ def get_histfiles_for_restarts(case, archive, archive_entry, restfile):
         cmd = "ncdump -v %s %s " %(rest_hist_varname, os.path.join(rundir, restfile))
         rc, out, error = run_cmd(cmd)
```
jedwards4b: This doesn't look right to me.  The error isn't going to be reported under normal run conditions, is that okay?

rljacob:
Yes @mvertens please do the doc remove on a separate branch/PR.  Its has nothing to do with st_archive fixes so shouldn't be on this branch.

utils/python/CIME/case_st_archive.py
mvertens: Yes - we do not want that reported each time - since its just to determine if that variable happens to exist and it was checking this on all components and files. It was very misleading to have it there in the first place. I have replace this with a much more extensive summary of the files that are being short term archived.

PR #918, opened 2016-12-12
runoff src mask = 1 globally

mnlevy1981:
To save computation, the runoff to ocean map originally only considered coastal points as potential sources, but we now potentially have runoff coming from the interior of continents so we don't want that mask any more.

Test suite: none
Test baseline: none
Test namelist changes: none
Test status: changes the runoff to ocean mask generated by the mapping tools

I don't believe this was every formally reported via issues

User interface changes?: none

Code review: @jtruesdal stood over my shoulder while I removed 4 lines of code and reformatted the whitespace of two more.

--- 

We don't want to limit the possible source points for the runoff to ocean
mapping file, so mask_a = 1 everywhere.

### DIFF
--- a/tools/mapping/gen_mapping_files/runoff_to_ocn/src/map_mod.F90
+++ b/tools/mapping/gen_mapping_files/runoff_to_ocn/src/map_mod.F90
@@ -428,13 +428,8 @@ SUBROUTINE map_gridRead(map, rfilename, ofilename, gridtype, lmake_rSCRIP)
 
          map%area_a(n) = 0.5 * 0.5 * cos(lat*DEGtoRAD) * DEGtoRAD * DEGtoRAD
 
-         if (abs(rdirc) < 0.5) then
-            map%mask_a(n) = 1
-            map%frac_a(n) = 1.0_r8
-         else
-            map%mask_a(n) = 0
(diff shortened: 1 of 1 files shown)


rljacob:
What does "runoff coming from the interior of continents" mean?  What physical situation is that trying to handle?

mnlevy1981:
I'm not an expert on this, but just put the source mods requested into the tool... but my understanding is that it's a combination of

1. The RTM runoff model routes all the runoff to the coast and then expects the coupler to distribute it out to the ocean, whereas MOSART maybe doesn't route to the coast?
2. The addition of ice runoff and the glacier model provides cases where ice from the middle of the glacier gets sent to the ocean
3. Sometimes interior lakes drain a little bit of excess volume to the ocean to maintain balance?

mnlevy1981:
To quote an email from the scientist that originally requested this change:

> I believe the original rtm file only mapped points at the coast to the ocean presumably for efficiency.  We had to unmask the mapping file to pick up interior points for interior land sink points, lakes, and glaciers.

jonbob:
Yes, MOSART handles it differently and we end up using unmasked versions of the different files to generate maps for that reason.

PR #919, opened 2016-12-12
Clone fix

### LINKED ISSUE #917: create_clone is not working across different users
I am trying to clone a case created by a different user - using as an example the command

`./create_clone --case testclone --clone /glade/scratch/klindsay/temp_caseroots/g.e20.GECO.T62_g16.mvertens_test.001`

create_clone is picking up the username from the clone directory and trying to write into that user's directory (which generates an error)

The output is the following:

 ```
Creating Case directory /glade/p/work/mvertens/cesm2_0_beta04.klindsay_spinup/cime/scripts/testclone
 Successfully created new case testclone from clone case g.e20.GECO.T62_g16.mvertens_test.001 
/glade/p/work/mvertens/cesm2_0_beta04.klindsay_spinup/cime/scripts/testclone/env_mach_specific.xml already exists, delete to replace
Checking file env_case
Creating batch script case.run
Writing case.run script from input template /glade/p/work/mvertens/cesm2_0_beta04.klindsay_spinup/cime/cime_config/cesm/machines/template.case.run
Writing case.st_archive script from input template /glade/p/work/mvertens/cesm2_0_beta04.klindsay_spinup/cime/cime_config/cesm/machines/template.st_archive
Writing case.lt_archive script from input template /glade/p/work/mvertens/cesm2_0_beta04.klindsay_spinup/cime/cime_config/cesm/machines/template.lt_archive
Locking file env_mach_pes.xml
ERROR: Could not make directory '/glade/scratch/klindsay/testclone/bld/cpl/obj', error: [Errno 13] Permission denied: '/glade/scratch/klindsay/testclone'

```

The problem is that CIME_OUTPUT_ROOT is resolved in case.py in the following statements:

```
        # Resolve the CIME_OUTPUT_ROOT variable, other than this
        # we don't want to resolve variables until we need them
        if output_root is None:
            output_root = self.get_value("CIME_OUTPUT_ROOT")
        self.set_value("CIME_OUTPUT_ROOT", output_root)
```

But this causes create_clone to look at a value of CIME_OUTPUT_ROOT that has the wrong user_name and is causing the above error.  

So I believe what happened is that a change was made to case.py and caused problems to create_clone as a side effect. So I would suggest that create_clone tests be carried out any time changes to case.py are made in the future.  This utility is extensively used by cesm scientists in their workflow. Maybe we can add this as a possible item in the testing summary for PRs.

### PR DESCRIPTION
**Pull request by jedwards4b**
Fixes an issue with create_clone when cloning from one user to another, adds two tests
to scripts_regression_tests to try to assure this works.   (It's not possible to write a test which
actually invokes two user accounts)

Test suite:  scripts_regression_tests
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #917 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/tests/scripts_regression_tests.py
+++ b/utils/python/tests/scripts_regression_tests.py
@@ -1,7 +1,7 @@
 #!/usr/bin/env python
 
 import io, glob, os, re, shutil, signal, sys, tempfile, \
-    threading, time, logging, unittest, getpass
+    threading, time, logging, unittest, getpass, string
 
 from xml.etree.ElementTree import ParseError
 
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: utils/python/tests/scripts_regression_tests.py [hunk 1]
```diff
@@ -170,44 +170,78 @@ def setup_proxy():
...
+        prevtestdir = self._testdirs[0]
+        self._testdirs.append(testdir)
+
+        run_cmd_assert_result(self, "%s/create_clone --clone %s --case %s --keepexe" %
+                              (SCRIPT_DIR, prevtestdir, testdir),from_dir=SCRIPT_DIR)
 
         self._do_teardown.append(testdir)
 
-    def tearDown(self):
+    def test_d_create_clone_new_user(self):
```
jgfouca: Thanks for adding test.

utils/python/tests/scripts_regression_tests.py
```diff
@@ -170,44 +170,78 @@ def setup_proxy():
 ###############################################################################
 class J_TestCreateNewcase(unittest.TestCase):
 ###############################################################################
-    def setUp(self):
-        self._testroot = MACHINE.get_value("CIME_OUTPUT_ROOT")
+    @classmethod
```
jgfouca: Why did this become a class method?

jedwards4b: I wanted the case from the create_newcase test left around so that I could reuse it for the create_clone tests - this is the solution I found.   Now the setup is only run at the beginning of the class and teardown only happens after all the class tests are complete.   I think that this is something we can build further on.

jgfouca: ok

jgfouca:
Should I wait for @mvertens to review?

PR #923, opened 2016-12-14
Mvertens/datm updates

mvertens:
obtain namelist defaults using regular expression matches rather than exact matches

- nmlgen.py now obtains default values from namelist_definition xml
  files using regular expression matches rather than exact matches

- using this refactored namelist_definition_datm.xml and namelist_definition_cice.xml
  this significantly reduced both the size and the complexity of the namelist_definition_datm.xml 
  However, If two matches are equivalent, the FIRST one will be used, so need to make sure
  that matches are not equivalent if possible

  As an example, say datm_mode=CLMCRUNCEP_V5, 

  the following order would result in an INCORRECT setting of streams for this datm_mode
`<value datm_mode="CLMCRUNCEP">CLMCRUNCEP.Solar,CLMCRUNCEP.Precip,CLMCRUNCEP.TPQW`
`<value datm_mode="CLMCRUNCEP_V5">CLMCRUNCEP_V5.Solar,CLMCRUNCEP_V5.Precip,CLMCRUNCEP_V5.TPQW`

  the following will results in the CORRECT set of streams for this datm_mode (notice the $ below)
`<value datm_mode="CLMCRUNCEP$">CLMCRUNCEP.Solar,CLMCRUNCEP.Precip,CLMCRUNCEP.TPQW 
`
`<value  datm_mode="CLMCRUNCEP_V5">CLMCRUNCEP_V5.Solar,CLMCRUNCEP_V5.Precip,CLMCRUNCEP_V5.TPQW`
 - several bug fixes 
    - set domainfile to null if CPLHISTForcingForOcnIce mode
    - fixed bug fixes - numa_IA and 5x5_amazon grids,  introduced in the config_grid.xml refactoring

Test suite: ran several test suites:
   aux_clm45 and aux_clm40 namelist on yellowstone and hobard, comparing to clm4_5_13_r205
   prealpha namelist compares on yellowstone and hobart comparing to cesm2_0_alpha05a
Test baseline: 
Test namelist changes: yes (see above)
Test status: bit-for-bit

Fixes [CIME Github issue #]

User interface changes?:  none

Code review: Bill Sacks, Erik Kluzek

### DIFF
--- a/cime_config/cesm/config_grids.xml
+++ b/cime_config/cesm/config_grids.xml
@@ -35,6 +35,7 @@
       <grid name="rof"    compset="RTM"	  >r05</grid>
       <grid name="rof"    compset="MOSART">r05</grid>
       <grid name="rof"    compset="DROF"  >rx1</grid>
+      <grid name="rof"    compset="DROF%CPLHIST">r05</grid>
       <grid name="glc"	  compset="SGLC"  >null</grid>
       <grid name="glc"	  compset="CISM1" >gland5UM</grid>
       <grid name="glc"	  compset="CISM2" >gland4</grid>
--- a/cime_config/cesm/config_grids.xml
+++ b/cime_config/cesm/config_grids.xml
@@ -64,7 +64,7 @@
 
     <model_grid alias="1x1_numaIA" compset="DATM.+CLM">
       <grid name="atm">1x1_numaIA</grid>
-      <grid name="lnd">1x1</grid>
+      <grid name="lnd">1x1_numaIA</grid>
     </model_grid>
 
     <model_grid alias="1x1_brazil" compset="DATM.+CLM">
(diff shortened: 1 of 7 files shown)


jedwards4b:
I'm concerned about reintroducing an order dependance here - can we use regular expressions in the file to clarify and generate an error if more than one match?   In your example (I think that it was incorrect in your original note):
```the following will results in the CORRECT set of streams for this datm_mode
<value datm_mode="CLMCRUNCEP_V5">CLMCRUNCEP_V5.Solar,CLMCRUNCEP_V5.Precip,CLMCRUNCEP_V5.TPQW
<value datm_mode="CLMCRUNCEP">CLMCRUNCEP.Solar,CLMCRUNCEP.Precip,CLMCRUNCEP.TPQW
```
the following order would result in an INCORRECT setting of streams for this datm_mode
```
<value datm_mode="CLMCRUNCEP">CLMCRUNCEP.Solar,CLMCRUNCEP.Precip,CLMCRUNCEP.TPQW
<value datm_mode="CLMCRUNCEP_V5">CLMCRUNCEP_V5.Solar,CLMCRUNCEP_V5.Precip,CLMCRUNCEP_V5.TPQW
```

Can we use ```datm_mode="CLMCRUNCEP$">``` so that both these don't match?

mvertens:
Ah good idea. I'll try that out.

Issue #848, opened 2016-11-29
Want better organization of xml variables, and more streamlined output from xmlquery


billsacks:
This request emerged from a discussion with @mvertens and @cecilehannay . It is slightly related to #767, in that it touches on the output format from xmlquery.

@cecilehannay and I would like two changes to make it easier to find xml variables of interest and see current values at a glance:

1. A new, streamlined output option from xmlquery that looks like this:

```
category A:
   foo = 1
   bar = 2

category B:
   baz = 3
   etc.
```

i.e., listing category headers and then just giving name/value for each xml variable.

2. Better organization of variables into categories, as well as sorting of categories.

In older versions of CESM, files like env_run.xml were organized so that the most commonly-changed variables were near the top. That is somewhat, but not entirely the case any more. In addition, variables that fit together (such as the various STOP_ variables, or all variables related to a branch run) were visually grouped together in the xml file. Again, this seems to be somewhat the case, but not entirely so.

Here are some examples from cesm1_0_6:

Top of env_run.xml:

```xml
<!--"A setting of TRUE implies a continuation run, valid values: TRUE,FALSE (logical) " -->
<entry id="CONTINUE_RUN"   value="FALSE"  />

<!--"if RESUBMIT is greater than 0, then case will automatically resubmit (integer) " -->
<entry id="RESUBMIT"   value="0"  />

<!-- ====================================== -->

<!--"CCSM tag (char) " -->
<entry id="CCSM_REPOTAG"   value="cesm1_0_6"  />

<!--"case description (char) " -->
<entry id="CASESTR"   value="UNSET"  />

<!-- ====================================== -->

<!--"sets the run length with STOP_N and STOP_DATE, valid values: none,never,nsteps,nstep,nseconds,nsecond,nminutes,nminute,nhours,nhour,ndays,nday,nmonths,nmonth,nyears,nyear,date,ifdays0,end (char) "\
 -->
<entry id="STOP_OPTION"   value="ndays"  />

<!--"sets the run length with STOP_OPTION and STOP_DATE (integer) " -->
<entry id="STOP_N"   value="5"  />

<!--"date in yyyymmdd format, sets the run length with STOP_OPTION and STOP_N (integer) " -->
<entry id="STOP_DATE"   value="-999"  />

<!-- ====================================== -->

<!--"sets frequency of model restart writes (same options as STOP_OPTION) (char) " -->
<entry id="REST_OPTION"   value="$STOP_OPTION"  />

[...]

PR #932, opened 2016-12-15
Refactor xml compvars 

### LINKED ISSUE #890: Reduce verbosity of some case xml files
Several variables in the case xml files which may be different for different components are currently set as XXX_ATM=blahm, XXX_LND=blah and so on for each component,   This proposed change will  
make the case xml files much shorter and easier to read.  Instead of a separate variable for each component the new format would be
```
<entry id="XXX">
<values>
  <value component="ATM">blahm</value>
  <value component="LND">blah</value>
...
</values>
</entry>
```

This makes the xml much more manageable and would also allow to do things with xmlquery and xmlchange to get the settings for all components of variable XXX.

jedwards4b:
In this PR I have combined various variables of the form XXX_COMPCLASS into variable XXX with values for each compclass.  xmlquery can use XXX_COMPCLASS or XXX --subgroup COMPCLASS
or simply XXX to get values for each compclass.  I also removed DRV from the comp classes and replaced it with CPL which cleans up a lot of code

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #890 
Addresses #848 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/build.py
+++ b/utils/python/CIME/build.py
@@ -184,8 +184,7 @@ def case_build(caseroot, case, sharedlib_only=False, model_only=False):
 
     complist = []
     for comp_class in comp_classes:
-        if comp_class == "DRV":
-            comp_class = "CPL"
+        if comp_class == "CPL":
             ninst = 1
             config_dir = None
         else:
@@ -193,7 +192,11 @@ def case_build(caseroot, case, sharedlib_only=False, model_only=False):
             config_dir = os.path.dirname(case.get_value("CONFIG_%s_FILE"%comp_class))
 
         comp = case.get_value("COMP_%s"%comp_class)
+        if ninst is None:
+            import pdb
+            pdb.set_trace()
         thrds =  case.get_value("NTHRDS_%s"%comp_class)
+        expect(ninst is not None,"Failed to get ninst for comp_class %s"%comp_class)
         complist.append((comp_class.lower(), comp, thrds, ninst, config_dir ))
         os.environ["COMP_%s"%comp_class] = comp
 
--- a/utils/python/CIME/build.py
+++ b/utils/python/CIME/build.py
@@ -192,9 +192,6 @@ def case_build(caseroot, case, sharedlib_only=False, model_only=False):
             config_dir = os.path.dirname(case.get_value("CONFIG_%s_FILE"%comp_class))
 
         comp = case.get_value("COMP_%s"%comp_class)
-        if ninst is None:
-            import pdb
-            pdb.set_trace()
         thrds =  case.get_value("NTHRDS_%s"%comp_class)
         expect(ninst is not None,"Failed to get ninst for comp_class %s"%comp_class)
         complist.append((comp_class.lower(), comp, thrds, ninst, config_dir ))
--- a/utils/python/CIME/case_run.py
+++ b/utils/python/CIME/case_run.py
@@ -89,11 +89,6 @@ def run_model(case):
     cmd = case.get_resolved_value(cmd)
     logger.info("run command is %s " %cmd)
 
-    with open("runtime_environment.txt", 'w') as fd:
-        fd.write("run command is: %s"%cmd)
-    run_cmd_no_fail("echo -e '\n' >> runtime_environment.txt && \
-                         env >> runtime_environment.txt")
-
     rundir = case.get_value("RUNDIR")
     run_cmd_no_fail(cmd, from_dir=rundir)
     logger.info("%s MODEL EXECUTION HAS FINISHED" %(time.strftime("%Y-%m-%d %H:%M:%S")))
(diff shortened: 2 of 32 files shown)


rljacob:
With this change, will the case xml files look like the example you gave in #890?

jedwards4b:
> With this change, will the case xml files look like the example you gave in #890?
yes

jgfouca:
It looks like there's a lot of overlap between this PR and other PRs, like the same commits are on both branches.

jedwards4b:
I couldn't do this without that so this branch was created from that branch - this PR will probably be easier to review after that one is accepted.

jgfouca:
Cool, so as soon as that one goes in, rebase this branch and force push.

jedwards4b:
This PR is now ready to be considered for merge.

jgfouca:
@jedwards4b did you rebase this branch?

jedwards4b:
Yes it should be at the top of master

utils/python/CIME/build.py
```diff
@@ -184,16 +184,19 @@ def case_build(caseroot, case, sharedlib_only=False, model_only=False):
...
-            comp_class = "CPL"
+        if comp_class == "CPL":
             ninst = 1
             config_dir = None
         else:
             ninst = case.get_value("NINST_%s"%comp_class)
             config_dir = os.path.dirname(case.get_value("CONFIG_%s_FILE"%comp_class))
 
         comp = case.get_value("COMP_%s"%comp_class)
+        if ninst is None:
```
jgfouca: You probably did not intend to keep this in the PR.

utils/python/CIME/case_run.py
```diff
@@ -89,6 +89,11 @@ def run_model(case):
     cmd = case.get_resolved_value(cmd)
     logger.info("run command is %s " %cmd)
 
+    with open("runtime_environment.txt", 'w') as fd:
```
jgfouca: This should probably follow the same process as case_setup for recording the environment:
```
        with open("software_environment.txt", "w") as f:
            f.write(env_module.list_modules())
        run_cmd_no_fail("echo -e '\n' >> software_environment.txt && env >> software_environment.txt")
```
This might be worth encapsulating in env_mach_specific since we are now doing it in two places.

utils/python/CIME/preview_namelists.py
```diff
@@ -58,16 +58,16 @@ def create_namelists(case):
...
+    # Note - cpl must be last in the loop below so that in generating its namelist,
     # it can use xml vars potentially set by other component's buildnml scripts
     models = case.get_values("COMP_CLASSES")
     models += [models.pop(0)]
     for model in models:
         model_str = model.lower()
         config_file = case.get_value("CONFIG_%s_FILE" % model_str.upper())
         config_dir = os.path.dirname(config_file)
-        if model_str == "drv":
+        if model_str == "cpl":
```
jgfouca: Is there guidance on when drv needs to be converted to cpl and vice versa?

utils/python/CIME/test_status.py
```diff
@@ -175,7 +175,8 @@ def set_status(self, phase, status, comments=""):
 
             if previous_core_phase in self._phase_statuses:
                 expect(self._phase_statuses[previous_core_phase][0] == TEST_PASS_STATUS,
-                       "Cannot move past core phase '%s', it didn't pass" % previous_core_phase)
+                       "Cannot move past core phase '%s', it didn't pass: %s" \
+                           % (previous_core_phase,self._phase_statuses[previous_core_phase][0]))
```
jgfouca: won't the previous core phase status always be FAIL in this case?

utils/python/tests/scripts_regression_tests.py
```diff
@@ -1192,92 +1229,6 @@ def test_save_timings_manual(self):
         self.simple_test(manual_timing=True)
 
 ###############################################################################
-class C_TestXMLQuery(unittest.TestCase):
```
jgfouca: Why were these tests deleted?

utils/python/CIME/build.py
jedwards4b: thanks - i'll remove it.

utils/python/CIME/case_run.py
jedwards4b: I'll take this out and leave it for another PR.

utils/python/tests/scripts_regression_tests.py
jedwards4b: They were not deleted, they were moved into TestCreateNewcase (and improved upon)

jedwards4b:
I think I've addressed all your comments @jgfouca - thanks!

jgfouca:
Should I merge or wait for the other three people to review?

jedwards4b:
I would like to leave it for a day to give them a chance to respond.

billsacks:
Sorry for my belated reply. I tried this out and like how this works - thank you, @jedwards4b !!!

PR #948, opened 2016-12-21
Build improvements

### LINKED ISSUE #600: build timing library first
I like to alter the build order for the shared libs so the timing library is built first.

Its a simpler library then MCT or PIO so it makes more sense to build it first.  Also this allows adding timers to MCT or PIO if one would like to (in a branch).

jedwards4b:
Build gptl prior to building mct so that mct can be instrumented for timing.
Also make sharedlibroot a subdir of output-root rather than test-root

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #600 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/build.py
+++ b/utils/python/CIME/build.py
@@ -423,14 +423,21 @@ def build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid,
     for shared_item in [shared_lib, shared_inc]:
         if (not os.path.exists(shared_item)):
             os.makedirs(shared_item)
-
-    libs = ["mct", "gptl", "pio", "csm_share"]
+    mpilib = case.get_value("MPILIB")
+#    if mpilib == "mpi-serial":
+#        libs = ["mct", "gptl", "pio", "csm_share"]
+#    else:
+    libs = ["gptl", "mct", "pio", "csm_share"]
+    if mpilib == "mpi-serial":
+        libs.insert(0, mpilib)
     logs = []
     sharedlibroot = case.get_value("SHAREDLIBROOT")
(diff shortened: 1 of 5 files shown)


utils/python/CIME/build.py
```diff
@@ -423,14 +423,21 @@ def build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid,
     for shared_item in [shared_lib, shared_inc]:
         if (not os.path.exists(shared_item)):
             os.makedirs(shared_item)
-
-    libs = ["mct", "gptl", "pio", "csm_share"]
+    mpilib = case.get_value("MPILIB")
+#    if mpilib == "mpi-serial":
```
jgfouca: Get rid of comment?

utils/python/CIME/build.py
```diff
@@ -423,14 +423,21 @@ def build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid,
...
+    libs = ["gptl", "mct", "pio", "csm_share"]
+    if mpilib == "mpi-serial":
+        libs.insert(0, mpilib)
     logs = []
     sharedlibroot = case.get_value("SHAREDLIBROOT")
     for lib in libs:
         if lib == "csm_share":
             # csm_share adds its own dir name
             full_lib_path = os.path.join(sharedlibroot, sharedpath)
+        elif lib == "mpi-serial":
```
jgfouca: why would mpi-serial build under mct?

jedwards4b: This is the way it was before, didn't see any need to change it.

PR #957, opened 2016-12-28
Implemented pelayout tool for viewing PE layouts

### LINKED ISSUE #953: pelayout tool
A tool to display and update the NTASKS, NTHRDS and ROOTPE fields from xml is desired.
This tool should have customizable io formats and should warn the user if an update is questionable. 
For example if ROOTPE_OCN < NTASKS_ATM in a B compset then a warning should be delivered.

gold2718:
New script to view the PE layout for a case.

Test baseline: NA
Test namelist changes: NA
Test status: bit for bit

Fixes #953

User interface changes?: New case script to view PE layouts

Code review:

### DIFF
--- a/scripts/Tools/pelayout
+++ b/scripts/Tools/pelayout
@@ -102,6 +102,7 @@ def format_pelayout(comp, ntasks, nthreads, rootpe, format):
     layout_str = re.sub(r"%([-+0-9]*)H", r"{H:\1}", layout_str)
     layout_str = re.sub(r"%([-+0-9]*)R", r"{R:\1}", layout_str)
     layout_str = layout_str.format(**subs)
+    return layout_str
 # End def format_pelayout
 
 ###############################################################################
--- a/scripts/Tools/pelayout
+++ b/scripts/Tools/pelayout
@@ -13,8 +13,10 @@ Interpreted format sequences are:"
 Standard format extensions, such as a field length and padding are supported.
 Example:
   For NTASKS_ATM=480, NTHREADS_ATM=2 and ROOTPE_ATM, \"%C: %06T/%+H\" will show
-ATM:    480/+2
+ATM: 000480/+2
 Note that since %R was omitted, the root PE was not shown.
+Python dictionary-format strings are also supported. For instance,
+--format "{C:4}", will print the component name padded to 4 spaces.
 =======================
 If this tool is missing any feature that you need, please go to github and
 create an issue (https://github.com/ESMCI/cime/issues). Detailed
@@ -48,14 +50,21 @@ def parse_command_line(args):
     parser.add_argument("-format" , "--format",
                         help="Format the PE layout items for each component (see below)")
 
+    parser.add_argument("-set-nasks" ,  "--set-ntasks", default=None,
+                        help="Total number of tasks to set for the case")
+
+    parser.add_argument("-set-nthrds" ,  "--set-nthrds", 
+                        "-set-nthreads", "--set-nthreads", default=None,
+                        help="Number of threads to set for all components")
+
     parser.add_argument("-caseroot" , "--caseroot", default=os.getcwd(),
-                        help="Case directory to build")
+                        help="Case directory to reference")
 
     args = parser.parse_args(args[1:])
 
     CIME.utils.handle_standard_logging_options(args)
 
-    return args.format, args.caseroot
+    return args.format, args.set_ntasks, args.set_nthrds, args.caseroot
 # End def parse_command_line
 
 
@@ -68,11 +77,6 @@ def get_value_as_string(case, var, attribute=None, resolved=False, subgroup=None
         value = convert_to_string(value, thistype, var)
     return value
 
-
-reComp = re.compile(r"%%([0-9]*)C")
-reTask = re.compile(r"%%([-+0-9]*)T")
-reThrd = re.compile(r"%%([-+0-9]*)H")
-reRoot = re.compile(r"%%([-+0-9]*)R")
 ###############################################################################
 def format_pelayout(comp, ntasks, nthreads, rootpe, format):
 ###############################################################################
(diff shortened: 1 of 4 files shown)


scripts/Tools/pelayout
```diff
@@ -0,0 +1,160 @@
+#!/usr/bin/env python
+
+"""
+This utility allows the CIME user to view and modify a case's PE layout.
```
jedwards4b: Perhaps you should not include the comment about modifying until that functionality is included.

gold2718: Yeah, I figure either take out that comment or add the modification options (see #953 for my suggestion additions).

scripts/Tools/pelayout
```diff
@@ -0,0 +1,160 @@
...
+                                     formatter_class=argparse.RawDescriptionHelpFormatter,
+                                     epilog=textwrap.dedent(__doc__) )
+    CIME.utils.setup_standard_logging_options(parser)
+
+    # Set command line options
+    parser.add_argument("-format" , "--format",
+                        help="Format the PE layout items for each component (see below)")
+
+    parser.add_argument("-caseroot" , "--caseroot", default=os.getcwd(),
+                        help="Case directory to build")
```
billsacks: This help text doesn't make sense to me for this tool. I'm guessing you copied it from xmlquery, which has the same non-sensical help-text.

scripts/Tools/pelayout
```diff
@@ -0,0 +1,160 @@
...
+    or using the format input, if it exists.
+    """
+    if (format is None):
+        layout_str = "%s: %6d/%d; %6d"%(comp, ntasks, nthreads, rootpe)
+    else:
+        layout_str = re.sub("%", "%%", format)
+        if (reComp.search(layout_str) is not None):
+            layout_str = re.sub(reComp, r"%\1s", layout_str)%comp
+        # End if
+        layout_str = re.sub("%", "%%", layout_str)
```
billsacks: I believe these repeated applications of `re.sub("%", "%%", layout_str)` should be inside the previous conditional: Currently, if the code within the previous conditional isn't executed, then you end up with extra occurrences of '%%'. For example, I tried:

```
./pelayout -format '%C: %06H'
```

and got:

```
CPL: %%000001
ATM: %%000001
LND: %%000001
ICE: %%000001
OCN: %%000001
ROF: %%000001
GLC: %%000001
WAV: %%000001
ESP: %%000001
```

Also, it would help if you added a comment about what you're doing here, since I couldn't figure this out until I ran the code in an interactive session. Something like: 

```python
# The string formatting operation (replacing  '%\1s' with comp) changes all instances of '%%' back to '%', so we need to change them back to '%%' again.
```

scripts/Tools/pelayout
```diff
@@ -0,0 +1,160 @@
...
+%%  a literal %
+%C  the component name
+%T  the task count for the component
+%H  the thread count for the component
+%R  the PE root for the component
+
+Standard format extensions, such as a field length and padding are supported.
+Example:
+  For NTASKS_ATM=480, NTHREADS_ATM=2 and ROOTPE_ATM, \"%C: %06T/%+H\" will show
+ATM:    480/+2
```
billsacks: When I ran this example, I got a bunch of leading zeroes, in contrast to what you gave as example output:

```
CPL: 000015/+1
ATM: 000015/+1
LND: 000015/+1
ICE: 000015/+1
OCN: 000015/+1
ROF: 000015/+1
GLC: 000015/+1
WAV: 000015/+1
ESP: 000015/+1
```

scripts/Tools/pelayout
billsacks: Actually: Rather than defining your own formatting mini-language here and then needing to write a parser for it, would it make sense to just use python's built-in string formatting? This would allow you to remove this whole function.

Assuming you have a dictionary defined like this:

```python
layout = {'C':'ATM', 'T': 480, 'H': 2, 'R': 0}
```

You can either allow the user to supply an old-style formatting string, such as:

```python
myformat = "%(C)s: %(T)06d/%(H)+d"
mystr = myformat%layout
```

Which gives:

```
'ATM: 000480/+2'
```

Or using the new-style formatting method (which I personally find more readable, though I know it isn't standard in cime):

```python
myformat = "{C}: {T:06}/{H:+}"
mystr = myformat.format(**layout)
```

which again gives:

```
'ATM: 000480/+2'
```

scripts/Tools/pelayout
gold2718: Thanks, I will fix that.

scripts/Tools/pelayout
gold2718: Did you mean nonsensical or non-sensible? How about:
> Case directory to reference
for pelayout and xmlquery and 
> Case directory to change
for xmlchange?

scripts/Tools/pelayout
gold2718: I would like some feedback on Bill's formatting change request. Will users be more comfortable with `%(C)s` instead of `%s`? Less likely to make mistakes? I see this as a UI issue more than a code issue (after all, if we weren't concerned with UI, we would just encourage users to use vi).

scripts/Tools/pelayout
billsacks: Sounds good.

scripts/Tools/pelayout
billsacks: I think there's a typo in your comment. I think the real choices are (for example):

1. Current code: `%C`

2. `%(C)s`

3. `{C}`

I think this is partly a question of UI, and partly a question of code maintainability. I like the idea of supporting user-settable formatting output, and could imagine that idea making its way into other scripts... so the question in my mind is: Do we want to define and parse our own mini-language for this, or just use python's? I'm inclined towards the latter, but I'll be quiet now and let others chime in.

scripts/Tools/pelayout
```diff
@@ -0,0 +1,160 @@
...
+def get_value_as_string(case, var, attribute=None, resolved=False, subgroup=None):
+###############################################################################
+    thistype = case.get_type_info(var)
+    value = case.get_value(var, attribute=attribute, resolved=resolved, subgroup=subgroup)
+    if value is not None and thistype:
+        value = convert_to_string(value, thistype, var)
+    return value
+
+
+reComp = re.compile(r"%%([0-9]*)C")
```
jgfouca: Why are these in the global namespace?

gold2718: I already moved them in my sandbox once I realized that the re.compile gets called once per script execution anyway so there is no efficiency to putting them in the global namespace. Good call -- now I wonder how many other gotchas I found will be caught by someone else :)

gold2718:
I think I have addressed the change requests. I also added the modification options (see #953) which allows you to set the number of threads or modify the layout size (it accepts new sizes which scale cleanly).
I also added a header and a --header option to change the header (or remove it via --no-header).
Finally, python format strings (e.g., "{C:4}: {T:6}/{H:2}; {R:6}") are supported in case folks want to play around with the various format suggestions (the original style also works).

mvertens:
I like Bill's new style formatting suggestion. But maybe have the default
have the rootpe next to tasks and threads on the right.

PR #964, opened 2016-12-30
Mvertens/nmlgen updates

### LINKED ISSUE #958: check_input_data doesn't find / download files needed by the coupler
I added some new mapping files in config_grids.xml. I imported these files to the inputdata repository. Then, when I tried to run on a different machine, I got this message:

```
ERROR: File not found: lnd2glc_fmapname = "cpl/gridmaps/fv0.9x1.25/map_fv0.9x1.25_TO_gland4km_aave.161223.nc"
```

Running `check_input_data` from my case seems to indicate everything is present.

I think the problem is that `check_input_data` looks for files named `*.input_data_list`, but there is no such file for the coupler, so it appears that check_input_data doesn't detect (or download) this missing file.

mvertens:
Performance improvements to namelist generation

- Added significant performance improvements to nmlgen.py to permit
  reuse of a target namelist entry node in the
  namelist_defaults_xxx.xml file rather than repeated searches over
  the namelist entry name.
- Removed the compression of output namelists produced by nmlgen. The
  overwhelming response from users was that this was very confusing.
  Changes to have this take place were very small and we can always
  revert to the compressed namelist output in the future.
- Fixed #958 
- Fixed a significant bug in how user_nl_xxx files were treated.
  A setting in the user_nl_xxx file should ALWAYS take precedence over
  namelist default settings from the call to add_default. There was not 
  a separate issue related to this.
- Changed entry_id and entry_id_namelist schemas to permit new
  requirements needed by pythonization of cesm prognostic components

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #958 

User interface changes?: none

Code review: jedwards

### DIFF
--- a/cime_config/xml_schemas/entry_id_namelist.xsd
+++ b/cime_config/xml_schemas/entry_id_namelist.xsd
@@ -3,7 +3,7 @@
   <!-- base schema -->
   <xs:include schemaLocation="entry_id_base.xsd"/>
   <!-- attributes -->
-  <xs:attribute name="modify_via_xml" type="xs:NCName"/>
+  <xs:attribute name="modify_via_xml" type="xs:string"/>
   <xs:attribute name="skip_default_entry" type="xs:boolean"/>
   <xs:attribute name="per_stream_entry" type="xs:boolean"/>
 
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,28 @@
 ======================================================================
 
+Originator: Chris Fischer 
+Date: 12-23-2016
+Tag: cime5.2.0-alpha.13.bhist
+Answer Changes: None
+Tests: prealpha pgi tests on hobart, BHIST restart on yellowstone
+
+Dependencies:
+
+Brief Summary: Fix pnetcdf module on hobart for pgi.  Update BHIST compset
+names
+
+User interface changes: 
+
+Modified files: git diff --name-status
+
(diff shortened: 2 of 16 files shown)


cime_config/xml_schemas/entry_id_namelist.xsd
```diff
@@ -3,7 +3,7 @@
   <!-- base schema -->
   <xs:include schemaLocation="entry_id_base.xsd"/>
   <!-- attributes -->
-  <xs:attribute name="modify_via_xml" type="xs:NCName"/>
+  <xs:attribute name="modify_via_xml" type="xs:string"/>
```
jedwards4b: Why is this change needed?

mvertens:
Because often a string like involving more than one xml variable will need
to be entered.

dabail10:
I believe this change has broken the logic in the CICE buildnml script. When I chose ice_ic = 'default' or ice_ic = 'none', it is prepending the inputdata root in front of it.

billsacks:
@dabail10 please open a new issue if there's an issue with the current cime that needs to be fixed. You can point from that issue to this PR if you think these changes are the source of the issue.

mvertens:
@jedwards4b - can you please help me look at this. @dbail10 - can you please open a CIME issue as @billsacks asked and assign it to me and @jedwards4b. That would be very helpful. I think we will need the fix both on the release and the master cime.

PR #965, opened 2017-01-01
Mvertens/nmlgen update2

mvertens:
More performance updates for namelist generation

The performance slowdown of the current namelist generation, particularly for multi-instance namelist generation has largely been addressed by removing xml node queries with simple dictionary access in namelist_definition.py. This has also had an impact on improving the performance of the single instance namelist generation. As an example, the time to create a pop namelist with the new namelist generator is now decreased by an order of magnitude.

The interfaces to nmlgen are different - but currently this only effects cime code.
- the instantiation of nmlgen is now done outside of the instance loop
`nmlgen = NamelistGenerator(case, definition_file)`
- the generation of namelist defaults for all entry ids whose group elements are not part of an optional skip_groups argument or do not have the entry attribute skip_default_entry="true" is done in one call within the multi instance loop (this is needed since infile is dependent on the particular instance)
`nmlgen.init_defaults(infile, config,skip_groups=skip_groups)
`
If skip_groups is empty it is not added in the call in any of the CIME buildnml scripts.

Test baseline: 
Test namelist changes: 
Test status: bit-for-bit

Fixes None

User interface changes?: interfaces to nmlgen

Code review:

### DIFF
--- a/utils/python/CIME/XML/namelist_definition.py
+++ b/utils/python/CIME/XML/namelist_definition.py
@@ -141,7 +141,7 @@ class NamelistDefinition(EntryID):
         raise TypeError, \
             "NamelistDefinition does not support `set_value`."
 
-    def get_value_match(self, item, attributes=None, exact_match=True):
+    def get_value_match(self, item, attributes=None, exact_match=True, entry_node=None):
         """Return the default value for the variable named `item`.
 
         The return value is a list of strings corresponding to the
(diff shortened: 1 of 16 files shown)


utils/python/CIME/XML/namelist_definition.py
```diff
@@ -405,5 +409,5 @@ def get_default_value(self, item, attribute=None):
         if attribute is not None:
             all_attributes.update(attribute)
 
-        value = self.get_value_match(item.lower(), all_attributes, True)
+        value = self.get_value_match(item.lower(), all_attributes, True, )
```
jedwards4b: this comma looks wrong to me

mvertens: Fixed the comma. 
I'd like to defer creating common library code to a future PR.

rljacob:
IIRC, at some point these changes are going to require changes in the ACME model build scripts.  Just a note to please tag master before that change goes in.

mvertens:
These changes should be completely backwards compatible with the current
ACME model build scripts.

jedwards4b:
Is the tag be2f946 (tag: cime5.2.0-alpha.13, cesmdev/master) not adequate?

rljacob:
An alpha-type tag will be fine.

PR #967, opened 2017-01-04
Pio namelist fix

jedwards4b:
The pio_numtasks wasn't being set correctly.  After each external call to a component buildnml we must update the case object to assure that changes made by the external are captured.

Test suite: scripts_regression_tests, GECO case in cesm
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/cesm/machines/config_compilers.xml
+++ b/cime_config/cesm/machines/config_compilers.xml
@@ -978,7 +978,9 @@ using a fortran linker.
   <FFLAGS>
     <append> -vec-report </append>
   </FFLAGS>
-
+  <CMAKE_OPTS>
+    <append DEBUG="TRUE"> -DPIO_ENABLE_LOGGING=ON </append>
+  </CMAKE_OPTS>
 </compiler>
 
 <compiler MACH="yellowstone" COMPILER="gnu">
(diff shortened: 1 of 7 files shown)


cime_config/cesm/machines/config_compilers.xml
```diff
@@ -978,7 +978,9 @@ using a fortran linker.
   <FFLAGS>
     <append> -vec-report </append>
   </FFLAGS>
-
+  <CMAKE_OPTS>
```
jgfouca: Does ACME need these XML changes?

cime_config/cesm/machines/config_machines.xml
```diff
@@ -956,6 +956,7 @@
     </module_system>
     <environment_variables>
       <env name="OMP_STACKSIZE">256M</env>
+      <env name="MPI_TYPE_DEPTH">16</env>
```
jgfouca: Ditto ^

jedwards4b: No - these are specific to the new ncar machine laramie

jgfouca:
I approve... will wait for @mvertens

PR #975, opened 2017-01-05
Add a cheesy test to get sha1 info on the dashboard

jgfouca:
This will create a AA_ShowVersion test on our CDash dashboard which should always pass. If you click on the test, you'll see the sha1 that was tested.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: None


billsacks:
This is fine with me. However, in trying to see what this might look like, I see that if you click on _any_ test currently from CDash, you see something like:

```
Testing commit 421b173d8f42c926a299e5e8e0efe46e699597cb
Using cime_model = cesm
```

So will this new test add any value beyond that? i.e., maybe we already have as much as we can get with a cheap & easy solution.

jgfouca:
Oh wow, I'm extra dense! Yeah, that's the info we want.

PR #978, opened 2017-01-06
Mfdeakin sandia/separate net cdf c fortran

### LINKED ISSUE #907: Allow build with NetCDF fortran and C
New NETCF distributions but C and Fortran in separate paths.  CIME needs to handle that.

mfdeakin-sandia:
Adds support for separate NetCDF C and Fortran library installations.

Test baseline: NA
Test namelist changes: NA
Test status: Pass

Fixes #907 

User interface changes?: To use separate NetCDF C and Fortran libraries, specify NETCDF_C_PATH and NETCDF_FORTRAN_PATH in config_compilers.xml before creating the case.

Code review:

### DIFF
--- a/cime_config/acme/machines/Makefile
+++ b/cime_config/acme/machines/Makefile
@@ -354,7 +354,7 @@ INCLDIR += -I$(INSTALL_SHAREDPATH)/include -I$(INSTALL_SHAREDPATH)/$(COMP_INTERF
 
 ifeq ($(NETCDF_SEPARATE), false)
   INCLDIR += -I$(INC_NETCDF)
-else
+else ifeq ($(NETCDF_SEPARATE), true)
   INCLDIR += -I$(INC_NETCDF_C) -I$(INC_NETCDF_FORTRAN)
 endif
 ifdef MOD_NETCDF
@@ -384,7 +384,7 @@ ifeq ($(MODEL),driver)
   ifeq ($(strip $(COMPILER)),nag)
      ifeq ($(NETCDF_SEPARATE), false)
        SLIBS += -Wl,-Wl,,-rpath=$(LIB_NETCDF)
-     else
+     else ifeq ($(NETCDF_SEPARATE), true)
        SLIBS += -Wl,-Wl,,-rpath=$(LIB_NETCDF_C)
        SLIBS += -Wl,-Wl,,-rpath=$(LIB_NETCDF_FORTRAN)
      endif
@@ -438,7 +438,7 @@ CONFIG_ARGS +=  CC="$(CC)" FC="$(FC)" MPICC="$(MPICC)" \
 
 ifeq ($(NETCDF_SEPARATE), false)
   CONFIG_ARGS += NETCDF_PATH=$(NETCDF_PATH)
-else 
+else ifeq ($(NETCDF_SEPARATE), true)
   # The mct library needs the NetCDF_C library
   CONFIG_ARGS += NETCDF_PATH=$(NETCDF_C_PATH)
 endif
@@ -493,7 +493,7 @@ endif
 ifndef SLIBS
   ifeq ($(NETCDF_SEPARATE), false)
     SLIBS := -L$(LIB_NETCDF) -lnetcdf
-  else
+  else ifeq ($(NETCDF_SEPARATE), true)
     SLIBS := -L$(LIB_NETCDF_C) -L$(LIB_NETCDF_FORTRAN) -lnetcdf
   endif
 endif
@@ -615,7 +615,7 @@ CMAKE_OPTS += -D CMAKE_Fortran_FLAGS:STRING="$(FFLAGS) $(INCLDIR)" \
 # Allow for separate installations of the NetCDF C and Fortran libraries
 ifeq ($(NETCDF_SEPARATE), false)
   CMAKE_OPTS += -D NETCDF_DIR:PATH=$(NETCDF_PATH)
-else
+else ifeq ($(NETCDF_SEPARATE), true)
   # NETCDF_Fortran_DIR points to the separate
   # installation of Fortran NetCDF for PIO
   CMAKE_OPTS += -D NETCDF_DIR:PATH=$(NETCDF_C_PATH) \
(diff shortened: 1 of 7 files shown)


cime_config/acme/machines/Makefile
```diff
@@ -185,14 +185,40 @@ ifndef AR
...
 endif
 
-ifdef NETCDF_PATH
-  ifndef INC_NETCDF
-    INC_NETCDF:=$(NETCDF_PATH)/include
-  endif
-  ifndef LIB_NETCDF
-    LIB_NETCDF:=$(NETCDF_PATH)/lib
+# To prevent breaking cleaning and building components which don't require NetCDF directly
+ifeq ($(if $(or $(filter pio, $(subst /,  , $(MAKECMDGOALS))), $(filter exec_se, $(subst /,  , $(MAKECMDGOALS)))), true, false), true)
```
jedwards4b: Can you explain what you are doing here - I'm sure we can find a less complicated method.

cime_config/acme/machines/Makefile
```diff
@@ -327,8 +352,10 @@ endif
 #===============================================================================
 INCLDIR += -I$(INSTALL_SHAREDPATH)/include -I$(INSTALL_SHAREDPATH)/$(COMP_INTERFACE)/$(ESMFDIR)/$(NINST_VALUE)/include
 
-ifdef INC_NETCDF
+ifeq ($(NETCDF_SEPARATE), false)
   INCLDIR += -I$(INC_NETCDF)
+else ifeq ($(NETCDF_SEPARATE), true)
```
jedwards4b: we don't need the comparison here - just else is fine

cime_config/acme/machines/Makefile
```diff
@@ -450,8 +491,13 @@ endif
 
 # System libraries (netcdf, mpi, pnetcdf, esmf, trilinos, etc.)
 ifndef SLIBS
-   SLIBS := -L$(LIB_NETCDF) -lnetcdf
+  ifeq ($(NETCDF_SEPARATE), false)
+    SLIBS := -L$(LIB_NETCDF) -lnetcdf
+  else ifeq ($(NETCDF_SEPARATE), true)
+    SLIBS := -L$(LIB_NETCDF_C) -L$(LIB_NETCDF_FORTRAN) -lnetcdf
```
jedwards4b: Actually you are only linking the netcdf c library here - the netcdf fortran library is netcdff.
You should be able to modify these SLIBS lines to have both, and I believe that it will then work for both shared and static links
-lnetcdf -lnetcdff

cime_config/acme/machines/Makefile
```diff
@@ -659,8 +714,13 @@ db_flags:
 
 test_fc: test_fc.o
 	$(LD) -o $@ test_fc.o $(LDFLAGS)
+ifeq ($(NETCDF_SEPARATE), false)
 test_nc: test_nc.o
 	$(LD) -o $@ test_nc.o -L$(LIB_NETCDF) -lnetcdf $(LDFLAGS)
+else ifeq ($(NETCDF_SEPARATE), true)
+test_nc: test_nc.o
+	$(LD) -o $@ test_nc.o -L$(LIB_NETCDF_C) -lnetcdf $(LDFLAGS)
```
jedwards4b: this should be -lnetcdf -lnetcdff

cime_config/acme/machines/Makefile
mfdeakin-sandia: Won't that fail if NETCDF_SEPARATE isn't defined?

cime_config/acme/machines/Makefile
mfdeakin-sandia: This detects if we're building either a PIO library or the coupled executable. I'd definitely prefer a simpler method; one thought I had was defining another environment variable for those build steps, but that involves modifying the Python.
The substitution is needed for the pio library because the target is actually a path which depends on the case name; so determining if pio is being built requires searching the string for it.
The substitution isn't actually needed for the coupled executable, so I can remove that.

cime_config/acme/machines/Makefile
mfdeakin-sandia: I'm linking the same libraries as the old code, so I'm not certain it's necessary?

cime_config/acme/machines/Makefile
jedwards4b: The variable MODEL in the makefile is what you want.   When pio is built MODEL=PIO1 or PIO2, when the executable is built MODEL = CIME_MODEL (acme or cesm)

cime_config/acme/machines/Makefile
jedwards4b: Okay, I see the point.

cime_config/acme/machines/Makefile
jedwards4b: No I'm pretty sure this is wrong - it's only working on skybridge because of the change you made in config_compilers.xml

mfdeakin-sandia:
I've implemented the requested changes. Anything else?

mfdeakin-sandia:
Also, @jgfouca, you might want to verify these changes on Skybridge yourself as well; as I'm not using the sems modules and have issues loading them.

jgfouca:
@mfdeakin-sandia as long as you're using the environment defined by config_machines.xml, should be fine.

cime_config/acme/machines/Makefile
```diff
@@ -185,14 +185,41 @@ ifndef AR
...
 
-ifdef NETCDF_PATH
-  ifndef INC_NETCDF
-    INC_NETCDF:=$(NETCDF_PATH)/include
-  endif
-  ifndef LIB_NETCDF
-    LIB_NETCDF:=$(NETCDF_PATH)/lib
+# To prevent breaking cleaning and building components which don't require NetCDF directly
+# NetCDF is only required for building the coupled model and the PIO library
+ifeq ($(if $(or $(findstring PIO, $(MODEL)), $(findstring CIME_MODEL, $(MODEL))), true, false), true)
```
jedwards4b: CIME_MODEL is not going to be found in $(MODEL) cesm or acme is.  cesm in the cesm Makefile, acme in the acme one.

mfdeakin-sandia:
@jgfouca The issue is that I changed the skybridge entry in config_compilers.xml, as it was attempting to use nf-config which doesn't exist with the netcdf-intel module I'm using.
@jedwards4b $(MODEL) is reported as driver when I print it in acme.bldlog, so I've changed it to that instead.

PR #982, opened 2017-01-07
new CESM2 defaults - 1hr pop cpl, Robert Filtering, grid files for  EBM, mushy ice

jtruesdal:
New defaults for ICE an OCN which match the current state of CESM2 development.  
   1hr Ocean coupling using Robert Filtering and RASM1 coupling order
   Separate LIQ and ICE runoff files for EBM (With EBM on LIQROF is no longer spread over many ocn pts.)
   Mushy Ice algorithm is now the default for providing ice temperatures.  Previously set to -1.8

Test suite: cime regression 
Test baseline: 
Test namelist changes: NA
Test status: climate changing - new ICE and POP defaults for CESM

Fixes [CIME Github issue #]

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/cesm/config_grids.xml
+++ b/cime_config/cesm/config_grids.xml
@@ -1218,20 +1218,20 @@
     </gridmap>
 
     <gridmap rof_grid="rx1" ocn_grid="gx3v7" >
-      <map name="ROF2OCN_LIQ_RMAPNAME">cpl/cpl6/map_rx1_to_gx3v7_e1000r500_090903.nc</map>
-      <map name="ROF2OCN_ICE_RMAPNAME">cpl/cpl6/map_rx1_to_gx3v7_e1000r500_090903.nc</map>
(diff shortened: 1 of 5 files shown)


rljacob:
These "CESM2" defaults change values to things used by ACME.  This should not be merged until we've had a chance to review it.

mvertens:
I think we should add a cime_model attribute to the xml files where so that
we can distinguish between cesm and acme specific settings. This would only
occur in a small handful of xml variables.

mvertens:
I agree with Rob that this should not have been merged until ACME reviewed
it. Not sure about the best way to proceed right now. I think adding this
extra attribute and adding the previous ACME defaults will solve the
problem. Is there agreement on this?

mvertens:
I think this will actually be very easy to do - so I am working on putting
together a PR to get this into cime.

rljacob:
That sounds good.

PR #984, opened 2017-01-08
added cime_model attribute to all value entries that had compset attr…

### LINKED ISSUE #983: driver config_component.xml with compset attributes need model specific attribute
The config_component.xml has compset attributes that are not always the same for cesm and acme. This became particularly true in the recent PR where cesm coupling frequencies were changed and this effected the acme defaults incorrectly. 
To mitigate this in the future, all value elements in config_component.xml  also need a cime_model attribute so that cesm developments can be independent of acme developments.

mvertens:
add cime_model attribute to driver_cpl config_component.xml

Value tags in driver_cpl/cime_config/ config_component.xml has compset attributes that are not always the same for cesm and acme. This became particularly true in the recent PR where cesm coupling frequencies were changed and this effected the acme defaults incorrectly.
To mitigate this in the future, all value elements in config_component.xml also need a cime_model attribute so that cesm developments can be independent of acme developments.

Test baseline: none
Test namelist changes: none
Test status: bit for bit

Fixes #983

User interface changes?: None

Code review:

### DIFF
--- a/driver_cpl/cime_config/config_component.xml
+++ b/driver_cpl/cime_config/config_component.xml
@@ -366,7 +366,8 @@
     <valid_values>none,never,nsteps,nstep,nseconds,nsecond,nminutes,nminute,nhours,nhour,ndays,nday,nmonths,nmonth,nyears,nyear,date,ifdays0,end</valid_values>
     <default_value>never</default_value>
     <values>
-      <value compset="_DATM.+_CLM">ndays</value>
+      <value cime_model="acme" compset="_DATM.+_CLM">ndays</value>
+      <value cime_model="cesm" compset="_DATM.+_CLM">ndays</value>
     </values>
     <group>run_begin_stop_restart</group>
     <file>env_run.xml</file>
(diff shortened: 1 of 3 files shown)


jgfouca:
Since it's pretty important that this works correctly, might be nice to have a test for this.

mvertens:
Agreed. But it would need to be a baseline comparison tests I think that
needs to include the prognostic components of the coupled model system. Not
sure how else you would test that this is being done correctly for the
target model system.
Could you clarify?

Right now - this builds in the backwards compatibility that the PR #982
broke for ACME. So I wanted to get this in as soon as possible since we
need PR #982 for the next upcoming CESM2 beta tag.

jgfouca:
@mvertens I was thinking something much simpler than that. I'd like a test to confirm that when you ask for the values of an entry-id that has model specific stuff that you in fact get the model specific values.

mvertens:
Ah.. Thanks for the clarification. So this should be added to the scripts
regression test right? Since this change is completely cime specific.

mvertens:
I've actually never added any test to the scripts regression test - anyway
some else who is an expert could add this test - given that I really don't
want to back out PR #982 and I agree that we should add this test as part
of this PR.

jgfouca:
@mvertens 

in scripts_regression_tests, K_TestCimeCase.test_cime_case, add this within the with Case ... block:

```
model_specific_val = case.get_value("X")
if CIME.utils.get_model() == 'cesm':
    self.assertEqual(model_specific_val, "expected cesm val")
else:
    self.assertEqual(model_specific_val, "expected acme val")
```

where X is one of the entries where acme and cesm differ.

mvertens:
Thank you!!! I'll add this and rerun the regression tests.

jgfouca:
Awesome. And just in case you didn't know, you can re-run just this test:

```
./scripts_regression_tests.py K_TestCimeCase
```

mvertens:
So I added the following entry and the test passed for both CIME_MODEL set
to acme and cesm

            model_specific_val = case.get_value("CPL_SEQ_OPTION")
            if CIME.utils.get_model() == 'cesm':
                self.assertEqual(model_specific_val, "RASM_OPTION1")
            else:
                self.assertEqual(model_specific_val, "CESM1_MOD")

I just pushed my changes back to github.

jgfouca:
Thanks, that looks perfect.

driver_cpl/cime_config/config_component.xml
```diff
@@ -366,7 +366,8 @@
     <valid_values>none,never,nsteps,nstep,nseconds,nsecond,nminutes,nminute,nhours,nhour,ndays,nday,nmonths,nmonth,nyears,nyear,date,ifdays0,end</valid_values>
     <default_value>never</default_value>
     <values>
-      <value compset="_DATM.+_CLM">ndays</value>
+      <value cime_model="acme" compset="_DATM.+_CLM">ndays</value>
```
jedwards4b: why is the cime_model modifier added here?  Just to document that you can?

mvertens:
The cime_model is added everywhere there is a compset attribute right now
to give acme the freedom to change the behavior. So there are currently
duplicates. But once this PR is in - acme is free to change any of these.

Separate files is an interesting idea - but for now I think this should go
in as soon as possible in order not to back out the previous PR. We can
certainly explore separate files moving forwards.

rljacob:
I like separate files better since cime/cime_config is where these differences are supposed to go.  And we're not in a huge hurry since ACME isn't using the head of master yet.

mvertens:
Thanks for clarifying. So I suggest that any variable in the driver_cpl
config_component.xml that has a compset attribute  in the value tag be
place in a component specific file.

jedwards4b:
I think I am suggesting that the driver config_components.xml be separate files. Perhaps config_acme_components.xml and config_cesm_components.xml and the only change required to achieve this is the file specification made in config_files.xml

mvertens:
I think this will lead to potentially having to change things in two places
when new driver changes are needed. I prefer to have most of the driver
settings that are not compset specific only defined in one place. A small
set of settings would then be model specific.

mvertens:
I agree. However, I still think we need two files - a config_component.xml
for variables that are neither acme or cesm dependent (i.e. have no compset
attribute seems to be the rule in my mind) and those that are. That way for
most variables we would only change things in one place. Does that make
sense?

rljacob:
Yes that makes sense.

PR #987, opened 2017-01-08
Presence of a memleak should not impact case.test success

### LINKED ISSUE #614: I'd prefer if tests generated a non-zero exit status under fewer conditions
I'll preface this by saying that I'm fine with being overruled here if the majority is happy with the current behavior. But I wanted to throw in my $0.02 on this issue.

I use email filters to separate out successful batch jobs from jobs that failed. In the past this has really helped when I run test suites with hundreds of jobs: I basically get instant email notification of any major problems.

In cime4, and for a long time before, a failed test job meant that the run failed. However, with cime5, a failure exit code is generated under many more conditions - such as a failed MEMLEAK test or a failed baseline comparison. This is a problem for me, because MEMLEAK failures are often expected / false positives (at least for now), and baseline comparison failures are often expected. With all of this "noise", the email filter becomes basically useless, and I'm going to end up having to filter out the failed jobs as well to avoid flooding my inbox.

I'd prefer if the behavior were changed back to the cime4 behavior, where a failure (non-zero) exit status were only generated in the case of a definitive problem. A model crash should generate a non-zero exit status, as it did before. I'd also be happy with a failure in the in-test comparison (e.g., base vs. rest) generating a non-zero exit status (though would also be okay if it didn't). But I feel that pieces of the test that are often expected to fail, such as baseline comparisons and (currently) memleak, should generate a zero exit status.

Maybe get_overall_test_status could be used for this exit status, with check_throughput=False, check_memory=False, ignore_namelists=True, ignore_memleak=True; and TEST_DIFF_STATUS would result in a 0 exit status.

jgfouca:
Test suite: scripts_regression_tests O_TestTestScheduler, code_checker
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #614 

User interface changes?: case.test returns zero regardless of MEMLEAK check

### DIFF
--- a/utils/python/CIME/SystemTests/system_tests_common.py
+++ b/utils/python/CIME/SystemTests/system_tests_common.py
@@ -158,7 +158,7 @@ class SystemTestsCommon(object):
             self._test_status.set_status(RUN_PHASE, status, comments=("time=%d" % int(time_taken)))
 
         # We only return success if every phase, build and later, passed
-        return self._test_status.get_overall_test_status(ignore_namelists=True) == TEST_PASS_STATUS
+        return self._test_status.get_overall_test_status(ignore_namelists=True, ignore_memleak=True) == TEST_PASS_STATUS
 
     def run_phase(self):
         """
--- a/utils/python/CIME/SystemTests/system_tests_common.py
+++ b/utils/python/CIME/SystemTests/system_tests_common.py
@@ -158,7 +158,7 @@ class SystemTestsCommon(object):
             self._test_status.set_status(RUN_PHASE, status, comments=("time=%d" % int(time_taken)))
 
         # We only return success if every phase, build and later, passed
-        return self._test_status.get_overall_test_status(ignore_namelists=True, ignore_memleak=True) == TEST_PASS_STATUS
+        return success
 
     def run_phase(self):
         """
--- a/utils/python/CIME/SystemTests/system_tests_common.py
+++ b/utils/python/CIME/SystemTests/system_tests_common.py
@@ -157,7 +157,8 @@ class SystemTestsCommon(object):
         with self._test_status:
             self._test_status.set_status(RUN_PHASE, status, comments=("time=%d" % int(time_taken)))
 
-        # We only return success if every phase, build and later, passed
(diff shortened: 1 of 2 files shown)


rljacob:
Why should a baseline failure be reported as success?

jgfouca:
@billsacks I'm a little hesitant to make baseline failure return 0 since that's an unambiguous problem.

jgfouca:
I'm trying to think of the consequences of case.test reporting 0 for a DIFF. I think it's actually pretty minimal, so I'm willing to concede this point if that's OK with others.

billsacks:
@rljacob and @jgfouca : First, to be clear, I'm NOT talking about the TestStatus here: I'm talking about the job status - e.g., what you'd get in the shell from checking `$?`. My argument was that the test job should return a 0 return code in most cases, reserving a non-zero return for cases where something went very wrong. I realize that "very" is subjective here, but to me, a difference from baseline doesn't constitute something going "very" wrong.

The practical reason for my argument is that I used to set up email filters so that I can be notified immediately of tests where something went very wrong and I need to dig into them further - whereas I dump all emails for tests that generate a 0 return into some folder that I ignore. For tests of active components, we often _expect_ baseline comparisons to fail. So if a baseline failure generates a non-zero return code, then this sort of email filtering becomes pretty useless.

All that said: If others feel that a baseline failure should generate a non-zero return code, I'm willing to defer to the majority opinion.

rljacob:
Oh the job status.  That makes sense then.

utils/python/CIME/SystemTests/system_tests_common.py
```diff
@@ -158,7 +158,7 @@ def run(self):
             self._test_status.set_status(RUN_PHASE, status, comments=("time=%d" % int(time_taken)))
 
         # We only return success if every phase, build and later, passed
```
billsacks: Thanks a lot, @jgfouca - I really appreciate it.

However, you should now remove this comment on line 160.

jgfouca:
@billsacks ok, addressed your requested change.

billsacks:
Looks good now - thanks again!

PR #988, opened 2017-01-08
Add support for --machine to scripts_regression_tests

### LINKED ISSUE #829: scripts_regression_tests needs --machine option
There are two machines possible from a cori (nersc) login node: cori-haswell and cori-knl
since the regex in config_machines.xml will match both, one must be selected as the default system.
This is fine except that scripts_regression_tests doesn't have any way to specify that the non-default system should be used for testing.

jgfouca:
In order to avoid having to add --machine arguments in 50 spots, add concept of CIME_MACHINE environment variable.

Test suite: scripts_regression_tests.py --machine sandia-srn-sems
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #829 

User interface changes?: Adds --machine option to scripts_regression_tests

Code review: @jedwards4b

### DIFF
--- a/utils/python/CIME/XML/machines.py
+++ b/utils/python/CIME/XML/machines.py
@@ -41,7 +41,10 @@ class Machines(GenericXML):
             GenericXML.read(self, local_infile)
 
         if machine is None:
-            machine = self.probe_machine_name()
+            if "CIME_MACHINE" in os.environ:
+                machine = os.environ["CIME_MACHINE"]
+            else:
+                machine = self.probe_machine_name()
 
         expect(machine is not None, "Could not initialize machine object from %s or %s" % (infile, local_infile))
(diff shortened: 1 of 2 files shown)


gold2718:
Any reason this is different from --mach in create_newcase?
Anyone know why create_newcase isn't --machine like create_test?
I'd love some consistency (and prefer --machine).

jgfouca:
Do not merge, by-hand testing reveals this PR doesn't work.

jgfouca:
OK, it's ready now.

jgfouca:
@gold2718 Consistency would be nice. I think --mach matches the legacy options used by CIME2 perl scripts. We've using --machine in most of the new tools, so that's why there's inconsistency.

PR #1005, opened 2017-01-12
Add overrides for pes

### LINKED ISSUE #1002: Settings in component config_compsets.xml not being honored
In cam config_compsets.xml there is:
```
    <entry id="NTHRDS_ATM">
      <values>
        <value compset="SPCAMS">1</value>
        <value compset="SPCAMM">1</value>
      </values>
    </entry>
```
This request does not appear to be honored when creating a test
SMS_D.f19_f19.FSPCAMS.yellowstone_intel.cam-outfrq9s

jedwards4b:
Add overrides section to the config_pes.xml file to allow altering pe layout definitions for certain cases.    Add schema for config_pes.xml and test file when opening

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1002  (these fields should not have been in config_compsets.xml but the support to
                      put them into config_pes.xml was incomplete) 

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/acme/allactive/config_pes.xml
+++ b/cime_config/acme/allactive/config_pes.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 
-<config_pes>
+<config_pes version="2.0">
 
   <grid name="any">
     <mach name="any">
--- a/cime_config/acme/config_files.xml
+++ b/cime_config/acme/config_files.xml
@@ -108,6 +108,7 @@
     <group>case_last</group>
     <file>env_case.xml</file>
     <desc>file containing specification of all pe-layouts for primary component (for documentation only - DO NOT EDIT)</desc>
+    <schema>$CIMEROOT/cime_config/xml_schemas/config_pes.xsd</schema>
   </entry>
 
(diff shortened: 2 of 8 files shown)


cime_config/acme/allactive/config_pes.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 
-<config_pes>
+<config_pes version="2.0">
```
rljacob: This needs to be added to acme/allactive/config_pesall.xml as well.  That's the file we're currently using.

jedwards4b: The file will still work without this change but it won't be checked against the schema.

billsacks:
@jedwards4b : Am I right in thinking that this change to the handling of overrides - such as for CISM1 - needs to be applied in all components that define PE layouts? e.g., CLM will need an identical section, as will CAM, and maybe others. I remember @mvertens and myself struggling with how to handle this (not liking the idea that information like this needs to be duplicated in all components' config_pes files), but I can't remember what, if anything, we concluded.

PR #1007, opened 2017-01-13
Mvertens/split drv config component

mvertens:
Separated out acme/cesm model specific config_component.xml

All compset dependent xml variables in driver_cpl/cime_config/config_component.xml have no been separated into config_component_cesm.xml and config_component_acme.xml in driver_cpl/cime_config. This will provide the ability for acme and cesm to modify these variables independently but still have all common non-compset dependent variables in one xml file.

Test baseline: NA
Test namelist changes: NA
Test status: bit for bit

Fixes: None

User interface changes?: None

Code review:

### DIFF
--- a/cime_config/acme/config_files.xml
+++ b/cime_config/acme/config_files.xml
@@ -204,7 +204,7 @@
 
   <entry id="CONFIG_CPL_FILE_MODEL_SPECIFIC">
     <type>char</type>
-    <default_value>$CIMEROOT/cime_config/$MODEL/config_component_drv.xml</default_value>
+    <default_value>$CIMEROOT/driver_cpl/cime_config/config_component_$MODEL.xml</default_value>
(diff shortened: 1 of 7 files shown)


rljacob:
Can other entries from config_component.xml be moved?   ACME wants DRV_THREADING and SAVE_TIMING to be default TRUE, possibly BFBFLAG as well.

mvertens:
Sure. I had a very simple rule when I implemented this - but I can easily
add the above variables before the PR is accepted. Anything else?

mvertens:
I believe that setting BFBFLAG to TRUE incurs a performance penalty.

rljacob:
That was the thinking about BFBFLAG but hasn't been verified at least in testing with ACME cases.

rljacob:
That's all the variables for now.

PR #1008, opened 2017-01-13
Machine updates

jedwards4b:
Updates for cori and cheyenne
Test suite: scripts_regression_tests on cori-knl and cheyenne
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/cesm/machines/config_batch.xml
+++ b/cime_config/cesm/machines/config_batch.xml
@@ -111,7 +111,7 @@
     <batch_query args="-f" >qstat</batch_query>
     <batch_submit>qsub </batch_submit>
     <batch_directive>#PBS</batch_directive>
-    <jobid_pattern>^(\d+)\.</jobid_pattern>
+    <jobid_pattern>^(\S+)$</jobid_pattern>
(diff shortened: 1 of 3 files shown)


cime_config/cesm/machines/config_batch.xml
```diff
@@ -111,7 +111,7 @@
     <batch_query args="-f" >qstat</batch_query>
     <batch_submit>qsub </batch_submit>
     <batch_directive>#PBS</batch_directive>
-    <jobid_pattern>^(\d+)\.</jobid_pattern>
+    <jobid_pattern>^(\S+)$</jobid_pattern>
```
rljacob: What does this do?  Wondering if ACME needs it too.

jedwards4b: The jobid string returned from pbs comes in the form 1234.pbsserver   we were only using the 1234 part and removing the pbsserver part, it turns out that we need this part for Cheyenne and I don't think that it has a negative consequence to use the whole string everywhere.

PR #1012, opened 2017-01-16
Mfdeakin sandia/no invalid characters fortran

### LINKED ISSUE #1000: cprnc should never produce invalid XML characters
The old version of CPRNC is sometimes producing output that XML cannot handle and therefore the nightly submission to cdash fails if this output is present. Example:

``` x2oacc_ox_cnt                                                                                                                                                                                                                                                                     
                   1        1^@^@^@^@
```

Which can be found in this file:

/sems-data-store/ACME/baselines/gnu/next/ERS_Ld5.T62_oQU120.CMPASO-NYF.melvin_gnu/cpl.hi.0001-01-06-00000.nc

mfdeakin-sandia:
I believe this fixes #1000, as it fixes it in my reproduction of the error, though more extensive testing is likely needed.

Test suite: Some of scripts_regression_test
Test status: BFB

### DIFF
--- a/tools/cprnc/compare_vars_mod.F90.in
+++ b/tools/cprnc/compare_vars_mod.F90.in
@@ -251,6 +251,12 @@ contains
     character(len=80) :: min_str(2), max_str(2),  dmax_str, rdmax_str, space
     logical :: compare2
 
+    min_str = ''
+    max_str = ''
+    dmax_str = ''
+    rdmax_str = ''
+    space = ''
+
     if(present(tindex)) then
        t = tindex
     else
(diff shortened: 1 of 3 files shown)


jgfouca:
Thanks @mfdeakin , I'm happy as long as this problem does not re-emerge in the python implementation of cprnc.

mfdeakin-sandia:
@jgfouca I'm pretty sure the issue was due to the strings memory not being initialized properly; so this is not something that could accidentally happen in python :)

mfdeakin-sandia:
Assuming no one objects, I will merge this

PR #1015, opened 2017-01-17
Fix race condition in Z_FullSystemTests test re-runs

### LINKED ISSUE #1014: There's a race condition in Z_TestSystemFull
There's a window in which the last wait_for_tests call will not wait for the re-run tests because it sees the initial PASS. This is what is causing the failures on skybridge.

jgfouca:
Fix is to manually set TestStatus files back to PEND
for the run_phase before resubmission.

Test suite: scripts_regression_tests.py Z_FullSystemTests
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1014

User interface changes?: None

Code review: @jedwards4b

### DIFF
--- a/utils/python/tests/scripts_regression_tests.py
+++ b/utils/python/tests/scripts_regression_tests.py
@@ -1115,6 +1115,16 @@ class Z_FullSystemTest(TestCreateTestCommon):
                 pass
             else:
                 casedir = os.path.join(TEST_ROOT, "%s.%s" % (test, self._baseline_name))
+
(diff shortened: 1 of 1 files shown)


jgfouca:
@jedwards4b whoops, didn't mean to submit this yet. Tests still running on skybridge.

jgfouca:
The ERI test is still broken if you try to re-run it.

jgfouca:
OK, this is ready. ERS and ERR also turn out to be broken.

PR #1025, opened 2017-01-19
Refactor implementation of forcing procs/threads in create_test

### LINKED ISSUE #1017: create_test --force-procs and --force-threads options should change test name
I see the convenience of the `--force-procs` and `--force-threads` options added to create_test in #1016 (addressing #897 ). However, for the sake of baseline generation and comparison, as well as tracking the history of passing / failing tests, I feel that it is a Bad Idea to allow changing test configuration without noting this in the test name.

Currently, you could run:

```
./create_test --force-procs 64 SMS.f45_g37.A.yellowstone_intel --generate foo
```

then later run

```
./create_test SMS.f45_g37.A.yellowstone_intel --compare foo
```

If there were differences due to the different PE layout (which is expected for some compsets, unexpected but possible due to bugs for others), then it would be very confusing to try to track down the cause of these differences.

I would suggest that specifying `--force-procs` or `--force-threads` modifies all of the test names to note this modification, by adding `_PNxM` to the test name - unless the test already has a `_PNxM` modifier, in which case it appears that that modifier currently takes precedence over the `--force` option. (I don't think we have a current naming convention for `--force-threads` without `--force-procs`, but we could use something like `PxM`.)

Tagging @jgfouca @jedwards4b

jgfouca:
Dynamically re-write test names to reflect these choices.

Add new TOTAL_CORES entry id in order to better support test_schedulers
scheduling.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1017 

User interface changes?: Auto test-renaming in create_test

Code review: @jedwards4b

### DIFF
--- a/driver_cpl/cime_config/config_component.xml
+++ b/driver_cpl/cime_config/config_component.xml
@@ -1967,6 +1967,14 @@
     <desc>total number of tasks and threads (setup automatically - DO NOT EDIT)</desc>
   </entry>
 
+  <entry id="TOTAL_CORES">
+    <type>integer</type>
+    <default_value>1</default_value>
+    <group>mach_pes_last</group>
+    <file>env_mach_pes.xml</file>
+    <desc>total number of cores used (setup automatically - DO NOT EDIT)</desc>
+  </entry>
+
   <entry id="MAX_TASKS_PER_NODE">
     <type>integer</type>
     <default_value>0</default_value>
--- a/utils/python/CIME/case.py
+++ b/utils/python/CIME/case.py
@@ -119,16 +119,19 @@ class Case(object):
         """
         env_mach_pes = self.get_env("mach_pes")
         comp_classes = self.get_values("COMP_CLASSES")
-        total_tasks = env_mach_pes.get_total_tasks(comp_classes)
+        total_tasks  = env_mach_pes.get_total_tasks(comp_classes)
+        pes_per_node = self.get_value("PES_PER_NODE")
 
         self.thread_count = env_mach_pes.get_max_thread_count(comp_classes)
         self.tasks_per_node = env_mach_pes.get_tasks_per_node(total_tasks, self.thread_count)
         logger.debug("total_tasks %s thread_count %s"%(total_tasks, self.thread_count))
         self.num_nodes = env_mach_pes.get_total_nodes(total_tasks, self.thread_count)
         self.tasks_per_numa = int(math.ceil(self.tasks_per_node / 2.0))
-        smt_factor = max(1,int(self.get_value("MAX_TASKS_PER_NODE")/self.get_value("PES_PER_NODE")))
+        smt_factor = max(1,int(self.get_value("MAX_TASKS_PER_NODE") / pes_per_node))
 
-        self.cores_per_task = self.thread_count / smt_factor
+        threads_per_node = self.tasks_per_node * self.thread_count
+        threads_per_core = 1 if (threads_per_node < pes_per_node) else smt_factor
+        self.cores_per_task = self.thread_count / threads_per_core
 
         return total_tasks
 
--- a/utils/python/CIME/case.py
+++ b/utils/python/CIME/case.py
@@ -130,7 +130,7 @@ class Case(object):
         smt_factor = max(1,int(self.get_value("MAX_TASKS_PER_NODE") / pes_per_node))
 
         threads_per_node = self.tasks_per_node * self.thread_count
-        threads_per_core = 1 if (threads_per_node < pes_per_node) else smt_factor
+        threads_per_core = 1 if (threads_per_node <= pes_per_node) else smt_factor
         self.cores_per_task = self.thread_count / threads_per_core
 
         return total_tasks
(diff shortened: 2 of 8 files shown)


driver_cpl/cime_config/config_component.xml
```diff
@@ -1967,6 +1967,14 @@
     <desc>total number of tasks and threads (setup automatically - DO NOT EDIT)</desc>
   </entry>
 
+  <entry id="TOTAL_CORES">
```
jgfouca: I think this is useful enough to add it as an entry.

jedwards4b: okay

utils/python/CIME/case_setup.py
```diff
@@ -190,6 +190,7 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False):
                    "it is not possible to run with OpenMP if using the NAG Fortran compiler")
             cost_pes = env_mach_pes.get_cost_pes(pestot, thread_count, machine=case.get_value("MACH"))
             case.set_value("COST_PES", cost_pes)
+            case.set_value("TOTAL_CORES", pestot * thread_count)
```
jedwards4b: I think that this is incorrect in some cases.   On SMT enabled systems we put two or four threads on a single core.    That's why I use the ratio PES_PER_NODE/MAX_TASKS_PER_NODE

jgfouca: Hmm, so what's the right thing to do here? I want test_scheduler to manage each thread as taking a core on melvin.

jgfouca: Should I set MAX_TASKS_PER_NODE or PES_PER_NODE to 1?

jedwards4b: No MAX_TASKS_PER_NODE=PES_PER_NODE in that case.

jgfouca: @jedwards4b melvin's current config is for PES_PER_NODE=64 and MAX_TASKS_PER_NODE=64, the behavior I'm seeing for COST_PES is still rounding up to 64. I'm wondering how to support your SMT case and the basic desktop case of 1-thread per core.

jedwards4b:
I'm not saying that you can't add the new field TOTAL_CORES and use it.  I'm just saying that your calculation for TOTAL_CORES is wrong in cases where MAX_TASKS_PER_NODE > PES_PER_NODE and pestot >= PES_PER_NODE.

utils/python/CIME/case.py
```diff
@@ -128,13 +128,10 @@ def initialize_derived_attributes(self):
         self.tasks_per_numa = int(math.ceil(self.tasks_per_node / 2.0))
         smt_factor = max(1,int(self.get_value("MAX_TASKS_PER_NODE")/self.get_value("PES_PER_NODE")))
 
-        self.cores_per_task = ((self.get_value("MAX_TASKS_PER_NODE")/smt_factor) \
-                               / self.tasks_per_node) * 2
+        self.cores_per_task = self.thread_count / smt_factor
```
jgfouca: @jedwards4b please take a close look at this. I was getting very odd results on melvin with the previous implementation of self.cores_per_task. The prior impl seemed to assume that it had to use every core for a problem. For example, for a 16x2 case on melvin (64 MAX_TASK, 64 PPN), the resulting cores_per_task was 8. This is not correct, at least not for what I intended, because the results TOTAL_CORES came out to 128.

utils/python/CIME/case.py
```diff
@@ -783,6 +782,8 @@ def configure(self, compset_name, grid_name, machine_name=None,
                 if pio_typename in ("pnetcdf", "netcdf4p"):
                     self.set_value(key, "netcdf")
 
+        # Set TOTAL_CORES
+        self.set_value("TOTAL_CORES", total_tasks * self.cores_per_task )
```
jgfouca: TOTAL_CORES set here.

utils/python/CIME/case.py
rljacob: Does this possibly change the pe-layout for create_newcase cases?

jedwards4b: This is currently only used on cori.   The new calculation makes a lot more sense for the variable name than the old one did and I think nersc has suggested that we don't need this field on cori anyway.  I'm testing cori (both knl and haswell) now with that field removed and will report results tomorrow.

jedwards4b: This flag is not needed for cori - all tests pass without it.  But the new computation in line 131 is still incorrect - ```cores_per_task = 1 if (pestot < PES_PER_NODE) else smt_factor```

utils/python/CIME/case.py
```diff
@@ -119,22 +119,22 @@ def initialize_derived_attributes(self):
...
         self.tasks_per_node = env_mach_pes.get_tasks_per_node(total_tasks, self.thread_count)
         logger.debug("total_tasks %s thread_count %s"%(total_tasks, self.thread_count))
         self.num_nodes = env_mach_pes.get_total_nodes(total_tasks, self.thread_count)
         self.tasks_per_numa = int(math.ceil(self.tasks_per_node / 2.0))
-        smt_factor = max(1,int(self.get_value("MAX_TASKS_PER_NODE")/self.get_value("PES_PER_NODE")))
+        smt_factor = max(1,int(self.get_value("MAX_TASKS_PER_NODE") / pes_per_node))
 
-        self.cores_per_task = ((self.get_value("MAX_TASKS_PER_NODE")/smt_factor) \
-                               / self.tasks_per_node) * 2
+        threads_per_node = self.tasks_per_node * self.thread_count
```
jgfouca: @jedwards4b OK, here's the latest

PR #1028, opened 2017-01-19
Science support field

### LINKED ISSUE #533: Want ability to distinguish compset aliases that should only be used for testing
CSEG - and particularly the CAM group ( @cacraigucar and others) - have asked for this feature. I am documenting it as an issue based on today's CSEG meeting.

Basically, we want a compset attribute that distinguishes whether a compset is scientifically supported / functionally supported / only available for testing.

Earlier idea was to have a test_alias (rather than alias for this).

But idea from today's discussion is to have a separate support_level attribute. Possible values are:
- Scientifically supported
- Functionally supported
- Testing only
- (And then there are user_compsets that don't match an alias, which is another level: "You're on your own").

Then, when you ask for a list of possible compsets, you'd by default only get a list of scientifically supported and functionally supported... but you could ask to see testing-only as well. (And I think the discussion is that you'd need to explicitly ask if you want to create a case with one of these testing-only compsets using create_newcase.)

It was also noted that scientific support is tied to grid. Maybe need an attribute describing that.

Tagging @mvertens @jedwards4b @mnlevy1981 , who are the ones who I remember contributing most to this discussion. Also Sheri Mickelson, but I can't find her github name.

jedwards4b:
Add a science_support field to the config_compsets.xml file.  This is implemented for CESM only and should not affect non cesm models in any way.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #533 

User interface changes?:   Added --use-unsupported flag to create_newcase for cesm only.

Code review:

### DIFF
--- a/utils/python/CIME/utils.py
+++ b/utils/python/CIME/utils.py
@@ -508,6 +508,7 @@ def get_timestamp(timestamp_format="%Y%m%d_%H%M%S", utc_time=False):
 def get_project(machobj=None):
     """
     Hierarchy for choosing PROJECT:
+    0. Command line flag to create_newcase or create_test
     1. Environment variable PROJECT
     2  Environment variable ACCOUNT  (this is for backward compatibility)
     3. File $HOME/.cime/config       (this is new)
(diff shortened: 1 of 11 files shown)


utils/python/CIME/utils.py
```diff
@@ -508,6 +508,7 @@ def get_timestamp(timestamp_format="%Y%m%d_%H%M%S", utc_time=False):
 def get_project(machobj=None):
     """
     Hierarchy for choosing PROJECT:
+    0. Command line flag to create_newcase or create_test
```
rljacob: Was this change supposed to be part of this commit?

jedwards4b: It's not relevant to this commit, but you aren't going to penalize me for improving documentation are you?

PR #1036, opened 2017-01-21
Fix submission of ERS and ERR tests.

### LINKED ISSUE #985: ERI test cases cannot be re-run
```
./create_test ERI.f45_g37.X
* wait till completes *
cd $casedir
./case.submit
* wait till fails *
```

Error:  ERROR: (seq_infodata_Init) :: rpointer file read returns an error condition

jgfouca:
The level of complexity of the ERI test is probably too much
to support resubmission.

Test suite: Z_FullSystemTest
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #985 

User interface changes?: None

Code review: @jedwards4b

### DIFF
--- a/cime_config/config_tests.xml
+++ b/cime_config/config_tests.xml
@@ -201,6 +201,10 @@ LII    CLM initial condition interpolation test
     <INFO_DBUG>1</INFO_DBUG>
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>11</STOP_N>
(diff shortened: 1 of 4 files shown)


jgfouca:
@jedwards4b I felt that ERI is far too complex to support resubmission. It would be a lot of work for little value. Fixing ERS and ERR was easy so I went ahead and did that instead.

PR #1039, opened 2017-01-24
Lock env_mach.xml file

jgfouca:
If user changes walltime or queue, this will not affect the batch
directives in case.run or case.test unless the user cleans and
re-runs case.setup. We've gotten complaints from users that their
WALLTIME changes are not taking affect. We've updated the ACME
config_batch settings to have these params be args to submit instead
of directives.

This branch also does some work to encapsulate the lockfile
concept better inside check_lockedfiles.py.

Test suite: scripts_regression_tests.py and by-hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

User interface changes?: env_batch.xml is now locked by case.setup

Code review: @jedwards4b

### DIFF
--- a/utils/python/CIME/case_setup.py
+++ b/utils/python/CIME/case_setup.py
@@ -98,8 +98,6 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False):
                 os.remove("case.testdriver")
                 logger.info("Successfully cleaned test script case.testdriver")
 
-        unlock_file("env_batch.xml")
-
         logger.info("Successfully cleaned batch script case.run")
 
         msg = "case.setup clean complete"
@@ -178,9 +176,6 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False):
                 elif job != "case.test":
                     logger.info("Writing %s script from input template %s" % (job, input_batch_script))
                     env_batch.make_batch_script(input_batch_script, job, case, pestot, tasks_per_node, num_nodes, thread_count)
-
-            # Lock env_batch
-            lock_file("env_batch.xml")
 
             # Make sure pio settings are consistant
             for comp in models:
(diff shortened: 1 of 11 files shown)


jgfouca:
This fixes an issue reported by @worleyph  on the ACME side.

jedwards4b:
I'm confused by this one because WALLTIME and QUEUE are arguments to the submit (qsub, squeue, etc) itself and no longer used in the batch directives - I wonder if some of the acme machines did not get changed in that way and thus caused this issue to be raised.   This change is not needed if the walltime and queue are configured in this way.   Can we reconsider this PR?

utils/python/CIME/case_setup.py
```diff
@@ -208,6 +179,9 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False):
                     logger.info("Writing %s script from input template %s" % (job, input_batch_script))
                     env_batch.make_batch_script(input_batch_script, job, case, pestot, tasks_per_node, num_nodes, thread_count)
 
+            # Lock env_batch
+            lock_file("env_batch.xml")
```
jedwards4b: Not all changes in env_batch.xml require a case.reset

jgfouca:
@jedwards4b yeah, it would be better if we didn't have to lock it. I will take a closer look at what's in env_batch.xml vs what's in config_batch.xml

jgfouca:
@jedwards4b I've updated the branch based on your recommendations. I hate to add a burden onto our users and that's what locking env_batch.xml would do. On the other hand, this opens the door for case.run and case.test not getting updated when they need to be. So we're sacrificing some safety for convenience here. I changed all the main batch definitions in acme, so hopefully this won't come up.

jedwards4b:
I think that a future change will be to identify the items in env_batch that require case.run and case.test to need updated - the locked files mechanism is too gross.

PR #1055, opened 2017-01-26
Add access to shell in XML files

### LINKED ISSUE #1040: Support some interface to the shell in our XML
Users like @amametjanov are asking for access to the shell within our XML files. Something like:

```
<env name="NETCDF_PATH">$SHELL{which nc-config | xargs dirname | xargs dirname}</env>
```

jgfouca:
Use /bin/bash{...}

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1040 

User interface changes?: $SHELL{...} is now recognized in our XML parsing system

Code review: @jedwards4b

### DIFF
--- a/utils/python/CIME/XML/generic_xml.py
+++ b/utils/python/CIME/XML/generic_xml.py
@@ -7,6 +7,8 @@ from distutils.spawn import find_executable
 from xml.dom import minidom
 from CIME.utils import expect, get_cime_root
 
+import getpass
+
 logger = logging.getLogger(__name__)
 
 class GenericXML(object):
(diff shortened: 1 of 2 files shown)


jgfouca:
Testing still running.

utils/python/CIME/XML/generic_xml.py
```diff
@@ -230,11 +240,9 @@ def get_resolved_value(self, raw_value):
             elif var == "SRCROOT":
                 srcroot = os.path.join(get_cime_root(),"..")
                 item_data = item_data.replace(m.group(), srcroot)
-            elif var in os.environ:
```
jgfouca: Significant change, please look at closely. I didn't like how we silently fell back to using the environment even if the user did not do the correct $ENV syntax.

utils/python/CIME/XML/generic_xml.py
```diff
@@ -230,11 +240,9 @@ def get_resolved_value(self, raw_value):
...
-                # this is a list of suppressed warnings (things normally expected to be resolved in env)
-                if var not in ("USER",):
-                    logging.debug("Resolved from env: " + var)
-                item_data = item_data.replace(m.group(), os.environ[var])
+            elif var == "USER":
+                item_data = item_data.replace(m.group(), getpass.getuser())
+
```
jgfouca: Should there be an error or warning if we were unable to resolve a reference?

utils/python/CIME/XML/generic_xml.py
jedwards4b: There may be unintended consequences here, but I can't think of any specific cases right now.

utils/python/CIME/XML/generic_xml.py
jedwards4b: We had something like that for a while and removed it because it was too noisy.   Maybe we've cleaned things up enough that we can put it back.

PR #1065, opened 2017-01-28
Merge the maint-cime5.1 branch including ACME 5.1 changes

### LINKED ISSUE #834: cime5.1 not generating correct aprun command on titan.
Originally reported by @worleyph.  Reproducing here to get more ideas.

Using a standard ACME layout for a fully coupled case with a mix of threaded and non-threaded, and some components stacked and some concurrent, the aprun command is wrong:

 run command is aprun  -d 2  -cc numa_node   -j 2  -S 8  -N 8  -n 1192 xxx/acme.exe  >> acme.log.xxx 2>&1  

This looks like an old Edison run command, or for the Intel compiler? We don't need to set -j or -cc for PGI - the defaults are sufficient (though setting them consistently is okay also). The error that causes the failure is -S 8 -N 8 -d 2. When using -d 2 (for 2 OpenMP threads), -S should be 4.

The second error is that the aprun command does not have separate -d, -N, -S for the different components on nonoverlapping nodes, which is an important performance optimization for Titan.

### LINKED ISSUE #1105: Adding a wind gustiness parameterization
The merge of PR #1065 proposed to introduce a gustiness parameterization.   This issue discusses how its currently done and how best to do it.

In module **seq_flux_mct**:  
In seq_flux_init_mct:  allocate prec_gust and set it to 0.

In seq_flux_atmocnexch_mct:  define gust_fac, set it to HUGE(1.0) and then overwrite with seq_infodata_GetData.   pass prec_gust and gust_fac to shr_flux_atmocn;  

In seq_flux_atmocn_mct:  set prec_gust to a2x%rAttr(index_a2x_Faxa_rainc,n);  define gust_fac, set it to HUGE(1.0) and then overwrite with seq_infodata_GetData (from drv_in).   pass prec_gust and gust_fac to shr_flux_atmocn

In module **seq_infodata_mod**
Add gust_fac to seq_infodata_type

In  seq_infodata_Init, declare gust_fac and add to /seq_infodata_inparm/ namelist block.   Set initial value to huge(1.0_SHR_KIND_R8), read value from seq_infodata_inparm namelist; copy value to infodata%gust_fac, 

In seq_infodata_GetData_explicit:    If present, gust_fac = infodata%gust_fac
In seq_infodata_PutData_explicit:  if present,    infodata%gust_fac  = gust_fac
In seq_infodata_bcast:  broadcast infodata%gust_fac
In seq_infodata_print:  print out gust_fac value.

In module **shr_flux_mod**
In shr_flux_atmOcn:  Add gust_fac and prec_gust to arguments.  
Define new function:
ugust(gprec) = gust_fac*log(1._R8+57801.6_R8*gprec-3.55332096e7_R8*(gprec**2.0_R8))
Alter vmag calculation to include gustiness:
vmag   = max(umin, sqrt( (ubot(n)-us(n))**2 + (vbot(n)-vs(n))**2) +  ugust(min(prec_gust(n),6.94444e-4_R8)))

jgfouca:
Merge the maint-cime5.1 branch along with ACME's changes to 5.1

XML changes:
* ACME compset additions
* ACME PEs changes
* ACME grids
* ACME build system updates (Depends files, Makefiles)
* ACME config changes (batch, compilers, machines)
* Minor modifications to config_component to split out entry ids that are not the same between ACME and CESM
* Lots of additions to config_component_acme

CPL/DRV changes:
* Addition of gust_fac into driver
* Addition of cron_script, cron equivalent of jenkins_script

Python changes:
* Remove obsolete ACME env_mach_specific files
* If model is ACME, auto-add compiler to baseline name (impacts create_test, bless_test_results, and compare_test_results)
* Fix HOMME test
* Updates to ACME performance archiving
* Raise exception if batch job-id fails to match pattern
* Bug fix for point calculation in pes.py
* Changes to build.py because clm is not a shared lib in ACME
* ACME now supports recommended test timings based on test-suite

Test baseline:  none
Test namelist changes: 
Test status: bit for bit

Closes #1105 
Fixes #834 

User interface changes?: See above

Code review: Mariana Vertenstein, Jim Edwards, Steve Goldhaber, Rob Jacob


jgfouca:
This is my first crack at this. Annotations and furthers commits to come.

jgfouca:
For some reason it's not letting me ask for a review from @mvertens or even mention her?

components/data_comps/datm/cime_config/config_component.xml
```diff
@@ -54,6 +54,7 @@
       <value compset="^RCP4_">rcp4.5</value>
       <value compset="^RCP2_">rcp2.6</value>
       <value compset="^HIST_">trans_1850-2000</value>
+      <value compset="^20TR_">trans_1850-2000</value>
```
jedwards4b: This is putting back a name (20TR) that we chose to replace with HIST.   I don't think we should do this - just use HIST.

driver_cpl/cime_config/buildnml
```diff
@@ -55,6 +55,57 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
     elif case.get_value('RUN_TYPE') == 'branch':
         config['run_type'] = 'branch'
 
+    cimeroot = case.get_value("CIMEROOT")
```
jedwards4b: I don't think that this code should be here, it's functionality has been moved in 5.2

driver_cpl/driver/seq_flux_mct.F90
```diff
@@ -1433,7 +1443,7 @@ subroutine seq_flux_atmocn_mct(infodata, tod, dt, a2x, o2x, xao)
                           !duu10n,ustar, re  , ssq, missval = 0.0_r8 )
                           cold_start=cold_start)
     else
-       call shr_flux_atmocn (nloc , zbot , ubot, vbot, thbot, &
+       call shr_flux_atmocn (nloc , zbot , ubot, vbot, thbot, prec_gust, gust_fac, &
```
jedwards4b: This may have an unintended affect on cesm atm and ocn components?

components/data_comps/datm/cime_config/config_component.xml
mvertens: This comment really applies to all changes in the data components config_componetn.xml files
- rather than having two lines that are repeated - it would be cleaner to have and entry of the form 
<value compset="^HIST_|^20TR_">trans_1850-2000</value>
However, I think this needs to be done in a separate PR to clean this up. Once this PR is accepted can we make sure to open a separate issue to cleanup duplicates like this?

share/csm_share/shr/shr_orb_mod.F90
```diff
@@ -78,8 +78,10 @@ real(SHR_KIND_R8) pure FUNCTION shr_orb_cosz(jday,lat,lon,declin,dt_avg)
       shr_orb_cosz =  shr_orb_avg_cosz(jday, lat, lon, declin, dt_avg)
    else
       shr_orb_cosz = sin(lat)*sin(declin) - &
-           cos(lat)*cos(declin) * &
-           cos((jday-floor(jday))*2.0_SHR_KIND_R8*pi + lon)
+      &              cos(lat)*cos(declin)*cos(jday*2.0_SHR_KIND_R8*pi + lon)
```
jedwards4b: I understand that you want to revert this for your testing, but it's gonna mess our testing up.   Can we maybe wrap this with an if(acme) conditional?

driver_cpl/bld/namelist_files/namelist_defaults_drv.xml
```diff
@@ -4,6 +4,275 @@
 
 <namelist_defaults>
 
+<!-- =========================================  -->
```
mvertens: This file is no longer in driver_cpl. It uses the old namelist generation. These entries need to be merged with the new driver_cpl/cime_config/namelist_default_drv.xml.  This is necessary for this PR to be accepted.

driver_cpl/bld/namelist_files/namelist_definition_drv.xml
```diff
@@ -2,7 +2,2782 @@
 
 <namelist_definition>
 
-  <version>1.0</version>
+<!-- Each namelist variable is defined in an <entry> element.  The
```
mvertens: This file is no longer in driver_cpl and should not be introduced as part of this PR. This file was part of the old namelist generation. These entries need to be merged with the new driver_cpl/cime_config/namelist_defaults_drv.xml which uses the new python nmlgen class and methods. The new namelist_defaults_drv.xml has both default values as well as definition values in one file.
Many of these settings should be the same in the new file.

share/esmf_wrf_timemgr/ESMF_TimeMod.F90
```diff
@@ -1080,16 +1080,20 @@ function ESMF_TimeDiff(time1, time2)
...
-         cmptime%Sn = 0
-         cmptime%Sd = 0
-         ESMF_TimeDiff%basetime = cmptime
-      else
+      !
+      ! Update: Actually, the timeintcmp() routine in ESMF_TimeIntervalMod.F90 is not capable
+      !    of comparing time intervals when one interval has a different year and month than
+      !    the other. So, it is best here to always compute the interval as day/seconds.
+
+!      if (d1 == d2 .and. s1 == s2) then
```
jedwards4b: It seems like the code has been commented out but no prevision has been put into place to handle the case it was intended for.

utils/python/CIME/build.py
```diff
@@ -237,6 +240,9 @@ def case_build(caseroot, case, sharedlib_only=False, model_only=False):
     os.environ["CISM_USE_TRILINOS"]    = stringify_bool(cism_use_trilinos)
     os.environ["MPASLI_USE_ALBANY"]    = stringify_bool(mpasli_use_albany)
 
+    if get_model() == "acme" and mach == "titan" and compiler == "pgiacc":
```
jedwards4b: This should be in xml not in code.

driver_cpl/cime_config/config_component_acme.xml
```diff
@@ -463,13 +644,22 @@
...
   </entry>
 
-  <entry id="TFREEZE_SALTWATER_OPTION">
-    <type>char</type>
-    <valid_values>minus1p8,linear_salt,mushy</valid_values>
-    <default_value>minus1p8</default_value>
-    <group>run_physics</group>
-    <file>env_run.xml</file>
-    <desc>Freezing point calculation for salt water.</desc>
+  <entry id="CCSM_GCOST">
```
mvertens: I thought these two went away

driver_cpl/cime_config/config_component_acme.xml
```diff
@@ -352,80 +498,114 @@
...
+    <file>env_case.xml</file>
+    <desc>If set to off, this component set/ grid specification is not scientifically supported.
+    If set to on, this component set/ grid specification is scientifically supported</desc>
   </entry>
 
-  <entry id="FLDS_WISO">
-    <type>logical</type>
-    <valid_values>TRUE,FALSE</valid_values>
-    <default_value>FALSE</default_value>
+  <entry id="CCSM_CCOST">
```
mvertens: I thought this entry went away.

share/csm_share/shr/shr_flux_mod.F90
```diff
@@ -267,7 +277,15 @@ real(R8)    :: psix2  ! stability function at ztref reference height
      if (mask(n) /= 0) then
     
         !--- compute some needed quantities ---
-        vmag   = max(umin, sqrt( (ubot(n)-us(n))**2 + (vbot(n)-vs(n))**2) )
+
```
mvertens: **This is a critical problem here.** CESM does not use these variables (ugust and gprec) and these changes will change a very fundamental calculation for CESM. These need to be brought into this routine in a way that permits backwards compatibility with the old calculation.

jedwards4b: It looks like the flux calculation they are used in is backward compatible, if vmag is 0 it reverts to the original formula.

share/csm_share/shr/shr_orb_mod.F90
mvertens: We need to have the current CIME routine in here and not the one brought in from ACME. As in PJR comment's it seems that the newer version (i.e. the one in the current CIME) should have been brought into AME in a separate PR. Not sure how you want to proceed but there should not be a change in shr_orb_mod.F90 in this PR.

share/esmf_wrf_timemgr/ESMF_TimeMod.F90
```diff
@@ -1080,16 +1080,20 @@ function ESMF_TimeDiff(time1, time2)
...
-      if (d1 == d2 .and. s1 == s2) then
-!         write(6,*) 'tcx timedifft ym'
-         ESMF_TimeDiff%YR = y1 - y2
-         ESMF_TimeDiff%MM = m1 - m2
-         cmptime%S  = 0
-         cmptime%Sn = 0
-         cmptime%Sd = 0
-         ESMF_TimeDiff%basetime = cmptime
-      else
+      !
```
mvertens: I am concerned about bringing this change in. We need to get a review from Tony who originally implemented this as to what effect this will have in CESM.

utils/python/CIME/SystemTests/homme.py
```diff
@@ -62,3 +61,12 @@ def run_phase(self):
             run_cmd_no_fail("%s -j 4 check >& %s" % (gmake, log), from_dir=exeroot)
         else:
             run_cmd_no_fail("%s -j 4 baseline >& %s" % (gmake, log), from_dir=exeroot)
+
```
mvertens: It would be useful to feedback from Steve on this routine.

mvertens:
It gives the right result only because ugust is zero. It would be very easy
for this to change and CESM to suddenly have a different calculation in a
very critical part of the code. I have proposed a very simple approach to
separate out what acme and cesm each want to do. But maybe its not the best
one.

gold2718:
@mvertens, homme.py tests is ACME-only as they decided to keep the separate HOMME standalone testing environment.

jgfouca:
Tests are now working on melvin.

driver_cpl/bld/namelist_files/namelist_defaults_drv.xml
rljacob: So backwards compatibility was broken?

cime_config/acme/allactive/config_pes.xml.cime2
```diff
@@ -454,6 +450,15 @@
     <PES_LEVEL>1r</PES_LEVEL>
 </pes>
 
+<pes OS="BGP">
```
rljacob: Notice the file has .cime2 on the end.  Here just for historical reference and not used/read.

cime_config/acme/allactive/config_pes.xml.cime2
```diff
@@ -469,6 +474,7 @@
 <NTASKS_GLC>2048</NTASKS_GLC> <NTHRDS_GLC>16</NTHRDS_GLC> <ROOTPE_GLC>0</ROOTPE_GLC> <NINST_GLC>1</NINST_GLC>
 <NTASKS_WAV>2048</NTASKS_WAV> <NTHRDS_WAV>16</NTHRDS_WAV> <ROOTPE_WAV>0</ROOTPE_WAV> <NINST_WAV>1</NINST_WAV>
 <NTASKS_CPL>2048</NTASKS_CPL> <NTHRDS_CPL>16</NTHRDS_CPL> <ROOTPE_CPL>0</ROOTPE_CPL>
+<PIO_NUMTASKS>-1</PIO_NUMTASKS><PIO_STRIDE>64</PIO_STRIDE> <PIO_TYPENAME>pnetcdf</PIO_TYPENAME> <PIO_ROOT>1</PIO_ROOT>
```
rljacob: Ignore any file ending in "cime2".  I kept those around in case I needed to look at them.

components/data_comps/datm/cime_config/config_component.xml
rljacob: We'll have to keep both until ACME makes that change as well.

share/csm_share/shr/shr_orb_mod.F90
rljacob: yes that's fine.

driver_cpl/driver/seq_flux_mct.F90
rljacob: yeah it might.

share/esmf_wrf_timemgr/ESMF_TimeMod.F90
rljacob: Not sure why git is showing a diff here.  We made no mods to this file since cime5.1.

utils/python/CIME/build.py
rljacob: Ultimately yes but we needed a way, at create_newcase time, to set the CAM source path based on machine and compiler.   We didn't want to add another create_newcase argument.

share/csm_share/shr/shr_orb_mod.F90
rljacob: As JimE suggested, just put an if(acme) around the old version.

mvertens:
A completely new namelist generation was put in place that is much more
robust and extensible for all for the data models and the driver. If you
can summarize what changes were put in place for the old
namelist_definition.xml and namelist_defaults.xml for the driver since acme
cime5.0 I am happy to help add these in.

share/csm_share/shr/shr_flux_mod.F90
mvertens: Assuming that the flux calculation will stay backwards compatible is not acceptable in my view. We need an if-block to ensure that if ugust is not zero then we do not get unintended consequences for CESM.

share/csm_share/shr/shr_orb_mod.F90
mvertens: That was my suggestion for shr_flux_mod - and I agree that it should be applied in shr_orb_mod also.

share/csm_share/shr/shr_flux_mod.F90
mvertens: Just as an additional note: It gives the right result only because ugust is zero. It would be very easy for this to change in the future and for CESM to suddenly have a different calculation in a very critical part of the code. I have proposed a very simple approach to
separate out what acme and cesm each want to do. But maybe its not the best
one.

rljacob: Perhaps @singhbalwinder can refresh my memory on this change.

rljacob:
What PR changes those 2 namelist files?   I'll have to compare ACME's version with those.

driver_cpl/bld/namelist_files/namelist_defaults_drv.xml
mvertens: Yes. The driver and data components namelists are now generated with the python nmlgen class which is much more robust and understandable than the previous build-namelist. As part of this we introduced a new schema combining namelist definition and namelists defaults into one file  in cime_config/namelist_definition_drv.xml. If you determine what differences there were between the acme cime5.0 and namelist_defaults.xml and namelists_definition.xml and what you are trying to merge in now - I am happy to help incorporate those into the new namelist_definition_drv.xml. My guess as that there will only be minor changes that need to be brought in.

utils/python/CIME/test_scheduler.py
```diff
@@ -574,13 +574,22 @@ def _run_catch_exceptions(self, test, phase, run):
             return False
 
     ###########################################################################
-    def _get_procs_needed(self, test, phase, no_batch=False):
+    def _get_procs_needed(self, test, phase, threads_in_flight=None, no_batch=False):
```
gold2718: Is the problem which caused this to be turned off resolved?

utils/python/CIME/build.py
jedwards4b: I believe that this can go into config_machines.xml  - without any new create_newcase arguments...

rljacob:
@mvertens ACME made this addition to namelist_defaults_drv.xml:
```
 <flux_diurnal>.false.</flux_diurnal>
+<gust_fac>0.0D0</gust_fac>
```

And this addtion to namelist_definition.xml
```
+<entry>
+id="gust_fac"
+type="real"
+category="control"
+group="seq_infodata_inparm">
+wind gustiness factor
+default: 0.0D0
+</entry>
```

mvertens:
@rjacob - all that is needed is to add the following entry to
cime_config/config_definition_drv.xml

  <entry id="gust_fac">
    <type>real</type>
    <category>control</category>
    <group>seq_infodata_inparm</group>
    <desc>wind gustiness factor</desc>
    <values>
      <value>0.0D0</default>
    </values>
  </entry>

You do not need to change build-namelist - since this will now
automatically get picked up in generating the driver namelist. Let me know
if you have any problem with implementing this change.

utils/python/CIME/test_scheduler.py
jgfouca: Yes

cime_config/acme/machines/config_batch.xml
```diff
@@ -44,8 +44,8 @@
        <arg flag="--cwd" name="CASEROOT"/>
        <arg flag="-A" name="PROJECT"/>
        <arg flag="-t" name="JOB_WALLCLOCK_TIME"/>
-       <!-- space is required here so that variable is resolved correctly -->
-       <arg flag="-n" name=" $TOTALPES / $PES_PER_NODE"/>
+       <!-- ceiling of $a/$b is ($a+$b-1)/$b -->
+       <arg flag="-n" name=" ($TOTALPES+$MAX_TASKS_PER_NODE-1)/$MAX_TASKS_PER_NODE"/>
```
jgfouca: Done.

driver_cpl/bld/namelist_files/namelist_defaults_drv.xml
jgfouca: @mvertens , I don't see this file: driver_cpl/cime_config/namelist_defaults_drv.xml. Did you mean driver_cpl/cime_config/namelist_definition_drv.xml?

jgfouca: @rljacob We did not make these changes on the ACME side. This is another file that we deleted on the ACME side but somehow got resurrected.

rljacob: You mean deleted on the ESMCI side?  Yes that's true.  As @mvertens said elsewhere, we just need to add the one thing we did change to the new file.

driver_cpl/driver/seq_flux_mct.F90
jgfouca: Solution?

utils/python/CIME/SystemTests/homme.py
jgfouca: @mvertens did you have something in mind? This code is specific to the ACME HOMME test hack. I don't CESM cares about this change.

share/csm_share/shr/shr_flux_mod.F90
jgfouca: Solution?

rljacob: Leave it as is.   This is guaranteed to be backwards compatible because gust_fac is 0 and math works.   Code does not spontaneously change.

share/csm_share/shr/shr_orb_mod.F90
rljacob: Just use the ESMCI version.  ACME will change to it eventually.

share/esmf_wrf_timemgr/ESMF_TimeMod.F90
rljacob: This change was made by @douglasjacobsen on the maint-cime5.1 branch.  See commit fddc0d05bb50

rljacob: @apcraig what do you think of this?

share/csm_share/shr/shr_flux_mod.F90
mvertens: First - from examining the code carefully, this is actually not correct. It  ugust is in fact zero - but prec_gust is NOT (see line 1397 in seq_flux_mct.F90)
             prec_gust (n) = a2x%rAttr(index_a2x_Faxa_rainc,n)

This is then used in shr_flux_mct.F90 in the following

       !--- reverts to original formula if gust_fac=0 
        vmag   = max(umin, sqrt( (ubot(n)-us(n))**2 + (vbot(n)-vs(n))**2) +  ugust(min(prec_gust(n),6.94444e-4_R8)))

From this, acme is introducing new science in a non-backwards compatible manner. So we cannot leave this code as it is.

Second,  uGust and prec_gust are not namelist parameters - they are currently hard-wired in seq_flux_mct.F90 to zero. Furthermore, this line is commented out - which could easily be uncommented by a user 
seq_flux_mct.F90:!           !!uGust(n) = 1.5_r8*sqrt(uocn(n)**2 + vocn(n)**2) ! there is no wind gust data from ocn

What needs to be done in seq_flux_mct.F90 is the following:

if (cime_model == 'cesm') then
     uGust(n) = 0._r8
     prec_gust(n) = 0._r8
else
!    !!uGust(n) = 1.5_r8*sqrt(uocn(n)**2 + vocn(n)**2) ! there is no wind gust data from ocn 
     prec_gust(n) = a2x%rAttr(index_a2x_Faxa_rainc,n)
     uGust(n) = 0.0_r8
end if

**However, my strong preference would be to pass cime_model to shr_flux_mct.F90
and have the following lines appear in shr_flux_mct.F90**

        !--- vmag+ugust (convective gustiness) Limit to a max precip 6 cm/day = 0.00069444 mm/s.
        !--- reverts to original formula if gust_fac=0 
if (trim(cime_model) == 'acme') then
        vmag   = max(umin, sqrt( (ubot(n)-us(n))**2 + (vbot(n)-vs(n))**2) +  ugust(min(prec_gust(n),6.94444e-4_R8)))
else
        !--- vmag+ugust (convective gustiness) Limit to a max precip 6 cm/day = 0.00069444 mm/s.
        !--- reverts to original formula if gust_fac=0 
        vmag   = max(umin, sqrt( (ubot(n)-us(n))**2 + (vbot(n)-vs(n))**2)
end if

This would be much cleaner and would clarify how those arguments are used.
This would also eliminate having to introduce multiple if-blocks in seq_flux_mct.F90 to set ugust and prec_gust to zero depending on acme or cesm.

gold2718: Every other science option is bracketed by logic, I do not see why this one is different. If folks do not like having model names (e.g., cesm & acme), then come up with a namelist parameter which indicates the scientific reason for the change and use that.

rljacob: This is getting a little long for the PR review so I'll move it to a separate issue. #1105

mvertens:
Thanks. I think that is a good idea.

jgfouca:
@rljacob unforunately, I'm pretty useless when it comes to the fortran driver/coupler code, so I'm going to hand this branch to you. I think that all the python and XML has been correctly merged.

mvertens:
I think that for the purposes of this PR all that needs to be done is:

1) Change line 486 of seq_infodata_mod.F90

       gust_fac  = huge(1.0_SHR_KIND_R8)   <==== current
       gust_fac  = 0._r8 <==== new

at line 486 of seq_infodata_mod.F90.

2) Change line 910 of seq_flux_mct.F90:

    real(r8)    :: gust_fac = huge(1.0_r8) !wind gust factor    <====
current
    real(r8)    :: gust_fac     <==== new

since you are obtaining gust_fac from seq_infodata in line 910 of
seq_flux_mct.F90.

    call seq_infodata_GetData(infodata, &
         read_restart=read_restart, &
         dead_comps=dead_comps,         &
         atm_nx=atm_nx, atm_ny=atm_ny,  &
         ocn_nx=ocn_nx, ocn_ny=ocn_ny,  &
         ocn_prognostic=ocn_prognostic, &
         flux_diurnal=flux_diurnal,     &
         gust_fac = gust_fac            )

Moving forwards, we will be getting input from CESM scientists as to how
they want to see this science change implemented to satisfy CESM
requirements. However, that can be put into a upcoming PR.

Mariana

components/data_comps/datm/cime_config/config_component.xml
rljacob: Sure thats ok.

driver_cpl/cime_config/config_component_acme.xml
rljacob: We still had this and CCSM_GCOST in the 5.1 branch we started with.  I'm not sure if they are used for anything and fine with removing them.

share/csm_share/shr/shr_orb_mod.F90
rljacob: We'll revert this change in this branch.

share/csm_share/shr/shr_flux_mod.F90
rljacob: Conclusion from issue #1105 was to leave as is and fix it later with a new PR.

rljacob:
With the conclusion in issue #1105 , everything in this PR has been agreed to except for the changes to ESMF_TimeMod.F90 by @douglasjacobsen .   What should we do about those?

rljacob:
A test merge to master shows only 3 files with conflicts:
```
	both modified:   cime_config/acme/machines/config_compilers.xml
	both modified:   cime_config/acme/machines/config_machines.xml
	both modified:   driver_cpl/shr/seq_infodata_mod.F90
```

The first 2 are pretty easy.  In seq_infodata_mod.F90, git is confused by the gust_fac addition and the ESP additions.  I accepted the "HEAD" version and then had to put declarations of gust_fac back in a few routines.

So if we can decide on the ESMF_TimeMod changes, this is ready to go.

mvertens:
@rjacob - I don't know enough about ESMF_TimeMod.F90 to determine if this
will effect CESM or not. Why was this change brought in? Is it really
needed in this PR?
I am really uncomfortable in bringing in an change to esmf_wrf_timemgr
without input from Tony - who is currently the main expert on this.

rljacob:
Here's the commit:  https://github.com/ESMCI/cime/commit/fddc0d05bb5084a8250b4ebcc57dff666f102fd8

Doug found this while doing some testing with MPAS I think.

mvertens:
Have we confirmed via testing Doug's assertion that "These changes
shouldn't affect intervals that previously worked, and

should only allow year and month intervals to be compared."

mvertens:
Tony has just confirmed that he thinks this change is okay. So I think this
PR is now good to go.

rljacob:
This merge was undone by a force push to the previous version of master.  Turns out github was right and there were to many commits/files in the original subtree split from ACME.  Replaced by PR #1123

PR #1069, opened 2017-01-28
updating pop defaults for cesm

jtruesdal:
Updating CESM defaults.  Changed map defaults for EBM and set coupling frequency to one hour/robert filtering for the ocean. Only grids gx1v6 and gx1v7 have the 1hr coupling interval for now.  As other resolutions are tested they will be updated.  

Test suite: Ran cime regression tests and pylint check.  When running my CESM pop tests I had one error in an ERR test that I traced to a recent cime mod.  Forwarded to Jim Edwards and he has a fix and PR in hand.
Test baseline: 
Test namelist changes: 
Test status: [climate changing]  New pop defaults, Estuary parameterization, and new maps give a different climate.

Fixes [CIME Github issue #]

User interface changes?: 

Code review: Mariana worked with me to create these cime mods.  Jim is handling ERR fix via a different PR.

### DIFF
--- a/cime_config/cesm/config_grids.xml
+++ b/cime_config/cesm/config_grids.xml
@@ -1221,16 +1221,16 @@
     </gridmap>
 
     <gridmap rof_grid="rx1" ocn_grid="gx3v7" >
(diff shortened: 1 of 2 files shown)


jtruesdal:
Forgot to add that these commits are CESM specific and don't affect ACME.

PR #1071, opened 2017-01-29
NOT A PR:  the pure ACME changes to CIME5.1

rljacob:
This branch was pushed at put up as a PR just to help with PR #1065 

The branch rljacob/acmemerge-to-cime51 was created by starting with a  recent clone of ACME and in the top level doing:

git subtree split --prefix cime/ --ignore-joins -b rljacob/acmemerge-to-cime51 

ACME started from cime5.1.4.   To see what changed, checkout the branch rljacob/acmemerge_to_cime51 and do
git diff cime5.1.4  "filename"


rljacob:
Note that the cime5.1.4 tag was made on the maint branch. cime5.1.1 was on master and could also be used for comparison.

jgfouca:
@rljacob we have major problems here. There is no way we made this many changes to CIME in the last couple months. If you look at the file:

cime/components/data_comps/datm/bld/namelist_files/namelist_defaults_datm.xml

In the ACME repo and look at commits touching this file, the only thing that's happened is that you removed it on Nov 13th. But according to this PR, we've made a bunch of mods.

components/data_comps/datm/bld/namelist_files/namelist_defaults_datm.xml
```diff
@@ -633,6 +645,83 @@ Currently the following streams are supported (term definitions precede the stre
   PSRF     pbot
 </strm_datvar>
 
+<!-- =============== -->
```
jgfouca: I am quite sure we did not make these changes on the ACME side.

rljacob:
I don't think we can trust what github displays on these subtrees that have been split out.   Best thing to do is checkout the branch and do "git diff cime5.1.1" on each file.

jgfouca:
You're right, the file I was concerned about doesn't even exist on the merge branch. How can the CESG folks possibly review our merges when GitHub is showing hundreds of thousands of lines of phantom diffs?

rljacob:
Well this is not a PR for review.  But I see its not even helping to have it up as a fake PR.  So I'll close it.

PR #1080, opened 2017-01-31
Config compilers conformity

### LINKED ISSUE #1078: Be consistent when accessing environment
config_compilers uses

```
<env>VARNAME</env>
```

everything else uses

```
$ENV{VARNAME}
```

config_compilers should be changed to be consistent with the other files.

jedwards4b:
Add support for $ENV{} $SHELL{} and ${} in config_compilers.xml to conform to usage in other
xml files.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1078 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/XML/compilerblock.py
+++ b/utils/python/CIME/XML/compilerblock.py
@@ -117,32 +117,41 @@ class CompilerBlock(object):
         output = elem.text
         if output is None:
             output = ""
-
+        logger.debug("Initial output=%s"%output)
+        #Brackets are manditory in this file
         reference_re = re.compile(r'\${?(\w+)}?')
         env_ref_re   = re.compile(r'\$ENV\{(\w+)\}')
         shell_ref_re = re.compile(r'\$SHELL\{([^}]+)\}')
 
+        for m in reference_re.finditer(output):
+            var_name = m.groups()[0]
+            if var_name not in ("SHELL","ENV"):
+                output = output.replace(m.group(), writer.variable_string(var_name))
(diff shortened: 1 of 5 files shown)


jedwards4b:
Don't merge this quite yet...

utils/python/CIME/XML/compilerblock.py
```diff
@@ -115,6 +117,35 @@ def _handle_references(self, elem, set_up, tear_down, depends):
         output = elem.text
         if output is None:
             output = ""
+
+        reference_re = re.compile(r'\${?(\w+)}?')
```
jgfouca: Why can't we reuse the code from generic_xml that already handles this kind of stuff?

jedwards4b: Because it doesn't handle it in quite the same way - 
In generic $ENV{NETCDF} is resolved at run time.   In compilerblock it is converted to make or cmake syntax but not resolved until used.

jedwards4b:
I've updated all the tests in scripts_regression_tests to test this new syntax and punted on the one I couldn't fix.

jgfouca:
Ready to merge?

jedwards4b:
Yes, if you're okay with it.   I see that you reverted the acme config_compilers.xml are you going to revert it again or shall i revert the cesm one?  (I prefer they use the same syntax whichever we choose)

jgfouca:
@jedwards4b please update both config_compilers.xml files.

Issue #1038, opened 2017-01-24
request for standardized return status codes reported in CaseStatus


bertinia:
It would be helpful if the return status codes for each reported operation in CaseStatus
conformed to an agreed upon nomenclature. 

For example,  we could use the following return status codes in a dictionary:

id   code              description
0    Succeeded   process successfully completed
1     Pending        process not yet started
2    Started          process has started            
3    Shutdown     process was shutdown           
4    Error              process error

And then each process that writes to CaseStatus could use the following message
template:

_status_dict = {
0 : 'Succeeded',
1 : 'Pending',
2 : 'Started',
3 : 'Shutdown',
4 : 'Error'
}
from string import Template
_status_template = Template($process: status id = $id; status code = _status_dict[$id])

msg = _status_template.substitute(process='case.run', id=$returncode)

The global variables, _status_dict and _status_template, could reside in the case.py class or
as a separate import file.

jedwards4b:
I think that we already do this to some extent - statuses are defined in CIME/test_status.py.   What problem is it you are trying to solve?

bertinia:
In archive_metadata, I need to report back to the expdb what the current status is of the case.
I'm currently parsing the CaseStatus file looking for the following strings:

# status strings to search for in CaseStatus                                                                                                                      
_run_msgs = ['Model did not complete','Run started','Run FAILED','Run SUCCESSFUL'];
_sta_msgs = ['st_archiving starting','st_archiving completed']
_lta_msgs = ['lt_archive starting','lt_archive completed']

And then translating those strings into corresponding lookup values in the database.
It would be nice to be able to search and parse a single format for status information.

jedwards4b:
thanks for the clarification.   This should be easy to do since all updates to the CaseStatus file go through the append_status routine in CIME/utils.py

PR #1090, opened 2017-02-02
fix issues with lt_archive

### LINKED ISSUE #1089: case.lt_archive
lt_archive script has several problems preventing functionality.

jedwards4b:
Several issues prevented lt_archive from functioning

Test suite: hand run of lt_archive (no tests available in suite)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1089 

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/cesm/machines/config_lt_archive.xml
+++ b/cime_config/cesm/machines/config_lt_archive.xml
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 
-<config_lt_archive version="1.0.0">
+<config_lt_archive version="1.0">
     <machine name="default">
       <lt_archive_args></lt_archive_args>
     </machine>
--- a/utils/python/CIME/case_lt_archive.py
+++ b/utils/python/CIME/case_lt_archive.py
@@ -12,17 +12,16 @@ def case_lt_archive(case):
     caseroot = case.get_value("CASEROOT")
 
     # max number of threads needed by scripts
-    os.environ["maxthrds"] = 1
+    os.environ["maxthrds"] = "1"
 
     # document start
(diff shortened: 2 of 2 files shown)


cime_config/cesm/machines/config_lt_archive.xml
```diff
@@ -1,6 +1,6 @@
 <?xml version="1.0"?>
 
-<config_lt_archive version="1.0.0">
+<config_lt_archive version="1.0">
```
jedwards4b: version handling in cime is implemented as real and so 1.0.0 doesn't work here.  We can revisit this issue if there is strong objection and alternative to the current implementation.

utils/python/CIME/case_lt_archive.py
```diff
@@ -12,17 +12,16 @@ def case_lt_archive(case):
...
 
     # max number of threads needed by scripts
-    os.environ["maxthrds"] = 1
+    os.environ["maxthrds"] = "1"
 
     # document start
     append_status("lt_archive starting",caseroot=caseroot,sfile="CaseStatus")
 
     # determine status of run and short term archiving
     runComplete = does_file_have_string(os.path.join(caseroot, "CaseStatus"),
```
jedwards4b: I'm really concerned about these checks - they are difficult to maintain because status strings they are testing for may change.  But also a single successful run in the case will make these true for all future runs and so they don't really provide anything useful for a long running case - I suggest they be removed and we find another method.

bertinia: Is this related to #1038?  I'll go ahead and approve this PR.

PR #1092, opened 2017-02-02
Add ESMF_TimeIntervalProdI8 routine

mgduda:
This PR adds a new routine, ESMF_TimeIntervalProdI8, for multiplying an ESMF_TimeInterval by an 8-byte integer.

To complement the existing routine ESMF_TimeIntervalProdI that implements the `*` operator for multiplying an ESMF_TimeInterval by a 4-byte integer, this PR adds a new routine, ESMF_TimeIntervalProdI8, that multiplies an ESMF_TimeInterval by an 8-byte integer. This new routine is accessed through the '*' operator in the same way as the existing ESMF_TimeIntervalProdI routine. This routine is useful when multiplying small intervals (e.g., 1 second) by large integers (e.g., the number of seconds in 1000 years).

Test suite: scripts_regression_tests.py passes with two errors related to ACME; according to @jedwards4b these failures are not problematic.
Test baseline: 
Test namelist changes: 
Test status: bit-for-bit -- only a new routine has been added

User interface changes: The `operator(*)` interface for multiplying an `ESMF_TimeInterval` by an integer now supports multiplication by either 4-byte or 8-byte integers.

Code review:

### DIFF
--- a/share/esmf_wrf_timemgr/ESMF_TimeIntervalMod.F90
+++ b/share/esmf_wrf_timemgr/ESMF_TimeIntervalMod.F90
@@ -103,6 +103,7 @@ module ESMF_TimeIntervalMod
 
       public operator(*)
       private ESMF_TimeIntervalProdI
+      private ESMF_TimeIntervalProdI8
 
 ! Inherited and overloaded from ESMF_BaseTime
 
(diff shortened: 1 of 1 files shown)


mgduda:
I think there a few open questions related to this PR:
1. Should the type of the integer argument to the `ESMF_TimeIntervalProdI` routine (i.e., the existing 4-byte multiplication routine) be changed to have an explicit kind of `ESMF_KIND_I4` to avoid an ambiguous interface if the ESMF time manager were to be compiled with default 8-byte integers?
2. Should we create an 8-byte version of the `ESMF_TimeIntervalQuotI` routine for symmetry?

jedwards4b:
I'll leave this up for a bit for others to review in particular @apcraig may be interested.

PR #1095, opened 2017-02-02
Add pause/resume to driver

### LINKED ISSUE #973: Data ESP model
To follow my own new proposed workflow rules, this issue is about a new data ESP (External System Processing) model which is part of the Pause/Resume project. The main purpose of this model is to test pause-resume functionality in the driver (which is also being implemented as part of this issue).

gold2718:
Add pause/resume signaling to driver
Create data ESP model (DESP) which tests pause/resume functionality
Create PRE test script which runs pause/resume tests
Create ADESP and ADESP_TEST compsets to test pause/resume with and without data modification (respectively).

Test suite: scripts_regression_test.py
Test baseline: NA
Test namelist changes: NA (for now)
Test status: bit for bit

Fixes #973

User interface changes?: Minor doc (--help) change to create_test

Code review: In progress

### DIFF
--- a/cime_config/config_tests.xml
+++ b/cime_config/config_tests.xml
@@ -260,6 +260,16 @@ LII    CLM initial condition interpolation test
     <DOUT_S>FALSE</DOUT_S>
   </test>
 
+  <test NAME="PRE">
+    <DESC>pause/resume test, default 5 hours, five pause/resume cycles, no data modification</DESC>
+    <INFO_DBUG>1</INFO_DBUG>
+    <STOP_OPTION>nhours</STOP_OPTION>
+    <STOP_N>5</STOP_N>
+    <HIST_OPTION>nhours</HIST_OPTION>
+    <HIST_N>1</HIST_N>
+    <DOUT_S>FALSE</DOUT_S>
+  </test>
+
   <test NAME="TESTBUILDFAIL">
     <DESC>For testing infra only. Insta-fail build step.</DESC>
     <INFO_DBUG>1</INFO_DBUG>
--- a/components/data_comps/desp/cime_config/namelist_definition_desp.xml
+++ b/components/data_comps/desp/cime_config/namelist_definition_desp.xml
@@ -1,0 +1,78 @@
+<?xml version="1.0"?>
+
+<?xml-stylesheet type="text/xsl" href="http://www.cgd.ucar.edu/~cam/namelist/namelist_definition.xsl"?>
+
+<entry_id version="2.0">
+
+  <!-- Each namelist variable is defined in an <entry> element.  The
+     content of the element is the documentation of how the variable is
+     used.  Other elements  are:
+     id
+          The variable's name.  *** N.B. *** The name must be lower case.
+          The module convert all namelist variable names to lower case
+          since Fortran is case insensitive.
+     type
+          An abbreviation of the fortran declaration for the variable.
+	  Valid declarations are:
+             char, integer, logical, real
+	  Any of these types may be followed by a comma separated list of
+	  integers enclosed in parenthesis to indicate an array.
+	  The current namelist validation code only distinquishes between
+	  string and non-string types.
+     category
+          A category assigned for organizing the documentation.
+     group
+          The namelist group that the variable is declared in.
+     valid_values
+          This is an optional attribute that is mainly useful for variables
+          that have only a small number of allowed values.
+  -->
+
+  <!-- =========================== -->
+  <!--   group desp_inparm         -->
+  <!-- =========================== -->
+
+  <entry id="desp_mode" modify_via_xml="DESP_MODE">
+    <type>char</type>
+    <category>desp</category>
+    <group>desp_nml</group>
+    <valid_values>NOCHANGE,DATATEST</valid_values>
+    <desc>
+      The mode of operation for the DESP model.
+      NOCHANGE: report  status of model 'pause' but make no changes
+      DATATEST: make a roundoff change  to the restart files of the
+                components in the 'PAUSE_COMPONENT_LIST' XML variable.
+      Default: NOCHANGE
+    </desc>
+    <values>
+      <value>$DESP_MODE</value>
+    </values>
+  </entry>
+
+  <entry id="info_debug">
+    <type>integer</type>
+    <category>desp</category>
+    <group>desp_nml</group>
+    <desc>
+      Level of debug output, 0=minimum, 1=normal, 2=more, 3=too much (default: 0)
+    </desc>
+    <values>
+      <value>0</value>
+    </values>
+  </entry>
+
+  <entry id="restfilm">
+    <type>char</type>
+    <category>desp</category>
+    <group>desp_nml</group>
+    <desc>
+      Model restart filename for the external system (ESP) model data.  This is
+      optional.  If restfilm is undefined, the restart filename will be
+      read from the DESP restart pointer file (or files for multiple instances).
+    </desc>
+    <values>
+      <value>undefined</value>
+    </values>
+  </entry>
+
+</entry_id>
--- a/components/data_comps/desp/desp_comp_mod.F90
+++ b/components/data_comps/desp/desp_comp_mod.F90
@@ -1,6 +1,3 @@
-#ifdef AIX
-@PROCESS ALIAS_SIZE(805306368)
-#endif
 module desp_comp_mod
 
 ! !USES:
--- a/components/data_comps/desp/desp_comp_mod.F90
+++ b/components/data_comps/desp/desp_comp_mod.F90
@@ -74,8 +74,9 @@ module desp_comp_mod
   integer,          parameter :: BAD_ID      = -1
   integer,          parameter :: NO_RPOINTER = -2
   integer,          parameter :: NO_RFILE    = -3
-  integer,          parameter :: NO_READ     = -4
-  integer,          parameter :: NO_WRITE    = -5
+  integer,          parameter :: NO_RPREAD   = -4
+  integer,          parameter :: NO_READ     = -5
+  integer,          parameter :: NO_WRITE    = -6
 
   type(shr_strdata_type)      :: SDESP
 
@@ -494,6 +495,8 @@ CONTAINS
         select case (errcode)
         case(NO_RPOINTER)
           call shr_sys_abort(subname//'Missing rpointer file for '//comp_names(ind))
+        case(NO_RPREAD)
+          call shr_sys_abort(subname//'Cannot read rpointer file for '//comp_names(ind))
         case(NO_RFILE)
           call shr_sys_abort(subname//'Missing restart file for '//comp_names(ind))
         case(NO_READ)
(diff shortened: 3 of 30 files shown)


cime_config/config_tests.xml
```diff
@@ -260,6 +260,16 @@ LII    CLM initial condition interpolation test
...
+  <test NAME="PRE">
+    <DESC>pause/resume test, default 5 hours, five pause/resume cycles, no data modification</DESC>
+    <INFO_DBUG>1</INFO_DBUG>
+    <STOP_OPTION>nhours</STOP_OPTION>
+    <STOP_N>5</STOP_N>
+    <HIST_OPTION>nhours</HIST_OPTION>
+    <HIST_N>1</HIST_N>
+    <DOUT_S>FALSE</DOUT_S>
+  </test>
+
```
gold2718: This test is not currently tested (neither is the newish DAE test). I propose adding these to cime_developer in update_acme_tests.py:
DAE.f19_f19.A
PRE.f19_f19_ADESP
PRE.f19_f19.ADESP_TEST

If that is too many, I can leave off the last one.

driver_cpl/cime_config/buildnml
```diff
@@ -5,7 +5,7 @@
 # pylint: disable=invalid-name
 
 # Disable these because this is our standard setup
-# pylint: disable=wildcard-import,unused-wildcard-import,wrong-import-position
+# pylint: disable=wildcard-import,unused-wildcard-import
```
gold2718: pylint complains that wrong-import-position is a bad option

driver_cpl/cime_config/namelist_definition_drv.xml
```diff
@@ -223,30 +223,6 @@
...
-    <group>seq_infodata_inparm</group>
-    <desc>
-      true => turn on ideal physics mode in cam
-    </desc>
-    <values>
-      <value>.false.</value>
-    </values>
-  </entry>
-
   <entry id="aqua_planet">
```
gold2718: CAM no longer looks at these since it is not the driver should not be involved in internal atmosphere model settings. 
This is also already fixed in the ACME atmosphere model.

driver_cpl/cime_config/user_nl_cpl
```diff
@@ -10,36 +10,10 @@
...
-! ice_rootpe     => ROOTPE_ICE	      ocn_rootpe   => ROOTPE_OCN
-! ice_pestride   => PSTRID_ICE	      ocn_pestride => PSTRID_OCN
-! ice_layout     => NINST_ICE_LAYOUT  ocn_layout   => NINST_OCN_LAYOUT
-! glc_ntasks     => NTASKS_GLC        rof_ntasks   => NTASKS_ROF
-! glc_nthreads   => NTHRDS_GLC	      rof_nthreads => NTHRDS_ROF
-! glc_rootpe     => ROOTPE_GLC	      rof_rootpe   => ROOTPE_ROF
-! glc_pestride   => PSTRID_GLC	      rof_pestride => PSTRID_ROF
-! glc_layout     => NINST_GLC_LAYOUT  rof_layout   => NINST_ROF_LAYOUT
+! For example, rather than set username to 'foo' in user_nl_cpl, call
+! ./xmlchange USER=foo
```
gold2718: This was already out of date so I substituted a more generic documentation.

driver_cpl/driver/cesm_comp_mod.F90
```diff
@@ -3679,6 +3723,33 @@ subroutine cesm_run()
...
+        end if
+        if (iamin_CPLID) then
+          call seq_rest_read(drv_resume, infodata,                            &
+               atm, lnd, ice, ocn, rof, glc, wav, esp,                        &
+               fractions_ax, fractions_lx, fractions_ix, fractions_ox,        &
+               fractions_rx, fractions_gx, fractions_wx)
+        end if
+      end if
+
+      !----------------------------------------------------------
```
gold2718: ESP needs to run after any model (including the coupler) has written a restart (pause) file.

driver_cpl/shr/seq_infodata_mod.F90
```diff
@@ -60,6 +63,27 @@ MODULE seq_infodata_mod
...
+      character(SHR_KIND_CL) :: atm_resume(num_inst_atm) = ' ' ! atm resume file
+      character(SHR_KIND_CL) :: lnd_resume(num_inst_lnd) = ' ' ! lnd resume file
+      character(SHR_KIND_CL) :: ice_resume(num_inst_ice) = ' ' ! ice resume file
+      character(SHR_KIND_CL) :: ocn_resume(num_inst_ocn) = ' ' ! ocn resume file
+      character(SHR_KIND_CL) :: glc_resume(num_inst_glc) = ' ' ! glc resume file
+      character(SHR_KIND_CL) :: rof_resume(num_inst_rof) = ' ' ! rof resume file
+      character(SHR_KIND_CL) :: wav_resume(num_inst_wav) = ' ' ! wav resume file
+      character(SHR_KIND_CL) :: cpl_resume = ' '               ! cpl resume file
+   end type seq_pause_resume_type
+
```
gold2718: Creating a sub-type for pause/resume means that all that data does not have to be allocated for other runs.

utils/python/CIME/XML/namelist_definition.py
```diff
@@ -139,8 +139,7 @@ def get_per_stream_entries(self):
     # good for that purpose anyway, so stop this function from being called.
     def set_value(self, vid, value, subgroup=None, ignore_type=True):
         """This function is not implemented."""
-        raise TypeError, \
-            "NamelistDefinition does not support `set_value`."
+        raise TypeError("NamelistDefinition does not support `set_value`.")
```
gold2718: pylint fix

components/data_comps/desp/cime_config/namelist_definition_desp.xml
```diff
@@ -0,0 +1,78 @@
...
+    <type>char</type>
+    <category>desp</category>
+    <group>desp_nml</group>
+    <desc>
+      Model restart filename for the external system (ESP) model data.  This is
+      optional.  If restfilm is undefined, the restart filename will be
+      read from the DESP restart pointer file (or files for multiple instances).
+    </desc>
+    <values>
+      <value>undefined</value>
```
jedwards4b: Do we need undefined here or would <value/> be sufficient?

components/data_comps/desp/desp_comp_mod.F90
```diff
@@ -0,0 +1,720 @@
+#ifdef AIX
```
jedwards4b: No more AIX machines and if there are any in the future they probably don't need this.

components/data_comps/desp/esp_utils.F90
```diff
@@ -0,0 +1,344 @@
...
+
+! !USES:
+
+  use shr_kind_mod, only: r8=>shr_kind_r8, CL=>SHR_KIND_CL
+  use shr_sys_mod,  only: shr_sys_abort, shr_sys_flush
+
+  implicit none
+  private
+#include <mpif.h>
+  save
```
jedwards4b: I don't think we want blanket save statements in modules and shouldn't we use mpi instead of include <mpif.h>?

driver_cpl/cime_config/buildnml
jedwards4b: which pylint is that?   Seems different pylint versions support different options and I'm at a bit of a loss regarding how to handle that.

driver_cpl/cime_config/user_nl_cpl
jedwards4b: We've tried to encode this interaction in xml now.

driver_cpl/driver/cesm_comp_mod.F90
```diff
@@ -2050,13 +2056,16 @@ end subroutine cesm_init
  !===============================================================================
 
  subroutine cesm_run()
-   use seq_comm_mct, only : atm_layout, lnd_layout, ice_layout, glc_layout, rof_layout, &
+   use seq_comm_mct,   only: atm_layout, lnd_layout, ice_layout, glc_layout, rof_layout, &
          ocn_layout, wav_layout, esp_layout
+   use shr_string_mod, only: shr_string_listGetIndexF
```
jedwards4b: Why remove implicit none here?

components/data_comps/desp/cime_config/namelist_definition_desp.xml
gold2718: Or would _what_ be sufficient?

gold2718: As for pylint, I think I learned my lesson which is only run pylint on yellowstone. I didn't realize how different a slightly newer version of pylint can be. We need to agree on a version with the DOE folks (if they can internally agree) and stick with that.

driver_cpl/driver/cesm_comp_mod.F90
gold2718: I don't understand your comment "We've tried to encode this interaction in xml now". My thought was to point anyone looking at user_nl_cpl to the correct interface. Do you have a better suggestion?
An implicit none at module scope applies to "applies to it, its module subprograms, and their internal subprograms" so the other ones are redundant. I removed all of them.

components/data_comps/desp/cime_config/namelist_definition_desp.xml
jedwards4b: Would an empty field be sufficient ```<value></value>```

driver_cpl/cime_config/user_nl_cpl
jedwards4b: In the namelist_definition.xml files we have explicitly indicated the variables which are updated via xml rather than via user_nl_* files for example:
``` <entry  id="flds_wiso" modify_via_xml="FLDS_WISO">``` 
if you try to modify flds_wiso in user_nl the namelist generation step will give you an error informing you that you must use FLDS_WISO in xml to modify this field.

cime_config/config_tests.xml
jedwards4b: How about if you put the first two in cime_developer and the last one in cesm prealpha?
Can you go ahead and add those to this PR?

driver_cpl/cime_config/user_nl_cpl
gold2718: I understand but what about users who do not look at xml files (because we tell them to use xmlquery and xmlchange) but do look at user_nl_<comp> files (because they are supposed to use those files)? Instruction via error message doesn't work for everyone. This is just a way to inform them what is going on. I think it is a helpful reinforcement (maybe it will cut down on a support call or two).

jedwards4b: If a user attempts to put this field into a user_nl_ file a nice, informative error is printed to stdout telling them they need to change the related XML variable instead.  This doesn't require users to look at xml files - I'm not criticizing your change here only trying to inform you of the new method which has been implemented.  If you have namelist variables in this pr which should only be change via changing a related xml variable then you should add this field to your namelist_definition xml file.

components/data_comps/desp/cime_config/namelist_definition_desp.xml
gold2718: You tell me, which is correct? This is how restart filenames are set up for datm, dice, dlnd, docn, drof and dwav.

driver_cpl/cime_config/user_nl_cpl
gold2718: Glad we are in agreement. I agree that the build_namelist error message is fine.

PR #1097, opened 2017-02-03
Tie timers to xml

### LINKED ISSUE #770: Timers not on by default
This issue is a twofer...

* By default, none of the internal timers in CESM are on, so POP outputs
```
Timing information:
 
Timer number  1 Time =       0.00 seconds   COMPUTE_COSZ
  Timer stats (node): min =        0.00 seconds
                      max =        0.00 seconds
                      mean=        0.00 seconds
  Timer stats(block): min =        0.00 seconds
                      max =        0.00 seconds
                      mean=        0.00 seconds
Timer number  2 Time =       0.00 seconds   ECOSYS_INTERIOR
  Timer stats (node): min =        0.00 seconds
                      max =        0.00 seconds
                      mean=        0.00 seconds
  Timer stats(block): min =        0.00 seconds
                      max =        0.00 seconds
                      mean=        0.00 seconds
```
and so on.
* @jedwards4b mentioned increasing the timer detail level (from default of 0 to new value of 4). This did produce timer output, but it is unclear exactly how to make this change. I initially thought I needed
```
$ ./xmlchange TIMER_DETAIL=4
```
but we actually needed to edit `user_nl_cpl` with
```
profile_detail_limit=4
```
So I think the coupler's `build-namelist` needs to be updated to use the XML variable.

### LINKED ISSUE #1084: Harmonize machine options for create_ commands
The arguments to specify a machine are different for create_newcae and create_test.  They should have one in common  (allowing for backward compatibility).

In create_test:
```
  -m MACHINE, --machine MACHINE
                        The machine for which to build tests, this machine
                        must be defined in the config_machines.xml file for
                        the given model. Default is to match the name of the
                        machine in the test name or the name of the machine
                        this script is run on to the NODENAME_REGEX field in
                        config_machines.xml. This option is highly unsafe and
                        should only be used if you know what you're doing.
                        (default: None)
```

In create_newcase:

```
  --mach MACH, -mach MACH
                        Specify a machine. default: match NODENAME_REGEX in
                        config_machines.xml To see list of current machines,
                        use the utility manage_case in this directory
                        (default: None)
```

### LINKED ISSUE #1096: timer defaults in cime
With the following code in namelist_definition_drv.xml it would seem that profile_detail_limit can never be set to a value > 1 without using user_nl_cpl.   (TIMER_LEVEL > 1 means timer_level='pos') 
A number of cesm users have requested that this variable have a different default (4) so that some timers are always on by default.  However acme wants 0 to be the default.   These values look backward to me - I think we want 0 when timer_level='neg' and TIMER_LEVEL when 'pos'  and then 
set TIMER_LEVEL=0 in the xml for acme.  Do you agree?
```
 <entry id="profile_detail_limit">                                                 
    <type>integer</type>                                                            
    <category>performance</category>                                                
    <group>prof_inparm</group>                                                      
    <desc>                                                                          
    </desc>                                                                         
    <values>                                                                        
      <value timer_level='neg'>$TIMER_LEVEL</value>                                 
      <value timer_level='pos'>0</value>                                            
    </values>                                                                       
  </entry>            
```

jedwards4b:
The settings for timers in xml TIMER_LEVEL and TIMER_DETAIL were a bit disconnected from their counterparts in namelist, this is an attempt to straighten out the problems.  Unrelated update netcdf on cheyenne.   

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1096 
Fixes #1084 
Fixes #770

User interface changes?:  create_newcase --mach is now --machine to match create_test 

Code review:

### DIFF
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -144,7 +144,7 @@ OR
         run_unsupported = args.run_unsupported
 
 
-    return args.case, args.compset, args.res, args.mach, args.compiler,\
+    return args.case, args.compset, args.res, args.machine, args.compiler,\
         args.mpilib, args.project, args.pecount, \
         args.user_mods_dir, args.user_compset, args.pesfile, \
         args.user_grid, args.gridfile, args.srcroot, args.test, args.ninst, \
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -12,7 +12,7 @@ def parse_command_line(args, cimeroot, description):
 ###############################################################################
     help_str = \
 """
-%s --case [CASE] --compset [COMPSET] --res [GRID] [--mach ...] [--compiler ...]
+%s --case [CASE] --compset [COMPSET] --res [GRID] [--machine ...] [--compiler ...]
 OR
 %s --help
 """ % ((os.path.basename(args[0]), ) * 2)
@@ -38,7 +38,7 @@ OR
                         help="(required) Specify a model grid resolution. "
                         "To see list of current compsets, use the utility manage_case in this directory")
 
-    parser.add_argument("--mach", "-mach",
+    parser.add_argument("--machine", "-mach",
                         help="Specify a machine. default: match NODENAME_REGEX in config_machines.xml "
                         "To see list of current  machines, use the utility manage_case in this directory"
                         )
--- a/cime_config/cesm/machines/config_machines.xml
+++ b/cime_config/cesm/machines/config_machines.xml
@@ -249,11 +249,10 @@
 	<command name="load">ncarcompilers/0.3.5</command>
       </modules>
       <modules mpilib="mpi-serial">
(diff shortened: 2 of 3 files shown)


scripts/create_newcase
```diff
@@ -12,7 +12,7 @@ def parse_command_line(args, cimeroot, description):
 ###############################################################################
     help_str = \
 """
-%s --case [CASE] --compset [COMPSET] --res [GRID] [--mach ...] [--compiler ...]
+%s --case [CASE] --compset [COMPSET] --res [GRID] [--machine ...] [--compiler ...]
```
rljacob: I think we want to keep --mach for backwards compatibility.  Just don't put it in the help.  Also this should be a separate PR.

jedwards4b: No we kept -mach for backward compatibility, I don't think that --mach has had much use.

PR #1098, opened 2017-02-03
Mfdeakin sandia/map field build

### LINKED ISSUE #956: Can't get map_field to build
I'm trying to build the map_field tool (in tools/mapping/map_field), and can't get it to build. I have an old version sitting around which I think I can use for now, so this isn't critical, but just something that should be fixed before the release.

I tried following something similar to the instructions in the INSTALL file, but adapted to what appears to be the current interface of the configure tool, but the build failed.

I ran:

```
cd tools/mapping/map_field/src
export CIME_MODEL=cesm
../../../configure --machine yellowstone --macros-format Makefile
. .env_mach_specific.sh
gmake
```

and got this output:

```
f77 -c   -I/glade/apps/opt/netcdf/4.3.3.1/intel/15.0.3/include -I/glade/apps/opt/netcdf/4.3.3.1/intel/15.0.3/include map_field.F90
f77: map_field.F90: linker input file unused because linking not done
f77 -o ../map_field map_field.o
f77: map_field.o: No such file or directory
Makefile:131: recipe for target '../map_field' failed
gmake: *** [../map_field] Error 1
```

mfdeakin-sandia:
Moves "CPPDEFS += -D${OS} -DCPRCOMPILER" to Make.macros. Also ensures that this is added for linux machines. This fixes #956.
Also fixes a typo in cime_config/acme/machines/config_machine.xml so Linux is properly cased.

Test suite: Most of scripts_regression_tests, particularly B_Check_Code
Test status: BFB

User interface changes?: 
Fixes #956 
Code review:

### DIFF
--- a/cime_config/acme/machines/Makefile
+++ b/cime_config/acme/machines/Makefile
@@ -78,7 +78,7 @@ endif
 # set CPP options (must use this before any flags or cflags settings)
 #===============================================================================
 
-CPPDEFS := $(USER_CPPDEFS) -DCPR$(shell echo $(COMPILER) | tr a-z A-z)
+CPPDEFS := $(USER_CPPDEFS)
 
 # Unless DEBUG mode is enabled, use NDEBUG to turn off assert statements.
 ifneq ($(strip $(DEBUG)),TRUE)
--- a/cime_config/acme/machines/Makefile
+++ b/cime_config/acme/machines/Makefile
@@ -78,7 +78,7 @@ endif
 # set CPP options (must use this before any flags or cflags settings)
 #===============================================================================
 
-CPPDEFS := $(USER_CPPDEFS) -D$(OS) -DCPR$(shell echo $(COMPILER) | tr a-z A-z)
+CPPDEFS := $(USER_CPPDEFS) -DCPR$(shell echo $(COMPILER) | tr a-z A-z)
 
 # Unless DEBUG mode is enabled, use NDEBUG to turn off assert statements.
 ifneq ($(strip $(DEBUG)),TRUE)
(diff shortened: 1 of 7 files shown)


mfdeakin-sandia:
I would really appreciate further testing of this, as I don't have access to yellowstone, so my tests for it were hacks.

cime_config/acme/machines/Makefile
```diff
@@ -78,7 +78,7 @@ endif
 # set CPP options (must use this before any flags or cflags settings)
 #===============================================================================
 
-CPPDEFS := $(USER_CPPDEFS) -D$(OS) -DCPR$(shell echo $(COMPILER) | tr a-z A-z)
+CPPDEFS := $(USER_CPPDEFS) -DCPR$(shell echo $(COMPILER) | tr a-z A-z)
```
jedwards4b: I think that the COMPILER part should be moved as well.

cime_config/acme/machines/config_compilers.xml
```diff
@@ -790,6 +790,9 @@ for mct, etc.
 </compiler>
 
 <compiler OS="Linux" COMPILER="intel">
+  <CPPDEFS>
```
jedwards4b: Rather than hard coding it this way we should use the OS and COMPILER variables.

mfdeakin-sandia: Attempting to do that gets me a circular reference error for both of those variables

cime_config/acme/machines/Makefile
mfdeakin-sandia: See my comment below

cime_config/acme/machines/config_compilers.xml
jedwards4b: I've pushed an update that solves the problem

PR #1104, opened 2017-02-03
only count tests in prealpha prebeta or aux_* categories

### LINKED ISSUE #1093: Don't count all tests when determining if a compset is unsupported
In the determination of whether a compset is listed as unsupported, currently all tests are treated as equal, and if any test exists for it, it is no longer "unsupported".  This causes issues with tests that are set up for development purposes, but are not ready for public use.  I would propose that the counting of tests only count tests from a subset of all the possible test names.  

For example, in CAM's case, we run "aux_cam" before we make a CAM tag.   Then "prealpha" and "prebeta" are also run prior to various CESM tags.  Any other compsets are not routinely tested and could cause issues for an end-user and would thus be identified as "unsupported".  

The list of supported tests (in CAM's case: aux_cam, prealpha, prebeta)  could be maintained in each component's cime_config directory to leave its definition under our control.

jedwards4b:
Limits the test count to prealpha, prebeta and aux_* test categories.  If you want to test a
compset but don't want it listed as tested call it something that doesn't match any of these prefixes. 

Test suite: hand tested 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1093 

User interface changes?: 

Code review:

### DIFF
--- a/utils/python/CIME/case.py
+++ b/utils/python/CIME/case.py
@@ -1121,4 +1121,5 @@ class Case(object):
             logger.info("\nThis compset and grid combination is not scientifically supported, however it is used in %d tests.\n"%(testcnt))
         else:
             expect(False, "\nThis compset and grid combination is unsupported in CESM.  "
-                   "If you wish to use it anyway you must supply the --run-unsupported option to create_newcase.")
+                   "If you wish to use it anyway you must supply the --run-unsupported option to create_newcase.",
+                   error_prefix="STOP: ")
--- a/utils/python/CIME/case.py
+++ b/utils/python/CIME/case.py
@@ -1113,8 +1113,12 @@ class Case(object):
 
         tests = Testlist(tests_spec_file, files)
         testlist = tests.get_tests(compset=compset_alias, grid=grid_name)
-        if len(testlist) > 0:
-            logger.info("\nThis compset and grid combination is not scientifically supported, however it is used in %d tests.\n"%(len(testlist)))
+        testcnt = 0
+        for test in testlist:
+            if test["category"] == "prealpha" or test["category"] == "prebeta" or "aux_" in test["category"]:
+                testcnt += 1
+        if testcnt > 0:
+            logger.info("\nThis compset and grid combination is not scientifically supported, however it is used in %d tests.\n"%(testcnt))
         else:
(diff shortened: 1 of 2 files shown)


utils/python/CIME/case.py
```diff
@@ -1113,8 +1113,13 @@ def _check_testlists(self, compset_alias, grid_name, files):
...
         tests = Testlist(tests_spec_file, files)
         testlist = tests.get_tests(compset=compset_alias, grid=grid_name)
-        if len(testlist) > 0:
-            logger.info("\nThis compset and grid combination is not scientifically supported, however it is used in %d tests.\n"%(len(testlist)))
+        testcnt = 0
+        for test in testlist:
+            if test["category"] == "prealpha" or test["category"] == "prebeta" or "aux_" in test["category"]:
+                testcnt += 1
+        if testcnt > 0:
+            logger.info("\nThis compset and grid combination is not scientifically supported, however it is used in %d tests.\n"%(testcnt))
```
mvertens: Can we please change the expect False here to simply state:
"WARNING: This compset and grid combination is not test in CESM and therefore is unsupported."
and then exit? Rather than using an expect that gives an error. 
We can move this to a separate PR if this causes too much difficulty here - otherwise I am fine with this PR.

jedwards4b: expect now has an optional prefix argument that changes this from ERROR: to STOP:  as discussed in yesterdays meeting - I thought people objected to WARNING:

mvertens:
STOP is better I agree. Thanks.

utils/python/CIME/utils.py
```diff
@@ -8,7 +8,7 @@
 TESTS_FAILED_ERR_CODE = 100
 logger = logging.getLogger(__name__)
 
-def expect(condition, error_msg, exc_type=SystemExit):
+def expect(condition, error_msg, exc_type=SystemExit, error_prefix="ERROR: "):
```
jgfouca: @jedwards4b this default value has a whitespace after the colon but then another whitespace is added later, this is causing some unit tests to fail since the exception string is slighting different. I'm going to hotfix this.

jedwards4b: Thanks

PR #1110, opened 2017-02-07
Jgfouca/fixes to recent new tests

jgfouca:
Disable PRE.f19_f19.ADESP test, it's far too expensive.
ACME config XML needed desp config component file.
Fix minor pylint error.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: Removes PRE test from cime_developer

Code review: @jedwards4b @gold2718

### DIFF
--- a/utils/python/update_acme_tests.py
+++ b/utils/python/update_acme_tests.py
@@ -48,8 +48,8 @@ _TEST_SUITES = {
                              "ERR.f45_g37_rx1.A",
                              "ERP.f45_g37_rx1.A",
                              "SMS_D_Ln9.f19_g16_rx1.A",
-                             "DAE.f19_f19.A",
-                             "PRE.f19_f19.ADESP")
+                             "DAE.f19_f19.A")
+#                             "PRE.f19_f19.ADESP")
                             ),
 
     #
--- a/cime_config/acme/allactive/config_pesall.xml
+++ b/cime_config/acme/allactive/config_pesall.xml
@@ -407,6 +407,7 @@
       </pes>
     </mach>
   </grid>
+
   <grid name="a%ne30np4">
     <mach name="melvin">
       <pes compset="any" pesize="any">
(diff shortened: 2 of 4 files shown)


jgfouca:
@gold2718 , the PRE.f19_f19.ADESP test is far far too expensive for a cime_developer test. I killed it after waiting 45 minutes. It also would have been re-run in Z_FullSystemTest, so that's another 45 minutes at least, so >90 minutes added to scripts_regression_test. If you can come up with a testcase for PRE that takes just a couple minutes on 64 cores, that would be good.

utils/python/update_acme_tests.py
```diff
@@ -48,8 +48,8 @@
                              "ERR.f45_g37_rx1.A",
                              "ERP.f45_g37_rx1.A",
                              "SMS_D_Ln9.f19_g16_rx1.A",
-                             "DAE.f19_f19.A",
-                             "PRE.f19_f19.ADESP")
```
gold2718: Instead of eliminating these, can you change it to f45_f45?

jgfouca: Tried that, still took a very long time (I killed it after 30 minutes).

jedwards4b: Is that an indication of some problem or is it realistic that it would take so long?

gold2718: Something is weird. f19_f19 is a 2 degree run which I can usually do on a workstation. f45_f45 should be very fast. Which run is hanging?

jgfouca: PRE.*.ADESP

jgfouca: The grid didn't seem to have much impact, in both cases I killed after 30+ minutes

gold2718: If you have any run directories lying around, I would love to see a grep 'atm: model date' from them to see how far they made it.

jgfouca:
Going to merge this to get the dashboard clean. I will file a ticket to look at the PRE test.

PR #1119, opened 2017-02-09
Rljacob/split cime51 5

rljacob:
ACME's changes since 5.1.4


rljacob:
This is much better then https://github.com/ESMCI/cime/pull/1071 which listed 250+ commits and 1000's of files.

PR #1128, opened 2017-02-09
xmlquery: allow partial match of variable name

### LINKED ISSUE #947: Allow partial matches in xmlquery
It would be nice if xmlquery allowed matches on partial variable names - e.g., query for all names matching 'STOP_*'.

This request is motivated by a discussion I had with @mvertens , in which we both confessed that we generally use grep rather than xmlquery. Reflecting on this, I think the main reason is that I often forget the full variable name, so grep on a piece of this. I'd be able to use xmlquery more if it supported partial matches.

I'm not sure exactly how this should be implemented: Should you explicitly give the wildcard characters? Or should it do partial matches by default? Should there be a flag that allows partial matches, or conversely a flag that forces exact matches? I'm happy to iterate on these details with others.

cc @mvertens @jedwards4b

jgfouca:
The new -p flag turns on this capability.

I also added some usage examples to the help.

I also made it so --listall would take --subgroup into account
if it was provided too.

Test suite: code_checker + by-hand + J_TestCreateNewcase
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #947 

User interface changes?: -p flag to xmlquery

### DIFF
--- a/scripts/Tools/xmlquery
+++ b/scripts/Tools/xmlquery
@@ -21,16 +21,41 @@ from standard_script_setup import *
 
 from CIME.case import Case
 from CIME.utils import expect, convert_to_string
-import textwrap
-import sys
+
+import textwrap, sys, re
 
 logger = logging.getLogger("xmlquery")
 
 ###############################################################################
 def parse_command_line(args):
 ###############################################################################
-    parser = argparse.ArgumentParser(   description="Query the xml files for attributes and return their values." ,
-                                        formatter_class=argparse.RawDescriptionHelpFormatter, epilog=textwrap.dedent(__doc__) )
+    parser = argparse.ArgumentParser(
+usage="""%s <VAR> [<VAR> <VAR> ...]
+
+Examples:
+
+# Show NTASKS for all components
+> %s NTASKS
+
+# Show RUNDIR
+> %s RUNDIR
+
+# Show all vars containing the substring ROOT
+> %s -p '*.ROOT.*'
+
+# Show all vars in the case
+> %s --listall
+
+# Show everything in the run_pio subgroup
+> %s --listall --subgroup=run_pio
+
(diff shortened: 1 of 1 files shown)


scripts/Tools/xmlquery
```diff
@@ -163,12 +189,40 @@ def _main_func():
...
-        if listall:
-            variables = sorted(case.get_record_fields(None, "varid"))
+        if listall or partial_match:
+            all_variables = sorted(case.get_record_fields(None, "varid"))
+            if partial_match:
+                all_matching_vars = []
+                for variable in variables:
+                    for all_variable in all_variables:
+                        regex = re.compile(variable)
+                        if regex.match(all_variable):
```
gold2718: Why regex.match instead of regex.search? That way, I can do something like:
`./xmlquery -p _ATM`
and get all the ATM variables.

scripts/Tools/xmlquery
```diff
@@ -163,12 +189,40 @@ def _main_func():
...
     with Case(caseroot) as case:
-        if listall:
-            variables = sorted(case.get_record_fields(None, "varid"))
+        if listall or partial_match:
+            all_variables = sorted(case.get_record_fields(None, "varid"))
+            if partial_match:
+                all_matching_vars = []
+                for variable in variables:
+                    for all_variable in all_variables:
+                        regex = re.compile(variable)
```
jedwards4b: Should the regex assignment be outside the inner loop?

jgfouca: Yes, thx.

scripts/Tools/xmlquery
jgfouca: Good idea.

billsacks:
@jgfouca - thank you very much for this! Sorry for my silence - I was in meetings all day.

Issue #1129, opened 2017-02-09
archive needs name of history file for config_archive.xml for MOSART and RTM


ekluzek:
This is with latest cime master. And it looks like it was something that came in between the update of cime for cesm2_0_beta04 and cesm2_0_beta05.

[erik@yslogin4 cime]$ git diff
diff --git a/cime_config/cesm/config_archive.xml b/cime_config/cesm/config_archive.xml
index 51b1831..ff484fa 100644
--- a/cime_config/cesm/config_archive.xml
+++ b/cime_config/cesm/config_archive.xml
@@ -43,7 +43,7 @@
   <comp_archive_spec compname="rtm" compclass="rof">
     <rest_file_extension>\.r.*</rest_file_extension>
     <hist_file_extension>\.h.*.nc$</hist_file_extension>
-    <rest_history_varname>unset</rest_history_varname>
+    <rest_history_varname>locfnh</rest_history_varname>
     <rpointer>
       <rpointer_file>rpointer$NINST_STRING.rof</rpointer_file>
       <rpointer_content>$CASE.rtm$NINST_STRING.r.$DATENAME.nc</rpointer_content>
@@ -53,7 +53,7 @@
   <comp_archive_spec compname="mosart" compclass="rof">
     <rest_file_extension>\.r.*</rest_file_extension>
     <hist_file_extension>\.h.*.nc$</hist_file_extension>
-    <rest_history_varname>unset</rest_history_varname>
+    <rest_history_varname>locfnh</rest_history_varname>
     <rpointer>
       <rpointer_file>rpointer$NINST_STRING.rof</rpointer_file>
       <rpointer_content>$CASE.mosart$NINST_STRING.r.$DATENAME.nc</rpointer_content>

bertinia:
This was fixed in #1131

Issue #1124, opened 2017-02-09
short term archiver updates to support DART1


bertinia:
The DART v1 code base outputs a specific set of files that
need to be archived. The case_st_archive.py _archive_history_files function
is looking to match filenames with the 'dart' sting included as the component 
identifier but DART v1 does not output any such files with that string in the filename. 
Consequently, most of the DART files are not being archived correctly. 

We now have an updated list of filename pattern strings for DART so we can
update the XML and _archive_history_files function accordingly.

Issue #416, opened 2016-08-17
Does the ERR test truly cover short term archiving?


billsacks:
Looking through the ERR test, it's not clear to me that it actually covers the short-term archiving capabilities. It looks like it's essentially the same as the ERS test, except with these lines:

```
            self._case.set_value("DOUT_S", True)
```

and later:

```
            dout_s_root = self._case.get_value("DOUT_S_ROOT")
            rundir = self._case.get_value("RUNDIR")
            logger.info("staging files from archive %s" % dout_s_root)
            for item in glob.glob(os.path.join(dout_s_root, "rest", "*", "*")):
                shutil.copy(item, rundir)
```

In PR #413 the for loop is changed to:

```
for item in glob.glob(os.path.join(dout_s_root, "*", "hist", "*base")):
```

As a thought experiment, I imagined what would happen if you left out the setting of DOUT_S to True. Then the glob wouldn't pick anything up, in either the old code or the new code. But I believe the second run would proceed just fine, as if you were doing an ERS test.

Which is to say: I believe that this test would pass even if setting DOUT_S to True has absolutely no effect, or if the short-term archiver does absolutely nothing.

Big caveat: I haven't tested this at all. I'd be happy to be shown to be wrong here.

Tagging @mvertens @jgfouca @jedwards4b @bertinia

billsacks:
I believe this could be fixed by having the test explicitly remove all restart files and history files from the run directory, then doing both the old glob (for restart files) and the new glob (for hist/*base files). In that case, the test would fail if the expected files did not exist in dout_s_root.

mvertens:
This sounds like a good idea. However, I am also wondering if implementing
the ERR test with the clone mechanism you have might not be a better longer
timer solution. Thoughts?

jedwards4b:
The ERR test also tests the RESUBMIT and the job dependency mechanism.  The clone method would not test these things.

billsacks:
I agree with @jedwards4b . In addition:

1. I think we'd have to change the ERR test to use a branch run rather than continue run... making it even harder (impossible?) to test the RESUBMIT and job dependency mechanism

2. We'd need to introduce a new hook in the clone-based infrastructure that is called between the two runs (that's not a big deal)

It might have been easier to implement ERR using the original, non-clone-based infrastructure I was working on - but with the changes @jgfouca recently introduced, the value added of that version is small, so I wouldn't suggest adding that as an alternative option.

[...]

PR #1131, opened 2017-02-10
Fix config_archive.xml and case_st_archive.py

bertinia:
Add fix to config_archive.xml for DART, mosart, rtm
and ww3 element values for filename pattern matching.
Update case_st_archive.py to correctly archive DART v1 files.

Test suite: ERR.f19_g16.B1850.yellowstone_intel.allactive-defaultio
NOTE - this PR did not test with DART data because we didn't have
any available at the time. I'll work with Kevin Raeder to test later
but wanted to get these other changes checked in today for mosart and rtm.

Fixes [Github issue #]: #1129 and #1124

User interface changes: None

Input data changes: None

Code review:

### DIFF
--- a/utils/python/CIME/SystemTests/err.py
+++ b/utils/python/CIME/SystemTests/err.py
@@ -29,6 +29,13 @@ class ERR(ERS):
             rundir = self._case.get_value("RUNDIR")
             logger.info("staging files from archive %s" % dout_s_root)
             for item in glob.glob(os.path.join(dout_s_root, "*", "hist", "*base")):
+                # NOTE - this loop doesn't do anything because the
+                # _ers_first_phase has already copied the history files in the
+                # rundir adding the .base suffix before the short term archiver runs.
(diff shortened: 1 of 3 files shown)


utils/python/CIME/SystemTests/err.py
```diff
@@ -29,6 +29,13 @@ def run_phase(self):
             rundir = self._case.get_value("RUNDIR")
             logger.info("staging files from archive %s" % dout_s_root)
             for item in glob.glob(os.path.join(dout_s_root, "*", "hist", "*base")):
+                # NOTE - this loop doesn't do anything because the
```
jedwards4b: If this loop doesn't do anything why leave it here?

bertinia: @fischer-ncar and I decided to leave this comment in place as a reminder
of what needs to happen in order to address issue #416.

PR #1140, opened 2017-02-14
Cleanbld refactor

### LINKED ISSUE #1138: bld/Depends can get out of sync with dependencies in evolving code
During development, I usually have a stand-alone case that I do test builds and runs in, prior to invoking create_test. As I change code, I regularly run case.build to ensure that the new code compiles.

A recent change I made introduced a new dependency between Fortran modules, A.F90 now uses B.F90, where it did not before. The new dependency was not reflected in the bld/Depends file, because the Depends file was never updated. I also ran case.build --clean, which removed the .mod files. It turns out that make tried to compile A.F90 prior to compiling B.F90, and the compiler couldn't find B.mod. So I couldn't get the build to work. It did work when I setup a completely new case.

I think there needs to be a mechanism to rebuild the Depends file.

jedwards4b:
Make case.build clean options clean more thoroughly. 

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1138 

User interface changes?: 

Code review:


utils/python/CIME/build.py
```diff
@@ -508,43 +508,54 @@ def _build_model_thread(config_dir, compclass, caseroot, libroot, bldroot, incro
...
-    os.environ["COMP_INTERFACE"]  = case.get_value("COMP_INTERFACE")
-    os.environ["PIO_VERSION"]     = str(case.get_value("PIO_VERSION"))
-    os.environ["CLM_CONFIG_OPTS"] = clm_config_opts  if clm_config_opts is not None else ""
-
-    cmd = gmake + " -f " + casetools + "/Makefile"
-    for item in cleanlist:
-        cmd = cmd + " clean" + item
-    logger.info("calling %s "%(cmd))
-    run_cmd_no_fail(cmd)
+        cmd = gmake + " -f " + casetools + "/Makefile"
```
gold2718: Why not use os.path.join here?

utils/python/CIME/build.py
```diff
@@ -508,43 +508,54 @@ def _build_model_thread(config_dir, compclass, caseroot, libroot, bldroot, incro
     for mod_file in glob.glob(os.path.join(bldroot, "*_[Cc][Oo][Mm][Pp]_*.mod")):
         shutil.copy(mod_file, incroot)
 
+def cleantree(base, ignoredirs=None):
+    """
+    clean all files below base level, leave directories and top level files
+    """
+    ignoredirs.append(base)
```
gold2718: Won't this crash if `ignoredirs` is `None`?

utils/python/CIME/build.py
```diff
@@ -508,43 +508,54 @@ def _build_model_thread(config_dir, compclass, caseroot, libroot, bldroot, incro
...
+    for root, _, files in os.walk(base, topdown=False):
+        if root not in ignoredirs:
```
gold2718: This also won't work if `ignoredirs` is `None`

utils/python/CIME/build.py
```diff
@@ -508,43 +508,54 @@ def _build_model_thread(config_dir, compclass, caseroot, libroot, bldroot, incro
...
-def clean(case, cleanlist=None):
+def clean(case, cleanlist=None, clean_all=False):
 ###############################################################################
+    caseroot = case.get_value("CASEROOT")
+    if clean_all:
+        # If cleanlist is empty just remove the bld directory
+        exeroot = case.get_value("EXEROOT")
+        if os.path.isdir(exeroot):
+            logging.info("cleaning directory %s" %exeroot)
+            cleantree(exeroot, ignoredirs=[os.path.join(exeroot,"lib")])
```
gold2718: Why ignore lib?

utils/python/CIME/build.py
```diff
@@ -508,43 +508,54 @@ def _build_model_thread(config_dir, compclass, caseroot, libroot, bldroot, incro
...
+        # if clean_all is True also remove the sharedlibpath
+        sharedlibroot = case.get_value("SHAREDLIBROOT")
+        if sharedlibroot != exeroot and os.path.isdir(sharedlibroot):
```
gold2718: Can `sharedlibroot` be `None`? Won't `os.path.isdir` crash?

utils/python/CIME/build.py
jedwards4b: Good point, i'll fix this.

utils/python/CIME/build.py
jedwards4b: This was part of trying to do --clean by just removing the directories.   I started by removing the bld directory but then several tests failed because they were expecting to find dependencies in bld/lib so then I went back to using make to do the clean operation.   Since this code is only for clean-all now I think I can go back to rmtree and just completely remove bld.

utils/python/CIME/build.py
jedwards4b: no sharedlibroot is always assigned.

utils/python/CIME/build.py
jedwards4b: old code, not my fault :-)

gold2718:
<rant>
What is the point of having a review process if you can just ignore it? What is the huge hurry that you can't wait a few hours for me (or anyone else) to review your changes?
It happens that I don't understand all of your comments to my requests but now it is nearly impossible to get a good look since you closed up shop (so github has a hard time showing me the conversation).
This is not the first or second time I have had an issue with your rushed merges. Why do you do this?
</rant>
Did you test that you can do:
```
./case.build --clean-all
./case.build
```
without errors?

PR #1145, opened 2017-02-15
Unit test cleanup

### LINKED ISSUE #1144: Clean up some minor issues with unit test refactor
PR #1135 left some minor unresolved issues. I'll clean them up in an upcoming pr.

billsacks:
Cleans up some minor outstanding issues from PR #1135, including:

1. Brings back a commented-out line in the circle_area CMakeLists.txt,
   and adds a README documenting the expected failure of one of the
   circle_area tests.

    Jim Edwards commented out this line, but I think it should not have
    been commented out. This test is currently failing with and without
    this line. I'm guessing he commented it out thinking it would make the
    test pass, but it doesn't.

2. Changes --build-type default back to CESM_DEBUG rather than
   CIME_DEBUG

    I started down the path of changing things to consistently use CIME_DEBUG, but
    then wasn't sure that was the right thing to do: Some cmake-related code seems
    to expect CESM or ACME there - e.g., in machine_setup.py: write_cmake_macros. I
    wasn't sure what the right thing to do was, so I decided it was safest for now
    to maintain this as CESM_DEBUG.

3. Change README.unit_testing to suggest creating a temporary
   directory for the unit test build.

    The main motivation is to isolate these files that otherwise get put
    in the current directory:

        .env_mach_specific.csh
        .env_mach_specific.sh
        Depends.intel
        Macros.cmake
        env_mach_specific.xml

Test suite: Fortran unit tests according to instructions in README.unit_testing
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes #1144 

User interface changes?: none

Code review: @jedwards4b

### DIFF
--- a/tools/unit_testing/run_tests.py
+++ b/tools/unit_testing/run_tests.py
@@ -36,7 +36,7 @@ builds and runs the tests defined via CMake."""
     parser.add_argument("--build-dir", default=".",
                         help="""Directory where tests are built and run. Will be created if it does not exist."""
                         )
-    parser.add_argument("--build-type", default="CIME_DEBUG",
+    parser.add_argument("--build-type", default="CESM_DEBUG",
                         help="""Value defined for CMAKE_BUILD_TYPE."""
                         )
 
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -3,7 +3,11 @@ include(ExternalProject)
 set(CIME_ROOT "${CMAKE_CURRENT_SOURCE_DIR}")
 
 project(cime_tests Fortran C)
+
+# We rely on pio for cmake utilities like findnetcdf.cmake, so that we don't
+# need to duplicate this cmake code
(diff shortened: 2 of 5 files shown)


tools/unit_testing/run_tests.py
```diff
@@ -36,7 +36,7 @@ def parse_command_line(args):
     parser.add_argument("--build-dir", default=".",
                         help="""Directory where tests are built and run. Will be created if it does not exist."""
                         )
-    parser.add_argument("--build-type", default="CIME_DEBUG",
+    parser.add_argument("--build-type", default="CESM_DEBUG",
```
jedwards4b: Not sure why you put this back, it doesn't do anything in either case.

billsacks: The shr_infnan_mod test triggers off of this - so while it's not used in many places, it is used a bit.

PR #1160, opened 2017-02-16
Get timing correction and misc cesm changes

jedwards4b:
Corrects some of the output from get_timing to account for threads and smt.   
Updates for cheyenne and yellowstone systems and for intel support for double double
Reduce default output from check_input_data

Test baseline: 
Test namelist changes: 
Test status: bit for bit 

Fixes 

User interface changes?: 

Code review: in progress

### DIFF
--- a/utils/python/CIME/check_input_data.py
+++ b/utils/python/CIME/check_input_data.py
@@ -135,7 +135,7 @@ def check_input_data(case, svn_loc=None, input_data_root=None, data_list_dir="Bu
 
     no_files_missing = True
     for data_list_file in data_list_files:
-        logging.info("Loading input file: '%s'" % data_list_file)
+        logging.info("Loading input file list: '%s'" % data_list_file)
         with open(data_list_file, "r") as fd:
             lines = fd.readlines()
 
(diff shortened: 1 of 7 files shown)


utils/python/CIME/check_input_data.py
```diff
@@ -186,10 +186,9 @@ def check_input_data(case, svn_loc=None, input_data_root=None, data_list_dir="Bu
                             else:
                                 no_files_missing = False
                         else:
-                            logging.info("  Already had input file: '%s'" % full_path)
+                            logging.debug("  Already had input file: '%s'" % full_path)
 
                 else:
                     model = os.path.basename(data_list_file).split('.')[0]
                     logging.warning("Model %s no file specified for %s"%(model,description))
-
```
jgfouca: It's nice to have a whitespace at the end of python blocks since python lacks the standard {} syntax.

utils/python/CIME/check_input_data.py
```diff
@@ -160,7 +160,7 @@ def check_input_data(case, svn_loc=None, input_data_root=None, data_list_dir="Bu
                                 logging.warning("    Cannot download file since it lives outside of the input_data_root '%s'" % input_data_root)
                             no_files_missing = False
                         else:
-                            logging.info("  Found input file: '%s'" % full_path)
+                            logging.debug("  Found input file: '%s'" % full_path)
```
jgfouca: I changed these to info a few weeks ago but don't feel too strongly either way.

utils/python/CIME/check_input_data.py
jedwards4b: i put it back

utils/python/CIME/check_input_data.py
jedwards4b: I found it a bit overwhelming, now it tells you each list it's checking then tells you if it can't find anything in the list.   Also changed the case.build output so that it shows you the log it's writing to.

PR #1164, opened 2017-02-17
Changes needed to get unit tests to build on my Mac

### LINKED ISSUE #1163: Need to tweak unit test-related scripts to allow running unit tests on my Mac
With the current master, I can't get the Fortran unit tests to build & run on my Mac. A few small tweaks should do the trick... incoming in a PR shortly.

billsacks:
See line comments in the PR for details.

   Also ran cime Fortran unit tests on my mac using the configuration here:
   https://github.com/billsacks/mac_cime_configuration
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes #1163 

User interface changes?: none

Code review: @jedwards4b

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -10,9 +10,8 @@ list(APPEND CMAKE_MODULE_PATH "${CIME_ROOT}/externals/pio2/cmake")
 
 list(APPEND CMAKE_MODULE_PATH ${CIME_CMAKE_MODULE_DIRECTORY})
 include(CIME_utils)
-find_package(NetCDF)
-include_directories(${NETCDF_INCLUDE_DIRS})
-
+find_package(NetCDF COMPONENTS C Fortran)
+include_directories(${NetCDF_C_INCLUDE_DIRS} ${NetCDF_Fortran_INCLUDE_DIRS})
 
 # TODO: Some of the below should be done in the relevant directories, not in
 # this top level CMakeLists.
--- a/cime_config/cesm/machines/config_compilers.xml
+++ b/cime_config/cesm/machines/config_compilers.xml
@@ -69,7 +69,6 @@ using a fortran linker.
      compilers -->
 <compiler>
   <CPPDEFS>
-    <append> -D<env>OS</env> </append>
     <append MODEL="pop"> -D_USE_FLOW_CONTROL </append>
   </CPPDEFS>
   <SUPPORTS_CXX>FALSE</SUPPORTS_CXX>
--- a/cime_config/cesm/machines/config_compilers.xml
+++ b/cime_config/cesm/machines/config_compilers.xml
@@ -69,6 +69,7 @@ using a fortran linker.
      compilers -->
 <compiler>
   <CPPDEFS>
+    <append> -D<env>OS</env> </append>
     <append MODEL="pop"> -D_USE_FLOW_CONTROL </append>
   </CPPDEFS>
   <SUPPORTS_CXX>FALSE</SUPPORTS_CXX>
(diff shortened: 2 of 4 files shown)


CMakeLists.txt
```diff
@@ -10,9 +10,8 @@ list(APPEND CMAKE_MODULE_PATH "${CIME_ROOT}/externals/pio2/cmake")
 
 list(APPEND CMAKE_MODULE_PATH ${CIME_CMAKE_MODULE_DIRECTORY})
 include(CIME_utils)
-find_package(NetCDF)
-include_directories(${NETCDF_INCLUDE_DIRS})
-
+find_package(NetCDF COMPONENTS C Fortran)
```
billsacks: I'm not sure if it was necessary to list both C and Fortran components. But it seems this could be needed on machines where the C and Fortran libraries are in different places. If you feel this change was wrong, I can try to revert it.

However, the change in the following line (include_directories) IS needed on my Mac: NETCDF_INCLUDE_DIRS was empty, but NetCDF_C_INCLUDE_DIRS and NetCDF_Fortran_INCLUDE_DIRS are non-empty (they are the same on my Mac but - tied in with the above comment - may not be the same on other machines?)

cime_config/cesm/machines/config_compilers.xml
```diff
@@ -69,7 +69,6 @@ using a fortran linker.
      compilers -->
 <compiler>
   <CPPDEFS>
-    <append> -D<env>OS</env> </append>
```
billsacks: This caused problems on my mac with gnu: it leads to a bare "-D" in the compilation line, leading to a compilation error. @jedwards4b - It looks like you added this line very recently, in trying to get the unit tests to work on yellowstone, but it appears not to be needed for that purpose (scripts_regression_tests pass without it). If you feel it's needed in some cases, then we'll need to find some other tweak to allow this to work - e.g., by always ensuring that the OS environment variable is defined.

share/csm_share/test/unit/dynamic_vector/CMakeLists.txt
```diff
@@ -2,7 +2,7 @@ include_directories(.)
 
 # Because these tests use type parameterization, unfortunately we need to
 # preprocess *before* running the pFUnit preprocessor, then again *after*.
-if(${CMAKE_C_COMPILER_ID} STREQUAL GNU)
+if(${CMAKE_C_COMPILER_ID} STREQUAL GNU OR ${CMAKE_C_COMPILER_ID} STREQUAL Clang)
```
billsacks: Macs ship with the Clang compiler. I'm not completely positive that it's right to treat Clang like GNU here, but it seemed to work.

tools/unit_testing/run_tests.py
```diff
@@ -289,7 +289,6 @@ def _main():
         os.environ["compile_threaded"] = "true"
     os.environ["CC"] = find_executable("mpicc")
     os.environ["FC"] = find_executable("mpif90")
-    os.environ["NETCDF_PATH"] = os.environ.get("NETCDF")
```
billsacks: I don't have the NETCDF environment variable defined on my Mac, so this line failed. Removing it allowed unit tests to build on my Mac, and didn't break them on yellowstone. However, if you feel this is needed in some cases, then we could do something like wrapping this in a conditional or try/except.

cime_config/cesm/machines/config_compilers.xml
jedwards4b: In your config_machines.xml you should have ```<os>Darwin</os>```

billsacks:
Okay, I see this works if I also set the OS environment variable in run_tests.py - see the commit I just added. Look okay?

billsacks:
Hold off on merging: suddenly shr_infnan is failing... I'm not sure why since it was working before. Let me figure this out and get back to you.

billsacks:
> Hold off on merging: suddenly shr_infnan is failing... I'm not sure why since it was working before. Let me figure this out and get back to you.

It turns out it's a system problem on pronghorn: /bin/true isn't world-executable. I'm opening a cisl ticket.

So I think this is ready to merge if it looks good to you.

PR #1166, opened 2017-02-17
Fixing mpi-serial builds on compute001

### LINKED ISSUE #667: mpi-serial not working on anlworkstation
Using mpi-serial, the netcdf library has an unresolved depency on MPI_File_open

jayeshkrishna:
We now only include netcdf-parallel and hdf5 for non-mpi-serial
builds and include netcdf-serial for mpi-serial builds. Without
the fix builds on compute001 with mpi-serial fails due to missing
MPI function (MPI_File*) references (Note that these functions
are not required by ACME+mpi-serial or Netcdf-serial).

Test suite: X case with and without mpi-serial
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #667

User interface changes?: None

Code review: @sarich

### DIFF
--- a/cime_config/acme/machines/config_machines.xml
+++ b/cime_config/acme/machines/config_machines.xml
@@ -657,13 +657,18 @@
       <cmd_path lang="csh">source /software/common/adm/packages/softenv-1.6.2/etc/softenv-aliases.csh ; soft</cmd_path>
       <cmd_path lang="sh">source /software/common/adm/packages/softenv-1.6.2/etc/softenv-aliases.sh ; soft</cmd_path>
       <modules compiler="gnu">
-	<command name="add">+gcc-6.2.0</command>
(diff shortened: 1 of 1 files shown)


jgfouca:
Could be a problem, cime_config/acme changes are supposed to happen on the ACME side.

jayeshkrishna:
So isn't ESMCI/cime the right place for this change?

jgfouca:
Not if we want to stay consistent. The XML files were a real problem in the last ESMCI/ACME integration.

jayeshkrishna:
ok, in that case we can close this one out and create it on the ACME side. @rljacob , any thoughts?

rljacob:
Yes that sounds good.

jgfouca:
Closed; it will go into ACME.

PR #1172, opened 2017-02-20
changes to get pop namelists working with user_nl_pop changes

mvertens:
changes to get pop namelists working with user_nl_pop changes

POP has the "feature" that multiple namelist groups often have the same variable name. To distinguish the various names in user_nl_pop settings, an @ symbol is appended to the variable to specify its group. Then this variable is stripped when the namelist is written out. These changes are minimal but permit POP requirements to be met for its namelist.

Test baseline: NA
Test namelist changes: NA
Test status: bit for bit

User interface changes?: None

Code review: mvertens

### DIFF
--- a/utils/python/CIME/namelist.py
+++ b/utils/python/CIME/namelist.py
@@ -1401,7 +1401,17 @@ class _NamelistParser(object): # pylint:disable=too-few-public-methods
         while self._curr() not in separators:
             self._advance()
         text = self._text[old_pos:self._pos]
-        if not is_valid_fortran_name(text):
+
+        # @ is used in a namelist to put the same namelist variable in multiple groups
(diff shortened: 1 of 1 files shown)


utils/python/CIME/namelist.py
```diff
@@ -1401,7 +1401,17 @@ def _parse_variable_name(self, allow_equals=True):
         while self._curr() not in separators:
             self._advance()
         text = self._text[old_pos:self._pos]
-        if not is_valid_fortran_name(text):
+
+        # @ is used in a namelist to put the same namelist variable in multiple groups
```
jedwards4b: Perhaps you should include an explanation of the use of % symbol here as well.

PR #1173, opened 2017-02-21
Revert "Get timing correction and misc cesm changes"

jedwards4b:
Reverts ESMCI/cime#1160


jedwards4b:
Close this in favor of just backing out the compiler change on yellowstone.

PR #1174, opened 2017-02-21
revert intel compiler change on yellowstone

jedwards4b:
Backout the compiler update on yellowstone

Test suite:  cesm testing of clm revealed a problem in the intel compiler, confirmed this change reverts the problem 

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Code review:

### DIFF
--- a/cime_config/cesm/machines/config_machines.xml
+++ b/cime_config/cesm/machines/config_machines.xml
@@ -1684,10 +1684,10 @@
 	<command name="load"> esmf-6.3.0rp1-ncdfio-uni-O</command>
       </modules>
       <modules compiler="pgi">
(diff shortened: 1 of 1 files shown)


billsacks:
Thanks a lot for this quick fix. This backs out the intel compiler version on yellowstone, but not some of the other changes you made. I guess it's fine to move ahead with the compiler update on cheyenne, but can you please also revert the pgi & gnu version changes on yellowstone? pgi should go back to 15.10 and gnu should go back to 5.2.0 -- unless you have done extensive testing that has shown these compiler changes not to be problematic.

jedwards4b:
done

PR #1180, opened 2017-02-22
add --mpilib option to create_test

### LINKED ISSUE #1083: Add --mpilib option to create_test
Add a --mpilib option to create_test.   This for machines that support more then one mpi library and you want to try to run an entire suite with the non-default library.

jedwards4b:
Add an mpilib option to create_test, this option will override the default but not any mpilib
explicitly added in an _M testname modifier. 
Test suite: scripts_regression_tests & create_test cime_developer --mpilib mpi-serial
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1083 

User interface changes?: adds --mpilib optional argument to create_test

Code review:

### DIFF
--- a/utils/python/CIME/build.py
+++ b/utils/python/CIME/build.py
@@ -95,9 +95,10 @@ def build_model(build_threaded, exeroot, clm_config_opts, incroot, complist,
     bldroot = os.path.join(exeroot, "cpl", "obj")
     if not os.path.isdir(bldroot):
         os.makedirs(bldroot)
+    logger.info("Building %s with output to %s"%(cime_model, file_build))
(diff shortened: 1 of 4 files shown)


utils/python/CIME/build.py
```diff
@@ -95,9 +95,10 @@ def build_model(build_threaded, exeroot, clm_config_opts, incroot, complist,
     bldroot = os.path.join(exeroot, "cpl", "obj")
     if not os.path.isdir(bldroot):
         os.makedirs(bldroot)
+    logger.info("Building %s with output to %s"%(cime_model, file_build))
```
jedwards4b: This is unrelated to the pr - makes the link step output match the build steps.

PR #1182, opened 2017-02-23
Config archive split

### LINKED ISSUE #1125: config_archive.xml file split into components
The config_archive.xml needs to be split up and distributed to the component/*/cime_config
directories. This change will satisfy a request from DART to support DART v2 when it is
brought in as an active component of ESP. It will also allow more flexibility for the other
components to modify their archiving rules in the XML without requiring them to update
CIME. 

To support this change, the CIME/XML/archive Archive class init method needs 
to include the ability to construct the env_archive.xml programmatically by gathering
all the component config_archive.xml files first.

jedwards4b:
Splits the config_archive.xml file into component parts so that each component can tune the default settings to meet its needs.   The location of all files is defined in config_files.xml and the default file may contain component settings if there is a conflict between the default and a specific component setting the component wins.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1125 
User interface changes?: 

Code review: in progress

### DIFF
--- a/cime_config/cesm/config_archive.xml
+++ b/cime_config/cesm/config_archive.xml
@@ -1,4 +1,4 @@
-<components>
+<components version="2.0">
 
   <comp_archive_spec compname="cam" compclass="atm">
     <rest_file_extension>\.[ri]\..*</rest_file_extension>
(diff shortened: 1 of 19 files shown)


rljacob:
So any component that doesn't have a config_archive.xml file in its cime_config will use the default?

jedwards4b:
Yes, that's correct.

bertinia:
Couple of issues that we found after merging this PR. @jedwards4b has decided to
issue a new PR rather then revert this one. 

1. need to test ERR with allactive configuration (in progress)
2. the esp comp_archive_spec for dart needs to be included in every generated env_archive.xml regardless of the compset. This is needed for  DART v1 and will change with DART v2.
3. the generated env_archive.xml no longer validates against the config_archive.xsd and it should.
4. the config_headers.xml needs the following update:

```
 <file name="env_archive.xml">
    <header>
    These are the variables specific to the short term archiver.
    For a detailed listing of the env_archive.xml file, run
    ./case.st_archive --help
    To validate the env_archive.xml file using xmllint, run
    xmllint -schema $SRCROOT/cime/cime_config/xml_schemas/config_archive.xsd env_archive.xml
    from the case root.
    </header>
  </file>
```

PR #1184, opened 2017-02-23
add dart to all cesm archive rules, add components element

jedwards4b:
Adds the dart component to all cesm cases.  Since it's external and not a component this is the
safest way to assure that it's available. 

Test suite: scripts_regression_tests.py several tests of create_newcase with comparison of env_archive.xml to previous versions.  ERR.f09_g16.B1850 using cesm2_0_alpha06d
A, B, and F compsets were created and the env_archive.xml file was checked for correctness
several rounds of introducing errors to env_run.xml, env_case.xml, env_build.xml and env_archive.xml were conducted assuring that the schema check was working.
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Code review: mvertens

### DIFF
--- a/cime_config/acme/config_files.xml
+++ b/cime_config/acme/config_files.xml
@@ -120,6 +120,30 @@
     <desc>file containing specification of all pe-layouts for primary component (for documentation only - DO NOT EDIT)</desc>
     <schema>$CIMEROOT/cime_config/xml_schemas/config_pes.xsd</schema>
   </entry>
+  <entry id="ARCHIVE_SPEC_FILE">
+    <type>char</type>
+    <values>
+      <value>$CIMEROOT/cime_config/acme/config_archive.xml</value>
+      <value component="cpl"      >$CIMEROOT/driver_cpl/cime_config/config_archive.xml</value>
+      <!-- data model components -->
+      <value component="drof">$CIMEROOT/components/data_comps/drof/cime_config/config_archive.xml</value>
+      <value component="datm">$CIMEROOT/components/data_comps/datm/cime_config/config_archive.xml</value>
+      <value component="dice">$CIMEROOT/components/data_comps/dice/cime_config/config_archive.xml</value>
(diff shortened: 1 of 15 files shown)


bertinia:
@jedwards4b can you list which compsets you used to run create_newcase and compare env_archive.xml files with?

bertinia:
please also address these outstanding issues from PR #1184:

1. the generated env_archive.xml no longer validates against the config_archive.xsd and it should.
2. the config_headers.xml needs the following update:
 <file name="env_archive.xml">
    <header>
    These are the variables specific to the short term archiver.
    See  ./case.st_archive --help for details on running the short term archiver script.
    To validate the env_archive.xml file using xmllint, run
    xmllint -schema $SRCROOT/cime/cime_config/xml_schemas/config_archive.xsd env_archive.xml
    from the case root.
    </header>
  </file>

jedwards4b:
The generated file does not validate, none of the case files do and I have opened a new issue to address that.  I would ask that this aspect be deferred to issue #1185

jedwards4b:
I have added schema checks for env_archve, env_run, env_case and env_build files.

mvertens:
The new code looks fine. Before I approve this PR I would like to see documentation as to

1) what compsets were run by had for the create_newcase with comparison of env_archive.xml to previous versions
2) what tests for run to verify that the new schema tests are working correctly. (i.e. I assume if a user were to incorrectly change an env_run.xml file, the new schema checks would detect an error). So this needs to be verified for the different xml files.

jedwards4b:
1. A, B and F compsets were compared.
2. I created a case and introduced errors one at a time in the env_archive.xml, env_run.xml, env_build.xml and env_case.xml files.  All were caught by the schema check.

mvertens:
Thanks for the clarifications. Please add this to the documentation summary of the case - and then I will accept the PR.

Issue #1150, opened 2017-02-16
Directory structure/naming clean-up


agsalin:
In trying to gain some fluency in CIME, I am finding instances where the directory structure and naming choices could use some clean-up. This is a small step towards decreasing the learning curve for new CIME developers.

Some ideas that feel like improvements to me:

1. Move driver_cpl, externals, share, and maybe components underneath a src directory.
2. Flatten director structure of: share/csm_share/shr ; remove csm_share layer perhaps. Reconsider if all externals really are external (I think the definition is: dropped in source code that others modify and so we don't). genf90 and CMake certainly don't fit here. 
3. Rename cime_config to be config or esm_configuration or model_configuration. cime/cime_config is redundant. Can have acme, cesm, and common sub-directories. 
4. Move the 4 buildlib scripts out of cime_config. I would expect them in the source code, not in configuration with XML lists.
5. CIME/tests is nearly empty --  move unit tests into the src to remove top-level clutter.
6. Remove utils from top level. No users need to see this, and is confused with tools, which they do need. Move code currently in utils/* to a level underneath where they are used in a /utils subdirectory. For example, move utils/python/CIME files to scripts/utils and scripts/Tools/utils.
7. Get rid of capital letters in names of scripts/Testing and scripts/Tools -- no others are capitalized. scripts/Testing can be combined with utils/python/CIME/SytemTests to form scripts/testutils
8. scripts_regression_tests needs to be very visible, such as in scripts directory itself. 
9. We need a file called something like acme_test_suite_definitions or acme_test_lists instead of this functionality being part of utils/python/update_acme_tests.py, and move to where it is visible so scientists can easily add tests.

Q1: Is there any agreement to make progress along these lines?

Q2: If so, is it best done incrementally or all at once in a bigger 6.0 release? Keeping scripts and tools the same preserves the user interface. so it may not be disruptive outside of the CIME team,

Q3: What is seq? (as in  seq_comm_mct.F90 )

mvertens:
These are interesting suggestions. However, many of these changes would put
a significant burden on CESM right now since we are on the verge of
freezing the CESM2 science and preparing for a release in the June time
frame. Changing the entire directory structure of cime and renaming a lot
of these directories - including the csm_share directory would require that
all of the CESM active components to update their cime_config directories
and their build scripts.
Our number one priority on the CESM side after the science freeze will be
to update the CIME documentation so that it truly reflects what is in CIME
right now rather than what was in CESM1.

[...]

PR #1186, opened 2017-02-23
Rename CIME Directories agsalin/mkdir_src

### LINKED ISSUE #1178: Group fortran code directory within src directory
Breaking out discussion from #1150 

I propose a small clean-up of the top-level CIME dir by moving
    **externals, share, driver_cp, components**
within a directory called **src**. Also, the **buildlib.*** files for compiling these libraries will be moved from cime_config to src.

Regarding comments from #1150 

- I like the name "src" over "source" because it is commonly used for source code, such as in virtually all ACME (and presumably CESM) components like cam, clm, cice, mosart. 
- I don't think that CIME infrastructure source (i.e. python) belongs here. To me, there is a distinction between infrastructure source code and model source. common_model_source_code

agsalin:
Re-organization of CIME directory structure based on Issue #1178 discussion.

Separate commits on this branch for each step -- tests pass at each stage:

1. Move components, driver_cpl,share, externals into new src dir. 
2. Move buildlib.xxxx files from cime_config to src/build_scripts 
3. Rename cime_config to config
4. Rename driver_cpl to drivers/mct and rename drivers/mct/driver to drivers/mct/main
5. Rename shr_RandNum to RandNum
6. Rename csm_share/share to csm_share/util
7. Move contents of share/csm_share up a level into share, and rmdir csm_share (retain library of that name)
8. Remove CIME/tests and CIME/scripts/Testing dirs -- unused per discussion
9. Move utils/python to scripts/lib
10. Move scripts/libs/tests up to scripts/tests (where scripts_regression_tests.py lives) and Move utils/data_assimilation to scripts/data_assimilation.

Test suite: scripts/tests/scripts_regression_tests.py
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

Fixes #1178 and #1150 

User interface changes?: No

Code review:


agsalin:
I think I did everything we discussed, as we decided. Only the data_assimilation move from utils to scripts was extra. 

I will start looking into how much work will be needed in ACME to match this change (and in CESM by analogy).

PR #1195, opened 2017-03-02
adjust pio layout if pe layout is changed

### LINKED ISSUE #1192: PIO settings not adjusted with pe changes
Currently both PIO_STRIDE and PIO_NUMTASKS is set at create_newcase time based on the NTASKS value.   But if NTASKS is changed the PIO settings are not - this can lead to some pathologically poor IO configurations.   I propose adjusting the pio settings upon running case.setup. Since users may not always want the pio settings adjusted for them I propose an additional command line argument to case.setup --no-adjust-pio

jedwards4b:
Adjust the pio layout if pe layout changes

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1192 

User interface changes?: case.setup has a new option --no-adjust-pio 

Code review:

### DIFF
--- a/scripts/Tools/case.setup
+++ b/scripts/Tools/case.setup
@@ -46,7 +46,10 @@ OR
                         help="Does a clean followed by setup")
 
     parser.add_argument("--no-adjust-pio", action="store_true",
-                        help="Do NOT adjust pio settings for new pelayout")
+                        help="Do NOT adjust pio settings for new pelayout."
+                                  "By default if the pelayout is changed the pio layout will be adjusted."
+                                  "This option overrides that feature and leaves the pio layout as currently set.")
+
     args = parser.parse_args(args[1:])
 
     CIME.utils.handle_standard_logging_options(args)
@@ -66,7 +69,6 @@ def _main_func(description):
 
 if __name__ == "__main__":
     _main_func(__doc__)
-
 
 
 
--- a/scripts/Tools/case.setup
+++ b/scripts/Tools/case.setup
@@ -41,14 +41,17 @@ OR
 
     parser.add_argument("-t", "--test-mode", action="store_true",
                         help="Keeps the test script when the --clean argument is used")
+
     parser.add_argument("-r", "--reset", action="store_true",
                         help="Does a clean followed by setup")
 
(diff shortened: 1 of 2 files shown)


scripts/Tools/case.setup
```diff
@@ -57,9 +60,16 @@ def _main_func(description):
...
 
-    caseroot, clean, test_mode, reset = parse_command_line(sys.argv, description)
+    caseroot, clean, test_mode, reset, adjust_pio = parse_command_line(sys.argv, description)
     with Case(caseroot, read_only=False) as case:
-        case_setup(case, clean, test_mode, reset)
+        case_setup(case, clean=clean, test_mode=test_mode, reset=reset, adjust_pio=adjust_pio)
 
 if __name__ == "__main__":
     _main_func(__doc__)
+
```
jgfouca: Don't need these newlines.

scripts/Tools/case.setup
```diff
@@ -41,14 +41,17 @@ OR
 
     parser.add_argument("-t", "--test-mode", action="store_true",
                         help="Keeps the test script when the --clean argument is used")
+
     parser.add_argument("-r", "--reset", action="store_true",
                         help="Does a clean followed by setup")
 
+    parser.add_argument("--no-adjust-pio", action="store_true",
+                        help="Do NOT adjust pio settings for new pelayout")
```
jgfouca: Maybe add just a bit of documentation to this option to explain why you might want this?

jedwards4b: Hows this?
     help="Do NOT adjust pio settings for new pelayout."
                                  "By default if the pelayout is changed the pio layout will be adjusted."
                                  "This option overrides that feature and leaves the pio layout as currently set."

jgfouca: That explains what the flag does, but I was looking more for an explanation of why anyone would want to do this.

jedwards4b: I'm not sure why someone would want to do this but I'm sure they would complain if they wanted to and couldn't.

jgfouca: OK, I guess it's fine as is then. Just take care of the whitespaces and I'll merge.

jedwards4b:
done

PR #1199, opened 2017-03-03
Big refactor of CaseStatus management

### LINKED ISSUE #1038: request for standardized return status codes reported in CaseStatus

jgfouca:
Write a special handler in utils.py in order to be able to
easily encapsulate the concept that a function needs to log
its status to CaseStatus and be robust in the face of exceptions.

Rearrange and rename some functions in build.py in order to make
it more clear what is the public API and what are internal functions.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1038 

User interface changes?: CaseStatus will look slightly different

Code review: @jedwards4b @bertinia

### DIFF
--- a/scripts/Tools/case.build
+++ b/scripts/Tools/case.build
@@ -7,8 +7,8 @@ Script to build a case.
 from standard_script_setup import *
 
 import CIME.build as build
-from CIME.case           import Case
-from CIME.utils           import  append_status, find_system_test
+from CIME.case            import Case
+from CIME.utils           import find_system_test
 from CIME.XML.files       import Files
 from CIME.XML.component   import Component
 from CIME.test_status     import *
@@ -94,20 +94,10 @@ def _main_func(description):
                     ts.set_status(phase_to_fail, TEST_FAIL_STATUS, comments="failed to initialize")
                 raise
 
-            append_status("case.testbuild starting ",
-                          caseroot=caseroot,sfile="CaseStatus")
             success = test.build(sharedlib_only=sharedlib_only, model_only=model_only)
-            if success:
-                append_status("case.testbuild complete",
-                              caseroot=caseroot,sfile="CaseStatus")
         else:
-            append_status("case.build starting",
-                          caseroot=caseroot,sfile="CaseStatus")
             success = build.case_build(caseroot, case=case, sharedlib_only=sharedlib_only,
-                             model_only=model_only)
-            if success:
-                append_status("case.build complete",
-                              caseroot=caseroot,sfile="CaseStatus")
+                                       model_only=model_only)
 
     sys.exit(0 if success else 1)
 
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -2,7 +2,7 @@
 functions for building CIME models
 """
 from CIME.XML.standard_module_setup  import *
-from CIME.utils                 import get_model, append_status, analyze_build_log, stringify_bool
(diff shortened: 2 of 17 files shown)


jgfouca:
@bertinia , I believe this should give you what you need even though it's not exactly what you asked for. You'll have to infer the pending case.run based on succeeded case.submit and not-started case.run. There's also no Shutdown state since I wasn't sure how to implement this.

scripts/Tools/case.build
```diff
@@ -94,20 +94,10 @@ def _main_func(description):
                     ts.set_status(phase_to_fail, TEST_FAIL_STATUS, comments="failed to initialize")
                 raise
 
-            append_status("case.testbuild starting ",
```
jgfouca: CaseStatus handling for case.build needs to be in the library to ensure it's always done.

scripts/lib/CIME/build.py
```diff
@@ -112,205 +112,9 @@ def build_model(build_threaded, exeroot, clm_config_opts, incroot, complist,
     return logs
 
 ###############################################################################
-def post_build(case, logs):
```
jgfouca: Most changes in this file were just shuffling things around.

scripts/lib/CIME/build.py
```diff
@@ -562,10 +366,210 @@ def clean(case, cleanlist=None, clean_all=False):
...
+        case.set_value("SMP_BUILD", os.environ["SMP_VALUE"])
+    case.flush()
+
+    lock_file("env_build.xml")
+
+    # must ensure there's an lid
+    lid = os.environ["LID"] if "LID" in os.environ else run_cmd_no_fail("date +%y%m%d-%H%M%S")
+    save_build_provenance(case, lid=lid)
+
+###############################################################################
```
jgfouca: Only interesting changes are here.

jgfouca: This approach encapsulates an approach that ensure the CaseStatus gets updated even in the presence of exceptions.

scripts/lib/CIME/utils.py
```diff
@@ -775,18 +775,27 @@ def compute_total_time(job_cost_map, proc_pool):
 
     return current_time
 
-def append_status(msg, caseroot='.', sfile="CaseStatus"):
+def append_status(msg, sfile, caseroot='.'):
```
jgfouca: API changes here.

jgfouca:
Annotations complete.

jgfouca:
@bertinia , I have some other work that's going to conflict with this, so I'm going to go ahead and merge. I'd still like your feedback, so please review if you want to. I'll make a separate PR to address your concerns if you have any.

PR #1200, opened 2017-03-03
cdf64 only valid for serial netcdf, fix seq test

### LINKED ISSUE #1114: ERIO test gives false positive in PIO1
The netcdf4p and netcdf4c pio types silently fail to open that format and fail over to netcdf and
so this test falsely indicates those tests as passing.

jedwards4b:
The cdf64 parameter is only needed for serial netcdf io.  The seq test was not
implemented as intended, each component should be on it's own set of tasks.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Partially Fixes #1114 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/SystemTests/seq.py
+++ b/scripts/lib/CIME/SystemTests/seq.py
@@ -3,7 +3,7 @@ CIME smoke test  This class inherits from SystemTestsCommon
 """
 from CIME.XML.standard_module_setup import *
 from CIME.SystemTests.system_tests_common import SystemTestsCommon
-from CIME.case_setup import case_setup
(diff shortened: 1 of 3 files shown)


jgfouca:
@jedwards4b looking at #1114 , does this PR make it so we'll detect silent fails?

jedwards4b:
No, in order to do that we should add a format check to the ERIO test and check that the
file is written in the expected format.

jgfouca:
@jedwards4b should that be in this PR or something else?

jedwards4b:
It's something else I think,  I've updated the comment to say partially fixes and I'll make sure that the issue remains open.

jgfouca:
thanks

PR #1203, opened 2017-03-03
Fix bug in ERP namelist comparison.

### LINKED ISSUE #1102: NLCOMP failures due to bless_test_results storing wrong namelist in ERP tests
All of CAM's ERP namelist comparison tests are failing

When we run an ERP test, the namelists which are used for the failed NLCOMP comparison are for the second run where the number of tasks are halved and the threads are 2.

The baselines (which have been stored via bless_test_results) appear to be from the first run where the number of threads are 1 (and the tasks are double the second run)

jgfouca:
The following pattern was broken...
% ./create_test -o -g ERP.f45_g37_rx1.A.melvin_gnu -b jimtest
% ./create_test -c -n ERP.f45_g37_rx1.A.melvin_gnu -b jimtest

In the first command, the cmpgen_namelists used to occur during the RUN
phase and therefore occurs after the ERP build phase which impacts
the contents of the namelist files. In the second command, the cmpgen_namelists
occurs after the setup phase and therefore will not match the first.

With this change, we have essentially re-introduced all the namelist
creations that I tried to remove a few months ago in order to
improve performance. The SETUP phase is back up to 15 seconds on melvin.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1102 

User interface changes?: Namelists in baselines are relative to post-SETUP phase

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/SystemTests/pem.py
+++ b/scripts/lib/CIME/SystemTests/pem.py
@@ -30,4 +30,4 @@ class PEM(SystemTestsCompareTwo):
             ntasks = self._case.get_value("NTASKS_%s"%comp)
             if ( ntasks > 1 ):
                 self._case.set_value("NTASKS_%s"%comp, int(ntasks/2))
-        case_setup(self._case, clean=False, test_mode=False, reset=True)
+        case_setup(self._case, reset=True)
--- a/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -65,7 +65,7 @@ class SystemTestsCommon(object):
             logging.warning("Resetting case due to detected re-run of phase %s" % phase)
             self._case.set_initial_test_values()
 
-            case_setup(self._case, reset=True, test_mode=True, no_status=True)
+            case_setup(self._case, reset=True, test_mode=True)
 
     def build(self, sharedlib_only=False, model_only=False):
         """
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -1062,7 +1062,7 @@ class Case(object):
         clonename = self.get_value("CASE")
         logger.info(" Successfully created new case %s from clone case %s " %(newcasename, clonename))
 
-        case_setup(newcase, clean=False, test_mode=False)
+        case_setup(newcase)
 
         return newcase
 
--- a/scripts/lib/CIME/case_setup.py
+++ b/scripts/lib/CIME/case_setup.py
@@ -232,13 +232,13 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False, 
         env_module.save_all_env_info("software_environment.txt")
 
 ###############################################################################
-def case_setup(case, clean=False, test_mode=False, reset=False, no_status=False, adjust_pio=True):
+def case_setup(case, clean=False, test_mode=False, reset=False, adjust_pio=True):
 ###############################################################################
     caseroot, casebaseid = case.get_value("CASEROOT"), case.get_value("CASEBASEID")
     phase = "setup.clean" if clean else "case.setup"
     functor = lambda: _case_setup_impl(case, caseroot, clean, test_mode, reset, adjust_pio)
 
-    if case.get_value("TEST") and not no_status:
+    if case.get_value("TEST") and not test_mode:
         test_name = casebaseid if casebaseid is not None else case.get_value("CASE")
         with TestStatus(test_dir=caseroot, test_name=test_name) as ts:
             try:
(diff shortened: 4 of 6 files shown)


jgfouca:
@mvertens @jedwards4b @rljacob I continue to believe that namelists are a source of problems and poor performance in our system. We're now back to square 1 with create_namelists being repeatedly called in every test phase.

jgfouca:
Current status of create_namelists:
1) Called post-SETUP by create_test (if compare or generate baselines)
2) Called in build.py for sharedlib build
3) Called in case.submit
4) Called once per invocation of case_run.run_model

Can we really not remove any of these?

gold2718:
What would you suggest to replace namelists (aside from rewriting the model in 'anything but Fortran')?

gold2718:
Can you elaborate why you had to add back a namelist creation?
If namelists are compared -- apples-to-apples -- after a RUN phase, why would this not work?
Is there something else about the ERP test that prevents this?

jgfouca:
@gold2718 , I have no problem with namelists and maybe my wording was poor. My problem is the repeated calls to create_namelists (previously called preview_namelists) and the subtle assumptions about namelists in the components that has forced us to do this.

jgfouca:
@gold2718 we cannot support the pattern in the PR description without doing a namelist generation after SETUP.

gold2718:
Sorry for being thick but I don't understand why. I thought an ERP test did two runs in two different run directories. At the end of the runs, we should compare the results (should match) but not the namelists (should _not_ match). Is that not what is happening?
The original issue (#1102) was about doing the wrong compare with the baseline where one run's namelist was compared with the other run's baseline namelist.
What am I missing?

jgfouca:
@jedwards4b yes, I think we discussed that way back. I made a brief attempt to implement this but I quickly realized that it was a lot harder than I thought. Every phase of a test makes env XML changes, so you'd have to know exactly how these changes could impact the namelists of any component.

jgfouca:
@gold2718, ERP only does namelist operations in its "main" directory. The issue here is a bad interaction between ERP and --namelists-only in create_tests. Note that for --namelists-only, the test will only go up through the SETUP phase.

gold2718:
I guess I have a lot to learn about the interaction between the tests and the baselines.
I thought the baselines (created either through `./create_test --generate` or via `bless_test_results`) would just reflect the state of a test after it is run. In the ERP case, that means saving the _first_ (main) run's namelist to the baseline. Why wouldn't that work?

jgfouca:
@gold2718 We would not be able to support a --namelists-only option if that were the case.

jgfouca:
@gold2718 oh, you're making a good point here afterall :) , part of the problem is that the BUILD phase of ERP is leaving the XML files in the state that matches the second run. Then, early in the RUN phase, these XML files are restored to match the first run. If the XML files were restored at the end of the BUILD phase, then maybe the original problem would not have come up.

However, fixing this issue that way leaves us open to similar problems in the future. The only-full proof way to ensure that --namelist-only works is to keep the fix the way it's done in the PR.

gold2718:
Boy am I thick today. In the `./create_test --generate --namelists-only` case, are the namelists 'generated' from the first run or the final run? Why can't they be done consistently so that compares always work?

I tried to do a `--generate` to see what happens but do not have permission:
`OSError: [Errno 13] Permission denied: '/home/climate1/acme/baselines/gnu/master'`

gold2718:
Well, maybe that's the best we can do at the moment. It seems to me that the real issue is trying to manage two different builds in the same directory. The SystemTestsCompareTwo class uses two different case directories so namelist files end up in two separate directories. If ERP did that, wouldn't this problem go away (in a fairly robust manner)?

jgfouca:
@gold2718 the whole point of --namelists-only is to allow the user to do namelist operations very cheaply, IE no build phase or run phase. So, with ERP, we never even make it to the code that sets up the first and second run if --namelists-only is provided.

jgfouca:
@gold2718 , unforunately, no. Unless we forbid the BUILD phase of our system tests (like ERP) from making changes to the env*.xml files, there is no way to guarantee that --namelists-only will produce the same namelists as are present at case.submit time. The only way we could do this is if --namelists-only also did a build of the case, which completely defeats the point of a --namelist-only option.

gold2718:
But if the ERP test had two separate case directories, then wouldn't it create the two namelists without having to build or run anything?

gold2718:
What if the ERP test responded to `--namelist-only` by using the 'restart' case XML values instead if the 'initial run' XML values?

jgfouca:
@gold2718 all the special setup for system tests (like ERP) happen in the BUILD phase. The class for the test overrides the build_phase method. You'll see this if you look at erp.py.

PR #1204, opened 2017-03-04
Refactor env_batch

### LINKED ISSUE #1116: BATCH_SYSTEM not available to xmlquery
Seems that env_batch.xml is not being parsed by xmlquery

jgfouca:
In places all over the code, env_batch was treated like a standard
env*.xml file, but it's format was not compatible due to the way
jobs were being handled. This required special code all over the place
to handle and it also rendered xmlquery useless when querying for
stuff in env_batch.

This branch fixes that by re-doing the way jobs are expressed in env_batch.xml
by treating them as groups instead of sub elements of the useless job_submission
group.

This also fixes a bug entry_id.get_value where it was throwing an error
complaining of too many matches even though the user provided a subgroup
that would have limited it to a single match.

Test suite: scripts_regression_tests (melvin and skybridge)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1116 

User interface changes?: env_batch.xml now looks different

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/XML/entry_id.py
+++ b/scripts/lib/CIME/XML/entry_id.py
@@ -130,7 +130,6 @@ class EntryID(GenericXML):
         # handled in classes
         return vid, None, False
 
-
     def _get_default(self, node):
         return self._get_node_element_info(node, "default_value")
 
@@ -151,7 +150,6 @@ class EntryID(GenericXML):
                 result.append(group.get("id"))
 
         return result
-
 
     def get_valid_values(self, vid):
         node = self.get_optional_node("entry", {"id":vid})
@@ -174,7 +172,6 @@ class EntryID(GenericXML):
 
     def get_nodes_by_id(self, vid):
         return self.get_nodes("entry", {"id":vid})
-
 
     def _set_valid_values(self, node, new_valid_values):
         old_vv = self._get_valid_values(node)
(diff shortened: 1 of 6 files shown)


jgfouca:
Annotations coming, this was a significant change. So much for thinking 1116 was easy.

scripts/lib/CIME/XML/entry_id.py
```diff
@@ -254,7 +251,8 @@ def get_value(self, vid, attribute=None, resolved=True, subgroup=None):
         or from the values field if the attribute argument is provided
         and matches
         """
-        node = self.get_optional_node("entry", {"id":vid})
+        root = self.root if subgroup is None else self.get_optional_node("group", {"id":subgroup})
```
jgfouca: Important change. Allows us to avoid "more than one match" errors when the user provided a subgroup

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -108,36 +91,41 @@ def get_type_info(self, vid):
...
+                results.append(group.get("id"))
+
+        return results
+
+    def create_job_groups(self, batch_jobs):
+        # Subtle: in order to support dynamic batch jobs, we need to remove the
+        # job_submission group and replace with job-based groups
 
-        expect(len(cjobs)==0," Looks like job groups have already been created")
+        orig_group = self.get_optional_node("group", {"id":"job_submission"})
```
jgfouca: Important change. job_submission group is removed entirely, each job gets a group now. This allows is to select data by job which we we're doing all over the place, without having to put hacks in for special treatment.

jgfouca:
Tests pending.

jgfouca:
Testing looks good.

PR #1207, opened 2017-03-06
update compare_xml to find more differences

### LINKED ISSUE #1205: changing env_mach_pes.xml without rerunning case.setup still permits you to run case.submit
This will cause a lot of confusion for users - since it will only result in a run-time failure. case.submit should die if this is the case with a user-friendly error as to what needs to be done.

jedwards4b:
The <values> subnode of entry id nodes was not being checked with compare_xml and so
changes to the pelayout were not being trapped by check_lockedfiles

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1205 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/check_lockedfiles.py
+++ b/scripts/lib/CIME/check_lockedfiles.py
@@ -9,6 +9,8 @@ from CIME.XML.env_mach_pes import EnvMachPes
 from CIME.XML.env_batch import EnvBatch
 from CIME.utils import run_cmd_no_fail
 
+logger = logging.getLogger(__name__)
+
 import glob, shutil
 
 LOCKED_DIR = "LockedFiles"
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/check_lockedfiles.py
```diff
@@ -9,6 +9,8 @@
 from CIME.XML.env_batch import EnvBatch
 from CIME.utils import run_cmd_no_fail
 
+logger = logging.getLogger(__name__)
```
jgfouca: This doesn't appear to be used anywhere in the file. Is pylint OK with this? If so, then I am too.

jedwards4b: Yes pylint is fine with it, I would like to leave it in place.

 Tools/code_checker check_lockedfiles.py
File scripts/lib/CIME/check_lockedfiles.py has no pylint problems

PR #1226, opened 2017-03-09
Compclasses cleanup

### LINKED ISSUE #1170: get_timing.py specifies component classes by hand
Are lines like this:
`for k in ['CPL', 'GLC', 'WAV', 'LND', 'ROF', 'ICE', 'ATM', 'OCN']:`
appropriate in get_timing.py? They prevent get_timing from automatically picking up new component types (e.g., ESP).

jedwards4b:
Get rid of hard coded comp_class settings and use get_values instead.
Punting on NCR test - it's broken, was broken before this PR will open a ticket. 

Test suite: scripts_regression_tests.py, hand tests of PET and PEM tests.         PEM.f19_g16.F1850.yellowstone_intel
      PET.f19_g16.F1850.yellowstone_intel
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1170 

User interface changes?: 

Code review: gold2718

### DIFF
--- a/scripts/lib/CIME/test_scheduler.py
+++ b/scripts/lib/CIME/test_scheduler.py
@@ -218,13 +218,15 @@ class TestScheduler(object):
                         else:
                             self._update_test_status(test, phase, TEST_PEND_STATUS)
                             self._update_test_status(test, phase, status)
+                logger.info("Using existing test directory %s"%(self._get_test_dir(test)))
         else:
             # None of the test directories should already exist.
             for test in self._tests:
                 expect(not os.path.exists(self._get_test_dir(test)),
                        "Cannot create new case in directory '%s', it already exists."
                        " Pick a different test-id" % self._get_test_dir(test))
-        logger.info("Created test in directory %s"%self._get_test_dir(test))
+                logger.info("Creating test directory %s"%(self._get_test_dir(test)))
+
         # By the end of this constructor, this program should never hard abort,
         # instead, errors will be placed in the TestStatus files for the various
         # tests cases
--- a/config/acme/machines/Makefile
+++ b/config/acme/machines/Makefile
@@ -614,7 +614,7 @@ CMAKE_OPTS += -D CMAKE_Fortran_FLAGS:STRING="$(FFLAGS) $(INCLDIR)" \
               -D CMAKE_VERBOSE_MAKEFILE:BOOL=ON \
               -D GPTL_PATH:STRING=$(INSTALL_SHAREDPATH) \
               -D PIO_ENABLE_TESTS:BOOL=OFF \
-              -D USER_CMAKE_MODULE_PATH:STRING=$(CIMEROOT)/src/externals/CMake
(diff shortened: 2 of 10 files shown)


scripts/lib/CIME/test_scheduler.py
```diff
@@ -218,13 +218,15 @@ def __init__(self, test_names, test_data=None,
                         else:
                             self._update_test_status(test, phase, TEST_PEND_STATUS)
                             self._update_test_status(test, phase, status)
+                logger.info("Using existing test directory %s"%(self._get_test_dir(test)))
```
jedwards4b: Unrelated change to improve output from create_test

gold2718:
Why are the PET and PEM tests not tested in scripts_regression_tests.py (I know, not really part of this PR review but I'm curious)?

scripts/lib/CIME/test_scheduler.py
gold2718: Since this log output is inside the self._tests loop, wouldn't you end up with identical 'Using existing test directory' log entries if there is more than one test? I suggest either adding the test name to the log output (see below) or move the log outside the for loop.
`"Using existing test directory %s for "%(self._get_test_dir(test), test)`

scripts/lib/CIME/test_scheduler.py
```diff
@@ -218,13 +218,15 @@ def __init__(self, test_names, test_data=None,
...
         else:
             # None of the test directories should already exist.
             for test in self._tests:
                 expect(not os.path.exists(self._get_test_dir(test)),
                        "Cannot create new case in directory '%s', it already exists."
                        " Pick a different test-id" % self._get_test_dir(test))
-        logger.info("Created test in directory %s"%self._get_test_dir(test))
+                logger.info("Creating test directory %s"%(self._get_test_dir(test)))
+
```
gold2718: As above, since you moved the statement inside the loop, please add the test name to the log.

jedwards4b: Try before you comment - the output from _get_test_dir includes the test name.

scripts/lib/CIME/test_scheduler.py
jedwards4b: The output from _get_test_dir includes the test name.

jedwards4b:
Regarding the PET and PEM tests - they aren't in cime standalone tests because none of the cime standalone code includes threading and so they wouldn't do anything.

scripts/lib/CIME/test_scheduler.py
gold2718: Oy, yes the test name is buried in the directory name, however, that's not what the user typed. It was just a suggestion to aid users who may not be used to parsing the test directory names.

jedwards4b: This change was in response to a user request that just seeing the test name here was not enough and they wanted the full path to the test.   I think that the test name is output elsewhere.

jedwards4b:
I have updated the documentation to reflect the PET and PEM tests that I ran.

PR #1231, opened 2017-03-13
Mvertens/drv flds in

### LINKED ISSUE #865: build-namelist reads XML variables from files, but doesn't abort if variable not found
_From @mnlevy1981 on March 16, 2016 22:57_

Some users copy "known good" versions of XML files from case to case. A recent CIME update added three variables to `env_run.xml` relating to the barrier settings (presumably to go along with the `COMP_RUN_BARRIERS` option). These three new variables are read in by the coupler's `build-namelist` and values in `drv_in` are set accordingly... however, if these variables are not present in the XML file then `build-namelist` assumes they are blank strings and behaves accordingly. It seems like better behavior would be for `build-namelist` to error out with a "variable not found" message - currently, the first indication of a problem is that the run crashes during the namelist read.

More specifics... by default, `env_run.xml` contains

```
  <entry id="BARRIER_DATE"  value="-999">
    <type>char</type> 
    <group>run_begin_stop_restart</group> 
    <desc>Alternative date in yyyymmdd format
      sets periodic model barriers with BARRIER_OPTION and BARRIER_N for synchronization</desc> 
  </entry> 

  <entry id="BARRIER_N"  value="1">
    <type>char</type> 
    <group>run_begin_stop_restart</group> 
    <desc>sets periodic model barriers with BARRIER_OPTION and BARRIER_DATE for synchronization</desc> 
  </entry> 

  <entry id="BARRIER_OPTION"  value="never">
    <type>char</type> 
    <valid_values>none,never,nsteps,nstep,nseconds,nsecond,nminutes,nminute,nhours,nhour,ndays,nday,nmonths,nmonth,nyears,nyear,date,ifdays0,end</valid_values> 
    <group>run_begin_stop_restart</group> 
    <desc>sets frequency of full model barrier (same options as STOP_OPTION) for synchronization with BARRIER_N and BARRIER_DATE</desc> 
  </entry> 
```

which produces the following in `drv_in`:

```
 barrier_n = 1
 barrier_option = 'never'
 barrier_ymd = -999
```

Deleting those lines from `env_run.xml` (or copying in a version that never had them to begin with) results in the following in `drv_in`:

```
 barrier_n =
 barrier_option = ''
 barrier_ymd =
```

I think catching this error in `build-namelist` would be preferable.

_Copied from original issue: CESM-Development/cime#395_

### LINKED ISSUE #1217: Driver allows a contradiction in drv_flds_in between CAM and CLM
I assume this goes back to the beginning of CIME5, but I know this was nicely working in CIME4.

I expect that if the drv_flds_in namelist created by CAM and CLM contradict each other, that the driver will point out the contradiction when preview_namelist is run. If the namelists are empty for either one, there is no contradiction, and if they both set a namelist item to the same value -- there's no contradiction either. But, if one sets something one way and the other to something else -- the driver should point out the contradiction. The driver build namelist should try to merge the drv_flds_in namelists created by components, and only die if the two are set differently.

An example of what I expect to NOT work is the following, where I contradict the drv_flds_in namelist between CAM and CLM.

./create_test SMS_D.f10_f10.FW2000.yellowstone_intel.clm-fire_emis -n --compare cesm2_0_alpha06d -r . --no-build
cd $CASENAME
./xmlchange CLM_BLDNML_OPTS="-fire_emis -bgc bgc -megan -drydep"
./preview_namelist

I expect it to die on preview_namelist and tell me that CAM and CLM are contradicting each other, and what namelist items were set differently.

mvertens:
added mct/cime_config/namelist_definition_drv_flds.xml with updated schema 

1) removed bld directory
2) updated schema for namelist_definition_drv_flds.xml 
3) put in error check that there are no duplicate entries in the drv drv_flds_in that have different values - verified that this works by having CLM change the same namelists that are set by CAM in drv_flds_in

In addition to the scripts regression test - verified that the following tests were bfb with cesm2_0_alpha06f

SMS_D_Ln9.f09_f09.FWAMIP.yellowstone_intel.cam-reduced_hist3s
ERS_Ld7.f19_g16.B1850.yellowstone_intel.allactive-defaultio
ERP_Ln9.f09_f09.FC55CLUBB.yellowstone_intel.cam-outfrq9s

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1217 

User interface changes?: None

Code review: jedwards


rljacob:
"mct/cime_config" ?

mvertens:
There is a new file mct/cime_config/namelist_definition_drv_flds.xml. This file is probably not used by ACME. This file list all possible fields that are contained in the files specified by the following entry in namelist_definitions_drv.xml:

  <entry id="drv_flds_in_files" skip_default_entry="true">
    <type>char(10)</type>
    <category>drv_physics</category>
    <group>default_settings</group>
    <desc>
      List of files to merge together that contains drv_flds_in namelists
      The paths are relative to the case directory. drv_flds_in include the namelists that
      the driver reads and gives information on additional fields to be passed to different
      components that need to look at the same data.
    </desc>
    <values>
      <value>Buildconf/camconf/drv_flds_in,Buildconf/clmconf/drv_flds_in</value>
    </values>
  </entry>

ACME probably does not use these. What happens is that CAM in certain cases will have a Buildconf/camconf/drv_flds_in that contains extra fields that are set by megan, drydep, etc.
Sometimes, these fields get quite complicated. The same applies to CLM creating namelists in Buildconf/clmconf/drv_flds_in. The driver buildnml reads the clmconf/drv_flds_in and the camconf/drv_flds_in as user_nl files and creates a concatenation of all possible namelist variables in the driver drv_flds_in. These are then read by the driver via the shr code like shr_dry_depmod.F90, etc. This permits CAM to specify a whole set of new fields as part of cam's namelist generation that the driver passes from cam to clm. At any rate - this xml file was buried in the mct/bld directory and was in the old format. Its now in the new xml schema and is now consistent with namelist_definition_drv.xml.
Hopefully, this is not too long-winded of an explanation - and I hope to have some time to document this in the cime user's guide.

rljacob:
Nevermind I forgot about the new directory structure.  I thought you had put this in src/externals/mct.

PR #1237, opened 2017-03-13
no longer need this code

### LINKED ISSUE #1222: Driver buildnml attempts to set atm_adiabatic or atm_ideal
yet these have been taken out of the namelist.

jedwards4b:
Driver code no longer needs to do this for cam 

Test suite: scripts_regression_tests.py + ERP_D_Ln9.T42z30_T42.FDABIP04.hobart_nag.cam-outfrq9s 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1222 

User interface changes?: 

Code review: @gold2718

### DIFF
--- a/src/drivers/mct/cime_config/buildnml
+++ b/src/drivers/mct/cime_config/buildnml
@@ -287,15 +287,6 @@ def buildnml(case, caseroot, component):
 
     # create cplconf/namelist
     infile_text = ""
-    if case.get_value('COMP_ATM') == 'cam':
-        # cam is actually changing the driver namelist settings
-        cam_config_opts = case.get_value("CAM_CONFIG_OPTS")
-        if "adiabatic" in cam_config_opts:
-            infile_text = "atm_adiabatic = .true."
-        if "ideal" in cam_config_opts:
-            infile_text = "atm_ideal_phys = .true."
-        if "aquaplanet" in cam_config_opts:
-            infile_text = "aqua_planet = .true. \n aqua_planet_sst = 1"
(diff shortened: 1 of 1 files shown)


src/drivers/mct/cime_config/buildnml
```diff
@@ -287,15 +287,6 @@ def buildnml(case, caseroot, component):
...
     infile_text = ""
-    if case.get_value('COMP_ATM') == 'cam':
-        # cam is actually changing the driver namelist settings
-        cam_config_opts = case.get_value("CAM_CONFIG_OPTS")
-        if "adiabatic" in cam_config_opts:
-            infile_text = "atm_adiabatic = .true."
-        if "ideal" in cam_config_opts:
-            infile_text = "atm_ideal_phys = .true."
-        if "aquaplanet" in cam_config_opts:
-            infile_text = "aqua_planet = .true. \n aqua_planet_sst = 1"
```
gold2718: Too much! We still need the aquaplanet stuff (last two lines plus the COMP_ATM if statement.

jedwards4b:
Fixed

rljacob:
The driver doesn't need to do this because CAM now does it?   ACMEs CAM too?

gold2718:
Currently, build-namelist in both CAM and the ACME atmosphere model (you really should come up with a name for that) sets the *driver* namelist aqua_planet variables. However, this is a clear layer violation since the driver should set its own namelist (all we need is different components disagreeing on how to set the driver's namelist).
IMHO, the atmosphere models should stop trying to set the driver namelist.
Counter-arguments?

PR #1256, opened 2017-03-16
Mvertens/phys cycle

mvertens:
New hooks for runnning external scripts for pre-run and post-run 

CESM POP needs new pre-run and post-run hooks to carry out the lengthy spinup of its ecosystem module. This addition provides additional hooks other than data assimilation to carry this out.
However, it is general enough to be leveraged by other future pre-run and post-run scripts should those be needed.

In addition, all log files for either data assimilation OR the new pre-run and post-run hooks are now output in $RUNDIR rather than $CASEROOT

Test suite: scripts_regression_test AND PRE.f19_f19.ADESP.yellowstone_intel
Test baseline: 
Test namelist changes: 

Fixes: None

User interface changes?: None

Code review:

### DIFF
--- a/scripts/lib/CIME/case_run.py
+++ b/scripts/lib/CIME/case_run.py
@@ -175,24 +175,23 @@ def resubmit_check(case):
         submit(case, job=job, resubmit=True)
 
 ###############################################################################
-def do_external(script_name, caseroot, cycle, lid):
+def do_external(script_name, caseroot, cycle, lid, prefix):
 ###############################################################################
-    cmd = script_name + " 1> external.log.%s %d %d 2>&1" %(lid, caseroot, cycle)
+    cmd = script_name + " 1> %s.external.log.%s %s %d 2>&1" %(prefix, lid, caseroot, cycle)
     logger.debug("running %s" %script_name)
     run_cmd_no_fail(cmd)
     # disposeLog(case, 'da', lid)  THIS IS UNDEFINED!
-
 
 ###############################################################################
 def case_run(case):
 ###############################################################################
     # Set up the run, run the model, do the postrun steps
     run_with_submit = case.get_value("RUN_WITH_SUBMIT")
-    expect(run_with_submit,
-           "You are not calling the run script via the submit script. "
-           "As a result, short-term archiving will not be called automatically."
-           "Please submit your run using the submit script like so:"
-           " ./case.submit")
+    # expect(run_with_submit,
+    #        "You are not calling the run script via the submit script. "
+    #        "As a result, short-term archiving will not be called automatically."
+    #        "Please submit your run using the submit script like so:"
+    #        " ./case.submit")
 
     prerun_script = case.get_value("PRERUN_SCRIPT")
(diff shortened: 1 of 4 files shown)


billsacks:
@mvertens - thanks a lot for putting this in place. I have a feeling it could be useful for the script used by the LIWG to update CAM's topographic heights before each resubmission. Am I right in understanding the code, that this is still somewhat tied in with the data assimilation capabilities? I'm particularly noticing that the new code is nested under

```python
     for cycle in range(data_assimilation_cycles):
```

which suggests to me that you still need to set data_assimilation_cycles in order to use this new feature... is that true? (I'm not asking you to change that - I'm just trying to understand the usage.)

@JeremyFyke and @lofverstrom - this may be of interest to you, too

jedwards4b:
@billsacks DATA_ASSIMILATION_CYCLES default value is 1, so that loop is always executed at least once.

scripts/lib/CIME/case_run.py
```diff
@@ -208,6 +219,9 @@ def case_run(case):
             case.set_value("CONTINUE_RUN", "TRUE")
             lid = new_lid()
 
+        if prerun_script != "UNSET":
```
jedwards4b: I've been trying to get rid of UNSET as a default value, preferring None.

src/components/data_comps/datm/cime_config/buildnml
```diff
@@ -147,11 +147,11 @@ def _create_namelists(case, confdir, inst_string, infile, nmlgen):
 
     # set variables that are not per-stream
     if datm_mode == 'CPLHISTForcingForOcnIce':
-        nmlgen.add_default("domainfile", value='null')
         if atm_domain_file != "UNSET":
             if case.get_value('ATM_DOMAIN_FILE') != 'UNSET':
-                case.add_default('ATM_DOMAIN_FILE', value='UNSET', ignore_abs_path=True)
-                case.flush()
+                nmlgen.add_default("domainfile", value=atm_domain_file, ignore_abs_path=True)
```
jedwards4b: This is an unrelated change?

mvertens:
The datm is a bug fix for the phys_cycle spinup work. I never put in an
issue since I just discovered it as part of trying to get the spinup
working out of the box.

mvertens:
@jedwards - are you suggesting that the prerun_script and postrun_script
should be set to None in the xml file as the default value?

mvertens:
@sacks - This is not tied to the data assimilation capabilities and I very
much wanted to keep it separate. As @jedwards pointed out - in
DATA_ASSIMILATION_CYCLES by default is 1- and I envisioned this as either
using data assimilation or the postrun or prerun script.

billsacks:
@jedwards4b and @mvertens - sounds great, thanks for the clarification!

jedwards4b:
Setting it to empty
<default_value/> will result in a None value in python

mvertens:
I have changed the <default_value> to empty - but it does NOT result in a None value in python - it results in a False value. I have verified that the behavior is working with the new empty default setting.

PR #1259, opened 2017-03-17
Index in namelist

### LINKED ISSUE #1248: CIME5 user_nl scripts do not recognize array elements
In CIME2, we could enter array elements in user_nl and have those changes picked up by the scripts. In CIME5, that no longer works. As an example, we change one array element in user_nl_datm:
   `tintalgo(3) = ‘coszen’`
That has always worked, but in CIME5 we get the following error:
   `ERROR: Error in parsing namelist: 'tintalgo(3)' is not a valid variable name at line 5, column 11`
It does work if we set the entire array, i.e.:
   `tintalgo = "linear", "linear", "coszen", "linear", "linear", "linear", "linear", "linear", "linear", "linear", "linear", “linear"`
but this seems like an unnecessary loss of functionality.

Note: this is also being tracked in [ACME-Climate/ACME #1311](https://github.com/ACME-Climate/ACME/issues/1311)

### LINKED ISSUE #1274: namelists in multiinstance cases
In multiinstance cases the namelist file from the first instance is copied to all other instances and so user_nl_* changes are only applied from the instance 1 file and not all the files as they should be.

jedwards4b:
Adds support for namelist entries of the form foo(3) = 'a'

Test suite: scripts_regression_tests.py, --namelist-only tests for cesm prealpha and aux_clm45
  Note there are some expected failures here due to bugs found in the process.  One bug was that
  for multi-instance cases all namelists were using the 0001 stream file and they each should have 
  a separate file.  Another bug was that only changes in user_nl_d***_0001 were being used in all
  instances (user_nl_d***_000n where n>1 was being ignored)
Test baseline: cesm2_0_alpha06g
Test namelist changes: 
Test status: bit for bit

Fixes #1248
Fixes #1274 

User interface changes?: 

Code review: @jgfouca @mvertens

### DIFF
--- a/scripts/lib/CIME/XML/namelist_definition.py
+++ b/scripts/lib/CIME/XML/namelist_definition.py
@@ -307,7 +307,7 @@ class NamelistDefinition(EntryID):
 
         # Check size of input array.
         if len(expand_literal_list(value)) > size:
-            expect(False, "Value index exceeds variable size for variable %s"%name)
+            expect(False, "Value index exceeds variable size for variable %s, allowed array length is %s value array size is %s"%(name, size, len(expand_literal_list(value))))
         return True
 
     def _expect_variable_in_definition(self, name, variable_template):
--- a/scripts/lib/CIME/XML/namelist_definition.py
+++ b/scripts/lib/CIME/XML/namelist_definition.py
@@ -346,7 +346,7 @@ class NamelistDefinition(EntryID):
                 qualified_variable_name = get_fortran_name_only(variable_name)
                 self._expect_variable_in_definition(qualified_variable_name, variable_template)
 
-                # Check if can actuax1lly change this variable via filename change
+                # Check if can actually change this variable via filename change
                 if filename is not None:
                     self._user_modifiable_in_variable_definition(qualified_variable_name)
 
--- a/scripts/lib/CIME/XML/namelist_definition.py
+++ b/scripts/lib/CIME/XML/namelist_definition.py
@@ -307,7 +307,7 @@ class NamelistDefinition(EntryID):
 
         # Check size of input array.
         if len(expand_literal_list(value)) > size:
-            return False
+            expect(False, "Value index exceeds variable size for variable %s"%name)
         return True
 
     def _expect_variable_in_definition(self, name, variable_template):
--- a/scripts/lib/CIME/XML/namelist_definition.py
+++ b/scripts/lib/CIME/XML/namelist_definition.py
@@ -14,7 +14,7 @@ import re
 
 from CIME.namelist import fortran_namelist_base_value, \
     is_valid_fortran_namelist_literal, character_literal_to_string, \
-    expand_literal_list, Namelist, get_variable_name
+    expand_literal_list, Namelist, get_fortran_name_only
 
 from CIME.XML.standard_module_setup import *
 from CIME.XML.entry_id import EntryID
@@ -342,21 +342,22 @@ class NamelistDefinition(EntryID):
         for group_name in namelist.get_group_names():
             for variable_name in namelist.get_variable_names(group_name):
                 # Check that the variable is defined...
-                self._expect_variable_in_definition(variable_name, variable_template)
+                qualified_variable_name = get_fortran_name_only(variable_name)
+                self._expect_variable_in_definition(qualified_variable_name, variable_template)
 
                 # Check if can actuax1lly change this variable via filename change
                 if filename is not None:
-                    self._user_modifiable_in_variable_definition(variable_name)
+                    self._user_modifiable_in_variable_definition(qualified_variable_name)
 
                 # and has the right group name...
-                var_group = self.get_group(variable_name)
+                var_group = self.get_group(qualified_variable_name)
                 expect(var_group == group_name,
                        (variable_template + " is in a group named %r, but should be in %r.") %
                        (str(variable_name), str(group_name), str(var_group)))
 
                 # and has a valid value.
                 value = namelist.get_variable_value(group_name, variable_name)
-                expect(self.is_valid_value(variable_name, value),
+                expect(self.is_valid_value(qualified_variable_name, value),
                        (variable_template + " has invalid value %r.") %
                        (str(variable_name), [str(scalar) for scalar in value]))
 
@@ -379,8 +380,9 @@ class NamelistDefinition(EntryID):
         groups = {}
         for variable_name in dict_:
             variable_lc = variable_name.lower()
-            self._expect_variable_in_definition(get_variable_name(variable_lc), variable_template)
-            group_name = self.get_group(variable_lc)
+            qualified_varname = get_fortran_name_only(variable_lc)
+            self._expect_variable_in_definition(qualified_varname, variable_template)
+            group_name = self.get_group(qualified_varname)
             expect (group_name is not None, "No group found for var %s"%variable_lc)
             if group_name not in groups:
                 groups[group_name] = {}
--- a/scripts/lib/CIME/XML/namelist_definition.py
+++ b/scripts/lib/CIME/XML/namelist_definition.py
@@ -11,6 +11,7 @@ This module contains only one class, `NamelistDefinition`, inheriting from
 # pylint:disable=wildcard-import,unused-wildcard-import
 
 import re
+import collections
 
 from CIME.namelist import fortran_namelist_base_value, \
     is_valid_fortran_namelist_literal, character_literal_to_string, \
@@ -385,7 +386,7 @@ class NamelistDefinition(EntryID):
             group_name = self.get_group(qualified_varname)
             expect (group_name is not None, "No group found for var %s"%variable_lc)
             if group_name not in groups:
-                groups[group_name] = {}
+                groups[group_name] = collections.OrderedDict()
             groups[group_name][variable_lc] = dict_[variable_name]
         return Namelist(groups)
 
--- a/scripts/lib/CIME/XML/namelist_definition.py
+++ b/scripts/lib/CIME/XML/namelist_definition.py
@@ -14,7 +14,7 @@ import re
 
 from CIME.namelist import fortran_namelist_base_value, \
     is_valid_fortran_namelist_literal, character_literal_to_string, \
-    expand_literal_list, Namelist
+    expand_literal_list, Namelist, get_variable_name
 
 from CIME.XML.standard_module_setup import *
 from CIME.XML.entry_id import EntryID
@@ -344,7 +344,7 @@ class NamelistDefinition(EntryID):
                 # Check that the variable is defined...
                 self._expect_variable_in_definition(variable_name, variable_template)
 
-                # Check if can actually change this variable via filename change
+                # Check if can actuax1lly change this variable via filename change
                 if filename is not None:
                     self._user_modifiable_in_variable_definition(variable_name)
 
@@ -379,7 +379,7 @@ class NamelistDefinition(EntryID):
         groups = {}
         for variable_name in dict_:
             variable_lc = variable_name.lower()
-            self._expect_variable_in_definition(variable_lc, variable_template)
+            self._expect_variable_in_definition(get_variable_name(variable_lc), variable_template)
             group_name = self.get_group(variable_lc)
             expect (group_name is not None, "No group found for var %s"%variable_lc)
             if group_name not in groups:
--- a/scripts/lib/CIME/namelist.py
+++ b/scripts/lib/CIME/namelist.py
@@ -1018,6 +1018,9 @@ class Namelist(object):
         [u'3']
         >>> x.get_variable_value('brack', 'bar')
         [u'4']
+        >>> x.set_variable_value('foo', 'red(2:6:2)', [u'2', u'4', u'6'], var_size=12)
+        >>> x.get_variable_value('foo', 'red')
+        ['', u'2', '', u'4', '', u'6']
         """
         group_name = group_name.lower()
 
@@ -1033,8 +1036,8 @@ class Namelist(object):
         if variable_name in self._groups[group_name]:
             tlen = len(self._groups[group_name][variable_name])
         else:
-            tlen = 0
-            self._groups[group_name][variable_name] = []
+            tlen = 1
+            self._groups[group_name][variable_name] = ['']
 
         if minindex > tlen:
             self._groups[group_name][variable_name].extend(['']*(minindex-tlen-1))
@@ -1043,7 +1046,10 @@ class Namelist(object):
             if i < tlen:
                 self._groups[group_name][variable_name][i] = value.pop(0)
             else:
+                if tlen < i:
+                    self._groups[group_name][variable_name].extend(['']*(i-tlen))
                 self._groups[group_name][variable_name].append(value.pop(0))
+            tlen = len(self._groups[group_name][variable_name])
             if len(value) == 0:
                 break
 
--- a/scripts/lib/CIME/namelist.py
+++ b/scripts/lib/CIME/namelist.py
@@ -635,8 +635,7 @@ def expand_literal_list(literals, nindex=None):
             num, _, value = literal.partition('*')
             expanded += int(num) * [value]
         elif nindex is not None:
-            for i in range(0,nindex-1):
-                expanded.append('')
+            expanded = ['']*(nindex-1)
             expanded.append(literal)
         else:
             expanded.append(literal)
--- a/scripts/lib/CIME/namelist.py
+++ b/scripts/lib/CIME/namelist.py
@@ -122,12 +122,6 @@ FORTRAN_NAME_REGEX = re.compile(r"""(^[a-z][a-z0-9_]{0,62})                     
                                   \))?\s*$"""                                           # end optional index expression
                                 , re.IGNORECASE | re.VERBOSE)
 
-
-
-
-
-
-
 FORTRAN_LITERAL_REGEXES = {}
 # Integer literals.
 _int_re_string = r"(\+|-)?[0-9]+"
--- a/scripts/lib/CIME/namelist.py
+++ b/scripts/lib/CIME/namelist.py
@@ -1993,17 +1993,17 @@ class _NamelistParser(object): # pylint:disable=too-few-public-methods
         >>> x = _NamelistParser("&group /")
         >>> x._parse_namelist_group()
         >>> x._settings
-        {u'group': {}}
+        OrderedDict([(u'group', {})])
         >>> x._curr()
         u'/'
         >>> x = _NamelistParser("&group\n foo='bar','bazz'\n,, foo2=2*5\n /")
         >>> x._parse_namelist_group()
         >>> x._settings
-        {u'group': {u'foo': [u"'bar'", u"'bazz'", u''], u'foo2': [u'5', u'5']}}
+        OrderedDict([(u'group', {u'foo': [u"'bar'", u"'bazz'", u''], u'foo2': [u'5', u'5']})])
         >>> x = _NamelistParser("&group\n foo='bar','bazz'\n,, foo2=2*5\n /", groupless=True)
         >>> x._parse_namelist_group()
         >>> x._settings
-        {u'foo': [u"'bar'", u"'bazz'", u''], u'foo2': [u'5', u'5']}
+        OrderedDict([(u'foo', [u"'bar'", u"'bazz'", u'']), (u'foo2', [u'5', u'5'])])
         >>> x._curr()
         u'/'
         >>> x = _NamelistParser("&group /&group /")
@@ -2016,15 +2016,15 @@ class _NamelistParser(object): # pylint:disable=too-few-public-methods
         >>> x = _NamelistParser("&group foo='bar', foo='bazz' /")
         >>> x._parse_namelist_group()
         >>> x._settings
-        {u'group': {u'foo': [u"'bazz'"]}}
+        OrderedDict([(u'group', {u'foo': [u"'bazz'"]})])
         >>> x = _NamelistParser("&group foo='bar', foo= /")
         >>> x._parse_namelist_group()
         >>> x._settings
-        {u'group': {u'foo': [u"'bar'"]}}
+        OrderedDict([(u'group', {u'foo': [u"'bar'"]})])
         >>> x = _NamelistParser("&group foo='bar', foo= /", groupless=True)
         >>> x._parse_namelist_group()
         >>> x._settings
-        {u'foo': [u"'bar'"]}
+        OrderedDict([(u'foo', [u"'bar'"])])
         """
         group_name = self._parse_namelist_group_name()
         if not self._groupless:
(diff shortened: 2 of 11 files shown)


scripts/lib/CIME/namelist.py
```diff
@@ -112,7 +112,7 @@
 
 # Fortran syntax regular expressions.
 # Variable names.
-FORTRAN_NAME_REGEX = re.compile(r"^[a-z][a-z0-9_]{0,62}$", re.IGNORECASE)
+FORTRAN_NAME_REGEX = re.compile(r"^[a-z][a-z0-9_]{0,62}(\((\d+)\))?$", re.IGNORECASE)
```
jedwards4b: This allows for the possibility of an index in the variable name such as foo(3)

scripts/lib/CIME/namelist.py
```diff
@@ -171,6 +171,20 @@ def is_valid_fortran_name(string):
     """
     return FORTRAN_NAME_REGEX.search(string) is not None
 
+def fortran_name_index(string):
```
jedwards4b: If a valid index exists return that value.   Note that we are not yet handling array sections such as (:2) or (3:5)

scripts/lib/CIME/namelist.py
```diff
@@ -1414,6 +1453,18 @@ def _parse_variable_name(self, allow_equals=True):
         if not is_valid_fortran_name(text_check):
             raise _NamelistParseError("%r is not a valid variable name at %s" %
                                       (str(text), self._line_col_string()))
+
+        # a particular value of a namelist array may be assigned using syntax
```
jedwards4b: We create a list of indices here and pop them off later where they are used.

scripts/lib/CIME/namelist.py
```diff
@@ -1905,17 +1968,23 @@ def parse_namelist(self):
         >>> _NamelistParser("! Comment \n &group /! Comment ").parse_namelist()
         {u'group': {}}
         >>> _NamelistParser("&group1\n foo='bar','bazz'\n,, foo2=2*5\n / &group2 /").parse_namelist()
-        {u'group1': {u'foo': [u"'bar'", u"'bazz'", u''], u'foo2': [u'2*5']}, u'group2': {}}
+        {u'group1': {u'foo': [u"'bar'", u"'bazz'", u''], u'foo2': [u'5', u'5']}, u'group2': {}}
```
jedwards4b: As a side effect these values are expanded a little earlier than they were before

scripts/lib/CIME/nmlgen.py
```diff
@@ -274,12 +274,24 @@ def get_default(self, name, config=None, allow_none=False):
...
     def get_streams(self):
         """Get a list of all streams used for the current data model  mode."""
         return self.get_default("streamslist")
 
+    def clean_streams(self):
+        for variable in self._streams_variables:
+            self._streams_namelists[variable] = []
+        self._streams_namelists["streams"] = []
+
+    def new_instance(self):
```
jedwards4b: Clean data stored from previous instances so that it isn't copied into new instance by instance here I am refering to the cesm ensemble mode functionality.

jgfouca:
I will wait for @mvertens before merging.

mvertens:
@jgfouca I think we need to do some testing on the cesm side before we
bring this in.

jedwards4b:
Having the doctests there really helped in the development process.

scripts/lib/CIME/namelist.py
gold2718: Why use an incomplete regex here (no sign, no array sections)?

jgfouca:
@mvertens OK, I will hold off until you guys tell me it's ready.

scripts/lib/CIME/namelist.py
gold2718: What about `foo(-3)` or `foo(+3)`, both of which are also valid?
I have a regex for that which also includes (integer constant) array sections if you are interested.
As a side note, since you use `re.IGNORECASE`, can we get rid of all the scary language in namelist_definition_drv.xml about the id having to be lower case?

scripts/lib/CIME/namelist.py
```diff
@@ -171,6 +171,20 @@ def is_valid_fortran_name(string):
...
+    2
+    >>> fortran_name_index("bar")
+
+    >>> fortran_name_index("foo(a)")
+
+    """
+    m = FORTRAN_NAME_REGEX.search(string)
+    if m is not None and m.group(2) is not None:
+        return int(m.group(2))
+
```
gold2718: Returning an int here is restrictive in that you will not be able to handle a list of ints, such as created by an array section. Of course, even a list will not always help (see below).

scripts/lib/CIME/namelist.py
```diff
@@ -618,6 +634,9 @@ def expand_literal_list(literals):
         if FORTRAN_REPEAT_PREFIX_REGEX.search(literal) is not None:
             num, _, value = literal.partition('*')
             expanded += int(num) * [value]
+        elif nindex is not None:
+            expanded = ['']*(nindex-1)
+            expanded.append(literal)
```
gold2718: This logic will not help much with an input like:
`foo(5:6) = 2*5`

scripts/lib/CIME/namelist.py
```diff
@@ -1414,6 +1453,18 @@ def _parse_variable_name(self, allow_equals=True):
...
+        # foo(2) = 7
+
+        nindex = fortran_name_index(text_check)
+        if nindex is not None:
+            text = text_check[:text.find('(')]
+            vname = text.lower()
+            if vname in self._nameindex:
+                self._nameindex[vname].append(nindex)
+            else:
+                self._nameindex[vname] = [nindex]
```
gold2718: If `fortran_name_index` returned a list of integers instead of a single int, you could always use extend.
Before you work on that, be sure you have figured out how to handle sections correctly.
For instance, `foo(2::2)` is valid as `foo(::)` or `foo(8,4,-1`.
Another issue with trying to insert values in lists (as opposed to just listing each section in order) is that you do now know the bounds of the actual array. It is perfectly valid for an array in a namelist to have negative bounds (e.g., `integer :: foo(-3:3)`). Not knowing those bounds, how will you process:
```
foo(:) = 16*'a'
foo(-3) = 'b'
```

jedwards4b:
@gold2718: I believe that implementing a complete fortran namelist parser may be beyond the scope or intent of this issue.  Do we want something that will be ready for cesm 2 or do we want a complete solution?

gold2718:
@jedwards4b, That questions is above my pay grade of course. I suggest this be brought up at the next CIME meeting.

Whatever solution is decided on (for any particular CIME tag, not just the CESM release) should be properly documented -- especially where it deviates from the Fortran standard.

If it is decided to keep the current version, documentation should at least have:

"The namelist code assumes that all arrays have a lower bound of one unless the user enters an index of zero in an array section in which case, the assumption is that the array has a lower bound of zero."

because that is how the current code sort of behaves. However, even that is _not_ correct.

For fun, stick this in your user\_nl\_datm and try `./preview\_namelists`:
```
dtlimit(6) = 6.0
DTLIMIT(3) = 3.0
dtlimit(0) = 0.0
dtlimit(1) = 1.0
```
Funny how `dtlimit(1)` overwrote `dtlimit(0)`. Any why is there one 'default' entry of 1.5 but not others? I don't think this is ready as it stands. I think rejecting zero as a valid index (and documenting it) would allow the assumption of a lower bound of one to work.

jedwards4b:
@gold2718: Out of curiosity I tried a few of these in a short fortran program.  None of intel, pgi and gnu compilers recognize array sections in input at all.    I think that it would be okay to document that we assume all arrays begin at 1 and I think that that is the case in all current namelists.    I will look into the dtlimit example you raised.

scripts/lib/CIME/namelist.py
```diff
@@ -1414,6 +1453,18 @@ def _parse_variable_name(self, allow_equals=True):
...
         return text.lower()
```
jedwards4b: fixed

jedwards4b:
@gold2718 The number of default values set in an array is something that was not changed with this PR - I have determined that this is done in nmlgen.py subroutine _to_namelist_literals and the number of values set is determined by enumerate(value).   I haven't figured out why the result of this appears generally to be 3.   I will open this as a separate issue.

gold2718:
@jedwards4b ???
Before I began, I also wrote a short program and had no issues reading in various array sections on Yellowstone (Intel) or Hobart (Nag). Program and one trial namelist below (I tried lots of variants):
```
program foo
  implicit none
  integer :: ierr
  integer :: unitn = 63
  integer :: item(-8:8)
  namelist /foo_nl/ item
  item(:) = -1
  open(unitn, file="foo.nl", status='old', delim='APOSTROPHE')
  read(unitn, foo_nl, iostat = ierr)
  if (ierr /= 0) then
    write(6, *) 'ERROR reading foo_nl, iostat = ',ierr
  else
    write(6, '(8i5)') item
  end if
  close(unitn)
end program foo
```
foo.nl
```
&foo_nl
  item(::2)    = -8, -6
  item(8:4:-1) = 8,,,,4
  ITEM(1:2) = ,2
  item(-3) = -3
  item(5:6) = 2*5
/
```
The result
```
   -8   -1   -6   -1   -1   -3   -1   -1
   -1   -1    2   -1    4    5    5   -1
    8
```

jedwards4b:
@gold2718 I stand corrected on array section syntax.  I did not appreciate the importance of the right hand side of the above assignments.

gold2718:
Before this PR is closed, I would like to hear from @mvertens, @rljacob, and @jgfouca on the issue of how to handle the issue of array sections in namelist input. Three clear options I see are:

1. Only handle whole arrays or single elements to the left of the equals sign along with the assumption that the namelist array has a lower bound of one. This option would include documentation of the limitation in the CIME manual (sections 6.3, 6.4, 6.5 & 6.6) and possibly an additional message when an unhandled case is detected.
2. Scrap this PR and implement the full standard.
3. Merge this PR with a plan to implement the full standard prior to the release (thus avoiding the need for the additional documentation mentioned in option 1).

Thoughts? Additional options? Please note a possible issue with options 2 & 3 in my next comment.

gold2718:
Why is it so hard to implement the full Fortran standard for namelist input? I think it is a combination of the fact that we do not know the bounds of the target array along with a decision to try and combine all mentions of the array into a single namelist entry. For instance, the sequence:
```
foo = 1, 2, 3, 4, 5
foo(-3) = 9
```
requires no knowledge of the array bounds, just an assumption that `foo` has at least five elements and bounds that include index minus three. However, if we want to combine these into a single entry, we need to know where to put the nine. Our current namelist definition file does not contain this information so I see the need for a decision before attempting to handle the full Fortran namelist input standard for arrays:

1. Extend the syntax of the `<type>` tag for namelist entries to allow specification of array bounds (as in Fortran programs). This allows for proper interpretation of non-positive integer as well as array-section element selections. It also allows for rejection of out-of-bounds elements.
2. Do not try to fold multiple array element setting into a single namelist entry (i.e., keep multiple entries as in the example above).

Thoughts on these options? Am I missing a third option?

rljacob:
So did this work in CIME2 because CIME2 didn't check what was in user_nl?

mvertens:
@rjacob - no CIME2 did check what was in user_nl - but it used the perl namelist generator which handled single array indices.
@gold2718 - I would recommend having **at least the behavior we had with the perl namelist parsing of user_nl_xxx.** At this point. I would like to understand if the perl namelist parsing handled array syntax behavior. If not - I would recommend going with (1) for now and dealing handling more sophisticated namelist parsing after the cesm2.0 release in an update.

rljacob:
Agreed.  The original issue was about the missing CIME2 functionality so restoring that should be enough for now.

ekluzek:
The perl handled array syntax. There was a test file that showed all the
things it handled.

mvertens:
@erik Could you please point us to this test file.

jedwards4b:
Replaced by #1267

ekluzek:
Looks like the test file I remembered went further back than I thought. And we didn't do a good job of keeping the unit testing working as we moved forward.

But, here's a test file that I used to make sure the namelist functionality worked that goes back to cam3_2_60

&ccmexp
 string = 4*"4times"
 string2 = "1times", 2*"2times","stuff", 		
	3*"3times"
 fourones = 4*1
 two5sa4a3andtwo8s = 2*5, 4, 3, 2*8
 one1two5s = 1,2*5
 a = 'b'
 absems_data    = '/fs/cgd/csm/inputdata/atm/ccm3/abs_ems_factors_fastvx.052001.nc'
 ncdAta = "as'dfasdfasdf'asdf", STUFF = -1, !comment
 adiabatic      = .false.,
 bndtvg         = '/fs/cgd/csm/inputdata/atm/ccm3/noaamisc.r8.nc'
 BNDTVo         = '/fs/cgd/csm/inputdata/atm/ccm3/noaao3.1990.21999.nc'
 bNDTvs         = '/fs/cgd/csm/inputdata/atm/ccm3/T21M5079.nc'
 caseid         = 'testaixeulT21L18'
 cTITle         = 'testaixeulT21L18 &;:[]{}+!=\@:# $|\^(*)~dyn: eul res: T21' !comment
 dif2           = 250000, dIF4           = 2e+16, fincl1(1)      = 'T:I', fincl1(2)      = 'PS:I'
 phys_grid%n    = 1, phys_grid%thing = 1e+16, pp_grid%value = -1.123456E-349,
 fincl2(1)      = 'T:A'	!comment
 fincl2(10)     = 'Z300:A'
 fincl2(11)     = 'Z500:A'
 fincl2(12)     = 'Z700:A'
 fincl2(2)      = 'PS:A'
 fincl2(3)      = 'PSL:A'
 fincl2(4)      = 'U200:A'
 fincl2(5)      = 'U850:A'
 fincl2(6)      = 'V200:A'
 fincl2(7)      = 'V850:A'
 fincl2(8)      = 'T300:A'
 fincl2(9)      = 'T850:A'
 fexcl1         = 'T', 'Q', 'U',
                  'Z', 'W', 'L',
                  'N', 'Z', 'W', 'Y', 'L', 'M', 'N'
                  'O', 'p', 'q', 'r', 's', 't', 'u', 'v', 'w', 'x', 'y', 'z', 'ZZ'
 fexcl2         = 'T' 'Q' 'U'


     			
 fexcl6         = 'TREFAT:I', 'TREFHA:I', 'aREFHT:I' 'TaEFHT:I'
                  'TREFBT:I', 'TREFHB:I', 'bREFHT:I' 'TbEFHT:I', 'PSL:X'
                  'TREFCT:I', 'TREFHc:I','cREFHT:I' 'TcEFHT:I',
                  'TREFDT:I', 'TREFHd:I', 'dREFHT:I' 'TdEFHT:I'
                  'TREFET:I', 'TREFHe:I', 'eREFHT:I' 'TeEFHT:I'
                  'TREFFT:I', 'TREFHf:I', 'fREFHT:I' 'TfEFHT:I'
                  'TREFGT:I', 'TREFHg:I','gREFHT:I' 'TgEFHT:I', 
                  ,  'TREFHT:I', 'TREFHh:I', 'hREFHT:I' 'ThEFHT:I',
                  ,'TREFIT:I', 'TREFHi:I', 'iREFHT:I' 'TiEFHT:I',
                  ,"TREFJT:I", 'TREFHj:I', "jREFHT:I" 'TjEFHT:I'
                  'TREFKT:I', 'TREFHk:I','kREFHT:I' 'TkEFHT:I'
                  'TREFLT:I', 'TREFHl:I', 'lREFHT:I' 'TlEFHT:A', 'OBLIQ_PSL:I',
                  'SPECHUM_1000:I', 'RELHUM:A', 'PSL_1000:A', 'U_850:I'
                  'OMEGA_850MB:X,q', '5MEGA_850MB:X' 'SPECHUM_1001:I'
                  '1MEGA_850MB:X,q', '4MEGA_850MB:X' 'SPECHUM_1002:I'
                  '2MEGA_850MB:X,q','3MEGA_850MB:X' 'SPECHUM_1003:I'
                  "3MEGA_850MB:X", '2MEGA_850MB:X' 'SPECHUM_1004:I'
                  "4MEGA_850MB:X", '1MEGA_850MB:X' 'SPECHUM_1005:I'
                  '5MEGA_850MB:X', '0MEGA_850MB:X' 'SPECHUM_1006:I'
                  '6MEGA_850MB:X','9MEGA_850MB:X' 'SPECHUM_1007:I'
                  '7MEGA_850MB:X', '8MEGA_850MB:X' 'SPECHUM_1008:I'
 thingy1        = 1900,1901, 1902, 1903 1904, 1905, 1906, 1907, 1908, 1909
                  1911,1911, 1912, 1913 1914, 1915, 1916, 1917, 1918, 1919
                  1921,1921, 1922, 1923 1924, 1925, 1926, 1927, 1928, 1929
                  1931,1931, 1932, 1933 1934, 1935, 1936, 1937, 1938, 1939
                  1941,1941, 1942, 1943 1944, 1945, 1946, 1947, 1948, 1949
                  1951,1951, 1952, 1953 1954, 1955, 1956, 1957, 1958, 1959
                  1961,1961, 1962, 1963 1964, 1965, 1966, 1967, 1968, 1969
                  1971,1971, 1972, 1973 1974, 1975, 1976, 1977, 1978, 1979
                  1981,1981, 1982, 1983 1984, 1985, 1986, 1987, 1988, 1989
                  1991,1991, 1992, 1993 1994, 1995, 1996, 1997, 1998, 1999
 logical = 4*.false.,2*.true.
 logical2 = .true.,2*.false.
 real =4*5.6,2*0.0
 real1 = -123.45678e+78
 real2 = +123.45678d-78
 real3 = -123.45678
 real4 = +1234567
 real5 =  123.45678
 real6 =  1234567
 real7 = +.123456D+23
 real8 = -.123456Q12
 real9 =  .123456q23
 real10 = +.123456
 real11 = -.123456
 real12 =  123e+8
 real13 = +123d-8
 real14 = -123q
 real15 =  123E
 real16 = +9.999D00,4*5.6,2*0.0
 naked          = .28789
 naked2         = .28789e-23
 naked3         = .28789d+23
 naked4         = -.28789
 complex = 100*(4,5.0e-10)
 complex2 = (-5.2345d+23,+5.98765E-100),2*(4,5.0e-10)
 complex3 = (4 ,  5.0e-10     )
 complex4 = (-5.2345d+23, +.98765E-100        ),2*( 4   ,5.0e-10)
 irt            = 0
 iyear_ad       = 1950
 ndens(1)       = 1
 ndens(2)       = 1
 nestep         = 30
 nhtfrq(1)      = 1
 nhtfrq(2)      = 10
 nnbdat         = 1231
 nsrest         = 0
 readtrace      = .false.
 trace_gas      = .true.
 hist_mfilt     =  1,1
/

jedwards4b:
This won't work because all variables are checked against namelist_definition.xml and I assume that
many of these (thingy for example) are not real variables.

gold2718:
Also, I do not see any array sections so we should probably take a closer look at what was, and was not supported in CIME2.

jedwards4b:
Can we please move this discussion to the currently open PR #1267?   If you can point to a problem with it I will try to address.

jedwards4b:
Reopening this PR as the preferred approach for solving this problem.  We are now explicitly assuming that all namelist arrays start at index 1 and are 1 dimensional.   I am not aware of a current namelist variable that falls outside of these parameters.

jedwards4b:
I see there is still a problem when the step is > 1, I'm working on it.

jedwards4b:
I've rerun scripts_regression_tests.py after these changes - all pass.

jedwards4b:
I have tested the cesm yellowstone prealphas in namelist-only mode with this tag, all tests pass except ERS_Ly20_N2_P2.f09_g16_gl10.TG1.yellowstone_pgi which failed due to an existing bug in multiinstance cases which this PR fixes.    I am on vacation next week so I would really appreciate a thorough review of this PR asap if you can.  Thanks

scripts/lib/CIME/namelist.py
```diff
@@ -884,25 +1000,59 @@ def get_value(self, variable_name):
...
+        ['', u'2', '', u'4', '', u'6']
         """
         group_name = group_name.lower()
-        variable_name = variable_name.lower()
+
+#        print "BEFORE name %s value %s"%(variable_name, value)
+        minindex, maxindex, step = get_fortran_variable_indices(variable_name, var_size)
+        variable_name = get_fortran_name_only(variable_name.lower())
+
+        expect(minindex > 0, "CIME indices < 1 not supported in CIME interface to fortran namelists minindex=%s"%minindex)
```
jedwards4b: But the expect logic is correct here  - if in user_nl you have dtlimit(6) then minindex=maxindex=6,step=1 .

jedwards4b: I have no way of knowing or trapping for fortran lower array bounds > 1

jedwards4b:
@gold2718 I believe that I've addressed all of the issues you raised - please try again.

Thanks

jedwards4b:
Reran scripts_regression_tests.py and cesm yellowstone prealphas in namelist-only mode

scripts/lib/CIME/XML/namelist_definition.py
```diff
@@ -342,21 +343,22 @@ def validate(self, namelist,filename=None):
         for group_name in namelist.get_group_names():
             for variable_name in namelist.get_variable_names(group_name):
                 # Check that the variable is defined...
-                self._expect_variable_in_definition(variable_name, variable_template)
+                qualified_variable_name = get_fortran_name_only(variable_name)
+                self._expect_variable_in_definition(qualified_variable_name, variable_template)
 
-                # Check if can actually change this variable via filename change
+                # Check if can actuax1lly change this variable via filename change
```
gold2718: I liked the original spelling better :)

jedwards4b:
In addition to the above I also ran the aux_clm45 tests:
```
:) git checkout master
Switched to branch 'master'
Your branch is up-to-date with 'origin/master'.
yslogin6: ~/sandboxes/cesm2_0_alpha/cime/scripts
:) execca ./create_test --namelists-only --xml-machine yellowstone --xml-category aux_clm45  --generate jpe_aux_clm45_master  1> cts.out 2>&1
yslogin6: ~/sandboxes/cesm2_0_alpha/cime/scripts
:) git checkout index_in_namelist
Switched to branch 'index_in_namelist'
yslogin6: ~/sandboxes/cesm2_0_alpha/cime/scripts
:) execca ./create_test --namelists-only --xml-machine yellowstone --xml-category aux_clm45  --compare jpe_aux_clm45_master  1> cts.out 2>&1
```
All pass with the exception of a multiinstance case that was broken in master and fixed in this PR.

scripts/lib/CIME/namelist.py
```diff
@@ -112,7 +113,21 @@
...
-FORTRAN_NAME_REGEX = re.compile(r"^[a-z][a-z0-9_]{0,62}$", re.IGNORECASE)
+#FORTRAN_NAME_REGEX = re.compile(r"(^[a-z][a-z0-9_]{0,62})(\([+-]?\d*:?[+-]?\d*:?[+-]?\d*\))?$", re.IGNORECASE)
+FORTRAN_NAME_REGEX = re.compile(r"""(^[a-z][a-z0-9_]{0,62})                             #  The variable name
+                                  (\(                                                   # begin optional index expression
+                                  (([+-]?\d+)                                           # Single valued index
+                                  |                                                     # or
+                                  (([+-]?\d+)?:([+-]?\d+)?:?([+-]?\d+)?))               # colon seperated triplet
+                                  \))?\s*$"""                                           # end optional index expression
+                                , re.IGNORECASE | re.VERBOSE)
+
```
gold2718: Why all these blank lines?

scripts/lib/CIME/namelist.py
```diff
@@ -884,25 +975,54 @@ def get_value(self, variable_name):
...
+            tlen = len(self._groups[group_name][variable_name])
+        else:
+            tlen = 1
+            self._groups[group_name][variable_name] = ['']
+
+        if minindex > tlen:
+            self._groups[group_name][variable_name].extend(['']*(minindex-tlen-1))
+
+        for i in range(minindex-1, maxindex+step, step):
+            while len(self._groups[group_name][variable_name]) < i+1:
```
gold2718: I think this logic is incorrect. For instance:
`dtlimit(20:22:2) = 20.0, 22.0, 24.0`
should generate an error since the slice only indicates two elements but you accept (and set) three. This is an ERROR in Fortran and should be trapped. If you trap the error earlier, then you can maybe keep the logic.
I believe the correct number of steps is:
`1 + ((maxindex - minindex) / step)`

gold2718:
Just tried:
`dtlimit(::2) = 1.0, 3.0, 5.0`
and got
`ERROR: Too many values for array dtlimit(::2)`

scripts/lib/CIME/namelist.py
```diff
@@ -884,25 +976,54 @@ def get_value(self, variable_name):
...
+        if variable_name in self._groups[group_name]:
+            tlen = len(self._groups[group_name][variable_name])
+        else:
+            tlen = 1
+            self._groups[group_name][variable_name] = ['']
+
+        if minindex > tlen:
+            self._groups[group_name][variable_name].extend(['']*(minindex-tlen-1))
+
+        for i in range(minindex-1, maxindex+step, step):
```
gold2718: This range does not seem to work when step < 0.

jedwards4b: fixed

PR #1260, opened 2017-03-17
Enable PIO rearranger options for PIO2

### LINKED ISSUE #1224: Allow users to set pio rearranger options in PIO2
Currently users can only set the pio rearranger options in PIO1.
Since the latest version of PIO2 in CIME has support for setting
rearranger options we need to modify the CIME framework to
allow users to set these options for PIO2.

jayeshkrishna:
The PIO rearranger options, set via env_run.xml, was only
available (set) with PIO1. Now enabling it for PIO2.

Test suite: X case with PIO2
Test baseline: 
Test namelist changes: None
Test status: bit for bit

Fixes #1224

User interface changes?: None

Code review: @jedwards4b


jayeshkrishna:
@jedwards4b : Can you take a look at this PR?

PR #1267, opened 2017-03-21
Index in namelist2

### LINKED ISSUE #1248: CIME5 user_nl scripts do not recognize array elements

jedwards4b:
Handle array sections in user_nl input.   This version does not try to merge variables, however it does check for valid values.   Values out of bounds such as assigning dtlimit(0) when dtlimit array starts at 1 in the fortran are not trapped and will result in a runtime error. 
This PR supersedes #1259 

Test suite: scripts_regression_tests.py and hand testing of various inputs 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1248 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/namelist.py
+++ b/scripts/lib/CIME/namelist.py
@@ -173,7 +173,22 @@ def is_valid_fortran_name(string):
     return FORTRAN_NAME_REGEX.search(string) is not None
 
 def get_fortran_name_only(varname):
-    
+    """ remove array section if any and return only the variable name 
+    >>> get_fortran_name_only('foo')
+    'foo'
+    >>> get_fortran_name_only('foo(3)')
+    'foo'
+    >>> get_fortran_name_only('foo(::)')
+    'foo'
+    >>> get_fortran_name_only('foo(1::)')
+    'foo'
+    >>> get_fortran_name_only('foo(:+2:)')
+    'foo'
+    >>> get_fortran_name_only('foo(::-3)')
+    'foo'
+    >>> get_fortran_name_only('foo(::)')
+    'foo'
+    """
     m = FORTRAN_NAME_REGEX.search(varname)
     if m is not None:
         return m.group(1)
--- a/scripts/lib/CIME/namelist.py
+++ b/scripts/lib/CIME/namelist.py
@@ -112,7 +112,7 @@ logger = logging.getLogger(__name__)
 
 # Fortran syntax regular expressions.
 # Variable names.
-FORTRAN_NAME_REGEX = re.compile(r"(^[a-z][a-z0-9_]{0,62})(\(([+-]?\d*:?){3}\))?$", re.IGNORECASE)
+FORTRAN_NAME_REGEX = re.compile(r"(^[a-z][a-z0-9_]{0,62})(\(((\s*[+-]?\d*\s*:?\s*){3},?\s*){7}\))?$", re.IGNORECASE)
 
 FORTRAN_LITERAL_REGEXES = {}
 # Integer literals.
@@ -153,6 +153,12 @@ def is_valid_fortran_name(string):
 
     >>> is_valid_fortran_name("")
     False
+    >>> is_valid_fortran_name("a(3, 2)")
+    True
+    >>> is_valid_fortran_name("a( 3: 2: 1)")
+    True
+    >>> is_valid_fortran_name("dtlimit(3,2 )")
+    True
     >>> is_valid_fortran_name("a")
     True
     >>> is_valid_fortran_name("A")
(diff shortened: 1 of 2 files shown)


jedwards4b:
Derived types are a special case handled at around line 1431 of namelist.py this change does not
affect that special case.   As I commented above out of bounds indices are not handled here and will produce a runtime error in the model.  Adding support for multidimensional is possible - but do we need it?

scripts/lib/CIME/namelist.py
```diff
@@ -112,7 +112,8 @@
 
 # Fortran syntax regular expressions.
 # Variable names.
-FORTRAN_NAME_REGEX = re.compile(r"^[a-z][a-z0-9_]{0,62}$", re.IGNORECASE)
+FORTRAN_NAME_REGEX = re.compile(r"(^[a-z][a-z0-9_]{0,62})(\(([+-]?\d*:?){3}\))?$", re.IGNORECASE)
```
gold2718: This regex has issues. For instance, it happily accepts:
`foo(-:2:1:)`
which has two issues. One is the third colon and the other is the allowance of a sign with no digits (which makes Intel choke but surprisingly, nag doesn't mind).
I have a regex that works (at least as far as I can tell):
`re.compile(r"(^[a-z][a-z0-9_]{0,62})((\(([+-]?\d+)\))|(\(([+-]?\d+)?:([+-]?\d+)?:?([+-]?\d+)?\)))?$", re.IGNORECASE)`
It seems a bit  clumsy but has the advantage that you can recover each item in a subsection (group(6-8)).

scripts/lib/CIME/namelist.py
```diff
@@ -171,6 +172,28 @@ def is_valid_fortran_name(string):
...
+    'foo'
+    >>> get_fortran_name_only('foo(1::)')
+    'foo'
+    >>> get_fortran_name_only('foo(:+2:)')
+    'foo'
+    >>> get_fortran_name_only('foo(::-3)')
+    'foo'
+    >>> get_fortran_name_only('foo(::)')
+    'foo'
+    """
```
gold2718: It would be good to add some error tests (invalid input).

gold2718: Also, I don't see bounds error checking. It seems to happily accept negative bounds (e.g., `dtlimit(-3:2) = 0.0`).

scripts/lib/CIME/namelist.py
```diff
@@ -112,7 +112,7 @@
 
 # Fortran syntax regular expressions.
 # Variable names.
-FORTRAN_NAME_REGEX = re.compile(r"(^[a-z][a-z0-9_]{0,62})(\(([+-]?\d*:?){3}\))?$", re.IGNORECASE)
+FORTRAN_NAME_REGEX = re.compile(r"(^[a-z][a-z0-9_]{0,62})(\(((\s*[+-]?\d*\s*:?\s*){3},?\s*){7}\))?$", re.IGNORECASE)
```
gold2718: I am confused by this change. Spaces are _not_ allowed in the identifier to the left of the equal sign (which includes all the array element or array section stuff).

gold2718: I could not get multi-dimensional constructs to work. I always get an error:
`dtlimit(1,2) = 5.0`
`ERROR: Error in parsing namelist: expected literal value, but got 'dtlimit(1' at line 13, column 9`
BTW, where do the line numbers come from? This line is a comment in my user_nl_datm file and there are not 13 data entries.

jedwards4b:
@gold2718 it turns out that the sort was done in the write step after all namelist values had been merged together.  In order to maintain the order of variables in user_nl_* I have to turn off the sorting of output but then that had a side effect that the defaults now follow the user_nl_* settings in the final namelist which is not desirable either.   merging as in #1259 solved this problem.

jedwards4b:
Closing in favor of an updated #1259

PR #1275, opened 2017-03-23
First try to support titan restart

### LINKED ISSUE #1233: CIME support for workaround for Titan node failures
I'm requesting CIME support for a run-script workaround for frequent node failures on Titan described in the link below in which we allocate more nodes than necessary for the job and put a loop around the aprun command, re-launching when failure occurs as it kicks out the bad node(s).

https://www.olcf.ornl.gov/kb_articles/aprun-tips/#nodefailures

Mark brought up the issue, so I'll tag him if he wants to give more information or tag it as critical: @mt5555

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1233 

User interface changes?: Support for restarting when node fails (if config_machines.xml has the right variables set)

Code review: JimE, Marianna


jgfouca:
This is just for review. I have not been able to test this. I tried to test on skybridge by using scancel to simulate a failed node, but that produces output in TestStatus.log, not acme.log.

config/acme/machines/config_machines.xml
```diff
@@ -1399,6 +1400,7 @@
     <SUPPORTED_BY>acme</SUPPORTED_BY>
     <GMAKE_J>8</GMAKE_J>
     <MAX_TASKS_PER_NODE>16</MAX_TASKS_PER_NODE>
+    <NUM_SPARE_NODES>2</NUM_SPARE_NODES>
```
jedwards4b: This should probably be a percentage of the total nodes instead of a fixed value.

scripts/lib/CIME/case_run.py
```diff
@@ -92,7 +95,35 @@ def _run_model_impl(case, lid):
...
+            node_fail_re = case.get_value("NODE_FAIL_RE")
+            if node_fail_re and num_spare_nodes > 0:
+                node_fail_regex = re.compile(node_fail_re)
+                model_logfile = os.path.join(rundir, model + ".log." + lid)
+                if os.path.exists(model_logfile):
+                    search = node_fail_regex.search(open(model_logfile, 'r').read())
+                    if search is not None:
+                        # We failed due to node failure!
+                        logger.warning("Detected model run failed due to node failure, restarting")
+
```
jedwards4b: Should include a comment here about why we are running the st_archive - ie that it will provide the latest available consistent set of restart files.

jedwards4b: I believe that the st_archive only provides part of the functionality that you need here.  It has found the latest complete set of restart files and put them in the archive/$CASE/restart/ directory you now need to find those file and copy them back to the run directory.

scripts/lib/CIME/case_run.py
```diff
@@ -92,7 +95,35 @@ def _run_model_impl(case, lid):
...
+                        case_st_archive(case, no_resubmit=True)
+
+                        case.set_value("CONTINUE_RUN", "TRUE")
```
jedwards4b: This won't work because there is no step to copy this xml value change into the model namelist where it is needed.   I think that you need to first check the current value of CONTINUE_RUN and if it's FALSE then you need to set it to try and run preview_namelists to update the model namelist files with the change.

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -1804,6 +1804,17 @@
     </desc>
   </entry>
 
+  <entry id="NODE_FAIL_REGEX">
+    <type>char</type>
+    <default_value>UNSET</default_value>
```
jedwards4b: I've been trying to move away from UNSET as a default <default_value\> will set it to None.

scripts/lib/CIME/case_run.py
jgfouca: Thanks, I was wondering about that.

config/acme/machines/config_machines.xml
jgfouca: Agree.

scripts/lib/CIME/case_run.py
jgfouca: OK, more preview namelist calls! :)

src/drivers/mct/cime_config/config_component.xml
jgfouca: Good point. This break my code since I check for None, not UNSET. If I leave this blank, do I get a python None if I ask for it and the machine doesn't specify this value?

jedwards4b: Yes

scripts/lib/CIME/case_run.py
jgfouca: @jedwards4b do I need to worry about things like RUN_TYPE, RUN_REFDATE, RUN_REFTOD , etc?

jedwards4b: No, only CONTINUE_RUN

jgfouca:
@jedwards4b I think I address all your concerns. Please re-review. I think I can add a Fake test that tests this.

jgfouca:
@jedwards4b , yes, I'm going to utilize the FakeTest mechanism we already have in place.

jgfouca:
@jedwards4b @mvertens I think I have this working and tested. Please review

jgfouca:
@jedwards4b please do not merge this PR until #1287 is merged. There are going to be some interesting conflicts between the "spare nodes" capability introduced here and the aprun restoration that happened on the ACME side. This PR is still worth reviewing though, especially the new NODEFAIL test.

jgfouca:
@jedwards4b , even better, it will work on any system period.

vanroekel:
@jedwards4b  and @jgfouca I wanted to confirm that the restart capability works in a large PE count (60K) titan production run.  I had a node failure on titan and the model restarted perfectly!  Thanks for adding this feature.

PR #1287, opened 2017-03-29
Merge changes from acme 03292017

### LINKED ISSUE #1271: Need support for aprun commands with different task/thread counts
An example of the syntax that needs to be supported (and was with the old task_maker).

`aprun   -S 4  -n 680 -N 8 -d 2 xxx/acme.exe  : -S 2  -n 128 -N 4 -d 4 xxx/acme.exe'` 

Translation: the first 680 MPI processes should be assigned to nodes 8 at a time, 4 per numa domain, with a separation of 2 between each task (allowing two OpenMP threads, though that is not required); the second 128 processes should be assigned 4 per node, with 2 per numa domain, and with a separation of 4.

agsalin:
Pull back ACME repo changes of CIME since last subtree split (late January?) up to March 29 back into CIME master.

Test suite: scripts_regression_tests on penn pass
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1271 

User interface changes?: 

Code review: PLEASE!


rljacob:
Was this first merged to a maint-cime5.2 branch?  I'd like to tag that (5.2.1) before we mingle it with cime5.3 code.

agsalin:
This branch agsalin/cime52-with-acmesplit-03292017 could be called maint-cime5.2 and the current head tagged 5.2.1. 
I checked out the cime5.2.0 tag, created a branch, and merged in ACME code into it.

scripts/lib/CIME/build.py
```diff
@@ -112,9 +112,206 @@ def _build_model(build_threaded, exeroot, clm_config_opts, incroot, complist,
     return logs
 
 ###############################################################################
-def _build_checks(case, build_threaded, comp_interface, use_esmf_lib,
-                  debug, compiler, mpilib, complist, ninst_build, smp_value,
-                  model_only):
+def post_build(case, logs):
+###############################################################################
```
jgfouca: This doesn't look right. It's showing lots of code added that should have already been there.

scripts/lib/CIME/aprun.py
```diff
@@ -0,0 +1,137 @@
...
+
+logger = logging.getLogger(__name__)
+
+###############################################################################
+def _get_aprun_cmd_for_case_impl(ntasks, nthreads, rootpes, pstrids,
+                                 max_tasks_per_node, pes_per_node,
+                                 pio_numtasks, pio_async_interface,
+                                 compiler, machine, run_exe):
+###############################################################################
+    """
```
jgfouca: @jedwards4b @mvertens this is the replacement for task_maker.

scripts/lib/CIME/build.py
jgfouca: @agsalin , the only recent change (last two months) to this file on the ACME side was one line:

```
@@ -292,6 +292,7 @@ def case_build(caseroot, case, sharedlib_only=False, model_only=False):
                                cimeroot, libroot, lid, compiler)
 
     if not sharedlib_only:
+        os.environ["INSTALL_SHAREDPATH"] = os.path.join(exeroot, sharedpath) # for MPAS makefile generators
```

agsalin: I just made a commit to add this one line to the current master. Should fix this.

scripts/lib/CIME/BuildTools/configure.py
```diff
@@ -37,7 +37,7 @@ def configure(machobj, output_dir, macros_format, compiler, mpilib, debug, sysos
     """
     # Macros generation.
     suffixes = {'Makefile': 'make', 'CMake': 'cmake'}
-    macro_maker = Compilers(machobj)
+    macro_maker = Compilers(machobj, compiler=compiler, mpilib=mpilib)
```
jgfouca: This was a critical fix for the v1 macro writer.

scripts/lib/CIME/BuildTools/macrowriterbase.py
```diff
@@ -15,6 +15,59 @@
 from CIME.XML.standard_module_setup import *
 logger = logging.getLogger(__name__)
 
+def _get_components(value):
+    """
+    >>> value = '-something ${shell ${NETCDF_PATH}/bin/nf-config --flibs} -lblas -llapack'
```
jgfouca: This was a big change that I can't quite remember why it was needed. shell and environment variables weren't being handled correctly.

scripts/lib/CIME/XML/compilers.py
```diff
@@ -149,7 +149,7 @@ def write_macros_file(self, macros_file="Macros.make", output_format="make", xml
             for compiler_node in reversed(self.compiler_nodes):
                 _add_to_macros(compiler_node, macros)
             write_macros_file_v1(macros, self.compiler, self.os,
-                                        self.machine, macros_file="Macros.make",
+                                        self.machine, macros_file=macros_file,
```
jgfouca: Key bug fix.

scripts/lib/CIME/test_scheduler.py
```diff
@@ -592,9 +592,10 @@ def _get_procs_needed(self, test, phase, threads_in_flight=None, no_batch=False)
             # Will force serialization of sharedlib builds
             # TODO - instead of serializing, compute all library configs needed and build
             # them all in parallel
-            for _, _, running_phase in threads_in_flight.values():
-                if (running_phase == SHAREDLIB_BUILD_PHASE):
-                    return self._proc_pool + 1
+            if get_model() == "cesm":
```
jgfouca: ACME still can't handle sharing shared libs.

scripts/lib/CIME/utils.py
```diff
@@ -777,6 +777,78 @@ def compute_total_time(job_cost_map, proc_pool):
 
     return current_time
 
+def format_time(time_format, input_format, input_time):
+    """
```
jgfouca: @mfdeakin added this, maybe he can say why it was needed.

mfdeakin-sandia: This was needed because CIME was failing with allocation walltimes greater than 23:59:59. This was a limitation of the Python date parser; it doesn't accept >=24 in the hours field; instead expecting it to roll over in to the days field.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -270,6 +269,10 @@ def get_submit_args(self, case, job):
                     if flag == "-n" and rval<= 0:
                         rval = 1
 
+                    if flag == "-q" and rval == "batch" and case.get_value("MACH") == "blues":
```
jedwards4b: You should be able to do this in xml and not in code.

scripts/lib/CIME/case.py
```diff
@@ -116,28 +118,31 @@ def check_if_comp_var(self, vid):
...
+
         self.tasks_per_numa = int(math.ceil(self.tasks_per_node / 2.0))
         smt_factor = max(1,int(self.get_value("MAX_TASKS_PER_NODE") / pes_per_node))
 
         threads_per_node = self.tasks_per_node * self.thread_count
         threads_per_core = 1 if (threads_per_node <= pes_per_node) else smt_factor
         self.cores_per_task = self.thread_count / threads_per_core
 
-        return total_tasks
+        if self.get_value("MACH") == "titan":
```
jedwards4b: can we check if mpirun executable is aprun instead of hard coding titan?

scripts/lib/CIME/case.py
```diff
@@ -1060,20 +1065,17 @@ def get_mpirun_cmd(self, job="case.run"):
...
-        if executable == "aprun":
-            totalpes = self.get_value("TOTALPES")
-            pes_per_node = self.get_value("PES_PER_NODE")
-            if totalpes < pes_per_node:
-                args["tasks_per_node"] = "-N "+str(totalpes)
-
-        mpi_arg_string = " ".join(args.values())
 
+        # special case for aprun
+        if executable == "aprun":
```
jedwards4b: Use this instead of test for titan above at line 142?

scripts/lib/CIME/case_setup.py
```diff
@@ -210,7 +210,7 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False,
                 logger.info("Finished testcase.setup")
 
         # Some tests need namelists created here (ERP) - so do this if are in test mode
-        if test_mode:
+        if test_mode or get_model() == "acme":
```
jedwards4b: Why?

scripts/lib/CIME/preview_namelists.py
```diff
@@ -94,9 +95,10 @@ def create_namelists(case):
             raise
 
         if do_run_cmd:
-            logger.debug("   Running %s buildnml"%compname)
+            logger.info("   Running %s buildnml"%compname)
```
jedwards4b: too noisy for default in my opinion.

src/share/util/shr_orb_mod.F90
```diff
@@ -78,6 +78,8 @@ real(SHR_KIND_R8) pure FUNCTION shr_orb_cosz(jday,lat,lon,declin,dt_avg)
       shr_orb_cosz =  shr_orb_avg_cosz(jday, lat, lon, declin, dt_avg)
    else
       shr_orb_cosz = sin(lat)*sin(declin) - &
+   !   &              cos(lat)*cos(declin)*cos(jday*2.0_SHR_KIND_R8*pi + lon)
```
jedwards4b: remove commented code?

scripts/lib/CIME/XML/env_batch.py
jgfouca: @jedwards4b I couldn't figure out how to do that in XML. It's a pretty unusual case: the -q <queue> option needs to be provided unless the queue is "batch". I don't think we can express that in XML.

jedwards4b: okay

scripts/lib/CIME/XML/env_mach_specific.py
```diff
@@ -372,16 +372,17 @@ def get_mpirun(self, case, attribs, check_members=None, job="case.run"):
...
-        if arg_node is not None:
-            arg_nodes = self.get_nodes("arg", root=arg_node)
-            for arg_node in arg_nodes:
-                arg_value = transform_vars(arg_node.text,
-                                           case=case,
-                                           subgroup=job,
-                                           check_members=check_members,
-                                           default=arg_node.get("default"))
-                args[arg_node.get("name")] = arg_value
+        if exe_only:
```
jedwards4b: This logic looks wrong - if exe_only is False then args is not assigned either?  I think you mean 
if not exe_only?

jgfouca: Thanks, yes

jedwards4b:
@mvertens is on vacation until next week.   Do we want another reviewer or do we want to wait?

rljacob:
Last week we agreed we just needed your review.

jgfouca:
@jedwards4b I think we're good. All tests are passing now with a fix for the problem you saw. I'll resolve conflicts and merge.

jedwards4b:
Sounds good - thanks

scripts/lib/CIME/case_setup.py
jgfouca: It's been a while since I fixed this. I think one of our models had buildnml stuff that depended on it being run at setup time.

scripts/lib/CIME/preview_namelists.py
jgfouca: I think I changed this because in the old implementation, all the output from buildnml was completely lost and I had trouble finding something.

src/share/util/shr_orb_mod.F90
jgfouca: @rljacob , thoughts?

rljacob: Yes you can.

PR #1299, opened 2017-04-03
testreporter rewrite to python

### LINKED ISSUE #1292: testreporter.pl not being copied to testing directory
Bug in test_scheduler.py was failing to copy testreporter.pl to the testing directory.

fischer-ncar:
Rewrite testreporter into python.  The perl version was having
issues handling web posts on different systems.
Fix issue where testreport failed to be copied to testing
director

Test suite: scripts_regression_tests.py, all pass except K_TestCimeCase, which is
                  also failing on master.
                 populated testdata base for alpha06i.
Test baseline: 
Test namelist changes: 
Test status:bit for bit

Fixes #1292 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/Tools/testreporter.py
+++ b/scripts/Tools/testreporter.py
@@ -1,0 +1,240 @@
+#!/usr/bin/env python
+
+"""
+
+Simple script to populate CESM test database with test results.  This can be 
+run with out any knowledge of CIME.
+
+"""
+
+import os
+import glob
+import xml.etree.ElementTree as ET
+from xml.dom import minidom
+import HTMLParser 
+import urllib
+import argparse
+
+
+
+# Parse command line options
+
+parser = argparse.ArgumentParser(description='Arguements for testreporter')
+parser.add_argument("--tagname", 
+                        help="Name of the tag being tested.")
+parser.add_argument("--testid", 
+                        help="Test id, ie c2_0_a6g,c2_0_b6g.")
+parser.add_argument("--testroot", 
+                        help="Root directory for tests to populate the database.")
+parser.add_argument("--testtype", 
+                        help="Type of test, prealpha or prebeta.")
+parser.add_argument("--dryrun",action="store_true", 
+                        help="Do a dry run, database will not be populated.")
+parser.add_argument("--dumpxml",action="store_true", 
+                        help="Dump XML test results to sceen.")
+args = parser.parse_args()
+
+# Fill in values needed from xml files (COMPILER, MPILIB, MACH, BASELINE_NAME_CMP.
+
+os.chdir(args.testroot)
+
+xml_file=glob.glob("*"+args.testid+"/env_build.xml")
+root = ET.parse(xml_file[0]).getroot()
+for child in root.iter('group'):
+   for entry in child.iter('entry'):
+     id = entry.get('id')
+     value = entry.get('value')
+     if id == "COMPILER":
+        compiler=value
+     if id == "MPILIB":
+        mpilib = value
+
+xml_file=glob.glob("*"+args.testid+"/env_case.xml")
+root = ET.parse(xml_file[0]).getroot()
+for child in root.iter('group'):
+   for entry in child.iter('entry'):
+     id = entry.get('id')
+     value = entry.get('value')
+     if id == "MACH":
+        machine=value
+
+xml_file=glob.glob("*"+args.testid+"/env_test.xml")
+root = ET.parse(xml_file[0]).getroot()
+for child in root.iter('group'):
+   for entry in child.iter('entry'):
+     id = entry.get('id')
+     value = entry.get('value')
+     if id == "BASELINE_NAME_CMP":
+        baseline=value
+
+
+#
+# Create XML
+#
+
+testrecord    = ET.Element("testrecord")
+tag_name      = ET.SubElement(testrecord,'tag_name').text=args.tagname
+mach          = ET.SubElement(testrecord,'mach').text=machine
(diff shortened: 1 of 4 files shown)


scripts/Tools/testreporter.py
```diff
@@ -0,0 +1,240 @@
+#!/usr/bin/env python
+
+"""
+
+Simple script to populate CESM test database with test results.  This can be 
+run with out any knowledge of CIME.
+
+"""
+
+import os
```
jedwards4b: Should use 
from Tools.standard_script_setup import *
which will import argparse and glob as well as some standard arguments.

scripts/Tools/testreporter.py
```diff
@@ -0,0 +1,240 @@
...
+                        help="Do a dry run, database will not be populated.")
+parser.add_argument("--dumpxml",action="store_true", 
+                        help="Dump XML test results to sceen.")
+args = parser.parse_args()
+
+# Fill in values needed from xml files (COMPILER, MPILIB, MACH, BASELINE_NAME_CMP.
+
+os.chdir(args.testroot)
+
+xml_file=glob.glob("*"+args.testid+"/env_build.xml")
```
jedwards4b: Please use CIME/XML/env_build.py for this.

jedwards4b: You should not use ET directly - if the CIME/XML interfaces are not adequate that's a bug in the interface that should be fixed.

scripts/Tools/testreporter.py
```diff
@@ -0,0 +1,240 @@
...
+        value = entry.get('value')
+     if test_id == "BASELINE_NAME_CMP":
+        baseline=value
+
+
+#
+# Create XML
+#
+
+testrecord    = ET.Element("testrecord")
```
jedwards4b: This code should be in a module in CIME/XML

scripts/Tools/testreporter.py
```diff
@@ -0,0 +1,240 @@
...
+
+#
+# Convert XML to a string
+#
+xmlstr = ET.tostring(testrecord,method="xml",encoding="UTF-8")
+
+#
+# Make the XML string human readable and print it out
+#
+xml=minidom.parseString(xmlstr)
```
jedwards4b: functionality provided by genericxml.py

scripts/Tools/testreporter.py
fischer-ncar: I wanted to keep testreporter independent of CIME for the times I populate the testdb later, after I've removed the source code.  Granted this isn't going to happen very often, if ever.  But I still have the problem of finding the CIMEROOT.  I'll either have to get CIMEROOT out of software_environment.txt, or parse env_case.xml.  Now that I think about it, I could set it up like cs.status.  cs.status copies the template to the testing directory, that then calls the cs.status script in CIME.

scripts/Tools/testreporter.py
fischer-ncar: Using GenericXML.write now.

scripts/Tools/testreporter.py
fischer-ncar: Created a CIME/XML/test-reporter.py module to interface with the xml string.

scripts/Tools/testreporter.py
fischer-ncar: Using en_build.py, env_case.py, and env_test.py to get values out of xml files.

scripts/Tools/testreporter.py
fischer-ncar: Now using Tools.standard_script_setup import *
glob isn't a part of the standard script setup.

scripts/lib/CIME/XML/test_reporter.py
```diff
@@ -0,0 +1,120 @@
...
+        tlelem.append(elem)
+        elem   = ET.Element('baselinetag')
+        elem.text   = baseline
+        tlelem.append(elem)
+    
+        self.root=tlelem
+
+
+
+    def dumpxml(self):
```
jedwards4b: why have an additional function here?  Just call GenericXML.write directly or if you really want the default outfile name at least call the inherited function write instead of dumpxml.

scripts/Tools/testreporter.py
```diff
@@ -0,0 +1,203 @@
+#!/usr/bin/env python
```
jedwards4b: Follow the template of other python scripts in cime.   Each script has a test to see if it was called directly
or from another python, and it has a parse_command_line subroutine which includes code for the standard cime options.   I think that create_clone provides a simple example.

scripts/Tools/testreporter.py
```diff
@@ -0,0 +1,203 @@
...
+
+#
+# Create lists on tests based on the testid in the testroot directory.
+#
+test_names=glob.glob("*"+args.testid)
+#
+# Loop over all tests and parse the test results
+#
+test_status={}
+for test_name in test_names:
```
fischer-ncar: The 3 "tests" above are not tests.  They're pulling out information needed to populated the test database.  It is possible for a test to fail without creating any env*.xml files.  So, testing env*xml files would prove nothing.  However, a TestStatus file  is always created.  So, I'll replace the check for cs.status+test_id, with a test for the existence of test_name/TestStatus.  That way I'll account for other files in the testroot directory which end with the testid string.

scripts/Tools/testreporter.py
fischer-ncar: Followed the template for python scripts.  create_clone was a good simple example.

PR #1300, opened 2017-04-03
Set a default value for esmf_logging

### LINKED ISSUE #1245: Call to ESMF_initialize should allow for no PET logfiles
While the default ESMF_initialize is a stub, when building with the ESMF library, a logfile is written for every PET. I propose that we add a driver namelist variable (drv_esmf_logkind?) to control logging by ESMF. The input would be ignored for the stub.

gold2718:
Set a default value for esmf_logging so it does not have to appear in drv_in

This change allows drv_in to be optional (helps with unit tests).

Test baseline: NA
Test namelist changes: NA
Test status: bit for bit

Fixes #1245 

User interface changes?: esmf_logging (and therefore drv_in) are now optional

Code review:

### DIFF
--- a/src/drivers/mct/shr/seq_comm_mct.F90
+++ b/src/drivers/mct/shr/seq_comm_mct.F90
@@ -359,7 +359,7 @@ contains
        cpl_pestride = 1
        cpl_nthreads = 1
 
-       esmf_logging = 'unset'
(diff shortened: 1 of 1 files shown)


billsacks:
Looks good to me. Thanks, @gold2718 , for making this change!

I see Jim already merged it, but I'll add my after-the-fact approval....

PR #1305, opened 2017-04-05
avoid a problem with pio_typename in clone with keepexe

### LINKED ISSUE #1304: create_clone cannot be called without case_setup
If create_clone is called before case.setup is called on a case bad things can happen.   
For example #1272 where the second case does not have the pio_typename set correctly.
I think that the correct action for this is to require case_setup is called on case1 before create_clone can be.  

This would require refactoring the system_tests_compare_two script so that case1 setup is done before the case clone.

jedwards4b:
Print a warning if create_clone is run before case1 is built.   Add code to system_tests_compare_two to update pio_typename in case2 after case1 build is complete.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1304 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -1009,7 +1009,7 @@ class Case(object):
             newcase.set_value("BUILD_COMPLETE","TRUE")
             orig_bld_complete = self.get_value("BUILD_COMPLETE")
             if not orig_bld_complete:
-                logger.warn("\nWARNING: Creating a clone before building the original case may cause PIO_TYPENAME to be invalid in the clone")
+                logger.warn("\nWARNING: Creating a clone with --keepexe before building the original case may cause PIO_TYPENAME to be invalid in the clone")
                 logger.warn("Avoid this message by building case one before you clone.\n")
         else:
             newcase.set_value("BUILD_COMPLETE","FALSE")
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -1007,6 +1007,10 @@ class Case(object):
             orig_exeroot = self.get_value("EXEROOT")
             newcase.set_value("EXEROOT", orig_exeroot)
             newcase.set_value("BUILD_COMPLETE","TRUE")
+            orig_bld_complete = self.get_value("BUILD_COMPLETE")
+            if not orig_bld_complete:
+                logger.warn("\nWARNING: Creating a clone before building the original case may cause PIO_TYPENAME to be invalid in the clone")
+                logger.warn("Avoid this message by building case one before you clone.\n")
         else:
             newcase.set_value("BUILD_COMPLETE","FALSE")
 
--- a/scripts/create_clone
+++ b/scripts/create_clone
@@ -25,7 +25,7 @@ def parse_command_line(args):
                         "is assumed to be under then current working directory ")
 
     parser.add_argument("--keepexe", "-keepexe", action="store_true",
-                        help="Sets EXEROOT to point to original build")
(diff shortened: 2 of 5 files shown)


scripts/lib/CIME/case.py
```diff
@@ -1006,6 +1006,10 @@ def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None, cime
             orig_exeroot = self.get_value("EXEROOT")
             newcase.set_value("EXEROOT", orig_exeroot)
             newcase.set_value("BUILD_COMPLETE","TRUE")
+            orig_bld_complete = self.get_value("BUILD_COMPLETE")
+            if not orig_bld_complete:
+                logger.warn("\nWARNING: Creating a clone before building the original case may cause PIO_TYPENAME to be invalid in the clone")
```
jedwards4b: The warning is only printed if you use --keepexe

jedwards4b: I've updated the warning to include --keepexe

billsacks: Ah, sorry I missed that (the conditional was just above github's context - now I see that). Anyhow, I like your new wording - thanks.

jedwards4b:
@gold2718: I have fixed the issue of checking against valid_values.    
On the second issue, if the clone is created after case.build then the clone valid_values are set correctly however if the clone is created prior to building the first case I can't think of any way
to update the valid_values in the clone.

gold2718:
The best solution would be to set `valid_values` to be correct during `create_newcase` since the compiler (and in principle it's limitations) are known then. That is, while having buildlib.pio act as a check, I feel we should have this information in config_compilers.xml.
For instance, for Hobart, there is no `PNETCDF_PATH` tag. This should equate to pnetcdf not showing up in valid_values. Similarly, it would be good to have a tag that specified NetCDF supported types (e.g., netcdf4p).

jedwards4b:
Whether the user has supplied the proper netcdf and pnetcdf libraries is not a compiler limitation and is determined during the pio build phase.  I would ask that you accept this PR as an improvement and open another issue if you feel that this issue has not been fully addressed.

gold2718:
Maybe it's not strictly a compiler limitation (bad phrasing on my part) but it is a feature of the CIME support for a machine (e.g., `NETCDF_PATH` and `PNETCDF_PATH`).
My preference is that if we are not going to fully address this issue (and I would like feedback from others, e.g, @mvertens) that we do not allow `./create_clone --keepexe` if the case is not built.
My justification is that `--keepexe` makes no sense if there is no exe to keep.
@mvertens? @billsacks?

jedwards4b:
I did consider that but it would require rewriting the system_test_compare_two.

billsacks:
Hmmm, tricky. I sympathize with @gold2718 's point here, but @jedwards4b raises a good point that @gold2718 's suggestion:

> we do not allow ./create_clone --keepexe if the case is not built

will require a massive overhaul of `system_test_compare_two`. I could also see this interfering with some people's desired workflow (maybe).

So my conclusion would be: For now, I'd say the current warning is sufficient; if this seems to bite people in practice, then we should revisit it.

mvertens:
@gold2718 @jedwards4b @billsacks 

At this point I agree with Bill. I think that an additional statement in create_clone --help stating that for the --keepexe option you MUST have built the reference executable first would be really helpful. Could we please add that new documentation to this PR and then go ahead and merge it.

gold2718:
Sorry, I keep getting interrupted.
I will create an issue with my current thinking on how to avoid leaving this hole open.

jedwards4b:
Rerunning scripts regression tests now, will merge upon completion.

PR #1316, opened 2017-04-06
Better handling of poorly implemented batch systems

### LINKED ISSUE #1270: Batch error messages not passed to user.
"Submitting a second job to the debug queue on Titan (where only one is allowed) generates a " " error message. Before it passed through the system error message (that there was already a debug job in the queue)."

This is a regression from 5.1. to 5.2.

jgfouca:
Test suite: by-hand, code-checker
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1270 

User interface changes?: None

Code review: @rljacob @jedwards4b

### DIFF
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -384,7 +384,7 @@ class EnvBatch(EnvBase):
                 submitcmd += string + " "
 
         logger.info("Submitting job script %s"%submitcmd)
-        output = run_cmd_no_fail(submitcmd)
+        output = run_cmd_no_fail(submitcmd + " 2>&1")
         jobid = self.get_job_id(output)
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/XML/env_batch.py
```diff
@@ -384,7 +384,7 @@ def submit_single_job(self, case, job, depid=None, no_batch=False, batch_args=No
                 submitcmd += string + " "
 
         logger.info("Submitting job script %s"%submitcmd)
-        output = run_cmd_no_fail(submitcmd)
+        output = run_cmd_no_fail(submitcmd + " 2>&1")
```
jedwards4b: same comment as 1317 - maybe just do this inside the run_cmd subroutine?

PR #1317, opened 2017-04-06
Better content in TestStatus.log file.

### LINKED ISSUE #1291: Remove python stack trace when a model run fails
When an ACME/CESM model run fails, the log files (TestStatus.log) contain
a python stack trace that is not very useful to the user (The failure is in
the model run, not in the CIME python framework). 
It would be useful to remove the python stack trace and include a comment
to the location of the log file that the user should look at to diagnose the
failure.

jgfouca:
1) There is no reason to print a full stacktrace when the model
fails. That just clutters the log file and provides no debugging value.
2) In some cases, create_test was losing stderr output. Now we redirect
stderr to stdout so it's never lost.
3) In some cases, the TestStatus claimed the RUN phase both passed
and failed. What actually happened was that the submit passed and the
run failed. The situation is now more clear.

Test suite: scripts-regression-tests --fast
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1291 

User interface changes?: Improved formatting/content of TestStatus.log

Code review: @jedwards4b


scripts/lib/CIME/test_scheduler.py
```diff
@@ -325,25 +325,25 @@ def _update_test_status(self, test, phase, status):
     def _shell_cmd_for_phase(self, test, cmd, phase, from_dir=None):
     ###########################################################################
         while True:
-            rc, output, errput = run_cmd(cmd, from_dir=from_dir)
+            rc, output, _ = run_cmd(cmd + " 2>&1", from_dir=from_dir)
```
jedwards4b: You can also do this with arg_stderr=subprocess.STDOUT,  it requires that you import subprocess but I think I prefer it to 2>&1

PR #1333, opened 2017-04-11
fix blues NETCDFROOT unresolved error.

### LINKED ISSUE #1309: case.setup broken on blues.
Using latest master.
```
./create_newcase -case Xcaseintel -compset X -res f19_g16 --compiler intel
cd Xcaseintel
blogin3[104]: ./case.setup 
ERROR: Undefined env var 'NETCDFROOT'
Exit 1
```

sarich:
NETCDFROOT as used here might not be stored in the environment dictionary yet, use full
path name (ESMCI/cime issue 1309)

Test baseline: 
Test namelist changes: 

Fixes #1309 

User interface changes?: None

Code review:

### DIFF
--- a/config/acme/machines/config_machines.xml
+++ b/config/acme/machines/config_machines.xml
@@ -900,8 +900,8 @@
       <env name="NETCDFROOT">/soft/climate/netcdf/4.3.3.1c-4.2cxx-4.4.2f-serial/intel-15.0.1</env>
       <env name="NETCDF_INCLUDES">/soft/climate/netcdf/4.3.3.1c-4.2cxx-4.4.2f-serial/intel-15.0.1/include</env>
       <env name="NETCDF_LIBS">/soft/climate/netcdf/4.3.3.1c-4.2cxx-4.4.2f-serial/intel-15.0.1/lib</env>
-      <env name="PATH">$ENV{NETCDFROOT}/bin:$ENV{PATH}</env>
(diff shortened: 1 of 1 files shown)


jgfouca:
@sarich , I don't know if we want to fix #1309 in this way. Hardcoded path is not very robust. Also, this pattern of using ENV to for netcdf root works fine on all our other machines. I suspect that the blues modules aren't setting it at all.

jayeshkrishna:
Blues uses soft and does not set NETCDFROOT. The best solution is for us to ask the systems folks to set NETCDF_PATH when user load the soft environment for netcdf etc.
@sarich : Correct me if I am wrong, but it looks like this PR is about using env set in config_machines via env tags within the same section (set NETCDFROOT via env tag, and using the value of NETCDFROOT in a subsequent tag to set PATH)

sarich:
@jayeshkrishna, I don't think there is a soft key for this version of netcdf, which is why the values are hard-coded.

> Correct me if I am wrong, but it looks like this PR is about using env set in config_machines via env tags within the same section (set NETCDFROOT via env tag, and using the value of NETCDFROOT in a subsequent tag to set PATH)
You are correct, the variables cannot be set and then also used within the same <environment> block

jayeshkrishna:
And yes, we don't have a soft key for the version of netcdf build used here. We should look into upgrading the versions of netcdf in blues (this requires some testing to make sure that these new libraries don't break anything).

rljacob:
We currently can run anything on blues with the current master and I would like a quick fix done now while we work on the correct fix.

jgfouca:
OK, then I approve.

PR #1337, opened 2017-04-11
validate the case and test_id arguments

### LINKED ISSUE #611: Change in default test id causes namelist failures
Recently, the following change went into `create_test` (line 260):

```
-        args.test_id = CIME.utils.get_timestamp()
+        args.test_id = "%s.%s"%(CIME.utils.get_timestamp(), CIME.utils.id_generator())
```

This changes the case name in a generated test from `TEST.RES.COMPSET.MACH_COMPILER.TESTMODS.ACTION.TESTID` to `TEST.RES.COMPSET.MACH_COMPILER.TESTMODS.ACTION.TESTID.MORETESTID` and the namelist comparison doesn't consider the last `.` as part of the `TESTID`. The result is unexpected failures in the namelist compare, as in the following example (from `SMS.T62_g37.CECO.hobart_nag.pop-ecosys.GC.20160929_143649.t7r5a0`):

```
Differences in namelist 'diagnostics_nml':
  BASE: diag_outfile = SMS.T62_g37.CECO.hobart_nag.pop-ecosys.ACTION.TESTID.pop.dd'
  COMP: diag_outfile = SMS.T62_g37.CECO.hobart_nag.pop-ecosys.ACTION.TESTID.t7r5a0.pop.dd'
  BASE: diag_transport_outfile = SMS.T62_g37.CECO.hobart_nag.pop-ecosys.ACTION.TESTID.pop.dt'
  COMP: diag_transport_outfile = SMS.T62_g37.CECO.hobart_nag.pop-ecosys.ACTION.TESTID.t7r5a0.pop.dt'
  BASE: diag_velocity_outfile = SMS.T62_g37.CECO.hobart_nag.pop-ecosys.ACTION.TESTID.pop.dv'
  COMP: diag_velocity_outfile = SMS.T62_g37.CECO.hobart_nag.pop-ecosys.ACTION.TESTID.t7r5a0.pop.dv'
```

A simple fix would be to replace the `.` with `_` (or some other character):

```
-        args.test_id = "%s.%s"%(CIME.utils.get_timestamp(), CIME.utils.id_generator())
+        args.test_id = "%s_%s"%(CIME.utils.get_timestamp(), CIME.utils.id_generator())
```

A more complicated fix would be to make sure the `.MORETESTID` is actually included in `TESTID` when running the namelist comparison.

### LINKED ISSUE #853: casename restrictions should be enforced
There are certain characters which cannot be included in a casename, these characters should 
be trapped and an error reported.   This issue is moved from CESM-DEVELOPMENT/cime 38

jedwards4b:
Checks test-id and case name for invalid characters - seeking feedback on other characters that should be included here

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #611 
Fixes #853

User interface changes?: 

Code review:

### DIFF
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -147,7 +147,7 @@ OR
     if model == "cesm":
         run_unsupported = args.run_unsupported
     try:
-        CIME.utils.check_name(args.case, fullpath=args.case.startswith('/'))
+        CIME.utils.check_name(args.case, fullpath=True)
     except:
         expect(False, "Illegal case name argument provided")
 
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -4,6 +4,9 @@ from Tools.standard_script_setup import *
 from CIME.utils         import expect, get_model
 from CIME.case          import Case
 from CIME.check_lockedfiles import lock_file
+import re
+
+invalid_case_name_re = re.compile(r"[\.<>/{}[\]~`@]");
 
 logger = logging.getLogger(__name__)
 
@@ -146,6 +149,10 @@ OR
     run_unsupported = False
     if model == "cesm":
         run_unsupported = args.run_unsupported
+
+    m = invalid_case_name_re.search(args.case)
+    if m is not None:
+        expect(False, "Invalid character in case name %s"%m.group(0))
 
 
     return args.case, args.compset, args.res, args.machine, args.compiler,\
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -146,11 +146,9 @@ OR
     run_unsupported = False
     if model == "cesm":
         run_unsupported = args.run_unsupported
-    try:
-        CIME.utils.check_name(args.case, fullpath=True)
-    except:
-        expect(False, "Illegal case name argument provided")
 
+    expect(CIME.utils.check_name(args.case, fullpath=True),
+           "Illegal case name argument provided")
 
     return args.case, args.compset, args.res, args.machine, args.compiler,\
         args.mpilib, args.project, args.pecount, \
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -4,9 +4,6 @@ from Tools.standard_script_setup import *
 from CIME.utils         import expect, get_model
 from CIME.case          import Case
 from CIME.check_lockedfiles import lock_file
-import re
-
-invalid_case_name_re = re.compile(r"[\.<>/{}[\]~`@]");
 
 logger = logging.getLogger(__name__)
 
@@ -149,10 +146,10 @@ OR
     run_unsupported = False
     if model == "cesm":
         run_unsupported = args.run_unsupported
-
-    m = invalid_case_name_re.search(args.case)
-    if m is not None:
-        expect(False, "Invalid character in case name %s"%m.group(0))
+    try:
+        CIME.utils.check_name(args.case, fullpath=args.case.startswith('/'))
+    except:
+        expect(False, "Illegal case name argument provided")
 
 
     return args.case, args.compset, args.res, args.machine, args.compiler,\
--- a/scripts/lib/CIME/utils.py
+++ b/scripts/lib/CIME/utils.py
@@ -39,13 +39,11 @@ def check_name(fullname, additional_chars=None, fullpath=False):
     writable
 
     >>> check_name("test.id", additional_chars=".")
-    Traceback (most recent call last):
-        ...
-    SystemExit: ERROR: Illegal character . found in name test.id
+    False
     >>> check_name("case.name", fullpath=False)
-
+    True
     >>> check_name("/some/file/path/case.name", fullpath=True)
-
+    True
     """
 
     chars = '<>/{}[\]~`@' # pylint: disable=anomalous-backslash-in-string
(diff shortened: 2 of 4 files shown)


scripts/create_newcase
```diff
@@ -4,6 +4,9 @@ from Tools.standard_script_setup import *
 from CIME.utils         import expect, get_model
 from CIME.case          import Case
 from CIME.check_lockedfiles import lock_file
+import re
+
+invalid_case_name_re = re.compile(r"[\.<>/{}[\]~`@]");
```
jgfouca: Don't need semicolon here.

jedwards4b:
I think that the requirements for the two names may be different.  If we end up with the same list in both places I will move them to CIME.utils and just have a single list.

scripts/create_newcase
jedwards4b: fixed

billsacks: Looks like you removed the semicolon from create_test but not create_newcase

billsacks: periods should definitely be allowed in the case name - e.g., they are used extensively in the case name for tests. But they should not be allowed in the testid, if I remember correctly

billsacks:
I like the changes you made, and feel these address my earlier comments well - thank you.

However, I realized there is a problem with trying to create a case using relative path, which is something I do very often:

```
[yslogin4:/glade/p/work/sacks/cime/scripts]$ mkdir testdir
[yslogin4:/glade/p/work/sacks/cime/scripts]$ ./create_newcase -case testdir/testcase0412a -compset A -res f45_g37_rx1
ERROR: Illegal character / found in name testdir/testcase0412a
None
ERROR: Illegal case name argument provided
  File "./create_newcase", line 150, in parse_command_line
    CIME.utils.check_name(args.case, fullpath=args.case.startswith('/'))
  File "/glade/p/work/sacks/cime/scripts/Tools/../../scripts/lib/CIME/utils.py", line 63, in check_name
    expect(False, "Illegal character %s found in name %s"%(match.group(0), name))
  File "/glade/p/work/sacks/cime/scripts/Tools/../../scripts/lib/CIME/utils.py", line 30, in expect
    raise exc_type("%s %s"%(error_prefix,error_msg))
SystemExit: ERROR: Illegal character / found in name testdir/testcase0412a
ERROR: Illegal case name argument provided
```

I believe this should be allowed. I think the problem is that you determine whether something is a full path based on whether it starts with '/', but that doesn't allow for relative paths.

jedwards4b:
Always assume create_newcase is passing a path and don't check the path argument at all here.
We will rely on python mkdirs to catch errors in the path.

scripts/lib/CIME/utils.py
```diff
@@ -26,11 +26,40 @@ def expect(condition, error_msg, exc_type=SystemExit, error_prefix="ERROR:"):
...
             import pdb
             pdb.set_trace()
-        raise exc_type("%s %s" % (error_prefix,error_msg))
+        logger.error("%s %s", error_prefix,error_msg, exc_info=True)
+        raise exc_type("%s %s"%(error_prefix,error_msg))
 
 def id_generator(size=6, chars=string.ascii_lowercase + string.digits):
     return ''.join(random.choice(chars) for _ in range(size))
 
+def check_name(fullname, additional_chars=None):
```
jgfouca: Thanks, this looks better. One thing that may be good is to have check_name return a true/false and then let the user decide if an exception is needed.

scripts/lib/CIME/utils.py
```diff
@@ -26,11 +26,40 @@ def expect(condition, error_msg, exc_type=SystemExit, error_prefix="ERROR:"):
...
+
+    chars = '<>/{}[\]~`@' # pylint: disable=anomalous-backslash-in-string
+    if additional_chars is not None:
+        chars += additional_chars
+    if fullpath:
+        _, name = os.path.split(fullname)
+    else:
+        name = fullname
+    match = re.search(r"["+re.escape(chars)+"]", name)
+    if match is not None:
```
jgfouca:
```
return match is None
```

scripts/create_newcase
```diff
@@ -146,6 +146,10 @@ OR
     run_unsupported = False
     if model == "cesm":
         run_unsupported = args.run_unsupported
+    try:
```
jgfouca:
```
expect(CIME.utils.check_name(args.case, fullpath=True), "Illegal case name argument provided")
```

billsacks:
Looks good to me now – thanks!

scripts/lib/CIME/utils.py
```diff
@@ -26,11 +26,40 @@ def expect(condition, error_msg, exc_type=SystemExit, error_prefix="ERROR:"):
...
             import pdb
             pdb.set_trace()
-        raise exc_type("%s %s" % (error_prefix,error_msg))
+        logger.error("%s %s", error_prefix,error_msg, exc_info=True)
+        raise exc_type("%s %s"%(error_prefix,error_msg))
 
 def id_generator(size=6, chars=string.ascii_lowercase + string.digits):
     return ''.join(random.choice(chars) for _ in range(size))
 
+def check_name(fullname, additional_chars=None, fullpath=False):
```
jgfouca: Thanks, this looks better. One thing that may be good is to have check_name return a true/false and then let the user decide if an exception is needed.

PR #1343, opened 2017-04-13
Unit testing: Determine machine-specific settings automatically

### LINKED ISSUE #872: Get some unit testing defaults from config_machines
_From @billsacks on August 4, 2015 17:6_

With the changes here:

https://github.com/CESM-Development/cime/pull/87

a number of machine-specific settings need to be explicitly specified when running run_tests.py; these should be obtained from config_machines:
- mpilib
- maybe the mpirun command
- maybe the default compiler for a machine

@quantheory suggested this; he and @jedwards4b seem like the ones who would know how to do this.

_Copied from original issue: CESM-Development/cime#92_

billsacks:
This PR allows you to run the unit tests by simply running
tools/unit_testing/run_tests.py without any additional arguments. The
machine, compiler, mpilib, and mpirun command are now determined
automatically. In addition, use-openmp is now assumed by default, since
that is the standard way we build pFUnit (but there is a mechanism to
turn off use-openmp).

This is important in order to support unit testing on additional
machines.

In order to accomplish this, I needed to tweak some pieces that are also
used by system builds:

(1) I added a unit_testing attribute to the mpirun command in
    config_machines.xml. This is needed on yellowstone, where we use a
    complex command to launch CESM, but just want to use a simple
    'mpirun.lsf' to launch unit tests.

(2) In machines.py: probe_machine_name: If no name is found with
    socket.getfqdn(), then try socket.gethostname(). This is needed in
    order to determine the machine when running on my laptop.

   Also ran unit tests on yellowstone using the trimmed-down command
   documented in README.unit_testing. And ran unit tests on my laptop
   using the same command.
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes #872

User interface changes?: Simplifies the typical call to run_tests.py

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/machines.py
+++ b/scripts/lib/CIME/XML/machines.py
@@ -95,8 +95,34 @@ class Machines(GenericXML):
         Find a matching regular expression for hostname
         in the NODENAME_REGEX field in the file.   First match wins.
         """
-        machine = None
+
+        names_not_found = []
+
         nametomatch = socket.getfqdn()
+        machine = self._probe_machine_name_one_guess(nametomatch)
+
+        if machine is None:
+            names_not_found.append(nametomatch)
+
+            nametomatch = socket.gethostname()
+            machine = self._probe_machine_name_one_guess(nametomatch)
+
+            if machine is None:
+                names_not_found.append(nametomatch)
+
+                names_not_found_quoted = ["'" + name + "'" for name in names_not_found]
+                names_not_found_str = ' or '.join(names_not_found_quoted)
+                logger.warning("Could not find machine match for %s" % names_not_found_str)
+
+        return machine
+
+    def _probe_machine_name_one_guess(self, nametomatch):
+        """
+        Find a matching regular expression for nametomatch in the NODENAME_REGEX
+        field in the file. First match wins. Returns None if no match is found.
+        """
+
+        machine = None
         nodes = self.get_nodes("machine")
 
         for node in nodes:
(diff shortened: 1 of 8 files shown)


scripts/lib/CIME/XML/machines.py
```diff
@@ -95,8 +95,34 @@ def probe_machine_name(self):
         Find a matching regular expression for hostname
         in the NODENAME_REGEX field in the file.   First match wins.
         """
-        machine = None
+
+        names_not_found = []
+
         nametomatch = socket.getfqdn()
+        machine = self._probe_machine_name_one_guess(nametomatch)
```
jedwards4b: Have you tested that this still works with the --machine command line option?

scripts/lib/CIME/case.py
```diff
@@ -1076,7 +1077,8 @@ def get_mpirun_cmd(self, job="case.run"):
         mpi_attribs = {
             "compiler" : self.get_value("COMPILER"),
             "mpilib"   : self.get_value("MPILIB"),
-            "threaded" : self.get_build_threaded()
+            "threaded" : self.get_build_threaded(),
+            "unit_testing" : False
```
jedwards4b: This attribute shouldn't be required, putting it into the case.py may cause confusion?

billsacks: Without that, create_newcase was failing due to this code in `get_mpirun`:

```python
            for key in xml_attribs:
                expect(key in attribs, "Unhandled MPI property '%s'" % key)
```

scripts/lib/CIME/XML/machines.py
billsacks: I hadn't before, but I just did: From yellowstone, I tested both --machine yellowstone and --machine cheyenne, and confirmed that MACH was set correctly.

scripts/lib/CIME/case.py
jedwards4b: I see - I think that this logic is messed up a little - we should get the list of allowed attributes from the config_machines.xsd file and check against that - but that's another issue.

PR #1347, opened 2017-04-13
add support for user_mods in compset definition

jedwards4b:
Allows for the specification of user_mods in the compset definition.   If both compset definition and
command line argument (--user-mods-dir) are specified then both are applied with command line applied first and compset second.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Code review:

### DIFF
--- a/config/xml_schemas/config_compsets.xsd
+++ b/config/xml_schemas/config_compsets.xsd
@@ -11,6 +11,7 @@
   <xs:element name="alias" type="xs:NCName"/>
   <xs:element name="lname" type="xs:string"/>
   <xs:element name="science_support"/>
+  <xs:element name="user_mods" type="xs:string"/>
 
   <!-- complex elements -->
 
@@ -32,6 +33,7 @@
         <xs:element ref="alias"/>
         <xs:element ref="lname"/>
 	<xs:element ref="science_support" minOccurs="0" maxOccurs="unbounded"/>
+	<xs:element ref="user_mods" minOccurs="0" maxOccurs="1"/>
       </xs:sequence>
       <xs:attribute ref="grid"/>
     </xs:complexType>
--- a/scripts/lib/CIME/XML/compsets.py
+++ b/scripts/lib/CIME/XML/compsets.py
@@ -36,8 +36,10 @@ class Compsets(GenericXML):
                 for node in science_support_nodes:
                     science_support.append(node.get("grid"))
 
+                user_mods = self.get_optional_node("user_mods", root=node)
+
                 logger.debug("Found node match with alias: %s and lname: %s" % (alias, lname))
-                return (lname, alias, science_support)
+                return (lname, alias, science_support, user_mods)
         return (None, None, [False])
 
     def get_compset_var_settings(self, compset, grid):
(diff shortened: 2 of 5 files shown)


billsacks:
Thanks, @jedwards4b , for implementing this. I have two thoughts:

First, I feel it's important that, if a user_mods is included in a compset, then that information should be printed to stdout when creating a case, and also echoed to the README.case file where (IIRC) the compset information is given. The reason is: If someone wants to create a one-off user-compset based on one of these compsets, then it's important for them to know that this compset includes a user_mods directory.

Second: I think some code mods are needed in user_mod_support.py in order to be consistent about which mods take precedence: the compset's mods or the mods specified on the command line. Personally, I feel that the mods specified on the command line should take precedence, because the user explicitly asked for those. However, the most important thing is consistency between user_nl files, shell_commands and SourceMods in terms of which takes precedence.

We ( @jedwards4b @mnlevy1981 and myself) spent some time making sure user_mod_support consistently enforces the rule that a given user_mods directory takes precedence over any mods that it includes. However, I don't think we considered the possibility that apply_user_mods may be called twice. If I understand the code correctly:

- For shell_commands, the second application overwrites values set by the first - i.e., second takes precedence

- For SourceMods, the first application remains in place, second generates a warning - i.e., first takes precedence

- For user_nl files: We currently prepend and the last instance of a variable is what matters, so the first application takes precedence.

However, if we implemented the changes suggested by this comment:

```python
    # it may be desireable to reverse include_dirs above the
    # previous loop and then append user_nl changes rather than prepend them
```

then that would change the precedence of user_nl files.

I would suggest the following:

1. Modify user_mod_support as suggested in the comment, so that:

    a. we reverse include_dirs as soon as it's created

    b. we change update_user_nl_file to append changes rather than prepend them

2. We change the application of SourceMods in user_mod_support so that existing files are overwritten, rather than keeping the existing file in place.

3. Reverse the order of this new line in case.py:

```python
for user_mods in (user_mods_dir, self._user_mods):
```

so that we apply the command line second. (However, I don't feel strongly about this last point.)

With these changes, I believe we would maintain the current behavior that a given user_mods directory takes precedence over any mods that it includes, and we would have the consistent rule that the second call to apply_user_mods takes precedence over the first - in terms of user_nl files, shell_commands and SourceMods. However, we should test (both for include dirs and for double-application of apply_user_mods) to make sure my understanding is correct.

billsacks:
@jedwards4b and @mvertens - I have added a commit where I think I have this working right in terms of precedence. This also fixes the problems that Mariana and I ran into (at least, I think it does: Mariana alluded to some other problem that I'm unclear about). I still want to do a bit more testing; I'll probably set up unit tests for apply_user_mods to make sure it's handling various recursive includes correctly. But I have at least tested the primary use case this PR is targeting.

billsacks:
Okay, I have finished addressing and testing the second point in my big comment. Someone should still address this one:

> First, I feel it's important that, if a user_mods is included in a compset, then that information should be printed to stdout when creating a case, and also echoed to the README.case file where (IIRC) the compset information is given. The reason is: If someone wants to create a one-off user-compset based on one of these compsets, then it's important for them to know that this compset includes a user_mods directory.

And, as @mvertens pointed out, my changes need a get_resolved_value (?) somewhere.

jedwards4b:
Superceeded by #1366

PR #1350, opened 2017-04-13
Fixes for PRE test

### LINKED ISSUE #1112: Figure out why PRE.*.ADESP tests run forever on desktops
Once the issue is resolved, re-enable this test in cime_developer suite.

gold2718:
New method for setting and detecting pause alarms in component models
 Refactor access to components which are paused
 Pause works with the driver and at least one active component (POP).
 Active component (POP) able to correctly process resume signal.
 PRE test refactored to be more flexible.
 Modify code and scripts to error if NINST_ESP > 1.
 Fixed NCK test to not set NINST_ESP > 1
 Added PRE test back to cime_developer.

Test suite: scripts_regression_tests.py on Yellowstone (CESM) and compute001 (ACME)
                   CESM prealpha on hobart_nag
Test baseline: NA
Test namelist changes: NA
Test status: bit for bit

Fixes #1112 

User interface changes?: NA

Code review:

### DIFF
--- a/scripts/lib/CIME/SystemTests/pre.py
+++ b/scripts/lib/CIME/SystemTests/pre.py
@@ -2,6 +2,7 @@
 Implementation of the CIME pause/resume test: Tests having driver
 'pause' (write cpl restart file) and 'resume' (read cpl restart file)
 without changing restart file. Compare to non-pause/resume run.
+Test can also be run with other component combinations.
 
 """
 
@@ -19,8 +20,9 @@ class PRE(SystemTestsCompareTwo):
 ###############################################################################
     """
     Implementation of the CIME pause/resume test: Tests having driver
-    'pause' (write cpl restart file) and 'resume' (read cpl restart file)
-    without changing restart file. Compare to non-pause/resume run.
+    'pause' (write cpl and/or other restart file(s)) and 'resume'
+    (read cpl and/or other restart file(s)) without changing restart
+    file. Compare to non-pause/resume run.
     """
 
     ###########################################################################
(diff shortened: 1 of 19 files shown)


scripts/lib/CIME/SystemTests/pre.py
```diff
@@ -31,26 +33,36 @@ def __init__(self, case):
                                        run_two_suffix='pr',
                                        run_one_description='no pause/resume',
                                        run_two_description='pause/resume')
+        self._stopopt  = 'ndays'
```
jedwards4b: Should these be in config_tests.xml and not here?

src/drivers/mct/cime_config/buildnml
```diff
@@ -23,6 +23,24 @@ from CIME.XML.files import Files
 logger = logging.getLogger(__name__)
 
 ###############################################################################
+def _get_time_in_seconds(time, unit):
```
jedwards4b: Isn't there already a function in CIME.utils for this?  If not wouldn't that be a better place for it?  Perhaps all of the time utilities should be moved to a new time_utils.py file.

scripts/lib/CIME/SystemTests/pre.py
gold2718: They were really placeholders. I modified the values to just initialize the variables in `__init__`, the working values do come from config_tests.xml.

src/drivers/mct/cime_config/buildnml
gold2718: I could not find such a function so I moved this function from buildnml to CIME.utils.

gold2718:
I believe I have addressed @jedwards4b useful comments (and rerun scripts_regression_tests).

PR #1354, opened 2017-04-14
Set rearr opts using a pio interface

### LINKED ISSUE #1296: pio_rearr_opts should not be exposed in shr_pio_mod.F90
In shr_pio_mod.F90 the internals of pio derived datatype pio_rearr_opt_t are used directly.
This creates an issue in which the internal data structures of pio are limited by usage in the model.
Settings for these data structures should happen internal to pio so that these internal data types are not exposed.

jayeshkrishna:
Instead of filling in and passing pio_rearr_opt_t type in pio_init,
we now use pio_set_rearr_opts() interface to set the rearranger
options.

Test suite: X case
Test baseline: NA
Test namelist changes: None
Test status: bit for bit

Fixes #1296

User interface changes?: None

Code review: @jedwards4b

### DIFF
--- a/src/share/util/shr_pio_mod.F90
+++ b/src/share/util/shr_pio_mod.F90
@@ -52,7 +52,11 @@ module shr_pio_mod
   integer, allocatable :: io_compid(:)
   integer :: pio_debug_level=0, pio_blocksize=0
   integer(kind=pio_offset_kind) :: pio_buffer_size_limit=-1
-  type(pio_rearr_opt_t)  :: pio_rearr_opts
+  integer :: pio_rearr_opt_comm_type, pio_rearr_opt_fcd
+  logical :: pio_rearr_opt_c2i_enable_hs, pio_rearr_opt_c2i_enable_isend
+  integer :: pio_rearr_opt_c2i_max_pend_req
+  logical :: pio_rearr_opt_i2c_enable_hs, pio_rearr_opt_i2c_enable_isend
+  integer :: pio_rearr_opt_i2c_max_pend_req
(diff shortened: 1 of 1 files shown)


jayeshkrishna:
Tested with an X case + (pio1 and pio2) on compute001.

src/share/util/shr_pio_mod.F90
```diff
@@ -177,7 +182,18 @@ subroutine shr_pio_init2(comp_id, comp_name, comp_iamin, comp_comm, comp_comm_ia
 
     if(pio_async_interface) then
 #ifdef PIO1
-       call pio_init(total_comps,mpi_comm_world, comp_comm, io_comm, iosystems, rearr_opts=pio_rearr_opts)
+       call pio_init(total_comps,mpi_comm_world, comp_comm, io_comm, iosystems)
```
jedwards4b: Shouldn't we be able to remove this #ifdef PIO1 now?

jayeshkrishna: We only have the "#ifdef PIO1" around the async interface. In PIO2 the async interface does not work in Fortran (for a while Ed was working on changing the async C interface, but I think now we should be able to implement the async fortran interface in pio2)

jayeshkrishna: But I think you are right, we should be able to get rid of the ifdef now because pio_set_rearr_opts() should work for async cases too (Although pio_init() does not
work for async+pio2).

jayeshkrishna: A PR is coming up soon

PR #1356, opened 2017-04-14
Fixes data model when using standard netcdf

### LINKED ISSUE #1355: Data model using incorrect setting for standard netcdf
The `if (pio_iotype == iotype_std_netcdf) then ` in [shr_dmodel_mod.F90#L774](https://github.com/ESMCI/cime/blob/master/src/share/util/shr_dmodel_mod.F90#L774) 
should be replaced by `if (pio_iotype == pio_iotype_netcdf) then `. 
See the following [comment](https://github.com/ESMCI/cime/issues/1286#issuecomment-294225625).

bishtgautam:
Uses pio_iotype_netcdf to determine if standard netcdf is
to be used by the data model

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes #1355

User interface changes?: 

Code review:

### DIFF
--- a/src/share/util/shr_dmodel_mod.F90
+++ b/src/share/util/shr_dmodel_mod.F90
@@ -60,8 +60,6 @@ module shr_dmodel_mod
   integer(IN),parameter,public :: shr_dmodel_gGridCompareMaskZerosMask  = 106 ! masks have same zeros
   integer(IN),parameter,public :: shr_dmodel_gGridCompareMaskSubsetMask = 107 ! mask is subset of other
 
-  integer(IN),parameter,public :: iotype_std_netcdf = -99 ! non pio option
-
 !--------------------------------------------------------------------------
 ! Private data
 !--------------------------------------------------------------------------
@@ -771,7 +769,7 @@ subroutine shr_dmodel_readstrm(stream, pio_subsystem, pio_iotype, pio_iodesc, gs
 
   call t_stopf(trim(lstr)//'_setup')
 
-  if (pio_iotype == iotype_std_netcdf) then
+  if (pio_iotype == pio_iotype_netcdf) then
 
      call t_startf(trim(lstr)//'_readcdf')
      if (my_task == master_task) then


bishtgautam:
This is my first time issue a PR for CIME, what tests do I need to run?

src/share/util/shr_dmodel_mod.F90
```diff
@@ -771,7 +769,7 @@ subroutine shr_dmodel_readstrm(stream, pio_subsystem, pio_iotype, pio_iodesc, gs
 
   call t_stopf(trim(lstr)//'_setup')
 
-  if (pio_iotype == iotype_std_netcdf) then
+  if (pio_iotype == pio_iotype_netcdf) then
```
jedwards4b: I'm afraid I gave you bad information, by setting this to -99 the logic was that if the PIO_TYPENAME was not explicitly set for this component bypass pio and use netcdf directly.   But in recent versions we always set the pio_typename and so this code is never exercised.   The change you've made will cause pio to be bypassed here if you set pio_typename='netcdf' I dont think that we want that.

bishtgautam: Ok, that is fine.

In the future, do you still plan on supporting the case when PIO_TYPENAME is not explicitly set?

jedwards4b: I suggest we leave the logic in place for now unless it's a hinderance to your changes

bishtgautam:
@jedwards4b : I'm attempting to add a new feature to allow data model to read multiple time slices at once and have added the code for `(pio_iotype /= iotype_std_netcdf)`. I'm wondering if it is worth the effort of adding code when `pio_iotype == iotype_std_netcdf` even though PIO_TYPENAME is now always set.

Maybe we can come up with an alternate solution that gets rid of the code within the `if (pio_iotype == iotype_std_netcdf)`. When PIO_TYPENAME isn't explicitly specified, could we use the code within the else block except use `pio_iotype_netcdf` instead of `pio_iotype` in [shr_dmodel_mod.F90#L838](https://github.com/ESMCI/cime/blob/f89e3bb105abd91611fd805cd0e84df0fae656ef/src/share/util/shr_dmodel_mod.F90#L838)? For example:

```
if (pio_iotype == iotype_std_netcdf) then
        rcode = pio_openfile(pio_subsystem, pioid, pio_iotype_netcdf, trim(filename), pio_nowrite)
else
        rcode = pio_openfile(pio_subsystem, pioid, pio_iotype, trim(filename), pio_nowrite)
endif
```

PR #1362, opened 2017-04-16
Force user to always go through case.submit

### LINKED ISSUE #347: RUN_WITH_SUBMIT set to TRUE, but never set to FALSE
This is a minor issue that I just noticed. Caveat: I haven't done any testing; this just looks wrong to me. (This also appears to have been an issue in cime4.)

The RUN_WITH_SUBMIT xml variable is set to TRUE the first time you run case.submit. But it is never set to FALSE. I _think_ this means that the scripts will let you do the following:

```
case.submit
(wait for run to finish)
bsub < case.run
```

whereas they should yell at you for the attempt to resubmit without use of the case.submit script.

I think the fix for this would be, after this block in case_run:

```
    run_with_submit = case.get_value("RUN_WITH_SUBMIT")
    expect (run_with_submit,
            "You are not calling the run script via the submit script. "
            "As a result, short-term archiving will not be called automatically."
            "Please submit your run using the submit script like so:"
            " ./case.submit")
```

insert:

```
case.set_value("RUN_WITH_SUBMIT", False)
```

so that it's reset for the next time around.

jgfouca:
This was already the policy, but was not being enforced correctly.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #347 

User interface changes?: None

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/case_run.py
+++ b/scripts/lib/CIME/case_run.py
@@ -245,6 +245,10 @@ def case_run(case):
            "Please submit your run using the submit script like so:"
            " ./case.submit")
 
+    # Forces user to use case.submit if they re-submit
+    if case.get_value("TESTCASE") is None:
+        case.set_value("RUN_WITH_SUBMIT", False)
+
     prerun_script = case.get_value("PRERUN_SCRIPT")
     postrun_script = case.get_value("POSTRUN_SCRIPT")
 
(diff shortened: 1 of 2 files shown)


jgfouca:
Testing problems, don't merge.

jgfouca:
Looks good now.

rljacob:
Can I still set RUN_WITH_SUBMIT to FALSE if I want too?

jedwards4b:
@rljacob it was never intended that you could.

rljacob:
Why is there a RUN_WITH_SUBMIT variable?

jedwards4b:
It's how we make sure that case.submit was used to call case.run, it was never intended to be a mechanism to override that method.

Issue #842, opened 2016-11-29
Allow CAM compsets "out-of-the-box" to specify CLM namelist options for specific F compsets 


cacraigucar:
Currently the F2000_DEV compset requires three CLM namelist options to be specified in order to run properly.  We want a way to be able to have this CAM compset run out-of-the-box without the user needing to know that a specific CLM user_mods file needs to be included.   Previous email discussions included  @fvitt, @jedwards4b , @mvertens, @cecilehannay, @cacraigucar

-------------------
Previous email discussion:

Mariana Vertenstein <mvertens@ucar.edu>
Oct 13
Reply
to Jim, me, Francis, Cecile 
I think this is the cleanest way forwards  for now. We can start thinking of linking in a user_mods directory to a compset - which we cannot do right now. But this method gives you the greatest flexibility of bringing in namelist changes from other components cleanly in a cam compset.

On Thu, Oct 13, 2016 at 3:18 PM, Jim Edwards <jedwards@ucar.edu> wrote:
I think Cheryl said that they already do it for testing but wanted to extend it to 
user mods.  I suggested user_mods but Cheryl didn't like the idea of having an additional command line argument to create_newcase.

On Thu, Oct 13, 2016 at 3:06 PM, Mariana Vertenstein <mvertens@ucar.edu> wrote:
One more question - do we only want this for testing? That is my understanding - so this could to in testmods.

On Thu, Oct 13, 2016 at 3:04 PM, Mariana Vertenstein <mvertens@ucar.edu> wrote:
I think we need to do this with user_mods for this compset. I think this is the most general way forwards to do this. Let me talk to Jim today/tomorrow before I leave to see if we can come up with a process that would be general moving forwards.

Mariana

On Thu, Oct 13, 2016 at 3:00 PM, Cheryl Craig <cacraig@ucar.edu> wrote:
Mariana-

Cecile is running with three CLM namelist options which are required for F2000_DEV to give the results that she is currently getting.   We want to add these into CAM so that anyone using the F2000_DEV compset gets these settings (but it doesn't affect any B compsets).

I spoke with Jim and he thought that I would not be able to put a CLM_NML_USE_CASE in CAM's config_component.xml as he believes that only the primary component retrieves the use_case from its own config_component.xml.  He thought I'd need to ask CLM to put this change into their trunk, but suggested I ask you.

[...]

PR #1366, opened 2017-04-17
Mvertens/usermods in compset

mvertens:
Can now specify <user_mods> directory as part of compset definition

Before this PR, could only specify the user_mods directory as an input argument to create_newcase. However, as part of bringing SOM aquaplanet functionality to the data ocean model, it became clear that it would be useful to have an alternative mechanism to specify a user_mods directory that was associated with the compset definition. This is the case when say a cam compset wants to have user_nl changes for another component, say the driver.

As part of this @billsacks implemented the following additional changes:

1. Modify user_mod_support so that:
- a. we reverse include_dirs as soon as it's created. 
- b. we change update_user_nl_file to append changes rather than prepend them

2. We change the application of SourceMods in user_mod_support so that existing files are overwritten, rather than keeping the existing file in place.

In addition, new changes have been put in to support aquaplanet SOM in the data ocean model. Additional changes for this capability will be brought into future PRs - particularly the ability to also have aquaplanet DOM.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: None

Code review: sacks, jedwards


scripts/lib/CIME/case.py
```diff
@@ -951,15 +953,26 @@ def create_caseroot(self, clone=False):
             self._create_caseroot_sourcemods()
         self._create_caseroot_tools()
 
+        # Apply user_mods if part of compset
+        if self._user_mods is not None:
+            self._user_mods = self.get_resolved_value(self._user_mods)
```
billsacks: Is there a reason that get_resolved_value is done here, at the last minute, rather than being done when this element is first retrieved from the xml file? It seems like the latter would be less error-prone as the code evolves. But I don't know what the standard convention is in this respect.

rljacob:
How does this change either a create_newcase line or a config_compset.xml definition?

billsacks:
Okay, I just pushed changes that fix the main problems I identified. I haven't addressed this comment:

> Is there a reason that get_resolved_value is done here, at the last minute, rather than being done when this element is first retrieved from the xml file? It seems like the latter would be less error-prone as the code evolves. But I don't know what the standard convention is in this respect.

I have changed where this is applied, but haven't made things any better in this respect.

Note that I have done only limited testing on my most recent commit. Probably scripts_regression_tests should be rerun.

mvertens:
The create_newcase line is not changed. The config_compset.xml definition
could look like the following:

  <compset>
    <alias>EC6AQUAP</alias>
    <lname>2000_CAM60%AQUAP_SLND_SICE_DOCN%SAQUAP_SROF_SGLC_SWAV</lname>

<user_mods>$SRCROOT/components/cam/cime_config/usermods_dirs/aquap_som</user_mods>
  </compset>

the nice thing about this is that the directory aquap_som has changes to
the orbital parameters that are needed for aqua-planet and this is the most
robust way to do this.

mvertens:
@sacks - this is a good idea, thanks.

@jedwards - could you please implement since I will be on travel this week?

rljacob:
Does this address isue #842 ?

mvertens:
Yes! Thanks for catching this. After a lot of discussion this week, we
decided that our new approach was a lot more robust than what was proposed
in #842.

billsacks:
I reworked @jedwards4b 's last commit to put the user_mod information more up-front, which I feel is important.

So now the first line of stdout and README.case look like this:

```
Compset longname is 2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV with user_mods directory /Users/sacks/temporary/user_mods_compset
```

I now feel this is ready to go, but scripts_regression_tests should probably be rerun on the latest version.

jedwards4b:
Reran scripts_regression_tests on cheyenne - all good except T_TestRunRestart which is in issue #1368

Issue #1141, opened 2017-02-15
Build pFUnit on some other machines


billsacks:
We should build pFUnit on some other machines so we can run unit tests there:

- cheyenne
- hobart

Instructions for building pFUnit are:

1. Download pFUnit from
   http://sourceforge.net/projects/pfunit/files/latest/download

2. Set the PFUNIT environment variable. For example::

     ```
     export PFUNIT=/glade/p/cesmdata/cseg/tools/pFUnit/pFUnit3.2.8_cheyenne_Intel17.0.1_MPI_openMP
     ```

3. Build pFUnit:

```
     mkdir build
     cd build
     cmake -DMPI=YES -DOPENMP=YES ..
     make -j 4
```

4. Run pFUnit's own unit tests:

```
     make tests
```

   Note: this didn't work on yellowstone. Instead, I had to do it by hand:

```
     DAV_CORES=4 execca
     mpirun.lsf tests/tests.x
```

5. Install pFUnit on your system:

```
     make install INSTALL_DIR=$PFUNIT
```

billsacks:
There are already some pFUnit builds on hobart, in Sean Santos's home directory. I could initially test with these, but eventually want to rebuild this using our standard build arguments (as given above), put in some more central location.

billsacks:
On cheyenne, I built pFUnit3.2.8 with:

```
module load intel/17.0.1
module load cmake

export PFUNIT=/glade/p/cesmdata/cseg/tools/pFUnit/pFUnit3.2.8_cheyenne_Intel17.0.1_MPI_openMP
mkdir build
cd build
cmake -DMPI=YES -DOPENMP=YES -DCMAKE_INSTALL_PREFIX=$PFUNIT ..
make -j 4
make install
```

rljacob:
FYI.  There is some work going on in CMDV to containerize pFUnit.

billsacks:
@rljacob - that sounds great, thanks for the heads-up. For now we at least need this on cheyenne (the yellowstone replacement), but that could be useful in the future.

billsacks:
I'm running into trouble on cheyenne. The fundamental problem is that (1) On the login nodes, I can build the unit tests, but cannot run them, but (2) On the compute nodes, I cannot build the unit tests. The unit test workflow requires building and running in the same place. I'm working with CISL to try to resolve this.

billsacks:
I got the pFUnit build and self-tests working on cheyenne with this procedure:

```
[...]

PR #1381, opened 2017-04-19
Get Fortran unit tests working on cheyenne

billsacks:
Changes needed to get the cime Fortran unit tests building and running
on cheyenne.

Some of these changes just involved setting the appropriate variables
for cheyenne. However, I needed to make more extensive changes in order
to support setting certain environment variabes only if we're doing unit
testing.

Note that the cheyenne unit tests can be built and run from either the
login nodes or the share queue - e.g.

```
qsub -I -l select=1:ncpus=1:mpiprocs=1 -l walltime=06:00:00 -q share -A P93300601
```

But running on the share queue causes some of the system tests run with
scripts_regression_tests to fail - so for now running on the login nodes
seems like the best bet.

Getting this working was painful - specifically due to complications of
how we can (and can't) build and run mpi programs on cheyenne. I'm going
to take a stab at allowing non-mpi builds and runs of the unit tests in
a separate PR to make porting the unit tests easier. But I still think
it's worth bringing this PR to master first.

Test suite: scripts_regression_tests on both yellowstone and cheyenne
(latter run from the login node). One test failed on cheyenne: `FAIL:
test_run_restart (__main__.T_TestRunRestart)`. It looks like this was
fixed recently in cime, so I assume this failure was expected from the
version on which this branch is based.
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes: Partially addresses #1141

User interface changes?: none

Code review:

### DIFF
--- a/scripts/tests/scripts_regression_tests.py
+++ b/scripts/tests/scripts_regression_tests.py
@@ -178,13 +178,19 @@ class N_TestUnitTest(unittest.TestCase):
         cls._testroot = os.path.join(TEST_ROOT, 'TestUnitTests')
         cls._testdirs = []
 
+    def _has_unit_test_support(self):
+        default_compiler = MACHINE.get_default_compiler()
+        compiler = Compilers(MACHINE, compiler=default_compiler)
+        pfunit_path = compiler.get_optional_compiler_node("PFUNIT_PATH")
+        if pfunit_path is None:
+            return False
+        else:
+            return True
+
     def test_a_unit_test(self):
         cls = self.__class__
-        machine           = MACHINE.get_machine_name()
-        compiler          = MACHINE.get_default_compiler()
-        if (machine != "yellowstone" or compiler != "intel"):
-            #TODO: get rid of this restriction
-            self.skipTest("Skipping TestUnitTest - only supported on yellowstone with intel")
+        if not self._has_unit_test_support():
+            self.skipTest("Skipping TestUnitTest - PFUNIT_PATH not found for the default compiler on this machine")
         test_dir = os.path.join(cls._testroot,"unit_tester_test")
         cls._testdirs.append(test_dir)
         os.makedirs(test_dir)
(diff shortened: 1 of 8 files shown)


scripts/tests/scripts_regression_tests.py
```diff
@@ -178,13 +178,19 @@ def setUpClass(cls):
         cls._testroot = os.path.join(TEST_ROOT, 'TestUnitTests')
         cls._testdirs = []
 
+    def _has_unit_test_support(self):
```
billsacks: Note that scripts_regression_tests will now run the Fortran unit tests on any machine for which PFUNIT_PATH is defined for the default compiler

src/externals/CMake/pFUnit_utils.cmake
```diff
@@ -215,7 +215,8 @@ function(create_pFUnit_test test_name executable_name pf_file_list fortran_sourc
   endif()
 
   # Prefix command with an mpirun command
-  set (MY_COMMAND ${PFUNIT_MPIRUN} ${MY_COMMAND})
+  separate_arguments(PFUNIT_MPIRUN_LIST UNIX_COMMAND ${PFUNIT_MPIRUN})
```
billsacks: This seems to be needed when PFUNIT_MPIRUN is made up of multiple, space-delimited strings - i.e., an mpirun command and some arguments.

tools/unit_testing/run_tests.py
```diff
@@ -277,8 +278,20 @@ def _main():
...
+    machspecific.load_env(compiler, debug, mpilib)
+    os.environ["OS"] = os_
+    os.environ["COMPILER"] = compiler
+    os.environ["DEBUG"] = stringify_bool(debug)
+    os.environ["MPILIB"] = mpilib
+    if use_openmp:
+        os.environ["compile_threaded"] = "true"
+    os.environ["CC"] = find_executable("mpicc")
+    os.environ["FC"] = find_executable("mpif90")
+    os.environ["UNIT_TEST_HOST"] = socket.gethostname()
```
billsacks: This environment stuff is mostly just moved up from below. But note that I also now set UNIT_TEST_HOST.

billsacks:
I'm done making line comments; this is ready for review.

PR #1385, opened 2017-04-19
return only aprun args and use executable from xml

### LINKED ISSUE #1384: Unable to use DDT with aprun on Titan
We determined the cause of the problem of trying to use DDT on Titan. For DDT to work, we need to call "ddt --connect aprun" instead of "aprun". I tried to do this through env_mach_specific.xml, changing the following:

122     <executable args="default">aprun</executable>
122     <executable args="default">ddt --connect aprun</executable>

But when I do this, the flags (e.g., -S8 -N16 -n16 -d1) all get dropped. The output from the standard "aprun" is:

run command is aprun -S 8 -n 16 -N 16 -d 1 /lustre/atlas1/stf006/scratch/imn/enable_debugger17.3/bld/acme.exe    >> acme.log.$LID 2>&1  

But the output from "ddt --connect aprun" is:

run command is ddt --connect aprun   /lustre/atlas1/stf006/scratch/imn/enable_debugger17.3/bld/acme.exe  >> acme.log.$LID 2>&1  

The DDT job successfully connects to my debugger but, of course, fails because it's only running one MPI task. I need some way of specifying "ddt --connect aprun" instead of "aprun" that keeps those flags in play.

I should also mention that this is pretty critical because PGI said they'd make much better progress if they could use a debugger on Titan to diagnose what's going on with their compiler bugs.

jedwards4b:
Return aprun args but use executable string from xml

Test suite: Someone with a titan account should test this
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1384 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/aprun.py
+++ b/scripts/lib/CIME/aprun.py
@@ -30,10 +30,10 @@ def _get_aprun_cmd_for_case_impl(ntasks, nthreads, rootpes, pstrids,
     >>> machine = "titan"
     >>> run_exe = "acme.exe"
     >>> _get_aprun_cmd_for_case_impl(ntasks, nthreads, rootpes, pstrids, max_tasks_per_node, pes_per_node, pio_numtasks, pio_async_interface, compiler, machine, run_exe)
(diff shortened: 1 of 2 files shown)


jgfouca:
Thanks.

PR #1388, opened 2017-04-20
Renamed manage_case to query_config

### LINKED ISSUE #1155: Rename manage_case to query_config
"manage" is for scripts that can edit xml files.  "query" just reads them.

"manage_case" in its current version only reads xml files so we should change its name to "query_case".

### LINKED ISSUE #1190: Change arguments and behavior of "manage_case"
I'm trying to write some documentation on using "manage_case" but don't really want to in its current form.  There is already issue #1155 for changing its name.  This issue is for other behavior.

1.  ./manage_case with no arguments quietly returns nothing.  It should print the help string instead.
2.  --query-compsets-setby  should be changed to just --compset.  This assumes the name is changed to "query_case". or "query_config"  The "setby" can be explained in the help.
3. --query-compsets-setby with no arguments gives an unhelpful error message that it "expects one argument".  If you supply it with on argument, but the wrong one, you get a more helpful error message listing all the valid options.  No arguments should provide the same error as invalid arguments.
4.  --query-compsets-setby should allow an "all" option that prints all the compsets it can find.
5. --query-compset-setby should not print the compset naming convention with every invocation.
6.  --query-component-name, similarly should just be --component
7.  --query-component-name has the same error message behavior as --query-compsets-setby
8.  --query-grids,  should just be --grids
9.  --grids-shortname and --grids-alias are "not implemented yet".  What are these supposed to do?
10.  --query-machines should just be --machines
11.   --query-machines should include in its output if cime knows the current platform.  Optional argument "current" would only print info about the current machine.
12. The --long additional argument only seems to do anything with --query-grids

erichlf:
Renamed manage_case to query_config. Additionally, changed all
references to manage_case to query_config. Added ability to query current machine and a single machine. Added 'all' option to --compsets. Made the help message much more useful.

Test suite: `scripts_regression_tests.py`
and
```
./query_config
./query_config --grids
./query_config --machines
./query_config --machines current
./query_config --machines redsky
./query_config --machines edison
./query_config --machines skybridge
./query_config --compsets
./query_config --compsets all
./query_config --compsets drv
./query_config --component
```
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #1155
Fixes #1190

User interface changes?:
manage_case -> query_config
--compset -> --compsets
--machine -> --machines
added 'all' option for --compsets
added 'current' and <machine> to --machines

Code review: @billsacks, @jgfouca


rljacob:
As @jedwards4b pointed out, later in #1155 we decided query_config is the better name.  I'll change the issue title and you should change this PR.

erichlf:
I will fix it.

erichlf:
Okay, should be fixed now.

erichlf:
How about we pause this PR until I fix #1190?

rljacob:
Sure.

For backward compatability, maybe we should include a link called "manage_case" that points to the new name.

mvertens:
At this point, I would vote for just calling it query_config. I'm not sure
a lot of people use manage_case yet - and its before the release - so I'm
fine without a link.

erichlf:
I believe I have now addressed all issues in #1190 and so this should be ready for review and merge.

rljacob:
This looks great!  I already know more about ACME by using this.  But there are a few issues.

The header in the output for --compset should say "Alias" instead of "Short Name".
```
       --------------------------------------
       Compset Short Name: Compset Long Name 
       --------------------------------------
```

The list of possible arguments to compset is the union of all the components in ACME and CESM.  Could it be just one model's worth?
```
./query_config  --compset blah
ERROR: Invalid input argument blah, valid input arguments are ['allactive', 'drv', 'cam', 'cism', 'clm', 'cice', 'pop', 'mpaso', 'mpasli', 'mpascice']
```

If I forget to give an argument to --component, the help stream gives the wrong list of input arguments:
```
CSI0350919:scripts jacob$ ./query_config  --component 
usage: query_config [-h] [-d] [-v] [-s] [--compset COMPSET]
                    [--component COMPONENT] [--grids]
                    [--machines [{current,empty}]] [--long]
query_config: error: argument --component: expected one argument
Valid input arguments are ['CPL', 'ATM', 'LND', 'ICE', 'OCN', 'ROF', 'GLC', 'WAV', 'ESP']
```
The help string for a bad argument is correct:
```
CSI0350919:scripts jacob$ ./query_config  --component blah
ERROR: Invalid input argument blah, valid input arguments are ['cam', 'datm', 'satm', 'xatm', 'vic', 'clm', 'dlnd', 'slnd', 'xlnd', 'mpascice', 'cice', 'dice', 'sice', 'xice', 'mpaso', 'pop', 'aquap', 'docn', 'socn', 'xocn', 'rtm', 'mosart', 'drof', 'srof', 'xrof', 'mpasli', 'mpaslisia', 'mpaslialb', 'cism', 'dglc', 'sglc', 'xglc', 'ww', 'dwav', 'swav', 'xwav', 'sesp', 'desp']
```

I was running this on my laptop which is not a machine CIME recognizes.  Everything worked except 
`./query-config --machines`
WIth no argument, it correctly listed all the ACME machines but also had this line between each entry:
`Could not find machine match for 'CSI0350919' or 'CSI0350919'`
CSI0350919 is the hostname of my laptop.

`./query-config --machines current `on my unsupported laptop gave the same string as above and a stack trace.

rljacob:
Regarding  separate --compset and --component arguments, this parallels the separation of these files within CIME.   I also like current output of just --compset.  If you want both, add a new argument for the combination of those 2 (but make a new issue or we'll never resolve this.)

billsacks:
> Regarding separate --compset and --component arguments, this parallels the separation of these files within CIME. I also like current output of just --compset. If you want both, add a new argument for the combination of those 2 (but make a new issue or we'll never resolve this.)

Okay, I'm okay with this if others like the separation, but I would request that `--component` be renamed, because it's very unclear what you can expect to get from an argument named `--component`.

rljacob:
How would you describe the information in config_compoents.xml?

billsacks:
> How would you describe the information in config_components.xml?

settings?

erichlf:
@rljacob As to your model aware components question-- I actually asked about this but I was told it was my machine and things should just work (unless I misunderstood). Any idea how to address this issue?
https://github.com/ESMCI/cime/issues/1190#issuecomment-296810299

@billsacks 
5. Where? I don't believe I have touched any such documentation, but I would be happy to fix it.
6. That text corresponds to the variable name. I don't have control over that without overloading argparse even more than I already have.  
7. I am unable to recreate your issue. I get the same component list either way.
The final comment seemed like a discussion point that I wasn't sure how to even address, and might be better dealt with in a separate issue.

billsacks:
> 5. Where? I don't believe I have touched any such documentation, but I would be happy to fix it.

Looks like it comes from lines like this (repeated in a few places):

```python
                        help="Query compsets that are set by the target component for {} model".format(cime_model))
```

> 7. I am unable to recreate your issue. I get the same component list either way.

I think I mis-stated this one. It seems the real issue is with the `--component` argument: without anything, I get:

```
[roo2:~/cime/scripts]$ ./query_config --component
usage: query_config [-h] [-d] [-v] [-s] [--compset COMPSET]
                    [--component COMPONENT] [--grids]
                    [--machines [{current,empty}]] [--long]
query_config: error: argument --component: expected one argument
Valid input arguments are ['CPL', 'ATM', 'LND', 'ICE', 'OCN', 'ROF', 'GLC', 'WAV', 'ESP']
```

but then I see:

```
[roo2:~/cime/scripts]$ ./query_config --component CPL
ERROR: Invalid input argument CPL, valid input arguments are ['cam', 'datm', 'satm', 'xatm', 'vic', 'clm', 'dlnd', 'slnd', 'xlnd', 'cice', 'dice', 'sice', 'xice', 'pop', 'mpas-o', 'aquap', 'docn', 'socn', 'xocn', 'rtm', 'mosart', 'drof', 'srof', 'xrof', 'cism', 'dglc', 'sglc', 'xglc', 'ww3', 'dwav', 'swav', 'xwav', 'desp', 'sesp']
```

And why would the allowed values for COMPONENT differ for `--compset` than for `--component`?

erichlf:
Maybe things are now up to par? I hope the current state is satisfactory to everyone.

billsacks:
This is getting closer, but I still have some outstanding requests, and some new ones based on this latest version:

1. I'd still like the `--component` argument to be renamed to `--settings`; does anyone object to this?

1. `--compsets` works without specifying any argument, but `--component` does not. I'm thinking `--component` (hopefully renamed to `--settings`) should also support the 'all' argument, which, like `--compsets`, should be the default.

1. The documentation for `--compsets` should add something like, "If no component is given, list compsets defined by all components."

1. I'd still like some overall documentation that appears when you run with `-h`. Something like, `Displays information about available compsets, component settings, grids and/or machines. Typically run with one of the arguments --compsets, --settings, --grids or --machines; if you specify more than one of these arguments, information will be listed for each.`

1. There is a typo in the help for `--machines` ('giving' instead of 'given', and some generally hard-to-parse text). I would reword this to, `Query supported machines for cesm model. If option is left empty then all machines are listed; if 'current' is given, machine details are listed for the current machine.`

rljacob:
How about adding "--settings" as an equivalent to "--components".   As long as we have config_components.xml, we need "--components".

billsacks:
@rljacob - I'm fine with these two being equivalent, though would argue that the one listed with `-h` should be `--settings` or something like that. Having it named `--component` made it hard for me to understand the usage of this tool initially: my initial intuition was that that was a subsetting operation that applied to all other arguments (e.g., only asking for the grids for the given component). Naming it something like 'settings' is more parallel with the other arguments, and thus IMHO makes it easier to understand the usage:

Question: "What are you querying?"

Answer: "I'm querying grids": `--grids`

Answer: "I'm querying machines": `--machines`

Answer: "I'm querying compsets": `--compsets`

What's the parallel construct for component settings? I'm not totally sure, but having it named `--component` implies:

Answer: "I'm querying components": `--component`

which I know isn't right.

"I'm querying settings" is a little better, though a bit vague....

rljacob:
Why are the many files called "config_components.xml" called that?

jedwards4b:
They define the xml variables (or settings) used by a given component.

rljacob:
I think "--settings" is unintuitive.  If I'm looking at the output for --compsets and I want to change it, I go to config_compsets.xml.  If I don't like what --grids is telling me, I edit config_grids.xml.   But if I want to change something from --settings, don't look for config_settings.xml, look for config_components.xml.

billsacks:
@rljacob I think we have to agree to disagree here. I'll defer to you on this unless others have opinions.

billsacks:
I'm happy with this now. Thanks for all your work, @erichlf !

PR #1392, opened 2017-04-20
Removed Deprecated acme/allactive/config_pes.xml

erichlf:
config/acme/allactive/config_pes.xml and
config/acme/allactive/config_pes.xml.cime2 are deprecated in favor of
config/acme/allactive/config_pesall.xml so they have been removed.

Fixes https://github.com/ACME-Climate/ACME/issues/1434


erichlf:
This should be done in ACME repo directly.

PR #1393, opened 2017-04-20
Conforming command line

### LINKED ISSUE #428: Using -testid yields misleading error message
The command:
./create_test --xml-machine yellowstone --xml-compiler intel --xml-category aux_cam -testid aux_cam_try8

gives the error message:
ERROR: Could not fill-out test name, partial string 'aux_cam_try8' had no grid information and you did not provide any

jedwards4b:
All command line arguments should conform.   I have allowed for one exception, the -value option to xmlquery.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #428 

User interface changes?:  YES, now accepting only conforming command line arguments 

Code review:

### DIFF
--- a/config/acme/machines/template.case.run
+++ b/config/acme/machines/template.case.run
@@ -55,9 +55,7 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
     parser.add_argument("--caseroot",
                         help="Case directory to build")
 
-    args = parser.parse_args()
-
-    CIME.utils.handle_standard_logging_options(args)
+    CIME.utils.parse_args_and.parse_args_and_handle_standard_logging_options(args, parser)
 
     if args.caseroot is not None:
         os.chdir(args.caseroot)
--- a/scripts/lib/CIME/utils.py
+++ b/scripts/lib/CIME/utils.py
@@ -637,7 +637,7 @@ class _LessThanFilter(logging.Filter):
         #non-zero return means we log this message
         return 1 if record.levelno < self.max_level else 0
 
-def handle_standard_logging_options(args):
+def handle_standard_logging_options(args, parser):
     """
     Guide to logging in CIME.
 
@@ -659,6 +659,11 @@ def handle_standard_logging_options(args):
     # Change warnings and above to go to stderr
     stderr_stream_handler = logging.StreamHandler(stream=sys.stderr)
     stderr_stream_handler.setLevel(logging.WARNING)
+
+    # scripts_regression_tests is the only thing that should pass a None argument in parser
+    if parser is not None:
+        _check_for_invalid_args(args[1:])
+        args = parser.parse_args(args[1:])
 
     # --verbose adds to the message format but does not impact the log level
     if args.verbose:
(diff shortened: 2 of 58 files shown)


scripts/lib/CIME/utils.py
```diff
@@ -1229,6 +1236,19 @@ def run_and_log_case_status(func, phase, caseroot='.'):
...
+def _check_for_invalid_args(args):
+    for arg in args:
+        if arg.startswith("--"):
+            continue
+        if arg.startswith("-") and len(arg) > 2:
+            if arg == "-value":
+                logger.warn("This argument is deprecated, please use -%s"%arg)
+            else:
+                expect(False, "Invalid argument %s\n  Multi-character arguments should begin with \"--\" and single character with \"-\"\n  Use --help for a complete list of available options"%arg)
+
```
jgfouca: Lots of unnecessary blank lines.

jedwards4b:
@jgfouca do you have a suggestion for how to handle -j8?   Note that -j 8 is okay.

jgfouca:
@jedwards4b , I don't have a suggestion unfortunately. It's impossible to tell the difference between a short option with no space vs. someone trying to use a non-complain long option. Maybe it's not that important to support short option + no space.

scripts/lib/CIME/utils.py
```diff
@@ -660,6 +660,11 @@ def handle_standard_logging_options(args):
     stderr_stream_handler = logging.StreamHandler(stream=sys.stderr)
     stderr_stream_handler.setLevel(logging.WARNING)
 
+    # scripts_regression_tests is the only thing that should pass a None argument in parser
+    if parser is not None:
+        _check_for_invalid_args(args[1:])
+        args = parser.parse_args(args[1:])
```
billsacks: If we're going to rely on this function to do the actual parsing of arguments, then it should be renamed. Currently, as I'm reading through the code, I am left wondering where the actual argument parsing happens: it is not intuitive that the argument parsing would happen inside a function named 'handle_standard_logging_options'. Maybe even rename this function to 'parse_args'?

jedwards4b:
We are now ready to merge this. @billsacks I have renamed as suggested.

jgfouca:
OK, I'm fine with this for now. Note that we are officially non-compliant with POSIX argument handing because we don't support short-option-without-space options (like -j8) like I mentioned above. That said, I'm OK with this as a temporary measure as we transition.

config/acme/machines/template.case.run
```diff
@@ -55,9 +55,7 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
     parser.add_argument("--caseroot",
                         help="Case directory to build")
 
-    args = parser.parse_args()
-
-    CIME.utils.handle_standard_logging_options(args)
+    CIME.utils.parse_args_and.parse_args_and_handle_standard_logging_options(args, parser)
```
billsacks: @jedwards4b - I think this line has a bug: see the extra `parse_args_and.`

PR #1395, opened 2017-04-20
Add Support of true and false in Namelist

### LINKED ISSUE #1283: perl namelist should allow both "false" and ".false."
The MPAS-CICE buildnml is complaining about a logical a user is trying to set using user_nl_mpascice.

`config_test_case_diag = false`

'ERROR(Build::Namelist::_parse_next): expect a F90 constant for a namelist instead got:  false

If they set it instead to

`config_test_case_diag = .false.`

There there's no error.    But both "false" and ".false." should be allowed.

erichlf:
The Namelist.pm only allowed for .false. or .true. in namelist files. I have
fixed the regex to allow for dots or no dots.

Fixes #1283


erichlf:
@jedwards4b Should be fixed now.

Issue #869, opened 2016-11-30
Speed up unit test build & run


billsacks:
_From @billsacks on December 21, 2015 23:21_

The unit test build and run can take longer than I'd like. This is particularly true for CLM, which now has 20 unit test executables; each of these is linked against the CLM library, so whenever you change any CLM code, all 20 need to relink. So the build takes some time due to that. In addition, particularly on yellowstone, the use of mpi-based tests means that each test executable takes more than 1 sec to run, leading to a total test time of about 30 sec (these tests should run in milliseconds).

A few ideas about how to speed things up:

(1) Create a single test executable for all of CLM (or perhaps one executable for each CLM subdirectory - e.g., a biogeophys test executable, etc.). However, by itself, this is considered unacceptable, because CTest doesn't give granular enough reporting. So, would it be possible to have a single executable, with CTest calling it multiple times in different ways (as in http://baoilleach.blogspot.com/2013/06/using-ctest-with-multiple-tests-in.html)? Having a single executable would (a) speed up the build, (b) speed up the run, (c) make it easier to move around code and its associated tests.

(2) Revisit issue https://github.com/CESM-Development/cime/issues/26 : allow tests to use a serial build of pFUnit when possible. This would allow you to run CLM unit tests from yellowstone (rather than needing to do so on caldera), and would speed up the run time if (1) isn't feasible.

_Copied from original issue: CESM-Development/cime#341_

billsacks:
cc @quantheory @bandre-ucar

rljacob:
@billsacks still relevant?

billsacks:
Still possibly relevant but never gonna happen.

billsacks:
(In practice, when I'm doing test-driven-development, after I do an initial build and run of the unit tests on a branch, I cd into the unit test build subdirectory for just the tests I'm working on and run something like 'make && make test' from there, so that I only end up rerunning the tests I'm currently working on. This works well for me, though is currently undocumented.)

PR #1396, opened 2017-04-20
Get unit test build and run working with serial or parallel pFUnit

### LINKED ISSUE #1141: Build pFUnit on some other machines

billsacks:
The primary aim of this PR is to enable serial unit test builds as well
as MPI-based builds. A given machine/compiler can now have multiple
pFUnit installations; the correct installation is chosen based on the
options given to run_tests.py (--use-mpi and --use-openmp).

The default is now to use the serial build and run (i.e., without
mpi). This is a change from the earlier behavior, where the default was
to use mpi. I made this change because I was finding it challenging to
get the mpi-based unit tests working on new supercomputers (due to
restrictions on where you can build and where you can launch mpi jobs -
often those two locations are different; there were also challenges with
setting up the mpirun command for unit testing). Currently we aren't
leveraging the parallel capabilities of pFUnit in any of our tests, so
it seemed reasonable to just use the serial version. Note that this
builds mpi-serial as part of the unit test build: this is needed by
tests that use MCT.

In addition, this PR adds unit testing support on hobart-intel (NCAR/CGD
machine), so that scripts_regression_tests on that system will now run
the Fortran unit tests. I have removed references to PFUNIT_PATH on some
other systems where Sean Santos had installed pFUnit in his home directory.

This changes the unit testing user interface somewhat:

- config_compilers should now give MPILIB and compile_threaded
  attributes for PFUNIT

- The default mode of operation for run_tests.py is now serial - with
  neither mpi nor openMP. (This means that, on yellowstone, you no
  longer need to run the unit tests on caldera.)

- Other components that have unit tests (CLM and CAM) will need to
  rework their top-level CMakeLists.txt file slightly. I can help with
  this. @ekluzek and @cacraigucar - take note.

Test suite: scripts_regression_tests on yellowstone and cheyenne
   All pass except K_TestCimeCase, which was also failing on the version
   of master on which this is based (6dd85dd)

   Also ran scripts_regression_tests N_TestUnitTest on hobart
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes #1141 , partially addresses #869 

User interface changes?: YES - see list above (under "This changes the
unit testing user interface somewhat") 

Code review:


scripts/lib/CIME/BuildTools/cmakemacroswriter.py
```diff
@@ -117,3 +118,42 @@ def end_ifeq(self):
...
+        Returns transformed value
+        """
+
+        value_transformed = value
+        if self._element_needs_whitespace_removal(name):
+            value_transformed = value_transformed.strip()
+
+        return value_transformed
+
+    def _element_needs_whitespace_removal(self, name):
```
jedwards4b: Do we really need this list?  Wouldn't it be enough to just always remove trailing (and leading) white space?

billsacks: I wondered about that. I thought there might be some variables where there was deliberately some leading or trailing white space so that when they are concatenated together you get something like `-foo -bar` rather than `-foo-bar`. But I didn't look closely or experiment with this. If you think it's safe to always remove whitespace, then I can give it a shot on the four machines that now have unit testing support.

PR #1413, opened 2017-04-24
Pass annual-average fields to GLC

billsacks:
These changes are from Mariana Vertenstein (primarily) and myself

Until now, the CESM coupler has been sending CISM daily-average fields,
and then CISM has been averaging these fields to annual-averages in
order to do its mass balance time step using annual-average SMB (and
temperature). This has a number of problems:

(1) We could not restart mid-year, because CISM doesn't write the
    necessary partial accumulation information to its restart files.

(2) With Bill Lipscomb's new remapping (bilinear + conservation
    correction; coming soon), there were too-large corrections needed
    when doing the remapping from CLM to CISM on daily fields.

(3) This led to a discrepancy between TG and BG/IG compsets: In BG/IG,
    CISM received the annual average of remapped daily fields; in TG,
    CISM received the remapped annual average fields. The remapping is
    not a linear operation, so there would typically be differences in
    these two runs.

This tag, together with cism2_1_30, allows the coupler to make annual
averages before remapping fields to CISM. This was a bit tricky: CISM
still needs to be called once per day in order to support mid-year
restarts (CESM can only restart at a frequency when all components are
called). So now the GLC calling frequency has been separated from the
GLC averaging frequency. Then GLC/CISM is passed a flag saying whether
the inputs are currently valid; CISM only accumulates averages of the
inputs when they are valid. (In the standard case, this means that CISM
accumulates a single value each year, leaving all of the averaging up to
the coupler. But this supports the ability to still do the averaging in
CISM if this is desired - however, note that mid-year restarts won't
work in that case [CISM now aborts if you try to do a mid-year restart
in that configuration].)

This is implemented in such a way that ACME runs with MPAS-LI should be
unaffected: the default GLC_AVG_PERIOD for ACME is
'glc_coupling_period', which makes things collapse to the old behavior.

There are some mostly-unrelated changes in
src/drivers/mct/cime_config/buildnml and
src/drivers/mct/cime_config/namelist_definition_drv.xml. These use
add_default rather than set_value for the various *_cpl_dt entries,
because in general we should not be calling set_value directly. And,
since we're using add_default, we needed to add
skip_default_entry="true" to the relevant xml variables.

In CESM, this change must be coordinated with cism2_1_30 and
https://github.com/CESM-Development/cime_config/pull/6

   Also, aux_glc test suite with cism2_1_30:
   - out-of-the-box: tests pass, and bit-for-bit except for IG tests
     without test_coupling
   - with GLC_AVG_PERIOD = 'glc_coupling_period': tests pass and
     bit-for-bit, except for expected failures in tests that restart
     mid-year
Test baseline: For aux_glc tests, baseline was cism2_1_29
Test namelist changes: YES - adds new glc_avg_period
Test status: greater than roundoff-level changes for multi-year runs
   with CISM, other than TG

Fixes none

User interface changes?: new xml variable, GLC_AVG_PERIOD

Code review: 

cc @whlipscomb @JeremyFyke @stephenprice @matthewhoffman


src/drivers/mct/main/cesm_comp_mod.F90
```diff
@@ -2171,6 +2190,8 @@ subroutine cesm_run()
       if (tod            == 0) t24hr_alarm = .true.
       if (month==1 .and. day==1 .and. tod==0) t1yr_alarm = .true.
 
+      ! TODO(wjs, 2017-04-05) I think glcrun_alarm can be removed from infodata: It used
```
jedwards4b: Can we go ahead and do this as part of the PR?

billsacks: I have made this change, and am rerunning testing

apcraig:
I'm unclear about this statement "Then GLC/CISM is passed a flag saying whether the inputs are currently valid; CISM only accumulates averages of the inputs when they are valid.".  I thought the change was that the coupler was accumulating the input.  Should this say, "CISM only runs when the inputs are valid"?

billsacks:
@apcraig - thanks for your review. Sorry, I realized the wording was confusing there because Ieft out a subtlety. I have reworded that paragraph to the following; I hope this makes things more clear:

This tag, together with cism2_1_30, allows the coupler to make annual
averages before remapping fields to CISM. This was a bit tricky: CISM
still needs to be called once per day in order to support mid-year
restarts (CESM can only restart at a frequency when all components are
called). So now the GLC calling frequency has been separated from the
GLC averaging frequency. Then GLC/CISM is passed a flag saying whether
the inputs are currently valid; CISM only accumulates averages of the
inputs when they are valid. (In the standard case, this means that CISM
accumulates a single value each year, leaving all of the averaging up to
the coupler. But this supports the ability to still do the averaging in
CISM if this is desired - however, note that mid-year restarts won't
work in that case [CISM now aborts if you try to do a mid-year restart
in that configuration].)

billsacks:
I have rerun scripts_regression_tests on the latest commit, and everything passes except for the expected failures in K_TestCimeCase

jedwards4b:
There are no expected failures on master  - the issues you opened on that have been resolved.
Perhaps you need to rebase?

src/drivers/mct/cime_config/namelist_definition_drv.xml
```diff
@@ -1387,7 +1387,7 @@
   <!-- group seq_timemgr_inparm    -->
   <!-- =========================== -->
 
-  <entry id="atm_cpl_dt" modify_via_xml="ATM_NCPL">
+  <entry id="atm_cpl_dt" modify_via_xml="ATM_NCPL" skip_default_entry="true">
```
rljacob: what are these changes for?

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -1,8 +1,4 @@
 !===============================================================================
```
rljacob: Looks like this file had a lot of clean-up in addition to the glc-required changes?

src/drivers/mct/cime_config/namelist_definition_drv.xml
billsacks: @mvertens - can you address this? I don't remember, if I ever knew at all.

src/drivers/mct/shr/seq_timemgr_mod.F90
billsacks: Yes, by @mvertens

mvertens:
Its been so long... Here is my current rationale. If skip_default_entry is
not set to "true" then the driver sets all the default values explicitly
from the xml file. However, if you have code in the buildnml that
explicitly sets these values than you want to ensure that this code is
always used - and so you want to have skip_default_entry="true" for those
variables.

mvertens:
Sorry - hit send too soon. The driver buildnml sets the NCPL values in the
following:

    for comp in comps:
        ncpl = case.get_value(comp.upper() + '_NCPL')
        if ncpl is not None:
            cpl_dt = basedt / int(ncpl)
            totaldt = cpl_dt * int(ncpl)
            if totaldt != basedt:
                expect(False, " %s ncpl doesn't divide base dt evenly"
%comp)
            nmlgen.set_value(comp.lower() + '_cpl_dt', value=cpl_dt)
            mindt = min(mindt, cpl_dt)

rljacob:
Why are the skip_default_entry="true" additions needed for the glc feature?

mvertens:
I don't recall. However, its correct to have them there to begin with -
since the drv buildnml is setting these values explicitly.

billsacks:
@mvertens - Sorry it took me so long to get this PR submitted from the time you handed me the code.

I have a feeling that the skip_default_entry="true" changes were needed in conjunction with this change:

```diff
--- a/src/drivers/mct/cime_config/buildnml
+++ b/src/drivers/mct/cime_config/buildnml
@@ -101,7 +101,7 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
             totaldt = cpl_dt * int(ncpl)
             if totaldt != basedt:
                 expect(False, " %s ncpl doesn't divide base dt evenly" %comp)
-            nmlgen.set_value(comp.lower() + '_cpl_dt', value=cpl_dt)
+            nmlgen.add_default(comp.lower() + '_cpl_dt', value=cpl_dt)
             mindt = min(mindt, cpl_dt)
 #        elif comp.lower() is not 'cpl':
```

although I'm not sure that brings us any closer to remembering why these changes were put in on this branch.

mvertens:
@sacks - thanks for pointing out this diff. The key issue is that if you
have add_default - then you MUST have the skip_default_entry="true" - since
a call to add_default does not change an already set default value. This is
a subtle point that has tripped me up before. And we want to add
add_default rather than set_value - so that if needed this can be changed
by changing the xml file. Does that make sense?

billsacks:
@mvertens - I just tried changing the ATM_NCPL xml variable and the atm_cpl_dt namelist item in user_nl_cpl, and got the same behavior in master and on this branch: Both allowed me to change ATM_NCPL, and neither allowed me to change atm_cpl_dt in user_nl_cpl.

I know you had a reason for this change: I remember your explaining it to me back in January. I just forget what it was....

src/drivers/mct/main/prep_glc_mod.F90
```diff
@@ -444,6 +446,31 @@ end subroutine prep_glc_map_one_field_lnd2glc
 
   !================================================================================================
 
+  subroutine prep_glc_zero_fields()
+
+    !---------------------------------------------------------------
+    ! Description
+    ! Set glc inputs to zero
```
matthewhoffman: I realize this is just to generate consistent 'garbage' for times when the field is not valid.  One thought would be to fill it with a very obvious garbage value to help protect against the possibility that GLC accidentally tries to actually use it.  But I realize just filling with zeroes might be more aesthetically pleasing and such an approach should not be necessary, so feel free to ignore this comment.

matthewhoffman:
@billsacks , a cursory review does not reveal anything that concerns me.  Thank you for implementing these changes with ACME considerations in mind.  I think this capability will be useful for us as well once we start working on refining our coupling interval.

rljacob:
I also didn't see any showstoppers.

billsacks:
@rljacob and @matthewhoffman - thanks for your reviews of this. 

@rljacob - I have added some summarizing information in the top comment, summarizing the changes @mvertens made to the driver namelist generation:

There are some mostly-unrelated changes in
src/drivers/mct/cime_config/buildnml and
src/drivers/mct/cime_config/namelist_definition_drv.xml. These use
add_default rather than set_value for the various *_cpl_dt entries,
because in general we should not be calling set_value directly. And,
since we're using add_default, we needed to add
skip_default_entry="true" to the relevant xml variables.

@matthewhoffman - I see your point about sending garbage, but given the tight timeline here and the fact that you didn't see that as critical, I'm going to leave that as is.

PR #1414, opened 2017-04-24
Update case members in case_setup

### LINKED ISSUE #646: Need better test case coverage in cime_developer
We are currently only testing NCK, ERI, SEQ, ERS, ERR, and SMS.

Untested: ERP, ERT, ICP, LII, NCR, NOC, PEA, PEM, PET, PFS, REP, SSP

It seems like we should be testing most if not all of these nightly.

jgfouca:
Also, ACME is making heavy use of PET, so it needs to be tested
in cime_developer.

Test suite: scripts_regression_tests Z_FullSystemTest
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #646 

User interface changes?: Adds PET test to cime_developer, PET behavior change

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/SystemTests/pet.py
+++ b/scripts/lib/CIME/SystemTests/pet.py
@@ -30,6 +30,12 @@ class PET(SystemTestsCompareTwo):
             if self._case.get_value("NTHRDS_%s"%comp) <= 1:
                 self._case.set_value("NTHRDS_%s"%comp, 2)
 
+                # Need to halve NTASKS since we have double the threads
+                ntasks = self._case.get_value("NTASKS_%s" % comp)
+                if ntasks > 1:
+                    ntasks /= 2
+                    self._case.set_value("NTASKS_%s" % comp, ntasks)
+
         # Need to redo case_setup because we may have changed the number of threads
         case_setup(self._case, reset=True)
 
--- a/scripts/lib/CIME/SystemTests/pet.py
+++ b/scripts/lib/CIME/SystemTests/pet.py
@@ -30,10 +30,6 @@ class PET(SystemTestsCompareTwo):
             if self._case.get_value("NTHRDS_%s"%comp) <= 1:
                 self._case.set_value("NTHRDS_%s"%comp, 2)
 
-        # Subtle: machines like titan will not let you overload the number of threads
-        # on a node.
-        self._case.num_nodes *= 2
-
         # Need to redo case_setup because we may have changed the number of threads
         case_setup(self._case, reset=True)
 
--- a/scripts/lib/CIME/SystemTests/pet.py
+++ b/scripts/lib/CIME/SystemTests/pet.py
@@ -30,11 +30,9 @@ class PET(SystemTestsCompareTwo):
             if self._case.get_value("NTHRDS_%s"%comp) <= 1:
                 self._case.set_value("NTHRDS_%s"%comp, 2)
 
-                # Need to halve NTASKS since we have double the threads
-                ntasks = self._case.get_value("NTASKS_%s" % comp)
-                if ntasks > 1:
-                    ntasks /= 2
-                    self._case.set_value("NTASKS_%s" % comp, ntasks)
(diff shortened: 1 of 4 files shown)


jgfouca:
@rljacob can you advise me here? I have ACME folks complaining that PET doesn't work on titan, but I understand @billsacks concerns too.

rljacob:
I think its true that changing both the threads and the tasks changes the character of PET so that its a different test.
We currently have PET (just change the threads), PEM ( just change the tasks) and ERP (halve both the threads and tasks within a restart run).   So we should add...a....PEMT test?  Do they have to be 3 letters?

A test that keeps the total core count the same while halving tasks and doubling threads makes sense even if Titan queue weirdness wasn't the motivation.

billsacks:
PEMT seems good to me. I don't think there's any requirement that test names are 3 letters.

jgfouca:
I think I have a better way of addressing this issue. Titan just needs more nodes in the 2-thread case.

rljacob:
Another suggestion is add an option to PET so that it does 2 batch submissions.

jgfouca:
@rljacob yes, I was thinking the same thing until I realized that that option would mean PET could no longer inherit SystemTestCompareTwo and would therefore be a full rewrite of the test. I think the node doubling should work.

jgfouca:
@billsacks ok, I think I addressed your concern.

billsacks:
Thanks, @jgfouca ! I have no idea what the general implications are of changing case.num_nodes, so will let @jedwards4b review this.

scripts/lib/CIME/SystemTests/pet.py
```diff
@@ -30,6 +30,10 @@ def _case_one_setup(self):
             if self._case.get_value("NTHRDS_%s"%comp) <= 1:
                 self._case.set_value("NTHRDS_%s"%comp, 2)
 
+        # Subtle: machines like titan will not let you overload the number of threads
```
jedwards4b: I don't understand why there was any overloading going on - the PET uses full nodes on the first test and undersubscribes on the second one.   Isn't this just a matter of updating the aprun command correctly for each of (between) the two runs?

jgfouca: @jedwards4b if you double the threads without reducing the tasks, you'd expect to have double the total threads. On a machine that allows you to put 2 threads on a single core, this isn't a problem, but apparently aprun/titan doesn't let you do that. Keep in mind that the num_nodes has already been computed at this point in the case object, so that fact the more threads requires more nodes on titan is lost unless we update by hand like we do here.

jedwards4b: In pet.py _case_one_setup you set NTHRDS = 2 for each component **then you run case_setup to recreate the case.run and case.test scripts for the correct node and task setup**
In _case_two_setup you set NTHRDS = 1 and MAX_TASKS_PER_NODE to half it's original value 
and run case.setup again to recreate the case.run and case.test scripts for the same number of nodes but undersubscribed.    I think that the issue that you are having is that the aprun command line is not getting updated to be consistent with these changes.

rljacob:
If the OOB case already has NTHRDS > 1, PET just uses that right?  It only sets it to 2 for components that don't already have NTRHDS >1.

jgfouca:
@rljacob yes that's correct. I think titan defaults to 1 thread.

scripts/lib/CIME/SystemTests/pet.py
jgfouca: @jedwards4b , I've been thinking and you're right. The real issue is that critical values that are set up in case._initialize_derived_attributes are not updated when case_setup is called, even though the values that went in to compute those derived attributes may have changed. I'm going to test a different fix.

jgfouca:
@jedwards4b , OK this is ready for a re-review.

rljacob:
Is the PR title still accurate?

jgfouca:
@rljacob it is now.

PR #1419, opened 2017-04-26
Xmlcq fileoption

### LINKED ISSUE #1216: Unable to set info_debug in the coupler namelist for tests
When I setup a test, I no longer have control of the info_debug namelist in the driver.

It used to be (years ago) that I could setup an SMS test, set INFO_DBUG to 2 in env_run.xml, and the driver would then run with info_debug set to 2 for any test.

Then a couple years ago, I could no longer set INFO_DBUG to 2 in env_run because the SMS test automatically reset it to 1 when it ran.  But I could force info_debug to 2 by setting that value in user_nl_cpl.  I wasn't a fan of this implementation but at least could work around it.

As of today, setting info_debug to 2 in user_nl_cpl for an SMS test produces

"ERROR: Cannot change info_debug in user_nl_xxx file, INFO_DBUG"

when I submit.  So I no longer have control of the info_debug namelist setting.  My feeling has always been that the tests should NOT overwrite a user change like this in the env_ settings.  There are some settings that the test has to change (like run length, restart info, etc in ERS tests), but there is other info that a user should be able to change (like INFO_DBUG).

Anyway, there seems to be some conflicting logic with regard to NOT allowing users to set namelist that are set via env_*.xml files and the fact that tests overwrite a number of env_*.xml values as it runs, thereby overwriting values that users might have set.

### LINKED ISSUE #1416: xmlchange does not enforce --file option
When using the `--file` option to `xmlchange`, incorrect or bogus files are not trapped. For instance:
`./xmlchange --file env_test.xml INFO_DBUG=3`
sets `INFO_DBUG` in env_run.xml instead of env_test.xml. Worse:
`./xmlchange --file blarg INFO_DBUG=2`
also sets the value in env_run.xml.

jedwards4b:
Add --file option to xmlquery and enable option in xmlchange.   Add code to allow changes in env_test.xml

Test suite: scripts_regression_tests.py hand tests of xmlchange and xmlquery
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1216 
FIxes #1416

User interface changes?:  new --file option in xmlquery

Code review:

### DIFF
--- a/scripts/Tools/xmlchange
+++ b/scripts/Tools/xmlchange
@@ -14,7 +14,6 @@ from standard_script_setup import *
 
 from CIME.utils import expect, convert_to_type, append_case_status
 from CIME.case import Case
-from CIME.XML.generic_xml import GenericXML
 
 import re
 
@@ -96,12 +95,7 @@ def xmlchange(caseroot, listofsettings, xmlfile, xmlid, xmlval, subgroup,
     with Case(caseroot, read_only=False) as case:
 
         if xmlfile:
-            expect(os.path.isfile(xmlfile), "Could not find file %s"%xmlfile)
-            gfile = GenericXML(infile=xmlfile)
-            ftype = gfile.get_id()
-            if os.path.abspath(os.path.join(caseroot,ftype)) != os.path.abspath(xmlfile):
-                logger.warn("setting case file to %s"%xmlfile)
-                case.set_file(xmlfile, ftype)
+            case.set_file(xmlfile)
 
         env_mach_pes = case.get_env("mach_pes")
         env_mach_pes.set_components(case.get_values("COMP_CLASSES"))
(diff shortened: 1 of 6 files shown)


gold2718:
I have a behavioral question. Is this expected?
```
% ./xmlquery --file env_build.xml INFO_DBUG
setting case file to env_build.xml
        INFO_DBUG: None
```
I guess I expected something like:
```
ERROR:  No results found for variable INFO_DBUG in env_build.html
```
which would mimic the error for a non-existent variable. Even the default error would be okay.
```
ERROR:  No results found for variable INFO_DBUG
```

gold2718:
`xmlquery` is behaving better but:
```
 ./xmlchange --file env_build.xml INFO_DBUG=2
setting case file to env_build.xml
ERROR: Could not find object for env_mach_pes.xml in case
```
just doesn't seem right.

jedwards4b:
@gold2718 I've addressed your comments - thanks.   @jgfouca can you review?

PR #1427, opened 2017-04-26
add missing term for NET WATER BUDGET tables

### LINKED ISSUE #1429: CIME Component Freshwater Budget incomplete
While tracking down an inconsistency in the freshwater budget between POP and what is reported in the coupler tables, I found that the coupler freshwater budget calculations were incomplete.  I've filed a pull request (#1427) to correct the error but would like to make sure this is not also a problem for ACME.

jtruesdal:
Correct a budget diagnostic printed in the coupler log files.  The fresh water budget was missing a freshwater term associated the salt flux due to ice melt/growth.  This salt flux should be  represented as an additional  freshwater flux in the coupler water budget.

The complete equation for net surface forcing of the POP model is 

 PREC_F+EVAP_F+ROFF_F+IOFF_F+MELT_F+SALT_F*(sflux_factor/salinity_factor) + QFLUX/lh_fusion*1e4*s_scaling

The freshwater representation of the SALT_F term was missing from the coupler budget.  I've added that term and verified that the net ocean fresh water budget better represents the freshwater tendency calculated by the ocean component.  I doubled checked the net budget against the global salinity trend in a model run.  The new term in the table for the missing flux is weqsalt which stands for the fresh water equivalent of the salt flux. 

Test suite: cime regression tests and a coupled run where I backed out the freshwater flux from the salinity timeseries.
Test baseline: cime_master 
Test namelist changes: 

Fixes #1429 

Code review: Tony, Rob, and Mariana.

### DIFF
--- a/src/drivers/mct/main/seq_diag_mct.F90
+++ b/src/drivers/mct/main/seq_diag_mct.F90
@@ -78,6 +78,13 @@ module seq_diag_mct
    &  - (shr_const_ocn_ref_sal-shr_const_ice_ref_sal) / &
    &    (shr_const_ocn_ref_sal*shr_const_latice)
 
+   real(r8),parameter :: SFLXtoWFLX = & ! water flux implied by salt flux
+                                        ! WFLX (kg/m^2s) = -SFLX (kg/m^2s) 
+                                        !                  / ocn_ref_sal (psu) (34.7g/kg)
+					!		   / 1.e-3 kg/g
+        -1._r8/(shr_const_ocn_ref_sal*1.e-3_r8)
+
+
    !--- C for component ---
    !--- "r" is recieve in the coupler, "s" is send from the coupler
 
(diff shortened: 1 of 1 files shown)


mvertens:
@rljacob - We need this PR for the upcoming CESM code freeze scheduled for May 19th. We also need it as soon as possible in CESM testing for the code freeze. What is the status of your approval?
Simulations on our end have shown that it improves the total budget.

rljacob:
I still need to do a simulation with and without it.  We don't think this term is necessary for MPAS.

mvertens:
How do you want to proceed? Can you do this simulation quickly so we can
know if we need do have cesm ifs around these terms - the way we already
have for some other terms? Would that be a better way to speed up this
process - that way this change would not effect ACME at all.

mvertens:
Thinking about this some more I think there are three possible steps forward. The basic issue is that the form of the water conservation table has changed by the introduction of a new column. There is no easy way I can see - given the existing implementation of budgets in seq_diag_mct.F90 to easily have different water budget tables for acme versus cesm. 

Given this, I think there are 3 possibilities.
1) ACME accepts this PR as is
2) We simply set the new water equivalent of salt flux to zero for ACME - but keep the table in its current form - thereby not impacting the current ACME budget 
3) We have two seq_diag_mct.F90 files  -and replace seq_diag_mct.F90 with  seq_diag_cesm.F90 and seq_diag_acme.F90 - thereby eliminating the need for ACME and CESM to coordinate their water and energy budget calculations. This might simplify things moving forwards.

@rljacob - how do you want to proceed, given that CESM now has a very tight timeline to get this into its release code base?

rljacob:
I prefer 1 but need to check.   2 if we can't.

rljacob:
Is Fioi_salt a new attribute that the sea ice passes to the coupler?

rljacob:
Looks like its not and mpas-cice sets that to 0 so I think we can go with 1 and if I find a problem with further testing I'll make a new PR.

mvertens:
@rljacob - Thanks for the quick response. Can I go ahead and merge this? You were the assignee for the merge - but I'm happy to do it.

akturner:
@rljacob: What do you mean MPAS-Seaice sets Fioi_salt to zero?

rljacob:
That was based on my quick read of the code which must have been wrong.

akturner:
If MPAS-Seaice is producing zeros for this field then that is a MAJOR bug

jonbob:
@rljacob - if that flux ever is zero, we have a major problem.  We may need to think about this PR and test it before implementing, because we do use those fields and are concerned about the budgets.  Chris has asked me to regenerate water and energy budgets again for the next runs, and we need to make sure this works for us.

rljacob:
Specifically:
i2x_i % rAttr(index_i2x_Fioi_salt ,n) = oceanSaltFlux(i) in ice_comp_mct.F.  oceanSaltFlux(i) is initialized to be zero and I didn't see where it was updated.  I now see its passed to ice_colpkg.

I have no knowledge about it being zero or not in a real run.  It just wasn't in the coupler diagnostic.

PR #1432, opened 2017-04-27
Update the top-level READMEs

rljacob:
Remove the "README" file which was an un-interesting (to users) list of the svn dirs that cime started with.  This is now a page in the wiki.

Update README.md with links to our documentation and ACME.  Also add more developers.  Modify the brief description of CIME.

Test suite:  none.

BFB


rljacob:
wiki page with the "README" content is here:  https://github.com/ESMCI/cime/wiki/History

PR #1436, opened 2017-04-27
Add --component Option to preview_namelist

### LINKED ISSUE #887: Add --component option to preview_namelists
It would be useful if users could specify a single component when running `preview_namelists` rather than building namelists for all components when the script is run from the command line in a case directory.

erichlf:
Added an option to preview_namelist, which allows the user to specify
the component for which to run the preview_namelist script.

Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #887

User interface changes?: added --component option to preview_namelist

Code review: @mnlevy1981, @jgfouca

### DIFF
--- a/scripts/Tools/preview_namelists
+++ b/scripts/Tools/preview_namelists
@@ -15,10 +15,10 @@ from CIME.utils             import expect
 import argparse, doctest
 
 ###############################################################################
-def parse_command_line(args):
+def parse_command_line(args, description):
 ###############################################################################
     cime_model = CIME.utils.get_model()
-    parser = argparse.ArgumentParser()
+    parser = argparse.ArgumentParser(description=description)
     CIME.utils.setup_standard_logging_options(parser)
 
     parser.add_argument("caseroot", nargs="?", default=os.getcwd(),
@@ -37,15 +37,14 @@ def parse_command_line(args):
 ###############################################################################
 def _main_func(description):
 ###############################################################################
-    args = parse_command_line(sys.argv)
+    args = parse_command_line(sys.argv, description)
     if args.test:
         test_results = doctest.testmod(verbose=True)
         sys.exit(1 if test_results.failed > 0 else 0)
 
-    caseroot = args.caseroot
-    expect(os.path.isfile(os.path.join(caseroot, "CaseStatus")),
+    expect(os.path.isfile(os.path.join(args.caseroot, "CaseStatus")),
            "case.setup must be run prior to running preview_namelists")
-    with Case(caseroot, read_only=False) as case:
+    with Case(args.caseroot, read_only=False) as case:
         create_namelists(case, component=args.component)
 
 if (__name__ == "__main__"):
--- a/scripts/Tools/preview_namelists
+++ b/scripts/Tools/preview_namelists
@@ -15,50 +15,38 @@ from CIME.utils             import expect
 import argparse, doctest
 
 ###############################################################################
-def parse_command_line(args, description):
+def parse_command_line(args):
 ###############################################################################
-    parser = argparse.ArgumentParser(
-        usage="""\n%s [--debug]
-OR
-%s --verbose
-OR
-%s --help
-OR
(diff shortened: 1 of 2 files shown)


scripts/Tools/preview_namelists
```diff
@@ -15,50 +15,38 @@ from CIME.utils             import expect
...
     CIME.utils.handle_standard_logging_options(args)
 
-    return args.caseroot
+    return args
 
 ###############################################################################
 def _main_func(description):
 ###############################################################################
-    if ("--test" in sys.argv):
+    args = parse_command_line(sys.argv)
```
jgfouca: In our other tools, we unpack args before returning it. It might be better to stay consistent.

billsacks: I personally find the unpacking of args to be tedious and error-prone, particularly when the argument list gets above 10 or so arguments. While I see the argument for consistency, I see a bigger argument for starting to migrate to returning the entire args structure.

gold2718: The problem I have with passing args is that it ties the user interface to lower code levels. I prefer the unpack.

billsacks: I agree that it's bad to tie user interface to lower-level code, but I feel this is worse:

```python
    output, build_dir, build_optimized, clean,\
        cmake_args, compiler, enable_genf90, machine, machines_dir,\
        make_j, use_mpi, mpilib, mpirun_command, test_spec_dir, ctest_args,\
        use_openmp, xml_test_list, verbose \
        = parse_command_line(sys.argv)
```

It's really hard to see if I have swapped some of the arguments in that list.

I agree that lower-level functions should accept specific arguments, but from the main program, it seems much better to me to replace the above with

```python
args = parse_command_line(sys.argv)
```

and then you can access the individual arguments as attributes of the args object. I do not see any downsides from either a practical standpoint or a philosophical elegance standpoint: If anything, I'd say that the latter is *less tied* to the user interface, because if there are arguments you don't care about, you don't have to list them explicitly here.

erichlf: I like the args being unpacked because then I only need to add an argument and don't have to then change the return and the function call itself. Much cleaner and simpler.

scripts/Tools/preview_namelists
```diff
@@ -15,50 +15,38 @@ from CIME.utils             import expect
 import argparse, doctest
 
 ###############################################################################
-def parse_command_line(args, description):
+def parse_command_line(args):
```
jgfouca: File does have a description, so it should be given to user with --help.

scripts/Tools/preview_namelists
jgfouca: It probably doesn't matter too much, so I won't hold up the PR based on this.

I understand that unpacking leads to more typing, but I think it's worth it for the reason that @gold2718 already described.

jgfouca: @billsacks you're going to have to unpack at some point, as we'd never want API that expects the args object. Why not unpack right away and be done with it? Another benefit, you don't have to prefix everything with args.

scripts/Tools/preview_namelists
erichlf: Fixed

scripts/Tools/preview_namelists
billsacks: > you're going to have to unpack at some point, as we'd never want API that expects the args object. Why not unpack right away and be done with it?

(1) At least with the scripts I've dealt with, the particular functions called by the main program typically only take a few of the total command-line arguments. Having, say, 4 positional arguments is less error-prone than having a 20-element positional list.

(2) Even if the main program turns around and immediately passes all arguments to some other routine, this can be done via keyword arguments rather than positional arguments - so again, much less error-prone.

> Another benefit, you don't have to prefix everything with args.

I'd gladly type a few extra characters for each argument if it frees me from the stress-inducing modification of these long position-dependent lists.

PR #1441, opened 2017-04-28
Pio default settings

### LINKED ISSUE #1433: pio strides reset by default in CIME
On master (cime5.3.0-alpha.07-144-g8924bb3) case.setup always resets
pio strides/numtasks to defaults (PES_PER_NODE) independent of the current
values in env_run.xml, if I don't specify "--no-adjust-pio". 
The default value of PIO_STRIDE in env_run.xml (after running ./create_newcase,
and before ./case.setup is run) comes from config_pio.xml .
 There are two cases that fail due to ./case.setup adjusting pio strides/numiotasks by
default,
* There is a machine specific setting for stride in config_pio.xml 
* The user manually modifies env_run.xml, to set pio strides to something different
  from PES_PER_NODE, after creating a new case (before ./case.setup is run)

jedwards4b:
remove scripts adjustment of pio settings - more trouble than its worth.

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1433

User interface changes?: removed --no-adjust-pio flag from case.setup and never attempt to adjust pio settings.  

Code review:  @jayeshkrishna

### DIFF
--- a/scripts/Tools/case.setup
+++ b/scripts/Tools/case.setup
@@ -45,16 +45,11 @@ OR
     parser.add_argument("-r", "--reset", action="store_true",
                         help="Does a clean followed by setup")
 
-    parser.add_argument("--no-adjust-pio", action="store_true",
-                        help="Do NOT adjust pio settings for new pelayout."
-                                  "By default if the pelayout is changed the pio layout will be adjusted."
-                                  "This option overrides that feature and leaves the pio layout as currently set.")
-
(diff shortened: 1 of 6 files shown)


jayeshkrishna:
Verified that case_setup no longer resets the stride explicitly specified by the user (1. via config_pio.xml 2. Modifying it in env_run.xml after create_newcase, before running case_setup)

PR #1445, opened 2017-05-01
Erp fix again

### LINKED ISSUE #1443: ERP test is broken
When I try to create an ERP test for CESM with the latest master I get the following error in the build:

```
2017-04-29 17:13:15: Exception during build:
ERROR: 
ERROR, SMP VALUES HAVE CHANGED
  SMP_BUILD = c1a1l1i1o1r1g1w1e1
  SMP_VALUE = c1a1l1i1o1r1g1w1e1
  smpstr = c0a0l0i0o0r0g0w0e0
  A manual clean of your obj directories is strongly recommended
  You should execute the following:
    ./case.build --clean
  Then rerun the build script interactively
  ---- OR ----
  You can override this error message at your own risk by executing:
    ./xmlchange -file env_build.xml -id SMP_BUILD -val 0
  Then rerun the build script interactively

```
I have narrowed down the problem to the following hash: 

`670cf97 Merge remote-tracking branch 'origin/master' into jgfouca/new_query_tools 

I believe the problem is related to the fact that when we are halving the threads - (which in the case of yellowstone goes from 2->1) the build-threaded option should NOT change. That was always the intent of this test. You want to keep the threading compilation the same. 

@jedwards noticed yesterday that this test was working on cheyenne - but this is a red herring since on cheyenne the threads are set to 1 in the initial pe-layout and so halving them keeps the threads at 1. On yellowstone the threads start at 2- so halving them gives the bad behavior we are seeing.

This is now a critical issue in our testing for the CESM2 release.

jedwards4b:
Needed to move build threaded flag to before shared lib build in erp test.  Test and fix issues with nag compiler.
                   PASS ERP_Ln9.f09_f09.F1850_DONOTUSE.yellowstone_intel.cam-outfrq9s_clm5 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1443 

User interface changes?: 

Code review: mvertens

### DIFF
--- a/config/cesm/machines/config_compilers.xml
+++ b/config/cesm/machines/config_compilers.xml
@@ -283,7 +283,6 @@ using a fortran linker.
     <!-- Runtime checks with OpenMP (in fact, all OpenMP cases) are WIP.       -->
     <append DEBUG="TRUE"> -C=all -g -time -f2003 -ieee=stop </append>
     <append DEBUG="TRUE" compile_threaded="false"> -gline </append>
-    <append compile_threaded="true"> -openmp </append>
     <!-- The SLAP library (which is part of the CISM build) has many instances of
        arguments being passed to different types. So disable argument type
        checking when building CISM. This can be removed once we remove SLAP from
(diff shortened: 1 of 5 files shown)


mvertens:
Did you run the CESM ERP test that failed before to verify that this fix also did not break that test. If so - could you please document that. If not - could you please run that test as well.

jedwards4b:
No, I can run it again.

Issue #1439, opened 2017-04-28
CIME should not over-ride anything explicitly set by the user


rljacob:
CIME should not over-ride anything explicitly set by the user.    Let it fail.

rljacob:
https://en.wikipedia.org/wiki/Principle_of_least_astonishment

jgfouca:
@rljacob I think I addressed the concern that led you to make this ticket.

PR #1450, opened 2017-05-01
Update queue selection to take walltime into account

### LINKED ISSUE #1255: Queue selection in env_batch is not adequate for blues
Some background: blues has two available queues, "shared" and "batch". The shared queue will throw an error if you try to submit a job and request more than an hour of runtime. The "shared" queue will accept num cores 1-64, the "batch" queue will accept 1-thousands. My problem is that we have a lot of tests that take 64 cores (fit in either queue), but some take three hours and some take one. env_batch.select_best_queue seems unable to select the appropriate queue for such jobs (1 hour or less should go to "shared" and the others should go to "batch"), instead it just selects "shared" for everything.

jgfouca:
Adds concept of strict walltime.

The idea here is to have better support for machines like blues that have a "debug" queue and a "standard" queue. The "debug" has strict limits on both walltime and num_pes and therefore should not be selected as the user's queue if they asked for a long walltime. For other machines, the maxwalltime setting is being used more like a default walltime than a true max.

Test suite: scripts_regression_tests (melvin and skybridge) and some by-hand testing on blues
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1255 

User interface changes?: Changes in how walltime is handled

Code review: @jedwards4b

### DIFF
--- a/config/acme/machines/config_batch.xml
+++ b/config/acme/machines/config_batch.xml
@@ -158,7 +158,7 @@
         <directive>-l nodes={{ num_nodes }}:ppn={{ tasks_per_node }}</directive>
       </directives>
       <queues>
-	<queue walltimemax="01:00:00" jobmin="1" jobmax="64">shared</queue>
+	<queue walltimemax="01:00:00" jobmin="1" jobmax="64" strict="true">shared</queue>
 	<queue walltimemax="03:00:00" jobmin="64" jobmax="4096" default="true">batch</queue>
       </queues>
     </batch_system>
@@ -177,8 +177,8 @@
     <!-- edison is SLURM as of Jan-4-2016 -->
     <batch_system MACH="edison" type="slurm" >
       <queues>
+	<queue walltimemax="00:30:00" jobmin="1" jobmax="12288" strict="true">debug</queue>
         <queue walltimemax="01:30:00" jobmin="1" jobmax="150000" default="true">regular</queue>
-	<queue walltimemax="00:30:00" jobmin="1" jobmax="12288">debug</queue>
       </queues>
     </batch_system>
 
@@ -198,8 +198,8 @@
         <directive> --constraint=haswell</directive>
       </directives>
       <queues>
+	<queue walltimemax="00:30:00" jobmin="1" jobmax="4096" strict="true">debug</queue>
         <queue walltimemax="01:00:00" jobmin="1" jobmax="10000" default="true">regular</queue>
-	<queue walltimemax="00:30:00" jobmin="1" jobmax="4096">debug</queue>
       </queues>
     </batch_system>
 
@@ -208,8 +208,8 @@
         <directive> --constraint=knl,quad,cache</directive>
       </directives>
       <queues>
+	<queue walltimemax="00:30:00" jobmin="1" jobmax="100000" strict="true">debug</queue>
         <queue walltimemax="01:00:00" jobmin="1" jobmax="3000000" default="true">regular</queue>
-	<queue walltimemax="00:30:00" jobmin="1" jobmax="100000">debug</queue>
       </queues>
     </batch_system>
 
@@ -329,8 +329,8 @@
        <directive>-l nodes={{ num_nodes }}</directive>
      </directives>
      <queues>
+       <queue walltimemax="01:00:00" jobmin="0" jobmax="64" strict="true">debug</queue>
        <queue walltimemax="24:00:00" default="true">batch</queue>
-       <queue walltimemax="01:00:00" jobmin="0" jobmax="64">debug</queue>
      </queues>
    </batch_system>
 
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -7,7 +7,7 @@ import math
 from CIME.XML.standard_module_setup import *
 from CIME.utils import format_time
 from CIME.XML.env_base import EnvBase
-from CIME.utils import transform_vars, get_cime_root
+from CIME.utils import transform_vars, get_cime_root, convert_to_seconds
 from copy import deepcopy
 
 logger = logging.getLogger(__name__)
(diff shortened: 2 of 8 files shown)


jgfouca:
This is worth a pretty close look. Significant change.

config/acme/machines/config_batch.xml
```diff
@@ -158,7 +158,7 @@
         <directive>-l nodes={{ num_nodes }}:ppn={{ tasks_per_node }}</directive>
       </directives>
       <queues>
-	<queue walltimemax="01:00:00" jobmin="1" jobmax="64">shared</queue>
+	<queue walltimemax="01:00:00" jobmin="1" jobmax="64" strict="true">shared</queue>
```
jgfouca: "strict" is to deal with the fact that sometimes walltimemax is acts as a default and sometimes it's truly the longest walltime that a queue will accept.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -195,18 +195,31 @@ def set_job_defaults(self, batch_jobs, pesize=None, walltime=None, force_queue=N
...
-
-            walltime = self.get_max_walltime(queue) if walltime is None else walltime
-            if walltime is None:
-                logger.warn("Could not find a queue matching task count %d, falling back to deprecated default walltime parameter"%task_count)
-                #if the user names a queue which is not defined in config_batch.xml and does not set a
-                #walltime, fall back to the max walltime in the default queue
-                if force_queue:
-                    self.get_default_queue()
-                walltime = self._default_walltime
+            if force_queue:
```
jgfouca: This is worth a close look. The behavior is basically the same as before expect that the algorithm will override a users requested walltime if the user walltime is causing us to fail to match any queues.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -418,36 +430,64 @@ def get_job_id(self, output):
         jobid = search_match.group(1)
         return jobid
 
-    def select_best_queue(self, num_pes, job=None):
+    def queue_meets_spec(self, queue, num_pes, walltime=None, job=None):
```
jgfouca: Basically just an encapsulation of what select_best_queue was doing.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -418,36 +430,64 @@ def get_job_id(self, output):
...
-                elif jobmin is not None and jobmax is not None and num_pes >= int(jobmin) and num_pes <= int(jobmax):
-                    return queue.text
+                qname = queue.text
+                if self.queue_meets_spec(qname, num_pes, walltime=walltime, job=job):
+                    return qname
+
         return None
 
-    def get_max_walltime(self, queue):
+    def get_queue_specs(self, queue):
```
jgfouca: Just a generalization of getting properties of a queue.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -418,36 +430,64 @@ def get_job_id(self, output):
...
+                return jobmin, jobmax, jobname, walltimemax, strict
+
+        expect(False, "Queue '%s' is unknown to this system" % queue)
 
     def get_default_queue(self):
         node = self.get_optional_node("queue", attributes={"default" : "true"})
         if node is None:
             node = self.get_optional_node("queue")
         expect(node is not None, "No queues found")
-        self._default_walltime = node.get("walltimemax")
```
jgfouca: I wasn't comfortable with a method called "get_default_queue" having an important side effect.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -418,36 +430,64 @@ def get_job_id(self, output):
...
+        if job is not None and jobname is not None:
+            return jobname == job
+
+        if jobmin is not None and num_pes < int(jobmin):
+            return False
+
+        if jobmax is not None and num_pes > int(jobmax):
+            return False
+
+        if walltime is not None and walltimemax is not None and strict:
```
jgfouca: The "strict" concept is implemented here.

scripts/lib/CIME/XML/env_batch.py
rljacob: In another issue (#1439), I was saying we should never override something explicitly set by the user.  The lower walltime you fill in may not be long enough for what the user intended to do.  They won't notice CIME submitted with a different walltime, the experiment will run and burn real time, and they they'll find it didn't run as long as they wanted/asked for.

jgfouca:
@rljacob the issue here is that automatically overriding makes this work better with create_test.

scripts/lib/CIME/XML/env_batch.py
jgfouca: Yeah, that true. I'm wondering how to support the following case:

A test that uses few procs but takes a long time, the test lives in acme_integration.

1) Due to being in acme_integration, 3 hours is requested
2) Due to using a small number of procs, blues will try to put it in the debug queue

With the changes below, the walltime is automatically reduced and the test will at least have a chance to run.

jgfouca: I could change the warning to something more conspicuous.

```
WARNING
WARNING Requested walltime '%s' could not be matched by any queue, using '%s' instead
WARNING
```

rljacob:
But for the majority of users who run create_newcase and then spend days/weeks fiddling with xml an user_nl variables, its better to not undo their work.  You can't print a warning big enough. 

How about another command-line option?  --override-walltime

jedwards4b:
I don't think that you need a new command line option - how about if you only override the command line if it's a test.   If calling directly from create_newcase then respect the command line.

rljacob:
Sure!  I didn't realize you could detect if its called from a test.  That works.

jgfouca:
OK, will do (override if test).

config/acme/machines/config_batch.xml
jayeshkrishna: Wouldn't adding "walltimedefault" be a better option? I always read "walltimemax" as the maximum walltime that is allowed for that queue.
(I feel the "strict" flag somehow doesn't convey the intent)

jayeshkrishna:
Like @rljacob mentioned above, If the idea is to reduce the time taken for tests, using a " --walltime" option to create_test (set walltime for all tests) for blues nightly tests would work, right (All nightly acme_integration test runs on blues would use "--walltime=1hr". By default, the acme_integration tests run by users without the "--walltime" option would go into slower queues.)?
It might be harder to understand the logic of resetting the walltime for tests if it is inside the scripts.

jgfouca:
@jayeshkrishna the default behavior for create_test is for the walltime of an individual test to be determined by the test_suite it lives in. For example, acme_developer tests have a limit of 45 minutes. Sometimes, the processors and walltimes requested by a test are not compatible with any queue available to the system. In this case, the walltime would be adjusted by env_batch with this PR.

jgfouca:
OK, I added tests and changed the behavior so that only tests can have their walltime overridden. This is ready for a final review.

jayeshkrishna:
@jgfouca : So adding a "--walltime" option to create_test should make sense, right (All tests created by this create_test command need to use this custom walltime, ./create_test acme_developer --walltime=01:00:00 implies all tests created would use 1hr as the walltime)?

jgfouca:
@jayeshkrishna all tests would *try* to use 01:00:00 as the walltime. If the system's queues could not handle that walltime for a certain test, then it would be adjusted. I figured this was better than just letting case.submit crash.

jayeshkrishna:
But there are queues in blues that can handle a walltime > 01:00:00

jgfouca:
@jayeshkrishna and I talked offline and he's cool with this PR.

jayeshkrishna:
@jgfouca mentioned the case below that would not run on blues without this modification,
* A test with 32 procs + 2hr walltime
As of now, the batch queue has a minimum requirement (jobmin) of 32 procs and the shared queue has a maximum walltime (walltimmax) of 1hr

PR #1451, opened 2017-05-02
New DOCN aquaplanet functionality

mvertens:
Creation of new aquaplanet capability in DOCN

Created new aquaplanet capability for both SOM and prescribed SST modes.
The prescribed modes are analytic and duplicate the capability in the CAM aquap (which is in fact part of CAM). This PR separates that out so that it is now in DOCN. As part of this, changes needed to be introduced so that no data was read if analytic data is present. 
In addition, aquaplanet capability has been introduced so that no changes in the domain files need to be made. This is done by overwriting the mask and fraction in aquaplanet mode with 1.

For the prescribed aquaplanet mode1, I have verified that the SST sent back to CAM is at most roundoff from that obtained with the current CAM aquap mode. However, the data ocean  areas used by CAM and sent to the coupler versus those sent by DOCN are different, and the flux correction calculation in the coupler introduces greater than roundoff differences in the solution due to differences in these areas.

Test suite: scripts_regression test
  also verified that the following tests are bfb with cesm2_0_alpha06j
     SMS_D_Ln9.f09_f09.FWAMIP.yellowstone_intel.cam-reduced_hist3s.C.aquaptest2/
     SMS_Ld1.f09_f09.FW1850.yellowstone_intel.cam-reduced_hist1d.C.aquaptest1/
Test baseline: cesm2_0_alpha06j
Test namelist changes: 
Test status: bit for bit (except for new prescribed aquaplanet mode)

Fixes: None
User interface changes?: None
Code review:

### DIFF
--- a/src/components/data_comps/docn/docn_comp_mod.F90
+++ b/src/components/data_comps/docn/docn_comp_mod.F90
@@ -297,6 +297,7 @@ subroutine docn_comp_init( EClock, cdata, x2o, o2x, NLFilename )
 
     ocn_mode = trim(SDOCN%dataMode)
 
+    ! Special logic for aquaplanet
     if (ocn_mode(1:5) == 'AQUAP') then
        if (len_trim(ocn_mode) == 6) then
           read(ocn_mode(6:6),'(i)') aquap_option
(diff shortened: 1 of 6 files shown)


src/components/data_comps/docn/docn_comp_mod.F90
```diff
@@ -587,16 +629,13 @@ subroutine docn_comp_run( EClock, cdata,  x2o, o2x)
...
-!  lsize = mct_avect_lsize(x2o)
-!  nflds_x2o = mct_avect_nRattr(x2o)
-
-!   do nf=1,nflds_x2o
-!   do n=1,lsize
-!     ?? = x2o%rAttr(nf,n)
-!   enddo
-!   enddo
-
+   !  lsize = mct_avect_lsize(x2o)
```
jedwards4b: do we need to keep this commented code?

rljacob:
Please provide a better PR title.

src/components/data_comps/docn/docn_comp_mod.F90
mvertens: I just removed it. I agree that it was confusing.

PR #1454, opened 2017-05-02
remove empty mpi-serial settings from config_machines.xml

### LINKED ISSUE #1220: missing mpi-serial mpirun case on hobart
hobart seems to be missing an option for mpirun for a mpi-serial build. This is with: cime5.3.0-alpha.02

It has

    <mpirun mpilib="default">
      <executable>mpiexec</executable>
      <arguments>
        <arg name="machine_file">--machinefile $ENV{PBS_NODEFILE}</arg>
        <arg name="num_tasks"> -n $TOTALPES </arg>
      </arguments>
    </mpirun>

and needs

    <mpirun mpilib="mpi-serial">
      <executable></executable>
    </mpirun>

Really can't the above be for every machine?

Sean Swenson ran into this problem trying to run with a hobart_intel build with mpi-serial on burnt, which is configured the same as hobart, but doesn't have a batch system.

Here's his details...
I created a case on cgd machine 'burnt'.  I used --mach hobart.  When building with mpi-serial, and running case.submit -- no-batch, I got this error:

ERROR: Undefined env var 'PBS_NODEFILE'

My workaround was to remove mpiexec and arguments from env_mach_specific.

jedwards4b:
Remove the requirement to have mpilib mpi-serial explicitly listed in the config_machines.xml file since it is assumed supported on all systems.   Remove the empty executable field for that library.

Test suite: scripts_regression_tests.py, preview_run (thanks for this!), SMS_Mmpi-serial.f19_g17.X.hobart_intel
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1220 

User interface changes?: no

Code review:

### DIFF
--- a/config/acme/machines/config_machines.xml
+++ b/config/acme/machines/config_machines.xml
@@ -61,7 +61,7 @@
   <NODENAME_REGEX>edison</NODENAME_REGEX>
   <TESTS>acme_developer</TESTS>
   <COMPILERS>intel,gnu,cray</COMPILERS>
-  <MPILIBS>mpt</MPILIBS>
+  <MPILIBS>mpt,mpi-serial</MPILIBS>
   <CIME_OUTPUT_ROOT>$ENV{SCRATCH}/acme_scratch/edison</CIME_OUTPUT_ROOT>
   <RUNDIR>$CIME_OUTPUT_ROOT/$CASE/run</RUNDIR>
   <EXEROOT>$CIME_OUTPUT_ROOT/$CASE/bld</EXEROOT>
@@ -185,7 +185,7 @@
     <NODENAME_REGEX>cori</NODENAME_REGEX>
     <TESTS>acme_developer</TESTS>
     <COMPILERS>intel,gnu,cray</COMPILERS>
-    <MPILIBS>mpt</MPILIBS>
+    <MPILIBS>mpt,mpi-serial</MPILIBS>
     <CIME_OUTPUT_ROOT>$ENV{SCRATCH}/acme_scratch</CIME_OUTPUT_ROOT>
     <RUNDIR>$CIME_OUTPUT_ROOT/$CASE/run</RUNDIR>
     <EXEROOT>$CIME_OUTPUT_ROOT/$CASE/bld</EXEROOT>
@@ -318,7 +318,7 @@
     <NODENAME_REGEX>cori-knl-haswell-is-default</NODENAME_REGEX>
     <TESTS>acme_developer</TESTS>
     <COMPILERS>intel,gnu,cray</COMPILERS>
-    <MPILIBS>mpt</MPILIBS>
+    <MPILIBS>mpt,mpi-serial</MPILIBS>
     <CIME_OUTPUT_ROOT>$ENV{SCRATCH}/acme_scratch</CIME_OUTPUT_ROOT>
     <RUNDIR>$CIME_OUTPUT_ROOT/$CASE/run</RUNDIR>
     <EXEROOT>$CIME_OUTPUT_ROOT/$CASE/bld</EXEROOT>
@@ -453,7 +453,7 @@
     <TESTS>acme_developer</TESTS>
     <OS>Darwin</OS>
     <COMPILERS>gnu</COMPILERS>
-    <MPILIBS>openmpi,mpich</MPILIBS>
+    <MPILIBS>openmpi,mpich,mpi-serial</MPILIBS>
     <RUNDIR>$ENV{HOME}/projects/acme/scratch/$CASE/run</RUNDIR>
     <EXEROOT>$ENV{HOME}/projects/acme/scratch/$CASE/bld</EXEROOT>
     <DIN_LOC_ROOT>$ENV{HOME}/projects/acme/cesm-inputdata</DIN_LOC_ROOT>    <DIN_LOC_ROOT_CLMFORC>$ENV{HOME}/projects/acme/ptclm-data</DIN_LOC_ROOT_CLMFORC>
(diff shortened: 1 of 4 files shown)


jayeshkrishna:
mpi-serial does not work on Mira/Cetus (and we had previously decided it wasn't worth the effort to look at it - no one uses mpi-serial on Cetus/Mira)

jedwards4b:
I don't think that this changes the current behavior of mpi-serial on cetus/mira - do you want to add a more specific error message if someone tries to use mpi-serial there?

jayeshkrishna:
On Mira I think we can currently just remove mpi-serial from the list of supported mpi libs and the user would get a meaningful error message (mpi-serial is not supported...). I wanted to retain that.

jedwards4b:
mpi-serial is now included by default in the  list of mpilibs - we need a special case for mira

jayeshkrishna:
@jedwards4b : How often is mpi-serial used (in your experience)? I am not sure if the library is useful to HPC devs anymore.

jedwards4b:
It's used regularly for testing - mira/cetus is the only machine it doesn't work on.

jayeshkrishna:
So should we fix it on Mira/Cetus rather than adding a special case?

And why is it used for testing, why not build with an MPI library?

rljacob:
mpi-serial was meant for developers working on their laptop or workstation so they can run without installing an MPI library.  If that use case doesn't exist, we don't need to support mpi-serial.  Its certainly not used by people developing/running on HPC platforms/clusters.

rljacob:
That can of worms should probably be a different issue.   For this one, we need a way of saying a machine doesn't support mpi-serial.  There may be others besides mira/cetus down the line.

jedwards4b:
How about this?
``` 
<mpirun mpilib="mpi-serial">		
    <executable> unsupported</executable>		
   </mpirun>		
```

jayeshkrishna:
Another soln would be to retain mpi-serial in MPILIBS for supported machine (machines without mpi-serial in MPILIBS don't support it) and get rid of mpirun section for mpi-serial (if the user has chosen mpi-serial you don't need mpirun settings)

ekluzek:
Let me point out some important aspects of mpi-serial. A set of users that do a lot of work with this are CLM users's doing single point simulations to compare to a single tower site. Also mpi-serial is the simplest system you can put together, so if you are porting a cime-driven model and all you care about is running on a single processor, it's going to be easier to setup mpi-serial than an mpi driven system. It's easier to get a debugger (or profiler) to be connected to mpi-serial and I think debugging, running is going to be simpler, easier, and probably faster than with mpi for a single task. It's nice to be able to show that the system works in the simplest case possible. And as I said there is a significant community of CLM users that do single point cases. Also note that CAM has SCAM the single column model, and I noticed that ocean/ice added single column capability recently as well.

jedwards4b:
It makes a lot more sense to me to have an exception for the one machine that doesn't support it than a required entry for all of the machines that do.

jedwards4b:
I've updated the PR to write a meaningful error message if config_machines.xml indicates that
mpi-serial is not supported using the syntax
```
<mpirun mpilib="mpi-serial">		
    <executable> unsupported</executable>		
   </mpirun>		
```
and added this to cetus and mira machine definitions.

jayeshkrishna:
@jedwards4b : How about leaving mpi-serial in MPILIBS for supported machines like I mentioned above (if mpi-serial is not in MPILIBS it is not supported on the machine)? And remove mpirun sections for mpi-serial (mpi-serial does not require any specific runtime env).
The MPILIBS section indicates the mpi libraries supported on a specific machine and specifying the support for mpi-serial there would make the intent clear.

rljacob:
Just leave things as they are.  It does what we need.

jgfouca:
I read the conversation chain. For what it's worth, I like @jedwards4b solution of explicitly adding unsupported for the couple machines that can't handle mpi-serial.

rljacob:
I went back and read the issue this is addressing and this comment https://github.com/ESMCI/cime/issues/1220#issuecomment-285175015 confused me because it said the requirement to list mpi-serial had already been removed.

jedwards4b:
@rljacob The problem was that although the requirement to list mpi-serial had been removed the block 
```
<mpirun mpilib="mpi-serial">		
    <executable></executable>		
   </mpirun>	
```
was still required.   This PR removes that requirement as well.

rljacob:
Ok can you do that while leaving the listing of mpi-serial in MPILIBS alone?

jedwards4b:
If you want to have the mpi-serial listed explicitly in acme config_machines.xml this PR does not disallow that.

jedwards4b:
I've added the mpi-serial back to the list of mpilibs in the acme config_machines.xml

rljacob:
If its not listed, and a user tries to build with it, will they still get an error message?

jedwards4b:
That error message is provided with the additional 
```
<mpirun mpilib="mpi-serial">		
    <executable> unsupported</executable>		
   </mpirun>		
```

rljacob:
All the issue asked to do is remove the empty mpirun blocks for mpi-serial.  Can't you do that and leave the rest alone?

jedwards4b:
No - for two reasons 
1.  It doesn't do what you seem to think it does now
2. It's not what we want on the cesm side. 

I have now updated to give you the functionality you seem to think was already there and to give us the functionality that we want.

rljacob:
No that's a bad idea too.  We need fewer acme/cesm special cases in the core library, not more.

jedwards4b:
So stop trying to create a special case for acme.

rljacob:
Fair enough.   I'm trying to find a solution that works for everybody.   Making mpi-serial in to an assumed available library has other implications that we need to think through.

jedwards4b:
But it's been that way for quite some time.

rljacob:
After talking on the CIME telecon, we decided to go with the original PR.  The error message will come from the failed build of mpi-serial.  We don't want those empty mpirun blocks in the xml.  We don't want to introduce more low-level diffs between acme and cesm.

jedwards4b:
Okay - I think that this is ready for a final review and merge.

PR #1455, opened 2017-05-02
add new CESM gx1v7 runoff maps for use with estuary parameterization.

jtruesdal:
Updated CESM mapping files that deal with gx1v7 runoff when using EBM (default for gx1v7).  These new files take a merged approach which 1) maps liquid runoff from the nearest coastal point in open ocean where EBM is active and 2) smoothly distributes liquid runoff over a spreading radius in marginal seas where EBM is currently not active.  Ice runoff is always spread.  gx3v7 maps for use with EBM are in the repository but EBM is off by default for resolution other than 1 degree.  Where EBM is not actively mixing we use the spreading maps for both liquid and ice.

Test suite: cime regression and one off tests to make sure the new datasets are being run for the gx1v7 but not for other resolutions.
Test baseline: 
Test namelist changes: 
Test status: climate changing.  Effects analysed and validated with science runs.

Fixes 

User interface changes?: 

Code review:

### DIFF
--- a/config/cesm/config_grids.xml
+++ b/config/cesm/config_grids.xml
@@ -1628,16 +1628,16 @@
     </gridmap>
 
     <gridmap rof_grid="rx1" ocn_grid="gx3v7" >
-      <map name="ROF2OCN_LIQ_RMAPNAME">cpl/gridmaps/rx1/map_rx1_to_gx3v7_e1000r500_161214.nc</map>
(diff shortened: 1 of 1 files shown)


mnlevy1981:
I talked to @klindsay28 and we do think that we should update the gx1v6 maps as well... so I'll make those maps and submit another pull request this afternoon.

PR #1462, opened 2017-05-02
in _check_for_invalid_args ignore quoted strings

### LINKED ISSUE #1460: test for -- options does not handle -- in quotes
example CLM_OPTIONS="-dothis -dothat"  causes problems (replace this with a better example)

jedwards4b:
If there is a space in an argument then it was quoted in the shell and should be ignored in this test.

Test suite: ./xmlchange --id CLM_BLDNML_OPTS      --val "-bgc bgc -crop -irrig=.true."
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1460 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/utils.py
+++ b/scripts/lib/CIME/utils.py
@@ -1238,6 +1238,9 @@ def run_and_log_case_status(func, phase, caseroot='.'):
 
 def _check_for_invalid_args(args):
     for arg in args:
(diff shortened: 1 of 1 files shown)


jgfouca:
Hope we can get away from doing our own arg parsing soon.

PR #1463, opened 2017-05-02
Quick fix of utils.py invalid parser args

agsalin:
Removing word "not" from if test. This was causing error for buildnml of cice with "-id" arg with single dash. From comment on line above, it seems clear that the True/False was reversed from intended.


agsalin:
Oops -- I jumped the gun. Code just failed differently with this fix. Need to revisit.

PR #1466, opened 2017-05-03
remove long-term archiving support in CIME

### LINKED ISSUE #1426: Remove the long-term archiver support in CIME
remove support for the long-term archiver (lt_archive.sh) and the batch submission for
this script from CIME.

bertinia:
Remove all references to long term archiving (lt_archive) in CIME for both CESM and ACME.
Update a the env_mach_specific.xsd schema so script_regression_tests pass.

Test suite: scripts_regression_tests.py and 
     create_newcase --compset A --res f45_g37_rx1
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes #1426

User interface changes?: Remove DOUT_L_* options from env_run.xml,
group case.lt_archive from env_batch.xml, and case.lt_archive from the caseroot.

Code review:

### DIFF
--- a/config/xml_schemas/env_mach_specific.xsd
+++ b/config/xml_schemas/env_mach_specific.xsd
@@ -86,13 +86,7 @@
       <xs:attribute name="name" use="required" type="xs:NCName"/>
     </xs:complexType>
   </xs:element>
-  <xs:element name="arguments">
-    <xs:complexType>
-      <xs:sequence>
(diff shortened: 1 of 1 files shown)


jedwards4b:
Why long term archiving is being removed:  

The cost of long term data storage is becoming prohibitive.
It is difficult to maintain a general interface to long term archiving because every hpc center is different.

PR #1469, opened 2017-05-03
Fixes for zlev problem in #1459

### LINKED ISSUE #1459: z level syntax stripped from atmosphere grid
Using the old (v1) config_grids.xml, one could use `z` to specify the number of levels, e.g., `ne30z60_ne30`. However, in v2, this information is stripped and not passed to the component buildcpp for processing.

jedwards4b:
The level specifier for atm and lnd grids was being ignored.
Test suite: scripts_regression_tests.py, SMS.ne30z60_ne30.F1850
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1459

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/grids.py
+++ b/scripts/lib/CIME/XML/grids.py
@@ -58,7 +58,7 @@ class Grids(GenericXML):
 
         #mechanism to specify lnd levels
         levmatch = re.match(r"(.*_)([^_]+)z(\d+)(_[^m].*)$", name)
-o        if  levmatch:
+        if  levmatch:
             lndnlev = levmatch.group(3)
             name = levmatch.group(1)+levmatch.group(2)+levmatch.group(4)
 
@@ -238,7 +238,7 @@ o        if  levmatch:
             levmatch = re.match(r"([^_]+)z(\d+)(.*)$", grid_name)
             if  levmatch:
                 grid_name_nonlev = levmatch.group(1)+levmatch.group(3)
-            levmatch = re.match(r"(.*_)([^_]+)z(\d+)(_[^m].*)$", name)
+            levmatch = re.match(r"(.*_)([^_]+)z(\d+)(_[^m].*)$", grid_name)
             if  levmatch:
                 grid_name_nonlev = levmatch.group(1)+levmatch.group(2)+levmatch.group(4)
 
--- a/scripts/lib/CIME/XML/grids.py
+++ b/scripts/lib/CIME/XML/grids.py
@@ -55,16 +55,12 @@ class Grids(GenericXML):
         if  levmatch:
             atmnlev = levmatch.group(2)
             name = levmatch.group(1)+levmatch.group(3)
-        else:
-            atmnlev = None
 
         #mechanism to specify lnd levels
         levmatch = re.match(r"(.*_)([^_]+)z(\d+)(.*)$", name)
         if  levmatch:
             lndnlev = levmatch.group(3)
             name = levmatch.group(1)+levmatch.group(2)+levmatch.group(4)
-        else:
-            lndnlev = None
 
         # determine component_grids dictionary and grid longname
         lname, component_grids = self._read_config_grids(name, compset, atmnlev, lndnlev)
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/XML/grids.py
```diff
@@ -204,6 +204,10 @@ def _read_config_grids_v2(self, name, compset):
                 lname = prefix[component_gridname]
             if model_grid[component_gridname] is not None:
                 lname += model_grid[component_gridname]
+                if component_gridname == 'atm' and atmnlev is not None:
+                    lname += "z" + atmnlev
+                elif component_gridname == 'lnd' and lndnlev is not None:
+                    lname += "z" + lndnlev
```
gold2718: I think we may need some error checking here. I tried an A compset with a grid alias of ne30z32_ne30z14 and ended up with LND_GRID="nullz14" in env_build.xml
I'm not sure this damages anything but I don't think this is intended.
Maybe modifying the land levmatch:
`levmatch = re.match(r"(.*_)([^_]+)z(\d+)(_[^m].*)$", name)`
and add an additional check to each if block above.

scripts/lib/CIME/XML/grids.py
```diff
@@ -57,7 +57,7 @@ def get_grid_info(self, name, compset):
             name = levmatch.group(1)+levmatch.group(3)
 
         #mechanism to specify lnd levels
-        levmatch = re.match(r"(.*_)([^_]+)z(\d+)(.*)$", name)
+        levmatch = re.match(r"(.*_)([^_]+)z(\d+)(_[^m].*)$", name)
```
jedwards4b: Is group 4 always expected to be present?

scripts/lib/CIME/XML/grids.py
```diff
@@ -205,9 +205,13 @@ def _read_config_grids_v2(self, name, compset, atmnlev, lndnlev):
             if model_grid[component_gridname] is not None:
                 lname += model_grid[component_gridname]
                 if component_gridname == 'atm' and atmnlev is not None:
-                    lname += "z" + atmnlev
+                    if re.search(r"a%null", lname) is None:
```
jedwards4b: You don't need a regex here   
```if not "a%null" in lname:```
 should do it.

File: scripts/lib/CIME/XML/grids.py [hunk 4]
```diff
@@ -205,9 +205,13 @@ def _read_config_grids_v2(self, name, compset, atmnlev, lndnlev):
...
+                        lname += "z" + atmnlev
+
                 elif component_gridname == 'lnd' and lndnlev is not None:
-                    lname += "z" + lndnlev
+                    if re.search(r"l%null", lname) is None:
```
jedwards4b: same here, don't use regex if you don't need to

scripts/lib/CIME/XML/grids.py
```diff
@@ -234,7 +238,7 @@ def _get_domains_v2(self, component_grids):
             levmatch = re.match(r"([^_]+)z(\d+)(.*)$", grid_name)
             if  levmatch:
                 grid_name_nonlev = levmatch.group(1)+levmatch.group(3)
-            levmatch = re.match(r"(.*_)([^_]+)z(\d+)(.*)$", grid_name)
+            levmatch = re.match(r"(.*_)([^_]+)z(\d+)(_[^m].*)$", grid_name)
```
jedwards4b: Can we reorganize so that the regex expression only appears in the code once?

mvertens:
@jedwards @gold2718  - there is a subtle problem with the naming convention we have. If you do not have a trigrid (atm/land on different grids) there is no way to distinguish if the zNN is meant for the atmosphere or for the land.

jedwards4b:
@mvertens that doesn't sound right - f09z30_f09 looks different than f09_f09z30

scripts/lib/CIME/XML/grids.py
gold2718: No but at the moment, there is no way to specify land levels if it is not present. For example:
`ne30z192_ne30` always specifies atmosphere levels. If you want to specify land levels, you need a tri-grid short name which means group 4 will be present.

scripts/lib/CIME/XML/grids.py
gold2718: Doh, will fix.

scripts/lib/CIME/XML/grids.py
gold2718: ditto

scripts/lib/CIME/XML/grids.py
gold2718: Is that the royal we?
I'll look into it.

gold2718:
but the second f09 is for the ocean/ice, not for the land. Do we really want to add land levels to the ocean grid name?

jedwards4b:
Oh - you got me there - would it work then to specify a tri-grid like this for this special case?
f09_f09z30_f09

gold2718:
That is the idea (and what I'm testing). That is why the (anti-)mask test is in the land regex.

gold2718:
Okay, changes are in. Reran scripts_regression_tests on Hobart and code checker on Yellowstone.

jedwards4b:
If you'll update the status I'll go ahead and merge.

PR #1471, opened 2017-05-03
Add skip-preview-namelist to case.submit and case_run

### LINKED ISSUE #1276: Add option to skip preview_namelist during case.run
Add an option to case.submit to skip calling preview_namelist during case.run.   This is for low-powered compute nodes that are spending a lot of time in preview_namelist.  Its up to the user to not modify any xml files while the job is in the queue.

erichlf:
Added '--skip-preview-namelist' to case.build

Added '--skip-preview-namelist' to case.build. This will allow the user
to specify if the namelist should be rebuilt from user_nl_*.

Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #1276, ACME-Climate/ACME#1466

User interface changes?: Added --skip-preview-namelist to case.build and
template.case.run

Code review: @jgfouca, @rljacob

### DIFF
--- a/config/acme/machines/template.case.run
+++ b/config/acme/machines/template.case.run
@@ -32,21 +32,21 @@ import argparse, doctest
 def parse_command_line(args, description):
 ###############################################################################
     parser = argparse.ArgumentParser(
-        usage="""\n%s [--verbose]
+        usage="""\n{} [--verbose]
 OR
-%s --help
+{} --help
 OR
-%s --test
+{} --test
 
 \033[1mEXAMPLES:\033[0m
     \033[1;32m# case.run SMS\033[0m
-    > %s
-""" % ((os.path.basename(args[0]), ) * 4),
+    > {}
+""".format(((os.path.basename(args[0]), ) * 4),
 
 description=description,
 
 formatter_class=argparse.ArgumentDefaultsHelpFormatter
-)
+))
 
     parser = argparse.ArgumentParser()
 
@@ -56,11 +56,16 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
                         help="Case directory to build")
 
     args = CIME.utils.parse_args_and_handle_standard_logging_options(args, parser)
+    parser.add_argument("--skip-preview-namelist", action="store_true",
+                        help="Skip calling preview-namelist during case.run")
 
     if args.caseroot is not None:
         os.chdir(args.caseroot)
 
-    return args.caseroot
+    if args.skip_preview_namelist is None:
+      args.skip_preview_namelist = False
+
+    return args.caseroot, args.skip_preview_namelist
 
 ###############################################################################
 def _main_func(description):
@@ -69,9 +74,9 @@ def _main_func(description):
         test_results = doctest.testmod(verbose=True)
         sys.exit(1 if test_results.failed > 0 else 0)
 
-    caseroot = parse_command_line(sys.argv, description)
+    caseroot, skip_pnl = parse_command_line(sys.argv, description)
     with Case(caseroot, read_only=False) as case:
-        success = case_run(case)
+        success = case_run(case, skip_pnl=skip_pnl)
 
     sys.exit(0 if success else 1)
 
--- a/config/acme/machines/template.case.run
+++ b/config/acme/machines/template.case.run
@@ -32,23 +32,19 @@ import argparse, doctest
 def parse_command_line(args, description):
 ###############################################################################
     parser = argparse.ArgumentParser(
-        usage="""\n{} [--verbose]
+        usage="""\n{0} [--verbose]
 OR
-{} --help
+{0} --help
 OR
-{} --test
+{0} --test
 
 \033[1mEXAMPLES:\033[0m
     \033[1;32m# case.run SMS\033[0m
-    > {}
-""".format(((os.path.basename(args[0]), ) * 4),
-
-description=description,
-
-formatter_class=argparse.ArgumentDefaultsHelpFormatter
-))
-
-    parser = argparse.ArgumentParser()
+    > {0}
+""".format(os.path.basename(args[0])),
+        description=description,
+        formatter_class=argparse.ArgumentDefaultsHelpFormatter
+    )
 
     CIME.utils.setup_standard_logging_options(parser)
 
@@ -63,7 +59,7 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
         os.chdir(args.caseroot)
 
     if args.skip_preview_namelist is None:
-      args.skip_preview_namelist = False
+        args.skip_preview_namelist = False
 
     return args.caseroot, args.skip_preview_namelist
 
--- a/scripts/Tools/case.submit
+++ b/scripts/Tools/case.submit
@@ -11,7 +11,19 @@ from CIME.case        import Case
 ###############################################################################
 def parse_command_line(args, description):
 ###############################################################################
-    parser = argparse.ArgumentParser(description=description)
+    parser = argparse.ArgumentParser(
+        usage="""\n{0} [<casedir>] [--verbose]
+OR
+{0} --help
+OR
+{0} --test
+\033[1mEXAMPLES:\033[0m
+    \033[1;32m# Setup case \033[0m
+    > {0}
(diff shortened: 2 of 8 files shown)


scripts/Tools/case.submit
```diff
@@ -11,26 +11,16 @@ from CIME.case        import Case
...
-%s --test
-
-\033[1mEXAMPLES:\033[0m
-    \033[1;32m# Setup case \033[0m
-    > %s
-""" % ((os.path.basename(args[0]), ) * 4),
-        description=description,
-        formatter_class=argparse.ArgumentDefaultsHelpFormatter
-    )
+    parser = argparse.ArgumentParser(description=description)
```
jedwards4b: Why was this code removed?

scripts/Tools/case.submit
```diff
@@ -45,27 +35,32 @@ OR
...
                         help="Used with tests only, to continue rather than restart a test. ")
 
+    parser.add_argument("--skip-preview-namelist", action="store_true",
+                        help="Skip calling preview-namelist during case.run")
+
     parser.add_argument("-a", "--batch-args",
                         help="Used to pass additional arguments to batch system. ")
 
-    args = parser.parse_args(args[1:])
+    args = parser.parse_args()
```
jedwards4b: This is in conflict with master - I recommend that you rebase and update the PR.

scripts/Tools/case.submit
```diff
@@ -45,27 +35,32 @@ OR
...
 
     CIME.utils.expect(args.prereq is None, "--prereq not currently supported")
 
     CIME.utils.handle_standard_logging_options(args)
 
-    return args.caseroot, args.job, args.no_batch, args.resubmit, args.batch_args
+    return args
```
jedwards4b: We prefer the old method of passing the individual fields of args not the args object itself.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -43,7 +43,7 @@ def set_value(self, item, value, subgroup=None, ignore_type=False):
                 elif value.count(":") == 2:
                     t_spec = "%H:%M:%S"
                 else:
-                    expect(False, "could not interpret format for wallclock time %s"%value)
+                    expect(False, "could not interpret format for wallclock time {}".format(value))
```
jedwards4b: I don't understand why these changes from %s to format were made?

erichlf: It is for python3 compatibility. Yes, we aren't using python3, but it is an easy change that doesn't hurt things and will make any future switch to python3 that much easier.

scripts/Tools/case.submit
billsacks: "We" is overly broad here. Some of us prefer passing the args object itself :-)

scripts/lib/CIME/XML/env_batch.py
billsacks: I personally prefer the new-style `format`, because I find it more readable. That said, I'm pretty sure that both are supported by python3.

scripts/Tools/case.submit
jedwards4b: I am asking for some uniformity across scripts and we have chosen to pass individual fields rather than the args object.  Same issue with format - we have standardized around %s and I see no reason to change it here.

billsacks: @jedwards4b I see your point about uniformity. I think what's going on here (if I can try to speak somewhat for @erichlf and somewhat for myself) is that there are certain coding standards like these that were put in place when the list of cime developers was very small (basically you and Jim F), and now, as there get to be more cime developers, we're not all in agreement about these coding standards.

Could I suggest, @erichlf - for now leave this as it was, but either open an issue for discussion or put this on the agenda of an upcoming Wednesday cime telecon to revisit our coding standards on these points (`%` vs. format and returning an args object vs. a big list of components).

scripts/lib/CIME/XML/env_batch.py
erichlf: I think you are right that both are currently supported, but if I recall the plan is not to support the old style.

jedwards4b: I do not object to changes that bring us closer to python3 compatibility - however I do object to having a mixture of %s and format and we are about to do a release.  I would prefer to wait until after the release and then apply this change uniformly throughout the code.

erichlf:
I believe I have addressed all issues, and now things should be working with no conflicts.

scripts/lib/CIME/XML/env_batch.py
erichlf: `{}` and `format` have already been extensively implemented throughout the code and not by me.

scripts/lib/CIME/case_run.py
```diff
@@ -120,7 +121,7 @@ def _run_model_impl(case, lid):
                         restore_from_archive(case)
 
                         orig_cont = case.get_value("CONTINUE_RUN")
-                        if not orig_cont:
+                        if not (orig_cont or skip_pnl):
```
jgfouca: Maybe @jedwards4b can help me here. I don't know if skipping create_namelists here is safe if orig_cont is False. run_model can be invoked several times for certain tests and in other situations.

jedwards4b: You still want to change the CONTINUE_RUN value in either case, so the skip_pnl should be a seperate conditional
if not orig_cont:
   case.set_value("CONTINUE_RUN", True)
   if not skp_pnl:
       create_namelists(case)

scripts/lib/CIME/XML/env_batch.py
erichlf: Absolutely happy to do so.

config/acme/machines/template.case.run
```diff
@@ -32,21 +32,21 @@ import argparse, doctest
...
 OR
-%s --test
+{} --test
 
 \033[1mEXAMPLES:\033[0m
     \033[1;32m# case.run SMS\033[0m
-    > %s
-""" % ((os.path.basename(args[0]), ) * 4),
+    > {}
+""".format(((os.path.basename(args[0]), ) * 4),
```
erichlf: Will fix. In this case it would be `{0}` since lists/arrays start at 0 in python.

config/acme/machines/template.case.run
```diff
@@ -56,11 +56,16 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
...
     args = CIME.utils.parse_args_and_handle_standard_logging_options(args, parser)
+    parser.add_argument("--skip-preview-namelist", action="store_true",
+                        help="Skip calling preview-namelist during case.run")
 
     if args.caseroot is not None:
         os.chdir(args.caseroot)
 
-    return args.caseroot
+    if args.skip_preview_namelist is None:
+      args.skip_preview_namelist = False
```
erichlf: doh!

scripts/lib/CIME/case_run.py
erichlf: Will fix.

scripts/lib/CIME/case_run.py
```diff
@@ -122,6 +123,7 @@ def _run_model_impl(case, lid):
                         orig_cont = case.get_value("CONTINUE_RUN")
                         if not orig_cont:
                             case.set_value("CONTINUE_RUN", True)
+                        if not skip_pnl:
```
jgfouca: This still isn't right. You've changed the original intent which was to re-run create_namelists if CONTINUE_RUN was False. For now, let's not worry about skip_pnl in this code. This code is for recovery from node failures which is a rare corner case and we don't need to worry about how it performs.

erichlf:
This should be ready now.

jgfouca:
@erichlf why did you assign this to @rljacob ? Usually Jim or I are the integrators.

erichlf:
@jgfouca Because he was on the origin issue.

rljacob:
With the new reviewing system in github, I'm not sure it matters.  As long as you or Jim review it.  But to get it merged quickly, probably want to re-assign it.

ndkeen:
When might we see this in ACME?

rljacob:
We are currently a little past alpha.10 we'd need alpha.12 to get this.

PR #1474, opened 2017-05-03
Only warning for single dash

jedwards4b:
Do not error on using one dash with a multi-character option. Just print a warning to stderr and move on.  It will take time to remove single-dash usage in various buildnml and buildlib and other scripts.

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes

User interface changes?:  Yes.  print a warning to stdout if single dash used with multi-char arg.

Code review:


rljacob:
Looks good.

rljacob:
Wait, should we warn the other way?  double dash with a single character argument?

PR #1475, opened 2017-05-03
Fix ACME config_machines.xml format

jgfouca:
So that it passes the XSD format.

Basically, individual module commands and environment variables do not
support attributes.

Test suite: cime_tiny
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: None

Code review: @jedwards4b

### DIFF
--- a/config/acme/machines/config_machines.xml
+++ b/config/acme/machines/config_machines.xml
@@ -542,8 +542,14 @@
         <command name="load">sems-env</command>
         <command name="load">sems-git</command>
         <command name="load">sems-python/2.7.9</command>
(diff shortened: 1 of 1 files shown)


jgfouca:
@jedwards4b I assigned you to review even though I plan on self integrating. I wanted you to see the changes I had to make to conform to the new XSD standard for this file. My initial reaction was that this felt like a regression in the usability of this file. As you can see, particularly for the environment_variable section for titan, the new format is much more cumbersome than the old. What the reasoning behind disallowing attributes for individual modules/env_vars?

PR #1480, opened 2017-05-04
Revert pio1 cmake path to point into pio2

agsalin:
Add back in the pointing of pio1 into pio2/cmake, for now.

The CMakelists.txt file for pio1/pio used to point into pio2/cmake for functions like FindNetCDF.cmake.
commit "Dont point to pio2 cmake modules by default"
PR#1352

This was changed in pio1/pio to point to "${CMAKE_CURRENT_SOURCE_DIR}/cmake" which doesn't exist.

Without a FindNetCDF, the ACMCE build of the HOMME test to fail. 

Changing this back to HOMME build works again until a better fix is found or we switch to pio2.

Test suite: Discovered in ACME -- just typed it in here without running tests.
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes: ACME integration of CIME master as of May 2


jedwards4b:
Andy,
Please use the template for the initial PR message and remove/replace text in [].

agsalin:
A fix that sets USER_CMAKE_MODULE_PATH sounds preferable. Hopefully @jayeshkrishna  can figure out where to do this. This is the ACME test to verify: HOMME_P24.f19_g16_rx1.A

jayeshkrishna:
The right fix for HOMME tests to work is in ACME-Climate/ACME#1480 . Specifically commit - ACME-Climate/ACME@443aa5e

agsalin:
OK, that sounds good. We're having trouble getting the sequence correct and finding the same bug twice. 

One thing that still looks wrong to me is pointing to the pio1/pio/cmake dir that doesn't exist, but that doesn't effect me.

PR #1482, opened 2017-05-04
Make alternative implementation for ERS

### LINKED ISSUE #1244: New option for ERS tests?
Right now, if you run an ERS_Ld5 test then CIME does the following:

1. Run for 5 days, writing restart file after day 3
1. Go back to the day 3 restart file, run for 2 days (days 4 & 5)
1. Compare the output from step 1 and step 2

I'd like an option to run the test in the following manner:

1. Run for 5 days without writing a restart file
1. Run for 3 days (days 1 - 3) and write a restart file at the end of the run
1. Continue the run for 2 days (days 4 & 5)
1. Compare the output from step 1 and step 3

It is extra work (initializing the model three times instead of twice and running an additional three days), but POP has a "feature" where writing a restart history file changes the eventual history file<sup>1</sup> that is written... so if we have a test that computes 5-day averages for tavg output then the current version of the ERS test will `PASS` and the proposed option will `FAIL`. Having a test that fails in this situation would be handy for when we fix this feature.

<sup>1</sup> Two notes... the change in the history file is round-off level, so it isn't terrible that the test currently passes. Also, the change is due to the fact that the POP restart history file is a normalized version of the accumulation buffer so writing a restart file normalizes the buffer, writes it, and then denormalizes it (hence the pass -- in the current test step 1 denormalizes the buffer after _writing_ the restart and step 2 denormalizes it in an identical fashion after _reading_ the restart). The proposed fix to POP is to write the denormalized buffer into the restart history file.

jgfouca:
Creatively named ERS2. First phase does no restarts. Second phase does two runs,
the first run makes a restart file and stops, second run resumes and then compares
to the first phase.

Test suite: ERS2 test, code checker
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1244 

User interface changes?: new ERS2 test type

Code review: @jedwards4b @mnlevy1981

### DIFF
--- a/config/config_tests.xml
+++ b/config/config_tests.xml
@@ -49,6 +49,13 @@ ERS    exact restart from startup (default 6 days + 5 days)
        if $IOP_ON is set then suffix is rest_iop
        compare component history files ".base" and ".rest" at day 11
 
+ERS2    exact restart from startup (default 6 days + 5 days)
+       do an 11 day initial test without making restarts     (suffix: base)
+       if $IOP_ON is set then suffix is base_iop
+       do an 11 day restart test stopping at day 6 with a restart, then resuming from restart at day 6 (suffix: rest)
+       if $IOP_ON is set then suffix is rest_iop
+       compare component history files ".base" and ".rest" at day 11
+
 ERP    pes counts hybrid (open-MP/MPI) restart bfb test from startup, default 6 days + 5 days (previousy PER)
        initial pes set up out of the box
        do an 11 day initial test - write a restart at day 6     (suffix base)
@@ -248,6 +255,19 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
   </test>
 
   <test NAME="ERS">
+    <DESC>exact restart from startup, default 6 days + 5 days</DESC>
+    <INFO_DBUG>1</INFO_DBUG>
+    <STOP_OPTION>ndays</STOP_OPTION>
+    <STOP_N>11</STOP_N>
+    <REST_N>$STOP_N / 2 + 1</REST_N>
+    <REST_OPTION>$STOP_OPTION</REST_OPTION>
+    <HIST_N>$STOP_N</HIST_N>
+    <HIST_OPTION>$STOP_OPTION</HIST_OPTION>
+    <DOUT_S>FALSE</DOUT_S>
+    <CONTINUE_RUN>FALSE</CONTINUE_RUN>
+  </test>
+
+  <test NAME="ERS2">
     <DESC>exact restart from startup, default 6 days + 5 days</DESC>
     <INFO_DBUG>1</INFO_DBUG>
     <STOP_OPTION>ndays</STOP_OPTION>
--- a/scripts/lib/CIME/SystemTests/ers2.py
+++ b/scripts/lib/CIME/SystemTests/ers2.py
@@ -1,0 +1,60 @@
+"""
+CIME restart test 2  This class inherits from SystemTestsCommon
+"""
+from CIME.XML.standard_module_setup import *
+from CIME.SystemTests.system_tests_common import SystemTestsCommon
+
+logger = logging.getLogger(__name__)
+
+class ERS2(SystemTestsCommon):
+
+    def __init__(self, case):
+        """
+        initialize an object interface to the ERS2 system test
+        """
+        SystemTestsCommon.__init__(self, case)
+
+    def _ers2_first_phase(self):
+        stop_n      = self._case.get_value("STOP_N")
+        stop_option = self._case.get_value("STOP_OPTION")
+        rest_n      = self._case.get_value("REST_N")
+
+        # Don't need restarts for first run
+        self._case.set_value("REST_OPTION","never")
+
+        expect(stop_n > 0, "Bad STOP_N: %d" % stop_n)
+        expect(stop_n > 2, "ERROR: stop_n value %d too short"%stop_n)
+
+        logger.info("doing an %s %s initial test with restart file at %s %s"
+                    %(str(stop_n), stop_option, str(rest_n), stop_option))
+        self.run_indv()
+
+    def _ers2_second_phase(self):
+        stop_n      = self._case.get_value("STOP_N")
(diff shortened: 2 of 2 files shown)


jgfouca:
I admit the ERS2 name leaves something to be desired. Also, I did not add an ERS2 test to cime_developer.

jgfouca:
@jedwards4b  What would the T stand for?

@mnlevy1981 how do you want to use this?

jedwards4b:
The next letter after S - I'm not sure any of these three letter designations are very meaningful.  ERT isn't any better than ERS2 I suppose.

jedwards4b:
ERT = E?   Restart Test
FRT = Full Restart Test

mnlevy1981:
@jedwards4b I thought the E was for "exact"? ERS = Exact ReStart? I'll admit to being 12 and really wanting a FRT test but FRS = Full ReStart might be more in line with the current naming convention? ERS2 makes just as much sense to me as the others, though; if Jim F hadn't said "I admit the ERS2 name leaves something to be desired" then I wouldn't even be thinking about the name.

@jgfouca I'll play with this new test tomorrow morning and get back to you - I think we have a test case that passes the ERS test but is expected to fail the [currently named] ERS2 test, but I need to double check (and also create a CESM sandbox with your CIME branch). I think my ideal use case is to add an ERS2 test for the test case that fails to the POP testlist as a not-so-subtle reminder that maybe we should fix this POP issue.

jedwards4b:
oh yea I meant FRS but I guess my reptilian brain took over

rljacob:
We already have an ERT test.

mnlevy1981:
Something doesn't seem quite right... here are the time stamps from an ERS_Ld5 run (I added a space between files to distinguish the different runs):

```
$ zgrep tStamp cpl.log.*
cpl.log.170505-104317.gz: tStamp_write: model date =    10102       0 wall clock = 2017-05-05 10:44:40 avg dt =    50.63 dt =    50.63
cpl.log.170505-104317.gz: tStamp_write: model date =    10103       0 wall clock = 2017-05-05 10:45:32 avg dt =    51.46 dt =    52.29
cpl.log.170505-104317.gz: tStamp_write: model date =    10104       0 wall clock = 2017-05-05 10:46:52 avg dt =    60.84 dt =    79.60
cpl.log.170505-104317.gz: tStamp_write: model date =    10105       0 wall clock = 2017-05-05 10:47:44 avg dt =    58.75 dt =    52.46
cpl.log.170505-104317.gz: tStamp_write: model date =    10106       0 wall clock = 2017-05-05 10:48:57 avg dt =    61.53 dt =    72.68

cpl.log.170505-104908.gz: tStamp_write: model date =    10105       0 wall clock = 2017-05-05 10:50:56 avg dt =    52.51 dt =    52.51
cpl.log.170505-104908.gz: tStamp_write: model date =    10106       0 wall clock = 2017-05-05 10:52:09 avg dt =    63.01 dt =    73.52
```

Note that this run outputs at the end of days 1-5, and there's a spike in the time at the end of day 3 when the restart is written. The next run reads the restart and runs days 4 & 5.

However, this is the output from and ERS2_Ld5:

```
$ zgrep tStamp cpl.log.*
cpl.log.170505-104257.gz: tStamp_write: model date =    10102       0 wall clock = 2017-05-05 10:44:21 avg dt =    50.74 dt =    50.74
cpl.log.170505-104257.gz: tStamp_write: model date =    10103       0 wall clock = 2017-05-05 10:45:13 avg dt =    51.51 dt =    52.29
cpl.log.170505-104257.gz: tStamp_write: model date =    10104       0 wall clock = 2017-05-05 10:46:05 avg dt =    51.76 dt =    52.27
cpl.log.170505-104257.gz: tStamp_write: model date =    10105       0 wall clock = 2017-05-05 10:46:58 avg dt =    51.88 dt =    52.21
cpl.log.170505-104257.gz: tStamp_write: model date =    10106       0 wall clock = 2017-05-05 10:48:11 avg dt =    56.21 dt =    73.55

cpl.log.170505-104822.gz: tStamp_write: model date =    10102       0 wall clock = 2017-05-05 10:49:45 avg dt =    50.60 dt =    50.60
cpl.log.170505-104822.gz: tStamp_write: model date =    10103       0 wall clock = 2017-05-05 10:51:05 avg dt =    65.54 dt =    80.48
cpl.log.170505-104822.gz: tStamp_write: model date =    10104       0 wall clock = 2017-05-05 10:51:59 avg dt =    61.46 dt =    53.31

cpl.log.170505-105214.gz: tStamp_write: model date =    10104       0 wall clock = 2017-05-05 10:54:03 avg dt =    52.60 dt =    52.60
cpl.log.170505-105214.gz: tStamp_write: model date =    10105       0 wall clock = 2017-05-05 10:54:58 avg dt =    53.78 dt =    54.97
```

Here the first run gets all the way through 5 days with no restarts written; the second run goes through day 3, but from the timing it appears as though the restart is written after day 2. Sure enough, the second run starts with day 3 instead of day 4, and doesn't run long enough to finish day 5.

I haven't dug through the python code to figure out what's going on, but it seems like `REST_N` is being set incorrectly in the ERS2.

billsacks:
I thought ERS = Exact Restart from Startup

mnlevy1981:
I think the issue is that

```
<entry id="REST_N" value="$STOP_N / 2 + 1">
```

In the first run, `STOP_N = 5` and `REST_N = 3`
In the second run, `STOP_N = 3`, so `REST_N = 2` and that's where we lose a day.

mnlevy1981:
I went looking in the log for this, and found an indication of the problem as well as superfluous comment. Here's the beginning of `ERS2_Ld5.T62_g16.GECO.yellowstone_intel.pop-ecosys_5day_tavg.20170505_103036_scqrx9.test.104887`

```
doing an 5 ndays initial test with restart file at 3 ndays
doing an 5 ndays startup test, no restarts written
```

(the "doing an 5 ndays initial test with restart file at 3 ndays" is incorrect, that run is not done at this point)

and the indication of a problem a little lower down:

```
doing first part 3 ndays restart test
doing an 3 ndays startup test, with restarts every 2 ndays
```

Issue #893, opened 2016-12-07
Output build time per component


rljacob:
In CIME2, there was a timestamp output to the screen when each component build started.  Example from an F-case:
```
      Wed Dec  7 14:06:40 2016 /lcrc/project/ACME/jacob/acme_scratch/FC5ne30c2gnuthrdblues/bld/lnd.bldlog.161207-140501
      Wed Dec  7 14:06:58 2016 /lcrc/project/ACME/jacob/acme_scratch/FC5ne30c2gnuthrdblues/bld/ice.bldlog.161207-140501
      Wed Dec  7 14:07:18 2016 /lcrc/project/ACME/jacob/acme_scratch/FC5ne30c2gnuthrdblues/bld/ocn.bldlog.161207-140501
      Wed Dec  7 14:07:20 2016 /lcrc/project/ACME/jacob/acme_scratch/FC5ne30c2gnuthrdblues/bld/glc.bldlog.161207-140501
      Wed Dec  7 14:07:21 2016 /lcrc/project/ACME/jacob/acme_scratch/FC5ne30c2gnuthrdblues/bld/wav.bldlog.161207-140501
      Wed Dec  7 14:07:22 2016 /lcrc/project/ACME/jacob/acme_scratch/FC5ne30c2gnuthrdblues/bld/rof.bldlog.161207-140501
      Wed Dec  7 14:07:28 2016 /lcrc/project/ACME/jacob/acme_scratch/FC5ne30c2gnuthrdblues/bld/cesm.bldlog.161207-140501

```

Those time stamps don't appear when building with cime5.  They are useful for finding which component is responsible for an increase in total build time.

Would like to at least get the build timestamps back in output.  Ideally, the build start and end times would also be logged in CaseStatus for each component.

jedwards4b:
Build start and end times are written to the CaseStatus file.  Since component builds now occur in parallel it may be a little trickier to get individual timing output.

rljacob:
Isn't that the full build start and end time written to CaseStatus (just one each)?

jedwards4b:
correct

rljacob:
I agree the parallelism means the starts and end timestamps could be out-of-order.  In that case, we would need to internally calculate the total build time and output that.

PR #1487, opened 2017-05-07
fix ccsm co2 ppmv for #863

### LINKED ISSUE #863: CCSM_CO2_PPMV logic issues
From CESM_DEVELOPMENT issue 121

There are some problems with CCSM_CO2_PPMV as it stands now:

It apparently doesn't do anything for many CLM/CAM compsets, or for any compsets without CLM. For transient runs, the CO2 isn't constant, regardless of configuration. For runs using BGC BPRP/BDRD, WACCM, or some CAM-CHEM compsets, the model chemistry is producing non-constant CO2 concentrations, and there is also typically some time/space varying boundary condition for concentrations or emissions at the surface. It is potentially confusing that we have this switch that looks very important, but often does nothing.

Note that some compsets actually set CCSM_CO2_PPMV to some tiny value to try to get the model to crash or do something crazy if CLM tries to actually use this value. However, besides being just one more hack that can confuse new users, this has some mixed results.

For instance, my recollection is that CESM 1.1.0 had exactly the kind of bug that the unrealistic CO2 value was intended to prevent, accidentally using -co2_type constant for a CAM compset. Indeed, the low value of CO2 was noticed by Mike Mills. However, this didn't happen until shortly after the release, which forced us to do an "emergency" CESM 1.1.1 release immediately afterward. If we still need to check for this error (which seems much less likely to happen in the current scripts), we should use a guard that will actually cause an automated test to fail, rather than requiring a human to notice that a bug is producing unrealistic output.

It's not actually necessary to coordinate this value between CAM and CLM. If CAM is on, CLM never uses -co2_type constant, and therefore CCSM_CO2_PPMV is ignored by CLM. If CAM is off, there is no possibility of having an inconsistent CO2 value between components. We still need CLM_CO2_TYPE to be set in the CIME scripts, because it has a different value based on whether or not CAM is on, but the actual CO2 concentration can be defined by the CAM/CLM use cases, without setting it at the CIME level.

In fact, CAM already sets the CO2 concentration in some use cases, but this is overridden by CIME, which is another potential source of confusion. Again, we have settings that look important, but are not actually used, except in CAM standalone builds. See, e.g., the discussion in #108.

If we remove CCSM_CO2_PPMV and just set the value in CAM/CLM use cases, the only real issue is that some CLM climatological runs could potentially drift from their CAM counterparts (e.g. I1850 runs could accidentally use a different value than F1850 for CO2). This does not seem too high a risk (especially since these values are changed so rarely), but it is present.

Note also that Erik proposed that DATM set this value for I compsets, rather than a CLM use case. It's not clear to me that either approach is superior. This value is used by CLM itself, not DATM, however, if a CLM use case is used, it may not be obvious to CLM users that the value in the use case is ignored when CAM is on. So it could be good to have the atmosphere component always be the one providing a value (even for data models).

There are a some little silly issues with the implementation that we wouldn't have to address if this variable just goes away. For instance, the outdated CCSM prefix, and the fact that you can't use scientific notation in the value (at least for CAM).

CC @ekluzek @brian-eaton @billsacks

I am marking this as a backwards compatibility issue, because the current {cam,clm}.buildnml scripts won't work properly without CCSM_CO2_PPMV defined, and because we don't want the components to start ignoring this variable until it's time to actually remove it. (Otherwise, some scientist using a beta tag might try to rely on this variable, unaware that it no longer does anything.)

mvertens:
Simplify settings in CCSM_CO2_PPMV in config_components_cesm.xml

Note that this PR ONLY effects CESM compsets.
I have decided to implement a compromise approach to remove the need to modify CLM and POP namelist generation scripts. This PR removes the CO2_TYPE_COSTANT_VALUE xml variable and simplifying the CCSM_CO2_PPVM variable in CIME to be the following:

  <entry id="CCSM_CO2_PPMV">
    <type>real</type>
    <valid_values></valid_values>
    <default_value>284.7</default_value>
    <values>
      <value compset="^2000">367.0</value>
      <value compset="^4804">367.0</value>
    </values>
    <group>run_co2</group>
    <file>env_run.xml</file>
    <desc>
      Mechanism for setting the CO2 value in ppmv for 
      CLM if CLM_CO2_TYPE is constant or for 
      POP if OCN_CO2_TYPE is constant.
    </desc>
  </entry>

This accomplishes the main goal of #893 but also provides backwards compatibility and does not require either a new CLM or POP tag and accomplish much of the cleanup that we wanted. I only drawback is keeping the old name of CCSM_CO2_PPMV - but the documentation for it now reflects what it is meant for. And this will not required any new CLM or POP tags. 

Test baseline: The following namelists comparisons were also validated:
    SSP_Ld10.f19_g16.I1850CLM45BGC.yellowstone_pgi.clm-rtmColdSSP
    ERP_D.f10_f10.IHISTCLM50BGC.yellowstone_pgi.clm-decStart
    SMS_Lm1_D.f10_f10.ICLM45BGCCROP.yellowstone_intel.clm-snowlayers_3_monthly
    ERS_Ld5.T62_g16.GECO.yellowstone_intel.pop-ecosys_5day_tavg
Test namelist changes: None
Test status: bit for bit

Fixes #863 
User interface changes?: None
Code review:

### DIFF
--- a/src/drivers/mct/cime_config/config_component_cesm.xml
+++ b/src/drivers/mct/cime_config/config_component_cesm.xml
@@ -384,58 +384,6 @@
   <entry id="CCSM_CO2_PPMV">
     <type>real</type>
     <valid_values></valid_values>
-    <default_value>379.000</default_value>
-    <values>
-      <value compset="^1850"                     >284.7</value>
-      <value compset="^HIST.+BGC%BPRP"           >284.7</value>
-      <value compset="^HIST_CAM4%FCHM"           >0.000001</value>
-      <value compset="^2000"                     >367.0</value>
-      <value compset="^1850.+DATM%NYF"           >379.000</value>
-      <value compset="^1850.+DATM%NYF.*_POP2%ECO">284.7</value>
-      <value compset="^2000.+DATM%NYF"           >379.000</value>
-      <value compset="^2000.+DATM%IAF"           >379.000</value>
-      <value compset="^HIST.+DATM"               >367.0</value>
-      <value compset="^4804.+DATM"               >367.0</value>
-      <value compset="^RCP2.+DATM"               >367.0</value>
-      <value compset="^RCP4.+DATM"               >367.0</value>
-      <value compset="^RCP6.+DATM"               >367.0</value>
-      <value compset="^RCP8.+DATM"               >367.0</value>
-      <value compset="^2000.+XATM"               >379.000</value>
-      <value compset="^HIST.+CAM"                >0.000001</value>
-      <value compset="^5505.+CAM"                >0.000001</value>
-      <value compset="^RCP2.+CAM"                >0.000001</value>
-      <value compset="^RCP4.+CAM"                >0.000001</value>
-      <value compset="^RCP6.+CAM"                >0.000001</value>
-      <value compset="^RCP8.+CAM"                >0.000001</value>
-      <value compset="^2013.+CAM"                >0.000001</value>
-      <value compset="^SDYN.+CAM"                >0.000001</value>
-      <value compset="^AMIP.+CAM"                >0.000001</value>
-      <value compset="^AR95"                     >368.9</value>
-      <value compset="^AR97"                     >368.9</value>
-      <value compset="^2003"                     >367.0</value>
-      <!-- Override values based on CAM (WACCM) chemistry -->
-      <value compset="CAM[456]0%W"               >0.000001</value>
-      <value compset="CAM[456]0%SVBS"            >0.000001</value>
-      <value compset="CAM[456]0%TMOZ"            >0.000001</value>
-      <!-- Mariana's fix to allow isotopes to get the proper CO2 value -->
-      <value compset="CAM..%WISOall"             >284.7</value>
-      <value compset="^2000.+CAM[45]%WCSC"       >368.9</value>
-      <value compset="^2000.+CAM[45]%MOZM"       >0.000001</value>
-      <value compset="^2000.+CAM[45]%MMOS"       >0.000001</value>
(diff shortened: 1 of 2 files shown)


billsacks:
@mvertens - this co2 specification isn't something I'm familiar with. Can this be reviewed by someone who knows more about this, like Erik?

mvertens:
@sacks - I've assigned this to Mike. This change is completely backwards
compatible - and Keith Lindsay is most familiar with the request - so Mike
can run this past him as well.

src/drivers/mct/cime_config/config_component_cesm.xml
```diff
@@ -384,68 +384,16 @@
...
-    <desc>This set the namelist values of CO2 ppmv for CAM and CLM. This variables is
-      introduced to coordinate this value among multiple components.</desc>
-  </entry>
-
-  <entry id="CO2_TYPE_COSTANT_VALUE">
-    <type>real</type>
-    <valid_values></valid_values>
     <default_value>284.7</default_value>
     <values>
       <value compset="^2000">367.0</value>
```
klindsay28: Can we also have the following here?
      <value compset="DATM.*POP2%ECO">284.7</value>
I think this will produce 1850 values for any compset using datm and ocean BGC,
whether or not the compset name starts with 2000.

klindsay28: sorry, github stripped the formatting of my proposed addition
Here's what I meant to write:
`      <value compset="DATM.*POP2%ECO">284.7</value>`

mvertens:
Thank you for the clarification. I am happy to add this in.

PR #1490, opened 2017-05-08
Fix pylint errors

jgfouca:
Also, add logging-format-interpolation to list of pylint warnings we don't care about. That way, we can use python3 string formatting without getting warnings.

Test suite: code_checker
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: None

Code review: @jedwards


jgfouca:
@jedwards4b This PR will get us back to a passing state in the nightlies, the question of string format can be decided later. The ignored waring is useless in either case.

PR #1491, opened 2017-05-08
Fix bug where file drv_flds_in wasn't being closed after being opened.

### LINKED ISSUE #1489: drv_flds_in being open multiple times without being close.
drv_flds_in is being opened in seq_drydep_mod.F90, but not closed.  Then it's opened again in shr_ndep_mod.F90.  This bug was only being caught by gnu and pgi, intel and nag where fine with this bug.

fischer-ncar:
A file close() statement that goes along with an file open() statement, was inside an if statement.  
This was causing the file to be open multi times, without it being closed.  This was throwing an error
for pgi and gnu.  Intel and nag didn't catch this.

Test suite: ./create_test cime_developer
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #1489

User interface changes?:

Code review:

### DIFF
--- a/src/drivers/mct/shr/seq_drydep_mod.F90
+++ b/src/drivers/mct/shr/seq_drydep_mod.F90
@@ -566,10 +566,10 @@ CONTAINS
                    call shr_sys_abort( subName//'ERROR: encountered end-of-file on namelist read' )
                 endif
              end do
-             close( unitn )
           else
              write(s_logunit,*) 'seq_drydep_read:  no drydep_inparm namelist found in ',NLFilename
           endif
+          close( unitn )
           call shr_file_freeUnit( unitn )
        end if
     end if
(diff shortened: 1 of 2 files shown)


mvertens:
If this is in the other routines that deal with drv_flds_in - that should
be fixed as well. I'm perplexed as to why this was not caught if the code
is identical. Could you please verify that this is indeed the case.

fischer-ncar:
There was only two routines with the close in the wrong place, seq_drydep_readnl and shr_ndep_readnl.

The reason this was never caught before was previously the last time drv_flds_in was accessed was in  seq_drydep_readnl.  Then drv_flds_in was never accessed again.  Then ndep was added after drydep.  At this point, drv_flds_in was still open when ndep tries to open it.  Thus causing the error for the first time.

mvertens:
Thanks for pointing this out. And I based my implementation on
seq_drydep_readnl. We need to fix both - could you please fix this in both
places and make sure the failed tests now work?

fischer-ncar:
It's been fixed in both places, and the tests have passed.  PR has been made.

PR #1494, opened 2017-05-08
Pesfile usage

jedwards4b:
Previously the user was only allowed to specify a file in the config_pes.xml format in the --pesfile argument to create_newcase.   This change allows the user to specify a file in either the config_pes.xml or the env_mach_pes.xml format.   Also add a test for this feature

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: --pesfile argument in create_newcase 

Code review:

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -126,8 +126,6 @@ class Case(object):
         for variable substitution using the {{ var }} syntax
         """
         env_mach_pes  = self.get_env("mach_pes")
-        print "HERE %s"%env_mach_pes.get_value("TOTALPES")
-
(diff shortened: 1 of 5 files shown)


scripts/lib/CIME/case.py
```diff
@@ -238,6 +238,7 @@ def flush(self, flushall=False):
         for env_file in self._env_files_that_need_rewrite:
             env_file.write()
         self._env_files_that_need_rewrite = set()
+        mach_pes = self.get_env("mach_pes")
```
jgfouca: I don't think you need this line, mach_pes is not used for anything.

jedwards4b: Thanks - removed.

PR #1502, opened 2017-05-09
Reset test case

### LINKED ISSUE #901: ERS test failure leaves namelist in unexpected state
I had the following happen with an ERS test:

1. Test failed in the first leg of the restart
1. I made necessary SourceMods and rebuilt / resubmitted the test
1. TestStatus reported a failure in the namelist comparison:
```
2016-12-09 08:24:57: Differences in namelist 'seq_timemgr_inparm':
  BASE: history_n = -999
  COMP: history_n = 11
  BASE: history_option = 'never'
  COMP: history_option = 'ndays'
  BASE: restart_n = 11
  COMP: restart_n = 6
```

`case.build` actually threw a warning about this failure
```
Found difference in HIST_N: case: 11 original value -999
 Use option --force to run the test with this value or --reset to reset to original
```

But (as @jedwards4b pointed out in Slack) that warning message is out of date; the `--force` and `--reset` options are not available in the build script (or submit, for that matter).

I think the main cause of this failure is that `case.test` does not abort if the first half of a restart test fails so it makes the necessary namelist changes to launch the restart run as well; if that's an accurate assessment of the issue, throwing an exception if the first run fails and aborting before setting up the second run would be a sufficient fix.

For cases where the ERS test aborts in the second run, though (i.e. if a bug caused a variable to be omitted from a restart file but a component tried to read it anyway), there needs to be a way to reset the namelist variables... ideally `case.test` would call this script in case of run failures but maybe make it available to users to allow re-running of these tests. Going off the original warning I got from `case.build`, I think it would be nice to be able to run

```
$ ./case.test --reset
```
to reset the XML variables to their expected values at the beginning of a test. (This approach would make it easy for `case.test` to call this routine when a run fails as well.)

jedwards4b:
adds a case.test --reset option to manually reset a test to initial conditions, also adds a teardown method to system_tests_common.py which assures that RUN_WITH_SUBMIT flag is reset to False
whenever a test method exits. 

Test suite: scripts_regression_tests.py ERR.f09_g16.B1850.cheyenne_intel.allactive-defaultio
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #901 

User interface changes?: Add --reset option to case.test 

Code review: jgfouca

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -35,6 +35,9 @@ class SystemTestsCommon(object):
         self._init_environment(caseroot)
         self._init_locked_files(caseroot, expected)
 
+    def __del__(self):
+        self._case.set_value("RUN_WITH_SUBMIT", False)
+
     def _init_environment(self, caseroot):
         """
         Do initializations of environment variables that are needed in __init__
--- a/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -53,14 +53,14 @@ class SystemTestsCommon(object):
         elif os.path.isfile(os.path.join(caseroot, "env_run.xml")):
             lock_file("env_run.xml", caseroot=caseroot, newname="env_run.orig.xml")
 
-    def _resetup_case(self, phase):
+    def _resetup_case(self, phase, reset=False):
         """
         Re-setup this case. This is necessary if user is re-running an already-run
         phase.
         """
         # We never want to re-setup if we're doing the resubmitted run
         phase_status = self._test_status.get_status(phase)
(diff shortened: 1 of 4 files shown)


scripts/lib/CIME/SystemTests/system_tests_common.py
```diff
@@ -35,6 +35,9 @@ def __init__(self, case, expected=None):
         self._init_environment(caseroot)
         self._init_locked_files(caseroot, expected)
 
+    def __del__(self):
+        self._case.set_value("RUN_WITH_SUBMIT", False)
```
jgfouca: Is this necessary? I added some code around line 250 in case_run to handle RUN_WITH_SUBMIT.

jedwards4b: Yes, I have noticed that case.test would attempt to run mpirun.exe when I did not use 
case.submit

Issue #1505, opened 2017-05-10
env_mach_specific.xsd not general enough to handle my Mac's config_machines


billsacks:
@jedwards4b (or someone else) - can you please help with this?

With recent versions of cime, I am no longer able to build the unit tests or set up a case on my Mac. I get this error:

```
ERROR: Command: '/usr/bin/xmllint --noout --schema /Users/sacks/cime/config/xml_schemas/env_mach_specific.xsd /Users/sacks/cime/unit_test_temp.bjkaBn/env_mach_specific.xml' failed with error '/Users/sacks/cime/unit_test_temp.bjkaBn/env_mach_specific.xml:8: element module_system: Schemas validity error : Element 'module_system': Missing child element(s). Expected is ( init_path ).
/Users/sacks/cime/unit_test_temp.bjkaBn/env_mach_specific.xml:9: element mpirun: Schemas validity error : Element 'mpirun': This element is not expected. Expected is ( environment_variables ).
/Users/sacks/cime/unit_test_temp.bjkaBn/env_mach_specific.xml fails to validate'
```

I am using the following config_machines.xml, which worked until recently:

```xml
<?xml version="1.0"?>
<!-- This is an ordered list, not all fields are required, optional fields are noted below. -->
<config_machines version="2.0">
<!-- MACH is the name that you will use in machine options -->
  <machine MACH="roo2">

    <!-- DESC: a text description of the machine, this field is current not used in code-->
    <DESC>roo2</DESC>

	<!-- NODENAME_REGEX: a regular expression used to identify this machine
	  it must work on compute nodes as well as login nodes, use machine option
	  to create_test or create_newcase if this flag is not available -->
    <NODENAME_REGEX>roo2</NODENAME_REGEX>

    <!-- OS: the operating system of this machine. Passed to cppflags for
	 compiled programs as -DVALUE  recognized are LINUX, AIX, Darwin, CNL -->
    <OS>Darwin</OS>

	<!-- COMPILERS: compilers supported on this machine, comma seperated list, first is default -->

    <COMPILERS>gnu</COMPILERS>

	<!-- MPILIBS: mpilibs supported on this machine, comma seperated list,
	     first is default, mpi-serial is assumed and not required in this list-->
    <MPILIBS>mpich</MPILIBS>

    <!-- CIME_OUTPUT_ROOT: Base directory for case output,
	 the case/bld and case/run directories are written below here -->
[...]

PR #1509, opened 2017-05-10
Fix env_mach_specific validation

jedwards4b:
Allow the user the option of skipping the xmllint validation of cime xml files. 

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Addresses #1505 

User interface changes?: Adds --skip-xml-validation flag to all user interface scripts

Code review:

### DIFF
--- a/config/xml_schemas/env_mach_specific.xsd
+++ b/config/xml_schemas/env_mach_specific.xsd
@@ -23,9 +23,9 @@
       <xs:sequence>
         <xs:element ref="header"/>
         <xs:element ref="module_system"/>
-        <xs:element maxOccurs="unbounded" ref="environment_variables"/>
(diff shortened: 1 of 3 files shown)


jgfouca:
@jedwards4b I agree with @billsacks .

jedwards4b:
I agree that the name setup_standard_logging_options is incorrect, but my intent was always that it be setup_standard_options.  This option was added based on my perception of @billsacks level of distress when an xml file didn't validate, but there may be other good reasons for it.   Do you have any suggestions other than the global for how to implement?

jgfouca:
The only other way would be a very-annoying pass-down of a parameter all the way down to GenericXML.

jedwards4b:
okay - only the schema is fixed

rljacob:
Update the description to match the title.

PR #1517, opened 2017-05-11
Set value valid

### LINKED ISSUE #389: Should case.set_value raise an exception if you try to set a variable not in an xml file?
I tried to force an error in a system test (LII.py) by calling self._case.set_value on some non-existent variable ('foo'). To my surprise, no exception was raised.

Looking at the Case.set_value code, it looks like it can work in two different ways: (1) If you specify a variable that exists in some xml file, then it sets that variable. I'm guessing that's the typical operation. But (2) if you specify a variable that is not in some xml file, then it sets this variable in its lookups dictionary.

The problem here is that this means that a typo in the variable to set will not be caught. For example, I ran an ERS test with this difference:

``` diff
--- a/utils/python/CIME/SystemTests/ers.py
+++ b/utils/python/CIME/SystemTests/ers.py
@@ -42,7 +42,7 @@ class ERS(SystemTestsCommon):
         expect(stop_new > 0, "ERROR: stop_n value %d too short %d %d"%(stop_new,stop_n,rest_n))

         self._case.set_value("STOP_N", stop_new)
-        self._case.set_value("CONTINUE_RUN", True)
+        self._case.set_value("CONTINUEEEE_RUN", True)
         self._case.set_value("REST_OPTION","never")
         self._case.flush()
         logger.info("doing an %s %s restart test"
```

The typo went undetected, and the ERS test ran two initial runs rather than doing a continue run.

This makes it more likely that system tests (and other code) won't work as intended.

I'd propose that set_value be split into two different routines:

(1) set_value works only on xml variables. If you give it an item not found in an xml file, it raises an exception

(2) set_lookup_value contains the last block of code currently in set_value:

``` python
            if item in self.lookups.keys() and self.lookups[item] is not None:
                logger.warn("Item %s already in lookups with value %s"%(item,self.lookups[item]))
            else:
                self.lookups[item] = value
```

jedwards4b:
Make sure that case.set_value is setting a valid variable and error out if it is not,
this requires allowing for exceptions during case configure and cleaning up a few places where
non-existant variables were being assigned.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #389

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/grids.py
+++ b/scripts/lib/CIME/XML/grids.py
@@ -70,6 +70,7 @@ class Grids(GenericXML):
 
         # determine domains given component_grids
         domains  = self._get_domains(component_grids, atmlevregex, lndlevregex)
+
         gridinfo.update(domains)
 
         # determine gridmaps given component_grids
@@ -250,8 +251,9 @@ class Grids(GenericXML):
             domain_node = self.get_optional_node(nodename="domain", attributes={"name":grid_name_nonlev})
             if domain_node is not None:
                 comp_name = grid[0].upper()
-                domains[comp_name + "_NX"] = int(self.get_element_text("nx", root=domain_node))
-                domains[comp_name + "_NY"] = int(self.get_element_text("ny", root=domain_node))
+                if not comp_name == "MASK":
+                    domains[comp_name + "_NX"] = int(self.get_element_text("nx", root=domain_node))
+                    domains[comp_name + "_NY"] = int(self.get_element_text("ny", root=domain_node))
                 domains[comp_name + "_GRID"] = grid_name
                 file_name = comp_name + "_DOMAIN_FILE"
                 path_name = comp_name + "_DOMAIN_PATH"
--- a/scripts/lib/CIME/case_st_archive.py
+++ b/scripts/lib/CIME/case_st_archive.py
@@ -212,7 +212,8 @@ def _archive_restarts(case, archive, archive_entry,
     rundir = case.get_value("RUNDIR")
     casename = case.get_value("CASE")
     archive_restdir = join(dout_s_root, 'rest', datename)
-    if not datename_is_last or case.get_value('DOUT_S_SAVE_INTERIM_RESTART_FILES'):
+    if not datename_is_last or case.get_value('DOUT_S_SAVE_INTERIM_RESTART_FILES') \
+       or case.get_value("TEST"):
         if not os.path.exists(archive_restdir):
             os.makedirs(archive_restdir)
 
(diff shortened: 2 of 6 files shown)


scripts/lib/CIME/XML/grids.py
```diff
@@ -250,8 +251,9 @@ def _get_domains_v2(self, component_grids, atmlevregex, lndlevregex):
             domain_node = self.get_optional_node(nodename="domain", attributes={"name":grid_name_nonlev})
             if domain_node is not None:
                 comp_name = grid[0].upper()
-                domains[comp_name + "_NX"] = int(self.get_element_text("nx", root=domain_node))
-                domains[comp_name + "_NY"] = int(self.get_element_text("ny", root=domain_node))
+                if not comp_name == "MASK":
```
billsacks: Is this change related to the overall PR, or is it something separate?

scripts/lib/CIME/case_st_archive.py
```diff
@@ -212,7 +212,8 @@ def _archive_restarts(case, archive, archive_entry,
     rundir = case.get_value("RUNDIR")
     casename = case.get_value("CASE")
     archive_restdir = join(dout_s_root, 'rest', datename)
-    if not datename_is_last or case.get_value('DOUT_S_SAVE_INTERIM_RESTART_FILES'):
+    if not datename_is_last or case.get_value('DOUT_S_SAVE_INTERIM_RESTART_FILES') \
+       or case.get_value("TEST"):
```
billsacks: Is this related to this PR, or a separate fix?

scripts/lib/CIME/case_submit.py
```diff
@@ -42,7 +42,8 @@ def _submit(case, job=None, resubmit=False, no_batch=False, skip_pnl=False, batc
     # flag will stay in effect for the duration of the RESUBMITs
     env_batch = case.get_env("batch")
     if not resubmit:
-        case.set_value("IS_FIRST_RUN", True)
+        if case.get_value("TEST"):
+            case.set_value("IS_FIRST_RUN", True)
```
billsacks: I think a similar change is needed a few lines lower, where IS_FIRST_RUN is set to False

scripts/lib/CIME/case_st_archive.py
jedwards4b: This is a separate fix noticed in testing.

scripts/lib/CIME/XML/grids.py
jedwards4b: This is related to the PR MASK_NX and MASK_NY were being set but are not variables that we ever define.

scripts/lib/CIME/case_submit.py
jedwards4b: Thanks for noticing, I don't think that second case would have been caught by any of our testing.

billsacks:
Okay, looks good now - thanks a lot for this change!

Issue #608, opened 2016-09-29
Want more user-friendly way to retrieve PE settings for a user compset


billsacks:
This is an enhancement that I'd like to see by the cesm2 release, but is not critical before then.

In order to specify a user-compset, you currently need to point to the xml file containing the PE settings you want to use (using the --pesfile argument to create_newcase). This requires expert knowledge of:
1. For a given compset, which component will contain the xml file you need
2. The path to the given xml file

I feel that users should not be required to know either of these things: the scripts should be able to figure this out for you. We have rules for where to find the appropriate config pes settings, based on what compset you're using. It may not be trivial to code up these rules, but I feel that it will be easier to do that than to try to train all CESM2 users on these rules.

This is especially important for CESM2, where we are planning to have very few out-of-the-box compsets, relying much more heavily on user-compsets.

ekluzek:
Agreeing with Bill that this is important. Having to know the specific path for the config_pes.xml is ridiculous and onerous for a new user.

billsacks:
I'm changing this from an enhancement to an issue, because I think it's really an issue that needs to be resolved for cesm2.

jedwards4b:
Please lead me through an example.  How do we determine that the user_compset input is a valid compset and then how do we determine what the principle component of that compset is?

mvertens:
One idea is to have a first pass match list for the principle component - 
- e.g. you have CAM + CLM  + DOCN  then the principle component is CAM
if a match is not found - then the user would be told they have to specify the pes file
thoughts?

billsacks:
@mvertens - yes, I agree with your idea. My conceptual thinking is: What is the process that you or I would currently go through to figure out where to find the pes file for a given user compset? We "just" need to put that process into code. I agree that the first step is to determine the principle component, as you say.

rljacob:
I didn't realize --pesfile had to point to a pre-existing config_pes.xml.   Shouldn't it point to a user-created env_mach_pes.xml?   It doesn't make sense for a user's custom compset to have its pe-layout in the same files as the regular compsets.

billsacks:
@rljacob - In most cases, I think user-defined compsets are just slight one-offs from existing compsets - e.g., using a combination of options for which there is no alias. Since the pe layout is usually more general than that (e.g., all I compsets at a given resolution should use this pe layout), it makes sense to reuse the existing config_pes file.

jedwards4b:
[...]

PR #1523, opened 2017-05-12
remove compset user_mods and add principle component user_mods

jedwards4b:
This PR removes user_mods support from the config_compsets.xml file and adds it to the config_component.xml file.  This allows for regular expression matching which greatly reduces the number of entries required.   The user_mods are only applied if they are defined in the config_component.xml file of the PRINCIPLE component.  

Test suite: scripts_regression_tests.py, Hand testing of 2000_CAM50_SLND_SICE_DOCN%AQP3_SROF_SGLC_SWAV with cam tag  https://svn-ccsm-models.cgd.ucar.edu/cam1/branch_tags/aquap_cesm2_tags/aquap_cems2_n12_cam5_4_118_alpha06h_plus
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -72,6 +72,7 @@ class Case(object):
         self._env_files_that_need_rewrite = set()
         self._read_only_mode = True
         self._force_read_only = read_only
+        self._principle_component = None
 
         self._env_entryid_files = []
         self._env_generic_files = []
@@ -98,7 +99,6 @@ class Case(object):
         self._is_env_loaded = False
         # these are user_mods as defined in the compset
         # Command Line user_mods are handled seperately
-        self._user_mods = None
         self.thread_count = None
         self.total_tasks = None
         self.tasks_per_node = None
@@ -430,7 +430,7 @@ class Case(object):
             # If the file exists, read it and see if there is a match for the compset alias or longname
             if (os.path.isfile(compsets_filename)):
                 compsets = Compsets(compsets_filename)
-                match, compset_alias, science_support, self._user_mods = compsets.get_compset_match(name=compset_name)
+                match, compset_alias, science_support = compsets.get_compset_match(name=compset_name)
                 if match is not None:
                     if pesfile is None:
                         self._pesfile = files.get_value("PES_SPEC_FILE"     , {"component":component})
@@ -443,13 +443,11 @@ class Case(object):
                     user_mods_dir     = files.get_value("USER_MODS_DIR"     , {"component":component}, resolved=False)
                     self.set_lookup_value("COMPSETS_SPEC_FILE" ,
                                    files.get_value("COMPSETS_SPEC_FILE", {"component":component}, resolved=False))
+                    self._principle_component = component.upper()
                     self.set_lookup_value("TESTS_SPEC_FILE"    , tests_filename)
                     self.set_lookup_value("TESTS_MODS_DIR"     , tests_mods_dir)
                     self.set_lookup_value("USER_MODS_DIR"      , user_mods_dir)
-                    compset_info = "Compset longname is %s"%(match)
-                    if self._user_mods is not None:
-                        compset_info += " with user_mods directory %s"%(self._user_mods)
-                    logger.info(compset_info)
+                    logger.info("Compset longname is %s"%(match))
                     logger.info("Compset specification file is %s" %(compsets_filename))
                     logger.info("Pes     specification file is %s" %(pesfile))
                     return compset_alias, science_support
@@ -556,6 +554,7 @@ class Case(object):
             result = self.set_value(key,value)
             if result is not None:
                 del self.lookups[key]
+
 
     def _setup_mach_pes(self, pecount, ninst, machine_name, mpilib):
         #--------------------------------------------
@@ -853,7 +852,8 @@ class Case(object):
         for name, value in matches:
             if len(value) > 0:
                 logger.debug("Compset specific settings: name is %s and value is %s"%(name,value))
-                self.set_value(name, value)
+                self.set_lookup_value(name, value)
+
 
     def set_initial_test_values(self):
         testobj = self.get_env("test")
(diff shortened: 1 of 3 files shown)


gold2718:
Could someone add a definition of the PRINCIPLE component to the documentation? True confession; I have not heard this term before today.

jedwards4b:
The PRINCIPLE component is the component which is determined to be the primary focus of an experiment.   For F or Q compsets this is the atmosphere, For C and G its ocn, for B compsets the principle is allactive. The principle component determines which cime_config directory will be used to locate control files such as config_pio.xml and config_pes.xml

mvertens:
Have you tested that a user-mods on the create_newcase command line actually takes precedence over the user-mods set by CAM_USER_MODS? I do not see this in the hand testing and I think its an important test to run.

jedwards4b:
```
./create_newcase --case foo --compset 2000_CAM40_SLND_SICE_DOCN%AQP3_SROF_SGLC_SWAV --res f09_f09_mg17 --run-unsupported --user-mods /glade/p/work/jedwards/sandboxes/cesm2_0_alpha/components/cam/cime_config/testdefs/testmods_dirs/cam/outfrq9s
Compset longname is 2000_CAM40_SLND_SICE_DOCN%AQP3_SROF_SGLC_SWAV
Compset specification file is /glade/p/work/jedwards/sandboxes/cesm2_0_alpha/cime/../components/cam/cime_config/config_compsets.xml
Pes     specification file is None
Machine is cheyenne
Pes setting: grid match    is a%0.9x1.25 
Pes setting: grid          is a%0.9x1.25_l%null_oi%0.9x1.25_r%null_g%null_w%null_m%gx1v7 
Pes setting: compset       is 2000_CAM40_SLND_SICE_DOCN%AQP3_SROF_SGLC_SWAV 
Pes setting: tasks       is {'NTASKS_ATM': -4, 'NTASKS_ICE': -4, 'NTASKS_CPL': -4, 'NTASKS_LND': -4, 'NTASKS_WAV': -4, 'NTASKS_ROF': -4, 'NTASKS_OCN': -4, 'NTASKS_GLC': -4} 
Pes other settings: {}
 Compset is: 2000_CAM40_SLND_SICE_DOCN%AQP3_SROF_SGLC_SWAV 
 Grid is: a%0.9x1.25_l%null_oi%0.9x1.25_r%null_g%null_w%null_m%gx1v7 
 Components in compset are: ['cam', 'slnd', 'sice', 'docn', 'srof', 'sglc', 'swav', 'sesp', 'cpl', 'dart'] 
Using project from env PROJECT: P93300606
cesm model version found: cesm2_0_alpha06j
 Creating Case directory /glade/p/work/jedwards/sandboxes/cesm2_0_alpha/cime/scripts/foo
This compset includes user_mods /glade/p/work/jedwards/sandboxes/cesm2_0_alpha/components/cam/cime_config/usermods_dirs/aquap_som
Adding user mods directory /glade/p/work/jedwards/sandboxes/cesm2_0_alpha/components/cam/cime_config/usermods_dirs/aquap_som
Adding user mods directory /glade/p/work/jedwards/sandboxes/cesm2_0_alpha/components/cam/cime_config/testdefs/testmods_dirs/cam/outfrq9s
```

scripts/lib/CIME/case.py
```diff
@@ -443,13 +443,11 @@ def _set_compset_and_pesfile(self, compset_name, files, user_compset=False, pesf
                     user_mods_dir     = files.get_value("USER_MODS_DIR"     , {"component":component}, resolved=False)
```
billsacks: Should this setting of user_mods_dir be removed now? I can't tell.

scripts/lib/CIME/case.py
```diff
@@ -443,13 +443,11 @@ def _set_compset_and_pesfile(self, compset_name, files, user_compset=False, pesf
...
                     self.set_lookup_value("COMPSETS_SPEC_FILE" ,
                                    files.get_value("COMPSETS_SPEC_FILE", {"component":component}, resolved=False))
+                    self._principle_component = component.upper()
```
billsacks: This determination of principle component will not work for a user-defined compset. I feel this is very important given that we encourage people to use compset long names rather than aliases. And, with @mvertens 's idea in #608 we'll need a way to determine the principle component for a user-defined compset for that purpose, too.

If it's critical to get this PR in without support for user-defined compsets, I'm okay with that - with a separate issue being opened that this needs to be fixed.

scripts/lib/CIME/case.py
```diff
@@ -1006,14 +1008,11 @@ def apply_user_mods(self, user_mods_dir=None):
         or they can be in the compset definition, or both.
         """
 
-        if self._user_mods is None:
-            compset_user_mods_resolved = None
-        else:
-            compset_user_mods_resolved = self.get_resolved_value(self._user_mods)
+        component_user_mods = self.get_value("%s_USER_MODS"%(self._principle_component))
```
billsacks: The previous code had get_resolved_value. Is that important? (I really don't know myself).

scripts/lib/CIME/case.py
mvertens: I would like this PR to go in without the support for user-defined compsets - which I agree needs to go into a new PR that could be integrated with #608.

scripts/lib/CIME/case.py
mvertens: I am confused about this as well.

scripts/lib/CIME/case.py
jedwards4b: Nope, still in use

scripts/lib/CIME/case.py
jedwards4b: I didn't understand why it had that and don't think that it needed it.,  the default returned from get_value is resolved=True

scripts/lib/CIME/case.py
billsacks: Okay, that's fine. Pre-emptively opened #1525

billsacks:
Okay, good to go as far as I'm concerned. Thanks!

PR #1531, opened 2017-05-14
add ability to remove tabs in shr_string_lalign

### LINKED ISSUE #862: Tab characters should be removed from strings in streams files
<a href="https://github.com/quantheory"></img></a> **Issue by [quantheory](https://github.com/quantheory)**
_Wednesday Jul 29, 2015 at 22:27 GMT_
_Originally opened as https://github.com/CESM-Development/cime/issues/72_

----

From a Bugzilla issue from Bill Sacks:

A user ran into a problem when editing a streams file (`dlnd.streams.txt.sno.cplhist`): there were some tab characters in the file both before and after some strings. It looks like these tab characters are not stripped from the strings (e.g., using `shr_string_leftAlign`). This led to an error that looked like this in the lnd log file:

```
(shr_sys_abort) ERROR: (shr_stream_getCalendar)  ERROR: nf90_open file      /glade/scratch/gailg        /           IG_spinup_smb4.cpl.hs2x.0067.4avgmax.nc
```

Since tabbing is a natural thing to do when editing text files, it seems it would be good to extend the `shr_string` utilities to strip tabs as well as spaces.

mvertens:
Removal of tabs from strings in stream files

 also manual verification that if tabs were introduced in a streams text file 
    (e.g. datm.streams.txt.CORE2_NYF.GISS) 
 then the resulting data model behaved correctly
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #862 
User interface changes?: None 
Code review:

### DIFF
--- a/src/share/util/shr_string_mod.F90
+++ b/src/share/util/shr_string_mod.F90
@@ -412,7 +412,7 @@ end function shr_string_endIndex
 ! !IROUTINE: shr_string_leftAlign -- remove leading white space
 !
 ! !DESCRIPTION:
-!    Remove leading white space
+!    Remove leading white space (spaces and tabs)
 !     \newline
 !     call shr\_string\_leftAlign(string)
 !
--- a/src/share/util/shr_string_mod.F90
+++ b/src/share/util/shr_string_mod.F90
@@ -412,7 +412,8 @@ end function shr_string_endIndex
 ! !IROUTINE: shr_string_leftAlign -- remove leading white space
 !
 ! !DESCRIPTION:
-!    Remove leading white space (spaces and tabs)
+!    Remove leading white space (spaces and tabs); if input str is entirely spaces and tabs,
+!    then the result has all tabs converted to spaces
 !     \newline
 !     call shr\_string\_leftAlign(string)
 !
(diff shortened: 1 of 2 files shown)


src/share/util/shr_string_mod.F90
```diff
@@ -445,30 +441,73 @@ subroutine shr_string_leftAlign(str,rc)
...
+   index = 0
+   inlength = len(str_input)
+   str_output = ''
+   do i = 1, inlength
+      if (str_input(i:i) /= achar(9)) then
+         index = index + 1
+         str_output(index:index) = str_input(i:i)
+      end if
+   end do
+   str_output(index+1:inlength) = ''
```
jedwards4b: Note that rc doesn't do anything here.

gold2718: I think it is just to match the interface used by all the other functions in this module. Some are actively used, most just set it to zero.

billsacks:
@mvertens , thanks for taking this on. I added some unit tests. These demonstrate that your implementation mostly works, but has the unexpected behavior that interior tabs are also removed: The test `test_shr_string_leftAlign_interiorTabs`, which started with the string `str = 'f' // tab_char // 'oo  '` failed with:

```
    expected: <"f	oo  ">
   but found: <"foo   ">
```

I'd argue that the removal of interior tabs in a subroutine named `shr_string_leftAlign` is a bug, and could cause problematic behavior.

I have reimplemented `shr_string_leftAlign` with one of the algorithms given in the link in #862 . Now all unit tests that I could think of pass. Can you please review this and see if it looks okay to you? One could argue that this is a bit riskier at this point in the release cycle, since it completely changes the implementation of `shr_string_leftAlign`. But given the now-strong coverage by unit tests, I personally feel comfortable with this.

Test suite: Just ran Fortran unit tests on my mac

PR #1551, opened 2017-05-17
Multi component user mods

jedwards4b:
Allows multiple component user_mods. but gives primary component precedent.  
Fixes warning issues for ocn grid non-idmap error if ocn is DOCN
Rearrange tests in scripts_regression_tests to avoid conflict between tests.

Test suite: scripts_regression_tests.py hand tests with F2000_DEV and contrived user_mods
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/grids.py
+++ b/scripts/lib/CIME/XML/grids.py
@@ -401,6 +401,8 @@ class Grids(GenericXML):
                 gridname = grid[0]
                 other_gridname = other_grid[0]
                 gridvalue = component_grids[grid[1]]
+                if gridname == "atm_grid":
+                    atm_gridvalue = gridvalue
                 other_gridvalue = component_grids[other_grid[1]]
                 gridmap_nodes = self.get_nodes(nodename="gridmap",
                                                attributes={gridname:gridvalue, other_gridname:other_gridvalue})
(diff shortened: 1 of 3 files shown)


scripts/lib/CIME/XML/grids.py
```diff
@@ -428,7 +430,10 @@ def _get_gridmaps_v2(self, component_grids):
                 if grid1_value != grid2_value and grid1_value != 'null' and grid2_value != 'null':
                     map_ = gridmaps[node.text]
                     if map_ == 'idmap':
-                        logger.warning("Warning: missing non-idmap %s for %s, %s and %s %s "
+                        if grid1_name == "ocn_grid" and grid1_value == atm_gridvalue:
```
jedwards4b: This avoids the warning: Warning: missing non-idmap ROF2OCN_LIQ_RMAPNAME for ocn_grid, 0.9x1.25 and rof_grid r05.  When the ocn grid matches the atm grid this file isn't used and there is no need for the warning.

scripts/lib/CIME/case.py
```diff
@@ -1053,11 +1049,12 @@ def create_caseroot(self, clone=False):
                           "README.case", caseroot=self._caseroot)
             append_status("%s is %s"%(comp_grid,self.get_value(comp_grid)),
                           "README.case", caseroot=self._caseroot)
-
-        if user_mods is not None:
-            note = "This compset includes user_mods %s"%user_mods
-            append_status(note, "README.case", caseroot=self._caseroot)
-            logger.info(note)
+            comp = str(self.get_value("COMP_%s"%component_class))
```
jedwards4b: Each component can now specify it's own user mods.

rljacob:
Can you give an example of how this works in a create_newcase command?

jedwards4b:
These are specified in the config_component.xml file for each component and not on the command line, there is no change in user interface.

jgfouca:
Waiting for @billsacks review.

Issue #1507, opened 2017-05-10
config_pes_converter is broken 


jedwards4b:
I'm not sure how it ever could have worked - I am getting an error:
 No such file or directory: '../config/cesm/machines/userdefined_laptop_template/config_pes_new.xml'

But it should be opening this file to write, not to read.

sarich:
The error probably means the directory doesn't exist, perhaps because these files and scripts have moved around a bit. I'll take a look.

jedwards4b:
There is another error in the file you'll need to fix
 
-    CIME.utils.parse_args_and_handle_standard_logging_options(args, parser)
+    args = CIME.utils.parse_args_and_handle_standard_logging_options(args, parser)

sarich:
For the first problem (no such file or directory), do you expect this script to recursively create directories if they don't exist, or just to quit gracefully with a message that the given directory is unavailable?

PR #1557, opened 2017-05-17
Sarich/fix config pes converter

sarich:
Remove assumption that a CIME5 file exists to compare entries with. Now if
no file exists it will still write recommended entries to a temp xml file and suggest
copying to the new file.

Also whitespace changes.

Test suite: No test applicable
Test baseline: 
Test namelist changes: 
Test status: 

Fixes Cime issue #1507 

User interface changes?: 

Code review:


rljacob:
@sarich instead of "Fixes Cime issue #1507 " just put "Fixes #1507" for the auto-close to work.

PR #1566, opened 2017-05-19
Support for multiple couplers

jedwards4b:
This adds support for mutiple couplers in multiinstance cases and improves performance of the multiinstance capability.   It adds a new system test MCC to compare multicoupler to single instance runs and adds a new test modifier _C# where # is the number of coupler instances.  

Test suite:  scripts_regression_tests.py ERS_C3.f19_g16_rx1.A.cheyenne_intel, MCC.f19_g16_rx1.A.cheyenne_intel
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes ESMCI/cime #1704 

Updated user doc to reflect this change.

User interface changes?:  Adds _C# modifier to test names for create_test adds --ncouplers option to create_newcase 
Code review:

### DIFF
--- a/scripts/Tools/check_case
+++ b/scripts/Tools/check_case
@@ -50,9 +50,8 @@ def _main_func(description):
 
     parse_command_line(sys.argv, description)
 
-    check_lockedfiles()
-
     with Case(read_only=False) as case:
+        check_lockedfiles(case)
         create_namelists(case)
         build_complete = case.get_value("BUILD_COMPLETE")
 
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -46,6 +46,10 @@ OR
                         help="Specify a compiler. "
                         "To see list of supported compilers for each machine, use the utility query_config in this directory")
 
+    parser.add_argument("--ncouplers",default=1,
+                        help="Specify number of coupler instances. "
+                        "Set the number of coupler instances in the case.")
+
     parser.add_argument("--ninst",default=1,
                         help="Specify number of component instances"
                         "Set the number of component instances in the case.")
@@ -142,6 +146,9 @@ OR
         expect(args.gridfile is not None,
                "User grid specification file must be set if the user grid is requested")
 
+    expect(not (int(args.ncouplers) > 1 and int(args.ninst) > 1),
+           "Only one component instance per coupler is allowed when using multiple couplers")
+
     run_unsupported = False
     if model == "cesm":
         run_unsupported = args.run_unsupported
@@ -155,7 +162,8 @@ OR
     return args.case, args.compset, args.res, args.machine, args.compiler,\
         args.mpilib, args.project, args.pecount, \
         args.user_mods_dir, args.pesfile, \
-        args.user_grid, args.gridfile, args.srcroot, args.test, args.ninst, \
+        args.user_grid, args.gridfile, args.srcroot, args.test, args.ncouplers, \
+        args.ninst, \
         args.walltime, args.queue, args.output_root, args.script_root, \
         run_unsupported, args.answer, args.input_dir
 
@@ -167,7 +175,7 @@ def _main_func(description):
     casename, compset, grid, machine, compiler, \
         mpilib, project, pecount,  \
         user_mods_dir, pesfile, \
-        user_grid, gridfile, srcroot, test, ninst, walltime, queue, \
+        user_grid, gridfile, srcroot, test, ncouplers, ninst, walltime, queue, \
         output_root, script_root, run_unsupported, \
         answer, input_dir = parse_command_line(sys.argv, cimeroot, description)
 
@@ -187,9 +195,11 @@ def _main_func(description):
 
     with Case(caseroot, read_only=False) as case:
         # Configure the Case
-        case.create(casename, srcroot, compset, grid, user_mods_dir=user_mods_dir, machine_name=machine, project=project,
+        case.create(casename, srcroot, compset, grid, user_mods_dir=user_mods_dir,
+                    machine_name=machine, project=project,
                     pecount=pecount, compiler=compiler, mpilib=mpilib,
-                    pesfile=pesfile,user_grid=user_grid, gridfile=gridfile, ninst=ninst, test=test,
+                    pesfile=pesfile,user_grid=user_grid, gridfile=gridfile,
+                    ncouplers=ncouplers, ninst=ninst, test=test,
                     walltime=walltime, queue=queue, output_root=output_root,
                     run_unsupported=run_unsupported, answer=answer,
                     input_dir=input_dir)
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -46,13 +46,13 @@ OR
                         help="Specify a compiler. "
                         "To see list of supported compilers for each machine, use the utility query_config in this directory")
 
-    parser.add_argument("--ninst-couplers",action="store_true",
+    parser.add_argument("--multi-coupler",action="store_true",
                         help="Specify that ninst should modify number of coupler instances "
                         "default is to have one coupler supporting multiple component instances.")
 
     parser.add_argument("--ninst",default=1,
                         help="Specify number of model ensemble instances. "
-                        "Default is multiple components and one coupler.  Use --ninst-couplers to "
+                        "Default is multiple components and one coupler.  Use --multi-coupler to "
                         "run multiple couplers in the ensemble.")
 
     parser.add_argument("--mpilib", "-mpilib",
@@ -157,7 +157,7 @@ OR
     if args.input_dir is not None:
         args.input_dir = os.path.abspath(args.input_dir)
 
-    if args.ninst_couplers:
+    if args.multi_coupler:
         ncouplers = args.ninst
         ninst = 1
     else:
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -48,11 +48,13 @@ OR
 
     parser.add_argument("--ncouplers",default=1,
                         help="Specify number of coupler instances. "
-                        "Set the number of coupler instances in the case.")
+                        "Set the number of coupler instances in the case. "
+                        "If ncouplers is > 1 then ninst should = 1.")
 
     parser.add_argument("--ninst",default=1,
-                        help="Specify number of component instances"
-                        "Set the number of component instances in the case.")
+                        help="Specify number of component instances. "
+                        "Set the number of component instances in the case. "
+                        "If ninst > 1 then ncouplers should = 1")
 
     parser.add_argument("--mpilib", "-mpilib",
                         help="Specify the mpilib. "
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -157,19 +157,11 @@ OR
     if args.input_dir is not None:
         args.input_dir = os.path.abspath(args.input_dir)
 
-    if args.multi_coupler:
-        ncouplers = args.ninst
-        ninst = 1
-    else:
-        ncouplers = 1
-        ninst = args.ninst
-
-
     return args.case, args.compset, args.res, args.machine, args.compiler,\
         args.mpilib, args.project, args.pecount, \
         args.user_mods_dir, args.pesfile, \
-        args.user_grid, args.gridfile, args.srcroot, args.test, ncouplers, \
-        ninst, args.walltime, args.queue, args.output_root, args.script_root, \
+        args.user_grid, args.gridfile, args.srcroot, args.test, args.multi_coupler, \
+        args.ninst, args.walltime, args.queue, args.output_root, args.script_root, \
         run_unsupported, args.answer, args.input_dir
 
 ###############################################################################
@@ -180,8 +172,8 @@ def _main_func(description):
     casename, compset, grid, machine, compiler, \
         mpilib, project, pecount,  \
         user_mods_dir, pesfile, \
-        user_grid, gridfile, srcroot, test, ncouplers, ninst, walltime, queue, \
-        output_root, script_root, run_unsupported, \
+        user_grid, gridfile, srcroot, test, multi_coupler, ninst, walltime, \
+        queue, output_root, script_root, run_unsupported, \
         answer, input_dir = parse_command_line(sys.argv, cimeroot, description)
 
     if script_root is None:
@@ -204,7 +196,7 @@ def _main_func(description):
                     machine_name=machine, project=project,
                     pecount=pecount, compiler=compiler, mpilib=mpilib,
                     pesfile=pesfile,user_grid=user_grid, gridfile=gridfile,
-                    ncouplers=ncouplers, ninst=ninst, test=test,
+                    multi_coupler=multi_coupler, ninst=ninst, test=test,
                     walltime=walltime, queue=queue, output_root=output_root,
                     run_unsupported=run_unsupported, answer=answer,
                     input_dir=input_dir)
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -46,15 +46,14 @@ OR
                         help="Specify a compiler. "
                         "To see list of supported compilers for each machine, use the utility query_config in this directory")
 
-    parser.add_argument("--ncouplers",default=1,
-                        help="Specify number of coupler instances. "
-                        "Set the number of coupler instances in the case. "
-                        "If ncouplers is > 1 then ninst should = 1.")
+    parser.add_argument("--ninst-couplers",action="store_true",
+                        help="Specify that ninst should modify number of coupler instances "
+                        "default is to have one coupler supporting multiple component instances.")
 
     parser.add_argument("--ninst",default=1,
-                        help="Specify number of component instances. "
-                        "Set the number of component instances in the case. "
-                        "If ninst > 1 then ncouplers should = 1")
+                        help="Specify number of model ensemble instances. "
+                        "Default is multiple components and one coupler.  Use --ninst-couplers to "
+                        "run multiple couplers in the ensemble.")
 
     parser.add_argument("--mpilib", "-mpilib",
                         help="Specify the mpilib. "
@@ -148,9 +147,6 @@ OR
         expect(args.gridfile is not None,
                "User grid specification file must be set if the user grid is requested")
 
-    expect(not (int(args.ncouplers) > 1 and int(args.ninst) > 1),
-           "Only one component instance per coupler is allowed when using multiple couplers")
-
     run_unsupported = False
     if model == "cesm":
         run_unsupported = args.run_unsupported
@@ -161,12 +157,19 @@ OR
     if args.input_dir is not None:
         args.input_dir = os.path.abspath(args.input_dir)
 
+    if args.ninst_couplers:
+        ncouplers = args.ninst
+        ninst = 1
+    else:
+        ncouplers = 1
+        ninst = args.ninst
+
+
     return args.case, args.compset, args.res, args.machine, args.compiler,\
         args.mpilib, args.project, args.pecount, \
         args.user_mods_dir, args.pesfile, \
-        args.user_grid, args.gridfile, args.srcroot, args.test, args.ncouplers, \
-        args.ninst, \
-        args.walltime, args.queue, args.output_root, args.script_root, \
+        args.user_grid, args.gridfile, args.srcroot, args.test, ncouplers, \
+        ninst, args.walltime, args.queue, args.output_root, args.script_root, \
         run_unsupported, args.answer, args.input_dir
 
 ###############################################################################
--- a/scripts/lib/CIME/XML/entry_id.py
+++ b/scripts/lib/CIME/XML/entry_id.py
@@ -121,7 +121,7 @@ class EntryID(GenericXML):
                     max_score = score
                     mnode = node
             else:
-                expect(False, 
+                expect(False,
                        "match attribute can only have a value of 'last' or 'first', value is %s" %match_type)
 
         return mnode.text
@@ -400,20 +400,27 @@ class EntryID(GenericXML):
                 if f1val is not None:
                     f2val = other.get_value(vid, resolved=False)
                     if f1val != f2val:
+                        logger.info("HERE %s %s "%(f1val, f2val))
                         xmldiffs[vid] = [f1val, f2val]
                 else:
-                    f1val = ET.tostring(node, method="text")
-                    f2val = ET.tostring(f2match, method="text")
-                    if f2val != f1val:
-                        f1value_nodes = self.get_nodes("value", root=node)
-                        for valnode in f1value_nodes:
-                            f2valnodes = other.get_nodes("value", root=f2match, attributes=valnode.attrib)
-                            for f2valnode in f2valnodes:
-                                if valnode.attrib is None and f2valnode.attrib is None or \
-                                   f2valnode.attrib == valnode.attrib:
-                                    if other.get_resolved_value(f2valnode.text) != self.get_resolved_value(valnode.text):
-                                        xmldiffs["{}:{}".format(vid, valnode.attrib)] = [valnode.text, f2valnode.text]
-
+                    for comp in self.get_values("COMP_CLASSES"):
+                        f1val = self.get_value("{}_{}".format(vid,comp), resolved=False)
+                        if f1val is not None:
+                            f2val = other.get_value("{}_{}".format(vid,comp), resolved=False)
+                            if f1val != f2val:
+                                xmldiffs[vid] = [f1val, f2val]
+                        else:
+                            f1val = ET.tostring(node, method="text")
+                            f2val = ET.tostring(f2match, method="text")
+                            if f2val != f1val:
+                                f1value_nodes = self.get_nodes("value", root=node)
+                                for valnode in f1value_nodes:
+                                    f2valnodes = other.get_nodes("value", root=f2match, attributes=valnode.attrib)
+                                for f2valnode in f2valnodes:
+                                    if valnode.attrib is None and f2valnode.attrib is None or \
+                                       f2valnode.attrib == valnode.attrib:
+                                        if other.get_resolved_value(f2valnode.text) != self.get_resolved_value(valnode.text):
+                                            xmldiffs["{}:{}".format(vid, valnode.attrib)] = [valnode.text, f2valnode.text]
         return xmldiffs
 
     def __iter__(self):
--- a/scripts/lib/CIME/XML/entry_id.py
+++ b/scripts/lib/CIME/XML/entry_id.py
@@ -400,7 +400,6 @@ class EntryID(GenericXML):
                 if f1val is not None:
                     f2val = other.get_value(vid, resolved=False)
                     if f1val != f2val:
-                        logger.info("HERE %s %s "%(f1val, f2val))
                         xmldiffs[vid] = [f1val, f2val]
                 else:
                     for comp in self.get_values("COMP_CLASSES"):
--- a/scripts/lib/CIME/XML/entry_id.py
+++ b/scripts/lib/CIME/XML/entry_id.py
@@ -416,11 +416,11 @@ class EntryID(GenericXML):
                                 f1value_nodes = self.get_nodes("value", root=node)
                                 for valnode in f1value_nodes:
                                     f2valnodes = other.get_nodes("value", root=f2match, attributes=valnode.attrib)
-                                for f2valnode in f2valnodes:
-                                    if valnode.attrib is None and f2valnode.attrib is None or \
-                                       f2valnode.attrib == valnode.attrib:
-                                        if other.get_resolved_value(f2valnode.text) != self.get_resolved_value(valnode.text):
-                                            xmldiffs["{}:{}".format(vid, valnode.attrib)] = [valnode.text, f2valnode.text]
+                                    for f2valnode in f2valnodes:
+                                        if valnode.attrib is None and f2valnode.attrib is None or \
+                                           f2valnode.attrib == valnode.attrib:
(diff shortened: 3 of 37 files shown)


scripts/Tools/check_case
```diff
@@ -50,9 +50,8 @@ def _main_func(description):
 
     parse_command_line(sys.argv, description)
 
-    check_lockedfiles()
-
     with Case(read_only=False) as case:
+        check_lockedfiles(case)
```
mvertens: Why has the check_lockedfiles been moved here with a new argument? Was this tied to multiple couplers - or was this needed regardless?

scripts/create_newcase
```diff
@@ -47,9 +47,15 @@ OR
...
+                        help="Specify number of coupler instances. "
+                        "Set the number of coupler instances in the case. "
+                        "If ncouplers is > 1 then ninst should = 1.")
+
     parser.add_argument("--ninst",default=1,
-                        help="Specify number of component instances"
-                        "Set the number of component instances in the case.")
+                        help="Specify number of component instances. "
+                        "Set the number of component instances in the case. "
+                        "If ninst > 1 then ncouplers should = 1")
```
mvertens: So you should allow the user to either specify --ncouplers OR --ninst and the appropriate ninst or ncouplers from the appropriate input. You should also make it clear from the documentation that if --ncouplers is set to N then the number of instances of each component will also be set to N. The key point is that ninst relates to the number of component instances for a **single coupler.** Not to the total number of component instances. This needs to be spelled out more clearly.
Also having the number of instances by a new argument to create_newcase is a major change in how we address multiple instances. It is assumed that the arguments to create_newcase (compset, grids) cannot be modified once the case is created. This has never been the case with the number of instances which could be modified once the case was created. I assume that you have this as an argument in order to have the correct pe-layout appear when you select multiple instances and that makes sense. Do we want to migrate away from permitting users to change the number of instances after they have created the case. We need to clarify this both as a requirement and in the documentation.

scripts/create_newcase
```diff
@@ -149,6 +155,9 @@ OR
         expect(args.gridfile is not None,
                "User grid specification file must be set if the user grid is requested")
 
+    expect(not (int(args.ncouplers) > 1 and int(args.ninst) > 1),
```
mvertens: This needs to change to address the above comment - either --ninst or --ncouplers should be allowed.

scripts/lib/CIME/XML/entry_id.py
```diff
@@ -400,20 +400,27 @@ def compare_xml(self, other):
...
-                        f1value_nodes = self.get_nodes("value", root=node)
-                        for valnode in f1value_nodes:
-                            f2valnodes = other.get_nodes("value", root=f2match, attributes=valnode.attrib)
-                            for f2valnode in f2valnodes:
-                                if valnode.attrib is None and f2valnode.attrib is None or \
-                                   f2valnode.attrib == valnode.attrib:
-                                    if other.get_resolved_value(f2valnode.text) != self.get_resolved_value(valnode.text):
-                                        xmldiffs["{}:{}".format(vid, valnode.attrib)] = [valnode.text, f2valnode.text]
-
+                    for comp in self.get_values("COMP_CLASSES"):
```
mvertens: Why did this need to get changed in entry_id.py as part of this PR? How does this change relate to this PR?

scripts/lib/CIME/check_lockedfiles.py
```diff
@@ -77,33 +77,34 @@ def check_pelayouts_require_rebuild(case, models):
 
         unlock_file("env_mach_pes.xml", case.get_value("CASEROOT"))
 
-def check_lockedfiles(caseroot=None):
+def check_lockedfiles(case):
```
mvertens: It seems that this change is independent of this PR. If this is the case - it should be brought in as a separate PR. If this is not the case - why did this change to the interface need to be part of this PR.

scripts/lib/CIME/get_timing.py
```diff
@@ -93,6 +93,14 @@ def gettime(self, heading_padded):
         return (0, 0, False)
 
     def getTiming(self):
+        ninst = self.case.get_value("NINST_CPL")
```
mvertens: This variable should be cpl_ninst  - not ninst (since the meaning of ninst really implies the number of instances of prognostic components for a single coupler).  This applies to all the ninst variables in the routine below. We should pick a standard variable name for the number of coupler instances - either ninst_cpl or cpl_ninst and be consistent in its usage. I would prefer ninst_cpl since this is what is used for the xml variable.

scripts/lib/CIME/preview_namelists.py
```diff
@@ -70,8 +72,20 @@ def create_namelists(case, component=None):
         config_dir = os.path.dirname(config_file)
         if model_str == "cpl":
             compname = "drv"
+            complist = [m for m in models if m.upper() != "CPL"]
+            if cpl_ninst > 1:
```
mvertens: Please document what the xmlfact dictionary does and why its initialized the way it is. Similarly for xmlsave. Why are these two dictionaries needed here.

src/drivers/mct/cime_config/config_archive.xml
```diff
@@ -4,8 +4,8 @@
     <hist_file_extension>\.h.*.nc$</hist_file_extension>
     <rest_history_varname>unset</rest_history_varname>
     <rpointer>
-      <rpointer_file>rpointer.cpl</rpointer_file>
-      <rpointer_content >$CASE.cpl.r.$DATENAME.nc</rpointer_content>
+      <rpointer_file>rpointer$NINST_STRING.cpl</rpointer_file>
```
mvertens: Where is NINST_STRING set. What tests were done with archiving to ensure that this is working correctly in both the new and older implementations?

src/drivers/mct/cime_config/namelist_definition_drv.xml
```diff
@@ -41,6 +41,22 @@
   -->
 
   <!-- =========================== -->
+  <!-- group cesm_cpl              -->
```
mvertens: Please lets call this ninst_cpl and not cpl_ninst for consistency.

src/drivers/mct/cime_config/testdefs/testlist_drv.xml
```diff
@@ -23,6 +23,15 @@
       </machine>
     </machines>
   </test>
+  <test name="ERS_C3" grid="f19_g17_rx1" compset="A">
```
mvertens: So does this test now really test the multiple instance functionality for an A compset? I think the answer is yes. We decided that ERS_N3 did not I believe.

src/drivers/mct/main/cesm_comp_mod.F90
```diff
@@ -4042,4 +4066,60 @@ subroutine cesm_comp_barriers(mpicom, timer)
...
+  integer , intent(inout) :: comm
+  integer , intent(out)   :: ninst
+  integer , intent(out)   :: id      ! instance ID, starts from 1
+  !
+  ! Local variables
+  !
+  integer :: ierr, inst_comm, mype, nu, numpes !, pes
+  integer :: cpl_ninst
+
+  namelist /cesm_cpl/ cpl_ninst
```
mvertens: I don't think we should be introducing "cesm" into any namelist or variable names. The namelist group could be multi_inst as a possibility.

src/drivers/mct/main/cesm_comp_mod.F90
```diff
@@ -4042,4 +4066,60 @@ subroutine cesm_comp_barriers(mpicom, timer)
...
+
+  call shr_mpi_commrank(comm, mype  , ' cesm_cpl_init')
```
mvertens: Again - we should not be introducing new strings with "cesm" in them.

src/drivers/mct/main/seq_hist_mod.F90
```diff
@@ -131,7 +131,7 @@ module seq_hist_mod
 subroutine seq_hist_write(infodata, EClock_d, &
      atm, lnd, ice, ocn, rof, glc, wav, &
      fractions_ax, fractions_lx, fractions_ix, fractions_ox, fractions_rx,  &
-     fractions_gx, fractions_wx)
+     fractions_gx, fractions_wx, tag)
```
mvertens: The argument should be cpl_inst_tag rather than just tag for clarity.

src/drivers/mct/shr/seq_comm_mct.F90
```diff
@@ -199,14 +199,15 @@ integer function seq_comm_get_ncomps()
...
-  subroutine seq_comm_init(Comm_in, nmlfile)
+  subroutine seq_comm_init(Comm_in, nmlfile, Comm_ID)
 
     !----------------------------------------------------------
     !
     ! Arguments
     implicit none
     integer, intent(in) :: Comm_in
     character(len=*), intent(IN) :: nmlfile
+    integer, optional, intent(in) :: Comm_ID
```
mvertens: Could you please add a sentence or two as to why we need the Comm_ID  optional argument. We need to explain how this new argument differentiates between multiple couplers and 1 component instance per coupler versus a single coupler and multi-component instances per that one coupler.

scripts/Tools/check_case
jedwards4b: In the process of testing with ERS_C4 I cleaned up implementation of this tool.

scripts/create_newcase
jedwards4b: If ninst is > 1, then ncouplers must == 1 and visa versa, this is implemented here.   ninst is not a new argument to create_newcase, it's on the trunk.   You can still change it after the case is created.  Happy to work on documentation with you but this is not a new change in this branch, the only change here is the addition of -ncouplers and its analogous to how ninst was already implemented here.

scripts/create_newcase
jedwards4b: Boy I'm really confused by your comments - maybe we need to talk? Either is allowed.

scripts/lib/CIME/XML/entry_id.py
jedwards4b: It was noted in testing this PR that the xml comparison implemented here could be improved and so I implemented an improvement.

src/drivers/mct/cime_config/config_archive.xml
jedwards4b: The ERR test covers this.

src/drivers/mct/cime_config/testdefs/testlist_drv.xml
jedwards4b: I think that you might be confusing this with anther test, ERS tests restart capability and works in an A case for either N3 or C3.

jgfouca:
Waiting for @billsacks review.

mvertens:
I think we can take Bill Sacks off of the review. I'd like to know if this
PR was run through a CESM pre-alpha suite as a sanity check that it does
not change any answers. Also - we do need to update the documentation in
create_newcase as well.

billsacks:
Thanks, @mvertens - yes, I don't think I'll have time to look at this closely this week.

I do have one comment based on skimming through the discussion here: I find it unintuitive to have `--ninst` and `--ncouplers`, where you can only specify one of them, and `-ncouplers` really means, "Run this number of instances with one coupler per instance." Instead, I'd find it more intuitive if the interface was: `--ninst` is always used to specify the number of instances. Then there is a separate argument like `--multiple-couplers` or `--one-coupler-per-inst` which simply sets a logical flag.

I'm not completely tied to that, though, if everyone else prefers the interface as it is now.

Another alternative would be to keep the idea as it is now, but to rename the arguments to `--ninst-single-coupler` and `--ninst-multiple-couplers`. Mainly I find `--ncouplers` to be an unintuitive argument name.

gold2718:
I second what @billsacks says and suggest:
--single-coupler
--ninst-couplers
with the default being whichever we think works better.

jedwards4b:
I disagree with the comments about changing the option names.   I do not think that the suggested changes add any clarity.

mvertens:
I've been thinking about this - and I think there are valid points raised
about changing the option names. If there is strong disagreement - we
should get other opinions before making a final decision. We don't want to
change the option names multiple times - and I think we should do what is
most intuitive to the largest number of people.

gold2718:
My point is to keep current usage of `--ninst` as meaning 'give me N instances in a default configuration'. The `--single-coupler` and/or `--ninst-couplers` option would only be needed to change the default behavior. I assume that the default behavior will be `--ninst-couplers` and if it is, we can even drop that option.

singhbalwinder:
I have a very limited knowledge of how this works. I think the way `ninst` works is that we have to create and provide multiple (equals the value of `ninst`) namelist files for each prognostic component. With multiple coupler option, the difference in the usage would be to provide multiple drv_in files also?

billsacks:
I'm happy with @gold2718 's suggestion

> My point is to keep current usage of `--ninst` as meaning 'give me N instances in a default configuration'. The `--single-coupler` and/or `--ninst-couplers` option would only be needed to change the default behavior. I assume that the default behavior will be `--ninst-couplers` and if it is, we can even drop that option.

and also agree with @mvertens 's suggestion

> If there is strong disagreement - we should get other opinions before making a final decision.

jedwards4b:
Option name change:

  --ninst-couplers      Specify that ninst should modify number of coupler
                        instances default is to have one coupler supporting
                        multiple component instances. (default: False)
  --ninst NINST         Specify number of model ensemble instances. Default is
                        multiple components and one coupler. Use --ninst-
                        couplers to run multiple couplers in the ensemble.
                        (default: 1)

I also ran PFS_C30.f10_f10.QPC4.cheyenne_intel and PFS_N30.f10_f10.QPC4.cheyenne_intel
performance was 400 ypd and 244 ypd respectively with each using 1080 total tasks.

jgfouca:
OK, is this ready to merge?

scripts/lib/CIME/test_scheduler.py
```diff
@@ -395,7 +395,7 @@ def _create_newcase_phase(self, test):
                     logger.debug (" NINST set to {}".format(ninst))
                 if case_opt.startswith('C'):
                     ncpl = case_opt[1:]
-                    create_newcase_cmd += " --ncouplers {}" .format(ncpl)
+                    create_newcase_cmd += " --ninst {} --ninst-coupler" .format(ncpl)
```
billsacks: This says `--ninst-coupler` but the actual argument is `--ninst-couplers`

scripts/create_newcase
```diff
@@ -168,11 +164,19 @@ OR
     if args.input_dir is not None:
         args.input_dir = os.path.abspath(args.input_dir)
 
+    if args.ninst_couplers:
+        ncouplers = args.ninst
+        ninst = 1
```
billsacks: Shouldn't this set `ninst = args.ninst`?

I tried `./create_newcase --ninst 3 --ninst-couplers --case test_ninst_0621b --compset A --res f45_g37_rx1`

and got:

```
[roo2:~/cime/scripts/test_ninst_0621b]$ ./xmlquery NINST
	NINST: ['CPL:3', 'ATM:1', 'LND:1', 'ICE:1', 'OCN:1', 'ROF:1', 'GLC:1', 'WAV:1', 'ESP:1']
```

whereas I expected NINST to be 3 for everything except for ESP.

jedwards4b: If ncouplers is > 1 then each component instance must == 1.

billsacks: Are you sure about that? That doesn't make sense to me. I think this misunderstanding (on one of our parts) is the cause for the different opinions on the best user interface here. I can't see any rationale for having multiple coupler instances with only one of each component instance. @mvertens or @gold2718 can you chime in here to say which is right?

jedwards4b: I am sure - because this is implemented over the existing multiinstance code some clever relationships had to be established - primarily if NINST_CPL>1 the NINST_* for all other components must == 1 and the entire model (coupler + components) is replicated NINST_CPL times.   If NINST_* > 1 for any component other than coupler then those components are replicated within  the single coupler.   There are reasons that we want to retain the old behavior SPCAM and any case in which we want multiple instances of one component but not others will need to use the old NINST behavior.

jedwards4b:
Thanks @bertinia I've made changes in code and doc to clarify.

billsacks:
@jedwards4b - thanks for the clarifications yesterday on how these ninst things work. Given this, I retract my earlier comment

> Shouldn't this set `ninst = args.ninst`?

However, I think three additional things are needed then:

(1) Need to fix the documentation of NINST. Currently if I set up a case with `--ninst 3 --ninst-couplers` and do `./xmlquery --full NINST`, I get

```
Results in group mach_pes
	NINST: value=['CPL:3', 'ATM:1', 'LND:1', 'ICE:1', 'OCN:1', 'ROF:1', 'GLC:1', 'WAV:1', 'ESP:1']
		type: i
		valid_values: []
		description: Number of instances for each component
		file: /Users/sacks/cime/scripts/test_multi_couplers_0622b/env_mach_pes.xml
```

But this documentation implies that I have 1 atmosphere instance, 1 lnd instance, etc. In fact, I have 3 atm instances, 3 lnd instances, etc. If I understand this correctly, the correct description for everything other than NINST CPL is something to the effect of, "Number of component instances per coupler instance. If CPL NINST > 1, then NINST should be 1 for each other component, and the number of instances of each component is equal to the CPL NINST value."

(2) I have the impression that (as my proposed documentation suggests) we have a rule: If CPL NINST > 1, then NINST must be 1 for all other components. But this does not seem to be enforced by the scripts. For example, if I set up `--ninst 3 --ninst-couplers`, then do `./xmlchange NINST_ATM=3`, I can get through case.setup without any errors. `./preview_namelists` generates this error message:

```
ERROR: Missing required user_nl_file /Users/sacks/cime/scripts/test_multi_couplers_0622a/user_nl_datm_0004
```

which does not really help a user identify the source of the problem.

If I'm right about this requirement, then I think it's important that the scripts check this rule and give a meaningful error message if someone tries to set both CPL NINST and some other NINST to a value > 1. I think this is something that users are likely to do, because it feels intuitive to me to set all component NINST values to the number of instances you want.

(3) I created a case without `--ninst` or `--ninst-couplers` then ran `./xmlchange NINST=3`. I got

```
[roo2:~/cime/scripts/test_multi_couplers_0622c]$ ./xmlquery --full NINST

Results in group mach_pes
	NINST: value=['CPL:3', 'ATM:3', 'LND:3', 'ICE:3', 'OCN:3', 'ROF:3', 'GLC:3', 'WAV:3', 'ESP:3']
		type: i
		valid_values: []
		description: Number of instances for each component
		file: /Users/sacks/cime/scripts/test_multi_couplers_0622c/env_mach_pes.xml
```

As I now understand it, this is wrong in two respects:

(i) You're not allowed to have NINST_ESP > 1, ever

(ii) It's incorrect to set both NINST_CPL > 1 and NINST of components > 1

The above xmlchange command feels like the intuitive way to change the number of instances after a case is set up, so it feels wrong to me that running this command leads to faulty behavior.

One way to fix this would be to:

- Remove ESP from the list of components in the NINST list, since this always has to be 1

- Put NINST_CPL in some separate list, not the generic NINST list, probably renaming this. So, for example, you could choose to do `./xmlchange NINSTANCES=3` which would give you 3 instances of the full system (equivalent to what you'd currently get by setting NINST_CPL=3), OR you could do `./xmlchange NINST=3`, which would set all NINST_XXX values to 3, but CPL would no longer be in this list. This would also help with the documentation of these options: NINSTANCES means number of instances of the full system (coupler and all components); NINST_FOO gives number of instances of component FOO, which is only valid if NINSTANCES=1.

I'm not tied to this exact proposed solution, though it would be nice if we had _some_ solution that led `./xmlchange NINST=3` to Just Work.

gold2718:
Can we have an item in, say, config_component.xml which specifies whether or not multi-instance is supported? Currently, that is (at least) ESP and WW3 so this would be a data-driven solution to either allow setting NINST via xmlchange or to cause an error if the NINST change tries to set a component which does not support the change.
We do want multi-instance with multi-coupler to work with ESP components -- that is one of the main reasons ESP was developed.

billsacks:
@gold2718 - that seems like a good solution to part of the problem. It feels like we'd still need to separate (and rename) the current NINST_CPL from the other NINST terms for this to work cleanly.

jedwards4b:
Putting the level of error checking in the code that @billsacks is suggesting would be a considerable new investment in this PR.   I agree that in order to do that step one will be to rename NINST_CPL to something else - what do you suggest calling it?

billsacks:
I suggested NINSTANCES, but I'm open to other suggestions. I'm okay with deferring the error checking to a later PR. I think the most important thing from my request is point (1) - i.e., the documentation of the meaning of these variables now. I'd be okay with opening a separate issue related to points (2) and (3) if you'd prefer.

gold2718:
I would prefer something like `MULTI_COUPLER` which would be a logical. Then, leave all of the other `NINST_<comp>` items alone so the user can see that the case is indeed configured for multiple instances (or not). To me, this is more intuitive for viewing.
When the case is setup, a `TRUE` value for `MULTI_COUPLER` would set the instance numbers of the components back to one for correct functioning.

jedwards4b:
Refactored and replaced NINST_CPL with COUPLER_COUNT.   Assure that NINST cannot be set > 1 if COUPLER_COUNT > 1 and visa versa.    preview_namelists needs a way to override this restriction (basically the buildnml scripts need to be fooled into doing the right thing) and we do this by allowing a negative value of NINST to override the check described above.

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -1933,10 +1933,17 @@
     <desc>ROOTPE (mpi task in MPI_COMM_WORLD) for each component</desc>
   </entry>
 
+  <entry id="COUPLER_COUNT">
+    <type>integer</type>
+    <default_value>1</default_value>
+    <group>mach_pes</group>
+    <file>env_mach_pes.xml</file>
+    <desc>Number of couplers in ensemble: if COUPLER_COUNT > 1 then NINST must be 1 for all components.</desc>
```
billsacks: I think this should say something more than this: My understanding is that this effectively sets the number of total instances of the system. e.g., if COUPLER_COUNT is 5, then we have 5 couplers, 5 atm instances, 5 lnd instances, etc.

gold2718:
@billsacks,  breaking things up to try and stick with one issue per post.
> Also, just to be clear: When you say:
> 
>     When the case is setup, a TRUE value for MULTI_COUPLER would set the instance numbers of the components back to one for correct functioning.
> 
> do you mean that this resetting happens internally or in the xml file? I'd be opposed to it happening in the xml file, because I'd find it confusing to have values changed out from under me like that.
I would like the resetting to all be internal and would like the XML file to show the true number of instances which will be running.

billsacks:
@gold2718 - thanks for the clarification. That sounds appealing if we can come up with a way to clarify / enforce the semantic difference I raised.

gold2718:
@billsacks & @jedwards4b,
I agree that we need to be careful of the semantics of the multi-coupler feature. I think discussion of this should not be here but in the issue that this PR is addressing. However, I notice that there is no such issue so I have created one, #1704. Please comment there.

scripts/lib/CIME/XML/env_mach_pes.py
```diff
@@ -19,6 +19,32 @@ def __init__(self, case_root=None, infile="env_mach_pes.xml", components=None):
...
+                value = -value
+        elif vid == "NINST":
+            if value > 1:
+                coupler_count = self.get_value("COUPLER_COUNT")
+                expect(coupler_count == 1,"Cannot change NINST value if COUPLER_COUNT > 1")
+            elif value < 0:
+                # negative value effectively overrides safety check
+                value = -value
+        return EnvBase.set_value(self, vid, value,subgroup=subgroup, ignore_type=ignore_type)
+
```
gold2718: If I specify `--ninst 4 --ninst-couplers` but then decide to only use one coupler with `./xmlchange COUPLER_COUNT=1`, I end up with a single-instance simulation which I think would be confusing to users who never intended to change the number of instances.

billsacks:
@jedwards4b - I think you have addressed most of my comments now - thanks! The one outstanding one is my [request for an explanatory comment in the code](https://github.com/ESMCI/cime/pull/1566/files/3228521efdc2be18b90296932a6edaef52080b04..fbc5bbaf980c4bd3ccccba9937936c3f5e9b1d82#diff-8032cd6a7aeca62edced917942b9d254)

scripts/lib/CIME/XML/env_mach_pes.py
```diff
@@ -19,6 +19,32 @@ def __init__(self, case_root=None, infile="env_mach_pes.xml", components=None):
...
+        if vid == "COUPLER_COUNT":
+            if value > 1:
+                for othercomp in self._components:
+                    if othercomp != "CPL":
+                        ninst_string = "NINST_{}".format(othercomp)
+                        expect(self.get_value(ninst_string)==1,
+                               "Cannot change COUPLER_COUNT value if {} > 1".format(ninst_string))
+            elif value < 0:
+                # negative value effectively overrides safety check
+                value = -value
```
jedwards4b: Because in order to get buildnml scripts to work correctly we must override this check.

gold2718: Or you could fix the interface so that secret backdoors are not needed. Right now, I can execute `./xmlchange COUPLER_COUNT=-15` and it happily sets `COUPLER_COUNT`  to `15`. Since `xmlchange` is part of our user interface, this feels like a bad feature to me.

scripts/lib/CIME/XML/env_mach_pes.py
jedwards4b: Do you not think it would be confusing to someone who thought that by changing COUPLER_COUNT=1 they were reverting to a single member if suddenly the NINST values were changed?

scripts/lib/CIME/case.py
```diff
@@ -1063,6 +1069,9 @@ def create_caseroot(self, clone=False):
                       "README.case", caseroot=self._caseroot)
         for component_class in self._component_classes:
             if component_class == "CPL":
+                append_status("Using %s coupler instances" %
+                              (self.get_value("COUPLER_COUNT")),
+                              "README.case", caseroot=self._caseroot)
```
gold2718: In addition to the number of couplers, please add the number of instances to README.case

scripts/lib/CIME/case.py
```diff
@@ -827,7 +830,10 @@ def configure(self, compset_name, grid_name, machine_name=None,
         env_mach_specific_obj.populate(machobj)
         self.schedule_rewrite(env_mach_specific_obj)
 
-        pesize = self._setup_mach_pes(pecount, ninst, machine_name, mpilib)
+        pesize = self._setup_mach_pes(pecount, ncouplers, ninst, machine_name, mpilib)
+
+        if ncouplers > 1:
+            logger.info(" Coupler has %s instances" % ncouplers)
```
gold2718: Component instance count should also be logged here.

scripts/lib/CIME/XML/env_mach_pes.py
gold2718: Well of course which is why I have advocated for leaving `NINST` as the user specified it and replacing the integer `COUPLER_COUNT` with a logical, `MULTI_COUPLER`. This accurately describes both the user's intent and the run and makes `NINST` and `MULTI_COUPLER` orthogonal variables which cleans up the case-control logic.
Then, the necessary settings can be effected in build.py.

jedwards4b:
I have removed COUPLER_COUNT and added MULTI_COUPLER logical flag along with replacing NINST_CPL.   If you toggle the MULTI_COUPLER flag on the max NINST_* value is copied to the coupler and all other values are set to 1.  If you toggle it off all values except ESP are set to NINST_CPL and NINST_CPL is set back to 1.  I have rerun scripts_regression_tests.py along with the 
MCC and PFS tests mentioned above.

jedwards4b:
@mvertens @gold2718 @billsacks please re-review

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -1947,7 +1960,8 @@
     </values>
     <group>mach_pes</group>
     <file>env_mach_pes.xml</file>
-    <desc>Number of instances for each component</desc>
+    <desc>Number of instances for each component.  If MULTI_COUPLER is True
+    only NINST_CPL is used, if MULTI_COUPLER is False NINST_CPL is 1.</desc>
```
billsacks: After "only NINST_CPL is used" I would add something like, "and all components have this many instances", then change the following comma to a semi-colon. So the full description will read:

```
Number of instances for each component. If MULTI_COUPLER is True, only NINST_CPL is used and all components have this many instances; if MULTI_COUPLER is False, NINST_CPL is 1.
```

scripts/lib/CIME/preview_namelists.py
```diff
@@ -68,11 +68,17 @@ def create_namelists(case, component=None):
...
+        multicoupler = case.get_value("MULTI_COUPLER")
         if model_str == "cpl":
             compname = "drv"
         else:
             compname = case.get_value("COMP_{}".format(model_str.upper()))
-
+        # Multicoupler mode (MULTI_COUPLER ) must temporarily change
+        # NINST and NTASKS settings so that the component buildnml scripts
+        # will work correctly.  After the call to buildnml the original values
+        # are restored.
```
billsacks: Is the above comment still correct?

jedwards4b:
I've done another major refactor of this PR and I must admit I'm getting a bit cross-eyed looking at it.   A careful review would be appreciated.

jedwards4b:
I am closing this PR and replacing with a multi-driver implementation to be posted soon.

PR #1559, opened 2017-05-17
Compset var settings

jedwards4b:
I introduced a bug which was preventing config_compsets from modifying xml variables. 

Test suite: scripts_regression_tests.py 
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1558 
Fixes #1538

User interface changes?: 

Code review: fischer

PR #1567, opened 2017-05-19
Point to gx1v7 domain files that exist

### LINKED ISSUE #1558: config_compsets not parsed correctly
This bug was introduced in 62fa492 (PR #1523) and is causing the time orientation of compsets to be incorrect.

### LINKED ISSUE #1565: Incorrect date stamps on two domain files with gx1v7 mask
When you set up a D compset with `g17_g17` resolution, you run into the following issue (this was reported in #1538)

```
$ grep DOMAIN env_run.xml 
    <entry id="ATM_DOMAIN_FILE" value="domain.ocn.gx1v7.151020.nc">
```

(The issue being that file does not exist). Also, when you run `preview_namelist` in an F compset with the `f09_f09_mg17 ` resolution you get an error pointing to a similar issue:

```
File not found: domainfile = "/glade/p/cesmdata/cseg/inputdata/share/domains/domain.ocn.fv0.9x1.25_gx1v7.151008.nc", will attempt to download in check_input_data phase
```

PR coming shortly.

mnlevy1981:
Test suite:  N/A; I created the two cases that caused problems detailed in issue ticket, verified problem went away
Test baseline: N/A
Test namelist changes: N/A
Test status: bit-for-bit, except cases that previously died at runtime (or that CIME refused to submit) now run to completion

Fixes #1565 

User interface changes?: N/A

Code review: @mvertens, this should fix your f09_f09_g17 issue. Can you verify? And @dabail10 this should fix g17_g17, please verify (I can't add you to official reviewer list)

Additional notes: @jedwards4b it looks like the pull request you reverted (#1559) also fixed #1558 -- I did not put that fix back in here.

### DIFF
--- a/config/cesm/config_grids.xml
+++ b/config/cesm/config_grids.xml
@@ -961,7 +961,7 @@
       <file grid="atm|lnd" mask="gx1v6">domain.lnd.fv0.9x1.25_gx1v6.090309.nc</file>
       <file grid="ocnice"  mask="gx1v6">domain.ocn.0.9x1.25_gx1v6_090403.nc</file>
       <file grid="atm|lnd" mask="gx1v7">domain.lnd.fv0.9x1.25_gx1v7.151020.nc</file>
-      <file grid="ocnice"  mask="gx1v7">domain.ocn.fv0.9x1.25_gx1v7.151008.nc</file>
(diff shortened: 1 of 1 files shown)


mvertens:
I have verified that  f09_f09_mg17 now works for an A compset.

Issue #1576, opened 2017-05-19
In datm, move replace datm mode %S1850 with %CPLHIST


mvertens:
CPLHISTForcing mode can span more than just an 1850 spinup, and this modification to the DATM% attribute to reflect this.

This change will also need the following changes in acme - 
The DATM%S1850 is still present in 
`config/acme/allactive/config_compsets.xml.cime2`
so if that file is still being used, then S1850 -> CPLHIST must be made.

PR #1578, opened 2017-05-20
DATM S1850 mode to CPLHIST

### LINKED ISSUE #1577: invalid quotes in DROF namelist_defaults.xml valid_values
Invalid double quotes in the valid values settings for 
taxmode, `decomp,mapmask,mapalgo`
need to be removed.

mvertens:
Use %CPLHIST for DATM coupler hist forcing

Coupler history forcing should have a %CPLHIST mode for any data component
that uses it. This change makes DATM conform to this convention.

It also fixes the invalid quotes in DROF valid values in #1577

    also verified that the following CESM spinup compset
    `1850_DATM%CPLHIST_SLND_CICE_POP2%ECO_DROF%CPLHIST_SGLC_WW3`
     gave the correct values
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1577 #1576 
User interface changes?: None
Code review:

### DIFF
--- a/src/components/data_comps/datm/cime_config/config_component.xml
+++ b/src/components/data_comps/datm/cime_config/config_component.xml
@@ -30,7 +30,7 @@
       <value compset="%CRU">CLMCRUNCEP</value>
       <value compset="%GSW">CLMGSWP3</value>
       <value compset="%1PT">CLM1PT</value>
-      <value compset="%S1850">CPLHISTForcing</value>
+      <value compset="%CPLHIST">CPLHISTForcing</value>
     </values>
   </entry>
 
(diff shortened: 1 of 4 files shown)


billsacks:
The core change here makes sense to me. But can you please explain these other changes:

1. Why did you change `strm_domdir` to `null` for the cplhist streams?

2. Why did you need to increase `dtlimit` to 3 in some places? My understanding is that values significantly greater than 1 generally indicate an error. If there is a true need for this to be 3, this should be documented for future reference.

mvertens:
1. Because for CPLHIST you want to get the domain from the first stream
file. Its important the the stream domain match the stream data. So when
strm_domdir is null shr_strdata ensures that this is the case. This is also
happening rof the drof CPLHIST mode.

2. By default it was 2.0. For the CPLHIST mode and the first time sample it
needs to be 3 - and we want this to work out of the box. 3 versus 2 is not
significantly different. Its a very large value - e.g. 100 that would be of
concern.

mvertens:
Can you please merge this.

billsacks:
You requested a review from @mnlevy1981 - do you want to wait for his review or should I merge it without that?

mvertens:
Sorry - I meant to delete it. Your feedback is all that's needed. Thanks for the quick response to this PR.

PR #1585, opened 2017-05-22
Updated grid aliases in  cesm/config_grids.xml

alperaltuntas:
Removed grid aliases that do not specify the mask.
Renamed grid aliases so the mask specifier always follow the ocean grid.

Test suite: 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Code review:

### DIFF
--- a/config/cesm/config_grids.xml
+++ b/config/cesm/config_grids.xml
@@ -161,24 +161,10 @@
       <mask>gx3v7</mask>
     </model_grid>
 
-    <model_grid alias="T42_T42" not_compset="_POP">
-      <grid name="atm">T42</grid>
-      <grid name="lnd">T42</grid>
-      <grid name="ocnice">T42</grid>
-      <mask>usgs</mask>
-    </model_grid>
-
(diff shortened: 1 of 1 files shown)


config/cesm/config_grids.xml
```diff
@@ -8,11 +8,11 @@
     =========================================
     The notation for the grid longname is
         a%name_l%name_oi%name_r%name_m%mask_g%name_w%name
-    where
+    where
```
jedwards4b: Please avoid adding unnecessary whitespace to files, it makes the review process difficult.  Most editors can be configured to do this.

config/cesm/config_grids.xml
```diff
@@ -1655,6 +1550,17 @@
       <map name="ROF2OCN_ICE_RMAPNAME">cpl/cpl6/map_r01_to_gx1v6_120711.nc</map>
     </gridmap>
 
+    <!--- river flooding variables -->
```
jedwards4b: These don't belong in this file and were removed in PR #1535

PR #1586, opened 2017-05-22
Update cime to support full X configuration

apcraig:
Update CIME to support full X configuration

Tested with ERS.f19_g16.X.yellowstone_intel only

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #] 1587

User interface changes?: 

Code review:

### DIFF
--- a/config/cesm/config_grids.xml
+++ b/config/cesm/config_grids.xml
@@ -35,14 +35,14 @@
       <grid name="rof"    compset="MOSART">r05</grid>
       <grid name="rof"    compset="DROF"  >rx1</grid>
       <grid name="rof"    compset="DROF%CPLHIST">r05</grid>
+      <grid name="rof"    compset="XROF"  >r05</grid>
       <grid name="glc"	  compset="SGLC"  >null</grid>
       <grid name="glc"	  compset="CISM1" >gland5UM</grid>
       <grid name="glc"	  compset="CISM2" >gland4</grid>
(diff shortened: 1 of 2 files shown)


config/cesm/config_grids.xml
```diff
@@ -41,6 +41,9 @@
       <grid name="wav"	  compset="SWAV"  >null</grid>
       <grid name="wav"	  compset="DWAV"  >ww3a</grid>
       <grid name="wav"	  compset="WW3"	  >ww3a</grid>
+      <grid name="rof"    compset="XROF"  >r05</grid>
```
jedwards4b: Can you please reorder these so that the rof and glc grids are together with the other rof and glc grids.

config/cesm/config_grids.xml
```diff
@@ -1154,7 +1157,7 @@
     </domain>
 
     <domain name="gland4">
-      <nx>376</nx> <ny>701</ny>
+      <nx>416</nx> <ny>704</ny>
```
jedwards4b: i don't think that this PR should have changed this field?

billsacks: Tony was right to change that. It turns out these values had been wrong for a long time but apparently this didn't cause any trouble outside of an X compset.

config/cesm/config_grids.xml
apcraig: I just modified, retested, and committed the change.  The PR is updated.

PR #1591, opened 2017-05-22
Improve comment for BASELINE status in TestStatus

### LINKED ISSUE #810: Test reporting feature request: give baseline tag in a comment in TestStatus
I think this feature was present in cime4, but was dropped in cime5: It's helpful if the baseline tag is given in a comment in TestStatus, so rather than just:

```
  PASS SMS_Ld5.f19_g16.IRCP45CLM45BGC.yellowstone_intel.clm-decStart BASELINE
```

we'd have something like:

```
  PASS SMS_Ld5.f19_g16.IRCP45CLM45BGC.yellowstone_intel.clm-decStart BASELINE clm4_5_12_r203
```

and same for FAILs

The same should be done for GENERATE.

jgfouca:
Adds the baseline name

Test suite: code-checker + by-hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #810 

User interface changes?: TestStatus changes

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -353,9 +353,10 @@ class SystemTestsCommon(object):
             success, comments = compare_baseline(self._case)
             append_testlog(comments)
             status = TEST_PASS_STATUS if success else TEST_FAIL_STATUS
-            ts_comments = comments if "\n" not in comments else None
+            baseline_name = self._case.get_value("BASECMP_CASE")
+            ts_comments = comments if "\n" not in comments else os.path.dirname(baseline_name)
             self._test_status.set_status(BASELINE_PHASE, status, comments=ts_comments)
-            basecmp_dir = os.path.join(self._case.get_value("BASELINE_ROOT"), self._case.get_value("BASECMP_CASE"))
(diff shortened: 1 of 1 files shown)


scripts/lib/CIME/SystemTests/system_tests_common.py
```diff
@@ -353,9 +353,10 @@ def _compare_baseline(self):
             success, comments = compare_baseline(self._case)
             append_testlog(comments)
             status = TEST_PASS_STATUS if success else TEST_FAIL_STATUS
-            ts_comments = comments if "\n" not in comments else None
+            baseline_name = self._case.get_value("BASECMP_CASE")
+            ts_comments = comments if "\n" not in comments else os.path.dirname(baseline_name)
```
billsacks: If there are comments, would it work to print the baseline_name in addition to the other comments - e.g., as `"%s %s"%(os.path.dirname(baseline_name), comments)`? That way we'd be guaranteed to see the baseline name, right?

jgfouca: Good point.

jgfouca:
@billsacks ok, what do you think now?

billsacks:
Looks good now -thanks.

PR #1593, opened 2017-05-23
Xmlquery and config compset

### LINKED ISSUE #946: Clean up of xmlquery usage
This request is partly about documentation and partly about what options are supported by xmlquery.

Looking at the new xmlquery (after #930), I see two classes of things that could be improved in the usage information:

1. I feel that there are more options than there need to be here: I don't see the need for the separate `--description`, `--type` and `--valid-values` options: I'd generally just want to run `--full` if I were interested in seeing that extra information. Also, the `--group` option appears not to work, though I don't see the need for that one either. I get overwhelmed trying to figure out how to use a tool when there are so many options. But I'm fine being overruled if others like having this flexibility.

2. The documentation (`-h`) could be improved:

    1. I get the impression that there are multiple possible uses: ./xmlquery -listall isn't compatible with some of the other options. This should be stated explicitly in the top usage message - i.e., it should give the different possible uses (similarly to what you get with, e.g., 'git help checkout')

    2. Important options should be listed first - as opposed to listing the deprecated '-file' first

    3. Need more information for some options like subgroup and no-resolve: I feel like these descriptions only make sense for someone who is intimately familiar with the system

    4. Is the documentation for 'caseroot' really correct? ("Case directory to build")

cc @jedwards4b @mvertens @bertinia

jedwards4b:
FIx an issue with xmlquery when the --file and --listall flags are used together, Update config_compsets.xml to remove grid attribute which isn't used.  Some minor updates for
cheyenne modules and omplace syntax.

Test suite: scripts_regression_tests.py, hand tests of xmlquery 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #946 

User interface changes?: 

Code review:

### DIFF
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -216,7 +216,7 @@
 	<arg name="anum_tasks"> -np $TOTALPES</arg>
 	<arg name="labelstdout">-p "%g:"</arg>
 	<!-- the omplace argument needs to be last -->
-	<arg name="zthreadplacement"> omplace </arg>
+	<arg name="zthreadplacement"> omplace -tm open64 </arg>
       </arguments>
     </mpirun>
     <mpirun mpilib="default" unit_testing="true">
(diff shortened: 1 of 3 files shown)


billsacks:
This doesn't completely address #946 - it just addresses the most recent comments there. I will merge it, but will keep #946 open.

PR #1594, opened 2017-05-23
Add some maps to config_grids.xml

### LINKED ISSUE #1539: Add f19_g17 resolution.
We are going to need 2-degree FV mapping files to gx1v7 very soon. I think for the tutorial we will be running f19_g37. However, we need some help on this.

Dave

mnlevy1981:
Added the f19_g17 maps, as wall as the ww3 <-> tx1v1 maps; all files are in inputdata as well.

Test suite: none, just some stand-alone checks
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit, plus resolutions that didn't run previously now run

Fixes #1539

User interface changes?: N/A

Code review: @dabail10 can you please look at this? (And maybe see about getting added to ESMCI so I can list you as an official reviewer?)

### DIFF
--- a/config/cesm/config_grids.xml
+++ b/config/cesm/config_grids.xml
@@ -1267,6 +1267,13 @@
       <map name="OCN2ATM_FMAPNAME">cpl/gridmaps/gx1v6/map_gx1v6_TO_fv1.9x2.5_aave.130322.nc</map>
       <map name="OCN2ATM_SMAPNAME">cpl/gridmaps/gx1v6/map_gx1v6_TO_fv1.9x2.5_aave.130322.nc</map>
     </gridmap>
+    <gridmap atm_grid="1.9x2.5" ocn_grid="gx1v7">
+      <map name="ATM2OCN_FMAPNAME">cpl/gridmaps/fv1.9x2.5/map_fv1.9x2.5_TO_gx1v7_aave.170518.nc</map>
+      <map name="ATM2OCN_SMAPNAME">cpl/gridmaps/fv1.9x2.5/map_fv1.9x2.5_TO_gx1v7_blin.170518.nc</map>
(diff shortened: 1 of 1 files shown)


mnlevy1981:
@dabail10 -- good catch! Updated that section as well

dabail10:
Also, you said that the domain.ocn files were slightly different from the earlier ones. Should we be using these?

mnlevy1981:
I think we want to stick with `domain.ocn.gx1v7.151008.nc` for the f19_g17 resolution, and this last commit should provide `domain.ocn.fv1.9x2.5_gx1v7.170518.nc` for cases with the f19_f19 resolution. I wasn't entirely clear on which compset / resolution pairs to create cases out of for testing purposes, though, so if you have an example of something that is still missing a map or domain file let me know and I'll make sure it gets updated.

mnlevy1981:
Sorry, that last comment should've specified f19_f19_mg17 instead of just f19_f19!

dabail10:
The domain.ocn.gx1v7 file is used for more than f19_g17 isn't it? Isn't for all DOCN/POP cases using the gx1v7 grid?

dabail10:
I'm fine with using the older one, but I just wondered if we should be consistent with all of the domains and maps.

mnlevy1981:
I don't think there's a good way to be consistent with all the domains and maps; the `domain.ocn.gx1v7` file is used by several different resolutions, and the other domain files used by some of those resolutions came from yellowstone while other resolutions use files from cheyenne. Unless we get inconsistency errors, which we shouldn't, differences are O(1e-8), I think we should stick with what we have.

jedwards4b:
@mnlevy1981 you should not assign a PR to yourself.   If this is ready assign to Chris or I and we will merge.

mnlevy1981:
I'll assign you both and see who gets there first :)

PR #1598, opened 2017-05-24
Don't ignore attributes for Compilers.get_value

### LINKED ISSUE #1387: compilers.py get_value doesn't use attributes argument
I noticed that class Compilers in compilers.py has a get_value method that doesn't use the attributes argument. Presumably it should?

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1387 

User interface changes?: None

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/XML/compilers.py
+++ b/scripts/lib/CIME/XML/compilers.py
@@ -122,7 +122,7 @@ class Compilers(GenericXML):
         expect(subgroup is None, "This class does not support subgroups")
         value = None
 
-        node = self.get_optional_compiler_node(name)
+        node = self.get_optional_compiler_node(name, attributes=attribute)
         if node is not None:
             value = node.text
 


scripts/lib/CIME/XML/compilers.py
```diff
@@ -122,7 +122,7 @@ def get_value(self, name, attribute=None, resolved=True, subgroup=None):
         expect(subgroup is None, "This class does not support subgroups")
         value = None
 
-        node = self.get_optional_compiler_node(name)
+        node = self.get_optional_compiler_node(name, attributes=attribute)
```
jgfouca: @jedwards4b would you take a close look at this? This seems like it could potentially have an impact on our system's behavior. I'm wondering if the ignoring of attributes was deliberate or just an oversight.

PR #1616, opened 2017-05-26
Fix race condition with case XML files when using set_file

jgfouca:
... and some other unrelated bugs.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: None

Code review: @jedwards4b


scripts/lib/CIME/SystemTests/system_tests_common.py
```diff
@@ -506,6 +506,9 @@ def build_phase(self, sharedlib_only=False, model_only=False):
         super(TESTTESTDIFF, self).build_phase(sharedlib_only=sharedlib_only,
                                               model_only=model_only)
 
+    def run_indv(self, suffix=None, st_archive=False ):
```
jgfouca: @jedwards4b turns out we need this override because it uses different defaults.

PR #1618, opened 2017-05-27
Add more error-checking with gnu in debug mode

### LINKED ISSUE #857: Stricter DEBUG checks when using GCC
_From @quantheory on August 4, 2015 17:5_

This is an extension of the issue here:

http://bugs.cgd.ucar.edu/show_bug.cgi?id=1773

Right now we really don't have any of the major run-time checks turned on for GNU compilers. We should probably use an option such as `-fcheck=all`. (This prints unnecessary creation of array temporary warnings, but unlike Intel's equivalent option, it should only print once per code location, so it shouldn't produce as much spam.)

_Copied from original issue: CESM-Development/cime#91_

### LINKED ISSUE #1615: Add stricter debug flags for gnu
I realized that my fix for #857 in #1339 only added flags for the C compiler, not the Fortran compiler. I want to add the same flags for the Fortran compilation, assuming this doesn't break our tests.

billsacks:
This adds more debug checks with gnu.

I thought this was done in #1339 (which claimed to fix #857), but I
realized that that only added flags for the C compiler, not the Fortran
compiler. This PR adds debug flags for the Fortran compiler, too.

Ideally, we would also have 'invalid' in the ffpe-trap list. But at
least with some versions of gfortran (confirmed with 5.3.0), gfortran's
isnan (which is called in cime via the CPRGNU-specific shr_infnan_isnan)
causes a floating point exception when called on a signaling NaN. (This
problem showed up in the unit tests on my Mac: Adding 'invalid' caused
the shr_strconvert unit tests to fail (specifically, the tests
toString_prints_snan_r4 and toString_prints_snan_r8).)

Test suite: Ran nearly all gnu debug tests that I could find in all CESM
test suites. This included:
- aux_clm45 yellowstone-gnu: all pass and are bit-for-bit
- The following yellowstone tests:

  ERS_D_Ly3.f09_g16_gl5.TG1.yellowstone_gnu.cism-noevolve
  SMS_D.T31_g37_gl20.IGCLM45.yellowstone_gnu.cism-test_coupling
  SMS_D.f09_g16_gl20.TG1.yellowstone_gnu
  SMS_D_Ly1.f09_g16_gl20.THISTG.yellowstone_gnu
  SMS_D_Ld5.f10_f10.IMCRUCLM50BGC.yellowstone_gnu.mosart-default
  SMS_D_Ld5.f10_f10.ICRUCLM50BGC.yellowstone_gnu.rtm-default
  SMS_D.f09_g16_gl20.TG1.yellowstone_gnu
  ERI_D.T31_g37_rx1.A.yellowstone_gnu
  SMS_D_Ld1.5x5_amazon.ITEST.yellowstone_gnu.clm-40default

  All pass except a build failure for
  SMS_D_Ld1.5x5_amazon.ITEST.yellowstone_gnu.clm-40default which doesn't
  look related to these new debug flags. This is a CLM40 test that isn't
  important.

- The following cheyenne tests:

  SMS_D.f09_g16_gl20.TG1.cheyenne_gnu
  ERI_D.T31_g37_rx1.A.cheyenne_gnu
  SMS_D_Ld1.5x5_amazon.ITEST.cheyenne_gnu.clm-40default

  The A test passed. The ITEST failed as for yellowstone. The TG1 test
  failed, unable to link some lapack routines (seems unrelated to the
  new build flags).

I ran these tests from
https://svn-ccsm-models.cgd.ucar.edu/clm2/branches/externals_update_cesm2_0_alpha06k
at r85267, with cime changes rebased onto cime5.3.0-alpha.13

Test baseline: For aux_clm45 tests, I compared against the above version
   of CLM unmodified. For other tests, I didn't do baseline comparisons
Test namelist changes: none
Test status: bit for bit

Fixes #1615

User interface changes?: none

Code review:

### DIFF
--- a/config/cesm/machines/config_compilers.xml
+++ b/config/cesm/machines/config_compilers.xml
@@ -128,7 +128,12 @@ using a fortran linker.
        so that these are passed to cmake builds (cmake builds don't use FIXEDFLAGS and FREEFLAGS). -->
     <base> -mcmodel=medium -fconvert=big-endian -ffree-line-length-none -ffixed-line-length-none </base>
     <append compile_threaded="true"> -fopenmp </append>
-    <append DEBUG="TRUE"> -g -Wall </append>
+    <!-- Ideally, we would also have 'invalid' in the ffpe-trap list. But at
+         least with some versions of gfortran (confirmed with 5.3.0), gfortran's
+         isnan (which is called in cime via the CPRGNU-specific
+         shr_infnan_isnan) causes a floating point exception when called on a
+         signaling NaN. -->
+    <append DEBUG="TRUE"> -g -Wall -ffpe-trap=zero,overflow -fcheck=bounds </append>
     <append DEBUG="FALSE"> -O </append>
   </FFLAGS>
   <FFLAGS_NOOPT>


billsacks:
@jtruesdal @mnlevy1981 @alperaltuntas @klindsay28 - I also found the following POP gnu debug tests in various test suites: I have **not** run these:

SMS_D.T62_s11.G.cheyenne_gnu.pop-cice : I can't tell if this is expected to pass: it looks like the last time it was run was alpha06c, when it failed on yellowstone and cheyenne as far as I can tell

aux_pop_obgc_se: PET_D_Ld2.T62_g37.C1DECO.userdefined_gnu.pop-ecosys_restore_gx3v7
aux_pop_obgc_se: ERS_D_Ld3.T62_g37.CECO.userdefined_gnu.pop-ciso_abio_dic_dic14
aux_pop_obgc_se: PET_D_Ld2.T62_g37.CECO.userdefined_gnu.pop-ecosys_restore_gx3v7
aux_pop_obgc_se: SMS_D_Ld2.T62_g37.CECO.userdefined_gnu.pop-ecosys

(I wasn't sure where the above are typically run, and if they are expected to pass.)

You may want to run these to make sure they still work with these new flags.

billsacks:
@rljacob - you probably will want these stricter debug checks in ACME, too. But I only added them in CESM so I wouldn't break your builds.

mnlevy1981:
The `aux_pop_obgc_se` testlist is no longer being used (it was originally used when we could test some MARBL compsets on a laptop), and should probably be removed. The `T62_s11` test likely failed because we din't make wav->ocn maps for the s11 grid until this past week... so hopefully everything will pass it the newest alpha tests.

As part of the MARBL development, I do run the `cheyenne_intel` aux_pop_MARBL tests with both intel and gnu, so we have some successful gnu debug tests in the development stream.

mnlevy1981:
So thanks for the heads-up about the stricter checks, if tests start to fail I should [hopefully] be easy to trace.

rljacob:
We have added more flags.  We're currently using:  -g -Wall -Og -fbacktrace -fcheck=bounds -ffpe-trap=invalid,zero,overflow

billsacks:
@rljacob - thanks. I missed that since it's in ACME but not yet in cime master.

PR #1623, opened 2017-05-30
step 1 in creating orphan branch and gh-pages

bertinia:
copied doc dir into master in order to create a gh-pages orphan branch


jhkennedy:
@bertinia I see the `build/html` directory and sub files are still in there. You should remove them so they aren't tracked in this branch, and only exist in the orphaned `gh-pages` branch.

PR #1666, opened 2017-06-09
add a test to assure flush is not triggered by instantiation of case

### LINKED ISSUE #1628: CIME is highly prone to race conditions
... in the various env xml files.

A recent minor change on the ACME side exposed the races and took many hours to sort out due to their intermittent nature.

Calls to case.set_value that occur outside of case_setup are very risky.

Other risky calls are things like restore, which swaps out the underlying XML file but doesn't necessarily update the in-memory values in the case object. Maybe some kind of tracking of timestamps or checksums of the underlying XML files should be done in the case object to ensure things are in sync.

jedwards4b:
Add a test to scripts_regression_tests.py to assure that instantiating a case does 
not invoke a flush command.   

Test suite: scripts_regression_tests.py J_TestCreateNewcase 
(I also ran a case with a modified case.py in which I added a
case.set_value call to initilize_derived_attributes to make sure the test failed as expected.  it did. )

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1628

User interface changes?: 

Code review:

### DIFF
--- a/scripts/tests/scripts_regression_tests.py
+++ b/scripts/tests/scripts_regression_tests.py
@@ -262,6 +262,15 @@ class J_TestCreateNewcase(unittest.TestCase):
 
         cls._do_teardown.append(testdir)
 
+    def test_aa_no_flush_on_instantiate(self):
+        testdir = os.path.join(self.__class__._testroot, 'testcreatenewcase')
+        with Case(testdir, read_only=False) as case:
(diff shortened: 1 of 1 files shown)


jgfouca:
@jedwards4b thanks, this is a good first step. I'm still worried about other types of race conditions, but this would have caught the one that caused me so much grief a couple weeks ago.

Issue #1665, opened 2017-06-09
Do we still want the `--pesfile` argument to create_newcase?


billsacks:
Now that `--pesfile` is no longer a required argument when you're running with a user-compset (thanks, @jedwards4b !): Do we still want this argument for create_newcase?

If we do keep it, the help text needs to be fixed. Currently it says:

```
  --pesfile PESFILE     Only used and required for --user-compset
                        argument.Full pathname of the pes specification
                        fileThis argument is required if --user-compset is
                        TrueThe file can follow either the config_pes.xml or
                        the env_mach_pes.xml format (default: None)
```

This is incorrect on two points:

(1) It can be used whether or not you're using a `--user-compset`

(2) It is **not** required any more.

jedwards4b:
I think that we do want to keep it with documentation:

Full pathname of an optional pes specification
file.  The file can follow either the config_pes.xml or
the env_mach_pes.xml format (default: None)

billsacks:
Okay, I'll add that in an upcoming PR.

Issue #1664, opened 2017-06-09
Pes specification file wrong in README.case for case with a compset alias and --pesfile


billsacks:
If you create a case with a compset alias using the `--pesfile` argument, README.case says:

```
2017-06-08 20:42:01: Pes     specification file is unset
```

rather than pointing to the `--pesfile` you gave.

I'm fixing this in an upcoming PR.

Issue #1658, opened 2017-06-08
--user-mods-dir not picked up for a user-compset


billsacks:
If you specify a --user-mods-dir that is relative to the primary component's cime_config/usermods_dirs area, it is not picked up when using a user-defined compset (i.e., a compset without an alias).

For example, I added the following:

```
--- src/drivers/mct/cime_config/usermods_dirs/wjs_test/shell_commands ---
./xmlchange HIST_OPTION=nyears,HIST_N=3
```

This is picked up if I do this:

```
./create_newcase --case usermods_0608a --compset A --res f45_g37_rx1 --user-mods-dir wjs_test
```

But not if I do this (note that I am using a bare "DATM" to force use of a user-compset):

```
./create_newcase --case usermods_usercompset_0608a --compset 2000_DATM_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV --res f45_g37_rx1 --user-mods-dir wjs_test --user-compset
```

This gives the error:

```
ERROR: Expected full directory path, got 'unset/wjs_test'
```

billsacks:
I think the problem is that some stuff in `Case._set_compset_and_pesfile` is only done for a compset with an alias. This causes the above problem, and could cause other problems as well.

The solution I see is to rework this code slightly: Always set primary component in the above routine (setting it earlier than is currently done for a user-defined compset); then move settings of the user_mods_dir and others to after the primary component is set. I'll take a stab at this.

PR #1667, opened 2017-06-09
User compset fixes

billsacks:
Main change is to have user-defined compsets still pick up a usermods
directory under [primary_component]/cime_config/usermods_dirs/. Along
with this, I have refactored where some other things are set that depend
on primary component. A key aim here was to have various things set in
the same place regardless of whether we're using a compset with or
without an alias. I find this makes it easier to understand the code,
and less likely to introduce discrepancies between user-defined and
alias-defined compsets in the future.

Also, manual testing:
- Compset with an alias: picks up PE settings and a usermods directory
- User compset picks up PE settings
- User compset picks up usermods directory, as in:
  ./create_newcase --case no_alias_0608d --compset 1850_DATM%GSWP3v1_CLM50%BGC-CROP_SICE_SOCN_RTM_CISM2%NOEVOLVE_SWAV --res f09_g16 --user-compset --user-mods-dir clm/clm5_OutputCrop
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#1658
Fixes ESMCI/cime#1664
Fixes ESMCI/cime#1665

User interface changes?: none

Code review:

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -410,21 +410,18 @@ class Case(object):
             if result is not None:
                 del self.lookups[key]
 
-    def _set_compset_and_pesfile(self, compset_name, files, user_compset=False, pesfile=None):
+    def _set_compset_and_primary_component(self, compset_name, files, user_compset=False):
         """
         Loop through all the compset files and find the compset
         specifation file that matches either the input 'compset_name'.
         Note that the input compset name (i.e. compset_name) can be
-        either a longname or an alias.  This will also set the
-        compsets and pes specfication files.
+        either a longname or an alias. This will set various compset-related
+        info and self._primary_component.
         """
         science_support = []
         compset_alias = None
         components = files.get_components("COMPSETS_SPEC_FILE")
         logger.debug(" Possible components for COMPSETS_SPEC_FILE are {}".format(components))
-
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/case.py
```diff
@@ -522,6 +513,33 @@ def _find_primary_component(self):
...
+        component = self._primary_component
+
+        if pesfile is None:
+            self._pesfile = files.get_value("PES_SPEC_FILE", {"component":component})
+            pesfile_unresolved = files.get_value("PES_SPEC_FILE", {"component":component}, resolved=False)
+            logger.info("Pes     specification file is {}".format(self._pesfile))
+        else:
+            self._pesfile = pesfile
+            pesfile_unresolved = pesfile
+        self.set_lookup_value("PES_SPEC_FILE", pesfile_unresolved)
```
billsacks: There are some subtle behavior changes here compared with the previous code:

1. Previously, this wasn't being done for a compset with an alias when a pesfile was specified explicitly (causing issue #1664 )

2. Previously, for a compset without an alias without an explicitly specified pesfile, this was set with resolved = True (implicitly); now it's set with resolved = False for consistency with what happens when there is an alias. I don't think this has any important impact, but I wanted to mention it in case it's important.

PR #1670, opened 2017-06-09
fixed description for CCSM_BGC

### LINKED ISSUE #1650: Need better description of CCSM_BGC and related namelist variables
The description of CCSM_BGC in src/drivers/mct/cime_config/config_component_cesm.xml and config_component_acme.xml needs to be improved: I can't understand what it does based on this description:

```xml
    <desc>Flag to turn on new fields in coupling.
    If the value is not none, the coupler is compiled so that optional
    BGC related fields are exchanged between component models.
    </desc>
```

I think it's wrong to say "the coupler is compiled...", because this is now a runtime option. More importantly, the meaning of each value (CO2A, CO2B, etc.) should be given here.

The descriptions of related namelist variables in namelist_definition_drv.xml should similarly be improved - e.g.

```xml
    <desc>
      Previously, new fields that were needed to be passed between components
      for certain compsets were specified by cpp-variables. This has been
      modified to now be use cases. This use cases are specified in the
      namelist cpl_flds_inparm and are currently triggered by the xml variable CCSM_BGC.
      If CCSM_BGC is set to 'CO2A', then flds_co2a will be set to .true.
    </desc>
```

I think most of this description (all but the last line) is wrong. And the last line tells me how this is set, but not what it does.

cc @mvertens @klindsay28 @ekluzek

mvertens:
Provides better description of CCSM_BGC xml variable for both ACME and CESM

Also removed the CO2_DMSA valid value for CESM since this is not supported by CESM.

Test suite: NA
Test baseline: NA 
Test namelist changes: NA
Test status: NA
Fixes #1650 
User interface changes?: NA
Code review: none

### DIFF
--- a/src/drivers/mct/cime_config/namelist_definition_drv.xml
+++ b/src/drivers/mct/cime_config/namelist_definition_drv.xml
@@ -49,11 +49,9 @@
     <category>seq_flds</category>
     <group>seq_cplflds_inparm</group>
     <desc>
-      Previously, new fields that were needed to be passed between components
-      for certain compsets were specified by cpp-variables. This has been
-      modified to now be use cases. This use cases are specified in the
-      namelist cpl_flds_inparm and are currently triggered by the xml variable CCSM_BGC.
-      If CCSM_BGC is set to 'CO2A', then flds_co2a will be set to .true.
+      If set to .true., adds prognostic CO2 and diagnostic CO2 at the lowest
+      model level to be sent from the atmosphere to the land and ocean.
+      If CCSM_BGC is set to 'CO2A', then flds_co2a will be set to .true. by default
     </desc>
     <values>
       <value>.false.</value>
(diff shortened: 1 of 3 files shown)


src/drivers/mct/cime_config/namelist_definition_drv.xml
```diff
@@ -117,12 +111,7 @@
     <category>seq_flds</category>
     <group>seq_cplflds_inparm</group>
     <desc>
-      Previously, new fields that were needed to be passed between components
-      for certain compsets were specified by cpp-variables. This has been
-      modified to now be use cases. This use cases are specified in the
-      namelist cpl_flds_inparm and are currently triggered by the xml variable ACME_BGC.
-      This is a new master switch for turning ACME BGC off and on, just for testing.
-      If ACME_BGC is set to 'TRUE', then flds_bgc will be set to .true.
```
billsacks: I removed this incorrect statement

mvertens:
@billsacks - this is fine. Thanks for the quick review and for catching the other problem. Please go ahead and merge.

Issue #1673, opened 2017-06-09
Top-level cime_config/usermods_dirs incorrectly added to list of all_user_mods


billsacks:
If a component defines a COMPNAME_USER_MODS (e.g., CAM_USER_MODS) entry in config_component.xml, but the value for the current case ends up being blank, we are getting incorrect behavior: The expected behavior is that no user mods directory would be included; the actual behavior is that it tries to pull in user mods from PRIMARY_COMPONENT/cime_config/usermods_dirs . This often will be harmless, because that directory won't contain any user_nl files or others (it typically just contains some subdirectories with the actual user mods). But it's confusing, and it could give wrong behavior if that directory happened to include any files that matched names that are expected in a user_mods directory (e.g., user_nl files).

I have seen this in the latest cesm2_0_alpha06 sandbox, with cime at master, using:

```
./create_newcase --case test_case_configure_0609h --res f09_g16 --mach yellowstone --project P93300601 --compset B1850
```

This prints:

```
This component includes user_mods
Adding user mods directory /glade2/scratch2/sacks/cesm_code/cesm2_0_alpha06_newCime/cime_config/usermods_dirs
```

As far as I can tell, this compset isn't supposed to be pulling in any user mods, at least based on this section in CAM's config_component.xml:

```xml
  <entry id="CAM_USER_MODS">
    <type>char</type>
    <valid_values></valid_values>
    <default_value></default_value>
    <values>
    <value compset="DOCN%AQP3">$SRCROOT/components/cam/cime_config/usermods_dirs/aquap</value>
    <value compset="DOCN%SOMAQP">$SRCROOT/components/cam/cime_config/usermods_dirs/aquap</value>
    <value compset="HIST_CAM60_CLM50%BGC">$SRCROOT/components/cam/cime_config/usermods_dirs/fhist</value>
    <value compset="2000_CAM60_CLM50%BGC">$SRCROOT/components/cam/cime_config/usermods_dirs/f2000</value>
    </values>
    <group>run_component_cam</group>
    <file>env_case.xml</file>
    <desc>User mods to apply to specific compset matches.   These usermods will be used only if
                cam is the principle component in the compset (F or Q compsets) </desc>
  </entry>
```

From tracing this in a debugger, I think the problem arises from case.py: apply_user_mods . I don't completely understand this, but my guess is that in this:

```python
            comp_user_mods = self.get_value("{}_USER_MODS".format(component.upper()))
            if comp_user_mods is not None:
[...]

PR #1675, opened 2017-06-11
Do not add empty user mods

billsacks:
Ignore empty COMP_USER_MODS values

If a component defined a COMPNAME_USER_MODS (e.g., CAM_USER_MODS) entry
in config_component.xml, but the value for the current case ends up
being blank, we were getting incorrect behavior: The expected behavior
is that no user mods directory would be included; the actual behavior
was that it tries to pull in user mods from
PRIMARY_COMPONENT/cime_config/usermods_dirs . This often will be
harmless, because that directory won't contain any user_nl files or
others (it typically just contains some subdirectories with the actual
user mods). But it's confusing, and it could give wrong behavior if that
directory happened to include any files that matched names that are
expected in a user_mods directory (e.g., user_nl files).

This PR fixes this behavior, so that if COMPNAME_USER_MODS evaluates to
an empty string, then we don't try to include it.

Also, manual tests of create_newcase:
- B1850 (no user_mods)
- BHIST (picks up user_mods from CAM)
- FHIST_DEV (picks up user_mods from CAM, the primary_component)

Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#1673 

User interface changes?: none

Code review:

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -1115,10 +1115,11 @@ class Case(object):
         Returns None if no value was found, or if the value is an empty string.
         """
         comp_user_mods = self.get_value("{}_USER_MODS".format(component.upper()))
-        if comp_user_mods and not comp_user_mods.isspace():
-            return comp_user_mods
-        else:
+        if (comp_user_mods is None or comp_user_mods == "" or
+            (isinstance(comp_user_mods, basestring) and comp_user_mods.isspace())):
             return None
+        else:
+            return comp_user_mods
 
     def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None, cime_output_root=None):
         if cime_output_root is None:
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -1115,11 +1115,17 @@ class Case(object):
         Returns None if no value was found, or if the value is an empty string.
         """
         comp_user_mods = self.get_value("{}_USER_MODS".format(component.upper()))
-        if (comp_user_mods is None or comp_user_mods == "" or
-            (isinstance(comp_user_mods, basestring) and comp_user_mods.isspace())):
+        if comp_user_mods is None or comp_user_mods == "":
             return None
         else:
-            return comp_user_mods
+            try:
+                is_space = comp_user_mods.isspace()
+            except AttributeError:
+                is_space = False
+            if is_space:
+                return None
+            else:
+                return comp_user_mods
 
     def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None, cime_output_root=None):
         if cime_output_root is None:
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -1069,7 +1069,7 @@ class Case(object):
             append_status("{} is {}".format(comp_grid,self.get_value(comp_grid)),
                           "README.case", caseroot=self._caseroot)
             comp = str(self.get_value("COMP_{}".format(component_class)))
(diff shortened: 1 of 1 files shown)


billsacks:
cc @mvertens

scripts/lib/CIME/case.py
```diff
@@ -1108,6 +1108,25 @@ def apply_user_mods(self, user_mods_dir=None):
...
+        """
+        comp_user_mods = self.get_value("{}_USER_MODS".format(component.upper()))
+        if comp_user_mods is None or comp_user_mods == "":
+            return None
+        else:
+            try:
+                is_space = comp_user_mods.isspace()
+            except AttributeError:
+                is_space = False
+            if is_space:
```
billsacks: Checking is_space isn't necessary right now, but is necessary to avoid cryptic error messages if someone had something like <default_value> </default_value>  (note the space)

scripts/lib/CIME/case.py
```diff
@@ -1108,6 +1108,25 @@ def apply_user_mods(self, user_mods_dir=None):
...
+        """
+        For a component 'foo', gets the value of FOO_USER_MODS.
+
+        Returns None if no value was found, or if the value is an empty string.
+        """
+        comp_user_mods = self.get_value("{}_USER_MODS".format(component.upper()))
+        if comp_user_mods is None or comp_user_mods == "":
+            return None
+        else:
+            try:
```
billsacks: After much trial and error, this try/except block was the only way I could get pylint 1.6.4 (on yellowstone) to accept the use of isspace.

scripts/lib/CIME/case.py
```diff
@@ -1108,6 +1108,25 @@ def apply_user_mods(self, user_mods_dir=None):
...
             apply_user_mods(self._caseroot, user_mods_path)
 
+    def _get_comp_user_mods(self, component):
+        """
+        For a component 'foo', gets the value of FOO_USER_MODS.
+
+        Returns None if no value was found, or if the value is an empty string.
+        """
+        comp_user_mods = self.get_value("{}_USER_MODS".format(component.upper()))
+        if comp_user_mods is None or comp_user_mods == "":
```
billsacks: Ideally it seems like we'd have some shared handling of empty strings in some common place, but I'm not sure where this should go (e.g., one of the get_value routines somewhere in the inheritance chain?), and I didn't want to risk breaking things this late in the release cycle, so for now I'm just applying this special handling for this one case where I know we need it.

scripts/lib/CIME/case.py
```diff
@@ -1108,6 +1108,19 @@ def apply_user_mods(self, user_mods_dir=None):
...
             apply_user_mods(self._caseroot, user_mods_path)
 
+    def _get_comp_user_mods(self, component):
+        """
+        For a component 'foo', gets the value of FOO_USER_MODS.
+
+        Returns None if no value was found, or if the value is an empty string.
+        """
+        comp_user_mods = self.get_value("{}_USER_MODS".format(component.upper()))
+        #pylint: disable=no-member
```
jedwards4b: I think that this is cleaner than the exception code - do you agree?

billsacks: This is definitely cleaner. But I was hesitant to bypass this pylint check, figuring that pylint might be seeing some real possibility for problems. In particular: If there is any possibility that comp_user_mods is **not** a string here, then this code will raise an exception. I'll defer to your understanding here: If you're confident that get_value will **always** return a string here, then I'm happy with your change.

jedwards4b: It should always return a string and if it doesn't it should be an exception.

PR #1684, opened 2017-06-13
fixing xmlquery issues

### LINKED ISSUE #946: Clean up of xmlquery usage

mvertens:
Fixes to xmlquery and extensive new documentation

This PR provides fixes #946 
Particularly  the arguments  --listall --full now work and --groups has been replaced by --get-groups and is now functional.

Test baseline: NA
Test namelist changes: NA
Test status: NA

Fixes #946 
User interface changes?: xmlquery --group has been changed go xmlquery --get-group
Code review: jedwards

### DIFF
--- a/scripts/Tools/xmlquery
+++ b/scripts/Tools/xmlquery
@@ -81,15 +81,15 @@ epilog=textwrap.dedent(__doc__))
                         help="Do not resolve variable values")
 
     parser.add_argument("-p", "--partial-match", action="store_true",
-                        help="Allow partial matches of variable names, treats args as regex. ")
+                        help="Allow partial matches of variable names, treats args as regex.")
 
     group = parser.add_mutually_exclusive_group()
 
     group.add_argument("--full", default=False, action="store_true",
-                        help="Print a full listing for a variable, or a list of variables ")
+                        help="Print a full listing for a variable, or a list of variables")
 
     group.add_argument("-fileonly", "--fileonly", default=False, action="store_true",
-                        help="Only print the filename that the field is defined in. ")
+                        help="Only print the filename that the field is defined in.")
 
     group.add_argument("-value", "--value", default=False, action="store_true",
                         help="Only print one value without newline character. "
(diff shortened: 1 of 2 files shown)


billsacks:
I just pushed a commit where I fixed some typos and added an example at the top of the Usage of querying a single variable (since this is the most common use).

scripts/Tools/xmlquery
```diff
@@ -127,17 +168,17 @@ def get_value_as_string(case, var, attribute=None, resolved=False, subgroup=None
...
+                 full=False, dtype=False, valid_values=False, xmlfile=None):
     """
     Return list of attributes and their values, print formatted
 
     """
     results = {}
     comp_classes = case.get_values("COMP_CLASSES")
-    if xmlfile:
-        case.set_file(xmlfile)
+
```
jedwards4b: I think that this was needed for the --file option.   Please clearly document the case that failed because this was here.   As an example I created a test with the setting of REST_OPTION='nsteps' in env_run.xml and 'none' in env_test.xml.    With this change I cannot view the value in env_test.xml, however I found a bug in master that causes an error with the flag, I'm pushing a change to your branch.

jedwards4b:
I readded the case.set_file(xmlfile) line (with a fix, it was broken on master).   For an example of where this is needed create a test and change a value in env_test.xml then try to query that new value.  If you don't have this line it will not look in this file.

billsacks:
Given the fix from @jedwards4b , should `--file` also be listed in the list of possible options to `--listonly`?

mvertens:
@billsacks - thank you for all of your thoughtful comments. I think I have resolved all of the issues you raised. Please have a look.

billsacks:
This is great now - thank you! I find this help text to be extremely helpful now!

I pushed a commit with fixes for some minor typos. I think this is good to go now.

Issue #1687, opened 2017-06-13
OCN_NCPL logic doesn't handle CLM5 correctly


billsacks:
src/drivers/mct/cime_config/config_component_cesm.xml has this (copied twice):

```xml
<value compset="_DATM.*_CLM4.*_SICE.*_SOCN">1</value>
```

This doesn't match CLM5. We should remove the 4 (and remove the duplicate line).

PR #1688, opened 2017-06-14
Apply OCN_NCPL setting to any version of CLM

billsacks:
Apply OCN_NCPL setting to any version of CLM

Also remove a duplicate line

Test suite: Ran a few tests in a CLM sandbox: I compsets with clm45 and clm50
before and after this change. Verified that, with this change,
OCN_NCPL is 1 as desired for both clm45 and clm50.
Test baseline: N/A
Test namelist changes: Changes ocn_cpl_dt in drv_in for I compsets with clm50 
Test status: bit for bit
I haven't tested for bit-for-bit, but this should be bit-for-bit, right?
i.e., answers shouldn't change based on the coupling frequency of a stub model, right?

Fixes ESMCI/cime#1687 

User interface changes?: none

Code review:


billsacks:
Thanks, @mvertens - were you planning to merge this or do you want me to?

mvertens:
@billsacks - If you could merge it - that would be great! Thanks.

PR #1692, opened 2017-06-14
Add a warning about use of component_defining_compset

billsacks:
In my ideal world, we wouldn't have this component_defining_compset variable at
all, and so wouldn't need this warning. Instead, _set_compset would set
self._primary_component directly for both user-defined compsets and compsets
with aliases (it could then be renamed to
_set_compset_and_primary_component). However, for this to work, we'd need to
call self.set_comp_classes before self._set_compset. I haven't tracked the
dependencies of self.set_comp_classes to see if this would be possible.

Test suite: scripts_regression_tests B_CheckCode
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

Fixes none

User interface changes?: none

Code review:

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -746,6 +746,10 @@ class Case(object):
         files = Files()
         compset_alias, science_support, component_defining_compset = self._set_compset(
             compset_name, files, user_compset=user_compset)
+        # WARNING: component_defining_compset will be None for a user-defined
+        # compset. If you use this variable, behavior of user-defined compsets
+        # will differ from behavior of compsets with aliases, which is typically
+        # undesirable. Instead, use self._primary_component, which is set below.
 
         self._components = self.get_compset_components()
         #--------------------------------------------


mvertens:
I agree with Bill. If we could not have this it would be great.

jedwards4b:
I have a fix as part of PR #1697 if you agree we can close this PR.

billsacks:
@jedwards4b - Your proposed fix represents a behavior change in the case where someone defines a compset in something other than the primary component. e.g., imagine that CAM defined an I compset in its config_compsets.xml file (maybe because they want an I compset that exactly matches one of their defined F compsets). In the old code, CAM would be considered the primary component; in the new code, CLM would be considered the primary component. I don't know which is right (though I do intuitively like the new code better), but behavior could theoretically differ in those two cases.

If this situation is not supposed to be allowed, then I'd suggest that we still return component_defining_compset from _set_compset, and just use it as an error check, like:

```python
self._primary_component = self._find_primary_component()
if component_defining_compset is not None:
    expect(component_defining_compset == self._primary_component, "Compsets must be defined in their primary component: compset was defined in {} but should be defined in {}".format(component_defining_compset, self._primary_component))
```

I'd prefer if your just-committed change were made its own PR, since it makes it hard to track down problems after-the-fact when behavior changes like this are combined with other big changes. But if you do keep this change in #1697 , the overall PR comment should be updated to mention this change, so that the change is obvious to readers of the git log.

jedwards4b:
There should not be a case in which cam defines an I compset in it's config_compsets.xml and even if it did then CLM should be the primary component as primary component is defined.

billsacks:
> There should not be a case in which cam defines an I compset in it's config_compsets.xml and even if it did then CLM should be the primary component as primary component is defined.

@jedwards4b - I'm totally fine with this if you (and @mvertens ) feel this should be the behavior. However, I'd still like to see:

1. Maintaining the error check I suggested above

2. Either moving this to its own PR, or at the very least updating the PR comments in #1697 to call out this change

rljacob:
Discussion from telecon:  internal rules and not alias now determines primary component.   If a B-case is defined in CAM config_compset,  primary component is considered "allactive" (by the internal rules) and not "CAM".  Source of confusion when considering attributed on compsets.  Refcase for the B-case-in-CAM would not work. pe-layout would come from allactive.

billsacks:
Closing this PR but opening issue #1722 in its place.

Issue #1683, opened 2017-06-12
$comp is hard-coded in buildlib.csm_share


dqwu:
```
my $COMP_INTERFACE = `./xmlquery COMP_INTERFACE	-value`;
...
my $comp="mct";
...
my $libdir = "$sharedlibroot/$comp/$useesmf/${NINST_VALUE}/csm_share";
```
In above code, a hard-coded string "mct" is assigned to $comp. This value is defined in config_component.xml and submit to change (PR #1663 changed it from "MCT" to "mct")
```
<entry id="COMP_INTERFACE">
  ...
  <valid_values>mct</valid_values>
  <default_value>mct</default_value>
  ...
</entry>
```
Instead of hard-coding, is it better to bind $comp to the value read from config_component.xml?
`my $comp = $COMP_INTERFACE;`

jedwards4b:
Yes

Issue #1522, opened 2017-05-12
FIXME at line 676 of case.py: validate compset settings for user-compset


jedwards4b:
I would like to remove this comment in case.py - can someone give me an example that
gets this far with invalid compset settings?

        self._components = self.get_compset_components()
        #FIXME - if --user-compset is True then need to determine that
        #all of the compset settings are valid

billsacks:
I'm not sure if this is what the FIXME note is referring to, but I see:

(1) Valid compset (as a "control"):

```
./create_newcase --compset 2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV --user-compset --res f45_g37 --run-unsupported --case test_0512a --pesfile ../src/drivers/mct/cime_config/config_pes.xml
```

(2) Invalid compset:

```
./create_newcase --compset 2000_DATM%FOO_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV --user-compset --res f45_g37 --run-unsupported --case test_0512c --pesfile ../src/drivers/mct/cime_config/config_pes.xml
```

Specifically, note the use of `DATM%FOO`. This happily moves ahead; I haven't checked what the resulting case looks like.

The only way I can think of to catch errors like this is if every component lists allowed settings in its config_components.xml file... actually, that might be a good requirement to catch typos in user_compsets.

jedwards4b:
So we could have a new variable in config_component.xml 
```
<entry id="COMPONENTNAME_MODIFIERS">
<valid_values>NYF, NXF, FOO</valid_values>

</entry>
```
I can make this optional so that all components do not need to supply it right away - if a component does supply it then any modifier provided for that component in the compset name will be checked against this list.   If it's not supplied than no error check is done for modifiers in that component.

billsacks:
@jedwards4b - yes, I really like that solution.

billsacks:
Actually, I _think_ (but am not sure) that datm, and maybe other components (probably mostly data models) require at least one modifier. i.e., I _think_ it's invalid to have a compset like `2000_DATM_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV`. If I'm right about that, then is there a way we can also denote that a modifier is required for a given component?

jedwards4b:
Yes, but can you confirm that?

billsacks:
@jedwards4b - I don't know how to confirm this.

[...]

PR #1697, opened 2017-06-15
Description field error check of user compsets.

jedwards4b:
This PR introduces version 3.0 schema for entry_id files bringing the description field to the top of the file and refining it to better define valid compsetname values.   Change definition of primary component as suggested in PR 1692.

Test suite: scripts_regression_tests.py, several F case user compsets with modified cam config_component.xml
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes ESMCI/cime#1522  
Fixes ESMCI/cime#1683 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/component.py
+++ b/scripts/lib/CIME/XML/component.py
@@ -226,6 +226,10 @@ class Component(EntryID):
         # The following is not allowed because the required WCCM field is missing
         >>> obj._get_description_match("1850_CAM50%RCO2_Barn",set(["CAM50", "WCCM"]), set(["CAM50","WCCM","RCO2"]),{"RCO2":"CRU"}, "*")
         False
+        >>> obj._get_description_match("1850_CAM50_Barn",set(["CAM50", "WCCM"]), set(["CAM50","WCCM","RCO2"]),{"RCO2":"CRU"}, "+")
+        False
+        >>> obj._get_description_match("1850_CAM50%WCCM_Barn",set(["CAM50", "WCCM"]), set(["CAM50","WCCM","RCO2"]),{"RCO2":"CRU"}, "+")
+        True
         """
         match = False
         comparts = compsetname.split('_')
--- a/scripts/lib/CIME/XML/component.py
+++ b/scripts/lib/CIME/XML/component.py
@@ -11,17 +11,18 @@ logger = logging.getLogger(__name__)
 
 class Component(EntryID):
 
-    def __init__(self, infile):
+    def __init__(self, infile, comp_class=None):
         """
         initialize an object
         """
         files = Files()
         schema = None
-        # not checking schema on external components yet
-        cimeroot = get_cime_root()
-        if  cimeroot in os.path.abspath(infile):
-            schema = files.get_schema("CONFIG_CPL_FILE")
-        EntryID.__init__(self, infile, schema=schema)
+        EntryID.__init__(self, infile)
+        if comp_class is None:
+            schema = files.get_schema("CONFIG_CPL_FILE", attributes={"version":"{}".format(self.get_version())})
+        else:
+            schema = files.get_schema("CONFIG_{}_FILE".format(comp_class), attributes={"version":self.get_version()})
+        self.validate_xml_file(infile, schema)
 
     #pylint: disable=arguments-differ
     def get_value(self, name, attribute=None, resolved=False, subgroup=None):
@@ -119,7 +120,7 @@ class Component(EntryID):
     def get_description(self, compsetname, component=None):
         component = component.lower()
         rootnode = self.get_node("description")
-        desc = None
+        desc = ""
         if self.get_version() == 3.0:
             expect(component is not None,"component argument required for version3 files")
             modifier_mode = rootnode.get('modifier_mode')
@@ -130,6 +131,13 @@ class Component(EntryID):
 
             desc_nodes = self.get_nodes("desc", root=rootnode)
             optiondesc = {}
+            if component == "forcing":
+                for node in desc_nodes:
+                    forcing = node.get('forcing')
+                    if forcing is not None and compsetname.startswith(forcing):
+                        desc = node.text
+                        return desc
+
             # first pass just make a hash of the option descriptions
             for node in desc_nodes:
                 option = node.get('option')
@@ -162,7 +170,7 @@ class Component(EntryID):
                             expect(len(complist) <= 2,
                                    "Expected 0 or one modifers found {}".format(len(complist)))
                         # found a match
-                        expect(desc is None
+                        expect(len(desc) == 0
                                ,"Found multiple matchs in file {} one: {} another: {}".format(self.filename,desc,node.text))
                         desc = node.text
                         for part in comp.split('%'):
--- a/scripts/lib/CIME/XML/component.py
+++ b/scripts/lib/CIME/XML/component.py
@@ -155,7 +155,8 @@ class Component(EntryID):
                     expect(len(desc)==0,
                            "Too many matches on forcing field {} in file {}".\
                                format(forcing, self.filename))
-                desc = node.text
+                    desc = node.text
+
             expect(len(desc) > 0,"No match found for forcing field {} in file {}".\
                        format(forcing, self.filename))
             return desc
--- a/scripts/lib/CIME/XML/component.py
+++ b/scripts/lib/CIME/XML/component.py
@@ -195,8 +195,8 @@ class Component(EntryID):
             fullset = set(parts+opt_parts)
             if self._get_description_match(compsetname, reqset, fullset, modifier_mode):
                 desc = node.text
-        # cpl component may not have a description, all others must
-        if comp_class != 'cpl':
+        # cpl and esp components may not have a description
+        if comp_class not in ['cpl','esp']:
             expect(len(desc) > 0,
                    "No description found for comp_class {} matching compsetname {} in file {}"\
                        .format(comp_class,compsetname, self.filename))
--- a/scripts/lib/CIME/XML/component.py
+++ b/scripts/lib/CIME/XML/component.py
@@ -118,21 +118,21 @@ class Component(EntryID):
         return match_value
 
     #pylint: disable=arguments-differ
-    def get_description(self, compsetname, component=None):
-        component = component.lower()
+    def get_description(self, compsetname, comp_class=None):
+        comp_class = comp_class.lower()
         rootnode = self.get_node("description")
         desc = ""
+        desc_nodes = self.get_nodes("desc", root=rootnode)
         if self.get_version() == 3.0:
-            expect(component is not None,"component argument required for version3 files")
+            expect(comp_class is not None,"comp_class argument required for version3 files")
             modifier_mode = rootnode.get('modifier_mode')
             if modifier_mode is None:
                 modifier_mode = '*'
                 expect(modifier_mode in ('*','1','?','+'),
                        "Invalid modifier_mode {} in file {}".format(modifier_mode, self.filename))
 
-            desc_nodes = self.get_nodes("desc", root=rootnode)
             optiondesc = {}
-            if component == "forcing":
+            if comp_class == "forcing":
                 for node in desc_nodes:
                     forcing = node.get('forcing')
                     if forcing is not None and compsetname.startswith(forcing):
@@ -144,9 +144,9 @@ class Component(EntryID):
                 option = node.get('option')
                 if option is not None:
                     optiondesc[option] = node.text
-            #second pass find a component match
+            #second pass find a comp_class match
             for node in desc_nodes:
-                compdesc = node.get(component)
+                compdesc = node.get(comp_class)
 
                 if compdesc is None:
                     continue
@@ -160,7 +160,7 @@ class Component(EntryID):
                     complist = comp.split('%')
 
                     cset = set(complist)
-                    if cset <= fullset:
+                    if cset == reqset or (cset > reqset and cset <= fullset):
                         if modifier_mode == '1':
                             expect(len(complist) == 2,
                                    "Expected exactly one modifer found {}".format(len(complist)))
@@ -177,8 +177,11 @@ class Component(EntryID):
                         for part in comp.split('%'):
                             if part in optiondesc:
                                 desc += optiondesc[part]
-
-        elif self.get_version <= 2.0:
+            if comp_class != 'cpl':
+                expect(len(desc) > 0,
+                   "No description found for comp_class {} matching compsetname {} in file {}"\
+                       .format(comp_class,compsetname, self.filename))
+        elif self.get_version() <= 2.0:
             for node in desc_nodes:
                 compsetmatch = node.get("compset")
                 if compsetmatch is not None and re.search(compsetmatch, compsetname):
--- a/scripts/lib/CIME/XML/component.py
+++ b/scripts/lib/CIME/XML/component.py
@@ -11,19 +11,19 @@ logger = logging.getLogger(__name__)
 
 class Component(EntryID):
 
-    def __init__(self, infile, comp_class=None):
+    def __init__(self, infile, comp_class):
         """
         initialize a Component obect from the component xml file in infile
         associate the component class with comp_class if provided.
         """
+        self._comp_class = comp_class
+        if infile == 'testingonly':
+            self.filename = infile
+            return
         files = Files()
         schema = None
         EntryID.__init__(self, infile)
-        if comp_class is None:
-            schema = files.get_schema("CONFIG_CPL_FILE", attributes={"version":"{}".format(self.get_version())})
-        else:
-            schema = files.get_schema("CONFIG_{}_FILE".format(comp_class), attributes={"version":"{}".format(self.get_version())})
-            self._comp_class = comp_class
+        schema = files.get_schema("CONFIG_{}_FILE".format(comp_class), attributes={"version":"{}".format(self.get_version())})
 
         if schema is not None:
             self.validate_xml_file(infile, schema)
@@ -135,6 +135,22 @@ class Component(EntryID):
 
 
     def _get_description_v3(self, compsetname, comp_class=None):
+        """
+        version 3 of the config_component.xml file has the description section at the top of the file
+        the description field has one attribute 'modifier_mode' which has allowed values
+        '*' 0 or more modifiers (default)
+        '1' exactly 1 modifier
+        '?' 0 or 1 modifiers
+        '+' 1 or more modifiers
+
+        modifiers are fields in the component section of the compsetname following the % symbol.
+
+        The desc field can have an attribute which is the component class ('cpl', 'atm', 'lnd' etc)
+        or it can have an attribute 'option' which provides descriptions of each optional modifier
+        or (in the config_component_{model}.xml in the driver only) it can have the attribute 'forcing'
+
+        component descriptions are matched to the compsetname using a set method
+        """
         comp_class = comp_class.lower()
         rootnode = self.get_node("description")
         desc = ""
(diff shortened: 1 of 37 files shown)


jedwards4b:
I would like someone to run scripts_regression_tests.py with MODEL=acme.

gold2718:
> I would like someone to run scripts_regression_tests.py with MODEL=acme.

I've started it up on compute001

scripts/lib/CIME/XML/component.py
```diff
@@ -11,17 +11,19 @@
 
 class Component(EntryID):
 
-    def __init__(self, infile):
+    def __init__(self, infile, comp_class=None):
```
billsacks: Please add documentation of the comp_class argument

scripts/lib/CIME/XML/component.py
```diff
@@ -116,14 +118,74 @@ def _get_value_match(self, node, attributes=None, exact_match=False):
         return match_value
 
     #pylint: disable=arguments-differ
-    def get_description(self, compsetname):
+    def get_description(self, compsetname, comp_class=None):
```
billsacks: Again, please add description of the comp_class argument.

Also, since comp_class is passed into the `__init__` method, is it possible to store comp_class as self._comp_class so you don't need to pass it in here, too? At a glance, it looks like it only needs to be passed in for drv to distinguish between the forcing and something else; otherwise you typically pass the same comp_class here as was used in `__init__`; if so that should be documented. Or maybe even more clear would be to introduce a separate get_description_forcing method, since it looks like the code path differs pretty significantly when comp_class is 'forcing' vs. anything else.

scripts/lib/CIME/XML/component.py
```diff
@@ -116,14 +118,74 @@ def _get_value_match(self, node, attributes=None, exact_match=False):
...
-            compsetmatch = node.get("compset")
-            if re.search(compsetmatch, compsetname):
-                desc += node.text
+        desc_nodes = self.get_nodes("desc", root=rootnode)
+        if self.get_version() == 3.0:
+            expect(comp_class is not None,"comp_class argument required for version3 files")
+            modifier_mode = rootnode.get('modifier_mode')
+            if modifier_mode is None:
+                modifier_mode = '*'
+                expect(modifier_mode in ('*','1','?','+'),
```
billsacks: This expect line should be out-dented one level

scripts/lib/CIME/XML/component.py
```diff
@@ -116,14 +118,74 @@ def _get_value_match(self, node, attributes=None, exact_match=False):
...
+                                   "Expected exactly one modifer found {}".format(len(complist)))
+                        elif modifier_mode == '+':
+                            expect(len(complist) >= 2,
+                                   "Expected one or more modifers found {}".format(len(complist)))
+                        elif modifier_mode == '?':
+                            expect(len(complist) <= 2,
+                                   "Expected 0 or one modifers found {}".format(len(complist)))
+                        # found a match
+                        expect(len(desc) == 0
+                               ,"Found multiple matchs in file {} one: {} another: {}".format(self.filename,desc,node.text))
```
billsacks: typo: matchs should be matches

scripts/lib/CIME/XML/component.py
```diff
@@ -116,14 +118,74 @@ def _get_value_match(self, node, attributes=None, exact_match=False):
...
+            if modifier_mode is None:
+                modifier_mode = '*'
+                expect(modifier_mode in ('*','1','?','+'),
+                       "Invalid modifier_mode {} in file {}".format(modifier_mode, self.filename))
+
+            optiondesc = {}
+            if comp_class == "forcing":
+                for node in desc_nodes:
+                    forcing = node.get('forcing')
+                    if forcing is not None and compsetname.startswith(forcing):
```
billsacks: I don't completely understand what's happening here, so this suggestion may not make sense, but: I think it would be more robust if you checked `compset.startswith(forcing + '_')`. This way you'd avoid matching a forcing that is a substring of another forcing. e.g., imagine a forcing 'RCP4' and another 'RCP4.5'.

scripts/lib/CIME/XML/component.py
```diff
@@ -116,14 +118,74 @@ def _get_value_match(self, node, attributes=None, exact_match=False):
...
+                        desc = node.text
+                        return desc
+
+            # first pass just make a hash of the option descriptions
+            for node in desc_nodes:
+                option = node.get('option')
+                if option is not None:
+                    optiondesc[option] = node.text
+            #second pass find a comp_class match
```
billsacks: I couldn't understand what the following code is doing, at a high level (so I haven't given it much of a review). This could use some high-level documentation and/or examples. Ideally, there would also be some unit tests - either for this whole get_description method or for the core algorithm (below), which could be extracted into its own method.

scripts/lib/CIME/XML/component.py
billsacks: This method now serves the additional important purpose of doing error checking on a compset name, right? If so, I'd vote for renaming it, even if the name becomes more verbose: something like get_compset_description_and_check_validity

src/drivers/mct/cime_config/config_component_cesm.xml
```diff
@@ -2,7 +2,31 @@
...
+    <desc forcing="2003">2002-2003</desc>
+    <desc forcing="HIST">Historic transient </desc>
+    <desc forcing="20TR">Twentieth century transient</desc>
+    <desc forcing="4804"></desc>
+    <desc forcing="RCP2">CMIP5 rcp 2.6 forcing</desc>
+    <desc forcing="RCP4">CMIP5 rcp 4.5 forcing</desc>
+    <desc forcing="RCP6">CMIP5 rcp 6.0 forcing</desc>
+    <desc forcing="RCP8">CMIP5 rcp 8.5 forcing</desc>
+    <desc cpl="BGC[%BDRD][%BPRP]">Biogeochemistry intercomponent </desc>
+    <desc option="BDRD"> does something</desc>
```
billsacks: BGC CO2=diagnostic, rad CO2=diagnostic

src/drivers/mct/cime_config/config_component_cesm.xml
```diff
@@ -2,7 +2,31 @@
...
+    <desc option="BPRP"> does another thing</desc>
```
billsacks: BGC CO2=prognostic, rad CO2=prognostic

gold2718:
I'm getting errors (in the ACME test) like:
```
2017-06-15 12:56:39: CREATE_NEWCASE FAILED for test 'TESTBUILDFAILEXC_P1.f19_g16_rx1.A.anlworkstation_gnu'.
Command: /nfs2/goldy/Projects/cime/scripts/create_newcase --case /homes/goldy/acme/scratch/scripts_regression_test.20170615_125418/TESTBUILDFAILEXC_P1.f19_g16_rx1.A.anlworkstation_gnu.fake_testing_only_20170615_125637-20170615_125637 --res f19_g16_rx1 --mach anlworkstation --compiler gnu --compset A --test --output-root /homes/goldy/acme/scratch/scripts_regression_test.20170615_125418  --pecount 1 --walltime 0:10:00
Output: Compset longname is 2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV
Compset specification file is /nfs2/goldy/Projects/cime/src/drivers/mct/cime_config/config_compsets.xml
ERROR: Incorrect number of matches, 2, for nodename 'description' and attrs 'None' in file '/nfs2/goldy/Projects/cime/src/drivers/mct/cime_config/config_component_acme.xml'
```

scripts/lib/CIME/XML/component.py
```diff
@@ -116,14 +118,74 @@ def _get_value_match(self, node, attributes=None, exact_match=False):
...
+                       "Invalid modifier_mode {} in file {}".format(modifier_mode, self.filename))
+
+            optiondesc = {}
+            if comp_class == "forcing":
+                for node in desc_nodes:
+                    forcing = node.get('forcing')
+                    if forcing is not None and compsetname.startswith(forcing):
+                        desc = node.text
+                        return desc
+
```
billsacks: I'm thinking this should error out if you haven't found an appropriate forcing? Otherwise it's going to fall through to the code below, which I'm guessing is the wrong thing to do.

scripts/lib/CIME/XML/component.py
jedwards4b: Thanks

scripts/lib/CIME/XML/component.py
jedwards4b: Changed to get_forcing_description.  Reuse comp_class from __init__ method.

jedwards4b: Since it doesn't do this for older versions I am leaving the name alone and adding a comment.

scripts/lib/CIME/XML/component.py
jedwards4b: good point, done

src/drivers/mct/cime_config/config_component_cesm.xml
jedwards4b: fixed

jedwards4b:
@gold2718 I think its fixed, please try again.
@billsacks I think I've resolved the issues you raised - thanks.

scripts/lib/CIME/XML/component.py
```diff
@@ -11,17 +11,22 @@
...
         """
         files = Files()
         schema = None
-        # not checking schema on external components yet
-        cimeroot = get_cime_root()
-        if  cimeroot in os.path.abspath(infile):
-            schema = files.get_schema("CONFIG_CPL_FILE")
-        EntryID.__init__(self, infile, schema=schema)
+        EntryID.__init__(self, infile)
+        if comp_class is None:
```
billsacks: I think you also need to set self._comp_class in this branch of the conditional. Is it supposed to be set to 'CPL' in this case (I can't tell for sure)?

scripts/lib/CIME/XML/component.py
```diff
@@ -117,12 +122,96 @@ def _get_value_match(self, node, attributes=None, exact_match=False):
...
+            expect(len(desc) > 0,"No match found for forcing field {} in file {}".\
+                       format(forcing, self.filename))
+            return desc
+
+        # first pass just make a hash of the option descriptions
+        for node in desc_nodes:
+            option = node.get('option')
+            if option is not None:
+                optiondesc[option] = node.text
+        #second pass find a comp_class match
```
billsacks: This is the one comment from before that you didn't resolve:

> I couldn't understand what the following code is doing, at a high level (so I haven't given it much of a review). This could use some high-level documentation and/or examples. Ideally, there would also be some unit tests - either for this whole get_description method or for the core algorithm (below), which could be extracted into its own method.

jedwards4b: I've added documentation and several unit tests.

jedwards4b:
I've repeated scripts_regression_tests.py.

gold2718:
Which means starting over with the ACME version . . .

gold2718:
End of the output of scripts_regression_tests.py on compute001 (ACME):
```
FAIL: test_save_timings (__main__.L_TestSaveTimings)
  File "./scripts_regression_tests.py", line 1658, in test_save_timings
    self.simple_test()
  File "./scripts_regression_tests.py", line 1653, in simple_test
    self.assertTrue(os.path.isdir(provenance_dir), msg="'%s' was missing" % provenance_dir)
AssertionError: '/homes/goldy/acme/scratch/scripts_regression_test.20170615_192759/timings/performance_archive/goldy/SMS_Ln9_P1.f19_g16_rx1.A.anlworkstation_gnu.fake_testing_only_20170615_193548/170615-193805' was missing

FAIL: test_save_timings_manual (__main__.L_TestSaveTimings)
  File "./scripts_regression_tests.py", line 1663, in test_save_timings_manual
    self.simple_test(manual_timing=True)
  File "./scripts_regression_tests.py", line 1653, in simple_test
    self.assertTrue(os.path.isdir(provenance_dir), msg="'%s' was missing" % provenance_dir)
AssertionError: '/homes/goldy/acme/scratch/scripts_regression_test.20170615_192759/timings/performance_archive/goldy/SMS_Ln9_P1.f19_g16_rx1.A.anlworkstation_gnu.fake_testing_only_20170615_193854/170615-194107' was missing

Ran 315 tests in 7412.369s

FAILED (failures=2, skipped=5)
Detected failed test or user request no teardown
Leaving files:
  /homes/goldy/acme/scratch/scripts_regression_test.20170615_192759/cs.status.fake_testing_only_20170615_193548
  /homes/goldy/acme/scratch/scripts_regression_test.20170615_192759/SMS_Ln9_P1.f19_g16_rx1.A.anlworkstation_gnu.fake_testing_only_20170615_193548
Detected failed test or user request no teardown
Leaving files:
  /homes/goldy/acme/scratch/scripts_regression_test.20170615_192759/cs.status.fake_testing_only_20170615_193854
  /homes/goldy/acme/scratch/scripts_regression_test.20170615_192759/SMS_Ln9_P1.f19_g16_rx1.A.anlworkstation_gnu.fake_testing_only_20170615_193854
Detected failed test or user request no teardown
Leaving files:
Detected failures, leaving directory: /homes/goldy/acme/scratch/scripts_regression_test.20170615_192759
```

jedwards4b:
@gold2718 Those errors don't seem at all related to this change, would you agree?

gold2718:
@jedwards4b, I have no idea. The output of scripts_regression_tests.py is still pretty murky for me. The system is not perfectly stable and if something in the changes broke something in ACME, I think it should at least be looked at. @jgfouca, can you provide any insight into why these tests might have failed?

jedwards4b:
After updating the acme machines definition of caldera and creating a SAVE_TIMING_DIR, I was able to get this test to pass:
```
 DAV_CORES=16 execca ./scripts_regression_tests.py  L_TestSaveTimings
 
  Requesting 16 core(s) to caldera queue, 
  to submit ./scripts_regression_tests.py L_TestSaveTimings the usage is to be charged into CESM0005 
  running

  bsub -Is -q caldera -n16 -PCESM0005 -W24:00 "./scripts_regression_tests.py L_TestSaveTimings"

  please wait.. 

Job <50916> is submitted to queue <caldera>.
<<Waiting for dispatch ...>>
<<Starting on caldera10-ib>>

Testing commit 86a876ecb014cb1d8823b88855f03a4c56504819
Using cime_model = acme
Testing machine = caldera
Test root: /glade/scratch/jedwards/scripts_regression_test.20170616_094527

test_save_timings (__main__.L_TestSaveTimings) ... ok
test_save_timings_manual (__main__.L_TestSaveTimings) ... ok

Ran 2 tests in 586.194s

OK
All pass, removing directory: /glade/scratch/jedwards/scripts_regression_test.20170616_094527
```

gold2718:
@billsacks, thanks for noticing that. That check will not work for long. For instance, the active data-assimilation ESP component is called DART, no ESP in the name at all.

scripts/lib/CIME/case.py
```diff
@@ -617,13 +620,19 @@ def _get_component_config_data(self, files):
             comp_config_file = self.get_resolved_value(comp_config_file)
             expect(comp_config_file is not None and os.path.isfile(comp_config_file),
                    "Config file {} for component {} not found.".format(comp_config_file, comp_name))
-            compobj = Component(comp_config_file)
-            self._component_description[comp_class] = compobj.get_description(self._compsetname)
-            expect(self._component_description[comp_class] is not None,"No description found in file {} for component {}".format(comp_config_file, comp_name))
+            compobj = Component(comp_config_file, comp_class)
+            compsetname = self._compsetname
+            if not compsetname.endswith('ESP'):
```
jedwards4b: I've pushed an alternative method to deal with it.   It leaves a potential for errors in the esp component but is otherwise more robust.

billsacks:
Okay, thanks @jedwards4b . I'll let @gold2718 comment on whether this ESP handling is okay or if he sees some better way to handle it. I approve these changes now, though I hope that @gold2718 will get a chance to review the core logic in `_get_description_v3` and `_get_description_match`, which I still don't completely understand (in part because I'm tired today).

gold2718:
The errors on compute001 still happen. There is no 'timings' directory -- could that be a hint?

jedwards4b:
Yes there needs to be a SAVE_TIMING_DIR which exists and is not associated with CIME_OUTPUT_ROOT

gold2718:
To make sure I understand, are you saying that this machine (anlworkstation) is not correctly configured to pass scripts_regression_tests.py. @jgfouca, please also weigh in here.

jedwards4b:
It has <SAVE_TIMING_DIR>$CIME_OUTPUT_ROOT/timings</SAVE_TIMING_DIR> but the scripts_regression_tests.py moves the location of CIME_OUTPUT_ROOT and this directory does not exist, so the test fails.  Not sure why they don't just create the directory @jgfouca

jgfouca:
@jedwards4b @gold2718 , this hits the acme provenance code which Worley doesn't like me to change. It might be easiest to just create the timings dir.

rljacob:
Can scripts_regression_tests be modified to skip the provenance tests if SAVE_TIMING_DIR doesn't exist?  We recently removed SAVE_TIMING_DIR from all the non-production machines.

jgfouca:
@rljacob how will we test that it works? I don't think there's any overlap between the production machines and our nightly testing machines.

rljacob:
run just the provenance test on a production machine?

jedwards4b:
@jgfouca I don't  believe that anything in this PR directly affected the acme provenance code.  The change must have been brought in with the latest acme merge?

jgfouca:
@jedwards4b yep, I'm quite sure that's it.

jedwards4b:
@gold2718 @mvertens  - is this review complete?

gold2718:
Sorry, still working through it. Will be done by the end of the day (my schedule, not yours :)

src/components/data_comps/desp/cime_config/config_component.xml
```diff
@@ -2,13 +2,21 @@
...
+       '1' exactly 1 modifier
+       '?' 0 or 1 modifiers
+       '+' 1 or more modifiers
+
+       This file may have esp desc entries.
+  -->
 
-  <!-- NOTE that the description block determines what DESP% values can appear in the compset name 
-       For DESP this is determined by the DESP_MODE values -->
   <description>
```
gold2718: This should have `modifier_mode='1'`

src/components/data_comps/dwav/cime_config/config_component.xml
```diff
@@ -1,12 +1,19 @@
...
+
+       This file may have wav desc entries.
+  -->
   <description>
-    <desc compset="_DWAV%NULL">dwav null mode:</desc>
-    <desc compset="_DWAV%CLIMO">dwav climatological mode:</desc>
+    <desc wav="DWAV[%NULL][%CLIMO]">Data wave model (DWAV)</desc>
+    <desc option="NULL">null mode</desc>
+    <desc option="CLIMO">climatological mode</desc>
   </description>
```
gold2718: Can dwav have both `%NULL` and `%CLIMO`? If not, maybe it needs `modifier_mode='1'` unless I misunderstand this interface.

src/drivers/mct/cime_config/config_component_cesm.xml
```diff
@@ -2,7 +2,31 @@
...
+    <desc forcing="20TR">Twentieth century transient</desc>
+    <desc forcing="4804"></desc>
+    <desc forcing="RCP2">CMIP5 rcp 2.6 forcing</desc>
+    <desc forcing="RCP4">CMIP5 rcp 4.5 forcing</desc>
+    <desc forcing="RCP6">CMIP5 rcp 6.0 forcing</desc>
+    <desc forcing="RCP8">CMIP5 rcp 8.5 forcing</desc>
+    <desc cpl="BGC[%BDRD][%BPRP]">Biogeochemistry intercomponent </desc>
+    <desc option="BDRD"> with diagnostic CO2</desc>
+    <desc option="BPRP"> with prognostic CO2 </desc>
+
```
gold2718: What are we calling things like `SDYN_`, `AR95_`, and `AR97_`? Aren't those forcings? Should they be here so CAM can update its `<description>` section?

scripts/lib/CIME/XML/component.py
```diff
@@ -117,12 +122,145 @@ def _get_value_match(self, node, attributes=None, exact_match=False):
...
+
+        modifiers are fields in the component section of the compsetname following the % symbol.
+
+        The desc field can have an attribute which is the component class ('cpl', 'atm', 'lnd' etc)
+        or it can have an attribute 'option' which provides descriptions of each optional modifier
+        or (in the config_component_{model}.xml in the driver only) it can have the attribute 'forcing'
+
+        component descriptions are matched to the compsetname using a set method
+        """
+        comp_class = comp_class.lower()
```
gold2718: `comp_class` defaults to `None`. Shouldn't there be a check here (like move the expect on line 159 before 154)?

scripts/lib/CIME/XML/component.py
```diff
@@ -117,12 +122,145 @@ def _get_value_match(self, node, attributes=None, exact_match=False):
...
+
+            if compdesc is None:
+                continue
+            opt_parts = [ x.rstrip("]") for x in compdesc.split("[%") ]
+            parts = opt_parts.pop(0).split("%")
+
+            reqset = set(parts)
+            fullset = set(parts+opt_parts)
+            if self._get_description_match(compsetname, reqset, fullset, modifier_mode):
+                desc = node.text
```
gold2718: If we found a desc, shouldn't we exit the `for` loop here?

scripts/lib/CIME/XML/component.py
```diff
@@ -117,12 +122,145 @@ def _get_value_match(self, node, attributes=None, exact_match=False):
...
+                           "Expected exactly one modifer found {}".format(len(complist)-1))
+                elif modifier_mode == '+':
+                    expect(len(complist) >= 2,
+                           "Expected one or more modifers found {}".format(len(complist)-1))
+                elif modifier_mode == '?':
+                    expect(len(complist) <= 2,
+                           "Expected 0 or one modifers found {}".format(len(complist)-1))
+                expect(not match,"Found multiple matches in file {} for {}".format(self.filename,comp))
+                match = True
+                # found a match
```
gold2718: Why not exit the `for` loop here?

src/components/data_comps/desp/cime_config/config_component.xml
jedwards4b: done

src/components/data_comps/dwav/cime_config/config_component.xml
jedwards4b: done

src/drivers/mct/cime_config/config_component_cesm.xml
jedwards4b: Yes, forcings are defined here.  You are invited to submit a PR with additional forcing descriptions if your component needs them.

scripts/lib/CIME/XML/component.py
jedwards4b: fixed

scripts/lib/CIME/XML/component.py
jedwards4b: In this one we want to pass all the way through to make sure that there is only one match and throw an error if there is more than one.

gold2718: Sorry, I did not process the expect at line 249 properly.

src/drivers/mct/cime_config/config_component_cesm.xml
gold2718: I can do that with a modified CAM `<description>` section but in the meantime, will all of the CAM compsets still work properly?

jedwards4b: Yes, as long as you don't change to the version="3.0" format things will work as they do now.

PR #1712, opened 2017-06-28
fix an issue with pio_numtasks setting and make sure that external sc…

### LINKED ISSUE #1711: PIO_NUMTASKS is getting reset
Setting env_run.xml's PIO_STRIDE to default -99 and PIO_NUMTASKS to 64 resets both PIO_STRIDE and PIO_NUMTASKS to different default values at run-time.

XML comments:
```
    <entry id="PIO_STRIDE">
      <desc>
     stride in compute comm of io tasks for each component, if this value is -99 it will 
     be computed based on PIO_NUMTASKS and number of compute tasks
    </desc>
```

but instead of 64, PIO_NUMTASKS is getting reset to 8 or 16 depending on the number of component's MPI tasks.

Note that setting PIO_STRIDE to a specific value and leaving PIO_NUMTASKS at default -99 works as expected: PIO_NUMTASKS is set to NTASKS / PIO_STRIDE.

jedwards4b:
…ripts are syncd with case.

We need to flush the case before running externals scripts and to reread xml after each script completes.   This assures that data in externals scripts is in sync with the case object.
Also fix pio namelist initialization of pio_numtasks.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1711

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/case_run.py
+++ b/scripts/lib/CIME/case_run.py
@@ -269,8 +269,10 @@ def case_run(case, skip_pnl=False):
             lid = new_lid()
 
         if prerun_script:
+            case.flush()
             do_external(prerun_script, case.get_value("CASEROOT"), case.get_value("RUNDIR"),
                         lid, prefix="prerun")
+            case.read_xml()
 
         lid = run_model(case, lid, skip_pnl)
         save_logs(case, lid)       # Copy log files back to caseroot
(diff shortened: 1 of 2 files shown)


jgfouca:
Waiting for @jayeshkrishna to approve.

PR #1719, opened 2017-06-30
Add Support for Email Options

### LINKED ISSUE #797: XML option to control email at end of the run
On slack, @rljacob said
> Some people want to turn off or on the "email at end of job" setting.

This is something I would like to be able to control, and I think `env_batch.xml` is the way to do it:

> I was thinking it would be nice to run `./xmlchange JOB_EMAIL_ON_COMPLETION=...` either as a binary or a string compare (`valid_values=always, if_success, if_failure,never`?)

I think the actual XML variable name and the best implementation is still open for debate.

erichlf:
Added an email option to env_batch.xml. Also, add the ability to send an email option through case.submit. The flags in case.submit follow slurm's usage, i.e. --mail-user and --mail-type, where type here is never, all, begin, end, fail.

Test suite: scripts_regression_tests.py (Anvil, Skybridge, and Climate)
various email options passed to case.submit on the above three machines, and similarly in env_batch.xml.
Test baseline: 
Test namelist changes:
Test status: bit for bit

Fixes #797

User interface changes?: Added --mail-user and --mail-type to case.submit

### DIFF
--- a/config/acme/machines/config_batch.xml
+++ b/config/acme/machines/config_batch.xml
@@ -86,6 +86,7 @@
       <directive default="acme.stdout" > -o {{ output_error_path }}.%J  </directive>
       <directive default="acme.stderr" > -e {{ output_error_path }}.%J  </directive>
       <directive                       > -J {{ job_id }} </directive>
+      <directive                       > -u {{ mail_user }} </directive>
     </directives>
   </batch_system>
 
@@ -106,8 +107,9 @@
       <directive default="n"> -r {{ rerunnable }} </directive>
       <!-- <directive> -j oe {{ output_error_path }} </directive> -->
       <directive> -j oe </directive>
-      <directive default="ae"  > -m {{ mail_options }} </directive>
       <directive> -V </directive>
+      <directive> -M {{ mail_user }}</directive>
+      <directive default="ae"> -m {{ mail_type }}</directive>
     </directives>
   </batch_system>
 
@@ -126,7 +128,8 @@
       <directive> -N {{ job_id }}</directive>
       <directive> -j oe </directive>
       <directive default="n"> -r {{ rerunnable }} </directive>
-      <directive default="ae"  > -m {{ mail_options }} </directive>
+      <directive> -M {{ mail_user }}</directive>
+      <directive default="ae"  > -m {{ mail_type }} </directive>
       <directive default="/bin/bash" > -S {{ shell }}</directive>
     </directives>
   </batch_system>
@@ -171,6 +174,8 @@
        <directive> --nodes={{ num_nodes }}</directive>
        <directive> --output={{ output_error_path }}.%j </directive>
        <directive> --exclusive                        </directive>
+       <directive> --mail-user={{ mail_user }}</directive>
+       <directive default="end,fail"> --mail-type={{ mail_type }}</directive>
      </directives>
    </batch_system>
 
--- a/config/acme/machines/config_batch.xml
+++ b/config/acme/machines/config_batch.xml
@@ -41,7 +41,7 @@
     <depend_string> --dependencies</depend_string>
     <walltime_format>%H:%M:%s</walltime_format>
     <batch_mail_flag>-M</batch_mail_flag>
-    <batch_mail_type_flag></batch_mail_type>
+    <batch_mail_type_flag></batch_mail_type_flag>
     <batch_mail_type></batch_mail_type>
     <submit_args>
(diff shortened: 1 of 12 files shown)


scripts/create_newcase
```diff
@@ -171,7 +171,7 @@ def _main_func(description):
         user_mods_dir, user_compset, pesfile, \
         user_grid, gridfile, srcroot, test, ninst, walltime, queue, \
         output_root, script_root, run_unsupported, \
-        answer, input_dir = parse_command_line(sys.argv, cimeroot, description)
+        answer, input_dir, = parse_command_line(sys.argv, cimeroot, description)
```
jedwards4b: This addition seems wrong.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -431,6 +450,69 @@ def _submit_single_job(self, case, job, depid=None, no_batch=False, skip_pnl=Fal
             logger.info("Submitted job id is {}".format(jobid))
             return jobid
 
+    def get_batch_mail_flag(self, batch_system):
```
jedwards4b: These should be in data not in python.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -431,6 +450,69 @@ def _submit_single_job(self, case, job, depid=None, no_batch=False, skip_pnl=Fal
...
+            'cobalt' : '-M',
+            'cobalt_theta' : '-M',
+            'pbs' : '-M',
+            'slurm' : '--mail-user',
+            'lsf' : '-u',
+        }
+
+        return mail_user_flag[batch_system]
+
+    def get_batch_mail_type(self, batch_system, mail_type='never'):
```
jedwards4b: Should not be defined in python - should be in config_batch.xml

scripts/create_newcase
erichlf: Not even sure how that got there. I don't remember doing that, but clearly I must have.

config/acme/machines/config_batch.xml
```diff
@@ -77,6 +77,7 @@
       <arg flag="-q" name="$JOB_QUEUE"/>
       <arg flag="-W" name="$JOB_WALLCLOCK_TIME"/>
       <arg flag="-P" name="$PROJECT"/>
+      <arg flag="-u" name="$MAIL_USER"/>
```
jgfouca: Remove these. Instead add a <batch_email_flag> attribute to the batch system.

Example
<batch_email_flag>u</batch_email_flag>

config/acme/machines/config_batch.xml
```diff
@@ -100,13 +101,14 @@
       <arg flag="-q" name="$JOB_QUEUE"/>
       <arg flag="-l walltime=" name="$JOB_WALLCLOCK_TIME"/>
       <arg flag="-A" name="$PROJECT"/>
+      <arg flag="-M" name="$MAIL_USER"/>
+      <arg flag="-m" name="$MAIL_TYPE"/>
```
jgfouca: Remove these. Instead add a <batch_email_type_flag> attribute to the batch system.

Example:
<batch_email_type_flag>, aeb, b, e, a</batch_email_type_flag>

scripts/lib/CIME/XML/env_batch.py
jgfouca: Yes, should be:

```
self.get_value("batch_email_flag")
```

scripts/lib/CIME/XML/env_batch.py
jgfouca: Yes, should be:

```
raw =  self.get_value("batch_email_type_flag")
items = [item.strip() for item in raw.split(",")]
idx = ["never", "all", "begin", "end", "fail"].index(mail_type)
return items[idx]
```

erichlf:
I am testing things, so don't merge yet.

rljacob:
Erich, are you using xmllint?

jedwards4b:
@rljacob I've confirmed that he used xmllint and updated the schema with the additional fields.
@erichlf Thanks for adding these changes to the cesm side as well.

rljacob:
Ok.  I'd suggest re-writing the history of this branch to remove some of the typo commits.

erichlf:
There is still at least one issue that I am working out, so don't merge this yet.

erichlf:
Okay, things should be finally working.

Issue #1718, opened 2017-06-30
Tests should work with longname not just alias.


jedwards4b:
Because we have chosen to minimize defined compset aliases we are in a situation in which we might regularly run compsets using the longname only.  We need the tests to work on any compset not just a subset defined by aliases.   I would like to add an option to create test --compset that will allow you to give a compset longname.   This means that the alias that is an integral part of the testname will have to have a special value - how about 'LONGNAME' - so for example a test might be run with 
```
./create_test SMS.f19_g17.LONGNAME.cheyenne_intel --compset AMIP_CAM50_CLM45%SP_CICE%PRES_DOCN%DOM_RTM_SGLC_SWAV
```

gold2718:
Is there any reason the longname can't just be included in the test name? E.g.,
```
./create_test SMS.f19_g17.AMIP_CAM50_CLM45%SP_CICE%PRES_DOCN%DOM_RTM_SGLC_SWAV.cheyenne_intel
```
Is the period an allowed character in compset names? I'm not sure we check for it but no one uses it so we could just declare it character _non grata_.

Issue #1706, opened 2017-06-25
--user-compset should imply --run-unsupported


gold2718:
In my opinion, if a user is entering expert mode by specifying `--user-compset`, then they should not also have to specify `--run-unsupported`. It seems to me that any user using this feature intends to be running unsupported code and even if they happen to specify an existing compset (supported or not). We should trust that they already signed the waiver that `--run-unsupported` represents.
Comments from @mvertens? @billsacks? @jedwards4b? Anyone at ACME (@jgfouca?).

jedwards4b:
1. --run-unsupported is a cesm only feature and does not affect acme.  
2. We had considerable discussion about this when the run-unsupported flag was implemented before deciding that using both would be required.

billsacks:
@gold2718 - thanks for bringing this up. I agree that you shouldn't have to specify both. I'd actually argue for ditching the `--user-compset` option entirely at this point, and _only_ requiring `--run-unsupported`. (I'm pretty sure that `--user-compset` predates `--run-unsupported`, and I don't ever remembering having a discussion about whether it's necessary to retain `--user-compset` now that we have the more general `--run-unsupported`.)

gold2718:
I am fine with ditching `--user-compset`.

billsacks:
Everyone at today's CSEG meeting was also fine with ditching `--user-compset`, just keeping `--run-unsupported` in this case.

PR #1721, opened 2017-06-30
eleminate --user-compset, allow longname in tests

jedwards4b:
Remove the --user-compset arg, it's no longer needed.   Allow create_test to accept compset longname as part of the testname, add test to scripts_regression_tests.py of these features.
I also change some more internal cime calls to xmlquery to use --value instead of -value.

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes ESMCI/cime#1718 
Fixes ESMCI/cime#1706 

User interface changes?: Remove --user-compset from create_newcase, allow compset longname in testname argument to create_test 

Code review:

### DIFF
--- a/config/cesm/machines/Makefile
+++ b/config/cesm/machines/Makefile
@@ -27,8 +27,8 @@ VPATH := $(subst $(space),:,$(VPATH))
 RM    := rm
 CP    := cp
 
-exec_se: $(EXEC_SE)  $(CURDIR)/Depends
-complib: $(COMPLIB)  $(CURDIR)/Depends
+exec_se: $(EXEC_SE)  Depends
+complib: $(COMPLIB)  Depends
 
 # Determine whether to compile threaded or not
 # Set the THREADDIR for the shared build
@@ -98,26 +98,26 @@ ifeq ($(compile_threaded), true)
 endif
 
 ifeq (,$(EXEROOT))
-  EXEROOT = $(shell ./xmlquery EXEROOT -value)
+  EXEROOT = $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) EXEROOT --value)
 endif
 ifeq (,$(BUILD_THREADED))
-  BUILD_THREADED = $(shell ./xmlquery BUILD_THREADED -value)
+  BUILD_THREADED = $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) BUILD_THREADED --value)
 endif
 
 ifeq (,$(LIBROOT))
-  LIBROOT = $(shell ./xmlquery LIBROOT -value)
+  LIBROOT = $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) LIBROOT --value)
 endif
 ifeq (,$(SHAREDLIBROOT))
-  SHAREDLIBROOT = $(shell ./xmlquery SHAREDLIBROOT -value)
+  SHAREDLIBROOT = $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) SHAREDLIBROOT --value)
 endif
 ifeq (,$(COMPILER))
-  COMPILER = $(shell ./xmlquery COMPILER -value)
+  COMPILER = $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) COMPILER --value)
 endif
 ifeq (,$(NINST_VALUE))
-  NINST_VALUE = $(shell ./xmlquery NINST_VALUE -value)
+  NINST_VALUE = $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) NINST_VALUE --value)
 endif
 ifeq (,$(MPILIB))
-  MPILIB = $(shell ./xmlquery MPILIB -value)
+  MPILIB = $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) MPILIB --value)
 endif
 ifeq ($(strip $(PIO_VERSION)),1)
   CPPDEFS += -DPIO1
@@ -616,7 +616,7 @@ $(PIO_LIBDIR)/Makefile:
 #-------------------------------------------------------------------------------
 
 touch_filepath:
-	touch $(CURDIR)/Filepath
+	touch Filepath
 
 # Get list of files and build dependency file for all .o files
 #   using perl scripts mkSrcfiles and mkDepends
@@ -626,19 +626,19 @@ BASENAMES := $(basename $(basename $(SOURCES)))
 OBJS    := $(addsuffix .o, $(BASENAMES))
 INCS    := $(foreach dir,$(cpp_dirs),-I$(dir))
 
-CURDIR := $(shell pwd)
+CASETOOLS := $(CASEROOT)/Tools
 
-$(CURDIR)/Depends: $(CURDIR)/Srcfiles $(CURDIR)/Deppath
+Depends: Srcfiles Deppath
 	$(CASETOOLS)/mkDepends $(USER_MKDEPENDS_OPTS) Deppath Srcfiles > $@
 
-$(CURDIR)/Deppath: $(CURDIR)/Filepath
-	$(CP) -f $(CURDIR)/Filepath $@
+Deppath: Filepath
+	$(CP) -f Filepath $@
 	@echo "$(MINCROOT)" >> $@
 
-$(CURDIR)/Srcfiles: $(CURDIR)/Filepath
+Srcfiles: Filepath
 	$(CASETOOLS)/mkSrcfiles
 
-$(CURDIR)/Filepath:
+Filepath:
 	@echo "$(VPATH)" > $@
 
 
@@ -852,7 +852,7 @@ realclean: clean cleancsmshare cleanpio cleanmct cleangptl
 ifneq ($(MAKECMDGOALS), db_files)
 ifneq ($(MAKECMDGOALS), db_flags)
 ifeq (,$(findstring clean,$(MAKECMDGOALS)))
-    -include $(CURDIR)/Depends $(CASEROOT)/Depends.$(COMPILER) $(CASEROOT)/Depends.$(MACH)
+    -include Depends $(CASEROOT)/Depends.$(COMPILER) $(CASEROOT)/Depends.$(MACH)
 endif
 endif
 endif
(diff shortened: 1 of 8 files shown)


config/cesm/machines/Makefile
```diff
@@ -852,7 +852,7 @@ realclean: clean cleancsmshare cleanpio cleanmct cleangptl
 ifneq ($(MAKECMDGOALS), db_files)
 ifneq ($(MAKECMDGOALS), db_flags)
 ifeq (,$(findstring clean,$(MAKECMDGOALS)))
-    -include $(CURDIR)/Depends $(CASEROOT)/Depends.$(COMPILER) $(CASEROOT)/Depends.$(MACH)
+    -include Depends $(CASEROOT)/Depends.$(COMPILER) $(CASEROOT)/Depends.$(MACH)
 endif
```
gold2718: Are the changes to Makefile required for this PR or they unrelated?

src/build_scripts/buildlib.csm_share
```diff
@@ -13,22 +13,22 @@ my ($sharedlibroot, $installroot, $CASEROOT) = @ARGV;
...
+my $NINST_ICE		= `./xmlquery NINST_ICE		--value`;
+my $NINST_GLC		= `./xmlquery NINST_GLC		--value`;
+my $NINST_LND		= `./xmlquery NINST_LND		--value`;
+my $NINST_OCN		= `./xmlquery NINST_OCN		--value`;
+my $NINST_ROF		= `./xmlquery NINST_ROF		--value`;
+my $NINST_WAV		= `./xmlquery NINST_WAV		--value`;
+my $NINST_ESP		= `./xmlquery NINST_ESP		--value`;
+my $NINST_VALUE		= `./xmlquery NINST_VALUE	--value`;
+$ENV{PIO_VERSION}  	= `./xmlquery PIO_VERSION	--value`;
 #--------------------------------------------------------------------
```
gold2718: These changes do not appear to be related to the PR subject. Please either use a separate PR or document the change in the PR text.

config/cesm/machines/Makefile
jedwards4b: Yes, % symbol in the compset longname is not appreciated by gnu make, but these explicit CURDIR fields aren't needed anyway.

src/build_scripts/buildlib.csm_share
jedwards4b: PR text updated.

gold2718:
Need citation for new word, "eleminate" #covfefe :)

rljacob:
@jedwards4b, for the full name of an issue id, don't put a space before the # sign.
"Esmci/cime#1718"  not "Esmci/cime #1718"

gold2718:
Suggestion, make [.] illegal in compset names. Above line 546 in case.py, add:
```
        # Period not allowed anywhere in a compset name
        expect("." not in compset,
               "ERROR: period (.) character not allowed in compset name")
```
@mvertens, any reason to not make this change? Checking cesm2_0_alpha07, I don't see any current usage of that character.

scripts/tests/scripts_regression_tests.py
```diff
@@ -1299,6 +1299,12 @@ def test_update_acme_tests_test_mods(self):
         #                  msg="COMMAND SHOULD HAVE WORKED\ncs.status output:\n%s\n\nerrput:\n%s\n\ncode: %d" % (output, errput, stat))
 
 ###############################################################################
+class W_TestCreateTest(TestCreateTestCommon):
+###############################################################################
+    def test_create_test_longname(self):
```
jgfouca: Why not add this to the existing CreateTest test class?

jedwards4b:
Since the testname needs to be parsed before the long compset name is an extra '.' in the compset name in a call to create_test will result in a create_test error: 
```
./create_test SMS.f19_g16.2000_DATM%QI.A_XLND_SICE_SOCN_XROF_XGLC_SWAV
ERROR: Expected 4th item of 'SMS.f19_g16.2000_DATM%QI.A_XLND_SICE_SOCN_XROF_XGLC_SWAV' ('A_XLND_SICE_SOCN_XROF_XGLC_SWAV') to be in form machine_compiler
```
I have added a check in case.py for valid compset names but the create_test code will fail before that point.

scripts/tests/scripts_regression_tests.py
jedwards4b: Maybe I'm just being dense - but I don't see one.  Which class specifically?

jgfouca: You're right. The class I was thinking of was TestTestScheduler. That said, this is probably the kind of test I'd put in K_TestCimeCase.

gold2718:
Sigh, 

It does seem like the logic in `parse_test_name` could be tightened up a bit. Can there really be more than 3 dots? Still, if the dot in the compset name is in the right place (e.g., a dot in the GLC component description), it could even pass the test that is in there but be bogus. You could add a check for a legal compset name (maybe add that as a function in utils.py and also call it from case.py) before deciding whether the test name is valid, and if invalid, where the fault probably lies.

BTW, I notice that a common usage of create_test (with a single positional argument) is not in the usage statement. Seems lacking.

billsacks:
Thanks for doing this, @jedwards4b - the ability to use a compset long name in a test will be very helpful! Did you end up using the solution proposed by @gold2718 https://github.com/ESMCI/cime/issues/1718#issuecomment-312312147 of putting the full compset longname in the test name? (That solution makes sense to me.)

jedwards4b:
@billsacks: Yes the full compset longname is in the test name.
@gold2718 I'm not sure what else you want in this PR, could you please clarify.
@jgfouca I've moved the test as suggested, retesting now.

gold2718:
@jedwards4b,

> I'm not sure what else you want in this PR, could you please clarify.

Here is my request (copied from above):

> You could add a check for a legal compset name (maybe add that as a function in utils.py and also call it from case.py) before deciding whether the test name is valid, and if invalid, where the fault probably lies.

To be more specific:
Add a new function, `parse_compset_name`, to utils.py which performs a function similar to `parse_test_name`. `parse_compset_name` will return a list of all the components but also perform error checking (e.g., no dot allowed anywhere in the compset name).
`parse_compset_name` should be called from `parse_test_name` as well as from `get_compset_components` in case.py as well as `_get_description_v3` and `_get_description_match` in component.py. This will ensure that compsets are always parsed and error-checked in a consistent manner.

jedwards4b:
@gold2718 and my reply was that this is a chicken and egg problem.  The code you propose is already there but the parse test name code must be done before the parse compset name code can be called.

gold2718:
@jedwards4b, the reply you reference is not in this conversation after my 
I do not understand why `parse_test_name` cannot call `parse_compset_name` on `rv[2]` before further processing anything in the test name. Also, the rules can be tightened because at least two dots are required (so the expect `num_dots <= 4` should have an accompanying expect of `num_dots > 1`).
While this is not perfect, if `parse_case_name` has sufficient error checking, it will eliminate many of the possible bogus test names which will currently either pass or result in confusing error messages. As a bonus, we have one place where compset names are parsed and checked and that function can have unit test cases built in.

jedwards4b:
@gold2718 I have added tests that may clarify the problem for you, of course future contributions which improve error checking will always be welcome.  This routine does not require that num_dots > 1 as illustrated by the test at line 303 of utils.py

gold2718:
@jedwards4b 
Your first new test case makes my point that it complains about the machine_compiler format when the problem is in fact the compset name. An independent compset name checker would catch most of these errors.
Second, your test on line 303 passes but this is in contradiction to the documentation in both the function (line 300) and in the CIME documentation (http://esmci.github.io/cime/users_guide/testing.html). Either the test case or the documentation is wrong -- please rectify this situation in this PR. The documentation (in both places) requires three elements which implies that two dots are required.

gold2718:
@jedwards4b, if you do not think a common compset name checking function is an improvement (to the current situation of compset names being parsed in three different modules with different error checking), please state your objection here so I can understand.

jedwards4b:
@gold2718 Please explain to me how I would extract and check the compsetname from the testname without first parsing the testname?  

Partial names ('ERS','ERS.f19_f19' are allowed here
and this code is used to parse partial testnames so a check on minimum number of dots would not work.   I have updated the documentation to reflect this.   Note that even if I changed the code to abort on incomplete test names it would not catch the error of an extra dot in the compsetname. 

A more complete check of the compset name is done in create_newcase after the testname has been parsed.

gold2718:
> Please explain to me how I would extract and check the compsetname from the testname without first parsing the testname?

I gave an explicit method above which you adopted with the `check_name` call on line 341.

jedwards4b:
And that works for the second new test, but the first one has an addtional dot and so the compset name is interpreted as 2000_DATM%QI
I believe the resulting error:
   ERROR: Expected 4th item of 'SMS.f19_g16.2000_DATM%QI.A_XLND_SICE_SOCN_XROF_XGLC_SWAV.mach-ine_compiler.test-mods' ('A_XLND_SICE_SOCN_XROF_XGLC_SWAV') to be in form machine_compiler
adequately informs the user as to the problem.

gold2718:
> Partial names ('ERS','ERS.f19_f19' are allowed here
> and this code is used to parse partial testnames so a check on minimum number of dots would not work.

I'm trying to figure out where this happens. One place is the function `_xml_phase`, is this function called anywhere? Why does this function call `parse_test_name` twice for the same info?
Is there any other place where `parse_test_name` is called where a partial test name is okay?

gold2718:
@jedwards4b,

If you had called a compset name checking function on the bogus compset field, `2000_DATM%QI`, it should have kicked an error since compset names currently require more than one underscore. While this is not a perfect check (I think compsets can currently range from 7 to 9 underscores), it would catch the majority of issues and point at the correct problem.

gold2718:
I'm willing to forgo more detailed compset name checking for this PR, please just show me where partial test names are used.

jedwards4b:
@gold2718 A compset name check at this point would also have to allow compset aliases and so a check for the number of underscores wouldn't work.    Partial test names are used in get_full_test_name which is in turn called by the acme test name parsing tools in update_acme_tests.py and scripts_regression_tests.  @jgfouca probably knows more about usage.

gold2718:
Yow, I guess `parse_test_name` is too overloaded for any user-level error checking. How about adding error checking (at least between 2 and 4 dots) after line 286 in create_test?
```
    for test in args.testargs:
        numdots = test.count(".")
        expect(numdots > 1 and numdots <= 4, "Invalid test name, '{}'".format(test))
```

PR #1725, opened 2017-07-07
fix error in path to env_mach_specific.xml file in case

### LINKED ISSUE #1607: compare_test_results generates error, "Could not find a matching MPI for attributes"
Out of cime master (on yellowstone), I did:

```
./create_test --test-root . --walltime 0:30 ERS_Ld3.f45_g37_rx1.A.yellowstone_intel
```

then when it completed

```
Tools/compare_test_results -b cesm2_0_alpha06m --baseline-root $CESMDATAROOT/ccsm_baselines -r . -t 20170525_094940_daztcr
```

This gave:

```
ERROR: Could not find a matching MPI for attributes: {'mpilib': 'mpich2', 'threaded': False, 'compiler': 'intel'}
```

I just ran into similar errors with somewhat older versions of cime - going back to at least cime5.3.0-alpha.13

jedwards4b:
The fullpath to env_mach_specific.xml was incorrectly evaluated in the init method in env_mach_specific.py
it doesn't need to be done here at all as it is done in env_base.py (the parent class)

Test suite: scripts_regression_tests.py, compare_test_results
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1607 

User interface changes?: 

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/env_mach_specific.py
+++ b/scripts/lib/CIME/XML/env_mach_specific.py
@@ -18,9 +18,8 @@ class EnvMachSpecific(EnvBase):
         """
         initialize an object interface to file env_mach_specific.xml in the case directory
         """
-        fullpath = infile if os.path.isabs(infile) else os.path.join(caseroot, infile)
         schema = os.path.join(get_cime_root(), "config", "xml_schemas", "env_mach_specific.xsd")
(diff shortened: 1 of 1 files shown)


billsacks:
I have verified that this fixes my compare_test_results problem - thanks!

Issue #1737, opened 2017-07-13
system_tests_compare_two broken for hybrid runs


jedwards4b:
If the compset is defined with RUN_TYPE = hybrid the initial startup files are not copied or linked into the second run directory and the second run fails.   Tested with
PET_Ln9.f09_f09_mg17.F2000_DEV.cheyenne_intel.cam-outfrq9s

PR #1738, opened 2017-07-13
System tests compare two stage two

jedwards4b:
The second test in a system_tests_compare_two type test was not getting its initial data in a RUN_TYPE='hybrid' case. 

Test suite: scripts_regression_tests.py & PET_Ln9.f09_f09_mg17.F2000_DEV.cheyenne_intel.cam-outfrq9s 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes ESMCI/cime#1737

User interface changes?: No

Code review:

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -26,6 +26,8 @@ In addition, they MAY require the following method:
 from CIME.XML.standard_module_setup import *
 from CIME.SystemTests.system_tests_common import SystemTestsCommon
 from CIME.case import Case
+from CIME.check_input_data import check_all_input_data
+from CIME.preview_namelists import create_namelists
 
 import shutil, os, glob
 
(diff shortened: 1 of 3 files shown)


billsacks:
I was trying to understand why these changes are needed, and I think I see why (correct me if I'm wrong): In a normal user workflow, the cloned case would be submitted with case.submit, whereas in system_tests_compare_two, we just run the second case without going through case.submit. When you go through case.submit, then case_submit:check_case gets called, and this has the necessary calls to create_namelists and check_all_input_data.

If that's the right reason, then would it make sense to have system_tests_compare_two call case_submit:check_case rather than duplicating part of the body of check_case here? It seems that would be easier to maintain moving forward, in case other things need to be added to the check_case body. One thing I'm not sure about is whether some other parts of check_case – specifically `check_lockedfiles` and the check of `BUILD_COMPLETE` – would ever cause problems if called from here.

billsacks:
Thanks. I'm happy with these changes. Have you confirmed that tests still pass with the most recent commit?

jedwards4b:
I've rerun scripts_regression_tests.py, still waiting for the PET test to complete.

jedwards4b:
testing complete - all good.

PR #1742, opened 2017-07-14
Adding components

jedwards4b:
Start a section on adding new components to cime.   This still needs a lot of work.

Test suite:  None - documentation only
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes

User interface changes?: 

Code review:

### DIFF
--- a/doc/source/users_guide/adding-cases.rst
+++ b/doc/source/users_guide/adding-cases.rst
@@ -151,7 +151,3 @@ The steps for adding a new component grid to the model system follow. gain, this
       Test the new grid with CAM(newgrid), CLM(newgrid), DOCN(gx1v6), DICE(gx1v6)
       (write an example)
 
-===================
-Adding components
-===================
-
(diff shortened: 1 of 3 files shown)


billsacks:
@kauff - you may be interested in this and/or able to contribute, with your ISSM integration work.

rljacob:
We definitely need this but I feel like its to much for Part II of the user's guide.   I see 3 classes of user:  The "pure" CIME user who only uses the scripts and xml files of a cime-driven model,  the coupled model developer that wants to  make their model part of a  "cime-driven" model and the cime developer who develops the core CIME python and Fortran services.

bertinia:
I agree with Rob and will work on splitting the User's Guide into 3 sections as suggested. 
For now, I'm going to merge this PR and build the html in the gh-pages branch so that other documentation work can be continuously 
integrated into the master like the testing section that @fischer-ncar is working on and
the grids that @mnlevy1981 is working on.

rljacob:
The current users's guide and its 2 parts is fine for the first class of user.   The separate "CIME programmers guide" is for the last class.   The middle one was a separate guide called "Buildling a Coupled Model with CIME"  and has a stub here:  http://esmci.github.io/cime/build_cpl/index.html

PR #1744, opened 2017-07-17
Implement NCR with SystemTestsCompareTwo

### LINKED ISSUE #444: Convert NCR test to SystemTestsCompareTwo infrastructure

mfdeakin-sandia:
This reimplements the NCR test with the SystemTestsCompareTwo infrastructure to significantly simplify the test. ACME currently fails the diff test; but at least it builds unlike on master.

Test suite: scripts_regression_tests - does not actually test NCR (AFAICT); so just going by the B_CheckCode test
Test baseline: 
Test namelist changes: 
Test status: PASS 

Fixes #444 

User interface changes?: N

Code review: @jgfouca @jedwards4b @billsacks

### DIFF
--- a/scripts/lib/CIME/SystemTests/ncr.py
+++ b/scripts/lib/CIME/SystemTests/ncr.py
@@ -9,103 +9,48 @@ import shutil
 from CIME.XML.standard_module_setup import *
 from CIME.case_setup import case_setup
 import CIME.utils
-from CIME.SystemTests.system_tests_common import SystemTestsCommon
+from CIME.SystemTests.system_tests_compare_two import SystemTestsCompareTwo
 from CIME.check_lockedfiles import *
 
 logger = logging.getLogger(__name__)
 
-class NCR(SystemTestsCommon):
+class NCR(SystemTestsCompareTwo):
 
     def __init__(self, case):
         """
-        initialize a test object
+        initialize an NCR test
         """
-        SystemTestsCommon.__init__(self, case)
+        SystemTestsCompareTwo.__init__(self, case,
+                                       separate_builds = True,
+                                       run_two_suffix = "multiinst",
+                                       run_one_description = "default build",
+                                       run_two_description = ("half the number of tasks, " +
+                                                              "twice the number of instances"))
 
-    def build_phase(self, sharedlib_only=False, model_only=False):
-        exeroot = self._case.get_value("EXEROOT")
-        cime_model = CIME.utils.get_model()
+    def _common_setup(self):
+        pass
 
-        machpes1 = "env_mach_pes.NCR1.xml"
-        if is_locked(machpes1):
-            restore(machpes1, newname="env_mach_pes.xml")
+    def _case_one_setup(self):
+        # Set the number of instances, the ROOTPEs, and the number of tasks
+        # The first case should have mostly default settings;
+        # though we apparently halve the number of tasks if greater than 1
+        for comp in ['ATM','OCN','WAV','GLC','ICE','ROF','LND']:
+            self._case.set_value("NINST_{}".format(comp), str(1))
+            self._case.set_value("ROOTPE_{}".format(comp), 0)
+            ntasks = self._case.get_value("NTASKS_{}".format(comp))
+            if ntasks > 1:
+                self._case.set_value("NTASKS_{}".format(comp), ntasks // 2)
+        case_setup(self._case, test_mode = True, reset = True)
 
-        # Build two exectuables for this test, the first is a default build, the
-        # second halves the number of tasks and runs two instances for each component
-        # Lay all of the components out concurrently
-        for bld in range(1,3):
-            logging.warn("Starting bld {}".format(bld))
-            machpes = "env_mach_pes.NCR{}.xml".format(bld)
-            ntasks_sum = 0
-            for comp in ['ATM','OCN','WAV','GLC','ICE','ROF','LND']:
-                self._case.set_value("NINST_{}".format(comp), str(bld))
-                ntasks      = self._case.get_value("NTASKS_{}".format(comp))
-                if(bld == 1):
-                    self._case.set_value("ROOTPE_{}".format(comp), 0)
-                    if ( ntasks > 1 ):
-                        self._case.set_value("NTASKS_{}".format(comp), ntasks/2)
-                else:
-                    self._case.set_value("ROOTPE_{}".format(comp), ntasks_sum)
-                    ntasks_sum += ntasks*2
-                    self._case.set_value("NTASKS_{}".format(comp), ntasks*2)
-            self._case.flush()
-
-            case_setup(self._case, test_mode=True, reset=True)
-            self.clean_build()
-            self.build_indv(sharedlib_only, model_only)
-            shutil.move("{}/{}.exe".format(exeroot,cime_model),
-                        "{}/{}.exe.NCR{}".format(exeroot,cime_model,bld))
-            lock_file("env_build.xml", newname="env_build.NCR{}.xml".format(bld))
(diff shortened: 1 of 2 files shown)


mfdeakin-sandia:
Note that I was unable to compare the output to the original NCR test, as I was getting sharedlib build failures

scripts/lib/CIME/SystemTests/ncr.py
```diff
@@ -1,111 +1,64 @@
...
+
+        for comp in comp_classes:
+            self._case.set_value("NINST_{}".format(comp), str(1))
+            self._case.set_value("ROOTPE_{}".format(comp), 0)
+            ntasks = self._case.get_value("NTASKS_{}".format(comp))
+            if ntasks > 1:
+                self._case.set_value("NTASKS_{}".format(comp), ntasks // 2)
+        case_setup(self._case, test_mode = True, reset = True)
+
+    def _case_one_setup(self):
```
jedwards4b: Can you put case_one first - just for continuity when you read the file.

mfdeakin-sandia: Done

scripts/lib/CIME/SystemTests/ncr.py
```diff
@@ -1,111 +1,61 @@
...
+
+        ntasks_sum = 0
+        # ['ATM','OCN','WAV','GLC','ICE','ROF','LND']
+        for comp in comp_classes:
+            self._case.set_value("NINST_{}".format(comp), str(1))
+            self._case.set_value("ROOTPE_{}".format(comp), ntasks_sum)
+            ntasks = self._case.get_value("NTASKS_{}".format(comp))
+            ntasks_sum += ntasks * 2
+            self._case.set_value("NTASKS_{}".format(comp), ntasks * 2)
+        case_setup(self._case, test_mode = False, reset = True)
```
billsacks: Why is test_mode False here? Please add a comment describing this unintuitive setting.

scripts/lib/CIME/SystemTests/ncr.py
```diff
@@ -1,111 +1,61 @@
 """
 Implementation of the CIME NCR test.  This class inherits from SystemTestsCommon
 
-Build two exectuables for this test, the first is a default build the
-second halves the number of tasks and runs two instances for each component
-Lay all of the components out concurrently
+Build two exectuables for this test:
+The first is a default build
+The second runs two instances for each component with the same total number of tasks,
+and runs each of them concurrently
```
billsacks: Given that (as you said) you haven't been able to fully test this, can you please add a comment here - and ideally also in the config_tests.xml file - noting that this is untested and may not be working quite right?

scripts/lib/CIME/SystemTests/ncr.py
```diff
@@ -1,111 +1,61 @@
...
+        comp_classes = self._case.get_values("COMP_CLASSES")
+        # Is this correct?
+        comp_classes.remove("CPL")
+
+        for comp in comp_classes:
+            self._case.set_value("NINST_{}".format(comp), str(1))
+            self._case.set_value("ROOTPE_{}".format(comp), 0)
+            ntasks = self._case.get_value("NTASKS_{}".format(comp))
+            if ntasks > 1:
+                self._case.set_value("NTASKS_{}".format(comp), ntasks // 2)
```
billsacks: I think it's wrong to both halve the number of tasks here and double the number of tasks in case_one. I think you want to do one or the other. My preference would be for following what's done in the NCK test: halving ntasks in both initially, then doubling them in the case that runs 2 instances - so that the case that runs 2 instances ends up with the original task count (though slightly different, by design, if the original task count was odd).

scripts/lib/CIME/SystemTests/ncr.py
jedwards4b: test_mode is False here because the test_mode flag indicates whether or not the case.test file should be updated (if test_mode is true it is not updated)

mfdeakin-sandia: @jedwards4b Could you clarify this for me? All I know is that it was running with the wrong number of nodes when it was True.

billsacks: @jedwards4b is there a general rule for when the case.test file should or should not be updated?

I'm especially wondering why test_mode is False here despite being True for both cases in the NCK test.

jedwards4b: If the two test runs have different pe counts as this one does then the run with the larger value should write the case.test file and the other one should not overwrite it.  That way the batch system will allocate enough resources for the test.

billsacks: Ah, great, thanks @jedwards4b . @mfdeakin-sandia can you please add Jim's comments as comments in the code?

scripts/lib/CIME/SystemTests/ncr.py
mfdeakin-sandia: Ah, I misunderstood the old code - I was treating it like it had two separate env_run.xml files with the same initial paramters, as is the case with SystemTestsCompareTwo. But this makes much more sense and explains my confusion.

rljacob:
This comment deserves more discussion:  " ACME currently fails the diff test; but at least it builds unlike on master."

Does CESM run the NCR test in its suites?  Because it sounds like it never worked prior to this PR.

mfdeakin-sandia:
@rljacob The error I was getting was that acme.exe didn't exist in the build directory after SHAREDLIBBUILD, which seems like it would fail for CESM as well.

billsacks:
@rljacob and @mfdeakin-sandia as far as I can tell, the NCR test is currently unused.

rljacob:
How does CESM test multi-instance?  NCK?

billsacks:
> How does CESM test multi-instance? NCK?

Yes. NCK tested a different multi-instance configuration, which would probably be useful to test, but I guess we hadn't been testing it for a while.

mfdeakin-sandia:
I think I've fixed everything, is there anything else?

scripts/lib/CIME/SystemTests/ncr.py
```diff
@@ -1,111 +1,68 @@
...
-
-        logger.info("default: doing a {} {} with NINST2".format(stop_n, stop_option))
-        self.run_indv(suffix="multiinst")
-
-        # Compare
-        self._component_compare_test("base", "multiinst")
+        SystemTestsCompareTwo.__init__(self, case,
+                                       separate_builds = True,
+                                       run_two_suffix = "singleinst",
+                                       run_one_description = "default build",
```
billsacks: run_one_description and run_two_description are reversed, I think

mfdeakin-sandia: I have them reversed because I'm uncertain it'll work otherwise; but I'll try it and see if it works

mfdeakin-sandia: Nope; it crashes if I have case_two setup this way. I'll just fix the decriptions

billsacks: I wonder why this doesn't work: In general, tests implemented on top of system_test_compare_two shouldn't care which is case one and which is case two. The one thing I can think of is that maybe this is tied in with the need to set test_mode to False in case_one - e.g., maybe case one needs to be the one with more tasks.

billsacks: Anyway, I'm fine with your just fixing the description here.

mfdeakin-sandia: Yes; that's exactly the reason. I've added a comment explaining this

PR #1745, opened 2017-07-17
an addition to pr 1728 for v3 compatibility

jedwards4b:
Need to add description fields for new datasets

Test suite: I compset tests by hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Code review:

### DIFF
--- a/src/components/data_comps/datm/cime_config/config_component.xml
+++ b/src/components/data_comps/datm/cime_config/config_component.xml
@@ -10,11 +10,13 @@
        This file may have atm desc entries.
   -->
   <description modifier_mode="1">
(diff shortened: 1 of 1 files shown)


billsacks:
Reassigning to @ekluzek because I don't know what's right here. Erik: is this correct? In particular, is it correct to have both CRU and CRUv7, and both GSW and GSWP3v1?

jedwards4b:
It seems there is some confusion here: I see "CRU", "GSW" and "GSW3P" still defined in several places, if your intent with PR 1728 was to replace them you need to try again.

billsacks:
@bandre-ucar or @mvertens - can one of you look at @jedwards4b 's comments above, since one of you made the original changes, I think?

PR #1751, opened 2017-07-18
remove untested/unused scripts in Tools directory

jedwards4b:
These scripts in the Tools directory appear to be unused and are untested.

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


jedwards4b:
There has been a request to keep concat_daily_hist.csh

scripts/Tools/create_train
```diff
@@ -1,244 +0,0 @@
-#!/usr/bin/env perl
```
jedwards4b: This script is completely broken - the only reason to support trains is to satisfy the needs of some HPC centers that want you to run very large jobs.   I submit that we should remove this and add a new script if we at some future time have a need to support trains.

scripts/Tools/concat_daily_hist.csh
```diff
@@ -1,71 +0,0 @@
-#!/bin/csh -f
```
jedwards4b: It did, this was the only file for which support was expressed.   I will add it back.

scripts/Tools/check_exactrestart.pl
```diff
@@ -1,145 +0,0 @@
-#!/usr/bin/env perl
```
jedwards4b: We have this functionality in python, this is a duplication.

scripts/Tools/archive_metadata.sh
```diff
@@ -1,446 +0,0 @@
-
```
jedwards4b: okay

scripts/Tools/load.awk
```diff
@@ -1,40 +0,0 @@
-BEGIN {
```
jedwards4b: It's easy to add something back if needed, I've already put out a call far and wide for users to speak up, i don't want to do it again.

PR #1753, opened 2017-07-20
Add support for arbitrary prerequisites to case.submit

### LINKED ISSUE #1710: Support --prereq argument to case.submit
The --prereq argument is still giving an error.
```
blogin1[247]: ./case.submit --prereq 1662097
ERROR: --prereq not currently supported
```

We should support or temporarily remove it from --help string.

mfdeakin-sandia:
This PR implements the --prereq option for case.submit, and also has some minor fixes of env_batch.py for some machines

Test baseline: 
Test namelist changes: 
Test status: ok (tenatively; passes the ones I'd expect to fail)

Fixes #1710 

User interface changes?: N

### DIFF
--- a/config/acme/machines/config_batch.xml
+++ b/config/acme/machines/config_batch.xml
@@ -86,7 +86,7 @@
     <batch_redirect>&lt;</batch_redirect>
     <batch_directive>#BSUB</batch_directive>
     <jobid_pattern>&lt;(\d+)&gt;</jobid_pattern>
-    <depend_string>-w 'ended(jobid)'</depend_string>
+    <depend_string>-w 'done(jobid)'</depend_string>
     <depend_separator>&amp;&amp;</depend_separator>
     <walltime_format>%H:%M</walltime_format>
     <batch_mail_flag>-u</batch_mail_flag>
@@ -117,7 +117,7 @@
     <batch_env>-v</batch_env>
     <batch_directive>#PBS</batch_directive>
     <jobid_pattern>^(\S+)$</jobid_pattern>
-    <depend_string>-W depend=afterany:jobid</depend_string>
+    <depend_string>-W depend=afterok:jobid</depend_string>
     <depend_separator>:</depend_separator>
     <walltime_format>%H:%M:%S</walltime_format>
     <batch_mail_flag>-M</batch_mail_flag>
@@ -144,7 +144,7 @@
     <batch_cancel>canceljob</batch_cancel>
     <batch_directive>#MSUB</batch_directive>
     <jobid_pattern>(\d+)$</jobid_pattern>
-    <depend_string>-W depend=afterany:jobid</depend_string>
+    <depend_string>-W depend=afterok:jobid</depend_string>
     <depend_separator>:</depend_separator>
     <walltime_format>%H:%M:%S</walltime_format>
     <batch_mail_flag>-M</batch_mail_flag>
@@ -199,7 +199,7 @@
     <batch_cancel>scancel</batch_cancel>
     <batch_directive>#SBATCH</batch_directive>
     <jobid_pattern>(\d+)$</jobid_pattern>
-    <depend_string> --dependency=afterok:jobid</depend_string>
+    <depend_string>--dependency=afterok:jobid</depend_string>
     <depend_separator>:</depend_separator>
     <walltime_format>%H:%M:%S</walltime_format>
     <batch_mail_flag>--mail-user</batch_mail_flag>
--- a/config/acme/machines/config_batch.xml
+++ b/config/acme/machines/config_batch.xml
@@ -35,27 +35,28 @@
     </directives>
   </batch_system>
 
-   <batch_system type="cobalt" >
-     <batch_query>qstat</batch_query>
-     <batch_submit>qsub</batch_submit>
-     <batch_cancel>qdel</batch_cancel>
-     <batch_env>-v</batch_env>
-     <batch_directive></batch_directive>
-     <jobid_pattern>(\d+)</jobid_pattern>
-     <depend_string> --dependencies</depend_string>
-     <walltime_format>%H:%M:%s</walltime_format>
-     <batch_mail_flag>-M</batch_mail_flag>
-     <batch_mail_type_flag></batch_mail_type_flag>
-     <batch_mail_type></batch_mail_type>
-     <submit_args>
-       <arg flag="--cwd" name="CASEROOT"/>
-       <arg flag="-A" name="PROJECT"/>
-       <arg flag="-t" name="JOB_WALLCLOCK_TIME"/>
-       <arg flag="-n" name=" ($TOTALPES + $MAX_MPITASKS_PER_NODE - 1)/$MAX_MPITASKS_PER_NODE"/>
-       <arg flag="-q" name="JOB_QUEUE"/>
-       <arg flag="--mode script"/>
-     </submit_args>
-   </batch_system>
+  <batch_system type="cobalt" >
+    <batch_query>qstat</batch_query>
+    <batch_submit>qsub</batch_submit>
+    <batch_cancel>qdel</batch_cancel>
+    <batch_env>-v</batch_env>
+    <batch_directive></batch_directive>
+    <jobid_pattern>(\d+)</jobid_pattern>
+    <depend_string> --dependencies jobid</depend_string>
+    <depend_separator>:</depend_separator>
+    <walltime_format>%H:%M:%s</walltime_format>
+    <batch_mail_flag>-M</batch_mail_flag>
+    <batch_mail_type_flag></batch_mail_type_flag>
+    <batch_mail_type></batch_mail_type>
+    <submit_args>
+      <arg flag="--cwd" name="CASEROOT"/>
+      <arg flag="-A" name="PROJECT"/>
+      <arg flag="-t" name="JOB_WALLCLOCK_TIME"/>
+      <arg flag="-n" name=" ($TOTALPES + $MAX_MPITASKS_PER_NODE - 1)/MAX_MPITASKS_PER_NODE"/>
+      <arg flag="-q" name="JOB_QUEUE"/>
+      <arg flag="--mode script"/>
+    </submit_args>
+  </batch_system>
 
   <batch_system type="cobalt_theta" >
     <batch_query>qstat</batch_query>
@@ -64,7 +65,8 @@
     <batch_env>-v</batch_env>
     <batch_directive>#COBALT</batch_directive>
     <jobid_pattern>(\d+)</jobid_pattern>
-    <depend_string> --dependencies</depend_string>
+    <depend_string>--dependencies jobid</depend_string>
+    <depend_separator>:</depend_separator>
     <batch_mail_flag>-M</batch_mail_flag>
     <batch_mail_type_flag></batch_mail_type_flag>
     <batch_mail_type></batch_mail_type>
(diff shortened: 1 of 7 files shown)


mfdeakin-sandia:
I should note that I changed the depend_string to be after_any in many cases. This affects how CIME runs the archiving scripts; and I'm uncertain if this is reasonable for most people

mfdeakin-sandia:
@jgfouca Is there a specific place you'd recommend place you'd recommend placing the test in scripts_regression_tests.py?

jgfouca:
@mfdeakin-sandia TestCimeCase

config/acme/machines/config_batch.xml
```diff
@@ -454,8 +461,10 @@
     <job name="case.st_archive">
       <template>template.st_archive</template>
       <task_count>1</task_count>
-      <!-- If DOUT_S is true and case.run (or case.test) exits successfully then run st_archive-->
-      <dependency>case.run or case.test</dependency>
+      <!-- If DOUT_S is true and case.run amd/or case.test exits (regardless of success) then run st_archive
+	   This dependence is 'and' if both are run before case.st, and is 'or' if one is not run
+      -->
+      <dependency>case.run case.test</dependency>
```
jedwards4b: I don't think that this change is necessary,  what was missing in the original logic?

config/xml_schemas/config_batch.xsd
```diff
@@ -57,9 +58,12 @@
         <xs:element minOccurs="0" ref="jobid_pattern"/>
 
   <!-- depend_string: How to indicate to batch system that this job should not run until
-       a previous job has completed successfully -->
+       a previous job has completed (possibly unsuccessfully) -->
```
jedwards4b: Was there any discussion and agreement regarding changing this functionality?  I don't recall it.

mfdeakin-sandia: No; I've been asking; but didn't get a response. Is this an okay thing to do? I can make a more complicated change and add a depend_any string.

config/acme/machines/config_batch.xml
mfdeakin-sandia: The original logic completely ignored the "or"; as it was looking for job names. While this change wasn't necessary for this PR; implying functionality that we don't (and can't, AFAICT most batch systems don't support OR in dependencies) have seems bad to me.

jedwards4b: The logic for this was done in the cime python and not by the underlying batch system.  What indication did you have that this logic was ignored?   I don't think that you should implement this change.

mfdeakin-sandia: This was the logic:
`for dep in deps:
  if dep in depid.keys() and depid[dep] is not None:`
depid is a map containing the job names which were started previously. There wasn't anything handling the logical statements; just some commented out code below saying that we doubted they would be used.

mfdeakin-sandia:
@jedwards4b Did you not want to discuss the probably more disruptive change of always running case.st_archive if DOUT_S is true?

config/xml_schemas/config_batch.xsd
jedwards4b: @mfdeakin-sandia There is a problem with changing from completes successfully to completes - if RESUBMIT > 0 and 
the first job fails, in the old method the job would complete on the first failure.   Now it will continue to 
submit jobs until the RESUBMIT flag is decremented to 0.  I don't think that is what you intended can you find a way to run the st_archive but then not RESUBMIT if the first job fails?

jedwards4b:
I did, I just did things in the wrong order.  I don't have a problem with always running st_archive if DOUT_S is true, but I think that the side-effect of resubmitting if RESUBMIT > 0 even though the job failed is not something we want.

mfdeakin-sandia:
No, I intended to have the next job run regardless of failure - this is something scientists on ACME want, as it will either get to a point where it always fails quickly (not spend much CPU time), or it will recover from a machine hiccup or a model bug.

jedwards4b:
This is the point that needs wider discussion,  I'm okay with running the st archiver after a failed job,
but starting subsequent jobs as a default behavior could lead to corruption of the run directory and confusion as to what the original problem that caused the failure was.   If this feature is added it should be made an optional feature and not default behavior.  @mvertens @rljacob please comment.

mvertens:
@jedwards4 - I agree with your point.

rljacob:
Yes I agree it should be off by default.

mfdeakin-sandia:
I need to finish implementing the unit test @jgfouca asked for; I'm not certain the way I'm doing so is ideal (using a bunch of run_cmd_no_fail's to directly run case.submit), but hope to finish this today

mfdeakin-sandia:
@jedwards4b Could you review my implementation of the test so far? Also, do you have suggestions on a generic way of verifying the queue is actually respecting the prerequisite? From what I recall, most queue systems don't seem to support in depth queries about jobs, which is how I'd prefer to check this.

config/xml_schemas/config_batch.xsd
jedwards4b: This comment is now incorrect, the job must complete successfully.

scripts/tests/scripts_regression_tests.py
```diff
@@ -1399,6 +1403,29 @@ def test_cime_case(self):
             self.assertEqual(case.get_value("TESTCASE"), "TESTRUNPASS")
 
     ###########################################################################
+    def test_cime_case_prereq(self):
```
jedwards4b: It seems to me that you should be using the existing system test framework in CIME/SystemTests.
You could create a new test based on the SMS test in which job one is an initial run which writes a restart at the final time and job two is started with the prereq flag and is a CONTINUE_RUN which reads the restart from job 1.  This would test that job 2 does not start before job 1 is complete and that job 2 successfully completes.

mfdeakin-sandia: I'm somewhat concerned that this would depend on the queue system not fortuitously scheduling the jobs regardless of whether the prerequisites have been met.
The other issue is it's not clear to me how to get job 1's queue id? run_indv calls case_run, which assumes it's already on the queue.

jedwards4b: You could include a negative test - make the first run fail and assure that the second doesn't start.    As for the jobid, code env_batch.py has a subroutine get_job_id.   Since you will need to submit two separate jobs for this test you should look at the ERR test for an example.

mfdeakin-sandia: Sorry for the silence on this; I've been focusing on other work recently.
I'm not certain what the best way to implement a negative test for this is - querying the batch system to verify the job will never be run seems just as much work as checking that the batch system added the dependency, neither of which seem doable with CIME currently.
I'd rather not add a timeout which might not be met even with the prerequisite running successfully, so querying the batch system seems like something I'd have to do.

mfdeakin-sandia: @jedwards4b @jgfouca Would there be any suggestions on a robust method of testing this without querying the queue? Waiting for some threshold and verifying the dependent one didn't run seems dangerous, especially with the recent changes on skybridge. AFAICT, there isn't a robust method, so just doing the simple positive test seems reasonable.

mfdeakin-sandia: The tests I'm envisioning is as follows:
Submit a job (done)
While that job is running, submit a job with the prerequisite argument (the dependent job) for the currently running job (done)
Ensure the job does not run if the other job has not finished successfully yet. (not done)

I can add a check to the dependent job to verify the first job has finished, but if the dependent job was not submitted to the queue correctly due to a bug in the prerequisite logic, this test may still pass if it makes it onto a node after the original finished.
@jedwards4b suggested a negative test, force the original job to fail, and then verify the dependent job is never run. I could add a timeout and verify the job hasn't run by this time; but this depends on the queue not submitting after the timeout.

Another option would be to write a fake batch system as @billsacks suggested; this would be a separate Python script which outputs the arguments to a file so we can verify they're correct. Given all the xml needed to support this fake batch system, I'm not certain this is the way we want to go, but if it's helpful for other parts of the code that could justify it.

billsacks:
I have just skimmed through these changes, so could easily be missing something, but here are my $0.02.

I like to think, "What are some simple, isolated unit tests that would cover logic errors in the new code?" In this case, it looks like the new code is mainly focused on making some changes to the batch submission command. So it seems like the most straightforward way to test the new code would be to look at the batch submission command and compare it with an expected batch submission command and say, "does this string match the expected string?".

At a glance (again, maybe incorrect), it looks like you are instead trying to test this with a more system-y test which, in addition to testing your new code, also is incidentally testing (a) whether the xml file describing the batch system is set up correctly, and (b) whether this machine's batch system actually honors job dependencies correctly. There is a place for such tests, but in my view, that place is (or should be) small. See also https://testing.googleblog.com/2015/04/just-say-no-to-more-end-to-end-tests.html

mfdeakin-sandia:
@billsacks Thanks for the comment, you have good points, and the article was interesting. I'll just remove the tests I was making and instead see about adding doc-tests today, which I think will suffice for testing

mfdeakin-sandia:
@jedwards4b @jgfouca I think this is ready now

scripts/tests/scripts_regression_tests.py
```diff
@@ -1463,6 +1467,33 @@ def test_cime_case(self):
...
+                                     % (SCRIPT_DIR, testcase_name, testdir, testdir)),
+                              from_dir=SCRIPT_DIR)
+
+        with Case(testdir, read_only=False) as case:
+            job_name = "case.run"
+            prereq_name = 'prereq_test'
+            batch_cmd = case.submit_jobs(prereq=prereq_name, job=job_name, skip_pnl=True, dry_run=True)[0][1].split()
+            jobid_ident = 'jobid'
+            dep_str = case.get_env('batch').get_value('depend_string', subgroup=None)[:-len(jobid_ident)]
+            found_dep_str = False
```
jedwards4b: Lines 1483 and 1485 have a lot going on in them and assume a certain format of the return values, can you
document here what the expected values of batch_cmd and dep_str are?

scripts/tests/scripts_regression_tests.py
```diff
@@ -22,6 +22,10 @@
 from  CIME.code_checker import check_code, get_all_checkable_files
 from  CIME.test_status import *
 
+from  CIME.case_setup import case_setup
+from  CIME.case_submit import submit
+import CIME.build as build
```
jedwards4b: Did you run pylint on this?  It seems you are not using submit or build and can remove them.

mfdeakin-sandia: I knew I was forgetting something. Fixed, and now passes pylint

scripts/tests/scripts_regression_tests.py
mfdeakin-sandia: Good point. I added tests for what these should be

jgfouca:
@mfdeakin-sandia , please do an upstream merge to resolve conflicts.

rljacob:
And change the title to "verb noun".

mfdeakin-sandia:
Rebased - running scripts_regression_tests now

mfdeakin-sandia:
This passed scripts regression tests; I think it's ready

jgfouca:
@mfdeakin-sandia , oh no! More conflicts :(

mfdeakin-sandia:
:(

jgfouca:
@mfdeakin-sandia , if the conflicts are minor, you don't need to re-do your testing after you rebase again.

mfdeakin-sandia:
Right, now this is ready and no conflicts :)

PR #1761, opened 2017-07-26
requested mods to ensemble.sh 

allibco:
Fixed these two issues (from Chris F.):
1.  There is a known issue that needs to be fixed in ensemble.sh.
         Line 244 in ensemble.sh needs to be changed to.
         CLONECOUNT=$((${Args[$i]}-1))
2.  The next issue is single_run.sh requires a project/account.  This is causing problems
      on systems that don't require a project/account.  It should be simple enough to not append
      --project $ACCOUNT to the NewCaseFlags string at line 312, when $ACCOUNT isn't being set.

Also changed default case from FC5 to F2000_DEV (since FC5 not an option anymore)

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/tools/statistical_ensemble_test/ensemble.sh
+++ b/tools/statistical_ensemble_test/ensemble.sh
@@ -241,7 +241,8 @@ while [ $i -le ${#Args[@]} ]; do
     --ensemble )
       i=$((i+1))
       # Set CLONECOUNT and runtype
-      CLONECOUNT=${Args[$i]}-1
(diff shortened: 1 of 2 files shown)


fischer-ncar:
We'll have to change the default case down the road.  When the atmosphere group has a final compset name.

Issue #140, opened 2016-04-21
Rework Fortran error handling to support unit tests


billsacks:
This isn't something needed immediately, but it would be very nice to have at some point, to make the pFUnit-based unit tests more useful. I'm opening this issue as a place to record some thoughts:

It would be good to rework the Fortran error handling code (currently, basically shr_sys_abort) so that you could plug in a different error handler when doing unit tests. For example, you could plug in an error handler that throws a pFUnit exception.

This would serve at least two purposes:

(1) Currently, there is no good way to test error-handling code: If you try to have a unit test that triggers an abort, then your whole test suite dies. I _believe_ (but I'm not 100% sure) you could handle this if you swapped out the production error handler for an error handler that raised a pFUnit "exception", then checked to make sure that the expected exception was raised.

(2) If there's a bug in either your production code or your test code that unintentionally causes shr_sys_abort to be triggered, then that whole suite of pFUnit tests aborts, and you aren't given very a very useful error message (e.g., it doesn't even point to the test that failed). If you could plug in a different error handler, which raised a pFUnit "exception" rather than aborting, then you would see useful information about which test(s) failed, and other tests would continue.

@quantheory went a long ways down the path of doing this two years ago. I can't remember why this never quite made it to the trunk. @jedwards4b also sent out an open source Fortran error handler recently, which I haven't looked at closely.

jedwards4b:
https://github.com/koenpoppe/ErrorHandlingInFortran2003

billsacks:
I realized that much of this benefit could be achieved by using Sean's stub version of shr_sys_mod: shr_sys_mod.nompi_abortthrows.F90.

billsacks:
I have some ideas for how to make this work that I'll work on today. It amounts to doing something similar to Sean's shr_sys_mod.nompi_abortthrows.F90, but without requiring fully replacing shr_sys_mod.

I can see 2 solutions:

(1) Use a preprocessor macro, like `TESTING_WITH_PFUNIT`, which will be defined in unit test builds

Then shr_sys_abort will have:

```Fortran
#ifdef TESTING_WITH_PFUNIT
   use pfunit
#endif
```

and, at the start of shr_sys_abort:

```
#ifdef TESTING_WITH_PFUNIT
    throw()
#endif
```

Don't have an #else. That way it's clear that we're not avoiding the abort stuff ever - we're just doing an additional thing first. (Point is: I don't want to accidentally miss doing the true abort in production code.)


[...]

PR #1765, opened 2017-07-26
Check flds for nan

### LINKED ISSUE #1757: Introduce NaN checks for fields passed between components
We want to introduce checks in the coupler that ensure that all fields passed from components are free of NaN values.

There is also value in having some other physical checks (like certain fields that should be non-negative), but feeling from a number of discussions is that those physical checks belong in components, not in the coupler.

jedwards4b:
Checks incoming component fields for NaN in the coupler, fail on first NaN found, report
component name and instance, field name and 1d global index into field.  Add a test in scripts_regression_tests.py that checks this feature is working.  

Test suite: scripts_regression_tests.py , PFS.f19_g17.B1850.cheyenne_intel with and without to measure impact on performance (negligible) 
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1757 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


rljacob:
There's a lot of trailing whitespace differences.  Didn't all the whitespace removal happen a long time ago?

src/drivers/mct/main/component_mod.F90
```diff
@@ -892,4 +898,41 @@ subroutine component_diag(infodata, comp, flow, comment, info_debug, timer_diag
...
+    integer(IN), pointer :: gpts(:)
+    character(len=CL) :: msg
+    real(r8), pointer :: rattr(:,:)
+
+    lsize = mct_avect_lsize(comp%c2x_cc)
+    nflds = size(comp%c2x_cc%rattr,1)
+    if(lsize <= 1 .and. nflds <= 1) return
+    rattr => comp%c2x_cc%rattr
+    do fld=1,nflds
+       nancnt = 0
```
billsacks: This variable `nancnt` is unused and should be removed

src/drivers/mct/main/component_mod.F90
```diff
@@ -892,4 +898,41 @@ subroutine component_diag(infodata, comp, flow, comment, info_debug, timer_diag
...
+    integer(in)   :: fld, n            ! iterators
+    integer(IN)  :: rank
+    integer(IN) :: ierr
+    integer(IN), pointer :: gpts(:)
+    character(len=CL) :: msg
+    real(r8), pointer :: rattr(:,:)
+
+    lsize = mct_avect_lsize(comp%c2x_cc)
+    nflds = size(comp%c2x_cc%rattr,1)
+    if(lsize <= 1 .and. nflds <= 1) return
```
billsacks: Why do you return in this case?

mvertens:
@jedwards4b - first thanks for doing this. Also - have we confirmed that the results are bfb in a B1850 test? That would be important to do.

billsacks:
In case you're just following along via email notifications (which don't get updated with edits), I have added a bit to my original review, including an important point (8).

billsacks:
> I'd like this check to be controlled by a flag. I'm concerned it will hurt performance at high resolution.

@rljacob - Would your concerns about performance be alleviated somewhat if the current doubly-nested loop were replaced with an initial check like this, so that shr_infnan_isnan only needs to be called once (noting that it's an elemental function):

```Fortran
if (any(shr_infnan_isnan(rattr))) then
   ! do the existing loop to find the first NaN
end if
```

or are you concerned about the performance of checking individual elements for NaN no matter how it's implemented?

src/drivers/mct/main/component_mod.F90
jedwards4b: The c2x_cc structure is allocated even if it is not used (stub components) in this case size is  1,1 and the field should not be tested.

src/drivers/mct/main/component_mod.F90
jedwards4b: Thanks, I've removed this.

jedwards4b:
For reasons that are not clear to me my final version of scripts_regression_tests.py got overwritten by an older version, rather than fix it again I've pulled those changes and will work on a unit test as @billsacks suggests.   I also agree with @rljacob that this check should be namelist controlled and will add a flag to do so.

src/drivers/mct/main/component_mod.F90
billsacks: Okay. Can you add a comment documenting that? Even better would be if there were a logical in component_type like `c2x_cc_used`, which is set to true in general, but false for stub components, and then you can test against this logical.

billsacks:
@jedwards4b - I have some thoughts this morning about how we could more easily support failure testing like this in unit tests. I'll try implementing this later today.

By the way, I'm fine accepting this PR without tests for now, with the plan to add unit tests later, if you prefer. (I'll be out of the office for about two weeks after today.)

jedwards4b:
@mvertens I have checked bfb against B1850 in alpha07b

mvertens:
@jedwards4b - thanks! I agree with @billsacks that a unit test can be
deferred and this PR should go ahead in order to get it into current CESM
simulations.

jedwards4b:
@rjacob I have added a namelist field nan_check_component_fields it is a logical which defaults to .false. for acme and .true. for cesm.   
@billsacks Unit tests are in place but cannot be turned on until issue #140 is resolved. 

I think that this PR is ready to merge - can you update your reviews if you agree.

src/drivers/mct/main/component_mod.F90
```diff
@@ -32,7 +32,7 @@ module component_mod
   use mct_mod   ! mct_ wrappers for mct lib
```
mvertens: There still seems to be a lot of white space diffs. Can we remove them?

src/drivers/mct/shr/seq_infodata_mod.F90
```diff
@@ -276,11 +276,11 @@ MODULE seq_infodata_mod
 !===============================================================================
```
mvertens: Same comment about white space diffs.

src/drivers/mct/main/component_mod.F90
jedwards4b: I have removed all whitespace in #1766 accepting that will remove the differences in this one.

billsacks:
Looks good now - thanks!

rljacob:
Bill asked: "are you concerned about the performance of checking individual elements for NaN no matter how it's implemented?"  Yes.

jedwards4b:
PR #1771 resolves the issue with unit tests in this PR so I have rebased and turned on the unit tests.

billsacks:
Thanks, @jedwards4b !

PR #1418, opened 2017-04-26
Adds support for data model to read multiple time slices at once

bishtgautam:
A new namelist variable ("readmode") is added that determines how
a temporal stream dataset is read. At present, "readmode" can be set to:
  - single    : Default method of reading a single time slice at a time
  - full_file : Reads all time slices in a given netcdf file at once

Test suite: 
Test baseline: 
Test namelist changes: 
Test status: [bit for bit, roundoff, climate changing]

Fixes [CIME Github issue #]

User interface changes?: new xml variable, readmode

Code review:

PR #1767, opened 2017-07-27
refactored datamodels

### LINKED ISSUE #1713: drivers and share circular dependency with shr_dmodel
In working on the build system, I've noticed that the drivers/mct/util/seq_*.F90 files depend on several share/utils/shr_*.F90 shared functions, as is the purpose of share/util. However, there is also one dependency that goes the backward direction, where share/util/shr_dmodel_mod.F90 has a dependency back into drivers/mct/shr/seq_flds_mod.F90:
`use seq_flds_mod, only : seq_flds_dom_coord, seq_flds_dom_other`
where the seq_ variables are used just once in this line:
```
  call mct_gGrid_init( GGrid=gGrid, CoordChars=trim(seq_flds_dom_coord), &
     OtherChars=trim(seq_flds_dom_other), lsize=lsize )
```

This is a minor annoyance to the build process, in that the share/util directory has to be broken up into multiple libraries to get the dependencies with drivers/mct/shr correct. I realized that this means that the share directory will not compile when another driver (nuopc) becomes an option, since a new driver presumably won't come with seq_flds_mod.F90. So thought it was worth calling attention to it.

Does anyone understand this code enough to untangle the circular dependency between these two directories, and have motivation to do so? (It isn't as simple as moving shr_dmodel into drivers/mct, since it is in turn used in other share/util files.) Thanks.

mvertens:
refactored datamodels for new modularity

This PR contains a refactorization of the data models in order to separate out the driver specfic parts of dxxx_comp_mod.F90 (which now contains datatypes such as cdata, infodata, etc. that are specific to the current mct cpl7 driver) and permit new driver implementations (e.g. nuopc) to use the core parts of the data models. In addition, a new directory data_comps/dshare now contains data model share code that was previously in share/utils.

The following tests had baselines generated without the refactor and tests were
run to verify that the data model refactor maintained bfb answers
```
- ERI.f09_g17.ETEST.cheyenne_intel.cice-default
- ERI.f09_g17.ETEST.cheyenne_intel.cice-default
- ERI.f09_g17.ETEST.cheyenne_intel.cice-default
- ERI.T62_g16.C1850ECO.cheyenne_intel.pop-ecosys
- ERI.T62_g16.C.cheyenne_intel.pop-default
- ERI.T62_g16.CIAF.cheyenne_intel.pop-default
- ERI.T62_g16.GIAF.cheyenne_intel.pop-default
- ERI.T62_g17.DTEST.cheyenne_intel.cice-default
- ERI.T62_g17.G.cheyenne_gnu.pop-cice
- ERP_D_Ln9.f19_f19_mg17.QPC6.cheyenne_intel.cam-outfrq9s
- ERP_D_Ln9.f19_f19_mg17.QSC6.cheyenne_intel.cam-outfrq9s
- ERP_Ln9.f09_f09_mg17.F1850_DONOTUSE.cheyenne_intel.cam-outfrq9s
- ERP_Ln9.f09_f09_mg17.FHIST_DEV.cheyenne_intel.cam-outfrq9s
- ERP_Ln9.f19_f19_mg17.F2000_DEV.cheyenne_gnu.cam-outfrq9s
- ERR.f45_g37_rx1.A.cheyenne_intel
- ERS.f09_g16.ADLND.cheyenne_intel
- ERS_Ld3.f45_g37_rx1.A.cheyenne_intel
- ERS_Lm3.T62_g16.AIAF.cheyenne_intel
- ERS_Ly20_N2_P2.f09_g16_gl20.T1850G1.cheyenne_intel
- ERS_Ly3.f09_g16_gl4.T1850G.cheyenne_intel
- PEM.f19_g16_rx1.A.cheyenne_intel
- PEM.f19_g16_rx1.A.cheyenne_intel
- SMS.f09_g16.ADWAV.cheyenne_intel
```
`Also ran most aux_clm5 tests and verified bfb with clm4_5_1_r251`
Test baseline: see above
Test namelist changes: 
Test status: none

Fixes #1713 

User interface changes?: None

Code review:


mvertens:
@billsacks thinks that I have moved too many routines into dshare. He would like for everything but shr_dmodel_mod.F90 to be moved back to share/utils since other components (e.g. clm and pop) use this code in non data model model. My rationale for moving these files is that they form a coherent unit of data model share code and they are bundled together because of that. However, they all still use routines from share/utils. Alternatively, @billsacks also suggested that dshare be moved under share instead of being in components/datacomps. That would enable the routines to be bundled together but be located in an area that was not totally specific to the data model code. Frankly, I am fine either way - but a wider consensus would be helpful.

mvertens:
I just realized that I'd like to move a few more things into the mct layer before the review continues.

jedwards4b:
@rljacob waiting for your review...

rljacob:
How does this relate to PR #1418 ?

jedwards4b:
It's independent of #1418 which is already on master.

jedwards4b:
I don't see any reason clm and the other component models can't use shr_dmodel_mod.F90 in it's new location, it would just mean adding a directory to the models FIlepath in buldlib.

mvertens:
No change is neeeded for calm or any of the prognostic components. The just
use the share library . See me testing summary.

jedwards4b:
Then I guess I don't understand @billsacks argument, this would seem to clarify and separate what is needed for the data models and what is shared code usage.

rljacob:
What functions from the 4 files in the new dshare directory are used by non-data models?   Its its small, maybe just move those to a new file back in share/utils.  I like the current dshare setup.

What is the other reorg trying to do?  Taking dlnd, there's now an mct subdir with one file in it and a new file called dlnd_shr_mod with one function in it.   If thats the only part shareable with nuopc, probably no point in trying to make the data models shareable.

rljacob:
On further thought, I'll have to look more closely at which mct datatype the data models are using.  If they don't require MCTWorld to be initialized, then they should be useable with a non-MCT coupler/driver.

mvertens:
There is no significant non-mct part to the data models. NOUPC for now will
be using mct-based data models. What I have done is remove the driver
specific parts of the data models (i.e. cdata, infodata, etc) that are not
used by nuopc. Now mct has a cap and nuopc will have a cap - but nuopc will
still have to translate its state to attribute vectors and gsmap and ggrid
in order to use the data models since those are intrinsic to the behavior
of the data models.

rljacob:
Oh I thought this was trying separate out the mct parts, not just the mct driver.   Its not a lot of code for its own subdir but the build system requires it.

Instead of an mct subdir in each data model directory, how about a "mctdrivers" directory at the same level as datm, dlnd, etc.  Put all the mct drivers there.   Then you'll just need one "nuopcdrivers" directory to match it.

mvertens:
I think that would be awkward. The datm/mct/atm_comp_mct. F90 and
dlnd/mct/dlnd_comp_mct.F90 are not the same and need to be different. This
structure matches what is done for the active components.

mvertens:
Given that this is accepted by both reviewers - please go ahead and merge it.
@apcraig - this is the basis of the NUOPC data model refactor if you want to take a look.

jedwards4b:
@mvertens there were a lot of merge conflicts between this PR and my PR#1778, I have rebased this branch in my sandbox and resolved the conflicts, I am testing now.

mvertens:
Thank you for doing this. I really appreciate it.

billsacks:
@jedwards4b @rljacob @mvertens  Sorry for not being able to comment on this sooner: I just got back from vacation. I wanted to reply to a seeming misunderstanding of my thoughts about the separation of the stream functionality into a new directory (`src/components/data_comps/dshare`): This has nothing to do with the model build, as @jedwards4b seems to have assumed: @mvertens set this up so that `src/components/data_comps/dshare` is built into the share library. Rather, my concern was with the understandability of the code with this new organization. Because this PR has been merged, I have opened a new "discussion" issue to clarify my thoughts: #1802

PR #1771, opened 2017-07-28
Facilitate expected error unit tests

billsacks:
Pull shr_sys_abort into its own module - now call shr_abort_abort in
shr_abort_mod. (The routine name, while redundant, follows the share
code convention of MODULENAME_VERB.) Other code can still use this via
shr_sys_mod: shr_sys_abort, via a rename done in that module. Two other
routines from shr_sys_abort that are used by shr_abort_abort are also
moved into shr_abort_abort to avoid circular dependencies.

There are no substantive changes in shr_abort_mod relative to the code
that used to be in shr_sys_abort: I have just moved code into this new
module and done some reformatting.

The purpose of this separation was to facilitate using a replacement
abort routine in pFUnit-based unit tests. This PR also adds a module
thath provides that replacement, which throws a pFUnit exception rather
than actually aborting. This also changes the cime unit test build so
that this new pFUnit-based abort routine is included in all drv unit
tests.

This also adds a simple unit test of the pFUnit-based abort
routine. This demonstrates how you can use this to write your own
"expected failure" unit tests:

```Fortran
  @Test
  subroutine test_abort(this)
    class(TestShrAbort), intent(inout) :: this

    call shr_abort_abort('Test message')
    @assertExceptionRaised('ABORTED: Test message')
  end subroutine test_abort
```

Test suite:
- scripts_regression_tests on yellowstone and cheyenne
- cime fortran unit tests with gnu on my Mac
- SMS.f09_g16.X and SMS_D.f09_g16.X with:
  - hobart_nag
  - hobart_pgi
  - yellowstone_intel
  - cheyenne_gnu
- SMS.f09_g16.X and SMS_D.f09_g16.X with cheyenne_intel, with a
  shr_sys_abort call inserted: made sure runs still die properly when
  they call shr_sys_abort
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#140

User interface changes?: none

Code review: 

@ekluzek @bandre-ucar - this could be useful for adding expected error
unit tests in CLM

### DIFF
--- a/src/share/test/unit/mock/shr_sys_mod.nompi_abortthrows.F90
+++ b/src/share/test/unit/mock/shr_sys_mod.nompi_abortthrows.F90
@@ -7,7 +7,6 @@ module shr_sys_mod
 use shr_kind_mod, only: &
      shr_kind_in, shr_kind_r8
 
-! This used to be in shr_sys_mod; we provide this rename for backwards compatibility
 use shr_abort_mod, only : shr_sys_abort => shr_abort_abort
 
 implicit none
@@ -15,7 +14,10 @@ private
 save
 
 ! Fake abort
-! Imported from shr_abort_mod and republished with renames for backwards compatibility
+! Imported from shr_abort_mod and republished with renames. Other code that wishes to use
+! these routines should use these shr_sys names rather than directly using the routines
+! from shr_abort_abort. (This is for consistency with older code, from when these routines
+! were defined in shr_sys_mod.)
 public :: shr_sys_abort
 
 ! Fake sleep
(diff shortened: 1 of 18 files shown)


src/share/test/unit/mock/shr_sys_mod.nompi_abortthrows.F90
```diff
@@ -7,11 +7,15 @@ module shr_sys_mod
 use shr_kind_mod, only: &
      shr_kind_in, shr_kind_r8
 
+! This used to be in shr_sys_mod; we provide this rename for backwards compatibility
```
jedwards4b: I think you should make it clear that client code should continue to use shr_sys_abort and not call shr_abort_abort directly.

src/share/util/shr_abort_mod.F90
```diff
@@ -0,0 +1,155 @@
+module shr_abort_mod
```
jedwards4b: Please add a note here that components should not use this module directly and should continue to use shr_sys_mod: shr_sys_abort. I think that it's important to maintain a single method in the API to access this function.

billsacks:
Thanks, @jedwards4b - good point. I have added comments to address your point.

PR #1776, opened 2017-07-29
dont use Machines object to get DIN_LOC_ROOT

### LINKED ISSUE #1773: NODENAME_REGEX is required - should it be?
Currently some tools such as check_input_data use the Machines object but do not have 
--machine as a command line argument This effectively means that the NODENAME_REGEX is a required field in config_machines.xml. 

An alternative would be to add the --machine argument to the default arguments in standard_script_setup.py so that it is available for all command line scripts.

jedwards4b:
Using the Machines object in check_input_data meant that either we needed to add --machine as a command line argument or we needed the optional NODENAME_REGEX defined for every machine.  Looking at the code I realized that there was no real requirement to have a Machines object and so I removed it and replaced with a call to xmlquery.  Also added a fix for a pylint error in case.qstatus and a unit test problem in check_fields.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1773 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/Tools/check_input_data
+++ b/scripts/Tools/check_input_data
@@ -7,7 +7,7 @@ Should be run from case.
 """
 from standard_script_setup import *
 
-from CIME.utils import get_model, run_cmd_no_fail
+from CIME.utils import get_model
 from CIME.check_input_data import check_input_data, SVN_LOCS
 from CIME.case import Case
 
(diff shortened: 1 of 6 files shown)


scripts/Tools/check_input_data
```diff
@@ -41,10 +38,11 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
     parser.add_argument("--svn-loc", default=SVN_LOCS[get_model()],
                         help="The input data repository from which to download data.")
 
-    parser.add_argument("-i", "--input-data-root", default=None,
-                        help="The root directory where input data goes. "
-                        "Default value will come from DIN_LOC_ROOT from case which itself "
-                        "defaults to '{}'".format(MACHINE.get_value("DIN_LOC_ROOT")))
+    din_loc_root = run_cmd_no_fail("./xmlquery --value DIN_LOC_ROOT")
```
jgfouca: I like the overall change, but not this approach. There's got to be a better way of getting DIN_LOC_ROOT than resorting to the shell which is very slow and error prone. Maybe default to None and then later resolve the variable using the case object?

jedwards4b: The only reason to have this there is to show the user the default value in the help message.
How about if instead the help says "Use xmlquery DIN_LOC_ROOT in your case directory to view default."

jedwards4b:
I've removed the xmlquery command and changed the help message.

jedwards4b:
reran scripts_regression_tests.py

PR #1778, opened 2017-07-31
Update dmodel ioformat

jedwards4b:
Update IO_format for data models, done in a backward compatible way so that externals can continue to use older method (but with io_format hard coded to 64bit_offset)

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/machines/config_batch.xml
+++ b/config/cesm/machines/config_batch.xml
@@ -397,7 +397,6 @@
     <batch_submit>sbatch</batch_submit>
     <directives>
       <directive>-C knl,quad,cache </directive>
-      <directive>-S 2 </directive>
     </directives>
     <queues>
       <queue walltimemax="02:00:00" jobmin="1" jobmax="45440">regular</queue>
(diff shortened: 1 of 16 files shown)


jedwards4b:
Reran scripts_regression_tests.py - all pass.

config/cesm/machines/config_batch.xml
```diff
@@ -397,6 +397,7 @@
     <batch_submit>sbatch</batch_submit>
     <directives>
       <directive>-C knl,quad,cache </directive>
+      <directive>-S 2 </directive>
```
jayeshkrishna: This change should be in a different PR

config/xml_schemas/env_mach_specific.xsd
```diff
@@ -4,7 +4,7 @@
 <xs:attribute name="id" type="xs:NCName"/>
 <xs:attribute name="version" type="xs:decimal"/>
 <xs:attribute name="lang" type="xs:NCName"/>
-<xs:attribute name="compiler" type="xs:NCName"/>
+<xs:attribute name="compiler" type="xs:string"/>
```
jayeshkrishna: This change should also be in a separate PR.

PR #1790, opened 2017-08-03
Add more default logging to desp

### LINKED ISSUE #1788: Data ESP model not verbose enough
Some warnings and messages should always be output.
As a start, move all current messages to default output.

gold2718:
Set default level for info_debug to one to create more logging 
Adjust some messages to avoid unnecessary warnings.

Test suite: scripts_regression_tests.py on Hobart
Test baseline: NA
Test namelist changes: NA
Test status: bit for bit

Fixes #1788

User interface changes?: No

Code review:

### DIFF
--- a/src/components/data_comps/desp/cime_config/namelist_definition_desp.xml
+++ b/src/components/data_comps/desp/cime_config/namelist_definition_desp.xml
@@ -49,18 +49,6 @@
     </values>
   </entry>
 
-  <entry id="info_debug">
-    <type>integer</type>
-    <category>desp</category>
-    <group>desp_nml</group>
-    <desc>
-      Level of debug output, 0=minimum, 1=normal, 2=more, 3=too much (default: 0)
-    </desc>
-    <values>
-      <value>1</value>
-    </values>
-  </entry>
-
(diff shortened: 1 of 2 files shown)


src/components/data_comps/desp/cime_config/namelist_definition_desp.xml
```diff
@@ -57,7 +57,7 @@
       Level of debug output, 0=minimum, 1=normal, 2=more, 3=too much (default: 0)
     </desc>
     <values>
-      <value>0</value>
+      <value>1</value>
```
jedwards4b: This isn't the same info_debug used by the coupler? I don't think that it should have the same name.  This is a cause for confusion as I was changing info_dbug in the xml and thought that I was toggling this value.   Why do we need this flag at all?  It seems that all of these messages written from mastertask to the desp log are appropriate.

gold2718: Okay, I just took out info_debug so you should see all the messages all the time (well, one warning only shows up if you are using DATATEST mode, not in NOCHANGE or NULL mode).

PR #1804, opened 2017-08-10
allow --compiler and --mpilib

### LINKED ISSUE #1770: cime nightly tests should test all supported compilers on the test machines
My understanding is that the cime nightly tests - and scripts_regression_tests in general - just tests the default compiler on the given machine. This fails to catch problems like #1769 . We should run the cime nightly tests across all supported compilers on the test machines, and possibly extend scripts_regression_tests so that it always runs at least one test from each supported compiler on the current machine.

I personally don't feel we need to run all system tests on all compilers. But ideally we'd have at least one debug test and one non-debug test on each compiler on each tested machine.

cc @jedwards4b @jgfouca @mvertens @rljacob @fischer-ncar

jedwards4b:
This PR allows ctest to use the new --compiler and --mpilib options in scripts_regression_tests.py
by using env variables CIME_COMPILER and CIME_MPILIB

Test suite: CIME_COMPILER=nag ./cime.ctest.hobart
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1770 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/tests/CMakeLists.txt
+++ b/scripts/tests/CMakeLists.txt
@@ -1,3 +1,8 @@
+#
+# Environment variables CIME_COMPILER and CIME_MPILIB
+# can be used to send --compiler and --mpilib settings to scripts_regression_tests.py
+#
+
 cmake_minimum_required(VERSION 2.8)
 
 include(CTest)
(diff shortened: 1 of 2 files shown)


billsacks:
> I gave them CIME_ prefix because that way it's unlikely to conflict with anything else. I don't think that this is worth the effort it would take to change, but you are welcome if you want to.

Okay, I can live with this, but can you please add documentation at the top of the two cmake files documenting the usage and the possible environment variables that can affect operation of these scripts? Once you have done that, I'll accept this PR.

jedwards4b:
added comments

PR #1816, opened 2017-08-15
Refactor how spare nodes get computed.

jgfouca:
The prior implementation was a bit too simple. Always giving the user
10 percent of nodes as spare nodes was way overkill for large jobs.
    
The new implementation maxes out the number of spare nodes at 10.
    
We also add a new variable to allow the user to pick the exact number
of spare nodes that they want.

Test suite: scripts_regression_tests T_TestRunRestart
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: Yes, new case variables controlling spare nodes

Code review: @jedwards4b


jgfouca:
Going to go ahead and merge. @vanroekel can ping me anytime to tweak the implementation.

PR #1823, opened 2017-08-17
Upgrade build system.

### LINKED ISSUE #1741: Options to case.build to speed incremental rebuilds
I feel like this was discussed earlier in one or more issues, but I can't find them, so I'm opening a new issue.

When I'm building new code that I've written, I usually have a lot of compilation errors initially (I know, it's probably just me...), and so I need to incrementally rebuild (rerunning ./case.build) a few times before the build is successful.

This currently takes quite a bit longer than it needs to, because before ./case.build gets to the library in question, it:

(a) Builds component namelists – sometimes twice, at least for some test types

(b) Checks whether all preceding libraries need to be rebuilt

These incremental rebuilds could be sped up if we had options to case.build to:

(1) skip namelist generation (addressing (a))

(2) build only a single, given library (addressing (b))... maybe this would imply also skipping namelist generation

What do people think about adding these or other options to speed up incremental rebuilds?

jgfouca:
1) Make incompatible flags to case.build mutually exclusive
2) Add new flag for building individual components
3) Never run buildnml when building individual libs/components

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1741 

User interface changes?: Yes, new -b flag to case.build.

Code review: jedwards, sacks

### DIFF
--- a/scripts/Tools/case.build
+++ b/scripts/Tools/case.build
@@ -56,7 +56,7 @@ OR
                              "if no arguments then clean all objects other than sharedlib objects")
 
     mutex_group.add_argument("-b", "--build", nargs="+", choices=allobjs,
-                             help="libs to build")
+                             help="libs to build. Will cause namelist generation to be skipped.")
 
     mutex_group.add_argument("--clean-all", action="store_true",
                              help="clean all objects including sharedlibobjects that may be used by other builds")
--- a/scripts/Tools/case.build
+++ b/scripts/Tools/case.build
@@ -36,11 +36,13 @@ OR
     parser.add_argument("caseroot", nargs="?", default=os.getcwd(),
                         help="Case directory to build")
 
-    parser.add_argument("--sharedlib-only", action="store_true",
-                        help="Only build sharedlibs")
+    mutex_group = parser.add_mutually_exclusive_group()
 
-    parser.add_argument("-m", "--model-only", action="store_true",
-                        help="Assume shared libs already built")
+    mutex_group.add_argument("--sharedlib-only", action="store_true",
+                             help="Only build sharedlibs")
+
+    mutex_group.add_argument("-m", "--model-only", action="store_true",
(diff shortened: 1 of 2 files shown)


scripts/Tools/case.build
```diff
@@ -49,18 +51,22 @@ OR
...
 
-    parser.add_argument("--clean", nargs="*", choices=allobjs,
-                        help="objects to clean"
-                        "if no arguments then clean all objects other than sharedlib objects")
+    mutex_group.add_argument("--clean", nargs="*", choices=allobjs,
+                             help="objects to clean"
+                             "if no arguments then clean all objects other than sharedlib objects")
+
+    mutex_group.add_argument("-b", "--build", nargs="+", choices=allobjs,
+                             help="libs to build")
```
billsacks: Please add documentation stating that the creation of namelists will be skipped if this argument is specified.

scripts/lib/CIME/build.py
```diff
@@ -248,9 +254,8 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
                 if re.search("Current setting for", line):
                     logger.warn(line)
 
-
     # clm not a shared lib for ACME
-    if get_model() != "acme":
+    if get_model() != "acme" and not (buildlist is not None and "lnd" not in buildlist):
```
billsacks: I'm finding the double (triple if you count "None") negatives here hard to parse. I'd find it easier to understand if this was refactored to the following (you should double-check this logic, though):

```python
if get_model() == "acme" or (buildlist is not None and "lnd" not in buildlist):
    pass
else
```

jgfouca:
@billsacks Done.

billsacks:
Thanks. @jgfouca I like your refactored condition better than my suggestion - thanks!

jgfouca:
You're welcome. @jedwards4b this is ready to integrate.

PR #1827, opened 2017-08-18
new capability to apply user mods in a case - needed for SCAM workflow

mvertens:
New case.setup to apply usermods after case creation.

This PR adds a new argument to case.setup that permits usermods to be applied after case creation. The caveat is that when these usermods are applied - all user_nl_xxx files are first scrubbed and empty ones are created. 

SCAM would like the capability to have one executable but easily change the IOP forcing data that is needed. Currently, they have a compset for each new IOP dataset. This is very cumbersome and is a step backwards from their previous workflow. With this change - a single SCAM compset can be user - with a time prefix SCAM. The a combination of 
./case.setup --apply-usermods and create_clone --keepexe can be used to satisfy the requirements of a SCAM workflow that would be suitable for scientific experimentation.

Test suite: None - hand tested the capability
Test baseline: None
Test namelist changes: None

Fixes: None 
User interface changes?: None
Code review: jedwards

### DIFF
--- a/scripts/Tools/case.setup
+++ b/scripts/Tools/case.setup
@@ -7,7 +7,11 @@ case.setup - create the $caseroot/case.run script and user_nl_xxx component name
 from standard_script_setup import *
 
 from CIME.case_setup import case_setup
-from CIME.case       import Case
+from CIME.case import Case
+from CIME.user_mod_support import apply_user_mods
+from CIME.utils import expect
+
+logger = logging.getLogger(__name__)
 
 ###############################################################################
 def parse_command_line(args, description):
@@ -45,9 +49,16 @@ OR
     parser.add_argument("-r", "--reset", action="store_true",
                         help="Does a clean followed by setup")
 
+    parser.add_argument("-a", "--apply-usermods",
+                        help="Specify a directory to apply user mods")
+
     args = CIME.utils.parse_args_and_handle_standard_logging_options(args, parser)
 
-    return args.caseroot, args.clean, args.test_mode, args.reset
+    if args.apply_usermods is not None:
+        expect(not args.reset and not args.clean,
+               "Error: cannot set --reset or --clean if are using --apply-usermods")
+
+    return args.caseroot, args.clean, args.test_mode, args.reset, args.apply_usermods
 
 ###############################################################################
 def _main_func(description):
@@ -56,9 +67,14 @@ def _main_func(description):
         test_results = doctest.testmod(verbose=True)
         sys.exit(1 if test_results.failed > 0 else 0)
 
-    caseroot, clean, test_mode, reset = parse_command_line(sys.argv, description)
-    with Case(caseroot, read_only=False) as case:
-        case_setup(case, clean=clean, test_mode=test_mode, reset=reset)
+    caseroot, clean, test_mode, reset, usermods_dir = parse_command_line(sys.argv, description)
+
+    if usermods_dir is not None:
+        logger.warn("WARNING: REMOVING all user_nl_xxx files as part of ./case_setup --apply-usermods")
+        apply_user_mods(caseroot, usermods_dir, cleanfirst=True)
+    else:
+        with Case(caseroot, read_only=False) as case:
+            case_setup(case, clean=clean, test_mode=test_mode, reset=reset)
 
 if __name__ == "__main__":
     _main_func(__doc__)
--- a/scripts/lib/CIME/user_mod_support.py
+++ b/scripts/lib/CIME/user_mod_support.py
@@ -8,7 +8,7 @@ import shutil, glob
 
 logger = logging.getLogger(__name__)
 
-def apply_user_mods(caseroot, user_mods_path):
+def apply_user_mods(caseroot, user_mods_path, cleanfirst=None):
     '''
     Recursivlely apply user_mods to caseroot - this includes updating user_nl_xxx,
     updating SourceMods and creating case shell_commands and xmlchange_cmds files
@@ -25,6 +25,15 @@ def apply_user_mods(caseroot, user_mods_path):
             os.remove(shell_command_file)
 
     include_dirs = build_include_dirs_list(user_mods_path)
+
+    # Clean all user_nl_xxx files
+    if cleanfirst is not None:
+        for filename in glob.iglob(os.path.join(caseroot,"user_nl_*")):
+            if os.path.exists(filename):
+                os.remove(os.path.join(caseroot, filename))
+                # Opening a blank user_nl file
+                open(filename, 'w').close()
+
     # If a user_mods dir 'foo' includes 'bar', the include_dirs list returned
     # from build_include_dirs has 'foo' before 'bar'. But with the below code,
     # directories that occur later in the list take precedence over the earlier


scripts/Tools/case.setup
```diff
@@ -45,9 +49,16 @@ OR
     parser.add_argument("-r", "--reset", action="store_true",
                         help="Does a clean followed by setup")
 
+    parser.add_argument("-a", "--apply-usermods",
+                        help="Specify a directory to apply user mods")
+
     args = CIME.utils.parse_args_and_handle_standard_logging_options(args, parser)
 
-    return args.caseroot, args.clean, args.test_mode, args.reset
+    if args.apply_usermods is not None:
```
jedwards4b: You might consider using the mutually exclusive options capability of parser here, as an example xmlquery uses this.

billsacks:
I just talked with @mvertens about this, and we came to the conclusion that this PR should be closed, and replaced with a different design: It seems that what's really wanted here is the ability to run create_newcase pointing to an existing executable, with some argument like `--use-exe /path/to/executable`. We felt that would be cleanest, and would avoid questions of how exactly the application of user mods operates in this case.

Mariana also asked me to share my more detailed thoughts about the current implementation, though, so I'll do that since I had already written them up - even though they are likely now a moot point:

I think there are a few principles at work in my head, which are aimed at (a) reducing the mental load required for users to understand how things work, and (b) reducing the potential for errors, which are more likely when things work in unintuitive or inconsistent ways:

(1) The Zen of Python: "There should be one-- and preferably only one --obvious way to do it". This is what made me uncomfortable with having it possible to apply user_mods either in create_newcase or in case.setup – unless there is a strong reason for needing that.

(2) If there are two similar use cases, they should be done in similar ways. Again, that made me uncomfortable with using create_newcase to apply user_mods in a new case, but case.setup to apply user_mods in a clone.

(3) If a flag or command does multiple things, then those multiple things should have symmetrical behavior. This is what makes me uncomfortable with removing user_nl files when you run create_clone (or case.setup) with a --user-mods flag: --user-mods applies user_nl changes and xmlchange commands (along with potentially other shell commands and SourceMods – but let's not worry about that for now). It is asymmetrical, and thus confusing and error-prone, if a clone overwrites anything in user_nl files from the base case, but (implicitly) adds to any xml changes from the base case.


As for a solution, I was wondering if this workflow could work for the SCAM folks:

- Make a base case without any changes, or only with changes that are meant to apply to all other cases.

- Do clones off of this, with create_clone --user-mods-dir ..., where the user mods append to anything that was in the original case.

(To be clear: that solution came from before my talk with Mariana; the idea laid out at the top of this comment is probably better.)


One more point that I'll share from our discussion is: It seems that my view of user_mods with create_clone differs from the views of @mvertens and @jedwards4b : In my view, create_clone is essentially the same as create_newcase, except you use some other case as a starting point, from which you make further modifications. In this view, it could make sense to have a `--user-mods-dir` argument to create_clone, which would mean: After you have set up the case using the original case as a starting point, apply user mods by appending to user_nl files and adding some extra shell commands. However, @mvertens and @jedwards4b seem to think this is confusing, so I'm fine not having this option - at least until others request it :-)

jedwards4b:
I'm not sure that I agree with this.   Are you saying that if the first case applies some user mods than those user mods will also be applied to all of the clones?  Why not allow for the user mods to be removed and completely replaced in a given case?  As a developer I might want to run a scam case over 9 different domains - if none of the changes required to change domain require rebuilding the case why do I need 9 different case directories?

billsacks:
@jedwards4b -

> Are you saying that if the first case applies some user mods than those user mods will also be applied to all of the clones?

Yes. I realize that you and @mvertens see this differently than I do, which is part of why @mvertens and I thought we should avoid this problem entirely and not provide a mechanism to apply user_mods in a clone case: If different experienced people have different ideas about how this should intuitively work, then maybe that means that it shouldn't work at all.

That said, I'll explain my rationale: Up until now, when you create a clone, any further changes you make in the new case are additive on top of the base case: nothing gets reset. And, up until now, if user mods are applied for multiple reasons (say, because the compset has user_mods attached and the user also specifies the `--user-mods-dir` argument), then these are also additive: nothing gets reset. But now you're saying that, if you have a clone with user mods, then the original case's user mods get reset. This leads to this confusing rule: "Clones are additive; user_mods are additive; but clones with user_mods get reset", as opposed to the more consistent rule that I'm suggesting: "Clones are additive; user_mods are additive; so clones with user_mods are also additive." Our system is complicated, so I feel that consistency like this is important for both users and developers to be able to understand the system.

> Why not allow for the user mods to be removed and completely replaced in a given case? 

In addition to the rationale above, there is one more key point: We don't have a way to completely replace user mods. user mods are not just user_nl files, but also consist of xmlchange commands (and possibly any arbitrary shell command, though we don't typically do that) and SourceMods. I see how you can remove all user_nl files (though I don't love that this removes all of the user's manual changes in addition to the changes from the user mods themselves). But we don't have a way to back out the xmlchange commands that were already applied to the base case. Again, with all this complexity, I feel it's important that we keep things as consistent and symmetrical as possible, so if you want to back out the user_nl changes from the user mods, then I'd only be comfortable if you could also come up with a way to back out whatever xmlchange commands were applied via the user_mods (not to mention other shell commands and SourceMods).

> As a developer I might want to run a scam case over 9 different domains - if none of the changes required to change domain require rebuilding the case why do I need 9 different case directories?

This feels antithetical to the very core of the cime / cesm workflow, which involves having a different case for each experimental setup. Isn't this use case what `create_clone --keep-exe` is designed for?

scripts/lib/CIME/user_mod_support.py
```diff
@@ -25,6 +25,15 @@ def apply_user_mods(caseroot, user_mods_path):
             os.remove(shell_command_file)
 
     include_dirs = build_include_dirs_list(user_mods_path)
+
+    # Clean all user_nl_xxx files
```
billsacks: This functionality doesn't belong in the `apply_user_mods` function. So if we keep this functionality at all, it should be put in its own function, called separately from case.setup.

scripts/Tools/case.setup
```diff
@@ -56,9 +67,14 @@ def _main_func(description):
...
 
-    caseroot, clean, test_mode, reset = parse_command_line(sys.argv, description)
-    with Case(caseroot, read_only=False) as case:
-        case_setup(case, clean=clean, test_mode=test_mode, reset=reset)
+    caseroot, clean, test_mode, reset, usermods_dir = parse_command_line(sys.argv, description)
+
+    if usermods_dir is not None:
+        logger.warn("WARNING: REMOVING all user_nl_xxx files as part of ./case_setup --apply-usermods")
+        apply_user_mods(caseroot, usermods_dir, cleanfirst=True)
+    else:
```
billsacks: Do you really mean for this to be in an `else` block? This means that if you run `case.setup --apply-user-mods` then the core of case_setup isn't done. That seems weird.

scripts/Tools/case.setup
```diff
@@ -56,9 +67,14 @@ def _main_func(description):
...
         sys.exit(1 if test_results.failed > 0 else 0)
 
-    caseroot, clean, test_mode, reset = parse_command_line(sys.argv, description)
-    with Case(caseroot, read_only=False) as case:
-        case_setup(case, clean=clean, test_mode=test_mode, reset=reset)
+    caseroot, clean, test_mode, reset, usermods_dir = parse_command_line(sys.argv, description)
+
+    if usermods_dir is not None:
+        logger.warn("WARNING: REMOVING all user_nl_xxx files as part of ./case_setup --apply-usermods")
+        apply_user_mods(caseroot, usermods_dir, cleanfirst=True)
```
billsacks: I thought we avoid having any substantive logic or behavior in the top-level scripts, instead putting it in the associated function (case_setup in this case).

rljacob:
> Currently, they have a compset for each new IOP dataset.

Why is that?

CIME tools should not do something different if the compset happens to have "SCAM" in it.

mvertens:
They need a different case directory for each IOP forcing. This is a run
time setting - so you want to not have the same executable be rebuilt each
time. And each new IOP forcing is accompanied by a different usermods
directory that has both a user_nl_clm, user_nl_cam and a series of
xmlchange commands that set starting dates, lat, lons, etc.

mvertens:
We will implement a different approach which is to have a user_mods input to create_clone.

rljacob:
Can you open an issue for what this PR was supposed to solve?

Issue #1813, opened 2017-08-13
proposed test of short term archiving 


jedwards4b:
A test should be created that does an initial run with the save interum restart option on and writes restart at some time which is not an even divisor of the total.   It would then clean the run directory and replace it with the files in the archive directory and restart from those files and run to the original runtime comparing history files from the initial run to the restart run.    Perhaps the ERS test could be refactored to do this instead of introducing an entirely new test?

billsacks:
@jedwards4b was this resolved by #1837 ?

jedwards4b:
Yes - closing.

PR #1830, opened 2017-08-20
Rewrite ERP and ERS to use SystemTestsCompareTwo

### LINKED ISSUE #1647: Would this be a way we could convert ERS and ERP to SystemTestsCompareTwo?
We previously felt that it would be challenging to rewrite ERS and ERP using the SystemTestsCompareTwo infrastructure because we couldn't see a way to have the clone case do a restart of the original case (see also some discussion in #448 and #824 ). But would this be a way to do it?: Name the clone case the same as the original case, just in a different location, and then link to the necessary rpointer and restart files before running case2.

The reason for doing this is that we keep running into problems with the ERP test (e.g., see #1640). I believe this test would be more robust and maintainable if it were converted to the SystemTestsCompareTwo infrastructure, since this would streamline the test and remove the need for resetting things in between runs.

More specifically: Currently, the SystemTestsCompareTwo infrastructure would set things up like this if we tried to write ERS using that framework:

```
Case directories:

ERS.f45_g37_rx1.A.cheyenne_intel
-- ERS.f45_g37_rx1.A.cheyenne_intel.rest

Build & run directories:

ERS.f45_g37_rx1.A.cheyenne_intel
-- bld
-- run

ERS.f45_g37_rx1.A.cheyenne_intel.rest
-- bld
-- run
```

What if we change the organization to this?

```
Case directories:

ERS.f45_g37_rx1.A.cheyenne_intel
-- rest_case
---- ERS.f45_g37_rx1.A.cheyenne_intel

Build & run directories:

ERS.f45_g37_rx1.A.cheyenne_intel
-- bld
-- run
-- rest_case
---- ERS.f45_g37_rx1.A.cheyenne_intel
------ bld
------ run
```

I **think** it works to have two cases have the same name as long as the case directories, bld & run directories all appear in different places, as laid out above. If so, this could be a way to have the clone case do a CONTINUE_RUN from the original case.

Does anyone see any critical flaws with this?

cc @mvertens @jedwards4b @jgfouca @rljacob @mfdeakin-sandia

jgfouca:
This required some significant changes, foremost it introduces
the concept of a cloned case to re-use the exact same RUNDIR as
its parent.

The other big change was adding the concept of disabling separate_runs
much like the preexisting option to disable separate_builds, in
SystemTestsCompareTwo.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1647 

User interface changes?: N

### DIFF
--- a/config/config_tests.xml
+++ b/config/config_tests.xml
@@ -244,7 +244,12 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
     <BFBFLAG>TRUE</BFBFLAG>
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>11</STOP_N>
+    <REST_N>$STOP_N / 2 + 1</REST_N>
+    <REST_OPTION>$STOP_OPTION</REST_OPTION>
+    <HIST_N>$STOP_N</HIST_N>
+    <HIST_OPTION>$STOP_OPTION</HIST_OPTION>
     <DOUT_S>FALSE</DOUT_S>
+    <CONTINUE_RUN>FALSE</CONTINUE_RUN>
   </test>
 
   <test NAME="ERS">
--- a/scripts/lib/CIME/SystemTests/erp.py
+++ b/scripts/lib/CIME/SystemTests/erp.py
@@ -23,7 +23,7 @@ class ERP(SystemTestsCompareTwo):
         """
         SystemTestsCompareTwo.__init__(self, case,
                                        separate_builds = True,
-                                       separate_runs = False,
+                                       separate_rundirs = False,
                                        run_two_suffix = 'rest',
                                        run_one_description = 'initial',
                                        run_two_description = 'restart')
--- a/scripts/lib/CIME/SystemTests/erp.py
+++ b/scripts/lib/CIME/SystemTests/erp.py
@@ -1,5 +1,5 @@
 """
-CIME ERP test.  This class inherits from SystemTestsCommon
+CIME ERP test.  This class inherits from SystemTestsCompareTwo
 
 This is a pes counts hybrid (open-MP/MPI) restart bfb test from
 startup.  This is just like an ERS test but the pe-counts/threading
@@ -8,132 +8,57 @@ count are modified on retart.
 (2) Do a restart test with half the number of tasks and threads (suffix rest)
 """
 
-import shutil
 from CIME.XML.standard_module_setup import *
 from CIME.case_setup import case_setup
-import CIME.utils
-from CIME.SystemTests.system_tests_common import SystemTestsCommon
+from CIME.SystemTests.system_tests_compare_two import SystemTestsCompareTwo
 from CIME.check_lockedfiles import *
 
 logger = logging.getLogger(__name__)
 
-class ERP(SystemTestsCommon):
+class ERP(SystemTestsCompareTwo):
 
     def __init__(self, case):
         """
         initialize a test object
         """
-        SystemTestsCommon.__init__(self, case)
+        SystemTestsCompareTwo.__init__(self, case,
+                                       separate_builds = True,
+                                       separate_runs = False,
+                                       run_two_suffix = 'rest',
+                                       run_one_description = 'initial',
+                                       run_two_description = 'restart')
 
-    def build_phase(self, sharedlib_only=False, model_only=False):
-        """
-        Build two cases.  Case one uses defaults, case2 uses half the number of threads
-        and tasks. This test will fail for components (e.g. pop) that do not reproduce exactly
-        with different numbers of mpi tasks.
-        """
+    def _common_setup(self):
         self._case.set_value("BUILD_THREADED",True)
-        if sharedlib_only:
-            return self.build_indv(sharedlib_only=sharedlib_only, model_only=model_only)
 
-        exeroot = self._case.get_value("EXEROOT")
-        cime_model = CIME.utils.get_model()
+    def _case_one_setup(self):
+        stop_n      = self._case.get_value("STOP_N")
 
-        # Make backup copies of the ORIGINAL env_mach_pes.xml and
-        # env_build.xml in LockedFiles if they are not there. If there
-        # are already copies there then simply copy them back to
-        # have the starting env_mach_pes.xml and env_build.xml
-        machpes1 = "env_mach_pes.ERP1.xml"
-        envbuild1 = "env_build.ERP1.xml"
-        if is_locked(machpes1):
-            restore(machpes1, newname="env_mach_pes.xml")
-        else:
-            lock_file("env_mach_pes.xml", newname=machpes1)
+        expect(stop_n > 2, "ERROR: stop_n value {:d} too short".format(stop_n))
 
-        if is_locked(envbuild1):
-            restore(envbuild1, newname="env_build.xml")
+    def _case_two_setup(self):
+        # halve the number of tasks and threads
+        for comp in self._case.get_values("COMP_CLASSES"):
+            ntasks    = self._case1.get_value("NTASKS_{}".format(comp))
+            nthreads  = self._case1.get_value("NTHRDS_{}".format(comp))
+            rootpe    = self._case1.get_value("ROOTPE_{}".format(comp))
+            if ( nthreads > 1 ):
+                self._case.set_value("NTHRDS_{}".format(comp), nthreads/2)
+            if ( ntasks > 1 ):
+                self._case.set_value("NTASKS_{}".format(comp), ntasks/2)
+                self._case.set_value("ROOTPE_{}".format(comp), rootpe/2)
 
-        # Build two executables, one using the original tasks and threads (ERP1) and
-        # one using the modified tasks and threads (ERP2)
-        # The reason we currently need two executables that CESM-CICE has a compile time decomposition
-        # For cases where ERP works, changing this decomposition will not affect answers, but it will
-        # affect the executable that is used
-        for bld in range(1,3):
-            logging.warn("Starting bld {}".format(bld))
+        stop_n = self._case.get_value("STOP_N")
 
-            if (bld == 2):
-                # halve the number of tasks and threads
-                for comp in self._case.get_values("COMP_CLASSES"):
-                    ntasks    = self._case.get_value("NTASKS_{}".format(comp))
-                    nthreads  = self._case.get_value("NTHRDS_{}".format(comp))
-                    rootpe    = self._case.get_value("ROOTPE_{}".format(comp))
-                    if ( nthreads > 1 ):
-                        self._case.set_value("NTHRDS_{}".format(comp), nthreads/2)
-                    if ( ntasks > 1 ):
-                        self._case.set_value("NTASKS_{}".format(comp), ntasks/2)
-                        self._case.set_value("ROOTPE_{}".format(comp), rootpe/2)
+        rest_n = stop_n/2 + 1
+        stop_new = stop_n - rest_n
+        expect(stop_new > 0, "ERROR: stop_n value {:d} too short {:d} {:d}".format(stop_new,stop_n,rest_n))
+        self._case.set_value("STOP_N", stop_new)
+        self._case.set_value("CONTINUE_RUN", True)
+        self._case.set_value("REST_OPTION","never")
 
-                # Note, some components, like CESM-CICE, have
-                # decomposition information in env_build.xml
-                # case_setup(self._case, test_mode=True, reset=True)that
-                # needs to be regenerated for the above new tasks and thread counts
-                case_setup(self._case, test_mode=True, reset=True)
-
-            # Now rebuild the system, given updated information in env_build.xml
-            self.build_indv(sharedlib_only=sharedlib_only, model_only=model_only)
-            shutil.move("{}/{}.exe".format(exeroot,cime_model),
-                        "{}/{}.ERP{}.exe".format(exeroot,cime_model,bld))
-
-            # Make copies of the new env_mach_pes.xml and the new
-            # env_build.xml to be used in the run phase
-            lock_file("env_mach_pes.xml", newname="env_mach_pes.ERP{}.xml".format(bld))
-            lock_file("env_build.xml", newname="env_build.ERP{}.xml".format(bld))
-
-    def run_phase(self):
-        # run will have values 1,2
-        for run in range(1,3):
-
-            expect(is_locked("env_mach_pes.ERP{:d}.xml".format(run)),
-                   "ERROR: LockedFiles/env_mach_pes.ERP{:d}.xml does not exist, run case.build".format(run ))
-
-            # Use the second env_mach_pes.xml and env_build.xml files
-            restore("env_mach_pes.ERP{:d}.xml".format(run), newname="env_mach_pes.xml")
-            restore("env_build.ERP{:d}.xml".format(run), newname="env_build.xml")
-
-            # update the case to use the new values
-            self._case.read_xml()
-
-            # Use the second executable that was created
-            exeroot = self._case.get_value("EXEROOT")
-            cime_model = CIME.utils.get_model()
-            exefile  = os.path.join(exeroot,"{}.exe".format(cime_model))
-            exefile2 = os.path.join(exeroot,"{}.ERP{:d}.exe".format(cime_model,run))
-            if (os.path.isfile(exefile)):
-                os.remove(exefile)
-            shutil.copy(exefile2, exefile)
-
-            stop_n      = self._case.get_value("STOP_N")
-            stop_option = self._case.get_value("STOP_OPTION")
-
-            if run == 1:
-                expect(stop_n > 2, "ERROR: stop_n value {:d} too short".format(stop_n))
-                rest_n = stop_n/2 + 1
-                self._case.set_value("REST_N", rest_n)
-                self._case.set_value("REST_OPTION", stop_option)
-                self._case.set_value("HIST_N", stop_n)
-                self._case.set_value("HIST_OPTION", stop_option)
-                self._case.set_value("CONTINUE_RUN", False)
-                suffix = "base"
-            else:
-                rest_n = stop_n/2 + 1
-                stop_new = stop_n - rest_n
-                expect(stop_new > 0, "ERROR: stop_n value {:d} too short {:d} {:d}".format(stop_new,stop_n,rest_n))
-                self._case.set_value("STOP_N", stop_new)
-                self._case.set_value("CONTINUE_RUN", True)
-                self._case.set_value("REST_OPTION","never")
-                suffix = "rest"
-
-            case_setup(self._case, test_mode=True, reset=True)
-
-            self.run_indv(suffix=suffix)
-
-        self._component_compare_test("base", "rest")
+        # Note, some components, like CESM-CICE, have
+        # decomposition information in env_build.xml
+        # case_setup(self._case, test_mode=True, reset=True)that
+        # needs to be regenerated for the above new tasks and thread counts
+        case_setup(self._case, test_mode=True, reset=True)
--- a/scripts/lib/CIME/SystemTests/erp.py
+++ b/scripts/lib/CIME/SystemTests/erp.py
@@ -57,6 +57,10 @@ class ERP(SystemTestsCompareTwo):
         self._case.set_value("CONTINUE_RUN", True)
         self._case.set_value("REST_OPTION","never")
 
+        # May need to change EXEROOT since separate_rundirs is False and separate_builds is True
+        if self._case1.get_value("EXEROOT") == self._case2.get_value("EXEROOT"):
+            self._case.set_value("EXEROOT", self._case.get_value("EXEROOT") + "_" + self._run_two_description)
+
         # Note, some components, like CESM-CICE, have
         # decomposition information in env_build.xml
         # case_setup(self._case, test_mode=True, reset=True)that
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -14,13 +14,24 @@ Classes that inherit from this are REQUIRED to implement the following methods:
 (2) _case_two_setup
     This method will be called to set up case 2, the "test" case
 
-In addition, they MAY require the following method:
+In addition, they MAY require the following methods:
 
 (1) _common_setup
     This method will be called to set up both cases. It should contain any setup
     that's needed in both cases. This is called before _case_one_setup or
     _case_two_setup.
 
+(2) _case_one_custom_prerun_action(self):
+    Use this to do arbitrary actions immediately before running case one
+
+(3) _case_two_custom_prerun_action(self):
+    Use this to do arbitrary actions immediately before running case two
+
+(4) _case_one_custom_postrun_action(self):
+    Use this to do arbitrary actions immediately after running case one
+
+(5) _case_two_custom_postrun_action(self):
+    Use this to do arbitrary actions immediately after running case two
 """
 
 from CIME.XML.standard_module_setup import *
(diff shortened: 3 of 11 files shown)


jgfouca:
This is worth a close look. Significant change.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -36,10 +36,12 @@ class SystemTestsCompareTwo(SystemTestsCommon):
 
     def __init__(self,
                  case,
-                 separate_builds,
+                 separate_builds=True,
```
jgfouca: Added new concepts: ability to use same rundir for both, multisubmit

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -132,6 +140,30 @@ def _common_setup(self):
         """
         pass
 
+    def _case_one_custom_prerun_action(self):
```
jgfouca: Add some new hooks into this class.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -161,28 +193,38 @@ def run_phase(self, success_change=False):  # pylint: disable=arguments-differ
...
-            check_case(self._case2, self._caseroot2)
-        self._force_case2_settings()
-        self.run_indv(suffix = self._run_two_suffix)
-
-        # Compare results
-        # Case1 is the "main" case, and we need to do the comparisons from there
-        self._activate_case1()
-        self._link_to_case2_output()
+        if not self._multisubmit or not first_phase:
+            with self._case2:
```
jgfouca: Key change, the with statement is needed so that case2 isn't read-only.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -198,7 +240,7 @@ def _get_caseroot2(self):
         casename1 = self._case1.get_value("CASE")
         caseroot1 = self._case1.get_value("CASEROOT")
 
-        casename2 = "{}.{}".format(casename1, self._run_two_suffix)
+        casename2 = "{}.{}".format(casename1, self._run_two_suffix) if self._separate_runs else casename1
```
jgfouca: case2 keeps same name if it's sharing rundir.

scripts/lib/CIME/case.py
```diff
@@ -1189,6 +1190,10 @@ def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None, cime
         else:
             newcase.set_value("BUILD_COMPLETE","FALSE")
 
+        if share_rundir:
```
jgfouca: Implementation of shared rundir for Case is here. Seems almost too easy?

scripts/lib/CIME/tests/SystemTests/test_system_tests_compare_two.py
```diff
@@ -395,3 +397,6 @@ def test_run2_fails(self):
         # Verify
         self.assertEqual(test_status.TEST_FAIL_STATUS,
                          mytest._test_status.get_status(test_status.RUN_PHASE))
+
+if __name__ == "__main__":
```
jgfouca: Add this so you can run the tests directly.

billsacks:
First off, thanks a lot for tackling this very difficult - but important - problem!

So do I understand right that, if we are running (for example) `ERS.f09_g16.X.cheyenne_intel`, then we'll have case directories:

`ERS.f09_g16.X.cheyenne_intel.testid` -- base case
`ERS.f09_g16.X.cheyenne_intel.testid/ERS.f09_g16.X.cheyenne_intel.testid` -- restart case

and both of them will point to the same run directory?

Actually, if I understand this right, they actually point to the same run directory for two reasons: (1) they have the same case name (unlike other uses of system_tests_compare_two, where the two cases have different case names); (2) they set the share_rundir argument to create_clone. It seems like (1) is necessary for the restart functionality to work (unless other code were added to make that work), but it seems to me like (2) is actually unnecessary given (1). Also, if I'm understanding this right, then it seems like the two cases will also share a build directory (by virtue of (1)), and so you shouldn't be allowed to set `separate_builds=True` if `separate_runs=False`.

If I understand this right, this is both clever and confusing. I'm not sure how I feel about it. My initial feeling is that I prefer to keep the build and run directories of the two cases separate, as suggested in #1647 , because that makes it easier to understand what's going on. I realize that would require some extra infrastructure to link to the necessary restart files... though much of this infrastructure could probably be borrowed from the short-term archiving infrastructure, which already knows how to identify all files needed for a restart. (I'm not suggesting actually using the short-term archiver, but rather calling whatever code gives the list of file names needed in the 'rest' directory, and using that list to determine the files to link to: see https://github.com/ESMCI/cime/issues/448#issuecomment-260838814 .) Can you comment on why you went this route of putting both cases in the same run directory rather than keeping them separate? Was it just because it involved fewer changes, or did you first try keeping them separate and run into some problems in doing so?

In any case, this definitely will need some careful thought before bringing it to master, given how critical the ERS and ERP tests are. I'm going to add some review comments tonight, but that's just from a quick initial pass, and doesn't at all constitute a sufficient review of these changes.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -49,17 +51,21 @@ def __init__(self,
...
+            separate_builds (bool): Whether separate runs are needed for the
+                two cases. If False, case2 uses the case1 run area. Can be useful for restart tests
             run_two_suffix (str, optional): Suffix appended to the case name for
                 the second run. Defaults to 'test'. This can be anything other
                 than 'base'.
             run_one_description (str, optional): Description printed to log file
                 when starting the first run. Defaults to ''.
             run_two_description (str, optional): Description printed to log file
                 when starting the second run. Defaults to ''.
+            multisubmit (bool): Do first and second runs as different submissions
```
billsacks: My understanding is that multi-submit depends on the test case setting RESUBMIT=1. If that's correct, then that should be added to the documentation here.

billsacks: Also, I'm a bit suspicious of whether the ERR and ERRI tests really work correctly with this new infrastructure. It looks to me like RESUBMIT will be set to 1 in both cases, since this is set via config_tests.xml. Does that do the right thing? If so, it's not very intuitive, and should be documented somewhere.

It also seems like ERR and ERRI are not re-runnable, because presumably RESUBMIT gets decremented to 0, and then if you try to rerun them they'd do the wrong thing because RESUBMIT would be 0 rather than 1. But if that's truly an issue, I don't think it's new to this PR.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -161,28 +193,38 @@ def run_phase(self, success_change=False):  # pylint: disable=arguments-differ
         Runs both phases of the two-phase test and compares their results
         If success_change is True, success requires some files to be different
         """
+        first_phase = self._case1.get_value("RESUBMIT") == 1
```
billsacks: This first_phase variable was confusing when I first read it. Maybe add a comment like, "Only relevant when self._multisubmit is True"?

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -161,28 +193,38 @@ def run_phase(self, success_change=False):  # pylint: disable=arguments-differ
...
+        run_type = self._case1.get_value("RUN_TYPE")
 
         # First run
-        logger.info('Doing first run: ' + self._run_one_description)
-        self._activate_case1()
-        run_type = self._case1.get_value("RUN_TYPE")
-        self.run_indv(suffix = self._run_one_suffix)
+        if not self._multisubmit or first_phase:
```
billsacks: You have introduced a fair amount of new complexity into run_phase. That's fine, but I'd like to see some extra tests covering this new stuff. Specifically (in test_system_tests_compare_two.py):

(1) I'd like to see `test_run_phase_internal_calls` extended so that it also checks that `_case_one_custom_prerun_action`, `_case_one_custom_postrun_action` and similarly for case_two are called at the right points in the sequence. I'm confident that the current code is correct in this respect, but I'd like to see this added to cover any future refactorings.

(2) I'd like to see one or more tests covering the case where multisubmit=True. Possibly this could be done via some new tests similar to `test_run_phase_internal_calls`, one with RESUBMIT=1 and one with RESUBMIT=0 -- making sure the right things do and don't happen in each case. Maybe there's something more clever that could be tested, too, though nothing pops to mind. As in (1), this is less for the sake of testing the correctness of the current code, and more for the sake of making sure that this remains correct under future refactorings, since we otherwise don't have any tests that test this code.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
billsacks: I find this long one-liner hard to read. Please separate it into multiple lines - i.e., a standard 'if' construct.

config/config_tests.xml
```diff
@@ -244,7 +244,12 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
     <BFBFLAG>TRUE</BFBFLAG>
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>11</STOP_N>
+    <REST_N>$STOP_N / 2 + 1</REST_N>
```
jgfouca: These were being set manually in the python. There's no reason to do that since it can be done in the XML. The tweaks needed for the second run of erp are still being done in case_two_setup.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -36,10 +36,12 @@ class SystemTestsCompareTwo(SystemTestsCommon):
 
     def __init__(self,
                  case,
-                 separate_builds,
+                 separate_builds=True,
+                 separate_runs=True,
```
jgfouca: Done.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -49,17 +51,21 @@ def __init__(self,
                 test. This is the main case associated with the test.
             separate_builds (bool): Whether separate builds are needed for the
                 two cases. If False, case2 uses the case1 executable.
+            separate_builds (bool): Whether separate runs are needed for the
```
jgfouca: Done.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
jgfouca: @billsacks I can't speak to err/erri working correctly; I can only saw confidently that the PR hasn't made the situation any worse that it was before. ERR should be re-runnable and I believe that's tested in the nightlies in Z_SystemTestFull. There's a mechanism in system_tests_common to detect if a test is being re-run and to reset all XML values based on config_tests.xml.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
jgfouca: Done.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
jgfouca: Done.

config/config_tests.xml
billsacks: I disagree in this case: At least some of these settings are only supposed to apply to one case, so I find it confusing to see them set in the xml, which presumably makes them apply to both cases.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
jgfouca: @billsacks This is tested by the ERR test in cime_developer. If you want to enhance the unittests for system_tests_compare_two, I'll leave that to you. I had a hard time figuring out what was going on in there.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
jgfouca: @billsacks, I understand your confusion; this one took me a while to figure out. Your a and b questions seem like the same question so I'll answer them together.

(a/b): This was exposed by the case.flush() in the NODEFAIL test. This flush will fail if the case is read-only.
(c): case1 is already in a writeable state. Recall that case1 is the original case while case2 is the close. That means case1 has already been opened with a with statement (if you look at the command-line tools in scripts, scripts/Tools, that's generally the first thing they do (take a caseroot and open a Case object with a with statement).

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
billsacks: Thanks for the clarification about rerunning. I'm still confused about how these tests work with the system_tests_compare_two infrastructure. Are you able to describe how this works? Specifically: It seems like both the base case and the restart case would have RESUBMIT set to 1. After the run of the base case, RESUBMIT would be decremented in the base case and the case would be resubmitted. Then the logic would cause it to execute the restart case. But then what happens with respect to RESUBMIT? Maybe the check for RESUBMIT happens in the base case, where it has already been decremented to 0, and so the RESUBMIT=1 value is ignored in the restart case? I could believe that. But if that's what's actually happening, it should be documented somewhere, because it's confusing.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
jgfouca: Why does everyone hate ternaries? :)

I'll leave this to the broader group since I've used this pattern in dozens of places. I personally like conciseness of a ternary for simple if/else assignments like this. 4 lines of code seems excessive to express such a simple concept.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
billsacks: Okay, thanks. Can you please add comments to this effect in the code so we don't have to dig up this PR discussion when we want to understand it in the future?

config/config_tests.xml
jgfouca: @billsacks I can understand that, but this approach is consistent with ERS and other tests.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
jgfouca: Done.

config/config_tests.xml
billsacks: Also, more generally, I find it confusing if critical test-specific settings are done in config_tests.xml, because I would expect to be able to understand the critical aspects of how a test operates by looking at that test's python file. But I realize that other people have different opinions on that...

jgfouca:
@billsacks I believe I addressed all your inline comments, so now on to your main comment.

Yes, your understanding of my approach is exactly right.

I selected this approach for a few reasons:
1) It seems the most elegant: no need to worry about copying files around or making special tweaks
2) It mimics the pre-PR behavior more closely; ERS and ERR were using the same run area for both runs
3) It minimized risk (in my eyes). I didn't need to worry if some artifact copied from the original rundir would force (or expect) the second phase to be run in the same dir. Likewise, don't have to worry about copying the wrong files or not enough files.

config/config_tests.xml
billsacks: @jgfouca Okay, I'll accept that.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
jgfouca: Yes, it is confusing. I believe the value of RESUBMIT in case2 is irrelevant. The resubmission logic is contained within the original case.

billsacks:
@jgfouca thanks a lot for your fast response to my comments. I agree that you have addressed most of my inline comments; the one that remains is adding some coverage of the new behavior via unit tests. I may take a stab at this, if Mariana gives her blessing for my spending more time on this. (I'm not supposed to be spending much time on cime stuff right now, and I think this (partial) review may have already used up my weekly allocation....)

As for my big, overall comment (https://github.com/ESMCI/cime/pull/1830#issuecomment-323628253): Thanks for describing your rationale. I'm coming around to this a bit. My main concern is that we don't have a precedent for two cases pointing into the same run area. I agree that it makes some things easier, but it also makes for more potential of the cases stomping on each other, and makes it harder to debug a failed run, in that the two cases aren't kept separate (no worse than the original ERS & ERP in this respect, but no better either). I want to see what others think about this (mainly @jedwards4b and @mvertens once they're back in the office).

There are two aspects of my big comment that you didn't address:

1. It seems that we don't actually need the argument to create_clone: won't the two cases point to the same run directory here because the cases have the same name? If you do keep the argument to create_clone, then you should add a comment around this line `        casename2 = "{}.{}".format(casename1, self._run_two_suffix) if self._separate_runs else casename1` that keeping the same case name is needed for restart tests – since otherwise this looks redundant and unnecessary.

2. Because the two cases have the same case name, they end up pointing to the same bld directory. So it seems like it won't work robustly to use this method for tests with `separate_builds=True`. **Actually, I just realized that the ERP test has separate_builds=True, which is needed for some sets of components. I feel like it won't work to do have these pointing to the same build directory, as they currently do.**

I'll also add a third comment, based on my experience now actually trying this out:

3. For the uninitiated, I feel it will be confusing to have a directory structure where (for example) `/glade/scratch/sacks/ERP.f09_g16.X.cheyenne_intel.20170821_122057_wklm6k` is the base case, and `/glade/scratch/sacks/ERP.f09_g16.X.cheyenne_intel.20170821_122057_wklm6k/ERP.f09_g16.X.cheyenne_intel.20170821_122057_wklm6k` is the restart case. There is no indication from the directory structure that the latter is the restart case, and instead you see the confusing situation where the case directory has a subdirectory with the same name. If we're going to keep this situation where the clone has the same name as the original case, then I feel it would be more clear if another level were added to the directory structure, so we had something like `/glade/scratch/sacks/ERP.f09_g16.X.cheyenne_intel.20170821_122057_wklm6k/rest_case/ERP.f09_g16.X.cheyenne_intel.20170821_122057_wklm6k`, as suggested by https://github.com/ESMCI/cime/issues/1647#issue-233941129 . If we did that, we might want to do it for all tests based on system_tests_compare_two, so that there is more consistency.

Finally, given this major overhaul, I feel it's important that some manual tests be done to confirm that at least ERS and ERP (and ideally the others, but I care most about ers and erp) still fail when they should. These manual tests can wait until everything else in this PR is finalized, though.

jgfouca:
@billsacks where are Jim and Marianna and when will they be back?

1. Done
2. That's an excellent point. I just tested ERP and only found one acme.exe in my area, so it didn't do two builds. I will fix this now.
3. Good idea, I will do that.

jgfouca:
@billsacks OK, I think I've addressed everything now.

scripts/lib/CIME/SystemTests/erp.py
```diff
@@ -57,6 +57,10 @@ def _case_two_setup(self):
         self._case.set_value("CONTINUE_RUN", True)
         self._case.set_value("REST_OPTION","never")
 
+        # May need to change EXEROOT since separate_rundirs is False and separate_builds is True
```
billsacks: I feel like this code belongs somewhere in system_tests_compare_two rather than in erp.py, so that other future tests that need it don't need to duplicate this. i.e., somewhere in system_tests_compare_two, if you have separate_rundirs=False and separate_builds=True, then it would set case2's EXEROOT like this. (Then you could possibly remove the error check that you just added about this in system_tests_compare_two.)

jgfouca: I thought to so, but it has to be in the setup (IE case_two_setup), so I didn't feel like I had much choice unless we wanted to provide a base-class version of case_two_setup.

billsacks:
@jgfouca - thanks, my comments from https://github.com/ESMCI/cime/pull/1830#issuecomment-323818141 are now mostly addressed. See my one inline comment. In addition, though, I'm wondering about this part of my earlier comment:

> It seems that we don't actually need the argument to create_clone: won't the two cases point to the same run directory here because the cases have the same name?

I partly ask this because, if it's not actually needed, then it might be good to remove this behavior from create_clone: It seems like, if possible, we should only have behavior in create_clone if it's something users might want, and I'm not sure that's the case here.

billsacks:
Oh, and Jim and Mariana are up in Wyoming watching the eclipse. I think they'll be back tomorrow.

scripts/lib/CIME/SystemTests/erp.py
billsacks: Would it work to put it in `_setup_cases`, either before or after the call to `_case_two_setup`?

jgfouca: Does changing EXEROOT make it necessary to call case_setup reset?

jgfouca:
@billsacks if @jedwards4b and @mvertens agreed that shared rundirs have no use in create_clone, I will remove it.

jedwards4b:
I've not yet fully reviewed this PR but I wonder if instead of reusing the same run directory we couldn't incorporate #1813 with this to also test st archiver functionality?

billsacks:
> I wonder if instead of reusing the same run directory we couldn't incorporate #1813 with this to also test st archiver functionality?

My current preferences - with (1) being most preferred - are:

1. Extracting just the piece of the short-term archiver that determines the list of files that are needed for restart, and having the ERS test (and related tests) call this directly, then using that list to make sym links to the necessary files.

2. Using the full short-term archiver, as @jedwards4b suggests.

3. Keeping the current implementation of having two cases write into a single run directory.

I rank (1) above (2) because (a) I prefer not to add another possible failure mode to the ERS and ERP tests, since that makes it harder to debug a failed test, and I'd prefer not to have our numerous ERS and ERP tests crippled by a bug in the short-term archiver; (b) running the short-term archiver would presumably be slower than just doing sym links.

However, I don't feel super-strongly about this, and am willing to accept any of these solutions.

jedwards4b:
@jgfouca @billsacks I think what I had in mind is some where between bills 1 and 2.   I'm going to checkout this branch and try my idea - I'll let you know.

billsacks:
@jgfouca - I talked to @jedwards4b and @mvertens . They were both against having two different cases point to a single run directory, either in testing or via create_clone.

So, our current feeling is that we should modify the behavior of these rewritten tests to use separate run directories - with the same case name, but in different locations. Then the ERS/ERP/etc. tests can use all or part of the short-term archiver. @jedwards4b proposes using the whole short-term archiver. However, I'm a little wary of that, because:

1. This adds additional failure modes to the ERS/ERP tests (if something goes wrong in short-term archiving)

2. It means you need to look in the archive location for history files produced during the base case, but not the restart case, which can be confusing to people trying to debug a run.

3. It slows down tests and takes up more disk space, due to copying files to the archive area, then copying restarts back. Note that some of our CLM tests produce a lot of history output, so the time and disk space needed to copy all of the history files over could be substantial.

So I'd personally prefer using just a subset of the short-term archiver, or using it in a slightly varied fashion. Some ideas of how to do this are:

1. Refactor the short-term archiver to separate the functionality of building a list of files that need to be copied to the rest directory. Then the test can call this and make sym links to the relevant files in the restart case. From an initial read, it looks like this refactor should be possible, though could be somewhat involved.

2. Add an option to the short-term archiver to only archive restart files, not history or log files.

3. Add an option to the short-term archiver to do sym links rather than copies (maybe implemented similar to the current copy_only option).

(2) and (3) could be combined.

jgfouca:
OK, I will refactor this PR when I get some time tmrw.

jedwards4b:
Check with me, I'm already working on it.

jgfouca:
@jedwards4b OK, that's fine. I will hold off until I hear something from you.

billsacks:
@jedwards4b - a note for the changes you're making: Personally, I think it will probably be best if you use the new organization for _all_ tests that use the system_test_compare_two infrastructure: i.e., always use the same case name for both, but with a nested directory structure, rather than using the old way for some tests and the new way for others. I think that having all tests be consistent in this respect will help both old and new developers. (I'm not sure if you were already doing that.)

jedwards4b:
New IRT test added in #1837

jedwards4b:
New ERP test added in #1845

PR #1843, opened 2017-08-28
Preserve test order in test_scheduler.

### LINKED ISSUE #1331: Set order that test suite builds / submits?
This is a very low priority issue that I'm bring over from the CESM-Development/cime issue page because it would be nice if this got addressed at some point.

> The aux_pop_MARBL test suite takes 40 minutes to build / submit 8 tests from a geyser node. Most of these are short, but we do have an ERI test that takes significantly longer than any other test to run (50 minutes instead of 5 or less)... so it would be nice to have that be the first test built and submitted. Currently it's the 5th test built, 20 minutes after the first job is submitted, and I don't see an easy way to change that. Perhaps a "priority" field in the xml file?

I'm not familiar with the parallelism in `create_test`; my impression is that the individual tests are built in parallel but processed serially (i.e. Test A is built in parallel and submitted, then Test B is built in parallel and submitted, ...). If the goal is to have all the tests processed in parallel, then this issue is not as important... though I'd argue being able to say "build / submit tests 1-4 first and 5-8 next" would still be useful.

jgfouca:
Also, sort acme tests by runtime so longer tests are started
earlier.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1331 

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/update_acme_tests.py
+++ b/scripts/lib/update_acme_tests.py
@@ -340,3 +340,25 @@ def get_recommended_test_time(test_full_name):
                     best_time = rec_time
 
     return best_time
+
+###############################################################################
+def sort_by_time(test_one, test_two):
+###############################################################################
+    """
+    >>> tests = get_full_test_names(["cime_tiny"], "melvin", "gnu")
+    >>> tests.extend(get_full_test_names(["cime_developer"], "melvin", "gnu"))
+    >>> tests.append("A.f19_f19.A.melvin_gnu")
+    >>> tests.sort(cmp=sort_by_time)
+    >>> tests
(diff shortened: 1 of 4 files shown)


billsacks:
This is great, thanks.

So how do we take advantage of this with CESM-style (xml-based) test lists? Is it sufficient to list the tests in the xml in the order in which we want them run?

scripts/lib/update_acme_tests.py
```diff
@@ -340,3 +340,25 @@ def get_recommended_test_time(test_full_name):
...
+###############################################################################
+def sort_by_time(test_one, test_two):
+###############################################################################
+    """
+    >>> tests = get_full_test_names(["cime_tiny"], "melvin", "gnu")
+    >>> tests.extend(get_full_test_names(["cime_developer"], "melvin", "gnu"))
+    >>> tests.append("A.f19_f19.A.melvin_gnu")
+    >>> tests.sort(cmp=sort_by_time)
+    >>> tests
+    ['DAE.f19_f19.A.melvin_gnu', 'ERI.f09_g16.X.melvin_gnu', 'ERIO.f09_g16.X.melvin_gnu', 'ERP.f45_g37_rx1.A.melvin_gnu', 'ERR.f45_g37_rx1.A.melvin_gnu', 'ERS.ne30_g16_rx1.A.melvin_gnu', 'IRT_N2.f19_g16_rx1.A.melvin_gnu', 'NCK_Ld3.f45_g37_rx1.A.melvin_gnu', 'PET_P32.f19_f19.A.melvin_gnu', 'PRE.f19_f19.ADESP.melvin_gnu', 'PRE.f19_f19.ADESP_TEST.melvin_gnu', 'SEQ_Ln9.f19_g16_rx1.A.melvin_gnu', 'SMS.T42_T42.S.melvin_gnu', 'SMS_D_Ln9.f19_g16_rx1.A.melvin_gnu', 'ERS.f19_g16_rx1.A.melvin_gnu', 'NCK.f19_g16_rx1.A.melvin_gnu', 'A.f19_f19.A.melvin_gnu']
```
rljacob: Does this list have to be kept in sync with something?

jgfouca: @rljacob yes, but this should be a trivial fix. scripts_regression_tests will catch it when cime_developer changes.

jgfouca:
@billsacks , @jedwards4b told me that CESM already returned the tests in the correct order, so I took his word for it.

PR #1852, opened 2017-08-31
Err 2 compare two

jedwards4b:
Convert the ERR and ERRI test to the compare2 framework

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/SystemTests/err.py
+++ b/scripts/lib/CIME/SystemTests/err.py
@@ -3,39 +3,40 @@ CIME ERR test  This class inherits from ERS
 ERR tests short term archiving and restart capabilities
 """
 from CIME.XML.standard_module_setup import *
-from CIME.SystemTests.ers import ERS
+from CIME.SystemTests.system_tests_compare_two import SystemTestsCompareTwo
+from CIME.SystemTests.irt import IRT
 
 import shutil, glob
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


scripts/lib/CIME/SystemTests/err.py
```diff
@@ -3,39 +3,33 @@
...
-import shutil, glob
+from CIME.SystemTests.system_tests_compare_two import SystemTestsCompareTwo
+from CIME.SystemTests.irt import IRT
+from CIME.case_st_archive import restore_from_archive
+from CIME.utils import ls_sorted_by_mtime
 
 logger = logging.getLogger(__name__)
 
-class ERR(ERS):
+class ERR(IRT):
```
billsacks: I find it more confusing than helpful that this inherits from IRT, particularly given that you override the most important method of IRT (`_case_one_custom_postrun_action`): this gives a false impression that this does something similar to IRT, when really the only functionality that it reuses seems to be the basic restart test stuff in _case_one_setup and _case_two_setup. I'd rather just see those duplicated. Or, if you'd like to avoid that duplication, you could make a base class for restart tests that has the body of IRT's _case_two_setup, and then have both IRT and ERR inherit from that restart test base class.

billsacks:
I'm happy with this now - thanks.

Issue #1853, opened 2017-08-31
st_archive: Failure to archive intermediate dart - and maybe other - restart files


billsacks:
There is a subtle issue is the current short-term archiver (case_st_archive.py) that prevents archiving intermediate restart files for dart, and possible other components.

The problem is in `_archive_process`: The relevant parts of the code look like this (some irrelevant pieces deleted for brevity):

```python
    for archive_entry in archive.get_entries():
        # determine compname and compclass
        compname, compclass = archive.get_entry_info(archive_entry)

        datenames = _get_datenames(case, last_date)
        for datename in datenames:

            # archive restarts
            histfiles_savein_rundir = _archive_restarts(case, archive, archive_entry,
                                                        compclass, compname,
                                                        datename, datename_is_last,
                                                        archive_file_fn)
```

The problem with this is that `_get_datenames` determines the list of dates based on what cpl restart files are present. If cpl is processed before other components, then those other components will only see the final restart files when they call `_get_datenames`, because all other cpl restart sets will have been moved.

I have only noticed this problem affecting dart restart sets, but in principle it seems that it could affect any components, depending on the ordering of the results from archive.get_entries()

I have a fix for this in an upcoming PR, where I have refactored `_archive_process` to have the date loop outside the component loop.

billsacks:
cc @mvertens @jedwards4b @bertinia

gold2718:
@billsacks, Would this also prevent archiving of intermediate initial files? I have a report from a user that intermediate CAM initial files are not being archived (i.e., only the last initial file written is archived). Since CAM uses initial files for data assimilation, this might be relevant to this issue.

billsacks:
> @billsacks, Would this also prevent archiving of intermediate initial files? I have a report from a user that intermediate CAM initial files are not being archived (i.e., only the last initial file written is archived). Since CAM uses initial files for data assimilation, this might be relevant to this issue.

Yes, I think the same mechanism is used for CAM initial files, so this issue would affect them (depending on the order in which entries were returned from archive.get_entries()). This should now be fixed on master, or if you want to test with a minimal set of changes, you can apply the minor diffs given by https://github.com/billsacks/cime/commit/02c1616

jedwards4b:
I am conducting this test now.

jedwards4b:
[...]

Issue #1851, opened 2017-08-30
Short-term archiver fails if there are no restart files


billsacks:
Currently, the short-term archiver fails in `_get_datenames` if there are no restart files.

While this shouldn't happen often in practice, it is possible that someone would do a run without any restart files, for which they want the history files archived.

I think that simply removing the error check wouldn't work, because of the logic in `_archive_process`, which only archives hist files if there is at least one datename.

I have a fix that involves refactoring `_archive_process`, which I want to do anyway to support rewriting ERP with system_tests_compare_two. I'll open a pr with that fix and some other refactoring of case_st_archive.

cc @jedwards4b @bertinia

Issue #1854, opened 2017-08-31
st_archive: DART intermediate rpointer files are rpointer.unset


billsacks:
Currently this issue is masked by https://github.com/ESMCI/cime/issues/1853 (which prevents DART intermediate restart / rpointer files from being archived at all); once that issue is resolved, this one will appear. As far as I can tell, this doesn't actually cause problems, but it does cause extraneous `rpointer.unset` files to appear in your archive directory.

This issue can be seen by introducing these diffs off of master: billsacks/cime@02c1616 . I plugged this into https://svn-ccsm-models.cgd.ucar.edu/clm2/trunk_tags/clm4_5_16_r253 but I think you could see the same issue from standalone cime. I then did:

```
./create_newcase --case check_clm_archiving_issue_0830b --compset I1850Clm45Bgc --res f10_f10_musgs --user-mods-dir /glade/scratch/sacks/cesm_code/clm4_5_16_r253_newCime2/components/clm/cime_config/testdefs/testmods_dirs/clm/default

cd check_clm_archiving_issue_0830b
./case.setup
./xmlchange HIST_OPTION=ndays,HIST_N=1,REST_OPTION=ndays,REST_N=1
./xmlchange DOUT_S_SAVE_INTERIM_RESTART_FILES=TRUE
./case.build && ./case.submit
```

Near the end of the short-term archiver output, I see:

```
-------------------------------------------
doing short term archiving for dart (esp)
-------------------------------------------
Archiving for date 0001-01-02-00000
writing rpointer_file /glade/scratch/sacks/archive/check_clm_archiving_issue_0830b/rest/0001-01-02-00000/rpointer.unset
Archiving for date 0001-01-03-00000
writing rpointer_file /glade/scratch/sacks/archive/check_clm_archiving_issue_0830b/rest/0001-01-03-00000/rpointer.unset
Archiving for date 0001-01-04-00000
writing rpointer_file /glade/scratch/sacks/archive/check_clm_archiving_issue_0830b/rest/0001-01-04-00000/rpointer.unset
Archiving for date 0001-01-05-00000
writing rpointer_file /glade/scratch/sacks/archive/check_clm_archiving_issue_0830b/rest/0001-01-05-00000/rpointer.unset
Archiving for date 0001-01-06-00000
```

and indeed, in my archive directory, I see a bunch of rpointer.unset files whose contents are simply `unset`.

[...]

PR #1845, opened 2017-08-29
mv erp test to compare two

jedwards4b:
Rewrite erp test using the compare two paradigm

Test suite: scripts_regression_tests.py, ERP_Ln9.f09_f09_mg17.F2000_DEV.cheyenne_intel.cam-outfrq9s
Test baseline: cesm2_0_alpha07d
Test namelist changes: 
Test status: bit for bit

Addresses #1647 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

PR #1855, opened 2017-08-31
Rewrite ERP with SystemTestsCompareTwo

billsacks:
Yet another attempt to rewrite the ERP test using the
SystemTestsCompareTwo infrastructure. If approved, this will supersede
#1830 and #1845.

This version is similar to the version of @jedwards4b in #1845. However,
rather than building our own mini-archiving functionality into
SystemTestsCompareTwo, I have refactored case_st_archive to extract a
function that can be called from SystemTestsCompareTwo. In this way,
there is very little test-specific code: we pretty much can just reuse
the existing logic for archiving restart files. This should reduce the
chance for errors and the maintenance cost compared with #1845.

In doing the refactoring for this, I stumbled upon and fixed two issues
with the short-term archiver: #1851 and #1853. Fixing #1853 uncovered
#1854, which is not fixed here.

Testing done: In addition to the standard scripts_regression_tests, I
have done the following testing, using
https://svn-ccsm-models.cgd.ucar.edu/clm2/trunk_tags/clm4_5_16_r253:

1. ERP_P24x2_Ld5.f10_f10.I1850Clm50Bgc.cheyenne_intel.clm-default
  - made sure it passes
  - made sure the two cases use the correct ntasks and nthreads
  - made sure it still passes if I rerun it

2. ERP_P24x2_Ln240.f10_f10.I1850Clm50Bgc.cheyenne_intel.clm-default
  - needs ./xmlchange GLC_NCPL=48,OCN_NCPL=48,ROF_NCPL=48

3. ERP_P24x2_Ld5.f10_f10.I1850Clm50Bgc.cheyenne_intel.clm-default with
  SourceMods to make the restart fail; confirmed it really does fail:

```diff
--- ../../../components/clm/src/biogeophys/WaterStateType.F902017-08-31 08:01:21.094907095 -0600
+++ SourceMods/src.clm/WaterStateType.F902017-08-31 08:54:32.910961000 -0600
@@ -862,7 +862,7 @@
     use clm_varcon       , only : denice, denh2o, pondmx, watmin, spval, nameg
     use landunit_varcon  , only : istcrop, istdlak, istsoil
     use column_varcon    , only : icol_roof, icol_sunwall, icol_shadewall
-    use clm_time_manager , only : is_first_step
+    use clm_time_manager , only : is_first_step, is_restart
     use clm_varctl       , only : bound_h2osoi
     use ncdio_pio        , only : file_desc_t, ncd_io, ncd_double
     use restUtilMod
@@ -875,7 +875,7 @@
     real(r8)         , intent(in)    :: watsat_col (bounds%begc:, 1:)  ! volumetric soil water at saturation (porosity)
     !
     ! !LOCAL VARIABLES:
-    integer  :: c,l,j,nlevs
+    integer  :: p,c,l,j,nlevs
     logical  :: readvar
     real(r8) :: maxwatsat    ! maximum porosity
     real(r8) :: excess       ! excess volumetric soil water
@@ -920,6 +920,14 @@
          long_name='canopy water', units='kg/m2', &
          interpinic_flag='interp', readvar=readvar, data=this%h2ocan_patch)

+    if (flag == 'read' .and. is_restart()) then
+       do p = bounds%begp, bounds%endp
+          if (this%h2ocan_patch(p) < 1._r8) then
+             this%h2ocan_patch(p) = this%h2ocan_patch(p) * (1._r8 + 1.e-14_r8)
+          end if
+       end do
+    end if
+
     call restartvar(ncid=ncid, flag=flag, varname='SNOCAN', xtype=ncd_double,  &
          dim1name='pft', &
          long_name='canopy snow water', units='kg/m2', &

--- ../../../components/clm/src/biogeophys/BalanceCheckMod.F902017-08-31 08:01:21.149685054 -0600
+++ SourceMods/src.clm/BalanceCheckMod.F902017-08-31 08:55:49.498168158 -0600
@@ -333,7 +333,6 @@
                   qflx_surf(indexc)+qflx_h2osfc_surf(indexc)+qflx_drain(indexc))

              write(iulog,*)'clm model is stopping'
-             call endrun(decomp_index=indexc, clmlevel=namec, msg=errmsg(sourcefile, __LINE__))

           else if (abs(errh2o(indexc)) > 1.e-5_r8 .and. (nstep > 2) ) then

@@ -361,7 +360,6 @@
              write(iulog,*)'qflx_snwcp_discarded_liq = ',qflx_snwcp_discarded_liq(indexc)*dtime
              write(iulog,*)'qflx_rootsoi_col(1:nlevsoil)  = ',qflx_rootsoi_col(indexc,:)*dtime
              write(iulog,*)'clm model is stopping'
-             call endrun(decomp_index=indexc, clmlevel=namec, msg=errmsg(sourcefile, __LINE__))
           end if
        end if
```

4. In this case, confirmed that the directory listings of both the
archive directory and the run directory (post-archiving) match master:

```
./create_newcase --case check_clm_archiving_0830a --compset I1850Clm45Bgc --res f10_f10_musgs --user-mods-dir /glade/scratch/sacks/cesm_code/clm4_5_16_r253_newCime2/components/clm/cime_config/testdefs/testmods_dirs/clm/default

./xmlchange HIST_OPTION=ndays,HIST_N=1,REST_OPTION=ndays,REST_N=1
```

5. Same as (4) but with:

```
./xmlchange DOUT_S_SAVE_INTERIM_RESTART_FILES=TRUE
```

In (5), differences from master are the presence of rpointer.unset files
in the interim restart directories. This is due to the pre-existing
issue #1854 (which was masked in master due to #1853)

Test suite: scripts_regression_tests.py on cheyenne
  Also tests listed above
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#1851
Fixes ESMCI/cime#1853
Addresses ESMCI/cime#1647 (rewrites ERP, not ERS)

User interface changes?: none

Code review:

### DIFF
--- a/config/cesm/config_archive.xml
+++ b/config/cesm/config_archive.xml
@@ -78,11 +78,11 @@
     <hist_file_extension>\.h.*.nc$|\.d[dovt]\.</hist_file_extension>
     <rest_history_varname>unset</rest_history_varname>
     <rpointer>
-      <rpointer_file>rpointer.ocn.restart$NINST_STRING</rpointer_file>
+      <rpointer_file>rpointer.ocn$NINST_STRING.restart</rpointer_file>
       <rpointer_content>./$CASE.pop$NINST_STRING.r.$DATENAME.nc,RESTART_FMT=nc</rpointer_content>
     </rpointer>
     <rpointer>
-      <rpointer_file>rpointer.ocn.ovf$NINST_STRING</rpointer_file>
+      <rpointer_file>rpointer.ocn$NINST_STRING.ovf</rpointer_file>
       <rpointer_content>./$CASE.pop$NINST_STRING.ro.$DATENAME</rpointer_content>
     </rpointer>
   </comp_archive_spec>
--- a/config/config_tests.xml
+++ b/config/config_tests.xml
@@ -245,6 +245,10 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>11</STOP_N>
     <DOUT_S>FALSE</DOUT_S>
+    <REST_N>$STOP_N / 2 + 1 </REST_N>
+    <REST_OPTION>$STOP_OPTION</REST_OPTION>
+    <HIST_OPTION>$STOP_OPTION</HIST_OPTION>
+    <HIST_N>$STOP_N</HIST_N>
   </test>
 
   <test NAME="ERS">
--- a/scripts/lib/CIME/SystemTests/erp.py
+++ b/scripts/lib/CIME/SystemTests/erp.py
@@ -1,5 +1,5 @@
 """
-CIME ERP test.  This class inherits from SystemTestsCommon
+CIME ERP test.  This class inherits from SystemTestsCompareTwo
 
 This is a pes counts hybrid (open-MP/MPI) restart bfb test from
 startup.  This is just like an ERS test but the pe-counts/threading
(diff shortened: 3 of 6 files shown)


billsacks:
@jedwards4b @jgfouca @mvertens @bertinia - I have added you all as reviewers, though I don't think all 4 of you need to review it. However, I'd love at least one (and preferably two) careful sets of eyes on my changes to case_st_archive, because I know there are some subtleties there that I might have missed.

config/cesm/config_archive.xml
```diff
@@ -78,11 +78,11 @@
     <hist_file_extension>\.h.*.nc$|\.d[dovt]\.</hist_file_extension>
     <rest_history_varname>unset</rest_history_varname>
     <rpointer>
-      <rpointer_file>rpointer.ocn.restart$NINST_STRING</rpointer_file>
+      <rpointer_file>rpointer.ocn$NINST_STRING.restart</rpointer_file>
```
billsacks: Changes in this file are unrelated fixes taken from @jedwards4b 's #1845

config/config_tests.xml
```diff
@@ -245,6 +245,10 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>11</STOP_N>
     <DOUT_S>FALSE</DOUT_S>
+    <REST_N>$STOP_N / 2 + 1 </REST_N>
```
billsacks: Changes in this file match @jedwards4b 's #1845

scripts/lib/CIME/SystemTests/erp.py
```diff
@@ -8,132 +8,58 @@
...
+        self._case.set_value("CONTINUE_RUN", True)
+        self._case.set_value("REST_OPTION","never")
+
+        # Note, some components, like CESM-CICE, have
+        # decomposition information in env_build.xml that
+        # needs to be regenerated for the above new tasks and thread counts
+        case_setup(self._case, test_mode=True, reset=True)
+
+    def _case_one_custom_postrun_action(self):
+        self.copy_case1_restarts_to_case2()
```
billsacks: Changes in this file basically match @jedwards4b 's #1845 , except for this line

scripts/lib/CIME/case_st_archive.py
```diff
@@ -247,6 +327,14 @@ def _archive_restarts(case, archive, archive_entry,
     # copy latest restart files to archive restart directory
     histfiles_savein_rundir = []
 
+    # determine function to use for last set of restart files
```
billsacks: This allows us to link rather than copy the restarts in erp and similar tests. This saves some time and disk space, but I'm not tied to this if others feel it adds too much complexity.

scripts/lib/CIME/case_st_archive.py
```diff
@@ -7,15 +7,43 @@
...
 
 logger = logging.getLogger(__name__)
 
 ###############################################################################
+def _get_archive_object(case):
+###############################################################################
+    """
+    Returns the EnvArchive object associated with this case
+    """
+    caseroot = case.get_value("CASEROOT")
```
jedwards4b: The archive object already exists in a case and you don't need to instantiate a new one
archive = case.get_env("archive")

scripts/lib/CIME/case_st_archive.py
```diff
@@ -341,50 +430,35 @@ def _archive_process(case, archive, last_date, archive_incomplete_logs, copy_onl
...
-                                                        archive_file_fn)
-
-            # if the last datename for restart files, then archive history files
-            # for this compname
-            if datename_is_last:
-                logger.info("histfiles_savein_rundir {} ".format(histfiles_savein_rundir))
-                _archive_history_files(case, archive, archive_entry,
-                                       compclass, compname, histfiles_savein_rundir,
-                                       archive_file_fn)
+    # archive restarts and all necessary associated files (e.g. rpointer files)
```
billsacks: This major refactor could use a careful review. Note that I have reversed the loop order, so that the date loop is now the outer loop. This was needed so that a test could easily call the restart archiving routine for a single date. Doing this refactor also incidentally fixed #1851 and #1853

scripts/lib/CIME/case_st_archive.py
```diff
@@ -410,6 +484,39 @@ def restore_from_archive(case, rest_dir=None):
         shutil.copy(item, rundir)
 
 ###############################################################################
+def archive_last_restarts(case, archive_restdir, link_to_restart_files=False):
```
billsacks: Most of the refactor supports sharing code between the real short-term archiving process and the erp test (and potentially future tests). This function is the one piece of new code that is NOT part of the standard short-term archiving process.

billsacks:
I'm done making line comments; this is ready for review.

scripts/lib/CIME/case_st_archive.py
```diff
@@ -62,6 +89,22 @@ def _get_ninst_info(case, compclass):
...
+    """
+    Each time this is generator function is called, it yields a tuple
+    (archive_entry, compname, compclass) for one component in this
+    case's compset components.
+    """
+    compset_comps = case.get_compset_components()
+    compset_comps.append('cpl')
+    compset_comps.append('dart')
+
+    for archive_entry in archive.get_entries():
```
jedwards4b: I think that this code is backward, instead of looping over all of the entries in env_archive and looking for a match in compname, one should just call self.get_optional_node('comp_archive_spec', attributes={"compname":compname}) for each compname.

scripts/lib/CIME/case_st_archive.py
```diff
@@ -7,15 +7,33 @@
...
 
 logger = logging.getLogger(__name__)
 
 ###############################################################################
+def _get_archive_file_fn(copy_only):
+###############################################################################
+    """
+    Returns the function to use for archiving some files
+    """
+    if copy_only is True:
```
jgfouca: Minor nit: "if copy_only is True" is a bit redundant, just do "if copy_only"

scripts/lib/CIME/case_st_archive.py
```diff
@@ -7,15 +7,33 @@
...
+        archive_file_fn = shutil.copyfile
+    else:
+        archive_file_fn = shutil.move
+    return archive_file_fn
```
jgfouca: I would replace this function implementation with:
```
return shutil.copyfile if copy_only else shutil.move
```

jedwards4b:
@jgfouca nit action taken

bertinia:
@billsacks as we discussed, it would be useful to run an ERR multi-instance, fully coupled
test to make sure all the changes to the STA get tested across all components.

jedwards4b:
@bertinia we have added to cime_config several fully active IRT test for this issue, including a multi-instance case.  It will currently fail because the pop model does not properly generate some of the multiinstance files - this is being worked on by pop developers.

billsacks:
For the record, I'm happy with @jedwards4b 's changes here - thank you for those!

@bertinia Based on the feedback from @jedwards4b I'll kick off ERR and IRT tests with single-instance, B compset; and ERR_N3 and IRT_N3 with F compsets. I'll let you know once those have passed.

billsacks:
> @bertinia Based on the feedback from @jedwards4b I'll kick off ERR and IRT tests with single-instance, B compset; and ERR_N3 and IRT_N3 with F compsets. I'll let you know once those have passed.

Actually, I'd prefer not to go down the rabbit hole of the ERR_N3 and IRT_N3 right now, since I don't see any such passing tests in our test suite.

jedwards4b:
IRT_N3_Ln9.f19_f19_mg17.F2000_DEV.cheyenne_intel.cam-outfrq9s   PASS
ERR_N3_Ln9.f19_f19_mg17.FHIST_DEV.cheyenne_intel.cam-outfrq9s PASS

billsacks:
Thanks for running those tests, @jedwards4b 

These two also pass, from cesm2_0_alpha07c with latest cime:

ERR.f19_g17.B1850.cheyenne_intel.allactive-defaultio
IRT.f19_g17.B1850.cheyenne_intel.allactive-defaultio

PR #1857, opened 2017-08-31
Erp to restart tests

### LINKED ISSUE #1856: SystemTestsCompareTwo multisubmit tries to do too much in phase 1
In comparing #1830 with what made it to master, I noticed that the indentation of this block is wrong:

```python
        # Compare results
        # Case1 is the "main" case, and we need to do the comparisons from there
        self._activate_case1()
        self._link_to_case2_output()

        self._component_compare_test(self._run_one_suffix, self._run_two_suffix, success_change=success_change)
```

-- this should be indented under the "Second run" conditional.

The current indentation leads the ERR test (and any other multi-submit test) to try to do component_compare_test after the first phase, leading to a FAIL result. This doesn't cause a test failure, because the FAIL is later overwritten with a PASS, but it is still incorrect.

I have a fix for this in an incoming PR.

jedwards4b:
merge new erp test into new restart_tests abstract class, fix indentation in system_test_compare_two

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1856 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -229,12 +229,12 @@ class SystemTestsCompareTwo(SystemTestsCommon):
                 self.run_indv(suffix = self._run_two_suffix)
                 self._case_two_custom_postrun_action()
 
-        # Compare results
-        # Case1 is the "main" case, and we need to do the comparisons from there
-        self._activate_case1()
-        self._link_to_case2_output()
+                # Compare results
+                # Case1 is the "main" case, and we need to do the comparisons from there
+                self._activate_case1()
+                self._link_to_case2_output()
 
-        self._component_compare_test(self._run_one_suffix, self._run_two_suffix, success_change=success_change)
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -229,12 +229,12 @@ def run_phase(self, success_change=False):  # pylint: disable=arguments-differ
                 self.run_indv(suffix = self._run_two_suffix)
                 self._case_two_custom_postrun_action()
 
-        # Compare results
-        # Case1 is the "main" case, and we need to do the comparisons from there
-        self._activate_case1()
-        self._link_to_case2_output()
+                # Compare results
+                # Case1 is the "main" case, and we need to do the comparisons from there
+                self._activate_case1()
```
billsacks: These lines need to be nested under the `if`, but I don't think they need to be nested under the `with` (and that's consistent with @jgfouca 's original implementation). See also #1858 

That said, I don't fully understand the subtleties here, so if you and @jgfouca think these lines should be nested under the `with`, then I'll accept that.

jedwards4b:
I'm fine with your version.

Issue #1846, opened 2017-08-29
Allow SCAM to run multiple IOPs with the same executable


mvertens:
This issue addresses a requirement that SCAM would like to have. They would like to create multiple cases with different SCAM IOPs - but using the same executable. Currently, the SCAM IOPs are obtained via different user_mods in the CAM user_mods directories. 
To address this use case - we would like to have the capability to have a "template" case that does not have any user_mods applied - and clone that directory multiple times with different user_mods - but with the executable created in the original create_newcase directory. 
To do this for SCAM - I propose that we create a "template" IOP that has NO user_mods. Build that case and the clone that case multiple times. 
The main new capability that this issue needs is to have a new user_mods argument to create_clone. 
Also - when --keepexe is added to create_clone
- any SourceMods in the original clone should point to the original template SourceMods. 
- the usermods directory CANNOT have any SourceMods (and an error should be thrown if that is the case)
- env_build.xml cannot modifications CANNOT be in the `shell_commands` file in usermods
- actually env_build.xml should not be allowed to have any modifications if `create_clone --keepexe `is used (so this new functionality needs to be added)

rljacob:
So what is the current workflow for doing different IOPs?

mvertens:
@rljacob - on the CESM side this is still being hashed out.  We have recently migrated SCAM away from using the stand-alone CAM scripts - and using CIME. The issue is that CAM can no longer configure either CLM or CICE - and so the stand-alone scripts would only permit CAM to configure an aquaplanet SCAM - which is very limiting. In addition, we are planning to remove all support for the CAM stand-alone scripts in the near future. The plan is to have a workflow with the above.

rljacob:
Is this it?
Old workflow:  build SCAM with CAM's custom scripts and driver.
current workflow:  Build SCAM with CIME (Create_newcase) but only aquaplanet available.
proposed workflow:  Build aquaplanet SCAM with create_newcase, use create_clone to then make land or sea-ice SCAM's.

jedwards4b:
SCAM = Single Column Atmospheric Model (there is by definition no scam land or scam sea-ice).  There are many possible single column locations and other settings that can be explored with a single executable,   This proposed workflow allows one to create a scam single executable and explore several specific cases through the create_clone tool.

rljacob:
But why isn't this a normal use of create_clone?  Clone my scam case, edit user_nl and .xml files to try a different location or setting?

jedwards4b:
[...]

PR #1866, opened 2017-09-03
add new user-mods option to create_clone 

mvertens:
New user-mods option to `create_clone`

This PR adds the capability to  clone a directory multiple times with different user_mods - but with the executable created in the original create_newcase directory. A new `user_mods` argument has been added to `create_clone`. If  `--keepexe `is an argument along with `--user_mods`  then
- any SourceMods in the original clone will  point to the original template SourceMods.
- the usermods directory CANNOT have any SourceMods (and an error us thrown if that is the case)
- env_build.xml  modifications CANNOT be in the usermods shell_commands file in usermods

In addition, this PR adds new functionality to scripts_regressions_tests and test_user_mod_support.py to test these new features.

Test suite: scripts_regression_tests (cheyenne)
Test baseline: No
Test namelist changes: None
Test status: bit for bit

Fixes: #1846 
User interface changes?: Yes - new argument to create_clone
Code review: @billsacks, @jedwards4b

### DIFF
--- a/scripts/create_clone
+++ b/scripts/create_clone
@@ -24,7 +24,7 @@ def parse_command_line(args):
                         "If not a full pathname, then the case to be cloned"
                         "is assumed to be under then current working directory ")
 
-    parser.add_argument("--user-mods-dir", "-user_mods_dir",
+    parser.add_argument("--user-mods-dir",
                         help="Path to directory with user_nl_* files and xmlchange "
                         "commands to utilize. This can also include SourceMods. "
                         "This can be an absolute path, a path relative to the "
@@ -32,11 +32,15 @@ def parse_command_line(args):
                         "cime_config/usermods_dirs/ under the primary component "
                         "for the given compset (for example, in an F compset "
                         "whose primary component is cam, '--user-mods-dir foo' "
-                        "could be found in components/cam/cime_config/usermods_dirs/foo).")
+                        "could be found in $SRCROOT/components/cam/cime_config/usermods_dirs/foo). "
+                        "If this argument is used in conjunction with the --keepexe flag, then "
+                        "no changes will be permitted to the env_build.xml in the --case entry directory. ")
 
     parser.add_argument("--keepexe", "-keepexe", action="store_true",
                         help="Sets EXEROOT to point to original build, it is highly recommended"
-                        "that the original case be built before cloning if using the --keepexe flag")
+                        "that the original case be built before cloning if using the --keepexe flag "
+                        "This flag will make SourceMods in the --case entry directory a symbolic link "
+                        " to the SourceMods in the --clone entry directory ")
 
     parser.add_argument("--mach-dir", "-mach_dir",
                         help="Specify the locations of the Machines directory, other than the default"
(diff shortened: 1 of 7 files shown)


scripts/create_clone
```diff
@@ -24,8 +24,19 @@ def parse_command_line(args):
                         "If not a full pathname, then the case to be cloned"
                         "is assumed to be under then current working directory ")
 
+    parser.add_argument("--user-mods-dir", "-user_mods_dir",
```
jedwards4b: Should we include here the limitations on user mods if the keepexe flag is used?

scripts/create_clone
```diff
@@ -51,20 +62,28 @@ def parse_command_line(args):
...
     expect(os.path.isdir(cloneroot),
            "Missing cloneroot directory %s " % cloneroot)
 
+    if user_mods_dir is not None:
+        if os.path.isdir(user_mods_dir):
+            user_mods_dir = os.path.abspath(user_mods_dir)
+
     with Case(cloneroot, read_only=False) as clone:
-        clone.create_clone(case, keepexe, mach_dir, project, cime_output_root)
+        if user_mods_dir is not None:
```
jedwards4b: Don't need this if else block.   clone.create_clone(case, keepexe, mach_dir, project, cime_output_root, user_mods_dir) will work in both cases.

scripts/lib/CIME/case.py
```diff
@@ -1225,6 +1230,33 @@ def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None, cime
         # lock env_case.xml in new case
         lock_file("env_case.xml", newcaseroot)
 
+        # apply user_mods if appropriate
+        newcase_root = newcase.get_value("CASEROOT")
+        if user_mods_dir is not None:
+            if keepexe:
+                # If keepexe CANNOT change any env_build.xml variables - so make a temporary copy of
+                # env_build.xml and verify that it has not been modified
+                shutil.copy(os.path.join(newcaseroot, "env_build.xml"),
```
jedwards4b: Is it possible that LockedFiles/env_build.xml already exists?   What is the result if that is that case?

scripts/tests/scripts_regression_tests.py
```diff
@@ -315,9 +315,11 @@ def test_c_create_clone_keepexe(self):
             shutil.rmtree(testdir)
         prevtestdir = cls._testdirs[0]
         cls._testdirs.append(testdir)
+        user_mods_dir = os.path.join(CIME.utils.get_python_libs_root(), "..", "tests", "user_mods_test3")
 
-        run_cmd_assert_result(self, "%s/create_clone --clone %s --case %s --keepexe" %
```
billsacks: Did you mean to delete this old test? Now we don't have a test that shows that using the `--keep-exe` flag leads to a pass. Maybe that's okay: it's not clear to me that this old test was providing much benefit (e.g., it didn't seem to even check anything to confirm that `--keep-exe` worked properly). But I wanted to make sure that its replacement was intentional.

scripts/create_clone
mvertens: I agree - I have put in this change and have removed the -user_mods_dir argument and only kept the --user-mods-dir argument. However, this is not consistent with the rest of the arguments in create_clone which all have a -xxx argument that accompanies the --xxx argument.

scripts/create_clone
mvertens: Good point.

scripts/tests/scripts_regression_tests.py
mvertens: @billsacks - @jedwards4b and I discussed this and decided that the next test captured the features of the old test as well as the new features.

scripts/create_clone
jedwards4b: The single dash variations were meant for backward compatibility and so don't need to be included with new options.

scripts/lib/CIME/case.py
mvertens: I have verified that this is not a problem.

mvertens:
@jedwards4b - I have updated the documentation for create_clone in the doc/ directory. I don't think there are any other outstanding issues.

PR #1875, opened 2017-09-06
improve error check for non-python sub processes

### LINKED ISSUE #1874: Deaths from perl-based build-namelists are silently suppressed
I noticed that `die` statements from perl-based build-namelists are being suppressed: their error message isn't printed, and the overall preview_namelists process simply moves on to the next component. This can lead to the given component's namelist file not being written at all, or - worse - it can be written incorrectly.

In CESM, here is a simple case that demonstrates the problem; you could probably come up with something similar in ACME:

With these diffs in components/cism/bld/build-namelist, off of the cesm2_0_alpha07d sandbox (also true in alpha07c):

```diff
Index: bld/build-namelist
===================================================================
--- bld/build-namelist	(revision 86564)
+++ bld/build-namelist	(working copy)
@@ -56,6 +56,10 @@
 EOF
 }

+print "I am about to die\n";
+die "a horrible death\n";
+print "I have been resurrected!\n";
+
 #-----------------------------------------------------------------------------------------------
 # Set some important constants, such as important values of certain config file parameters
```

I ran

```
./create_newcase --case test_die_0906a --compset T1850G --res f09_g16 --run-unsupported
cd test_die_0906a/
./case.setup
./preview_namelists
```

and got

```
Creating component namelists
   Calling /glade2/scratch2/sacks/cesm_code/cesm2_0_alpha07/cime/src/components/stub_comps/satm/cime_config/buildnml
   Calling /glade2/scratch2/sacks/cesm_code/cesm2_0_alpha07/cime/src/components/data_comps/dlnd/cime_config/buildnml
   Calling /glade2/scratch2/sacks/cesm_code/cesm2_0_alpha07/cime/src/components/stub_comps/sice/cime_config/buildnml
   Calling /glade2/scratch2/sacks/cesm_code/cesm2_0_alpha07/cime/src/components/stub_comps/socn/cime_config/buildnml
   Calling /glade2/scratch2/sacks/cesm_code/cesm2_0_alpha07/cime/src/components/stub_comps/srof/cime_config/buildnml
   Running /glade2/scratch2/sacks/cesm_code/cesm2_0_alpha07/components/cism/cime_config/buildnml
I am about to die
   Calling /glade2/scratch2/sacks/cesm_code/cesm2_0_alpha07/cime/src/components/stub_comps/swav/cime_config/buildnml
   Calling /glade2/scratch2/sacks/cesm_code/cesm2_0_alpha07/cime/src/components/stub_comps/sesp/cime_config/buildnml
   Calling /glade2/scratch2/sacks/cesm_code/cesm2_0_alpha07/cime/src/drivers/mct/cime_config/buildnml
Finished creating component namelists
```

No cism namelist file was created.

cc @mvertens @jedwards4b @fischer-ncar @jgfouca @rljacob

jedwards4b:
Subprocesses which are not python are not causing failures when they exit abnormally.
This fixes that problem for build-namelist.

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1874 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/utils.py
+++ b/scripts/lib/CIME/utils.py
@@ -199,8 +199,7 @@ def _convert_to_fd(filearg, from_dir):
 _hack=object()
 
 def run_sub_or_cmd(cmd, cmdargs, subname, subargs, logfile=None, case=None,
-                   input_str=None, from_dir=None, verbose=None,
-                   arg_stdout=_hack, arg_stderr=_hack, env=None, combine_output=False):
+                   from_dir=None, combine_output=False):
 
     # This code will try to import and run each buildnml as a subroutine
     # if that fails it will run it as a program in a seperate shell
@@ -227,7 +226,8 @@ def run_sub_or_cmd(cmd, cmdargs, subname, subargs, logfile=None, case=None,
         logger.info("   Running {} ".format(cmd))
         if case is not None:
             case.flush()
-        output = run_cmd_no_fail("{} {}".format(cmd, cmdargs), combine_output=True)
+        output = run_cmd_no_fail("{} {}".format(cmd, cmdargs), combine_output=combine_output,
+                                 from_dir=from_dir)
         logger.info(output)
         # refresh case xml object from file
         if case is not None:
--- a/scripts/lib/CIME/utils.py
+++ b/scripts/lib/CIME/utils.py
@@ -227,11 +227,7 @@ def run_sub_or_cmd(cmd, cmdargs, subname, subargs, logfile=None, case=None,
         logger.info("   Running {} ".format(cmd))
         if case is not None:
             case.flush()
-        stat, output, errput = run_cmd("{} {}".format(cmd, cmdargs), input_str=input_str, from_dir=from_dir,
-                                 verbose=verbose, arg_stdout=arg_stdout, arg_stderr=arg_stderr, env=env,
-                                 combine_output=combine_output)
-        expect(stat==0, errput)
-
+        output = run_cmd_no_fail("{} {}".format(cmd, cmdargs), combine_output=True)
         logger.info(output)
         # refresh case xml object from file
         if case is not None:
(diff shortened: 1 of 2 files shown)


jgfouca:
@jedwards4b see my comment on #1874. You probably want to turn combine_output on and maybe change the run_cmd to run_cmd_no_fail.

billsacks:
I'll wait to review until you've responded to @jgfouca

scripts/lib/CIME/utils.py
```diff
@@ -227,10 +227,7 @@ def run_sub_or_cmd(cmd, cmdargs, subname, subargs, logfile=None, case=None,
         logger.info("   Running {} ".format(cmd))
         if case is not None:
             case.flush()
-        stat, output, errput = run_cmd("{} {}".format(cmd, cmdargs), input_str=input_str, from_dir=from_dir,
-                                 verbose=verbose, arg_stdout=arg_stdout, arg_stderr=arg_stderr, env=env,
-                                 combine_output=combine_output)
-
+        output = run_cmd_no_fail("{} {}".format(cmd, cmdargs), combine_output=True)
```
jgfouca: I like this better, although it makes most of the optional arguments to run_sub_or_cmd useless. I'd recommend removing the ones that aren't being used.

jgfouca: FYI, pylint should be flagging the unused args.

scripts/lib/CIME/utils.py
```diff
@@ -199,8 +199,7 @@ def _convert_to_fd(filearg, from_dir):
 _hack=object()
 
 def run_sub_or_cmd(cmd, cmdargs, subname, subargs, logfile=None, case=None,
-                   input_str=None, from_dir=None, verbose=None,
-                   arg_stdout=_hack, arg_stderr=_hack, env=None, combine_output=False):
+                   from_dir=None, combine_output=False):
```
jgfouca: Do you think a user will ever not want to see errput from buildnml? If not, than we should removed the arg from this function and just set it to True in the run_cmd_no_fail call below.

jedwards4b: buildnml isn't necessarily the only user of run_sub_or_cmd

jgfouca: Oh ok.

jgfouca:
I accept.

PR #1878, opened 2017-09-08
Allow queue selection by node count instead of task count

### LINKED ISSUE #1861: Allow ability to select batch queue by num_nodes instead of tasks
Tasks is not necessarily a good metric on queue selection since the number of tasks per node will vary based on the number of threads.

jgfouca:
All the HPCs care about is how many nodes you request, so it makes
more sense to do selection by nodes instead of tasks. Also, the
mapping of tasks to nodes depends on the number of threads, so
task count alone is not a detemistic way of know how many nodes
you'll need.

1) Chama and skybridge to select queue based on node count
2) Change job-based task_count override to a node_count override
3) XSD changes to support these changes
4) Batch system setup must happen after initialize_derived_attributes (computes num_nodes)
5) Lots of changes to env_batch to support nodemin, nodemax

Test suite: scripts_regression_tests (TBD)
Test baseline: 
Test namelist changes: 

Fixes #1861 

User interface changes?: Yes, new ways to select queue in config_batch.xml

Update gh-pages html (Y/N)?:N

Code review: @jedwards4b

### DIFF
--- a/config/acme/machines/config_batch.xml
+++ b/config/acme/machines/config_batch.xml
@@ -445,17 +445,15 @@
     <!-- order matters, with no-batch jobs will be run in the order listed here -->
     <job name="case.run">
       <template>template.case.run</template>
-      <node_count>default</node_count>
       <prereq>$BUILD_COMPLETE and not $TEST</prereq>
     </job>
     <job name="case.test">
       <template>template.case.test</template>
-      <node_count>default</node_count>
       <prereq>$BUILD_COMPLETE and $TEST</prereq>
     </job>
     <job name="case.st_archive">
       <template>template.st_archive</template>
-      <node_count>1</node_count>
+      <task_count>1</task_count>
       <!-- If DOUT_S is true and case.run (or case.test) exits successfully then run st_archive-->
       <dependency>case.run case.test</dependency>
       <prereq>$DOUT_S</prereq>
--- a/config/acme/machines/config_batch.xml
+++ b/config/acme/machines/config_batch.xml
@@ -312,7 +312,8 @@
        <directive>--ntasks-per-node={{ tasks_per_node }}</directive>
     </directives>
     <queues>
-      <queue jobmin="1" jobmax="1024" walltimemax="06:00:00" default="true">ec</queue>
+      <queue nodemin="1" nodemax="12" walltimemax="04:00:00" default="true">short</queue>
+      <queue nodemin="1" nodemax="64" walltimemax="24:00:00">batch</queue>
     </queues>
   </batch_system>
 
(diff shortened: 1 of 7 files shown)


jgfouca:
@jedwards4b , skybridge is down so I'm not able to fully test just yet. by-hand testing worked on chama. I'll let you know when I get a chance to test. For now, just review.

config/acme/machines/config_batch.xml
```diff
@@ -435,25 +437,25 @@
...
     <job name="case.test">
       <template>template.case.test</template>
-      <task_count>default</task_count>
+      <node_count>default</node_count>
       <prereq>$BUILD_COMPLETE and $TEST</prereq>
     </job>
     <job name="case.st_archive">
       <template>template.st_archive</template>
-      <task_count>1</task_count>
+      <node_count>1</node_count>
```
jedwards4b: I don't think you want to change this, case.st_archive in particular is a single task job and other jobs you may optionally change here may need the finer control of task count.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -162,14 +162,15 @@ def make_batch_script(self, input_template, job, case, total_tasks, tasks_per_no
...
-        if task_count == "default":
+        if node_count == "default":
             self.total_tasks = total_tasks
             self.num_nodes = num_nodes
         else:
-            self.total_tasks = task_count
-            self.num_nodes = int(math.ceil(float(task_count)/float(tasks_per_node)))
+            self.total_tasks = 1
+            self.num_nodes = int(node_count)
+            expect(self.num_nodes == 1, "Invalid setting for node_count, must be '1' or 'default'")
```
jedwards4b: This restriction was never intended - why add it?

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -185,32 +186,34 @@ def make_batch_script(self, input_template, job, case, total_tasks, tasks_per_no
...
-            task_count = jsect["task_count"]
-            if task_count is None or task_count == "default":
+            node_count = jsect["node_count"]
+            if node_count is None or node_count == "default":
                 task_count = pesize
+                node_count = num_nodes
             else:
-                task_count = int(task_count)
+                task_count = 1
+                node_count = int(node_count)
```
jedwards4b: Although we have only used a task_count of 1 here, it was never intended that that be the only choice.

jgfouca:
@jedwards4b I agree this is a little bit awkward, but the old way had it's problems too. If you told st.archive job to run with a task_count of 2, it would still just run a single process. Maybe it makes sense to allow either?

jedwards4b:
@jgfouca but that's only because st.archive is a serial process.   Imagine for example a post processing step were added to the workflow and that postprocessing step could use 16 tasks.  What is the problem in the old way that you are trying to overcome?  This change doesn't seem to have anything to do with the queue selection issue.

jgfouca:
@jedwards4b this is ready for a re-review.

jgfouca:
I'll let you know when I get a chance to re-test. For now, just review.

jgfouca:
Tests passed, I'll click the button.

PR #1884, opened 2017-09-12
Add a timeout option to wait_for_tests

### LINKED ISSUE #1867: wait_for_tests needs a timeout option
We run scripts_regression_tests.py from cron on some batch systems.  Jobs submitted to the batch system from the test may end up in the queue and not run if resources are not available.  I think that we need to add a timeout to wait_for_tests so that the nightly cdash testing can report the partial results even if batch jobs fail to run within a defined time window.  This option needs to be available at the scripts_regression_tests.py command line.

jgfouca:
Our current timeout system is dependent on Jenkins hitting processes
with signals. This option will allow for a timeout mechanism independent
of Jenkins.

Also add a regression test for this feature.

Test suite: scripts_regression_tests M_TestWaitForTests, code_checker
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1867 

User interface changes?: Yes, new --timeout flag for wait_for_tests

Code review: @jedwards4b

### DIFF
--- a/scripts/tests/scripts_regression_tests.py
+++ b/scripts/tests/scripts_regression_tests.py
@@ -25,6 +25,7 @@ from  CIME.test_status import *
 SCRIPT_DIR  = CIME.utils.get_scripts_root()
 TOOLS_DIR   = os.path.join(SCRIPT_DIR,"Tools")
 TEST_COMPILER = None
+GLOBAL_TIMEOUT = None
 TEST_MPILIB = None
 MACHINE     = Machines()
 FAST_ONLY   = False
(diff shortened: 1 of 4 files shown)


scripts/tests/scripts_regression_tests.py
```diff
@@ -610,6 +610,12 @@ def test_wait_for_test_no_wait(self):
         self.simple_test(self._testdir_unfinished, expected_results, "-n")
 
     ###########################################################################
+    def test_wait_for_test_timeout(self):
```
jedwards4b: We need to add the --timeout flag to the scripts_regression_tests.py command line as well.

jgfouca:
@jedwards4b  OK, let me add that real quick.

scripts/tests/scripts_regression_tests.py
```diff
@@ -741,15 +742,7 @@ def setUp(self):
...
-        if TEST_COMPILER is None:
-            self._compiler          = MACHINE.get_default_compiler()
-        else:
-            self._compiler = TEST_COMPILER
-        if TEST_MPILIB is None:
-            self._mpilib = MACHINE.get_default_MPIlib()
-        else:
-            self._mpilib = TEST_MPILIB
-
+        self._compiler          = MACHINE.get_default_compiler() if TEST_COMPILER is None else TEST_COMPILER
```
jedwards4b: You left out the mpilib?

jedwards4b: Oh I see.

PR #1891, opened 2017-09-14
rename PES_PER_NODE to MAX_MPITASKS_PER_NODE

### LINKED ISSUE #1885: Rename PES_PER_NODE to MAX_MPITASKS_PER_NODE
It seems that the total number of MPI tasks is incorrect when ntasks is set to a negative value (number of machine nodes) and nthreads is greater than 1.  The total of  MPI tasks is nthreads times too large.  My workaround has been to set ntasks to the number of MPI tasks I want (positive), rather than the number of machine nodes.

jedwards4b:
Some people found the name PES_PER_NODE less than fully descriptive
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1885 

User interface changes?: 

Code review:

### DIFF
--- a/config/acme/allactive/config_pesall.xml
+++ b/config/acme/allactive/config_pesall.xml
@@ -4181,7 +4181,7 @@
     <mach name="anvil">
       <pes compset="any" pesize="S">
         <comment>ne30_ne30 grid on 45 nodes 32 ppn pure-MPI</comment>
-        <PES_PER_NODE>32</PES_PER_NODE>
+        <MAX_MPITASKS_PER_NODE>32</MAX_MPITASKS_PER_NODE>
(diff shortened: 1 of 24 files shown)


rljacob:
@jedwards4b please remember to add labels to your PRs.

rljacob:
Note @amametjanov

billsacks:
FYI, for future changes like this: This change is causing me a bit of pain with respect to my Mac port (in my .cime directory): I needed to change my config_machines.xml to get things working again, but then I need to revert back to my old config_machines.xml when running tests with code versions (e.g., of CLM) that use older versions of cime. This is a short-term pain that I can live with, but I think things like this need to be done with more backwards compatibility in mind in the future, particularly as more and more people may have their own machine ports.

PR #1895, opened 2017-09-15
Test fixes

### LINKED ISSUE #1890: remove debug print statements from docn_shr_mod.F90
These statements should not be in the code:
```
components/data_comps/docn/docn_shr_mod.F90:    write(6,*)'DEBUG: ocn_present is ',ocn_present
components/data_comps/docn/docn_shr_mod.F90:    write(6,*)'DEBUG: ocn_prognostic is ',ocn_prognostic
```

### LINKED ISSUE #1892: tests fail if --compare is used but baseline directory does not exist
Failing because cpl log file not found.

### LINKED ISSUE #1894: no error if ninst > ntasks in multidriver mode
is in 
   cime/scripts/lib/CIME/case_setup.py; line 139.

jedwards4b:
On yellowstone if the model run expected less that the allocated tasks for a job, it would use all the tasks anyway and extra tasks would just fall through to finalize.  This broke with the multi-driver PR.
The fix is to specify the number of tasks used on the mpirun command line - we already do this on 
all other machines so only yellowstone was affected.  There are also a couple of miscellaneous fixes
for multi driver and removing some debug print statements 

Test suite: scripts_regression_tests.py on yellowstone PEM_Ln9.f19_g16_rx1.A.yellowstone_intel
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1892 
Fixes #1890 
Fixes #1894

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -388,7 +388,8 @@ class SystemTestsCommon(object):
 
             # compare memory usage to baseline
             newestcpllogfiles = self._get_latest_cpl_logs()
-            memlist = self._get_mem_usage(newestcpllogfiles[0])
+            if len(newestcpllogfiles) > 0:
+                memlist = self._get_mem_usage(newestcpllogfiles[0])
             for cpllog in newestcpllogfiles:
                 m = re.search(r"/(cpl.*.log).*.gz",cpllog)
                 if m is not None:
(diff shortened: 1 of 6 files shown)


scripts/lib/CIME/SystemTests/system_tests_common.py
```diff
@@ -388,7 +388,8 @@ def _compare_baseline(self):
 
             # compare memory usage to baseline
             newestcpllogfiles = self._get_latest_cpl_logs()
-            memlist = self._get_mem_usage(newestcpllogfiles[0])
+            if len(newestcpllogfiles) > 0:
+                memlist = self._get_mem_usage(newestcpllogfiles[0])
```
jgfouca: Could this result in a variable-unset error below for memlist?

jedwards4b: No - memlist is not used if the newestcpllogfiles array is empty.

PR #1900, opened 2017-09-18
Ntasks per inst xxx

### LINKED ISSUE #1897: Request for derived variable: NTASKS_PER_INST_COMPNAME
Particularly with the new multi-driver code, it would be helpful if we had a set of derived variables (not to be set directly) in env_mach_pes.xml like `NTASKS_PER_INST_COMPNAME` for each component.

Reasons why this would be useful:

1. With the old multi-instance code, it was often confusing to users that the number of tasks per instance was _not_ NTASKS_COMPNAME but rather was (I think) floor(NTASKS_COMPNAME/NINST_COMPNAME). With the new multi-driver capability, this relationship is apparently more confusing: the number of tasks per instance is floor(NTASKS_COMPNAME/NINST_COMPNAME) if running without multi-driver, but is simply NTASKS_COMPNAME if running with multi-driver. Number of tasks per instance is something that users want to know, and it is asking a lot for them to all know this subtlety, so it would be great if this were listed explicitly for users.

2. Some components need to know the number of tasks per instance at build time. Before multi-instance, this was as simple as querying NTASKS_COMPNAME. With multi-instance, this needed to be expanded to also consider NINST_COMPNAME. With multi-driver, this gets more complicated, requiring all of these components to duplicate the logic where ntasks per instance depends on whether multi-driver is set. And what if the logic for number of tasks per instance changes again in the future? I am a firm believer in having any rule be coded in one and only one place, arguing for a central place in the cime scripts where the rule "number of tasks per instance" is encoded.

cc @gold2718 @dabail10 (for cice) @alperaltuntas (for pop, where this might also be an issue)

jedwards4b:
Add a derived variable NTASKS_PER_INST_COMP where COMP is the component name (ATM, LND, etc) 

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1897 
User interface changes?:  N

Code review:

### DIFF
--- a/src/drivers/mct/cime_config/config_component.xml
+++ b/src/drivers/mct/cime_config/config_component.xml
@@ -1928,7 +1928,7 @@
     </values>
     <group>mach_pes</group>
     <file>env_mach_pes.xml</file>
-<desc>Number of tasks per instance for each component. DO NOT EDIT: Set automatically by case.setup based on NTASKS, NINST and MULTI_DRIVER</desc>
+    <desc>Number of tasks per instance for each component. DO NOT EDIT: Set automatically by case.setup based on NTASKS, NINST and MULTI_DRIVER</desc>
   </entry>
 
   <entry id="NTHRDS">
--- a/src/drivers/mct/cime_config/config_component.xml
+++ b/src/drivers/mct/cime_config/config_component.xml
@@ -1928,7 +1928,7 @@
     </values>
     <group>mach_pes</group>
     <file>env_mach_pes.xml</file>
-    <desc>number of tasks per instance for each component (readonly)</desc>
+<desc>Number of tasks per instance for each component. DO NOT EDIT: Set automatically by case.setup based on NTASKS, NINST and MULTI_DRIVER</desc>
   </entry>
 
   <entry id="NTHRDS">
--- a/src/drivers/mct/cime_config/config_component.xml
+++ b/src/drivers/mct/cime_config/config_component.xml
@@ -1913,6 +1913,24 @@
     <desc>number of tasks for each component</desc>
   </entry>
 
+  <entry id="NTASKS_PER_INST">
+    <type>integer</type>
+    <values>
+      <value component="ATM"> 0</value>
+      <value component="CPL"> 0</value>
+      <value component="OCN"> 0</value>
+      <value component="WAV"> 0</value>
+      <value component="GLC"> 0</value>
+      <value component="ICE"> 0</value>
+      <value component="ROF"> 0</value>
+      <value component="LND"> 0</value>
+      <value component="ESP"> 0</value>
+    </values>
+    <group>mach_pes</group>
+    <file>env_mach_pes.xml</file>
+    <desc>number of tasks per instance for each component (readonly)</desc>
+  </entry>
+
   <entry id="NTHRDS">
     <type>integer</type>
     <values>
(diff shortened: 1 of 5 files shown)


billsacks:
@jedwards4b thanks a lot for taking this on.

One minor issue is that a basic xmlquery gives extra output (note the NINST print):

```
$ ./xmlquery NTASKS_PER_INST_GLC
NINST = 1
	NTASKS_PER_INST_GLC: 4
```

However, a bigger set of issues is that it leaves out some functionality that is typically provided for other xml variables:

1. Doing a partial query doesn't work (`xmlquery -p NTASKS_PER`, for example)

2. Listing NTASKS_PER_INST for all components doesn't work (`xmlquery NTASKS_PER_INST`)

3. Trying to get documentation of this variable with something like `./xmlquery --full NTASKS_PER_INST_GLC` gives an error:

```
$ ./xmlquery --full NTASKS_PER_INST_GLC
NINST = 1

Results in group none
  File "./xmlquery", line 365, in <module>
    _main_func()
  File "./xmlquery", line 352, in _main_func
    if results[group][var]['desc'][0] is not None:
IndexError: list index out of range
```

4. Auto-generated web documentation will not be created for these variables

My impression is that solving these problems would require making these "true" xml variables, rather than variables that are created on demand. This could follow the example of other derived variables, such as TOTALPES.

If others ( @gold2718 @mvertens @dabail10 @alperaltuntas @bertinia  for example) feel that these limitations are okay, then I can live with this. But I'm reluctant to sign off on this without approval from others.

jedwards4b:
I don't like the idea of making this a true xml variable because that implies that the user can change it (for example the user can change TOTALPES but it has no effect) and because I think that we should not introduce new xml variable(s) as a solution to every situation.

billsacks:
> I think that we should not introduce new xml variable(s) as a solution to every situation.

I agree with this in a general, philosophical sense. But in this case, part of the issue was having this be something that users can query. In that case, it seems useful to have some documentation on this, so that they know it can be queried.

But if others weigh in and say that there's no need for documentation of this new set of variables, then I'll defer to that opinion.

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -1913,6 +1913,24 @@
...
+      <value component="WAV"> 0</value>
+      <value component="GLC"> 0</value>
+      <value component="ICE"> 0</value>
+      <value component="ROF"> 0</value>
+      <value component="LND"> 0</value>
+      <value component="ESP"> 0</value>
+    </values>
+    <group>mach_pes</group>
+    <file>env_mach_pes.xml</file>
+    <desc>number of tasks per instance for each component (readonly)</desc>
```
billsacks: I'd like to see a bit more explicit documentation on how this is determined, as well as consistency with other options that say, "DO NOT EDIT". What about changing the description to:

```xml
<desc>Number of tasks per instance for each component. DO NOT EDIT: Set automatically by case.setup based on NTASKS, NINST and MULTI_DRIVER</desc>
```

jedwards4b:
Updated based on review - note that the comment in config_headers.xml about variables
MAX_TASKS_PER_NODE, MAX_MPITASKS_PER_NODE was incorrect and I have removed it.

billsacks:
Okay, this looks good now - thanks.

@jedwards4b did you run scripts_regression_tests on something close to the latest version (9beb46e or later), or only on the initial version?

jedwards4b:
@billsacks Just completed, all systems go.

PR #1901, opened 2017-09-18
We don't want check_if_comp_var to return a new vid if it's not a comp var

jgfouca:
This was causing very confusing messages from xmlquery.

Test suite: scripts_regression_tests J_TestCreateNewcase, code-checker
Test baseline: 
Test namelist changes: 
Test status: bit for bit 

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -112,14 +112,12 @@ class Case(object):
             self.initialize_derived_attributes()
 
     def check_if_comp_var(self, vid):
-        vid = vid
-        comp = None
-        iscompvar = False
         for env_file in self._env_entryid_files:
             new_vid, new_comp, iscompvar = env_file.check_if_comp_var(vid)
             if iscompvar:
                 return new_vid, new_comp, iscompvar
-        return vid, comp, iscompvar
+
+        return vid, None, False
 
     def initialize_derived_attributes(self):
         """
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -116,9 +116,9 @@ class Case(object):
         comp = None
         iscompvar = False
         for env_file in self._env_entryid_files:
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/case.py
```diff
@@ -116,9 +116,9 @@ def check_if_comp_var(self, vid):
         comp = None
         iscompvar = False
         for env_file in self._env_entryid_files:
-            vid, comp, iscompvar = env_file.check_if_comp_var(vid)
+            new_vid, new_comp, iscompvar = env_file.check_if_comp_var(vid)
```
jedwards4b: Wouldn't it be better to change this inside check_if_comp_var so that the original vid is returned if iscompvar is False?

jgfouca: Yeah, that's probably better.

jedwards4b:
```
diff --git a/scripts/lib/CIME/XML/env_base.py b/scripts/lib/CIME/XML/env_base.py
index 1540139c1..1622adbc3 100644
--- a/scripts/lib/CIME/XML/env_base.py
+++ b/scripts/lib/CIME/XML/env_base.py
@@ -46,13 +46,13 @@ class EnvBase(EntryID):
 
         for comp in self._components:
             if "_"+comp in vid:
-                vid = string.replace(vid, '_'+comp, '', 1)
+                new_vid = string.replace(vid, '_'+comp, '', 1)
                 break
             elif comp+"_" in vid:
-                vid = string.replace(vid, comp+'_', '', 1)
+                new_vid = string.replace(vid, comp+'_', '', 1)
                 break
-        if vid in self._component_value_list:
-            return vid, comp, True
+        if new_vid in self._component_value_list:
+            return new_vid, comp, True
         return vid, None, False
```

jgfouca:
OK, @jedwards4b please re-review.

PR #1902, opened 2017-09-19
Sarich/tools/load balancing tool

sarich:
Rewrite of the load balancing tool. This tool will model the component layouts as an linear program
and solve for optimum throughput. Timing information can be provided manually in a file or can be
generated through automatic submissions given a list of desired pe configurations.
The solve is no longer modeled in AMPL and solved by submitting to the NEOS server, instead
it is modeled and solved in Python using the PuLP package (https://pythonhosted.org/PuLP) which can
usually be installed with pip install --user pulp

Test suite: load_balancing_test.py (new), scripts_regression_tests.py

(Not sure I did this right)

Code review:


tools/load_balancing_tool/load_balancing_submit.py
```diff
@@ -0,0 +1,326 @@
...
+                        help='casename prefix to use for all timing runs')
+    parser.add_argument('--force_purge', action='store_true')
+
+    args = CIME.utils.parse_args_and_handle_standard_logging_options(args, parser)
+
+    return (args.compset, args.res, args.pesfile,
+            args.compiler, args.project, args.machine, args.extra_options_file,
+            args.casename_prefix, args.force_purge)
+
+def _set_xml_val(option, value, casedir):
```
jedwards4b: It's not clear to me why you are using a shell interface to xmlchange instead of instantiating a case object and using case.set_value?

tools/load_balancing_tool/load_balancing_submit.py
```diff
@@ -0,0 +1,326 @@
...
+                                "information available. Either run with "
+                                "--force_purge\n  (remove all timing runs) or "
+                                "remove this directory and try again", casedir)
+                sys.exit(1)
+
+        cmd = '%s/create_newcase --case %s --output-root %s ' % \
+              (script_dir, casename, casedir)
+        cmd += create_newcase_flags
+        logger.info(cmd)
+        run_cmd_no_fail(cmd, from_dir=script_dir)
```
jedwards4b: Why use this instead of calling case.configure directly?

tools/load_balancing_tool/load_balancing_submit.py
```diff
@@ -0,0 +1,326 @@
...
+        logger.info(cmd)
+        run_cmd_no_fail(cmd, from_dir=casedir)
+
+        cmd = './case.build'
+        logger.info(cmd)
+        run_cmd_no_fail(cmd, from_dir=casedir)
+
+        cmd = './case.submit'
+        logger.info(cmd)
+        run_cmd_no_fail(cmd, from_dir=casedir)
```
jedwards4b: I think that you should consider using the PFS system test instead of create_newcase - the test will do the setup build and submit steps internally.

tools/load_balancing_tool/optimize_model.py
```diff
@@ -0,0 +1,361 @@
...
+        raise NotImplementedError
+
+    def write_xml_changes(self, outfile):
+        """
+        Write out a list of xmlchange commands to implement
+        the optimal layout
+        """
+        raise NotImplementedError
+
+    def write_pe_template(self, pefilename, ntasks, nthrds, roots):
```
jedwards4b: Why are you reinventing the method to write env_mach_pes.xml - you should use CIME/XML/env_mach_pes.py

sarich: Excellent questions, I have no good answers. I will attempt the suggested edits, thanks.

tools/load_balancing_tool/tests/load_balancing_test.py
```diff
@@ -0,0 +1,288 @@
...
+            output = run_cmd_no_fail(cmd, from_dir=CODE_DIR)
+            self._check_solution(output, "NTASKS_ATM", 992)
+        
+    def test_xcase_submit(self):
+        with tempfile.NamedTemporaryFile('w+') as tfile, tempfile.NamedTemporaryFile('w+') as xfile:
+            tfile.write(PES_XML)
+            tfile.flush()
+            xfile.write(X_OPTIONS)
+            xfile.flush()
+            cmd = "./load_balancing_submit.py --pesfile %s --res ne30_g16 --compset X --casename_prefix test_lbt_  --extra_options_file %s" % (tfile.name, xfile.name)
```
jedwards4b: I don't think that this test should run by default under the scripts directory.   This directory may not be writable by the user.  This is a moot point if we choose to use the PFS test instead of create_newcase.

jedwards4b:
Working on the load_balancing_test.py I have a few issues.
I had to change ne30_g16 to f19_g16 because of a mapping file issue with ne30_g16 - that's not a problem with this PR.
When I ran the test I got an error 
```
OSError: [Errno 2] No such file or directory: '/glade/u/home/jedwards/sandboxes/cesm2_0_alpha/cime.jet/scripts/test_lbt_2/timing'' from dir '/glade/u/home/jedwards/sandboxes/cesm2_0_alpha/cime.jet/scripts/../tools/load_balancing_tool'
```
However I see that the directory is there and running load_balance_solve.py from the command line works - is there a timing issue here in which you are looking for timing files before the test_lbt_2 run is complete?

tools/load_balancing_tool/optimize_model.py
```diff
@@ -0,0 +1,361 @@
...
+
+            # add in data for ntasks=1 if not provided
+            if m.ntasks[0] > 1:
+                m.cost.insert(0, m.ntasks[0] * m.cost[0])
+                m.ntasks.insert(0, 1)
+                m.extrapolated.insert(0, True)
+
+            # add in data for maxtasks if not available
+            # assume same scaling factor as previous interval
+            if m.ntasks[-1] < self.maxtasks:
```
jedwards4b: There seems to be a bug here.  If len(m.ntasks) <= 1 then the code will crash on line 130.

sarich: Yes, that's a bug. Thanks for the catch.

sarich: I think the first failure with ne30_g16 left the directory in an unknown state that the solver didn't handle correctly. (the test_lbt_2 directory is there but not the timing directory). I need to fix this.

jedwards4b:
I've started a refactor of this code based on some of the ideas I've had while reviewing.  I think that the best approach is to accept this PR and offer my refactor as a separate one.  @rljacob do you want to review this before it's merged?

doc/source/misc_tools/load-balancing-tool.rst
```diff
@@ -0,0 +1,368 @@
...
+
+     --pesfile <filename>    (required)
+          This file is used to designated the pes layout that
+	  are used to create the timing data. The format is the same used
+	  by CIME pes_files, but note that the 'pesize' tag will be used
+	  to generate the casename. Also, this file will not be directly
+	  passed through to CIME, but rather it will trigger xmlchange
+	  commands to execute based on the values in the file.
+
+     --casename_prefix <prefix>
```
jedwards4b: Following cime conventions options should use dash '-' not underscore '_'

rljacob:
Did you try it out?  That's what I was going to do.

jedwards4b:
I did try it out - just a few minor problems.   I originally suggested merging this one and then my PR #1923 on top, but then I found that 1923 was ready so I would suggest closing this one without merge and taking that one instead.

sarich:
I think that @jedwards4b has supplanted this pr with
https://github.com/ESMCI/cime/pull/1923

sarich:
I thought we were rejecting this one

rljacob:
No I want this to go in and then the rewrite.

sarich:
OK

PR #1912, opened 2017-09-21
Err test rework

### LINKED ISSUE #1909: Want confirmation of some change made to system_tests_compare_two in #1904
@jedwards4b can you clarify / confirm this change that you made to system_tests_compare_two.py in #1904 

```diff
+            # Go to case2 for the generate and compare baselines
+            self._activate_case2()
```

I am pretty sure that is NOT what we want to do in general. system_tests_compare_two is set up to put all necessary files for comparison / generation in the case1 directory. I'm pretty sure that some things will break if this is done from case2. Can you clarify what the problem was that you were trying to solve with that change?

jedwards4b:
Addresses the problem in ERR test baseline generation and compare by linking the 
.base files created in the case1 run directory back to the original history file names. 

Test suite:  scripts_regression_tests.py ERR.f19_g16_rx1.A with generate and compare options.
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1909 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/SystemTests/err.py
+++ b/scripts/lib/CIME/SystemTests/err.py
@@ -2,6 +2,7 @@
 CIME ERR test  This class inherits from ERS
 ERR tests short term archiving and restart capabilities
 """
+import glob, os
 from CIME.XML.standard_module_setup import *
 from CIME.SystemTests.restart_tests import RestartTest
 from CIME.case_st_archive import restore_from_archive
@@ -31,4 +32,10 @@ class ERR(RestartTest):
             os.path.join(rest_root, restart_list[0]))
 
     def _case_two_custom_postrun_action(self):
-        restore_from_archive(self._case1)
+        # Link back to original case1 name
+        for case_file in glob.iglob(os.path.join(self._case1.get_value("RUNDIR"),
+                                                 "*.nc.{}".format(self._run_one_suffix))):
+            orig_file = case_file[:-(1+len(self._run_one_suffix))]
+            if not os.path.isfile(orig_file):
+                os.symlink(case_file, orig_file)
+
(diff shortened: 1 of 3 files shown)


scripts/lib/CIME/SystemTests/err.py
```diff
@@ -29,3 +30,12 @@ def _case_two_custom_prerun_action(self):
         expect(len(restart_list) >= 1, "No restart files found in {}".format(rest_root))
         restore_from_archive(self._case, rest_dir=
             os.path.join(rest_root, restart_list[0]))
+
+    def _case_two_custom_postrun_action(self):
+        # Link back to original case1 name
```
billsacks: Can you add a brief comment saying why this is needed - e.g.

```python
        # This is needed so that the necessary files are present for baseline comparison and generation,
        # since some of them may have been moved to the archive directory
```

scripts/lib/CIME/SystemTests/err.py
```diff
@@ -29,3 +30,12 @@ def _case_two_custom_prerun_action(self):
...
+        for case_file in glob.iglob(os.path.join(self._case1.get_value("RUNDIR"),
+                                                 "*.nc.{}".format(self._run_one_suffix))):
+            orig_file = case_file[:-(1+len(self._run_one_suffix))]
+            if not os.path.isfile(orig_file):
+                os.symlink(case_file, orig_file)
```
billsacks: Will a symlink work robustly if you try to rerun a test? My head is spinning trying to think this through, but I could foresee possible problems. A copy seems safer in case the test gets rerun, even though it's a bit more expensive.

jedwards4b:
@billsacks ready for rereview, I also moved the DOUT_S argument so that only case1 has DOUT_S=True.

jedwards4b:
All tests pass.

PR #1918, opened 2017-09-22
Big refactor to queue selection.

### LINKED ISSUE #1913: Queue selection is not updated when PE layout changes
I've hit a case that CIME does not handle very well at all. Here is the problematic pattern:

- create_test FOO.a.b
  - create_newcase ...
    - Selects a certain queue for case.test job
  - case.setup
  - case.build: FOO test case makes changes that require more nodes to be used, in this case the additional nodes mean that the queue selected during create_newcase is no longer able to run the job
  - case.submit FAILS

In order to solve the above I proposed:
* Changing all queue specifications to use nodemin and nodemax instead of min/max tasks
* Queue selections will be recomputed in case.initialize_derived_attributes unless the user has forced us to try to use a certain queue
  - initialze_derived_attributes always computes a fresh value for num_nodes, this value should then be fed into env_batch so that we can check if the queue selections for case.run and case.test are still valid

jgfouca:
Main goal was to dynamically reselect queue during case.setup. This allows us to
reselect an appropriate queue when PE layout changes are made.
    
Change list:
1) Remove jobmin and jobmax entirely from the system, nodemin and nodemax should be used instead
2) Remove output_error_path, it just obfuscated that it's the same as job_id
3) Refactor transform_vars and its usage.
    Remove check_members: allows us to avoid setting otherwise pointless members on our objects
4) Re-select queue for primary job (case.run/case.test) during case setup
5) Need to store user-requested queue and walltime so it can be reused during case setup queue selection

Test suite: scripts_regression_tests, melvin and chama, by-hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1913 

User interface changes?: Y, jobmin and jobmax gone from config_batch.xml

Update gh-pages html (Y/N)?:  N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -178,8 +178,8 @@ class EnvBatch(EnvBase):
         os.chmod(job, os.stat(job).st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
 
     def set_job_defaults(self, batch_jobs, case):
-        walltime    = case.get_value("USER_REQUESTED_WALLTIME")
-        force_queue = case.get_value("USER_REQUESTED_QUEUE")
+        walltime    = case.get_value("USER_REQUESTED_WALLTIME") if case.get_value("USER_REQUESTED_WALLTIME") else None
+        force_queue = case.get_value("USER_REQUESTED_QUEUE") if case.get_value("USER_REQUESTED_QUEUE") else None
 
         if self._batchtype is None:
             self._batchtype = self.get_batch_system_type()
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -21,8 +21,8 @@ class EnvBatch(EnvBase):
         """
         initialize an object interface to file env_batch.xml in the case directory
         """
-        self.prereq_jobid = None
-        self.batchtype = None
+        self._prereq_jobid = None
+        self._batchtype = None
         # This arbitrary setting should always be overwritten
         self._default_walltime = "00:20:00"
         schema = os.path.join(get_cime_root(), "config", "xml_schemas", "env_batch.xsd")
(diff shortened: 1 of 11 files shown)


jgfouca:
@jedwards4b tests still running but I think this is very close.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -155,68 +155,60 @@ def set_batch_system(self, batchobj, batch_system_type=None):
...
+        output_text = transform_vars(open(input_template,"r").read(), case=case, subgroup=job, overrides=overrides)
         with open(job, "w") as fd:
             fd.write(output_text)
         os.chmod(job, os.stat(job).st_mode | stat.S_IXUSR | stat.S_IXGRP | stat.S_IXOTH)
 
-    def set_job_defaults(self, batch_jobs, pesize=None, num_nodes=None, tasks_per_node=None, walltime=None, force_queue=None, allow_walltime_override=False):
-        if self.batchtype is None:
-            self.batchtype = self.get_batch_system_type()
+    def set_job_defaults(self, batch_jobs, case):
+        walltime    = case.get_value("USER_REQUESTED_WALLTIME") if case.get_value("USER_REQUESTED_WALLTIME") else None
```
jgfouca: This is where I had problems getting back empty string instead of None.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -155,68 +155,60 @@ def set_batch_system(self, batchobj, batch_system_type=None):
...
-        if task_count is None:
-            self.total_tasks = total_tasks
-            self.num_nodes = num_nodes
-        else:
-            self.total_tasks = int(task_count)
-            self.num_nodes = int(math.ceil(float(task_count)/float(tasks_per_node)))
-
-        self.pedocumentation = ""
-        self.job_id = case.get_value("CASE") + os.path.splitext(job)[1]
+        overrides["pedocumentation"] = "" # TODO?
```
jedwards4b: Not used - should we just remove?

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -2581,6 +2581,20 @@
   </entry>
 
  <!-- Job Submission stuff -->
+ <entry id="USER_REQUESTED_QUEUE">
+   <type>char</type>
```
jedwards4b: If you define an empty  <default_value></default_value> you should get None?

jedwards4b: Line 20 of entry_id.py sets value to "" if value is None - this seems wrong, at the very least it should only do that if the variable is a string.

scripts/lib/CIME/XML/env_batch.py
jgfouca: Yeah, I was about to remove it myself.

src/drivers/mct/cime_config/config_component.xml
jgfouca: OK, i'll make an issue for this.

Issue #1914, opened 2017-09-21
More path length issues with new system_tests_compare_two


billsacks:
We have been running into a number of path length issues with the new system_tests_compare_two, where CIME_OUTPUT_ROOT for case2 looks like `$CIME_OUTPUT_ROOT/TESTNAME/case2/TESTNAME`, with the bld and run directories under there. For long testnames, this creates some very long source file names, which exceed various maximum limits for compilers. The latest issue I'm running into is that CMake tries to compile a test program in a subdirectory of the build directory, and ifort barfs when the abs path to this test source file is too long.

billsacks:
I propose fixing this by moving the location of the case2 bld directory.

So we'll still have the run directory here: 

`$CIME_OUTPUT_ROOT/TESTNAME/case2/TESTNAME/run`

But we'll have the bld directory here:

`$CIME_OUTPUT_ROOT/TESTNAME/case2/bld`

I'll probably put in a symlink: `$CIME_OUTPUT_ROOT/TESTNAME/case2/TESTNAME/bld -> $CIME_OUTPUT_ROOT/TESTNAME/case2/bld` to make it easier for people to find the bld directory.

billsacks:
@fischer-ncar @ekluzek and @jedwards4b I think this will solve the cism intel build problems Chris ran into.

jedwards4b:
Maybe just change both EXEROOT and RUNDIR to take out the $CASE?
 ./xmlquery EXEROOT --no-resolve
	EXEROOT: $CIME_OUTPUT_ROOT/$CASE/bld
yslogin6: ~/sandboxes/cesm2_0_alpha/cime/scripts/ERR.f19_g16_rx1.A.yellowstone_intel.GC.20170921_090015_k45txl
:) ./xmlquery RUNDIR --no-resolve
	RUNDIR: $CIME_OUTPUT_ROOT/$CASE/run

I don't think that you need to add the link to the old path, no-one has been using this enough for that to be a difficulty.

billsacks:
@jedwards4b okay, I'm happy with that suggestion. So to be clear, we'd have:

Run directory:

`$CIME_OUTPUT_ROOT/TESTNAME/case2/run`

Bld directory:

`$CIME_OUTPUT_ROOT/TESTNAME/case2/bld`

Is that what you're thinking?

I'm thinking we'd still have case2's `CIME_OUTPUT_ROOT` be 

[...]

PR #1919, opened 2017-09-23
Shorter paths for case2 using system_tests_compare_two

billsacks:
The recently-changed version of system_tests_compare_two nested case2's
bld and run directories under
`$CIME_OUTPUT_ROOT/TESTNAME/case2/TESTNAME`. Since the TESTNAME is often
long, this was leading to very long paths to some things in the bld
directory. This was causing some compilation errors.

With this PR, case2's bld and run directories appear in
`$CIME_OUTPUT_ROOT/TESTNAME/case2/bld` and
`$CIME_OUTPUT_ROOT/TESTNAME/case2/run`. This shortens the bld directory
paths enough to get some formerly-failing builds to pass. The change in
run directory is just for consistency.

I was hoping to do this just via setting EXEROOT and RUNDIR in
system_tests_compare_two, to avoid cluttering up create_clone with
arguments that are just needed for tests. But it turns out that we need
to do this in create_clone, because create_clone calls case_setup, which
creates the EXEROOT and RUNDIR directories.

Also confirmed that the following tests build successfully:
`ERP.f10_f10_musgs.I1850Clm45BgcCru.yellowstone_pgi`
`ERP_P60x2_Lm36.f10_f10_musgs.I2000Clm50BgcCrop.yellowstone_intel.clm-clm50cropIrrigMonth_interp`
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#1914

User interface changes?: none

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -1395,9 +1395,11 @@ class Case(object):
                     logger.warn("Leaving broken case dir {}".format(self._caseroot))
 
             raise
-    def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None, cime_output_root=None,
-                 user_mods_dir=None):
+    def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None,
+                     cime_output_root=None, exeroot=None, rundir=None,
+                     user_mods_dir=None):
         """ moved to case_clone """
         return create_case_clone(self, newcase, keepexe=keepexe, mach_dir=mach_dir,
-                          project=project, cime_output_root=cime_output_root,
-                          user_mods_dir=user_mods_dir)
+                                 project=project, cime_output_root=cime_output_root,
+                                 exeroot=exeroot, rundir=rundir,
+                                 user_mods_dir=user_mods_dir)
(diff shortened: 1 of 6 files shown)


billsacks:
@fischer-ncar this fixes some of the build problems you have run into that are due to too-long paths in ERP tests.

scripts/lib/CIME/case.py
```diff
@@ -1395,9 +1395,11 @@ def create(self, casename, srcroot, compset_name, grid_name,
                     logger.warn("Leaving broken case dir {}".format(self._caseroot))
 
             raise
-    def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None, cime_output_root=None,
-                 user_mods_dir=None):
+    def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None,
```
jedwards4b: It's not clear to me why you added these options to create_clone instead of just setting
EXEROOT and BLDDIR in the _setup section of system_tests_compare_two.py?

billsacks: See the comments I made in the PR:

> I was hoping to do this just via setting EXEROOT and RUNDIR in
system_tests_compare_two, to avoid cluttering up create_clone with
arguments that are just needed for tests. But it turns out that we need
to do this in create_clone, because create_clone calls case_setup, which
creates the EXEROOT and RUNDIR directories.

PR #1926, opened 2017-09-26
Remove developers table on README.md

rljacob:
Remove the hard-to-maintain developers table and replace
with a alphabetical-by-last-name list of develoeprs in two
catagories: primary and also.

Also change ACME to E3SM

Test suite: none
Test baseline: N/A
Test status: BFB

Fixes ESCMI/cime#1562

Code review:

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,13 +1,14 @@
 # cime
 Common Infrastructure for Modeling the Earth
 
-**cime** (pronounced: seem) represents the infrastructure code for the 
+ **cime** (pronounced: seem)providess scripts for configuration, build, and testing of
+models via the command line, data and stub models for climate components,
+and a driver code for bringing all the climate components together in to a single executable.
+
+*cime* (pronounced: seem) is currently used by the
 <a href="http://www2.cesm.ucar.edu">Community Earth System Model </a>
-     (CESM) and the <a href="https://climatemodeling.science.energy.gov/projects/accelerated-climate-modeling-energy">
-Accelerated Climate Model for Energy </a> (ACME).
-*cime* providess scripts for configuration, build, and testing of
-models, as well as code for data and stub models for climate components,
-and a driver code for bringing all the climate components together in a single executable.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -1,48 +1,23 @@
 # cime
 Common Infrastructure for Modeling the Earth
 
-**cime** (pronounced: seem) represents the infrastructure code for the 
+ **cime** (pronounced: seem)providess scripts for configuration, build, and testing of
```
mvertens: I think that the right content here is to put in a more comprehensive statement - like what is contained at the top of the User's guide:
CIME, pronounced “SEAM”, contains the support scripts (configure, build, run, test), data models, essential utility libraries, a “main” and other tools that are needed to build a single-executable coupled Earth System Model. CIME is available in a stand-alone package that can be compiled and tested without active prognostic components but is typically included in the source of a climate model. CIME does not contain: any active components, any intra-component coupling capability (such as atmosphere physics-dynamics coupling).

rljacob: Sure I'll make that change.

rljacob:
It occurred to me that maybe we should list Primary developers by area?
Case Control System, Data Models and MCT-driver.  That covers the big three parts of CIME.  Then the "also contributing" category will include everyone who worked on any of those.

mvertens:
What about the fact that some of the developers have worked in all three
areas (like me?). Would it be odd to have the names repeated?

agsalin:
We need to add back in some attribution of who funded the development. DOE needs to be able to show that it contributed, which is why I added the paragraph and table in the first place. I agree the complexity of the table is no longer needed, and like the simple contributor lists. 

How about adding:
> CIME is jointly developed with support from the Earth System Modeling program of DOE's BER office and <some NCAR/UCAR/NSF attribution>.

README.md
mvertens: You did not make the above change . Could you please do that and I can then accept the PR.

rljacob:
@mvertens I think I fixed it.

Issue #1929, opened 2017-09-27
Add ability to query test types


billsacks:
There was a request from users of CLM to be able to get a list of the different test types (ERS, ERP, etc.) and their meaning.

Coming in a PR shortly....

rljacob:
That should also be in our Users's Guide under "testing".

billsacks:
@rljacob there is already a todo in the User's guide to add this. We also need documentation of query_testlists in there, which will presumably go where there is currently a place-holder for "manage_testlists" (which no longer works with new-style test lists). Do you think we should open a separate issue for these documentation needs?

As an aside: I started thinking about updating that page (http://esmci.github.io/cime/users_guide/testing.html) yesterday, but quickly got overwhelmed due to the completely different workflows in ACME vs. CESM. I'm not quite sure how to organize this to make it useful to both CESM and ACME users.

Issue #1685, opened 2017-06-13
Documentation for different Test Codes (SMS, ERS, ERI ...)


fischer-ncar:
I've been working on getting documentation for the different tests for create_test by going
through the comments in the scripts and an old README.  It would be great if people could
look through the different tests and fill in or fix any missing details for there favorite tests.

DAE
    Implementation of the CIME data assimilation test: Tests simple DA script
    which does not change the CAM input. Compares answers to non-DA run.
    
ERI   
    hybrid/branch/exact restart test, default (by default STOP_N is 22 days)
    (1) ref1case
         do an initial for ${STOP_N}/6 writing restarts at ${STOP_N}/6
         ref1 case is a clone of the main case (by default this will be 4 days)
         short term archiving is on
    (2) ref2case
         do a hybrid for ${STOP_N}-${STOP_N}/6 running with ref1 restarts from ${STOP_N}/6
         and writing restarts at ( ${STOP_N} - ${STOP_N}/6 )/2 +1
         (by default will run for 18 days and write a restart after 10 days)
         ref2 case is a clone of the main case
         short term archiving is on
    (3) case
         do a branch run starting from restart written in ref2 case
         and run for ???  days
    (4) case do a restart run from the branch case

ERIO
    tests restart with different PIO methods

ERP   
    pes counts hybrid (open-MP/MPI) restart bfb test from startup, default 6 days + 5 days (previousy PER)
    initial pes set up out of the box
    do an 11 day initial test - write a restart at day 6     (suffix base)
    half the number of tasks and threads for each component
    do a  5  day restart test starting from restart at day 6 (suffix rest)
       this is just like an ERS test but the pe-counts/threading count are modified on retart

ERR
    does an ERS test except that after the initial run the short term archive tool is run
    which moves model output out of the run directory into the short-term archive directory
    then the restart run is staged from the short term archive directory.  In batch mode there are
    four submitted jobs for this test (mira excepted) these are run1, sta1, run2 and sta2
    run1 and sta1 are submitted together with RESUBMIT=1.  sta1 has a batch system dependancy
    on successful completion of run1, when sta1 is completed it uses the cime resubmit capabilty
    to submit run2.

ERS   
    exact restart from startup (default 6 days + 5 days)
    do an 11 day initial test - write a restart at day 6     (suffix: base)
    if $IOP_ON is set then suffix is base_iop
    do a  5  day restart test starting from restart at day 6 (suffix: rest)
[...]

PR #1930, opened 2017-09-27
Add option to define test types in query_testlists

billsacks:
As far as I could tell, we didn't have any way to query the different
test types and their meaning (ERS, ERP, etc.). This PR adds this
capability to query_testlists.

I'm not sure if query_testlists is exactly the right place for this, but
I couldn't think of anywhere better. I thought about query_config, but I
didn't want to clutter that tool with this option that is more for
developers than users. So I have implemented this as an optional header
that can be output with query_testlists. I'm open to suggestions for a
better user interface.

The output looks like this:

```
$ ./query_testlists --define-testtypes
########################################################################
Test types
----------
DAE: data assimilation test, default 1 day, two DA cycles, no data modification
ERI: hybrid/branch/exact restart test, default 3+19/10+9/5+4 days
ERP: pes counts hybrid (open-MP/MPI) restart bfb test from startup, default 6 days + 5 days
ERS: exact restart from startup, default 6 days + 5 days
ERS2: exact restart from startup, default 6 days + 5 days
IRT: exact restart from startup, default 4 days + 7 days with restart from interim file
ERIO: exact restart from startup with different PIO methods, default 6 days + 5 days
ERR: exact restart from startup with resubmit, default 4 days + 3 days
ERRI: exact restart from startup with resubmit, default 4 days + 3 days. Tests incomplete logs option for st_archive
ERT: exact restart from startup, default 2 month + 1 month (ERS with info dbug = 1)
HOMME: Run homme tests.
PRE: pause/resume test, default 5 hours, five pause/resume cycles, no data modification
ICP: cice performance test
PEA: single pe bfb test (default length)
PEM: pes counts mpi bfb test (seq tests; default length)
PET: openmp bfb test (seq tests; default length)
PFS: performance test setup
MCC: multi-driver validation vs single-instance (default length)
NCK: multi-instance validation vs single instance (default length)
NCR: multi-instance validation sequential vs concurrent (default length)
OCP: pop performance test
REP: reproducibility test: do two runs give the same answers?
SBN: smoke build-namelist test (just run preview_namelist and check_input_data)
SEQ: sequencing bfb test (10 day seq,conc tests)
SMS: smoke startup test (default length)
########################################################################
jim        : SMS.f19_g16.X.cheyenne_intel
jim        : SMS.f19_g16.X.cheyenne_gnu
jim        : SMS.f19_g16.X.cheyenne_pgi
pauseresume: PRE.f09_f09.ADESP.hobart_nag
pauseresume: PRE.f09_f09.ADESP_TEST.hobart_nag
prealpha   : ERS_C3.f19_g17_rx1.A.cheyenne_pgi
```

(and then a lot more tests following that, since this lists out all the
tests defined in cime).

Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#1929

User interface changes?: adds --define-testtypes option to query_testlists

Code review:

### DIFF
--- a/config/config_tests.xml
+++ b/config/config_tests.xml
@@ -339,7 +339,7 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
   </test>
 
   <test NAME="HOMME">
-    <DESC>Run homme tests.</DESC>
+    <DESC>Run homme tests. Only works with the ACME version of the atmosphere component.</DESC>
     <INFO_DBUG>1</INFO_DBUG>
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>11</STOP_N>
--- a/config/config_tests.xml
+++ b/config/config_tests.xml
@@ -16,6 +16,9 @@ The following are the test functionality categories:
   11) other component-specific tests
 
 NOTES:
+- specifying the attribute INFRASTRUCTURE_TEST="TRUE" declares the given
+  test type to be an infrastructure-only test that is only meant to be
+  run by scripts_regression_tests.
 - unless otherwise noted everything is run in one executable directory
 - suffix: denotes the component history file suffixes that are added as part of the test
 - IOP test is done along with regular tests - not as a separate test
@@ -354,7 +357,7 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
     <DOUT_S>FALSE</DOUT_S>
   </test>
 
-  <test NAME="TESTBUILDFAIL">
+  <test NAME="TESTBUILDFAIL" INFRASTRUCTURE_TEST="TRUE">
     <DESC>For testing infra only. Insta-fail build step.</DESC>
     <INFO_DBUG>1</INFO_DBUG>
     <STOP_OPTION>ndays</STOP_OPTION>
@@ -363,7 +366,7 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
     <DOUT_S>FALSE</DOUT_S>
   </test>
 
-  <test NAME="TESTBUILDFAILEXC">
+  <test NAME="TESTBUILDFAILEXC" INFRASTRUCTURE_TEST="TRUE">
     <DESC>For testing infra only. Insta-fail build step by failing to init.</DESC>
     <INFO_DBUG>1</INFO_DBUG>
     <STOP_OPTION>ndays</STOP_OPTION>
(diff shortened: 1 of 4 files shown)


config/config_tests.xml
```diff
@@ -354,7 +357,7 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
     <DOUT_S>FALSE</DOUT_S>
   </test>
 
-  <test NAME="TESTBUILDFAIL">
+  <test NAME="TESTBUILDFAIL" INFRASTRUCTURE_TEST="TRUE">
```
billsacks: I don't know if this was the right way to do this. I wanted some way to exclude these infrastructure-only tests, because otherwise they clutter the output. I'm open to suggestions of a better way to denote this.

jgfouca: Seems like a decent approach to me. The other way would be check if the test is a subclass of FakeTest.

scripts/query_testlists
```diff
@@ -61,15 +67,38 @@ def parse_command_line(args, description):
...
+    """Ensures there are no incompatible arguments
+
+    If incompatible arguments are found, aborts with a helpful error
+    message.
+    """
+
+    expect(not(args.count and args.list_type),
+           "Cannot specify both --count and --list arguments.")
+
+    if args.count:
```
jgfouca: This might be cleaner to achieve with ArgParse's mutually exclusive group (see case.build).

billsacks:
> Nearly all of the tests except "homme" work with stand alone cime. "homme" in particular only works with the acme version of the atmosphere component. Can we make that clear in this output?

Would the right thing to do here be to move the homme test into ACME's CAM's cime_config? In the meantime, I'll just edit the DESC string for that test to add, "Only works with the ACME version of the atmosphere component."

billsacks:
I just rediscovered #1685 after submitting this PR. I still feel like it's useful to have this information available via query_testlists, but maybe we want it in create_test, too? @jedwards4b @jgfouca @fischer-ncar if it's okay with you, I'll leave the extension to create_test up to @fischer-ncar as part of resolving #1685 . You should be able to reuse much of what I did here.

scripts/query_testlists
billsacks: That's a useful-looking feature. But I'm struggling with how to do that here. I see how I can replace this with a mutually exclusive group:

```python
    expect(not(args.count and args.list_type),
           "Cannot specify both --count and --list arguments.")
```

But I can't see how to express the rest of the logic with that mechanism: Although it's not allowed to combine either `--show-options` or `--define-testtypes` with either `--list` or `--count`, it _is_ allowed to combine `--show-options` with `--define-testtypes`. Can you see how to express this with a mutually exclusive group?

jgfouca: OK, if it's not obvious then don't worry about it.

PR #1932, opened 2017-09-27
Add ability to cancel batch jobs to the system

jgfouca:
Jenkins needs this for finer control of batched jobs in
the presence of timeouts and potentially multiple concerrunt
runs of jenkins_generic_job.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: Yes, new batch_cancel field in config_batch

Update gh-pages html (Y/N)?:N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/jenkins_generic_job.py
+++ b/scripts/lib/jenkins_generic_job.py
@@ -10,7 +10,7 @@ def cleanup_queue(test_root, test_id):
     """
     Delete all jobs left in the queue
     """
-    for teststatus_file in glob.glob("{}/*{}*/TestStatus".format(test_root, test_id)):
+    for teststatus_file in glob.iglob("{}/*{}*/TestStatus".format(test_root, test_id)):
         case_dir = os.path.dirname(teststatus_file)
         with Case(case_dir, read_only=True) as case:
             jobmap = case.get_job_info()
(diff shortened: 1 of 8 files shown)


scripts/lib/jenkins_generic_job.py
```diff
@@ -1,22 +1,25 @@
...
 ###############################################################################
-def cleanup_queue(set_of_jobs_we_created):
+def cleanup_queue(test_root, test_id):
 ###############################################################################
     """
     Delete all jobs left in the queue
     """
-    current_jobs = set(CIME.utils.get_my_queued_jobs())
-    jobs_to_delete = set_of_jobs_we_created & current_jobs
+    for teststatus_file in glob.glob("{}/*{}*/TestStatus".format(test_root, test_id)):
```
jedwards4b: I think that this should be glob.iglob ?

jgfouca: Why iglob?

jedwards4b:
```
glob.iglob(pathname)
Return an iterator which yields the same values as glob() without actually storing them all simultaneously.
```
seems that is all that you need here.

jgfouca: Thanks, yes that would be better.

jgfouca: Done.

PR #1936, opened 2017-09-29
Centos7 linux

jedwards4b:
A port to a generic Linux workstation.

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -191,6 +191,64 @@
     </environment_variables>
   </machine>
 
+  <machine MACH="centos7_linux">
+    <DESC>
+	Example port to centos7 linux system with gcc, netcdf, pnetcdf and mpich
+        using modules from http://www.admin-magazine.com/HPC/Articles/Environment-Modules         
+    </DESC>
+    <NODENAME_REGEX>regex.expression.matching.your.machine</NODENAME_REGEX>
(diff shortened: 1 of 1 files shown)


billsacks:
At a glance, I'm happy with this. But I have no way of testing it. Should it be reviewed by someone with a linux machine who can test it?

jedwards4b:
I don't think there is any point unless you want someone to start from an OS install and do the whole thing.

billsacks:
Have you tested this yourself? I'd be a little wary of adding something like this if it hasn't undergone any testing, since it might give the false impression that we expect it to wor.

billsacks:
Though I'm willing to accept it if we at least add a comment saying that this hasn't been tested at all.

jedwards4b:
Yes - I have tested as documented.

billsacks:
The documentation just says you ran scripts_regression_tests - but as far as I know that wouldn't actually test this new stuff at all. What I meant was, have you actually tested using this port in any way?

jedwards4b:
I ran scripts_regression_tests on the centos7-linux system.

billsacks:
Ah, that makes me feel much better – I couldn't tell that from your initial PR comments. I'll merge this, then.

PR #1942, opened 2017-10-02
Load balancing tool mods

jedwards4b:
Automatically determine the layout to use if one is not offered on the command line.
If both ATM and OCN are prognostic but WAV is not use IceLndAtmOcn if
all of ATM, OCN, and WAV are prognostic use IceLndWavAtmOcn.

Test suite: scripts_regression_tests.py + load_balancing_test.py 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes
User interface changes?: 

Code review:


scripts/lib/CIME/case.py
```diff
@@ -462,6 +462,11 @@ def _set_compset(self, compset_name, files):
 
         return None, science_support
 
+    def get_primary_component(self):
+        if self._primary_component is None:
+            self._primary_component = self._find_primary_component()
```
jgfouca: Please remove 

```
         # This needs to be called after self.set_comp_classes, which is called
         # from self._get_component_config_data
         self._primary_component = self._find_primary_component()
```
... from configure method.

In _set_info_from_primary_component, change
```
component = self._primary_component
```

to 
```
component = self.get_primary_component()
```

tools/load_balancing_tool/README
```diff
@@ -9,352 +9,6 @@
...
-      $ ./load_balancing_solve.py ... --layout my_new_layout.MyNewLayout
-
-   -- to permanently add to CIME --
-
-   1. add MyNewLayout class to layouts.py
-   2. run using '--layout MyNewLayout'
-   3. add test in tests/load_balance_test.py that uses that name in command
-      line argument (see test for atm_lnd)
-   4. make pull request
+Complete documentation of this tool is available in the CIME Miscellaneous Tools section.
```
rljacob: Is there a --help message?   Also I think we should indicate the 2 modes in the README.  Something like this:
"This Load Balancing tool finds reasonable PE layouts for CIME-driven models.  It will find these from timing files you provide or from runs done by the tool.

For more information, see the help message or complete documentation in the Miscellaneous Tools section of the CIME user's guide."

jedwards4b: yes --help works

jedwards4b:
@rljacob your issue has been addressed.
@sarich can you review?

jedwards4b:
@jgfouca ready

PR #1949, opened 2017-10-04
Py3 port branch

### LINKED ISSUE #1047: Update to python 3.0 compliance
End of life for python 2.7 is 2020.   What needs to be done to update cime to python 3?

jedwards4b:
This is a port to python3 while maintaining compatibility with python 2.7.  The new module six.py provides portability between the two.   This PR deserves careful testing and discussion.

Test suite: scripts_regression_tests.py with both python2 and python3
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1047 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


billsacks:
@jedwards4b thanks a lot for taking this on!

Can you elaborate a bit on this?:

> This PR deserves careful testing and discussion.

From a 2-minute glance, the changes look relatively straightforward (though I'm sure it took a while to find all the places that needed to be changed!). Are there particular aspects you feel we should examine closely in our reviews?

rljacob:
Merging this PR will depend a lot on testing.  We'll have to run everything with python2.7 and python3 indefinitely.   Will developers have to develop in both?

jedwards4b:
@billsacks the changes are straightforward and I think that it's a solid PR  but there are a lot of files touched and I would feel better if it had more eyes than mine look at it. 
@rljacob I am setting up two platforms with python3 for cdash - I think we can develop in one or the other and let cdash help alert us of any compatibility issues.

billsacks:
Thanks for the clarification, @jedwards4b .

Regarding @rljacob 's question of testing: I think I'm comfortable trying @jedwards4b 's suggestion of having some cdash tests with both python2 and python3. If we find that problems are frequently making their way to master (because new code works in python2 but not python3, or vice versa), then I wonder if it would be feasible to have scripts_regression_tests run code_checker with both python2 and python3, so that developers' pre-PR testing covers both? (Would code_checker catch most problems in this respect?)

jedwards4b:
@billsacks I'm not sure that code_checker is sufficient - I had all of the code passing code_checker in both versions before I had all of the other testing working.

gold2718:
@jedwards4b, I read @billsacks suggestion as an addition to the cdash tests, not a replacement. Without some additional support, perhaps each PR will need to go through a testing cycle and cdash consulted before it can be merged and this could slow down our beloved frenetic development pace.

rljacob:
Time for a "next" branch!

jedwards4b:
I think that we should try the cdash route without a next branch and evaluate.

billsacks:
@gold2718 's interpretation of my suggestion is right. I'm willing to go with @jedwards4b 's suggestion of trying it out, though: If the frequency of breakage of master is no more often than, say, once every few months, then I can live with that for the simpler workflow.

That said: If having code_checker run with both python2 and python3 would catch, say, 90% of problems, then that could be a useful addition to the developer tests.

jedwards4b:
agreed - but how to implement that?  I think that we would need pylint2 and pylint3 wrapper scripts on each system or something like the second answer here https://stackoverflow.com/questions/23052637/specify-which-python-version-pylint-should-evaluate-for

billsacks:
On my mac, after doing a `pip3 install pylint` (which generated some permissions errors but I think generally worked), I'm able to do:

```
python -m pylint /path/to/file
python3 -m pylint /path/to/file
```

But I'm not sure if that's a generally applicable solution.

billsacks:
Never mind on my last comment... from some quick searches (e.g. https://github.com/vim-syntastic/syntastic/issues/1767), I have a feeling that's doing the wrong thing. I see suggestions for using a virtual environment. But this is all feeling like a lot of trouble, so I'm okay with trying for now with only having the developer tests run with a single version.

scripts/lib/CIME/XML/entry_id.py
```diff
@@ -360,7 +360,7 @@ def add_elements_by_group(self, srcobj, attributes=None, infile=None):
                 gname = "group_not_set"
             # If group with id=$gname does not exist in self.groups
             # then create the group node and add it to infile file
-            if gname not in self.groups.keys():
+            if gname not in list(self.groups.keys()):
```
billsacks: I don't think `list` is needed here, and I think you can just say `if gname not in self.groups:`

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -123,7 +123,7 @@ def create_job_groups(self, batch_jobs):
         for name, jdict in batch_jobs:
             new_job_group = ET.Element("group")
             new_job_group.set("id", name)
-            for field in jdict.keys():
+            for field in list(jdict.keys()):
```
billsacks: Remove `list`, possibly replace with simply `for field in jdict:`

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -359,7 +359,7 @@ def submit_jobs(self, case, no_batch=False, job=None, skip_pnl=False,
             if self._prereq_jobid is not None:
                 jobid = self._prereq_jobid
             for dep in deps:
-                if dep in depid.keys() and depid[dep] is not None:
+                if dep in list(depid.keys()) and depid[dep] is not None:
```
billsacks: Remove `list`, possibly replace with just `if dep in depid`

scripts/lib/CIME/XML/grids.py
```diff
@@ -533,7 +533,7 @@ def _print_values_v1(self, long_output=None):
                 logger.info("   short name: {}".format(sname))
             if alias is not None:
                 logger.info("   alias: {}".format(alias))
-            for attr, value in grid_node.items():
+            for attr, value in list(grid_node.items()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/XML/grids.py
```diff
@@ -582,7 +582,7 @@ def _get_all_values_v1(self):
                 "alias" : alias,
                 }
             compset_list = list()
-            for attr, value in grid_node.items():
+            for attr, value in list(grid_node.items()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/XML/testlist.py
```diff
@@ -115,7 +115,7 @@ def _get_testsv2(self, machine=None, category=None, compiler=None, compset=None,
             machnodes = self.get_nodes("machine",machatts,root=tnode)
             if machnodes:
                 this_test_node = {}
-                for key, value in tnode.items():
+                for key, value in list(tnode.items()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/XML/testlist.py
```diff
@@ -124,7 +124,7 @@ def _get_testsv2(self, machine=None, category=None, compiler=None, compset=None,
                 for mach in machnodes:
                     # this_test_node can include multiple tests
                     this_test = dict(this_test_node)
-                    for key, value in mach.items():
+                    for key, value in list(mach.items()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/build.py
```diff
@@ -387,7 +387,7 @@ def _case_build_impl(caseroot, case, sharedlib_only, model_only, buildlist):
     # Retrieve relevant case data
     # This environment variable gets set for cesm Make and
     # needs to be unset before building again.
-    if "MODEL" in os.environ.keys():
+    if "MODEL" in list(os.environ.keys()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/case.py
```diff
@@ -402,15 +403,15 @@ def set_valid_values(self, item, valid_values):
                 return result
 
     def set_lookup_value(self, item, value):
-        if item in self.lookups.keys() and self.lookups[item] is not None:
-            logger.warn("Item {} already in lookups with value {}".format(item,self.lookups[item]))
+        if item in list(self.lookups.keys()) and self.lookups[item] is not None:
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/case.py
```diff
@@ -402,15 +403,15 @@ def set_valid_values(self, item, valid_values):
...
+            logger.warning("Item {} already in lookups with value {}".format(item,self.lookups[item]))
         else:
             logger.debug("Setting in lookups: item {}, value {}".format(item,value))
             self.lookups[item] = value
 
     def clean_up_lookups(self, allow_undefined=False):
         # put anything in the lookups table into existing env objects
-        for key,value in self.lookups.items():
+        for key,value in list(self.lookups.items()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/case.py
```diff
@@ -694,7 +695,7 @@ def _setup_mach_pes(self, pecount, multi_driver, ninst, machine_name, mpilib):
         mach_pes_obj = self.get_env("mach_pes")
 
         if other is not None:
-            for key, value in other.items():
+            for key, value in list(other.items()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/case.py
```diff
@@ -733,10 +734,10 @@ def _setup_mach_pes(self, pecount, multi_driver, ninst, machine_name, mpilib):
             mach_pes_obj.set_value(key, ninst)
 
             key = "NTASKS_{}".format(compclass)
-            if key not in pes_ntasks.keys():
+            if key not in list(pes_ntasks.keys()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/case.py
```diff
@@ -733,10 +734,10 @@ def _setup_mach_pes(self, pecount, multi_driver, ninst, machine_name, mpilib):
...
                 mach_pes_obj.set_value(key,1)
             key = "NTHRDS_{}".format(compclass)
-            if compclass not in pes_nthrds.keys():
+            if compclass not in list(pes_nthrds.keys()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/case.py
```diff
@@ -768,7 +769,7 @@ def configure(self, compset_name, grid_name, machine_name=None,
         gridinfo = grids.get_grid_info(name=grid_name, compset=self._compsetname)
 
         self._gridname = gridinfo["GRID"]
-        for key,value in gridinfo.items():
+        for key,value in list(gridinfo.items()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/case.py
```diff
@@ -967,7 +968,7 @@ def _set_pio_xml(self):
         mpilib = self.get_value("MPILIB")
         defaults = pioobj.get_defaults(grid=grid,compset=compset,mach=mach,compiler=compiler, mpilib=mpilib)
 
-        for vid, value in defaults.items():
+        for vid, value in list(defaults.items()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/check_lockedfiles.py
```diff
@@ -112,7 +112,7 @@ def check_lockedfiles(case):
             diffs = f1obj.compare_xml(f2obj)
             if diffs:
                 logging.warn("File {} has been modified".format(lfile))
-                for key in diffs.keys():
+                for key in list(diffs.keys()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/check_lockedfiles.py
```diff
@@ -124,7 +124,7 @@ def check_lockedfiles(case):
                 elif objname == "env_build":
                     logging.warn("Setting build complete to False")
                     case.set_value("BUILD_COMPLETE", False)
-                    if "PIO_VERSION" in diffs.keys():
+                    if "PIO_VERSION" in list(diffs.keys()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/get_timing.py
```diff
@@ -145,7 +145,7 @@ def _getTiming(self, inst=0):
         else:
             pecost = totalpes
 
-        for m in self.models.values():
+        for m in list(self.models.values()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/get_timing.py
```diff
@@ -215,20 +215,20 @@ def _getTiming(self, inst=0):
         if inittype == "TRUE":
             odays = odays - (tlen/ocn_ncpl)
 
-        peminmax = max([m.rootpe for m in self.models.values()])+1
+        peminmax = max([m.rootpe for m in list(self.models.values())])+1
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/get_timing.py
```diff
@@ -215,20 +215,20 @@ def _getTiming(self, inst=0):
...
         if ncpl_base_period in ["decade","year","day"] and int(adays) > 0:
             adays = int(adays)
             if tlen % ocn_ncpl == 0:
                 odays = int(odays)
         self.adays = adays
         maxoffset = 40
         extraoff = 20
-        for m in self.models.values():
+        for m in list(self.models.values()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/get_timing.py
```diff
@@ -293,7 +293,7 @@ def _getTiming(self, inst=0):
         tmax = tmax + wtmin + correction
         ocn.tmax += ocnrunitime
 
-        for m in self.models.values():
+        for m in list(self.models.values()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/hist_utils.py
```diff
@@ -44,7 +44,7 @@ def _get_latest_hist_files(testcase, model, from_dir, suffix=""):
         ext = get_extension(model, hist)
         latest_files[ext] = hist
 
-    for key in latest_files.keys():
+    for key in list(latest_files.keys()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/namelist.py
```diff
@@ -899,9 +899,9 @@ def get_group_names(self):
         >>> Namelist().get_group_names()
         []
         >>> sorted(parse(text='&foo / &bar /').get_group_names())
-        [u'bar', u'foo']
+        ['bar', 'foo']
         """
-        return self._groups.keys()
+        return list(self._groups.keys())
```
billsacks: I'm wondering if the conversion to `list` here was done just to get the doctests to pass? If so, I'd say a better solution would be to convert the result to `list` in the doctests - keeping the production code returning an iterator in python3.

scripts/lib/CIME/namelist.py
```diff
@@ -912,38 +912,38 @@ def get_variable_names(self, group_name):
...
         >>> x = parse(text='&foo bar(::)=,bazz=false,bazz(2)=true,bazz(:2:)=6*""/')
         >>> sorted(x.get_variable_names('fOo'))
-        [u'bar(::)', u'bazz', u'bazz(2)', u'bazz(:2:)']
+        ['bar(::)', 'bazz', 'bazz(2)', 'bazz(:2:)']
         """
         group_name = group_name.lower()
         if group_name not in self._groups:
             return []
-        return self._groups[group_name].keys()
+        return list(self._groups[group_name].keys())
```
billsacks: I'm wondering if the conversion to `list` here was done just to get the doctests to pass? If so, I'd say a better solution would be to convert the result to `list` in the doctests - keeping the production code returning an iterator in python3.

scripts/lib/CIME/test_scheduler.py
```diff
@@ -643,7 +643,7 @@ def _get_procs_needed(self, test, phase, threads_in_flight=None, no_batch=False)
                 # Will force serialization of sharedlib builds
                 # TODO - instead of serializing, compute all library configs needed and build
                 # them all in parallel
-                for _, _, running_phase in threads_in_flight.values():
+                for _, _, running_phase in list(threads_in_flight.values()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/test_scheduler.py
```diff
@@ -774,7 +774,7 @@ def _producer(self):
                 # No free resources, wait for something in flight to finish
                 self._wait_for_something_to_finish(threads_in_flight)
 
-        for unfinished_thread, _, _ in threads_in_flight.values():
+        for unfinished_thread, _, _ in list(threads_in_flight.values()):
```
billsacks: Another unnecessary `list`?

scripts/lib/CIME/test_status.py
```diff
@@ -192,7 +192,7 @@ def set_status(self, phase, status, comments=""):
                 if subsequent_phase in self._phase_statuses:
                     del self._phase_statuses[subsequent_phase]
                 if subsequent_phase.startswith(COMPARE_PHASE):
-                    for stored_phase in self._phase_statuses.keys():
+                    for stored_phase in list(self._phase_statuses.keys()):
```
billsacks: Another unnecessary `list`?

mfdeakin-sandia:
Nice. I'm glad to see this was done sooner than I could get to it :+1:
I don't know much about six, but I notice sometimes you check that some variables are instances of six.string (utils.py), while other times you check that they are instances of str (compilers.py). Is there a preferred practice here?
All of the other changes look good to me, though I didn't look at what didn't change, so hopefully tests will catch anything else

scripts/Tools/cimeteststatus
```diff
@@ -354,10 +353,10 @@ def getSuiteInfo(specfile):
 
     for t in root.findall('test'):
         testlist.append(t.get('case'))
-        if 'machine' not in suiteinfo.keys():
+        if 'machine' not in list(suiteinfo.keys()):
```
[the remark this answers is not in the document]
mfdeakin-sandia: I'm reasonably certain that's correct

rljacob:
I see six.py was just added to scripts/lib.    Where is that developed?   Does it get updated a lot?  Does it come with tests?

jedwards4b:
six came from:
https://pypi.python.org/pypi/six

jedwards4b:
@mfdeakin-sandia I think that it should be six.string, using str will only cover a subset of possibilities.   I'll try to find these and fix.

jgfouca:
@jedwards4b did you use a conversion script to do this?

jedwards4b:
yes - that explains all of the list()s

scripts/lib/update_acme_tests.py
```diff
@@ -355,11 +355,15 @@ def sort_by_time(test_one, test_two):
...
-        return cmp(test_one, test_two)
+        return (test_one > test_two) - (test_two < test_one)
     else:
         if rec2 is None:
             return -1
         elif rec1 is None:
             return 1
         else:
-            return cmp(convert_to_seconds(rec2), convert_to_seconds(rec1))
+            a = convert_to_seconds(rec2)
```
jgfouca: This change confuses me.

jedwards4b: cmp was removed in python3, this function is equivalent.
https://stackoverflow.com/questions/15556813/python-why-cmp-is-useful

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -359,7 +359,7 @@ def submit_jobs(self, case, no_batch=False, job=None, skip_pnl=False,
             if self._prereq_jobid is not None:
                 jobid = self._prereq_jobid
             for dep in deps:
-                if dep in list(depid.keys()) and depid[dep] is not None:
+                if dep in depid:
```
billsacks: did you mean to drop the `depid[dep] is not None` condition?

billsacks:
Looks good to me now, thanks!

scripts/lib/update_acme_tests.py
billsacks: Would it be worth having our own utils.cmp function that does this, since I agree with @jgfouca that this code is not intuitive?

scripts/tests/list_tests
```diff
@@ -1,26 +1,59 @@
 #!/usr/bin/env python
-
+# This script will print the list of test classes in
```
jedwards4b: I haven't figured out how to make discover work in python3,  as a consequence the list of tests is hardcoded here - the list is confirmed when testing is done with python2 so any problems should be caught fairly quickly.  But the A_RunUnitTests also uses discover and are not working in python3.   The problem with discover in list_tests is only encountered with cmake/ctest but the A_RunUnitTests problem happens even when scripts_regression_tests.py is run directly (tests are not run but report as passing).

billsacks:
@jedwards4b I am trying to reproduce this problem:

> But the A_RunUnitTests also uses discover and are not working in python3.

I ran into a number of other issues when trying to get this to work, which I'll describe in a separate comment. So in order to test it, I introduced a new file in , `temporary.py`, with the following contents:

```python
#!/usr/bin/env python
import unittest
import logging
import os
import sys

LIB_DIR = os.path.join(os.path.dirname(os.path.abspath(__file__)),"..","lib")
sys.path.append(LIB_DIR)

class A_RunUnitTests(unittest.TestCase):
    def test_unittests(self):
        # Finds all files contained in CIME/tests or its subdirectories that
        # match the pattern 'test*.py', and runs the unit tests found there
        # (i.e., tests defined using python's unittest module).
        #
        # This is analogous to running:
        #     python -m unittest discover -s CIME/tests -t .
        # from cime/scripts/lib
        #
        # Yes, that means we have a bunch of unit tests run from this one unit
        # test.

        testsuite = unittest.defaultTestLoader.discover(
            start_dir = os.path.join(LIB_DIR,"CIME","tests"),
            pattern = 'test*.py',
            top_level_dir = LIB_DIR)

        testrunner = unittest.TextTestRunner(buffer=False)

        # Disable logging; otherwise log messages written by code under test
        # clutter the unit test output
        log_lvl = logging.getLogger().getEffectiveLevel()
        logging.disable(logging.CRITICAL)
        try:
            results = testrunner.run(testsuite)
        finally:
            logging.getLogger().setLevel(log_lvl)

        self.assertTrue(results.wasSuccessful())

if __name__ == "__main__":
    unittest.main(verbosity=2, catchbreak=True)
```

Then, on my mac, with python3.5.1, I ran: `python3 ./temporary.py`. This seemed to run all of the unit tests, which I confirmed by introducing an error:

```diff
diff --git a/scripts/lib/CIME/tests/SystemTests/test_system_tests_compare_two.py b/scripts/lib/CIME/tests/SystemTests/test_system_tests_compare_two.py
index 342fe4ecd..08ab93d21 100644
--- a/scripts/lib/CIME/tests/SystemTests/test_system_tests_compare_two.py
+++ b/scripts/lib/CIME/tests/SystemTests/test_system_tests_compare_two.py
@@ -491,6 +491,7 @@ def test_run1_fails(self):
         # Verify
         self.assertEqual(test_status.TEST_FAIL_STATUS,
                          mytest._test_status.get_status(test_status.RUN_PHASE))
+        self.assertEqual(2,3)

     def test_run2_fails(self):
         # Make sure that a failure in run2 is reported correctly
```

This error caused a failure, as expected.

So I wonder if the problem you're having is specific to particular versions of python3?

billsacks:
I'll also note - though I'm not sure if there's anything we can do about it: I did find that I got a lot more output with python3 than python2 when running the unit tests:

python2:

```
$ python -m unittest discover -s CIME/tests -t .
No handlers could be found for logger "CIME.SystemTests.system_tests_common"
.....................
Ran 21 tests in 0.279s

OK
```

python3:

```
$ python3 -m unittest discover -s CIME/tests -t .
Exception during run:
caseroot not in run_pass_caseroot
  File "/Users/sacks/cime/scripts/lib/CIME/SystemTests/system_tests_common.py", line 148, in run
    self.run_phase()
  File "/Users/sacks/cime/scripts/lib/CIME/SystemTests/system_tests_compare_two.py", line 216, in run_phase
    self.run_indv(suffix = self._run_one_suffix)
  File "/Users/sacks/cime/scripts/lib/CIME/tests/SystemTests/test_system_tests_compare_two.py", line 176, in run_indv
    raise RuntimeError('caseroot not in run_pass_caseroot')
RuntimeError: caseroot not in run_pass_caseroot

.Exception during run:
caseroot not in run_pass_caseroot
  File "/Users/sacks/cime/scripts/lib/CIME/SystemTests/system_tests_common.py", line 148, in run
    self.run_phase()
  File "/Users/sacks/cime/scripts/lib/CIME/SystemTests/system_tests_compare_two.py", line 234, in run_phase
    self.run_indv(suffix = self._run_two_suffix)
  File "/Users/sacks/cime/scripts/lib/CIME/tests/SystemTests/test_system_tests_compare_two.py", line 176, in run_indv
    raise RuntimeError('caseroot not in run_pass_caseroot')
RuntimeError: caseroot not in run_pass_caseroot

.......WARNING: Test case setup failed. Case2 has been removed, but the main case may be in an inconsistent state. If you want to rerun this test, you should create a new test rather than trying to rerun this one.
.........WARNING: Overwriting existing SourceMods in /var/folders/qt/mpzbvd5n2nn9x9g_jyv99344z_kvxg/T/tmpxo9qu5ws/SourceMods/src.drv/myfile.F90
WARNING: Overwriting existing SourceMods in /var/folders/qt/mpzbvd5n2nn9x9g_jyv99344z_kvxg/T/tmpxo9qu5ws/SourceMods/src.drv/myfile.F90
WARNING: Overwriting existing SourceMods in /var/folders/qt/mpzbvd5n2nn9x9g_jyv99344z_kvxg/T/tmpxo9qu5ws/SourceMods/src.drv/myfile.F90
.WARNING: Overwriting existing SourceMods in /var/folders/qt/mpzbvd5n2nn9x9g_jyv99344z_kvxg/T/tmp0_rf82o1/SourceMods/src.drv/myfile.F90
..WARNING: Overwriting existing SourceMods in /var/folders/qt/mpzbvd5n2nn9x9g_jyv99344z_kvxg/T/tmphbt64nk4/SourceMods/src.drv/myfile.F90
.
Ran 21 tests in 0.237s

OK
```

(The exceptions and warnings are all things that are expected to arise as part of various unit tests.)

jedwards4b:
@billsacks I have fixed all of the unittest issues for python3.

sarich:
Checking the python version needs to be changed to allow python3, or this will fail:

(in standard_script_setup.py)
`CIME.utils.check_minimum_python_version(2, 7)
`
```
def check_minimum_python_version(major, minor):
    expect(sys.version_info[0] == major and sys.version_info[1] >= minor,
           "Python {:d}, minor version {:d}+ is required, you have {:d}.{:d}".format(major, minor, sys.version_info[0], sys.version_info[1]))
```



I also got this error running scripts_regression_tests.py
```
> (2) In scripts_regression_tests.py: TypeError: cannot use a string pattern on a bytes-like object
> 
> Traceback (most recent call last):
>   File "./scripts_regression_tests.py", line 2449, in <module>
>     _main_func()
>   File "./scripts_regression_tests.py", line 2426, in _main_func
>     if check_for_pylint():
>   File "./scripts_regression_tests.py", line 2337, in check_for_pylint
>     pylintver = re.search(r"pylint\s+(\d+)[.](\d+)[.](\d+)", output)
>   File "/usr/local/Cellar/python3/3.5.1/Frameworks/Python.framework/Versions/3.5/lib/python3.5/re.py", line 173, in search
>     return _compile(pattern, flags).search(string)
> TypeError: cannot use a string pattern on a bytes-like object
```

subprocess communication can return bytes-like objects and not just strings, so they may need to be decoded before getting returned. Possible fix in CIME.utils.run_cmd():
```
    output, errput = proc.communicate(input_str)
    try:
        output = output.strip()
        output = output.decode("utf-8")
    except AttributeError:
        pass
    try:
        errput = errput.strip()
        errput = errput.decode("utf-8")
    except AttributeError:
        pass
```

I had trouble also with is_python_executable when it tried to check .jpg files:
```
  File "./scripts_regression_tests.py", line 2450, in <module>
    _main_func()
  File "./scripts_regression_tests.py", line 2428, in _main_func
    files_to_test = get_all_checkable_files()
  File "/home/sarich/working/cime/scripts/tests/../lib/CIME/code_checker.py", line 72, in get_all_checkable_files
    files_to_test = [item for item in all_git_files
  File "/home/sarich/working/cime/scripts/tests/../lib/CIME/code_checker.py", line 73, in <listcomp>
    if ((item.endswith(".py") or is_python_executable(os.path.join(cimeroot, item))) and not _should_pylint_skip(item))]
  File "/home/sarich/working/cime/scripts/tests/../lib/CIME/utils.py", line 1309, in is_python_executable
    raise e
  File "/home/sarich/working/cime/scripts/tests/../lib/CIME/utils.py", line 1306, in is_python_executable
    first_line = f.readline()
  File "/usr/local/lib/python3.6/codecs.py", line 321, in decode
    (result, consumed) = self._buffer_decode(data, self.errors, final)
UnicodeDecodeError: 'utf-8' codec can't decode byte 0xff in position 0: invalid start byte
```
Not sure what to do here, perhaps filter out obvious binary files by their endings and/or return False on exceptions


Escaping the errors above, scripts_regression_test.py is at least running for me, though with many failures and errors using python 3.4.3 
[test.log](https://github.com/ESMCI/cime/files/1360384/test.log)

billsacks:
This works for me now - thanks!

sarich:
Many of the errors I got were due to using a string for the version in the MockMachines constructor in scripts_regression_tests.py. 
       maker = Compilers(MockMachines("mymachine", "SomeOS"), version="2.0")

Using 2.0 instead of "2.0" fixes that

Also, six.assertRaisesRegex() is only available in six 1.11 and higher. Perhaps it should be documented somewhere.

jedwards4b:
@sarich thanks for the feedback.  I think I've addressed these issues.

jgfouca:
Waiting on approval from @rljacob and @mvertens

jedwards4b:
I would like to rebase this branch - any objections?

billsacks:
@jedwards4b no objections to rebasing here - thanks a lot for checking

billsacks:
@jedwards4b I have a sense that you've made a lot of changes since the initial reviews. Do you feel this needs a re-review at this point, or were the additional changes fairly minor and straightforward?

jedwards4b:
@billsacks Additional changes have pretty much been more of the same - Changes to exception handling and data type conversions.   I would say they were minor and straightforward, but I would also welcome additional review as I think that there are probably still a few problems I've missed.

billsacks:
> Additional changes have pretty much been more of the same - Changes to exception handling and data type conversions. I would say they were minor and straightforward, but I would also welcome additional review as I think that there are probably still a few problems I've missed.

Okay, thanks. I have enough of a backlog right now that I'm inclined to leave the final review to one of the other reviewers in that case - but I can try to look at this if necessary.

rljacob:
Jim, can you rewrite history on this branch to reduce the number of commits?  All those "fixes" ones can be collapsed.

jedwards4b:
That may be beyond my git ability.

jedwards4b:
It seems that I can do this in the process of merging to master - if you are ready I propose to do:
```
git checkout master
git merge --squash py3_port_branch
git commit -m'backward compatible port to python3 using six.py'
git push origin/master
```

rljacob:
Yes that's fine.  It will collapse everything to a single commit I think.

billsacks:
@jedwards4b if you do this the way you suggest, can you please:

1. Maintain the original PR comment in the log message (as we typically do)

2. Reference this PR number somewhere in the commit message (as is typically done when merging PRs)

jedwards4b:
Merged in commit `c5cf2af`  (the squashed commit was not recognized by github)

PR #1950, opened 2017-10-04
refactor CPLHIST mode and add DATM CPLHIST topo capability 

### LINKED ISSUE #1407: For datm cplhist cases, get topo from cplhist file
Currently, datm always gets the topo stream (atmosphere topographic height) from a particular topo dataset. However, for cplhist cases, we really want to get the topo field from the cplhist file.

Changes to write the topo field to the cplhist file are already in place. So this requires extending the logic in datm's namelist / streams file generation to get the topo field from the correct place.

I think this is related to https://github.com/CESM-Development/cime/issues/99

One tricky part of this is how to support backwards compatibility for old cplhist files. When we talked about it in Nov, 2015, @ekluzek suggested: We could have an xml variable that says where to look for topo when running cpl hist. By default it will look on the cplhist files themselves. But the user can change this xml variable to make it look on the static file that is used for other datm cases.

### LINKED ISSUE #1877: Review mode implementation in data models
The data models need to review implementation of the different mode support.  It looks like in the latest refactor, only datm, desp, dice, and docn have a "select case for the mode".   Different modes should be supported in all data models, at least for NULL and not NULL to provide clearer paths for extensions.

mvertens:
This PR has several features:
1) unifies the CPLHIST naming convention across the components by
   adding the following new xml variables in DATM, DLND and DOCN
      `DATM_CPLHIST_DOMAIN_FILE, DLND_CPLHIST_DOMAIN_FILE, DROF_CPLHIST_DOMAIN_FILE`
   These new xml variables are the full pathname for domain file
   for datm, dlnd and drof when the corresponding data model mode
   is set to CPLHIST.
   NOTE: if any of these xml variables set to 'null',
   then domain information is read in from the first coupler
   history file in the target stream.
   NOTE: it is assumed that when `DXXX_CPLHIST_DOMAIN_FILE='null'`
   that the first coupler stream file that is pointed
   to contains the domain information for that stream.
   NOTE: This is the default mode that should be used when the data model
   `DXXX_MODE is CPLHIST`.
2) puts in new capability for DATM CPLHIST mode to read in topo data
3) replaces % formatting in the data models with new .format()
4) address issue #1877
5) replaced `xxx_mode (e.g. atm_mode) `variables in dxxx_comp_mod.F90 and dxxx_shr_mod.F90 with `datamode `- to reduce confusion.

Additional testing:
Verified that using this CIME branch to replace cime/ in cesm2_0_alpha07d,
the following cases ran successfully on cheyenne with data that was already there:

```
1850_DATM%CPLHIST_CLM50%BGC_SICE_SOCN_MOSART_CISM2%NOEVOLVE_SWAV (f09_f09_mg17)
   the following xmlchanges neededed to be made:
   ./xmlchange DATM_CPLHIST_DOMAIN_FILE=null
   ./xmlchange DATM_CPLHIST_DIR=/glade/p/cesm/bgcwg_dev/forcing/b.e20.B1850.f09_g17.pi_control.all.179.cplhist/cpl/hist.mon
   ./xmlchange DATM_CPLHIST_CASE=b.e20.B1850.f09_g17.pi_control.all.179.cplhist
   ./xmlchange DATM_CPLHIST_YR_ALIGN=1
   ./xmlchange DATM_CPLHIST_YR_START=37
   ./xmlchange DATM_CPLHIST_YR_END=37
```
```
1850_DATM%CPLHIST_SLND_CICE_POP2%ECO_DROF%CPLHIST_SGLC_WW3 (f09_g17)
   the following xmlchange commands needed to be run:
   ./xmlchange DATM_CPLHIST_DOMAIN_FILE=null
   ./xmlchange DATM_CPLHIST_DIR=/glade/p/cesm/bgcwg_dev/forcing/b.e20.B1850.f09_g17.pi_control.all.179.cplhist/cpl/hist.mon
   ./xmlchange DATM_CPLHIST_CASE=b.e20.B1850.f09_g17.pi_control.all.179.cplhist
   ./xmlchange DATM_CPLHIST_YR_ALIGN=1
   ./xmlchange DATM_CPLHIST_YR_START=37
   ./xmlchange DATM_CPLHIST_YR_END=37
   ./xmlchange DROF_CPLHIST_DOMAIN_FILE=null
   ./xmlchange DROF_CPLHIST_DIR=/glade/p/cesm/bgcwg_dev/forcing/b.e20.B1850.f09_g17.pi_control.all.179.cplhist/cpl/hist.mon
   ./xmlchange DROF_CPLHIST_CASE=b.e20.B1850.f09_g17.pi_control.all.179.cplhist
   ./xmlchange DROF_CPLHIST_YR_ALIGN=1
   ./xmlchange DROF_CPLHIST_YR_START=37
   ./xmlchange DROF_CPLHIST_YR_END=37

```
```
1850_SATM_DLND%SCPL_SICE_SOCN_SROF_CISM2%EVOLVE_SWAV (f09_g16_gl20)
    no xmlchange commands were needed since in this case we have
      DLND_CPLHIST_DOMAIN_FILE" value="$LND_DOMAIN_PATH/$LND_DOMAIN_FILE"
    which is obtained from the config_grids.xml file

```
Test baseline: did not compare against baselines
Test namelist changes: None
Test status: should be bfb

Fixes #1877 
Fixes #1407 

User interface changes?: added new xml variables discussed above
Code review:


doc/source/data_models/data-atm.rst
```diff
@@ -24,42 +24,47 @@ These variables will appear in ``env_run.xml`` and the resulting values are comp
...
+   "DATM_CPLHIST_YR_END",      "Coupler history forcing data model ending year to loop data over"
+
+   "DATM_CLMNCEP_YR_ALIGN",    "I compsets only - simulation year corresponding to data starting year"
+   "DATM_CLMNCEP_YR_START",    "I compsets only - data model starting year to loop data over"
+   "DATM_CLMNCEP_YR_END",      "I compsets only - data model ending year to loop data over"
+
+.. note:: If ``DATM_MODE`` is set to ``CPLHIST``, it is normally assumed that the model domain will be identical to **all** of the stream domains. To ensure this, the xml variables ``ATM_DOMAIN_PATH`` and ``ATM_DOMAIN_FILE`` are ignored and a valid setting **must be given** for ``DATM_CPLHIST_DOMAIN_FILE``. If ``DATM_CPLHIST_DOMAIN_FILE`` is set to ``null``, then the datm component domain information is read in from the first coupler history file in the target stream and  it is assumed that the first coupler stream file that is pointed to contains the domain  information for that stream. This is the default mode that should be used for this mode. Alternatively, ``DATM_CPLHIST_DOMAIN_FILE`` can be set to ``$ATM_DOMAIN_PATH/$ATM_DOMAIN_FILE`` in a non-default configuration.
```
gold2718: In another discussion, it was thought that `'none'` is more user accessible than `'null'`. Is it possible to use `'none'` in this context?

gold2718:
Is #1877 'fixed' by this change? If so, please add it to the 'Fixes' section of the header.

billsacks:
#1407 is fixed by this, right? So it should be added to the "Fixes" section.

billsacks:
Minor documentation point: In your big initial comment, you have three blocks showing the tests you ran. The second is missing information on what compset that's referring to.

billsacks:
From your documentation in this PR, it looks like you need to explicitly set `./xmlchange DATM_CPLHIST_DOMAIN_FILE=null`? From our earlier discussions, I thought this was going to be the out-of-the-box default?

mvertens:
It is the out of the box default - but its always good to explicitly state
it for consistency of specifying what xmlchange commands are needed.

doc/source/data_models/data-atm.rst
mvertens: Unfortunately, this is not really possible in this case. A 'null' value is hard-wired as the default value in the data model shr strdata input namelist - and it would be good to be consistent with the code. This could be changed -but I'm really wary to do this now.

mvertens:
@billsacks -. It is the default but it helps to have it explicitly set as well. I have also addressed your comment as to the documentation.

mvertens:
@billsacks, @gold2718 - I have explicitly added the fixes in the PR.

billsacks:
@mvertens I tried setting up a case (out of alpha07d with new cime) with:

```
./create_newcase --case tg_new_1011 --compset T1850G --res f09_g16 --run-unsupported
cd tg_new_1011
./case.setup
./preview_namelists
```

This gave (among other things):

```
   Calling /glade2/scratch2/sacks/cesm_code/cesm2_0_alpha07d_newCime/cime/src/components/data_comps/dlnd/cime_config/buildnml
   ....  Obtaining DLND model domain info from stream sno.cplhist
   ....    lonname  must be xc
   ....    latname  must be yc
   ....    areaname must be area
   ....    fracname must be frac
```

Do these "must be" messages indicate problems? If not, I find them confusing, because this wording suggests to me that something is wrong.

mvertens:
These do not indicate problems. They need to let the user know that if you
use a stream file to obtain the domain and you are picking the stream file
- rather than just having the first stream file be picked - then you need
to have the following naming convention. I think  its important to point
this out - but maybe another wording would be useful.

src/components/data_comps/datm/cime_config/config_component.xml
```diff
@@ -111,61 +112,69 @@
...
     <valid_values></valid_values>
     <default_value>1</default_value>
     <values match="last">
-      <value compset="1850_DATM%CPLHIST">1</value>
+      <value compset="1850_DATM%CPLHIST">-999</value>
     </values>
     <group>run_component_datm</group>
     <file>env_run.xml</file>
-    <desc>Simulation year corresponding to starting year (only used for CPLHIST3HrWx mode)</desc>
+    <desc>Simulation year corresponding to starting year (only used when DATM_MODE is CPLHIST)</desc>
```
billsacks: Since the user now needs to set this manually, I think the documentation should be expanded. Here's my suggestion, but I'd like you to confirm this, because I'm not confident of my understanding of this variable:

```xml
    <desc>Simulation year corresponding to DATM_CPLHIST_YR_START. This will usually be the same as the year in RUN_STARTDATE, but it can be set to a different year to start the simulation with a different year of forcing data. (Only used when DATM_MODE is CPLHIST.)</desc>
```

Whatever documentation we add for this one, the same documentation should be added for DLND and DROF CPLHIST_YR_ALIGN.

src/components/data_comps/datm/cime_config/namelist_definition_datm.xml
```diff
@@ -1616,8 +1628,6 @@
       scientific modes supported by the model, they are listed below.  The
       mode is selected by a character string set in the strdata namelist
       variable dataMode.
-      datamode = "CPLHIST"
```
billsacks: I think CPLHIST should also be removed from the list of valid values for this variable; do you agree?

src/components/data_comps/dlnd/cime_config/config_component.xml
```diff
@@ -44,6 +44,28 @@
     set to zero and not utilized. The default is NULL.</desc>
   </entry>
 
+  <entry id="DLND_CPLHIST_DOMAIN_FILE">
+    <type>char</type>
+    <valid_values></valid_values>
+    <default_value>null</default_value>
+    <values match="last">
+      <value compset="2000.*_DLND%SCPL" grid="l%0.9x1.25">$LND_DOMAIN_PATH/$LND_DOMAIN_FILE</value>
```
billsacks: Why are there these duplicated entries for all the different time periods? Why not have a single entry with a compset match of "_DLND%SCPL"?

src/components/data_comps/dlnd/cime_config/config_component.xml
```diff
@@ -44,6 +44,28 @@
...
+      <value compset="20TR.*_DLND%SCPL" grid="l%0.9x1.25">$LND_DOMAIN_PATH/$LND_DOMAIN_FILE</value>
+    </values>
+    <group>run_component_dlnd</group>
+    <file>env_run.xml</file>
+    <desc>
+      Full pathname for domain file for dlnd when DLND mode is CPLHIST or GLC_CPLHIST.
+      NOTE: if this is set to 'null', then domain information is read in from the first
+      coupler history file in the target stream and it is assumed that the first coupler
+      stream file that is pointed to contains the domain information for that stream.
+      This is the default mode that should be used when DLND_MODE is CPLHIST or GLC_CPLHIST.
```
billsacks: This last sentence is wrong for DLND, I think.

src/components/data_comps/dlnd/cime_config/config_component.xml
```diff
@@ -76,7 +98,7 @@
 
   <entry id="DLND_CPLHIST_YR_ALIGN">
     <type>integer</type>
-    <default_value>1</default_value>
+    <default_value>-999</default_value>
```
billsacks: See above comment for DATM_CPLHIST_YR_ALIGN: whatever documentation is added for that one should also be added for DLND_CPLHIST_YR_ALIGN

src/components/data_comps/datm/cime_config/config_component.xml
```diff
@@ -17,7 +17,7 @@
     <desc option="CRUv7"> CLM CRU NCEP v7 data set </desc>
     <desc option="GSW"> GSWP3 data set </desc>
     <desc option="GSWP3v1"> GSWP3v1 data set </desc>
-    <desc option="CPLHIST"> Coupler hist data set </desc>
+    <desc option="CPLHIST"> Coupler hist data set (assumes that model grid and stream domains are the same)</desc>
```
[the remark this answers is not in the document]
mvertens: I think this is the default IF the domain file is set to null. In this case if you ask for a different model grid and you obtain the domain from the first stream file then you will have an inconsistent set of mapping files relative to the actual domain that is being used. In that case I think the model should abort - but I've never tried it.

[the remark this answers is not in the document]
billsacks: Thanks for the clarification. But I guess my point is: I don't think CPLHIST mode itself assumes that the model grid and stream domains are the same. Rather, the *default* if the domain file is set to null assumes that these two domains are the same.

I'm not quite sure if I'm right in my understanding there... and if I am right, I'm not quite sure how and where to document this... but I can think more about how to document this clearly if you want me to.

src/components/data_comps/datm/cime_config/config_component.xml
mvertens: I'm not sure this is right. Say you had an 1850 simulation and you were writing coupler history output for years 220-230. Now you wanted to use that output to spin-up the ocean. You now start from year 1 in the ocean - but cycle over years 220-230. I think it will be more confusing to add this documentation. Frankly, I think that cplhist mode is really geared to more expert users. We could give some examples in the documentation for clarity - but I think that would be better in the user's guide as opposed to the xml file.

src/components/data_comps/datm/cime_config/namelist_definition_datm.xml
mvertens: Yes! Good catch.

src/components/data_comps/dlnd/cime_config/config_component.xml
mvertens: Another good catch! You're right. I'll change that.

src/components/data_comps/dlnd/cime_config/config_component.xml
mvertens: Thanks for catching this as well. Actually this was wrong in DATM, DROF and DLND. I've changed the xml documentation in all three config_component.xml files.

src/components/data_comps/datm/cime_config/config_component.xml
billsacks: My understanding of the align year is that, in order to do what you're suggesting, you would set the align year to 1, which I think is consistent with my suggested wording. And I do still think that, for spinup runs, the typical operation would be to set the align year to the year in RUN_STARTDATE... and if that is correct (we could ask Keith L & Keith O to confirm), I do feel it would be helpful to add a brief sentence to that effect so that people don't have to dig all through the user's guide to see how to set this unintuitive value.

But I'll let that go if you still disagree :-) - in which case, I'm fine with putting additional documentation in the user's guide instead of the xml file, as long as it appears somewhere. Your idea of providing some examples of typical use cases sounds like a good one, regardless of whether this extra documentation appears in the xm file. (And I'm fine with that being deferred to a separate issue/pr - in that case we should open an issue for it.)

src/components/data_comps/datm/cime_config/config_component.xml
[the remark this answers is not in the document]
mvertens: Yes - I agree with your statement that the default if the domain file is set to null assumes that these two domains are the same. I think that one possible way to state this is that _CPLHIST mode is strongly recommended to be run only when the model domain and the couple history forcing are on the same domain_.  Thoughts?

mvertens:
I would run this by Keith L and Keith O to determine if how they use this.
If we can clarify this in one or two sentences then I am fine with bringing
this into config_component.xml. I just would rather not have a long
explanation.

src/components/data_comps/datm/cime_config/config_component.xml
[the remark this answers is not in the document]
billsacks: Yes, I like that wording.

PR #1965, opened 2017-10-16
Iscompvar cleanup

jedwards4b:
This is a cleanup of the compvar implementation.   There was a confusion between component attributes which represented components (cam, datm, etc) vs those which represented component classes (ATM, LND, etc).  I've renamed the attributes associated with component classes to be compclass instead of component.   This change also allowed me to get rid of the explicit list of compvars in env_mach_pes.py and env_run.py

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes:

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/utils.py
+++ b/scripts/lib/CIME/utils.py
@@ -1222,7 +1222,6 @@ def find_system_test(testname, case):
 
             if tdir is not None:
                 tdir = os.path.abspath(tdir)
-                print (tdir)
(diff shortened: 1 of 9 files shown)


scripts/lib/CIME/utils.py
```diff
@@ -1222,6 +1222,7 @@ def find_system_test(testname, case):
 
             if tdir is not None:
                 tdir = os.path.abspath(tdir)
+                print (tdir)
```
jgfouca: Remove.

PR #1981, opened 2017-10-21
Resolve compset issue

jedwards4b:
A subtle bug fix - the discovery of compsets needs to have information about available components so that it can look in the correct cime_config directories for config_compsets.xml files. 

Test suite: scripts_regression_tests.py + cesm F cases
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/entry_id.py
+++ b/scripts/lib/CIME/XML/entry_id.py
@@ -422,6 +422,16 @@ class EntryID(GenericXML):
                                                 xmldiffs["{}:{}".format(vid, valnode.attrib)] = [valnode.text, f2valnode.text]
         return xmldiffs
 
+    def overwrite_existing_entries(self):
+        # if there
+        for node in self.get_nodes("entry"):
+            vid = node.get("id")
+            samenodes = self.get_nodes_by_id(vid)
+            if len(samenodes) > 1:
+                expect(len(samenodes) == 2, "Too many matchs for id {} in file {}".format(vid, self.filename))
+                logger.debug("Overwriting node {}".format(vid))
+                self.root.remove(samenodes[0])
+
     def __iter__(self):
         for node in self.get_nodes("entry"):
             vid = node.get("id")
--- a/scripts/lib/CIME/XML/entry_id.py
+++ b/scripts/lib/CIME/XML/entry_id.py
@@ -423,7 +423,7 @@ class EntryID(GenericXML):
         return xmldiffs
 
     def overwrite_existing_entries(self):
-        # if there
+        # if there exist two nodes with the same id delete the first one.
         for node in self.get_nodes("entry"):
             vid = node.get("id")
             samenodes = self.get_nodes_by_id(vid)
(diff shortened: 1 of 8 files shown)


jedwards4b:
@mvertens I have rerun testing on this branch.  All pass

scripts/lib/CIME/XML/files.py
```diff
@@ -23,6 +24,29 @@ def __init__(self):
         expect(os.path.isfile(infile), "Could not find or open file {}".format(infile))
         schema = os.path.join(cimeroot, "config", "xml_schemas", "entry_id.xsd")
         EntryID.__init__(self, infile, schema=schema)
+        config_files_override = os.path.join(os.path.dirname(cimeroot),".config_files.xml")
+        # .config_file.xml at the top level may overwrite COMP_ROOT_DIR_ nodes in config_files
+        if os.path.isfile(config_files_override):
+            self.read(config_files_override)
+            for vid in ("COMP_ROOT_DIR_LND", "COMP_ROOT_DIR_ATM", "COMP_ROOT_DIR_ICE",
```
mvertens: It would be better not to hardwire the _XXX names such as _LND.

scripts/lib/CIME/case_st_archive.py
```diff
@@ -180,6 +180,8 @@ def _archive_history_files(case, archive, archive_entry,
     if not os.path.exists(archive_histdir):
         os.makedirs(archive_histdir)
         logger.debug("created directory {}".format(archive_histdir))
+    if compname == 'drv':
```
mvertens: Please add a comment as to why you are doing this.

scripts/lib/CIME/XML/entry_id.py
```diff
@@ -422,6 +422,16 @@ def compare_xml(self, other):
...
 
+    def overwrite_existing_entries(self):
+        # if there
+        for node in self.get_nodes("entry"):
+            vid = node.get("id")
+            samenodes = self.get_nodes_by_id(vid)
+            if len(samenodes) > 1:
+                expect(len(samenodes) == 2, "Too many matchs for id {} in file {}".format(vid, self.filename))
+                logger.debug("Overwriting node {}".format(vid))
+                self.root.remove(samenodes[0])
```
mvertens: Is it always the case that the new node will be after the original node - this is assumed in that your are removing samenodes[0] - the first node. Also - could you please document what this routine is doing in terms of the new functionality we are building in (you are taking the entry from the .config_files in $SRCROOT and overwriting the entry here with that one).

scripts/lib/CIME/case_st_archive.py
```diff
@@ -180,6 +180,7 @@ def _archive_history_files(case, archive, archive_entry,
     if not os.path.exists(archive_histdir):
         os.makedirs(archive_histdir)
         logger.debug("created directory {}".format(archive_histdir))
+    # the compname is drv but the files are named cpl
```
mvertens: Could you please give an example of what files are named cpl currently so that we can have this documented.

jedwards4b:
@jgfouca this is ready

PR #1987, opened 2017-10-24
Fix several data component issues

### LINKED ISSUE #1802: Discussion on new organization of shr_stream_mod and related code
In #1767 , @mvertens moved shr_stream_mod and related code out of `src/share/util` into `src/components/data_comps/dshare`. There was some discussion of this move in the comments of #1767 (https://github.com/ESMCI/cime/pull/1767#issuecomment-318493072 and then later comments by @mvertens @jedwards4b and @rljacob ), but I'm not sure that my points were understood correctly. Since that PR has been merged, I wanted to continue the discussion in a new issue. To be clear: I am perfectly happy to be overruled on this point, but I just want to make sure that my points come across clearly before this decision is finalized.

Note that @mvertens set the build up so that `src/components/data_comps/dshare` is compiled into the share library, similarly to what is done for `src/drivers/mct/shr`. 

I have two related concerns with the new organization:

1. For my first few years of working with CESM, I found it confusing that the share library was built from code from two completely different places: the share directory itself as well as the shr subdirectory of the drv code. This move introduces a third location included in the share library. It would be easier to understand how the source code is built into libraries – and thus, crucially, what modules are available to components using cime – if the organization into libraries more closely matched the source code directory structure.

2. For someone not intimately familiar with the cime source code, a logical place to look for modules that can be utilized by component models is `src/share`. This will now fail to turn up the stream-related code - which can be a very useful set of utilities for active components as well as data components. Putting this code under `src/components/data_comps/dshare` sends the message that this code is only meant for use by data models, which is not at all the case. I am concerned that the new organization will therefore cause component model developers to overlook this code, or at the very least have a hard time finding it.

Here are the uses of the 4 modules included in this new `dshare` directory, in CESM:

• shr_strdata_mod: This is the key one: it is used extensively by CLM, as well as by POP and CICE:

  ```
  ./cime/src/components/data_comps/datm/datm_comp_mod.F90
  ./cime/src/components/data_comps/desp/desp_comp_mod.F90
  ./cime/src/components/data_comps/dice/dice_comp_mod.F90
  ./cime/src/components/data_comps/dlnd/dlnd_comp_mod.F90
  ./cime/src/components/data_comps/docn/docn_comp_mod.F90
  ./cime/src/components/data_comps/drof/drof_comp_mod.F90
  ./cime/src/components/data_comps/dwav/dwav_comp_mod.F90
  ./cime/src/share/util/shr_scam_mod.F90
  ./components/cice/src/drivers/cesm/ice_prescribed_mod.F90
  ./components/clm/src/biogeochem/ch4FInundatedStreamType.F90
  ./components/clm/src/biogeochem/CNFireBaseMod.F90
  ./components/clm/src/biogeochem/SatellitePhenologyMod.F90
  ./components/clm/src/biogeophys/UrbanTimeVarType.F90
  ./components/clm/src/main/ndepStreamMod.F90
  ./components/clm/src_clm40/main/ndepStreamMod.F90
  ./components/pop/source/strdata_interface_mod.F90
  ```

• shr_tInterp_mod: only used by shr_strdata_mod.F90 (so this should go in whatever directory shr_strdata_mod.F90 goes in)

• shr_stream_mod: In addition to shr_dmodel_mod.F90 and shr_strdata_mod.F90, this is used directly by two modules in CLM

• shr_dmodel_mod: in addition to the data models, this is used by:
  - cime/src/share/util/shr_scam_mod.F90
  - cime/src/share/util/shr_strdata_mod.F90
    - This use suggests to me that some reorganization of these modules may be needed: active components use shr_strdata_mod, which uses shr_dmodel_mod, which doesn't make sense conceptually
  - components/cice/src/drivers/cesm/ice_prescribed_mod.F90

### LINKED ISSUE #1937: Case created even with no domain file
This sample `create_newcase` call:
```
./create_newcase --case QPC5 --compset QPC5 --res ne30_ne30_mg17 --compiler nag --run-unsupported
```
builds with no issues even though there is no matching domain file for the underlying grid file in config_grids.xml. It crashes at run time because the domain file XML variables (e.g., `OCN_DOMAIN_FILE`) are `UNSET`.
In this case, there is no domain entry for `ne30np4` with a `mask="gx1v7"` (there is one for `mask="gx1v6"`).

### LINKED ISSUE #1960: Q compset DOCN streams not zeroing land mask with SE dycore
https://github.com/CESM-Development/cime/tags/cime5.4.0-alpha.03
https://github.com/CESM-Development/cime_config/tags/cime_config0.1.0.alpha.02/
https://svn-ccsm-models.cgd.ucar.edu/cam1/branches/physgrid/components/cam

./create_newcase --case ~/work/QPC6/QPC6.ne30 --compset QPC6 --res ne30_ne30_mg16 --project XXXXXXXXX --run-unsupported

This configuration works as expected if one of the analytic SST profiles is chosen for aquaplanet simulations.

However, if I try to read a data stream... add user_docn.streams.txt.aquapfile (NCAR glade example at /glade/p/work/zarzycki/initial/SST/user_docn.streams.txt.aquapfile) and set DOCN_MODE to "sst_aquapfile"
... CAM happily runs and loads the correct SST stream file but does NOT zero out the land mask. I have reproduced this with both a user-generated variable-resolution SE grid/domain file and one of the default grids/domain files in the repo (ne30)

This user_docn.streams seems to behave correctly with FV, so it appears this problem is only with A) the SE core and then B) only when a data stream file is being read for DOCN. Perhaps an issue with the unstructured grid and zeroing out the mask?

mvertens:
This PR fixes several outstanding issues:

1) src/components/data_comps/dshare was moved to src/share/streams in order
to resolved issue #1802.

2) For aquaplanet runs - the landmask was reset in docn_comp_mod.F90
and in fact this does work if you are reading an input sst file as the
aquaplanet forcing.  The right place to put this is in
`shr_strdata_init.F90 `right after the model grid is read in. An
optional argument was introduced in `shr_strdata_init.F90`
(reset_domain_mask) that will allow this in a backwards compatible manner.
This resolves issue #1960.

3) If the data component domain file is equal to `'UNSET'` the `'domainfile'` namelist
variable should be set to` 'null' `which will assume that the model domain is contained
in the first stream file. This resolves issue #1937.

  The following extra tests were also run on cheyenne and compared to cesm2_0_alpha07f:
   ```
ERI.T62_g16.C1850ECO.cheyenne_intel.pop-ecosys
   ERI.T62_g37.G.cheyenne_intel.pop-cice
   ERP_D_Ln9.f19_f19_mg17.QPC6.cheyenne_intel.cam-outfrq9s
   ERP_D_Ln9.f19_f19_mg17.QSC6.cheyenne_intel.cam-outfrq9s
   ERP_Ln9.f09_f09_mg17.F1850_DONOTUSE.cheyenne_intel.cam-outfrq9s
   ERP_P180x2_D_Ld5.f19_g17_gl4.I1850Clm50BgcCropG.cheyenne_intel.clm-default
   ERR.f45_g37_rx1.A.cheyenne_intel
   ERS_IOP.T62_g16.CIAF.cheyenne_intel.pop-default
   ERS_IOP.T62_g16.GIAF.cheyenne_intel.pop-default
   ERS_Lm3.T62_g16.AIAF.cheyenne_intel
   ERS_Ly3.f09_g16_gl4.T1850G.cheyenne_intel
```
Test baseline: cesm2_0_alpha07b
Test namelist changes: none
Test status: bit for bit

Fixes #1802 
Fixes #1960 
Fixes #1937

User interface changes?: None
Update gh-pages html (Y/N)?:N
Code review:


apcraig:
It looks like the dependency on the driver share code was not removed.  I think that is important.

mvertens:
This belongs in a separate PR. This PR only related to the data model share
code. Please raise an issue for addressing the driver share code and what
routines need to be moved.

mvertens:
We also need to be careful that driver share code that is different between
nuopc and cesm is handled correctly.

apcraig:
I am not raising a driver share code issue, I am raising an issue that is directly related to the shr stream code.  It currently depends on driver share code.  That dependency should be removed.

mvertens:
You are right. That said - I would like to simply put this in another PR.
Lots of other things were also fixed in this PR.

PR #1990, opened 2017-10-24
Fix COST_PES and TOTALPES for titan (aprun)

jgfouca:
COST_PES should not take spare nodes into consideration.

Most fields that get set in case.initialize_derived_attributes are impacted
by aprun.

Fix string search/replace that capitalized regular stack variables by mistake.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b @worleyph

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -129,18 +129,9 @@ class Case(object):
         env_mach_pes  = self.get_env("mach_pes")
         env_mach_spec = self.get_env('mach_specific')
         comp_classes  = self.get_values("COMP_CLASSES")
-        MAX_MPITASKS_PER_NODE  = self.get_value("MAX_MPITASKS_PER_NODE")
-        self.total_tasks = env_mach_pes.get_total_tasks(comp_classes)
+        max_mpitasks_per_node  = self.get_value("MAX_MPITASKS_PER_NODE")
+
         self.thread_count = env_mach_pes.get_max_thread_count(comp_classes)
-        self.tasks_per_node = env_mach_pes.get_tasks_per_node(self.total_tasks, self.thread_count)
-        logger.debug("total_tasks {} thread_count {}".format(self.total_tasks, self.thread_count))
-
-        self.tasks_per_numa = int(math.ceil(self.tasks_per_node / 2.0))
-        smt_factor = max(1,int(self.get_value("MAX_TASKS_PER_NODE") / MAX_MPITASKS_PER_NODE))
-
-        threads_per_node = self.tasks_per_node * self.thread_count
-        threads_per_core = 1 if (threads_per_node <= MAX_MPITASKS_PER_NODE) else smt_factor
-        self.cores_per_task = self.thread_count / threads_per_core
 
         mpi_attribs = {
             "compiler" : self.get_value("COMPILER"),
@@ -148,16 +139,28 @@ class Case(object):
             "threaded" : self.get_build_threaded(),
             }
 
-        os.environ["OMP_NUM_THREADS"] = str(self.thread_count)
-
         executable = env_mach_spec.get_mpirun(self, mpi_attribs, job="case.run", exe_only=True)[0]
         if executable is not None and "aprun" in executable:
-            self.num_nodes = get_aprun_cmd_for_case(self, "acme.exe")[1]
+            _, self.num_nodes, self.total_tasks, self.tasks_per_node, self.thread_count = get_aprun_cmd_for_case(self, "acme.exe")
             self.spare_nodes = env_mach_pes.get_spare_nodes(self.num_nodes)
             self.num_nodes += self.spare_nodes
         else:
+            self.total_tasks = env_mach_pes.get_total_tasks(comp_classes)
+            self.tasks_per_node = env_mach_pes.get_tasks_per_node(self.total_tasks, self.thread_count)
+
             self.num_nodes, self.spare_nodes = env_mach_pes.get_total_nodes(self.total_tasks, self.thread_count)
             self.num_nodes += self.spare_nodes
+
+        logger.debug("total_tasks {} thread_count {}".format(self.total_tasks, self.thread_count))
+
+        self.tasks_per_numa = int(math.ceil(self.tasks_per_node / 2.0))
+        smt_factor = max(1,int(self.get_value("MAX_TASKS_PER_NODE") / max_mpitasks_per_node))
+
+        threads_per_node = self.tasks_per_node * self.thread_count
+        threads_per_core = 1 if (threads_per_node <= max_mpitasks_per_node) else smt_factor
+        self.cores_per_task = self.thread_count / threads_per_core
+
+        os.environ["OMP_NUM_THREADS"] = str(self.thread_count)
 
     # Define __enter__ and __exit__ so that we can use this as a context manager
     # and force a flush on exit.
@@ -728,10 +731,10 @@ class Case(object):
             mach_pes_obj.set_value(rootpe_str, rootpe)
 
         pesize = 1
-        MAX_MPITASKS_PER_NODE = self.get_value("MAX_MPITASKS_PER_NODE")
+        max_mpitasks_per_node = self.get_value("MAX_MPITASKS_PER_NODE")
         for val in totaltasks:
             if val < 0:
-                val = -1*val*MAX_MPITASKS_PER_NODE
+                val = -1*val*max_mpitasks_per_node
             if val > pesize:
                 pesize = val
         if multi_driver:
@@ -890,7 +893,7 @@ class Case(object):
         charge_account = get_charge_account(machobj)
         if charge_account is not None:
             self.set_value("CHARGE_ACCOUNT", charge_account)
-            
+
 
         # Resolve the CIME_OUTPUT_ROOT variable, other than this
         # we don't want to resolve variables until we need them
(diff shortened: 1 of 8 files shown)


jgfouca:
Let me try to add @worleyph as a reviewer.

jgfouca:
Might not let me do it until he accepts invitation to CIME. In any event, he should have gotten emails regarding this PR.

scripts/lib/CIME/case.py
```diff
@@ -129,36 +129,39 @@ def initialize_derived_attributes(self):
...
             "mpilib"   : self.get_value("MPILIB"),
             "threaded" : self.get_build_threaded(),
             }
 
-        os.environ["OMP_NUM_THREADS"] = str(self.thread_count)
-
         executable = env_mach_spec.get_mpirun(self, mpi_attribs, job="case.run", exe_only=True)[0]
         if executable is not None and "aprun" in executable:
-            self.num_nodes = get_aprun_cmd_for_case(self, "acme.exe")[1]
+            _, self.num_nodes, self.total_tasks, self.tasks_per_node, self.thread_count = get_aprun_cmd_for_case(self, "acme.exe")
```
jgfouca: Important change.

scripts/lib/CIME/case_setup.py
```diff
@@ -160,16 +160,12 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False):
...
-            logger.debug("at update TOTALPES = {}".format(pestot))
-            case.set_value("TOTALPES", pestot)
-            thread_count = env_mach_pes.get_max_thread_count(models)
-            cost_pes = env_mach_pes.get_cost_pes(pestot, thread_count, machine=case.get_value("MACH"))
-            case.set_value("COST_PES", cost_pes)
 
             case.initialize_derived_attributes()
 
+            cost_per_node = 16 if case.get_value("MACH") == "yellowstone" else case.get_value("MAX_MPITASKS_PER_NODE")
+            case.set_value("COST_PES", (case.num_nodes - case.spare_nodes) * cost_per_node)
```
jgfouca: Important change.

worleyph:
Hi @jgfouca , I'm not competent to review this. I don't know what these are used for, nor can I read the python with any degree of confidence. Even though these are currently often wrong on Titan, they do not seem to prevent correct execution. My ACME github issue was just to point out that they are either incorrect or the descriptions are incorrect (or both), in case anyone cares.

Based on the description for COST_PES ("pes or cores used relative to PES_PER_NODE for accounting"), perhaps the spare nodes should be accounted for, since they will be charged for? Again, I do not know how these are used.

jgfouca:
@jedwards4b @worleyph @rljacob @mvertens : the presence of aprun forces us to think about concepts like num_threads and thread_count on a per-component basis rather than as global values. This is fundamentally incompatible with CIME's current implementation. This is why, without a major refactor, things like COST_PES and other things in getTiming etc will always be wrong on titan for some pe-layouts. We've currently hacked our way around this problem by using special code to generate aprun commands for titan.

I think this PR puts us in slightly better shape since it at least ensures TOTALPES is correct on titan.

worleyph:
@jgfouca , please put this into an ACME branch when you get the chance and I will try it on Titan. The aprun command is correct currently on Titan, and I want to exercise your update to ensure that this continues to hold. Thanks.

(I haven't checked yet, but Theta at ALCF also uses aprun. Anyone know whether it also supports different process count/thread count per process for each node? If so, this will be important on more than just Titan.)

jgfouca:
@worleyph I'm pretty confident in the aprun command since it is well-protected by unit-tests. This change will be brought in during the next ESMCI merge to ACME and we can test it then.

worleyph:
Okay. Thanks.

jgfouca:
Maybe COST_PES is OK since it uses num_nodes (which is correct for aprun too) and just multiplies is by MAX_MPITASKS_PER_NODE.

Maybe having case.thread_count and case.tasks_per_node be wrong for aprun is irrelevant as long as the aprun command is being generated correctly?

rljacob:
FYI: Anyone with a valid github account can comment on issues on an open repo like CIME.  Accepting an invitation is only necessary for write access to the repo itself.

worleyph:
@jgfouca , COST_PES is not correct in my one example described in ACME github issue #1426  (unless you are saying the fix in this PR is correct).

Again, from what I can tell TOTAL_PES is an accurate count of the number of MPI tasks, though that is not what the descripion implies that it should be, and I am not sure that this is the information needed in get_timing.py either.

scripts/lib/CIME/case_setup.py
jedwards4b: Perhaps we should not attempt to automatically set COST_PES - since it often ends up incorrect.   Instead maybe we should consider adding a COST_PER_NODE variable to the machine definition in config_machines.xml?

jgfouca:
@rljacob , I invited @worleyph to ESMCI so I could request a review from him for this PR.

jgfouca:
@jedwards4b , I'm fine with continuing to include spare nodes in the COST_PES as long as @worleyph is OK with that. I would just need some guidance on how to change getTiming to report something useful.

jedwards4b:
What about adding COST_PER_NODE to the config_machines.xml file?

jgfouca:
@jedwards4b , I'm struggling with this PR since I have never used COST_PES for anything. I would be fine with anything both you and @worleyph would be happy with.

worleyph:
@jgfouca , as I indicated in the my first comment, I'm not competent to review the code. I was just pointing out that the values and/or descriptions of  COST_PES and TOTAL_PES are often incorrect on Titan. I can verify the fix when it is put into a version of ACME that I can try it out with, and once it is clarified what these values are supposed to be. 

I agree with @jedwards4b that, based on its name, it is reasonable that COST_PES should include the spare nodes. Depends on whether this is some evaluation of the model run cost as a performance metric (so would not include the spare nodes?) or whether it is the true charged cost for the run (in which case it should include the spare nodes?).

jgfouca:
@worleyph , that's fine, don't worry about doing a full review.

I believe this PR fixes TOTALPES, since it's now computed in the aprun module for titan (assuming TOTALPES = total number of MPI tasks).

If you are OK with COST_PES including spare nodes, then, basically, COST_PES = NUM_NODES * MAX_MPITASKS_PER_NODE. If that's OK then I think we're good to go. I'll just push a small update to add spare nodes.

worleyph:
I don't use COST_PES and TOTAL_PES for anything either. That said, I don't know if they are used by anything that I do use.

jedwards4b:
@jgfouca COST_PES = NUM_NODES * MAX_MPITASKS_PER_NODE is not always correct - yellowstone and cori-knl do not generally follow this rule.  This is why I want to introduce COST_PER_NODE.

worleyph:
@jgfouca , if TOTALPES is now the total number of MPI tasks then the description needs to be modified. It is currently

     desc>total number of tasks and threads (setup automatically - DO NOT EDIT)</desc>

which would imply a different value.

jgfouca:
@worleyph , OK I will fix.

jedwards4b:
TOTALPES should include the threads.  
COST_PER_NODE is the number of physical cores on a node and should be set in config_machines.xml COST_PER_NODE for yellowstone is 16 and for cori-knl is 68.
In most other cases it is MAX_MPITASKS_PER_NODE.

jgfouca:
I'm going to hold off further communication about this PR until the CIME call this afternoon. @worleyph , if you want to join it's at 1:00PM MDT, https://global.gotomeeting.com/join/243702717

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -2047,7 +2047,7 @@
     <default_value>0</default_value>
     <group>mach_pes_last</group>
     <file>env_mach_pes.xml</file>
-    <desc>total number of tasks and threads (setup automatically - DO NOT EDIT)</desc>
+    <desc>total number of MPI tasks (setup automatically - DO NOT EDIT)</desc>
```
jedwards4b: Now I'm confused - this should be the total number of physical cores used this would include threads unless the threads are using SMT.

jgfouca:
@jedwards4b , TOTALPES seems to be widely-used throughout CIME as the -np argument to mpiexec. That would mean it should not mean threads, right?

src/drivers/mct/cime_config/config_component.xml
worleyph: That is what I would have thought TOTALPES meant, but it has been returning number of MPI tasks recently.

rljacob:
I've always found the various PES, TASKS and other variables like that in CIME to be frustratingly difficult to remember what they are and what they mean.    What about a node does CIME need to know to provide its services?  The number of physical cores,  how many mpi-tasks can you put on those cores (sometimes not 1-1), how many nodes/cores do you need to ask for in a submission, how many nodes/cores are you going to run on, how many nodes/cores/tasks/threads are going to be given to each major component.  There's probably more.

jgfouca:
Discussed in meeting: TOTALPES is the number of mpi tasks, does not include thread info. We will make a follow-on PR to do something about COST_PES, COST_PER_NODE.

rljacob:
Then should we change TOTALPES to TOTALTASKS  ?   A Processor Element could be assigned a thread or a task.

PR #2003, opened 2017-10-27
Fix critical bug with py3 str handling

jgfouca:
There were many places hardcoded to check for the 'str' type.
With the python3 conversion, some of our strings have become
type unicode instead. This caused CIME to mistakenly think
these values were not strings in some cases.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/XML/entry_id.py
+++ b/scripts/lib/CIME/XML/entry_id.py
@@ -8,6 +8,7 @@ from CIME.utils import expect, convert_to_string, convert_to_type
 from CIME.XML.generic_xml import GenericXML
 
 from copy import deepcopy
+import six
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


jgfouca:
@jedwards4b i'm shocked and unnerved that nothing in scripts_regression_tests caught a critical mistake in generic_xml.get_resolved_value which is so widely used.

jgfouca:
This was caught by ACME testing for the latest cime merge. I'm not sure how to test this in general though.

billsacks:
@jgfouca are you able to briefly describe what problems this caused? I'm wondering if we need to flag all cime versions between the python3 conversion and this current PR as bad and tell people not to use them?

jgfouca:
@billsacks , this manifested on the ACME side with unresolved case variables on some platforms. For example, edison and cetus were trying to run their mpiexec equivalent with "-n $TOTALPES" which obviously didn't work.

billsacks:
Thanks @jgfouca . @jedwards4b do you have a sense of whether this will affect CESM users?

jedwards4b:
It may affect users on for example edison or cori, but we've tested our core machines pretty thoroughly and not run into any problems - this is also in our beta07 tag.   We can list it in the known problems as a potential problem, but I don't think that we need to make any statements about not using these cime versions.

billsacks:
Sounds good, thanks.

Issue #2006, opened 2017-10-28
run_merge_mapping_files.sh gives an error if no source point maps to marginal seas


billsacks:
When using run_merge_mapping_files.sh (in tools/mapping/gen_mapping_files/runoff_to_ocn), mapping a regional Greenland grid to an ocean grid, no source point maps to marginal seas.

This generates the following error:

```
fatal:Subscript out of range, error in subscript #0
fatal:An error occurred reading map_in_ms_S
fatal:["Execute.c":8640]:Execute: Error occurred at or near line 177
in file ncl/merge_mapping_files.ncl
```

Fix incoming.

PR #2007, opened 2017-10-28
Handle the case where no points are mapped to marginal seas

billsacks:
This case came up in mapping the regional Greenland grid to an ocean
grid. Without this fix in place, we got the following error:

```
fatal:Subscript out of range, error in subscript #0
fatal:An error occurred reading map_in_ms_S
fatal:["Execute.c":8640]:Execute: Error occurred at or near line 177
in file ncl/merge_mapping_files.ncl
```

Mike Levy gets most of the credit for this fix.

Test suite: Just ran the run_merge_mapping_files.sh case that was giving problems
Test baseline: n/a
Test namelist changes: n/a
Test status: bit for bit

Fixes ESMCI/cime#2006

User interface changes?: no

Code review:

### DIFF
--- a/tools/mapping/gen_mapping_files/runoff_to_ocn/ncl/merge_mapping_files.ncl
+++ b/tools/mapping/gen_mapping_files/runoff_to_ocn/ncl/merge_mapping_files.ncl
@@ -146,12 +146,12 @@ begin
 
   print("determining map_ms vals where REGION_MASK<0")
   sign_REGION_MASK_ms = sign_matlab(REGION_MASK_1d(map_in_ms_row-1))
-  if (any(sign_REGION_MASK_ms .lt. 0)) then
-    ind_vals_ms = ind(sign_REGION_MASK_ms .lt. 0)
-    n_s_subset_ms = dimsizes(ind_vals_ms)
-  else
+  ind_vals_ms = ind(sign_REGION_MASK_ms .lt. 0)
+  if (all(ismissing(ind_vals_ms))) then
     print("No source points mapped to marginal seas: output will just contain points mapped to open ocean")
     n_s_subset_ms = 0
+  else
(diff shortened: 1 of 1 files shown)


billsacks:
@klindsay28 can you please review this? (It won't let me add you as an official reviewer, maybe because you're not part of the cime group, but @mnlevy1981 and I would like to get your review on this change.)

klindsay28:
I would probably use the following (untested) diff on the first block, which has 1 fewer passes of the array sign_REGION_MASK_ms. But this is a minor point, take it or leave it at your discretion.

```
--- merge_mapping_files.ncl	2017-03-08 17:41:31.253293000 -0700
+++ merge_mapping_files.ncl.tmp	2017-10-28 07:41:52.408653000 -0600
@@ -103,7 +103,11 @@
   print("determining map_ms vals where REGION_MASK<0")
   sign_REGION_MASK_ms = sign_matlab(REGION_MASK_1d(map_in_ms_row-1))
   ind_vals_ms = ind(sign_REGION_MASK_ms .lt. 0)
-  n_s_subset_ms = dimsizes(ind_vals_ms)
+  if (all(ismissing(ind_vals_ms))) then
+    n_s_subset_ms = 0
+  else
+    n_s_subset_ms = dimsizes(ind_vals_ms)
+  end if
 
   n_s_out = n_s_subset_oo + n_s_subset_ms
 
```

billsacks:
Thanks @klindsay28 - I made your suggested change.

mnlevy1981:
Since the changes are so small, should we also update the script to handle the case where no points are mapped to the open ocean? @billsacks and I talked about it on Friday and at that time I didn't think it was necessary... but maybe it would be better to add similar checks to `ind_vals_oo` and `ns_subset_oo` than to wait until someone comes up with a regional map that only sends runoff to marginal seas and creates another issue ticket?

PR #2008, opened 2017-10-30
Fixes for ACME MPAS builds

jgfouca:
It  makes it so acme does not try to reuse sharedlib for both cases.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @billsacks

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -190,7 +190,8 @@ class SystemTestsCompareTwo(SystemTestsCommon):
                 # the MPAS build system
                 self._case2.set_value("SHAREDLIBROOT",
                                       self._case1.get_value("SHAREDLIBROOT"))
-                self.build_indv(sharedlib_only=sharedlib_only, model_only=model_only)
+
+            self.build_indv(sharedlib_only=sharedlib_only, model_only=model_only)
         else:
             self._activate_case1()
             self.build_indv(sharedlib_only=sharedlib_only, model_only=model_only)
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -39,6 +39,7 @@ from CIME.SystemTests.system_tests_common import SystemTestsCommon
 from CIME.case import Case
 from CIME.case_submit import check_case
 from CIME.case_st_archive import archive_last_restarts
+from CIME.utils import get_model
 
 import shutil, os, glob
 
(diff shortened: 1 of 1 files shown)


scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -184,8 +185,9 @@ def build_phase(self, sharedlib_only=False, model_only=False):
             # Although we're doing separate builds, it still makes sense
             # to share the sharedlibroot area with case1 so we can reuse
             # pieces of the build from there.
-            self._case2.set_value("SHAREDLIBROOT",
-                                  self._case1.get_value("SHAREDLIBROOT"))
+            if get_model() != "acme":
```
billsacks: Can you please add a comment describing why it doesn't work to do this for acme?

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -299,7 +301,7 @@ def _get_case2_exeroot(self):
             # to this case. (If case2's CIME_OUTPUT_ROOT were in some
             # more generic location, then this would result in its bld
             # directory being inadvertently shared with other tests.)
-            case2_exeroot = os.path.join(self._get_output_root2(), "bld")
+            case2_exeroot = os.path.join(self._get_output_root2(), self._case1.get_value("CASE"), "bld")
```
billsacks: This was not an inadvertent change - it was changed very intentionally. Please see the above comment.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -311,7 +313,7 @@ def _get_case2_rundir(self):
         """
         # Put the case2 run directory alongside its bld directory for
         # consistency. (See notes about EXEROOT in _get_case2_exeroot.)
-        case2_rundir = os.path.join(self._get_output_root2(), "run")
+        case2_rundir = os.path.join(self._get_output_root2(), self._case1.get_value("CASE"), "run")
```
billsacks: Again, this was changed intentionally - see above comment.

jgfouca:
@billsacks , sorry, didn't realize that was intentional. I reverted those changes and kept the the sharedlib change which is the only important one for MPAS.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -184,9 +185,12 @@ def build_phase(self, sharedlib_only=False, model_only=False):
...
             # pieces of the build from there.
-            self._case2.set_value("SHAREDLIBROOT",
-                                  self._case1.get_value("SHAREDLIBROOT"))
-            self.build_indv(sharedlib_only=sharedlib_only, model_only=model_only)
+            if get_model() != "acme":
+                # We need to turn off this change for ACME because it breaks
+                # the MPAS build system
+                self._case2.set_value("SHAREDLIBROOT",
+                                      self._case1.get_value("SHAREDLIBROOT"))
+                self.build_indv(sharedlib_only=sharedlib_only, model_only=model_only)
```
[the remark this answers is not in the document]
jgfouca: Yikes, nice catch!

jgfouca:
@jedwards4b this is ready.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
jedwards4b: Is this a deficiency in the mpas build that should be corrected?

jgfouca: Yes, at some point it will need to be fixed. I've noticed that ACME doesn't do well with sharedlibs in general, which is why we still force every case to build its own sharedlibs.

Issue #2009, opened 2017-10-31
runoff_to_ocn mapping: mask_b and frac_b in nn smooth map are mostly 0


billsacks:
When using a file_ocn_coastal_mask that is a subset of the full file_ocn in the runoff_to_ocn tool, the mask_b and frac_b values in the resulting nearest-neighbor smoothed map are mostly 0. Specifically, it appears that they are only 1 over coastal points, whereas they should be 1 over all ocean points.

I'm not sure if these incorrect mask_b and frac_b values would have any impact on the operation of the coupled system (but we should check that). However, they do call check_map to fail with conservation errors.

PR #2013, opened 2017-10-31
For runoff map: fix mask_b and frac_b in nearest neighbor map

billsacks:
Main change: For runoff map: fix mask_b and frac_b in nearest neighbor map.
Previously, mask_b and frac_b were taken from the nearest neighbor map
to the coastal ocean, which is a small subset of what this mask/frac
should be.

Also, in the check_map tool: Do not try to get title attribute.
This wasn't being used, and I was worried that it might cause a crash
when trying to run check_map on files created with
run_merge_mapping_files.sh (because no title attribute was present on
the resulting files).

Test suite: manual checks of mapping files generated with runoff_to_ocn tool
   and run of check_map.sh tool
Test baseline: n/a
Test namelist changes: n/a
Test status: bit for bit

Fixes ESMCI/cime#2009

User interface changes?: none

Code review:

### DIFF
--- a/tools/mapping/check_maps/src/ESMF_RegridWeightGenCheck.F90
+++ b/tools/mapping/check_maps/src/ESMF_RegridWeightGenCheck.F90
@@ -57,7 +57,7 @@ program OfflineTester
 
       type(ESMF_VM) :: vm
 
-      character(ESMF_MAXSTR) :: wgtfile, title
+      character(ESMF_MAXSTR) :: wgtfile
 
       real(ESMF_KIND_R8), pointer :: factorList(:)
       integer, pointer            :: factorIndexList(:,:)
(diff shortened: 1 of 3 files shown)


billsacks:
@mnlevy1981 - note that this also includes the changes from #2007 . I couldn't get the PR workflow we discussed to work right, so this branch includes both. So:

(1) For your review, do not bother reviewing the first two commits on this branch – just review f4f923f and 4a3fac5.

(2) You should merge #2007 before merging this one - otherwise I think this will auto-close #2007 .

billsacks:
@mnlevy1981 very good point - I have removed the offending comment.

billsacks:
@mnlevy1981 is this ready to merge to master?

billsacks:
@mnlevy1981 oh wait, never mind - I forgot my own comment saying that we'll merge the other PR first.

PR #2021, opened 2017-11-02
Fix cprnc build on my mac

### LINKED ISSUE #2020: Makefile-based cprnc build fails on my Mac
When I try to build cprnc on my Mac with the Makefile-based build, the link step fails because none of the LDFLAGS settings are added to the link line. Fix incoming.

billsacks:
The original version was overwriting LDFLAGS with SLIBS, where the
latter was unset. The new version preserves both.

It's possible that the overwriting of LDFLAGS with SLIBS (from
e2323e10f1) was needed on some systems, but that seems like the wrong
thing to do in general.

Test suite: Just tested cprnc build on my Mac (with gnu), hobart (intel) and edison (intel)
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

Fixes #2020 

User interface changes?: N

Code review:

### DIFF
--- a/tools/cprnc/Makefile
+++ b/tools/cprnc/Makefile
@@ -45,7 +45,6 @@ GENF90 = ../../src/externals/genf90/genf90.pl
 
 FC := $(SFC)
 FFLAGS += -I$(INC_NETCDF) -I.
-LDFLAGS=$(SLIBS)
 #------------------------------------------------------------------------
 # Default rules and macros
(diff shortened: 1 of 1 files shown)


billsacks:
@jedwards4b it looks like you added the setting of LDFLAGS to SLIBS a couple of years ago (see commit referenced above), so I'll let you comment on whether my change might break something that you can remember.

Issue #2010, opened 2017-10-31
CESM's rof2ocn liquid maps likely have incorrect mask_b and frac_b


billsacks:
Due to #2009 @mnlevy1981 and I feel that we should remake all of CESM's rof2ocn liquid maps (once #2009 is fixed). For at least one case, we should then check if this changes answers at all - i.e., whether the old, buggy mask_b and frac_b had any impact on the system.

alperaltuntas:
I generated two sets of rx1->gx37 maps: (1) using the mapping tool without the latest fixes, and (2) using the mapping tool with the latest fixes (#2007 #2013 #2022). I ran C cases with both maps and compared the results. It appears that answers do NOT change, so I think it's not necessary to remake rof2ocn liq maps. (Answer changes I saw earlier were because I have not specified the coastal grid files in nml file for the map generating tool, and so the nearest neighbor mappings were not confined to the coast.)

billsacks:
Thanks for checking. That's a relief.

PR #2022, opened 2017-11-02
If area units on source grid file are in sq deg, convert to sq radians

### LINKED ISSUE #2018: Runoff mapping generates incorrect weights for scrip grid files with area in square degrees
When running `runoff_map` (in `tools/mapping/gen_mapping_files/runoff_to_ocn`) with a file with `gridtype = 'scrip'`, which has area in square degrees: The tool generates incorrect weights because it assumes that areas are in square radians. The resulting weights are off by a factor of about 3300 ((180/pi)^2).

It turns out that this doesn't affect the values mapped by the coupler (at least not much), maybe because the area correction in the coupler takes care of this? But we should still fix this so that we have correct mapping weights on the runoff mapping files.

billsacks:
This just applies to source grid files in the scrip format.

Without this fix, areas were assumed to be in sq radians, which led to
mapping weights that were off by about a factor of 3000 if areas were
actually in sq degrees.

Test suite: Tested runoff_map tool with a source file with units in sq
            degrees; confirmed that area_a and S values now have the
            right order of magnitude. Also tested with a source file
            with units in sq radians, and confirmed that the resulting
            mapping file is the same as before.
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes #2018 

User interface changes?: none

Code review:

### DIFF
--- a/tools/mapping/gen_mapping_files/runoff_to_ocn/src/map_mod.F90
+++ b/tools/mapping/gen_mapping_files/runoff_to_ocn/src/map_mod.F90
@@ -644,11 +644,21 @@ SUBROUTINE map_gridRead(map, rfilename, ofilename, gridtype, lmake_rSCRIP)
 
       rcode = nf_inq_varid     (fid,'grid_imask',vid )
       rcode = nf_get_var_int   (fid,vid     ,map%mask_a)
+
       rcode = nf_inq_varid     (fid,'grid_area',vid )
-      if (rcode.eq.0) then
-         rcode = nf_get_var_double(fid,vid     ,map%area_a)
-      else
+      if (rcode.ne.0) then
          write(6,*) "ERROR: could not find variable grid_area in source grid input file!"
+         stop
+      end if
+      rcode = nf_get_var_double(fid,vid     ,map%area_a)
+      units = "" ! units needs to be emptied before reading from netCDF file
+      rcode = nf_get_att_text(fid, vid, "units", units)
+      if (trim(units).eq."square radians") then
+         ! Nothing to do
+      else if (trim(units).eq."square degrees") then
+         map%area_a = map%area_a * DEGtoRAD * DEGtoRAD
+      else
+         write(6,*) "ERROR: Unrecognized units for source grid_area: ", trim(units)
          stop
       end if
 
--- a/tools/mapping/gen_mapping_files/runoff_to_ocn/src/map_mod.F90
+++ b/tools/mapping/gen_mapping_files/runoff_to_ocn/src/map_mod.F90
@@ -653,12 +653,18 @@ SUBROUTINE map_gridRead(map, rfilename, ofilename, gridtype, lmake_rSCRIP)
       rcode = nf_get_var_double(fid,vid     ,map%area_a)
       units = "" ! units needs to be emptied before reading from netCDF file
       rcode = nf_get_att_text(fid, vid, "units", units)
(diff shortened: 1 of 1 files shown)


billsacks:
@alperaltuntas - @mnlevy1981 suggested that we start adding you as a reviewer for runoff mapping-related changes, so I have done that here. I also went ahead and made you the assignee. Let me know if you have questions about the process of being an assignee, because it looks like this may be the first cime PR for which you're the assignee.

billsacks:
Good catch. I have just pushed a commit that aborts if no units are found. I'm more comfortable with this conservative approach rather than assuming units of square radians.

mnlevy1981:
@billsacks  -- I'm okay with aborting if no units are found, but I'd like to wait until cheyenne is back up and test this by re-generating `r05` and `r01` to `gx3v7` maps (something I think we need to do anyway because of #2010); I don't like introducing new exit criteria without verifying that our current examples don't trigger it. If it turns out that we need to add units to an existing SCRIP grid file, then we should also update the corresponding `nml` files that come bundled with this tool.

billsacks:
@mnlevy1981 The r05, r01 and rx1 mappings do not enter the block of code modified in this PR, because they use 'rtm'-formatted files (for r05 and r01) or 'obs'-formatted files (for rx1).

The two `map_wr50a_*` template namelist files do use 'scrip'-formatted inputs. I'm not sure what these are for: I don't see anything referring to wr50 in either CESM or ACME's config_grids file.

As an aside: In looking into this, I noticed that (if I'm reading the code correctly) incorrect areas and mapping weights will currently be used if you try to generate mappings with the r01 grid, because 'rtm'-formatted files use this block of code, which assumes a 0.5 deg grid:

```fortran
         map%area_a(n) = 0.5 * 0.5 * cos(lat*DEGtoRAD) * DEGtoRAD * DEGtoRAD
```

ekluzek:
@mnlevy1981 I'm pretty sure we had problems with the r01 grid and hence haven't used it. There is a 8th degree grid for MOSART, but that hasn't been a standard grid that we run with either. Collin Z. used it in some high res. simulations that he did, but there were problems with it. So I don't know that we have any working high resolution ROF grids. 

But, I do think there is a need for this for high resolution work. So we might want to check with scientists on the need for it.

billsacks:
> Is the r01 grid used at all? I have a vague memory of noticing the hard-coded 0.5 in the line you mentioned, but I didn't do anything about it because I figured it was only encountered with the r05 grid. Another possible fix would be to add `rtm_cell_width_deg` or something similar to the namelist (with default value of `0.5`) and instructions to set it properly when the runoff grid is rtm-formatted.

We have old mapping files from r01 to gx1v6 `map_r01_to_gx1v6_120711.nc` but not to gx1v7 or any other grid.

It feels like the best solution would be to require rtm-formatted files to have the necessary metadata on them. That would require making a new version of the r05 (and r01) file with this metadata added. I'm also okay with your proposed namelist option, though - as long as the default is some 'unset' value so that it is required to be set explicitly. I think this should be deferred to a separate issue (as opposed to folding it into this PR).

mnlevy1981:
> It feels like the best solution would be to require rtm-formatted files to have the necessary metadata on them. That would require making a new version of the r05 (and r01) file with this metadata added.

Oh, that's a clever solution

> I think this should be deferred to a separate issue (as opposed to folding it into this PR

I agree

Issue #2024, opened 2017-11-02
aprun sub needs to handle all options 


jedwards4b:
acme machine theta is currently bypassing the aprun function call in case.py because that function doesn't handle the -j option, but that breaks runs on theta where $TOTALTASKS < $MAX_MPITASKS_PER_NODE.     I think that the aprun function should be expanded to handle -j so that we don't need an exception for theta.

jgfouca:
@rljacob , who is the POC for theta? They should probably take this ticket since I know next to nothing about that machine.

rljacob:
@amametjanov is the POC for theta.

jgfouca:
@amametjanov , would you mind taking this one?

jedwards4b:
This really isn't particularly about theta - I think that we need to extend the aprun function to handle more (all?) of the options of the aprun command.

rljacob:
Yes I think this needs a CIME developer more then a Theta expert.

jgfouca:
I can take this ticket if someone can explain the nuances of aprun to me.

jgfouca:
I found some documentation online; I will attempt this ticket.

jedwards4b:
@jasonb5 have you looked at this at all yet.   I would prefer a more general solution than #4332

jasonb5:
@jedwards4b I've looked at this a little, my understanding is we need CIME to support all of `aprun`'s flags in order for machines like `theta` and the ones listed in #4332 to utilize CIME's `aprun` capability.

Is there a link to documentation for `aprun`'s flags? I found this [document](https://bluewaters.ncsa.illinois.edu/using-aprun).

jedwards4b:
Or [this](https://support.hpe.com/hpesc/public/docDisplay?docLocale=en_US&docId=a00114008en_us&page=Run_Applications_Using_the_aprun_Command.html)

jasonb5:
@jedwards4b Ok great, I'm looking at this now.

PR #2025, opened 2017-11-02
Allow overrides in config batch

jedwards4b:
Allow duplicate entries in config_batch.xml with later entry taking precidence.
Fix more issues on bluewaters

Test suite: scripts_regression_tests.py on bluewaters
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


scripts/lib/CIME/case.py
```diff
@@ -1238,7 +1238,7 @@ def get_mpirun_cmd(self, job="case.run"):
         executable, mpi_arg_list = env_mach_specific.get_mpirun(self, mpi_attribs, job=job)
 
         # special case for aprun
-        if executable is not None and "aprun" in executable and "titan" in self.get_value("MACH"):
+        if executable is not None and "aprun" in executable and not "theta" in self.get_value("MACH"):
```
jedwards4b: I have opened issue #2024 - I don't think we should have an exception here.

PR #2026, opened 2017-11-02
add function to rebuild dependencies

### LINKED ISSUE #1889: Wanted: command to clean the Srcfiles and Depends files for a component
At least in some circumstances, if you rebuild a case (i.e., it already has been built, then you rerun `./case.build`) after adding a new source file, the build fails because the new source file and the changed dependency tree are not picked up.

For example, I added a file named NumericsMod.F90 to the CLM source tree, and added `use NumericsMod` in SoilHydrologyMod.F90. Then I reran `./case.build` and got:

```
   use NumericsMod       , only : truncate_small_values
      1
Fatal Error: Can't open module file 'numericsmod.mod' for reading at (1): No such file or directory
compilation terminated.
```

We discussed this on a recent cime concall, and felt that it would be too hard to automatically detect this situation. What I'm asking for here is something simpler: It would be great if there were some `./case.build` option that reset a given component's build _just enough_ to allow picking up any new source files and regenerating the dependency tree. From a quick look at the CESM Makefile, I think this would involve a new set of clean commands that just remove Srcfiles from the given component (which it seems would also force a regeneration of Depends).

So can we have something like `./case.build --reset-srcfiles [component]` that calls something like `make clean-srcfiles-component` for the given component?

jedwards4b:
Add an argument to case.build --clean-depends that will go through each of the component obj files and remove file Srcfiles which will cause this file and the Depends file to be recreated.   This allows a user to add files to SourceMods and rebuild without requiring a full clean build.

Test suite: scripts_regression_tests.py, hand tests 
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1889 

User interface changes?: Yes add --clean-depends to case.build, this argument can take a component list and will clean all components if one is not provided

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/acme/machines/Makefile
+++ b/config/acme/machines/Makefile
@@ -868,6 +868,39 @@ $(COMPLIB): $(OBJS)
 %.F90: %.F90.in
 	$(CIMEROOT)/src/externals/genf90/genf90.pl $< > $@
 
+clean_dependsatm:
+	$(RM) -f $(EXEROOT)/atm/obj/Srcfiles
+
+clean_dependscpl:
+	$(RM) -f $(EXEROOT)/cpl/obj/Srcfiles
+
+clean_dependsocn:
+	$(RM) -f $(EXEROOT)/ocn/obj/Srcfiles
+
+clean_dependswav:
+	$(RM) -f $(EXEROOT)/wav/obj/Srcfiles
+
+clean_dependsglc:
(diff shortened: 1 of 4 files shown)


jedwards4b:
I do it now through gmake so it will handle shared directories correctly.

PR #2028, opened 2017-11-03
seperate data_assimilation by component and add to fortran

### LINKED ISSUE #1980: Importing DO_DATA_ASSIMILATION into components
This issue is related to #1807 (CESM+DART tag testing)
and is created to handle the inclusion of DART SourceMods
into the CESM trunk, so that they are not needed.
After some discussions it appears (to me) that the most
pervasive change needed is to provide the value of
env_run.xml:DO_DATA_ASSIMILATION to the components,
so that they can take appropriate actions ( restart in a certain
mode, change printed output, etc.).  

I can start looking at how to implement that, but I'll be starting
from mostly ignorance.  It would be helpful if someone would
outline how an env_run.xml variable should be communicated
to the components.

I may not be understanding all of the important implications of this.
From email discussions:
@mvertens  (2017-10-4) CIME/drv:
"there will need to be source mods or a more extensible implementation
to set read_restart = .false. in dart mode."
@mvertens 2017-10-4) POP: 
"(from DART's SourceMod)
    !*****hardcode read_restart so that dlnd does not look for rpointers
    !**** or restart files
If this is really needed,  we need to come up with a more extensible scheme to put it in."
@billsacks  (2017-10-6) CLM:
"it might be possible to do something a bit more nuanced than just disabling the balance checks entirely. For example, maybe it's only a problem for the first time step or two after restarting in a data assimilation context, in which case it might be best if we only bypass the balance checks for that first time step or two."
Tim Hoar (2017-10-6) CLM:
"I'd like CLM to recompute the snow layers, etc. Snow is a real nuisance when it comes 
to DA, it is easiest to modify total SWE - but the prognostic variables in CLM are the individual snow layers. Going from SWE to the layers changes with the versions of CLM,
so I'd like to not have to try to track those changes in the DART scheme.
Fei has moved on to the CICE model, but has similar desires for the ice classes 
(not a CLM problem, I know)."
Kevin; We decided in a meeting that this is a separate issue to be fixed later,
but it will be another example of CLM doing something differently based on 
DO_DATA_ASSIMILATION.

Thanks!
Kevin, for DART

jedwards4b:
Separate DATA_ASSIMILATION flag by component and add to fortran infodata
components can now query infodata for the flag data_assimilation_xxx where xxx is one of
(atm, ocn, wav, glc, ice, rof, lnd)

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1980 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/SystemTests/dae.py
+++ b/scripts/lib/CIME/SystemTests/dae.py
@@ -49,7 +49,7 @@ class DAE(SystemTestsCompareTwo):
         expect(os.path.isfile(da_file), "ERROR: da_file, '{}', does not exist".format(da_file))
 
         # Set up two data assimilation cycles each half of the full run
-        self._case.set_value("DATA_ASSIMILATION", "TRUE")
+        self._case.set_value("DATA_ASSIMILATION_ATM", "TRUE")
         self._case.set_value("DATA_ASSIMILATION_SCRIPT", da_file)
         self._case.set_value("DATA_ASSIMILATION_CYCLES", 2)
         stopn = self._case.get_value("STOP_N")
--- a/scripts/lib/CIME/case_run.py
+++ b/scripts/lib/CIME/case_run.py
@@ -270,7 +270,7 @@ def case_run(case, skip_pnl=False):
     prerun_script = case.get_value("PRERUN_SCRIPT")
     postrun_script = case.get_value("POSTRUN_SCRIPT")
 
-    data_assimilation = case.get_value("DATA_ASSIMILATION")
+    data_assimilation = any(case.get_values("DATA_ASSIMILATION"))
     data_assimilation_cycles = case.get_value("DATA_ASSIMILATION_CYCLES")
     data_assimilation_script = case.get_value("DATA_ASSIMILATION_SCRIPT")
 
--- a/scripts/lib/CIME/case_submit.py
+++ b/scripts/lib/CIME/case_submit.py
@@ -118,7 +118,7 @@ def check_case(case):
     logger.info("Check case OK")
 
 def check_DA_settings(case):
-    if case.get_value("DATA_ASSIMILATION"):
+    if any(case.get_values("DATA_ASSIMILATION")):
         script = case.get_value("DATA_ASSIMILATION_SCRIPT")
         cycles = case.get_value("DATA_ASSIMILATION_CYCLES")
         logger.info("Data Assimilation enabled using script {} with {:d} cycles".format(script,cycles))
--- a/src/drivers/mct/cime_config/config_component.xml
+++ b/src/drivers/mct/cime_config/config_component.xml
@@ -2575,11 +2575,20 @@
   <entry id="DATA_ASSIMILATION">
     <type>logical</type>
     <valid_values>TRUE,FALSE</valid_values>
-    <default_value>FALSE</default_value>
     <group>external_tools</group>
     <file>env_run.xml</file>
     <desc> Run the external tool pointed to by DATA_ASSIMILATION_SCRIPT after the model run completes </desc>
+    <values>
+      <value compclass="ATM">FALSE</value>
+      <value compclass="OCN">FALSE</value>
+      <value compclass="WAV">FALSE</value>
+      <value compclass="GLC">FALSE</value>
+      <value compclass="ICE">FALSE</value>
+      <value compclass="ROF">FALSE</value>
+      <value compclass="LND">FALSE</value>
+    </values>
   </entry>
+
   <entry id="DATA_ASSIMILATION_CYCLES">
     <type>integer</type>
     <valid_values></valid_values>
--- a/src/drivers/mct/cime_config/namelist_definition_drv.xml
+++ b/src/drivers/mct/cime_config/namelist_definition_drv.xml
@@ -3934,4 +3934,88 @@
     </values>
   </entry>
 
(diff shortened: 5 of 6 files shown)


kdraeder:
Thanks for all the work on this!
On first reading, it looks like what is needed.

I want to confirm that any combination of DATA_ASSIMILATION_CCC
can be chosen by the user.   Then these will be available to 
DATA_ASSIMILATION_SCRIPT, which will be responsible for doing
the assimilation in the chosen components.

I'm a bit concerned that
data_assimilation = any(case.get_values("DATA_ASSIMILATION"))
may capture DATA_ASSIMILATION_SCRIPT and 
DATA_ASSIMILATION_CYCLES too.  Is that intended?

I haven't tested this at all.
Kevin

kdraeder:
Steve, thanks for reviewing.

We currently can do assimilation in CAM, CLM, POP, and CICE.
I'm hoping to add CISM during CESM2's lifetime, and there are
people doing river runoff (hydrology) data assimilation with other models, 
so that's on the horizon.  That leaves only WAV as a component
with no anticipated data assimilation.  I can see leaving out ROF
and WAV at the moment, which would be easy to add later since
this template will exist.

This feature is part of the effort to merge the DART SourceMods
into the CESM2 trunk.  Those mods have/need/want a logical variable
in each component to control the bits of code needed by DA.  
The sourcemods of the 4 components listed above have been tested 
independently, some only in earlier CESMs.  We would like to test
them all in CESM2, which will be much easier if the mods are
already in the trunk.

This mechanism came originally from CSEG in the context of CAM.
I've recently met with CLM, POP, and CICE developers, who agree 
that this is reasonable thing to do and a good mechanism.

jedwards4b:
@kdraeder This any(case.get_values("DATA_ASSIMILATION")) does not capture DATA_ASSIMILATION_SCRIPT or DATA_ASSIMILATION_CYCLES.
@gold2718 the coupler is providing this flag which components may or may not use.   I see no reason to wait until a component uses it for the coupler to provide it.    It won't do anything in the components until a component chooses to use it, but if we don't provide it then no component can use it.  I don't understand your point.

gold2718:
My point is that the components do not need any flag. If they are participating in data assimilation, they should have a component-specific namelist variable for this purpose.
If the components need a signal for when to process data assimilation input (i.e., update their state), the coupler should use the resume signal which was designed for this purpose.
We have suffered in the past by having the coupler try to direct component behavior (e.g., adiabatic or ideal mode in CAM) and have finally removed these situations. I feel that this PR goes back down that path.

jedwards4b:
I think that this will require a wider discussion.  @mvertens can you review and comment.

mvertens:
Could @jedwards4b or @kraeder please explain how they envision these DATA_ASSIMILATION_XXX variables to be used once they are in infodata.
Lets look at an example that @gold2718  would have:
CAM wants to do data assimilation 
1. It could query the xml variables DATA_ASSIMILATION_CYCLES and DATA_ASSIMILATION_SCRIPT and the number of instances - if those are evaluated to trigger data assimilation - the CAM would set its internal namelist to carry out data assimilation with the relevant functionality needed to do this.
2. Alternatively, CAM could look at the xml variables DATA_ASSIMILATION_ATM and if that is set - then it could set its own namelist variable - and not query infodata.
I think the key issue is not the introduction of the xml variable - but if that variable gets used by the driver in infodata - or if that variable gets used by CAM to generate its own namelist. I can see the point that @gold2718 is making in that I don't think we want to be introducing new variables into infodata if we can avoid it. However, I think the introduction of these new xml variables are a clean way to trigger the component to perform the data assimilation functionality it needs.
Does what I am saying make sense?

jedwards4b:
We can't use the _CYCLES or _SCRIPT variables since we want to distinguish which components have DATA_ASSIMILATION enabled.    Looking at the resume variables introduced by @gold2718 it may be possible to use these and avoid the new variable.   But this mechanism is needed for current data_assimilation system, not a theoretical future one.

gold2718:
@jedwards4b, why are you referring to pause/resume signaling as a theoretical future system? The functionality is available in CIME and has been for months. While the names (not chosen by me) suggest use only with a pause/resume system, the signaling can be used today to send signals to models at prescribed intervals.
What might be a (simple) desired addition in this case is an additional `PAUSE_OPTION` (e.g., `'startup'`) which would tell the driver to send a resume signal only at the beginning of a run (during initialization).
However, I am not convinced that even that is necessary -- it all depends on how the components are informed that they should ingest an assimilated state (as opposed to an initial state or a classic restart state). If the data assimilation script is going to set the necessary model-specific namelist variable(s), then the model will know what to do when it starts up. In this case, I see no need for any driver intervention or signaling (including resume signals).

jedwards4b:
My understanding is that the component models need to know if they are in assimilation mode so that they can optionally choose the modified code path currently handled in SourceMods by the dart team.  If this can be done with the current resume flag I have no problem with that, but it strikes me that just knowing you are in da mode might be different information than what the resume flag is intended to convey.

gold2718:
I agree that the models need to know when to do assimilation. However, over at issue #1980, @kdraeder seems to indicate that this is already being done (in the SourceMods version) through model namelist variables (different variables for different components). I still would like to understand what is missing from the SourceMods version that requires _any_ signaling from the driver.

jedwards4b:
Replaced by PR #2039

PR #2030, opened 2017-11-06
Convert SEQ to use compare-two

### LINKED ISSUE #1862: Get SEQ using compare-two test infrastructure
This test is important to ACME and it's still doing old-school XML/exe shuffling.

jgfouca:
Test suite: code_checker, seq by-hand and with re-run
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1862 

User interface changes?: N

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -53,8 +53,7 @@ class SystemTestsCompareTwo(SystemTestsCommon):
                  run_two_suffix = 'test',
                  run_one_description = '',
                  run_two_description = '',
-                 multisubmit = False,
-                 expected=None):
+                 multisubmit = False):
         """
         Initialize a SystemTestsCompareTwo object. Individual test cases that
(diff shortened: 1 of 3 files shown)


scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -53,7 +53,8 @@ def __init__(self,
                  run_two_suffix = 'test',
                  run_one_description = '',
                  run_two_description = '',
-                 multisubmit = False):
+                 multisubmit = False,
+                 expected=None):
```
billsacks: Can you please add some documentation of what this "expected" argument is for?

Actually, from a quick read of the code, I wonder if this is really needed now: It looks like this is just needed for working with locked files, and I'm wondering if that is no longer needed now that you're using the system_tests_compare_two infrastructure.

jgfouca:
@billsacks , thanks, I think you're right about expected.

jgfouca:
@jedwards4b , I think this is ready.

PR #2034, opened 2017-11-07
Add retry capability to create_test

### LINKED ISSUE #1865: To deal with system issues, automatically retry tests
This is a low-priority enhancement request; I just want to record some thoughts on it for the future.

When running a test suite, we often run into problems where a small number of tests failed due to system issues. It slows down the testing/tagging process to have to go back and manually rerun these tests. It would be nice if cime's testing facilities had an option to automatically retry tests N times before giving up.

@ekluzek raised this at a CSEG meeting June 12, 2017. @jedwards4b felt this could cause too much confusion. But I was just thinking: A way around the confusion problem would be for the test system to create a totally new version of the test, with a new test id (say, by appending "try2" to the testid) (I realize that may not be trivial to implement, though). @mnlevy1981 suggests having a flag to --create-test (like --num-tries  – which is my name, not his). This would default to 1, but you could set it to something like 3 if you want

jgfouca:
Plus new regression test to exercise this capability.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1865 

User interface changes?: Yes, new --retry option to create_test

Code review: @jedwards4b  @billsacks

### DIFF
--- a/scripts/create_test
+++ b/scripts/create_test
@@ -234,6 +234,9 @@ OR
                         "file. The file can follow either the config_pes.xml or "
                         "the env_mach_pes.xml format.")
 
+    parser.add_argument("--retry", type=int, default=0,
+                        help="Automatically retry failed tests. >0 implies --wait")
+
     args = CIME.utils.parse_args_and_handle_standard_logging_options(args, parser)
 
     # generate and compare flags may not point to the same directory
@@ -256,6 +259,9 @@ OR
 
     expect(not (args.namelists_only and not (args.generate or args.compare)),
            "Must provide either --compare or --generate with --namelists-only")
+
+    if args.retry > 0:
+        args.wait = True
 
     if args.parallel_jobs is not None:
         expect(args.parallel_jobs > 0,
@@ -379,7 +385,7 @@ OR
         args.namelists_only, args.project, \
         args.test_id, args.parallel_jobs, args.walltime, \
         args.single_submit, args.proc_pool, args.use_existing, args.save_timing, args.queue, \
-        args.allow_baseline_overwrite, args.output_root, args.wait, args.force_procs, args.force_threads, args.mpilib, args.input_dir, args.pesfile
+        args.allow_baseline_overwrite, args.output_root, args.wait, args.force_procs, args.force_threads, args.mpilib, args.input_dir, args.pesfile, args.retry
 
 ###############################################################################
 def single_submit_impl(machine_name, test_id, proc_pool, _, args, job_cost_map, wall_time, test_root):
@@ -514,7 +520,7 @@ def create_test(test_names, test_data, compiler, machine_name, no_run, no_build,
         # Create submit script
         single_submit_impl(machine_name, test_id, proc_pool, project, sys.argv[1:], job_cost_map, walltime, test_root)
 
-    return 0 if success else CIME.utils.TESTS_FAILED_ERR_CODE
+    return success
 
 ###############################################################################
 def _main_func(description):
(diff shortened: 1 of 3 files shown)


billsacks:
@jgfouca - Thanks a lot for taking this on!

My understanding of the requests from #1865 - particularly from @jedwards4b and @gold2718 - was that people wanted the ability for the test system to create totally new versions of the failed tests, rather than just rerunning out of the existing test directories. My understanding is that that isn't done here. I don't personally have strong feelings about that, but I'd like to let them comment on whether they feel that's important.

jgfouca:
@billsacks that would make for a much-more complicated PR. I'd prefer to stay with this approach unless people really don't want it. The logs will contain the failures if they happened.

jedwards4b:
I think this approach is fine, thanks.

billsacks:
Okay, that's fine with me, then. I'll give @gold2718 a little time to reply in case he feels strongly.

I have a few other questions:

(1) Just for my own understanding: What are the implications of setting `--wait`? Does this simply mean that the create_test process doesn't exit until all batch jobs have completed (by doing something like periodically polling the jobs it has spawned)?

And two questions regarding testing:

> Test suite: scripts_regression_tests --fast

(2) It seems like the full scripts_regression_tests should be run, without '--fast', right?

(3) I appreciate that you have added an automated test of this feature. However, I don't love seeing test-specific code in the production code (it makes it harder to understand both the tests and the production code, increases the chances that test logic would accidentally pollute the production environment, etc.). Would it be feasible to do these tests by introducing new fake test classes that fail the first time (either in the build or run) but pass the second time? I don't think it would work to store state in the object, but could you do this with something like this?:

```python
class TESTBUILDFAIL_THEN_PASS:

    def build_phase(self, sharedlib_only=False, model_only=False):
        already_run_filename = os.path.join(self._case.get_value("CASEROOT"), "already_run")
        if os.path.isfile(already_run_filename):
            # Do whatever is needed to pass the test
        else:
            open(already_run_filename, 'a').close()
            # Do whatever is needed to fail the test
```

I'd imagine having three classes like that:

(a) One that fails the build the first time, passes on subsequent runs

(b) One that fails the run the first time, passes on subsequent runs

(c) One that PASSES the run the first time, FAILS on subsequent runs.

(a) and (b) reimplement what you already have, but note that (c) goes beyond the tests you already have, and I feel this would be a helpful addition to ensure that already-passed tests are NOT rerun.

I don't feel super-strongly about the need to reimplement the tests this way, though I would like to see the addition of something like (c) using some mechanism.

gold2718:
While I think having the new test version would be good, I do not feel it is a high priority item. My usual workaround is to move the failed test so that a new test is created. Can that work with test suites in this system?

jgfouca:
@billsacks 

> What are the implications of setting --wait?

It means create_test will wait until all tests have gone through the batch system. This was necessary in order to support retrying tests that had RUN failures.

> It seems like the full scripts_regression_tests should be run, without '--fast'

I think only a code-check and testing retry is necessary. There are plenty of tests in the --fast suite to make sure that normal create_test usage isn't broken.

> However, I don't love seeing test-specific code in the production code

Yeah, I didn't like that either. But it was sooo much simpler than doing it another way.

jgfouca:
@gold2718 , I'm not sure I understand the question. If you turn on retry, you won't have to move anything.

billsacks:
Thanks for the replies, @jgfouca 

>> However, I don't love seeing test-specific code in the production code

> Yeah, I didn't like that either. But it was sooo much simpler than doing it another way.

I'm willing to hold my nose and accept that.

The one thing I would like to see added before accepting this PR is a test of:

> (c) One that PASSES the run the first time, FAILS on subsequent runs... to ensure that already-passed tests are NOT rerun

Since the `--retry` mechanism mostly just leverages the `--use-existing` mechanism, I don't care whether this test is done for `--retry` or as part of `O_TestTestScheduler`:`test_c_use_existing`.

jgfouca:
@billsacks , that's a good idea.

gold2718:
@jgfouca: Sorry, I do the move so I can compare the two runs. Sometimes, two failures still points at a system problem so I like to keep the first failed run around. It is not (IMHO) a requirement for CIME.

jgfouca:
@gold2718 , oh, I see. The current approach will preserve the log info from the original failed run, so that info is not lost.

gold2718:
Well, if you insist on making my life easier, I suppose I can adapt.

jgfouca:
@billsacks , I added the test you asked for.

billsacks:
Okay, I'm happy with this now. Thanks again, @jgfouca !

PR #2035, opened 2017-11-08
Improve email handling

### LINKED ISSUE #2004: Return default behavior of emailing user upon job completion/abort
As @mnlevy1981 noticed: Sometime in the last few months - probably in #1719 from @erichlf - the default behavior for emails at the end of a job changed so that, by default, you no longer receive emails. I don't remember any discussion of this change, and I feel the default behavior should revert back to how it was a few months ago, which was to send emails upon either a successful or unsuccessful completion of the job.

In addition to reverting that behavior, I feel it's important that we avoid hidden behavior changes like this in the future: These sorts of changes should be accompanied by discussion on github, ideally in an issue or PR whose title makes it clear what's going on, so others can notice the topic and reply if they're interested. "Add Support for Email Options" (for example) does not make it clear that the default behavior is changing.

jgfouca:
Improve email handling
    
1) Allow multiple event selections in case.submit
2) Remove hardcoded 'never' as mail_type
3) Allow batch systems to specify their own default
4) Allow mail settings to be configured by a users's cime_config

Test suite: by-hand on skybridge, code_checker
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2004 

User interface changes?: Yes. Allows multiple -m args to case.submit. Allows user to put email settings in .cime/config

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,27 @@
 ======================================================================
 
+Originator: Chris Fischer
+Date: 11-07-2017
+Tag: cime5.4.0-alpha.08
+Answer Changes: None
+Tests: Hand test B1850 
+Dependencies:
+
+Brief Summary:
+    - Handle case when cmdargs is not a list
+
+User interface changes: 
+
+PR summary: git log --oneline --first-parent [previous_tag]..master 
+fc72203 Merge pull request #2032 from jedwards4b/cmdargs_fix
+
+Modified files: git diff --name-status [previous_tag]
+M       scripts/lib/CIME/utils.py
+
+======================================================================
+
+======================================================================
+
 Originator: Chris Fischer 
 Date: 11-06-2017
 Tag: cime5.4.0-alpha.07
--- a/config/acme/machines/config_batch.xml
+++ b/config/acme/machines/config_batch.xml
@@ -133,7 +133,6 @@
       <directive default="n"> -r {{ rerunnable }} </directive>
       <!-- <directive> -j oe {{ job_id }} </directive> -->
       <directive> -j oe </directive>
-      <directive default="ae"  > -m {{ mail_options }} </directive>
       <directive> -V </directive>
     </directives>
   </batch_system>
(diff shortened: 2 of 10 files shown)


billsacks:
Adding @mnlevy1981 rather than myself for review of this one. He said he's in the process of reviewing it but it may be a little while before he completes it because he wants to do testing on cheyenne (which is down until tomorrow).

mnlevy1981:
To summarize above:

#. `./case.submit -m end -m fail` needs to update `qsub` with `-m ea` rather than `-m e -m a` (the `-m e` is ignored in the latter)
#. `./case.submit -m end,fail` should be functionally equivalent to `./case.submit -m end -m fail` rather than producing an error about invalid options
#. When do the defaults get updated?

As @billsacks mentioned, cheyenne went down unexpectedly so I'll test this code when the machine is back up (hopefully tomorrow, because I'm taking PTO Friday and Monday).

jedwards4b:
@mnlevy1981 can't this be tested on hobart?

mnlevy1981:
@jedwards4b I was planning to test on both hobart and cheyenne; as noted above, hobart testing found a couple of issues but I don't know if those are also issues on cheyenne (or if there are issues on cheyenne that don't appear on hobart)

jgfouca:
@mnlevy1981 , thanks for the review. I agree, it would be nice to support -m end,fail for case.submit, but that means I can't leverage the choices feature of argparse. Maybe that's not the end of the world since I can check validity by hand.

Oh man, that's so annoying that PBS does not support multiple -m flags. That means it will have to work differently than slurm.

I will let CSEG/CESM decide what defaults they want in config_batch.xml. All they need to do is add something for batch_mail_default.

mnlevy1981:
@jgfouca if you use `action=store, nargs='+'` then you can pass space-separated options:

```
$ ./case.submit -m end fail
```

Stack overflow hasn't shown me how to replace the space separated values with comma-separated values, though.

mnlevy1981:
Judging from [this stray comment](https://bugs.python.org/issue9338#msg200858) in a Python bug ticket, (and the one that follows it)

> The splitting on space is done by the shell.

So I think the best I can hope for is

```
$ ./case.submit -m end fail
```

(and you need to make sure there are no positional arguments following `'-m', '--mailtype'` otherwise those arguments will be slurped into processing `-m`). Does it make sense to also update `.cime/config` to use space-delimiting for multiple options instead of comma-delimiting? So

```
[main]
MAILTYPE=end fail
```

Or are there too many options for which comma-separating is already ingrained?

jgfouca:
@mnlevy1981 , I think your concerns should be addressed now.

mnlevy1981:
@jgfouca Verified that on hobart:

1. default is still no email
2. `case.submit -m end,fail` works as expected
3. `MAILTYPE=end,fail` works as expected

I'd like to check those three points on cheyenne when it comes back up.

Question for everyone: I noticed that there is no `--mailtype` option for `create_test`. Should there be one, or is it okay to assume that users running `create_test` have set `~/.cime/config` accordingly / are happy with the default settings?

jedwards4b:
I think that we want the command line option for create_test as well, shouldn't be too difficult to add, but if a test is expected to fail (TESTRUNFAIL) should it send email?

mnlevy1981:
I'm okay with expected failures being treated as successes. One possible issue with adding `--mailtype` to `create_test` is that `-m` is already taken (short for `--machine`) so there will be some consistency issues to work out.

jgfouca:
OK, I will add flags to create_test and create_newcase.

jgfouca:
Actually, create_newcase shouldn't need this since the user is running case.submit themselves in that case anyway.

jgfouca:
OK, I think this is ready now.

mnlevy1981:
It looks like both `case.submit` and `create_test` will default to

```
[main]
MAILTYPE=...
```

Which means users can't specify ([for example](https://github.com/ESMCI/cime/issues/2004#issuecomment-340498588)) to have `create_test` email only tests that fail while stand-alone cases email upon failure or success. Would it be possible to implement either

```
[main]
MAILTYPE=...
MAILTYPE_TEST=...
```

or

```
[main]
MAILTYPE=...
[test]
MAILTYPE=...
```

for some appropriate value of `[test]` (I don't know what else is configurable in the `config`file).

jedwards4b:
@mnlevy1981 This 
```
[main]
MAILTYPE=...
[create_test]
MAILTYPE=...
```
is consistent with the configparser and should be easy to do.

jgfouca:
@mnlevy1981 , yes, that should be doable. Let me do that real quick.

mnlevy1981:
@jgfouca I think @jedwards4b beat you to it with e053fb1 and then 9084dd6

jgfouca:
@mnlevy1981 , you're right. You guys already did it :)

I just removed the conflict and changed MAILUSER to MAIL_USER to be consistent. I think this is ready now.

jgfouca:
@mnlevy1981 great, thanks for the good suggestions for this PR. I think this will be a very nice feature for our users. Maybe we should document it somewhere?

Issue #2031, opened 2017-11-06
Compare two doesn't work on when casedirs and rundirs are on separate file systems


jgfouca:
On titan, cases are put here:
~/acme_scratch/$PROJECT/$CASE

... and RUNDIRS are put here:
/lustre/atlas/proj-shared/$PROJECT/$USER/$CASE/run

SystemTestCompareTwo sets the OUTPUT_ROOT for case2 to ~/acme_scratch/$PROJECT/$CASE/case2, which means the RUNDIR for case2 is on the wrong file system.

billsacks:
Okay, I think I see the problem. My mind is working slowly today, though, so it may take me a little while to come up with a robust solution (there may be an easy one, I'm just not grasping it immediately). @jgfouca or @jedwards4b feel free to chime in if you see one quickly :-)

The problem is that we were assuming that machines always put the bld and run directories under CIME_OUTPUT_ROOT. I think this is always true for all CESM machines, but that's not the case on titan:

```xml
    <CIME_OUTPUT_ROOT>$ENV{HOME}/acme_scratch/$PROJECT</CIME_OUTPUT_ROOT>
    <RUNDIR>$ENV{PROJWORK}/$PROJECT/$USER/$CASE/run</RUNDIR>
    <EXEROOT>$CIME_OUTPUT_ROOT/$CASE/bld</EXEROOT>
```

I think the setting of OUTPUT_ROOT for case2 is done correctly:

```python
        output_root2 = os.path.join(self._case1.get_value("CIME_OUTPUT_ROOT"),
                                    self._case1.get_value("CASE"), "case2")
```

but this is wrong:

```python
    def _get_case2_rundir(self):
        """
        Gets rundir for case2.
        """
        # Put the case2 run directory alongside its bld directory for
        # consistency. (See notes about EXEROOT in _get_case2_exeroot.)
        case2_rundir = os.path.join(self._get_output_root2(), "run")
        return case2_rundir
```

The problem is: Given that case1 and case2 have the same case name, and given that we want to keep their run directories separate (not critical, but highly desired): How do we specify the RUNDIR for case2 in a general, robust way? We figured out how to do this if we can assume that bld and run are nested under CIME_OUTPUT_ROOT, but I'm struggling with how to do this if we can't assume that.

jedwards4b:
How about putting the case2 run dir under the case1 run directory:
case2_rundir = os.path.join(case1_rundir, "case2run")

jgfouca:
[...]

PR #2036, opened 2017-11-08
Change paths to case2 rundir and exeroot

billsacks:
Previously, the code assumed that rundir and exeroot were subdirectories
of CIME_OUTPUT_ROOT. This is the case for CESM machines, but not for all
ACME machines. The new version doesn't make this assumption.

Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#2031

User interface changes?: none

Code review:

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -282,9 +282,13 @@ class SystemTestsCompareTwo(SystemTestsCommon):
 
         Assumes that self._case1 is already set to point to the case1 object
         """
-        # Since case 2 has the same name as case1 its CIME_OUTPUT_ROOT must also be different
+        # Since case2 has the same name as case1, its CIME_OUTPUT_ROOT
+        # must also be different, so that anything put in
+        # $CIME_OUTPUT_ROOT/$CASE/ is not accidentally shared between
+        # case1 and case2. (Currently nothing is placed here, but this
+        # helps prevent future problems.)
         output_root2 = os.path.join(self._case1.get_value("CIME_OUTPUT_ROOT"),
-                                    self._case1.get_value("CASE"), "case2")
+                                    self._case1.get_value("CASE"), "case2_output_root")
         return output_root2
 
     def _get_case2_exeroot(self):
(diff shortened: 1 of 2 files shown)


billsacks:
I kicked off testing last night, but the tests hung (probably a system issue). I'm redoing tests now, but wanted to go ahead and open this for review. I'll check the tests a little later this morning and post results.

scripts/lib/CIME/tests/case_fake.py
```diff
@@ -97,9 +98,11 @@ def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None,
         os.makedirs(newcaseroot)
         clone = self.copy(newcasename = newcasename, newcaseroot = newcaseroot)
         if cime_output_root is not None:
-            self.set_value('CIME_OUTPUT_ROOT', cime_output_root)
+            clone.set_value('CIME_OUTPUT_ROOT', cime_output_root)
+        if exeroot is not None:
+            clone.set_value('EXEROOT', exeroot)
         if rundir is not None:
-            self.set_value('RUNDIR', rundir)
+            clone.set_value('RUNDIR', rundir)
```
billsacks: Change from self.set_value to clone.set_value was a mostly unrelated bug fix that just affects unit tests (and not in any substantive way).

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -282,9 +282,13 @@ def _get_output_root2(self):
...
         """
-        # Since case 2 has the same name as case1 its CIME_OUTPUT_ROOT must also be different
+        # Since case2 has the same name as case1, its CIME_OUTPUT_ROOT
+        # must also be different, so that anything put in
+        # $CIME_OUTPUT_ROOT/$CASE/ is not accidentally shared between
+        # case1 and case2. (Currently nothing is placed here, but this
+        # helps prevent future problems.)
         output_root2 = os.path.join(self._case1.get_value("CIME_OUTPUT_ROOT"),
-                                    self._case1.get_value("CASE"), "case2")
+                                    self._case1.get_value("CASE"), "case2_output_root")
```
billsacks: I debated what to do about CIME_OUTPUT_ROOT. Now that we set RUNDIR and EXEROOT separately, nothing is placed here. Yet this directory is still created by the scripts. So we get a directory like `/glade/scratch/sacks/TESTNAME/case2_output_root` with nothing in it. I found this a little confusing. On the other hand, I'm nervous about leaving case2's CIME_OUTPUT_ROOT at its default value, because that opens the door for things from case2 to clobber things from case1 (in case anything is put in `$CIME_OUTPUT_ROOT/$CASE/` - since that location will be the same for case1 and case2).

As you'll see below, the solution I came up with was adding a README file to this location, so that if someone stumbles upon this directory, they'll be pointed to where they likely really want to go. But I'm open to suggestions if you feel there's a better solution.

billsacks:
I was having system issues with the testing, but now have run scripts_regression_tests on hobart and all tests pass.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
jgfouca: @billsacks , case2 will involve 3 different directions, it's RUNDIR, EXEROOT (blddir), and it's case dir. I would expect the case dir to be under the case2 output root. Is this not right?

billsacks: @jgfouca No, that's not right. case2's case directory gets nested under case1's case directory:

```python
        caseroot2 = os.path.join(caseroot1, "case2", casename2)
```

which, in general, bears no relationship to either the case1 or case2 CIME_OUTPUT_ROOT.

It's confusing, I know.

jgfouca: OK, thanks.

PR #2039, opened 2017-11-09
Data assimilation to fortran try1

### LINKED ISSUE #1980: Importing DO_DATA_ASSIMILATION into components

jedwards4b:
Separate DATA_ASSIMILATION flag by component and add to Fortran infodata.
Components can now query infodata for the flag data_assimilation_xxx where xxx is one of
(cpl, atm, ocn, wav, glc, ice, rof, lnd)
Removes variable PAUSE_COMPONENT_LIST
Note that while this brings the appropriate signals into CIME, no components yet test for their presence at the appropriate times. These tests will be part of future PRs.

Supersedes #2028 
Test baseline: NA
Test namelist changes: XML variable PAUSE_COMPONENT_LIST eliminated
        namelist variable pause_component_list eliminated
        XML variables DATA_ASSIMILATION_XXX added for XXX = [ CPL, ATM, OCN, WAV, GLC, ICE, ROF, LND ]
        namelist variables data_assimilation_xxx added for xxx = [ cpl, atm, ocn, wav, glc, ice, rof, lnd ]
Test status: bit for bit

Partially fixes #1980
This PR will close #1980 and new issues to address the remaining requests.

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @gold2718

### DIFF
--- a/scripts/lib/CIME/case_submit.py
+++ b/scripts/lib/CIME/case_submit.py
@@ -123,7 +123,7 @@ def check_case(case):
     logger.info("Check case OK")
 
 def check_DA_settings(case):
-    if case.get_value("DATA_ASSIMILATION"):
+    if any(case.get_values("DATA_ASSIMILATION")):
         script = case.get_value("DATA_ASSIMILATION_SCRIPT")
         cycles = case.get_value("DATA_ASSIMILATION_CYCLES")
         logger.info("Data Assimilation enabled using script {} with {:d} cycles".format(script,cycles))
--- a/scripts/lib/CIME/case_submit.py
+++ b/scripts/lib/CIME/case_submit.py
@@ -123,7 +123,7 @@ def check_case(case):
     logger.info("Check case OK")
 
 def check_DA_settings(case):
-    if any(case.get_values("DATA_ASSIMILATION")):
-        script = case.get_value("DATA_ASSIMILATION_SCRIPT")
-        cycles = case.get_value("DATA_ASSIMILATION_CYCLES")
+    script = case.get_value("DATA_ASSIMILATION_SCRIPT")
+    cycles = case.get_value("DATA_ASSIMILATION_CYCLES")
+    if len(script) > 0 and os.path.isfile(script) and cycles > 0:
         logger.info("Data Assimilation enabled using script {} with {:d} cycles".format(script,cycles))
--- a/src/drivers/mct/cime_config/namelist_definition_drv.xml
+++ b/src/drivers/mct/cime_config/namelist_definition_drv.xml
@@ -1643,13 +1643,13 @@
     <category>time</category>
     <group>seq_timemgr_inparm</group>
     <desc>
-      true => ESP component runs after driver 'pause cycle'
-      If any component 'pauses' (see PAUSE_OPTION, PAUSE_N and
-      PAUSE_COMPONENT_LIST XML variables), the ESP component (if
-      present) will be run to process the component 'pause' (restart)
-      files and set any required 'resume' signals.
-      If true, esp_cpl_dt and esp_cpl_offset settings are ignored.
-      default: true
+      true => ESP component runs after driver 'pause cycle' If any
+      component 'pauses' (see DATA_ASSIMILATION_PAUSE_OPTION,
+      DATA_ASSIMILATION_PAUSE_N and DATA_ASSIMILATION_XXX XML
+      variables), the ESP component (if present) will be run to
+      process the component 'pause' (restart) files and set any
+      required 'resume' signals.  If true, esp_cpl_dt and
+      esp_cpl_offset settings are ignored.  default: true
     </desc>
     <values>
       <value>.true.</value>
@@ -2024,52 +2024,38 @@
     </values>
   </entry>
 
-  <entry id="pause_option" modify_via_xml="PAUSE_OPTION">
+  <entry id="data_assimilation_pause_option" modify_via_xml="DATA_ASSIMILATION_PAUSE_OPTION">
     <type>char</type>
-    <category>time</category>
+    <category>data_assimilation</category>
     <group>seq_timemgr_inparm</group>
     <valid_values>none,never,nsteps,nstep,nseconds,nsecond,nminutes,nminute,nhours,nhour,ndays,nday,monthly,nmonths,nmonth,nyears,nyear</valid_values>
     <desc>
-      sets the pause frequency with pause_n
-      pause_option alarms are:
+      sets the pause frequency with data_assimilation_pause_n
+      data_assimilation_pause_option alarms are:
       [none/never], turns option off
-      [nstep/s]   , pauses every pause_n nsteps  , relative to start or last pause time
-      [nsecond/s] , pauses every pause_n nseconds, relative to start or last pause time
-      [nminute/s] , pauses every pause_n nminutes, relative to start or last pause time
-      [nhour/s]   , pauses every pause_n nhours  , relative to start or last pause time
-      [nday/s]    , pauses every pause_n ndays   , relative to start or last pause time
-      [nmonth/s]  , pauses every pause_n nmonths , relative to start or last pause time
+      [nstep/s]   , pauses every data_assimilation_pause_n nsteps  , relative to start or last pause time
+      [nsecond/s] , pauses every data_assimilation_pause_n nseconds, relative to start or last pause time
+      [nminute/s] , pauses every data_assimilation_pause_n nminutes, relative to start or last pause time
+      [nhour/s]   , pauses every data_assimilation_pause_n nhours  , relative to start or last pause time
+      [nday/s]    , pauses every data_assimilation_pause_n ndays   , relative to start or last pause time
+      [nmonth/s]  , pauses every data_assimilation_pause_n nmonths , relative to start or last pause time
       [monthly/s] , pauses every        month    , relative to start or last pause time
-      [nyear/s]   , pauses every pause_n nyears  , relative to start or last pause time
+      [nyear/s]   , pauses every data_assimilation_pause_n nyears  , relative to start or last pause time
     </desc>
     <values>
-      <value>$PAUSE_OPTION</value>
+      <value>$DATA_ASSIMILATION_PAUSE_OPTION</value>
     </values>
   </entry>
 
-  <entry id="pause_n" modify_via_xml="PAUSE_N">
+  <entry id="data_assimilation_pause_n" modify_via_xml="DATA_ASSIMILATION_PAUSE_N">
     <type>integer</type>
-    <category>time</category>
+    <category>data_assimilation</category>
     <group>seq_timemgr_inparm</group>
     <desc>
-      Sets the pause frequency with pause_option
+      Sets the pause frequency with data_assimilation_pause_option
     </desc>
     <values>
-      <value>$PAUSE_N</value>
-    </values>
-  </entry>
-
-  <entry id="pause_component_list" modify_via_xml="PAUSE_COMPONENT_LIST" skip_default_entry="true">
-    <type>char</type>
-    <category>time</category>
-    <group>seq_timemgr_inparm</group>
-    <desc>
-      A colon-separated list of component types (e.g., ocn,atm) to receive a
-      resume signal after a pause has happened. Special values "all" or "none"
-      may be used in place of a list
-    </desc>
-    <values>
-      <value>$PAUSE_COMPONENT_LIST</value>
+      <value>$DATA_ASSIMILATION_PAUSE_N</value>
     </values>
   </entry>
 
@@ -3937,7 +3923,7 @@
   <entry id="data_assimilation_atm" modify_via_xml="DATA_ASSIMILATION_ATM">
     <type>logical</type>
     <category>data_assimilation</category>
-    <group>seq_infodata_inparm</group>
+    <group>seq_timemgr_inparm</group>
     <desc>
       Whether Data Assimilation is on for component atm
     </desc>
@@ -3946,10 +3932,22 @@
     </values>
   </entry>
 
+  <entry id="data_assimilation_cpl" modify_via_xml="DATA_ASSIMILATION_CPL">
+    <type>logical</type>
+    <category>data_assimilation</category>
+    <group>seq_timemgr_inparm</group>
+    <desc>
+      Whether Data Assimilation is on for component CPL
+    </desc>
+    <values>
+      <value>$DATA_ASSIMILATION_CPL</value>
+    </values>
+  </entry>
+
   <entry id="data_assimilation_ocn" modify_via_xml="DATA_ASSIMILATION_OCN">
     <type>logical</type>
     <category>data_assimilation</category>
-    <group>seq_infodata_inparm</group>
+    <group>seq_timemgr_inparm</group>
     <desc>
       Whether Data Assimilation is on for component ocn
     </desc>
@@ -3961,7 +3959,7 @@
   <entry id="data_assimilation_wav" modify_via_xml="DATA_ASSIMILATION_WAV">
     <type>logical</type>
     <category>data_assimilation</category>
-    <group>seq_infodata_inparm</group>
+    <group>seq_timemgr_inparm</group>
     <desc>
       Whether Data Assimilation is on for component wav
     </desc>
@@ -3973,7 +3971,7 @@
   <entry id="data_assimilation_glc" modify_via_xml="DATA_ASSIMILATION_GLC">
     <type>logical</type>
     <category>data_assimilation</category>
-    <group>seq_infodata_inparm</group>
+    <group>seq_timemgr_inparm</group>
     <desc>
       Whether Data Assimilation is on for component glc
     </desc>
@@ -3985,7 +3983,7 @@
   <entry id="data_assimilation_rof" modify_via_xml="DATA_ASSIMILATION_ROF">
     <type>logical</type>
     <category>data_assimilation</category>
-    <group>seq_infodata_inparm</group>
+    <group>seq_timemgr_inparm</group>
     <desc>
       Whether Data Assimilation is on for component rof
     </desc>
@@ -3997,7 +3995,7 @@
   <entry id="data_assimilation_ice" modify_via_xml="DATA_ASSIMILATION_ICE">
     <type>logical</type>
     <category>data_assimilation</category>
-    <group>seq_infodata_inparm</group>
+    <group>seq_timemgr_inparm</group>
     <desc>
       Whether Data Assimilation is on for component ice
     </desc>
@@ -4009,7 +4007,7 @@
   <entry id="data_assimilation_lnd" modify_via_xml="DATA_ASSIMILATION_LND">
     <type>logical</type>
     <category>data_assimilation</category>
-    <group>seq_infodata_inparm</group>
+    <group>seq_timemgr_inparm</group>
     <desc>
       Whether Data Assimilation is on for component lnd
     </desc>
--- a/src/drivers/mct/cime_config/namelist_definition_drv.xml
+++ b/src/drivers/mct/cime_config/namelist_definition_drv.xml
@@ -3934,4 +3934,88 @@
     </values>
   </entry>
 
+  <entry id="data_assimilation_atm" modify_via_xml="DATA_ASSIMILATION_ATM">
+    <type>logical</type>
+    <category>data_assimilation</category>
+    <group>seq_infodata_inparm</group>
+    <desc>
+      Whether Data Assimilation is on for component atm
+    </desc>
+    <values>
+      <value>$DATA_ASSIMILATION_ATM</value>
+    </values>
+  </entry>
+
+  <entry id="data_assimilation_ocn" modify_via_xml="DATA_ASSIMILATION_OCN">
+    <type>logical</type>
+    <category>data_assimilation</category>
+    <group>seq_infodata_inparm</group>
+    <desc>
+      Whether Data Assimilation is on for component ocn
+    </desc>
+    <values>
+      <value>$DATA_ASSIMILATION_OCN</value>
+    </values>
+  </entry>
+
+  <entry id="data_assimilation_wav" modify_via_xml="DATA_ASSIMILATION_WAV">
+    <type>logical</type>
+    <category>data_assimilation</category>
+    <group>seq_infodata_inparm</group>
+    <desc>
+      Whether Data Assimilation is on for component wav
+    </desc>
+    <values>
+      <value>$DATA_ASSIMILATION_WAV</value>
+    </values>
+  </entry>
+
+  <entry id="data_assimilation_glc" modify_via_xml="DATA_ASSIMILATION_GLC">
+    <type>logical</type>
+    <category>data_assimilation</category>
+    <group>seq_infodata_inparm</group>
+    <desc>
+      Whether Data Assimilation is on for component glc
+    </desc>
+    <values>
+      <value>$DATA_ASSIMILATION_GLC</value>
+    </values>
+  </entry>
+
+  <entry id="data_assimilation_rof" modify_via_xml="DATA_ASSIMILATION_ROF">
+    <type>logical</type>
+    <category>data_assimilation</category>
+    <group>seq_infodata_inparm</group>
+    <desc>
+      Whether Data Assimilation is on for component rof
+    </desc>
+    <values>
+      <value>$DATA_ASSIMILATION_ROF</value>
+    </values>
+  </entry>
+
+  <entry id="data_assimilation_ice" modify_via_xml="DATA_ASSIMILATION_ICE">
+    <type>logical</type>
+    <category>data_assimilation</category>
+    <group>seq_infodata_inparm</group>
+    <desc>
+      Whether Data Assimilation is on for component ice
+    </desc>
+    <values>
+      <value>$DATA_ASSIMILATION_ICE</value>
+    </values>
+  </entry>
+
+  <entry id="data_assimilation_lnd" modify_via_xml="DATA_ASSIMILATION_LND">
+    <type>logical</type>
+    <category>data_assimilation</category>
+    <group>seq_infodata_inparm</group>
+    <desc>
+      Whether Data Assimilation is on for component lnd
+    </desc>
+    <values>
+      <value>$DATA_ASSIMILATION_LND</value>
+    </values>
+  </entry>
+
 </entry_id>
--- a/src/drivers/mct/cime_config/namelist_definition_drv.xml
+++ b/src/drivers/mct/cime_config/namelist_definition_drv.xml
@@ -1644,8 +1644,8 @@
     <group>seq_timemgr_inparm</group>
     <desc>
       true => ESP component runs after driver 'pause cycle' If any
-      component 'pauses' (see DATA_ASSIMILATION_PAUSE_OPTION,
-      DATA_ASSIMILATION_PAUSE_N and DATA_ASSIMILATION_XXX XML
+      component 'pauses' (see PAUSE_OPTION,
+      PAUSE_N and DATA_ASSIMILATION_XXX XML
       variables), the ESP component (if present) will be run to
       process the component 'pause' (restart) files and set any
       required 'resume' signals.  If true, esp_cpl_dt and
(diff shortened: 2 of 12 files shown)


src/drivers/mct/cime_config/namelist_definition_drv.xml
```diff
@@ -2024,52 +2024,38 @@
     </values>
   </entry>
 
-  <entry id="pause_option" modify_via_xml="PAUSE_OPTION">
+  <entry id="data_assimilation_pause_option" modify_via_xml="DATA_ASSIMILATION_PAUSE_OPTION">
     <type>char</type>
-    <category>time</category>
+    <category>data_assimilation</category>
```
gold2718: Since I don't feel that PAUSE is only for data assimilation, I don't think this is the best place for this entry id.

src/drivers/mct/cime_config/namelist_definition_drv.xml
```diff
@@ -2024,52 +2024,38 @@
...
-    <category>time</category>
-    <group>seq_timemgr_inparm</group>
-    <desc>
-      A colon-separated list of component types (e.g., ocn,atm) to receive a
-      resume signal after a pause has happened. Special values "all" or "none"
-      may be used in place of a list
-    </desc>
-    <values>
-      <value>$PAUSE_COMPONENT_LIST</value>
+      <value>$DATA_ASSIMILATION_PAUSE_N</value>
```
gold2718: Please change back to pause_option / PAUSE_OPTION and pause_n / PAUSE_N.

src/drivers/mct/cime_config/namelist_definition_drv.xml
```diff
@@ -2024,52 +2024,38 @@
...
-      <value>$PAUSE_OPTION</value>
+      <value>$DATA_ASSIMILATION_PAUSE_OPTION</value>
     </values>
   </entry>
 
-  <entry id="pause_n" modify_via_xml="PAUSE_N">
+  <entry id="data_assimilation_pause_n" modify_via_xml="DATA_ASSIMILATION_PAUSE_N">
     <type>integer</type>
-    <category>time</category>
+    <category>data_assimilation</category>
```
gold2718: Since I don't feel that PAUSE is only for data assimilation, I don't think this is the best place for this entry id.

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -304,9 +304,18 @@ subroutine seq_timemgr_clockInit(SyncClock, nmlfile, restart, restart_file, pioi
     character(SHR_KIND_CS)  :: restart_option        ! Restart option units
     integer(SHR_KIND_IN)    :: restart_n             ! Number until restart interval
     integer(SHR_KIND_IN)    :: restart_ymd           ! Restart date (YYYYMMDD)
-    character(SHR_KIND_CS)  :: pause_option          ! Pause option units
-    integer(SHR_KIND_IN)    :: pause_n               ! Number between pause intervals
-    character(SHR_KIND_CS)  :: pause_component_list  ! Pause - resume components
+    character(SHR_KIND_CS)  :: data_assimilation_pause_option          ! Pause option units
+    integer(SHR_KIND_IN)    :: data_assimilation_pause_n               ! Number between pause intervals
```
gold2718: Please change back to pause_option and pause_n

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -541,9 +567,10 @@ subroutine seq_timemgr_clockInit(SyncClock, nmlfile, restart, restart_file, pioi
        write(logunit,F0I) trim(subname),' restart_n      = ',restart_n
        write(logunit,F0I) trim(subname),' restart_ymd    = ',restart_ymd
        write(logunit,F0L) trim(subname),' end_restart    = ',end_restart
-       write(logunit,F0A) trim(subname),' pause_option   = ',trim(pause_option)
-       write(logunit,F0I) trim(subname),' pause_n        = ',pause_n
-       write(logunit,F0A) trim(subname),' pause_component_list = ',trim(pause_component_list)
+       write(logunit,F0A) trim(subname),' data_assimilation_pause_option   = ',&
+            trim(data_assimilation_pause_option)
+       write(logunit,F0I) trim(subname),' data_assimilation_pause_n        = ',&
+            data_assimilation_pause_n
```
gold2718: Please change back to pause_option and pause_n

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -630,9 +657,17 @@ subroutine seq_timemgr_clockInit(SyncClock, nmlfile, restart, restart_file, pioi
     call shr_mpi_bcast( restart_n,            mpicom )
     call shr_mpi_bcast( restart_option,       mpicom )
     call shr_mpi_bcast( restart_ymd,          mpicom )
-    call shr_mpi_bcast( pause_n,              mpicom )
-    call shr_mpi_bcast( pause_option,         mpicom )
-    call shr_mpi_bcast( pause_component_list, mpicom )
+    call shr_mpi_bcast( data_assimilation_pause_n,              mpicom )
+    call shr_mpi_bcast( data_assimilation_pause_option,         mpicom )
```
gold2718: Please change back to pause_option and pause_n

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -701,48 +736,34 @@ subroutine seq_timemgr_clockInit(SyncClock, nmlfile, restart, restart_file, pioi
...
+    endif
+    if (data_assimilation_lnd) then
+       pause_active(seq_timemgr_nclock_lnd) = .true.
+    endif
+
     if ( ANY(pause_active) .and.                                              &
-         (trim(pause_option) /= seq_timemgr_optNONE)  .and.                   &
-         (trim(pause_option) /= seq_timemgr_optNever)) then
+         (trim(data_assimilation_pause_option) /= seq_timemgr_optNONE)  .and.                   &
+         (trim(data_assimilation_pause_option) /= seq_timemgr_optNever)) then
```
gold2718: Please change back to pause_option.

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -896,8 +917,8 @@ subroutine seq_timemgr_clockInit(SyncClock, nmlfile, restart, restart_file, pioi
        if (pause_active(n)) then
          call seq_timemgr_alarmInit(SyncClock%ECP(n)%EClock,                  &
               EAlarm  = SyncClock%EAlarm(n,seq_timemgr_nalarm_pause),         &
-              option  = pause_option,                                         &
-              opt_n   = pause_n,                                              &
+              option  = data_assimilation_pause_option,                                         &
+              opt_n   = data_assimilation_pause_n,                                              &
```
gold2718: Please change back to pause_option and pause_n.

jedwards4b:
@gold2718 sure seems like we had a conversation and agreed that these variables should be renamed, I guess I don't mind undoing it but geeze.

gold2718:
@jedwards4b, apparently, we suck at communication (or maybe it's just me).
We definitely agreed on renaming the resume signalling by using the new `DATA_ASSILIMATION_XXX` variables. I do not recall addressing `PAUSE_OPTION` and `PAUSE_N`.
If you recall differently, please remind me why we wanted to rename these two.

scripts/lib/CIME/case_submit.py
```diff
@@ -123,7 +123,7 @@ def check_case(case):
     logger.info("Check case OK")
 
 def check_DA_settings(case):
-    if case.get_value("DATA_ASSIMILATION"):
-        script = case.get_value("DATA_ASSIMILATION_SCRIPT")
-        cycles = case.get_value("DATA_ASSIMILATION_CYCLES")
+    script = case.get_value("DATA_ASSIMILATION_SCRIPT")
+    cycles = case.get_value("DATA_ASSIMILATION_CYCLES")
+    if len(script) > 0 and os.path.isfile(script) and cycles > 0:
```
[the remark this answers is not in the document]
jedwards4b: It does.

src/drivers/mct/cime_config/namelist_definition_drv.xml
```diff
@@ -2026,7 +2026,7 @@
 
   <entry id="pause_option" modify_via_xml="PAUSE_OPTION">
     <type>char</type>
-    <category>time</category>
+    <category>data_assimilation</category>
```
gold2718: Since PAUSE is not data assimilation specific, can't we put this in some other category?

src/drivers/mct/cime_config/namelist_definition_drv.xml
```diff
@@ -2049,7 +2049,7 @@
 
   <entry id="pause_n" modify_via_xml="PAUSE_N">
     <type>integer</type>
-    <category>time</category>
+    <category>data_assimilation</category>
```
gold2718: Since PAUSE is not data assimilation specific, can't we put this in some other category?

jedwards4b:
The dae test already does this.

gold2718:
Sorry for being thick but can you point out how this is done? I see that you set `DATA_ASSIMILATION_ATM` in dae.py but I do not see any check that the appropriate signal actually reaches the atmosphere model at the appropriate time (which in this case is at initialization but not during subsequent run steps).
Am I missing a functional test for this?

gold2718:
If there is no functionality test yet, we need to adjust the description of what this PR does and make sure there is an open issue (or issues) for what comes next.
My preference is to modify the description of what this PR does and does not do and open a new issue (or issues) which will be closed as we approach the full need described in #1980.
In particular, this PR imports the data assimilation variables into CIME but not yet into any components which is the request in #1980. We need to work with @kdraeder to make sure that the new issues are focused, clearly state the needs, and are modular enough to each fit into a single PR (#1980 is too broad IMHO).
Any objection to that?

jedwards4b:
No objection here.

kdraeder:
I'm happy to leave it to @jedwards4b  and @gold2718 to organize 
the sub-tasks of #1980.  Thanks!

The CAM sourcemods I provided do use a logical variable to control
the flow in the context of the data assimilation capability.  That variable 
is hard-coded to .true., so that needs to be changed to get the value
from the namelist variable data_assimiliation_atm.  But after that, it sounds
like it should be possible to test these xml and namelist variables in the
CAM context.  I have a test case to do that after cheyenne returns to
service (or do you know of a way to make beta07 run on yellowstone?).

jedwards4b:
beta07 works fine on yellowstone

kdraeder:
That's not my experience yet.  It's helpful to know that 'user error' is
the likely cause.
Here's the error I'm getting, in case you recognize it easily:

Before case.setup modules are
Currently Loaded Modules:
  1) ncarenv/1.0          5) netcdf/4.3.0        9)  ncview/2.1.1
  2) ncarbinlibs/1.1      6) job_memusage/1.0    10) nco/4.6.9
  3) intel/12.1.5         7) perlmods/5.0        11) git/2.3.0
  4) ncarcompilers/1.0    8) python/2.7.7        12) matlab/R2015b
ERROR: module command
/glade/u/apps/ch/opt/lmod/7.2.1/lmod/lmod/libexec/lmod
python purge   failed with message:
/glade/u/apps/ch/opt/lua/5.3.3/bin/lua: /lib64/libc.so.6: version
`GLIBC_2.14' not found
(required by /glade/u/apps/ch/opt/lua/5.3.3/bin/lua)
ERROR: Case could not be set up.

jedwards4b:
You are trying to run a case defined for cheyenne on yellowstone.

kdraeder:
I just figured that out too.  Sorry for the distraction.

jedwards4b:
@jgfouca  this is ready to merge

kdraeder:
@jedwards4b 
@gold2718 
@mvertens 
It looks like the next steps are for me to:
1) check out the/a version of cime which has the new data_assim structure.
   ?   Should I get the commit  53e1ddf, or the latest master?
   I hope to test this in cesm2_0_beta07, if that makes a difference.
2) Merge the remaining CAM and CLM SourceMods into my beta07.
3) Modify the Mods to use the new namelist variables.
4) Modify the DATA_ASSIMILATION_SCRIPT to xmlchange the assimilation variables
   for CAM and CLM.
5) Test it with and without assimilation.
6) Commit changes to CAM and CLM (hopefully not CIME).

Am I missing anything?

Should this be a new issue, or the final stage of this issue?

kdraeder:
@jedwards4b 
@gold2718 
I believe that I downloaded the ESMCI:master after the merge.
And I see data_assimilation_atm in namelist_definition_drv.xml
and seq_timemgr_mod.F90.
But I don't see it in seq_infodata_mod.F90,
which is confirmed by the inability of my
atm_comp_mct.F90:seq_infodata_GetData
to fetch the value of data_assimilation_atm.
Am I understanding how this should work?
Should I be figuring out how to include data_assimilation_xxx
into seq_infodata_mod.F90?
Thanks for any tips.
Kevin

kdraeder:
I added some data_assimilation sections to seq_infodata_mod.F90,
which enable it to run and finish the assimilation, but I may not have
added all the sections that should be added for completeness.
Should I test it and issue a pull request, or have someone look at it first? 
It's in
/glade/p/work/raeder/Models/cesm2_0_beta07_clm4_5_16_r259_dart/cime/src/drivers/mct/shr
Thanks,
Kevin

jedwards4b:
I am deferring to @gold2718 here because my original PR which included this variable in infodata was rejected by him and now I cannot remember his alternative suggestion.

gold2718:
Well to quote my response (above in this thread):

> If there is no functionality test yet, we need to adjust the description of what this PR does and make sure there is an open issue (or issues) for what comes next.
> My preference is to modify the description of what this PR does and does not do and open a new issue (or issues) which will be closed as we approach the full need described in #1980.
> In particular, this PR imports the data assimilation variables into CIME but not yet into any components which is the request in #1980. We need to work with @kdraeder to make sure that the new issues are focused, clearly state the needs, and are modular enough to each fit into a single PR (#1980 is too broad IMHO).
> Any objection to that?

There was no objection but I do not think it was done either (not assigning blame here as I approved it). The PR states that it will

> Separate DATA_ASSIMILATION flag by component and add to Fortran infodata,

however, as @kdraeder points out, this was not in fact done in this PR. There are resume signals which _could_ be sent to component models but at the moment, the driver is not participating in data assimilation and has no functionality to set any DA signals.

I think what is missing is that if (for instance), `DATA_ASSIMILATION_ATM` is `.TRUE.`, then `atm_resume` should be set at initialization Then, each model would be able to check their resume signal during initialization to see if they need to do something. At that point, they would consult the relevant namelist variable to, for example, perform post data assimilation processing.

The sequence would be something like:
In `seq_timemgr_clockInit` (or a new function if it is decided that this function if not appropriate), set `xxx_resume` iff `DATA_ASSIMILATION_XXX` is `.TRUE.`
Somewhere near the end of `cime_init`, set `xxx_resume` to `.FALSE.`

Is this okay? If so, an issue should be opened (as mentioned in the PR) requesting this.
If not, what is missing / wrong?

kdraeder:
I don't know the bigger context of pause-resume very well,
but this plan seems redundant or circular.  DATA_ASSIM determines
both atm_resume (in cpl) and data_assimilation_atm (in the component),
but then the component consults both atm_resume and data_assim,
which say the same thing.  Maybe there's temporarily hidden functionality;
when we want to assimilate in POP (every 24 hours) and CAM (every 6 hours)
will we pause both components every 6 hours?  Then POP would need to resume
without doing the things required by the assimilation 3 out of 4 times. 
In this case, pop_resume would not be iff DATA_ASSIMILATION_OCN.
Kevin

gold2718:
In the current scheme, the component does not see any `DATA_ASSIMILATION_ATM` signal from the coupler, only `atm_resume`.
The `atm_resume` (or `pop_resume`) signal merely means that there is a new restart file for the component to integrate into its state.
_How_ the component integrates the new file is up to the component and should be influenced by an appropriate namelist variable (e.g., `data_assimilation = 'kalman'`).
I do not see any redundancy there, am I missing something?

jedwards4b:
I think that @kdraeder wants to know if a model is running in data assimilation mode whether or not it is directed to read a new restart file at the current time.

gold2718:
> Maybe there's temporarily hidden functionality;
> when we want to assimilate in POP (every 24 hours) and CAM (every 6 hours)
> will we pause both components every 6 hours? Then POP would need to resume
> without doing the things required by the assimilation 3 out of 4 times.
> In this case, pop_resume would not be iff DATA_ASSIMILATION_OCN.

There are two issues here, one of which, I believe is just a misunderstanding.

1. The `pop_resume` signal will only be set if `DATA_ASSIMILATION_OCN` is `.TRUE.`. Currently, this flag is under control of the DA script so I do not see any problem.
2. However, when it becomes time to 'pause' the model (i.e., not the current approach), we may need to expand the semantics as there is currently only one 'pause clock' defined. This can easily be remedied if we decide on a new semantics. I do not believe this is a problem for the current approach as all the components have to restart so they must all drop a restart file.

gold2718:
> I think that @kdraeder wants to know if a model is running in data assimilation mode whether or not it is directed to read a new restart file at the current time.

My proposal is that this is controlled by a component namelist variable such as the example given above. Anything wrong with that?

kdraeder:
I outlined a case where DATA_ASSIMILATION_POP would be false,
say at hour 6 of some day, but POP was paused(?) so that CAM could
do an assimilation, so POP would need to resume.  Am I misunderstanding
whether POP would need to be paused in that situation?

gold2718:
@kdraeder,
`DATA_ASSIMILATION_OCN` (not `DATA_ASSIMILATION_POP`)  is only used with the current DA scheme where the entire model is 'paused' (i.e., halted) so every component will have to restart. However, if `DATA_ASSIMILATION_OCN` is `.FALSE.`, POP will not receive an `ocn_resume` signal and would restart normally.
However, if you are talking about a future version of data assimilation using pause/resume, I think we may need to expand the current namelist scheme to allow for greater user control over when individual components are paused (i.e., required to write a restart file) and when they receive a resume signal (which is when they would perform state update based on the DA file).
Can you please clarify whether your concern is addressed towards the current (CESM 2.0) version or a future pause/resume version?

kdraeder:
I like the first paragraph.  That will accomplish in CESM2 what
Alicia Karspeck has already developed outside of CESM2 via scripting.
It sounds to me like users will already have complete control
over which components take part in assimilation at which times.
Assuming that all components will be paused at the end of a forecast,
where 'pause' seems to mean 'write out a restart, and wait for instructions',
then DATA_ASSIMILATION_XXX can be set by the assimilation script.

At the beginning of the next forecast, each component will query 
its data_assimilation_xxx namelist variable (set to the value in DATA_ASSIMILATION_XXX).
Then they either restart (from unchanged restart file) or  
resume (from updated restart file and do extra work to accommodate 
the updated state).  

In the future, it would be great if OCN didn't need to pause at hours 
6, 12, and 18, when CAM does assimilations, But we don't need
that functionality.  It would just be a savings of the time
needed to write and read the OCN restart file when there is 
no OCN assimilation done.

gold2718:
I think we have a plan. Two quick notes:

> At the beginning of the next forecast, each component will query
> its data_assimilation_xxx namelist variable (set to the value in DATA_ASSIMILATION_XXX).

The `DATA_ASSIMILATION_XXX` variables do _not_ reach the components, they are only signals to the driver to tell components to perform a 'resume' operation. Each component needs to define a regular namelist variable which determines what it does when it gets an `xxx_resume` signal. The namelist variable needs to follow that component's naming conventions.

As for the future, we will figure out a way to properly configure pause/resume DA if (when?) the pause/resume project 'resumes'.

kdraeder:
I know that DATA_ASSIMILATION_XXX don't directly reach the components,
but they are used to define data_assimilation_xxx in the cpl namelist:
drv_in: 
&seq_timemgr_inparm
  data_assimilation_atm = .true.
which are passed to the components via infodata:
atm_comp_mct.F90:
       call seq_infodata_GetData( infodata,                                           &
            data_assimilation_atm=data_assimilation_atm)

Maybe they shouldn't be, but at some level this was agreed to
before Jim and I coded it.  And it works in the way I tested it.

gold2718:
@kdraeder,

I am really confused. Four days ago you said (above):

> I believe that I downloaded the ESMCI:master after the merge.
> And I see data_assimilation_atm in namelist_definition_drv.xml
> and seq_timemgr_mod.F90.
> But I don't see it in seq_infodata_mod.F90,
> which is confirmed by the inability of my
> atm_comp_mct.F90:seq_infodata_GetData

That comment spawned this whole thread. Now it seems you are saying the opposite. Which is it?
Is this new code you are talking about? I do not see it on ESMCI master.

kdraeder:
The next comment after that one is:

"I added some data_assimilation sections to seq_infodata_mod.F90,
which enable it to run and finish the assimilation, but I may not have
added all the sections that should be added for completeness.
Should I test it and issue a pull request, or have someone look at it first?
It's in
/glade/p/work/raeder/Models/cesm2_0_beta07_clm4_5_16_r259_dart/cime/src/drivers/mct/shr
Thanks,
Kevin"

No one answered my question about pushing my changes in seq_infodata_mod.F90
to Github and issuing a pull request, or taking a preliminary look.  We just launched
into the discussion.  So I have not pushed my changes to Github, and you can only
see them in the file I listed.  
Should I have put some sort of flags in my 2nd comment to draw attention to it?

gold2718:
@kdraeder,

> Should I have put some sort of flags in my 2nd comment to draw attention to it?

Yeah, flashing lights or something. Sorry, this is my bad for missing this post (multiple times). The problem with this approach is that we discussed it and rejected it in PR #2028 (which is why this PR replaced it). You can read that discussion but the TL;DR is that we decided that the driver should not be telling the components how to operate and took that sort of flag out of infodata.

kdraeder:
@gold2718 @jedwards4b 
Here's where we are, from my perspective.

1) There is (or could be) code in each component's fortran code
   that needs to be executed only when the state used to start the
   next forecast has been modified by data assimilation (or some other
   ESP operation).

2) The modified states are stored in CESM restart files,
   except for CAM, which uses initial files.
   All these files use standard CESM names.

3) There is an xml variable DATA_ASSIMILATION_$COMP for each component.
   These can be changed by DATA_ASSIMILATION_SCRIPT to tell the components,
   through some script and code path to be named,
   whether to execute DA related code before starting the next forecast.

4) There are resume_$comp variables in the driver(?), which tell the components
   how to start (not whether to start) the next forecast.
   I don't know how those are set.  If they can be set according to the values
   in DATA_ASSIMILATION_$COMP, then I think that we are virtually done.
   Steve's description leads me to believe that this is the case:
   "if DATA_ASSIMILATION_OCN is .FALSE., POP will not receive an 
    ocn_resume signal and would restart normally."
   In this case, we'll remove the variables data_assimilation_$comp
   from everywhere, and use resume_$comp instead.

   If not, then we need to figure out how the information in 
   DATA_ASSIMILATION_$COMP will be propagated into the components
   and not conflict with resume_$comp.  The existing data_assimilation_$comp
   do this, but maybe not in the way they should according to CESM
   architecture standards.

5) If we want the resume_$comp variables to handle more than 1 ESP
   component, then they may need to be a non-logical variables,
   so that they can tell the components *which* specialized code to
   execute.

jedwards4b:
@gold2718 @kdraeder I think that an in person discussion is needed to progress here.

kdraeder:
I need to leave at 2:30 for the dentist, but I'm available tomorrow until
3pm.
Kevin

gold2718:
I should be around tomorrow -- depends on how today's appt. goes . . .

All the same, I think @kdraeder's last post sums things up pretty well.

jedwards4b:
> There is an xml variable DATA_ASSIMILATION_$COMP for each component.
> These can be changed by DATA_ASSIMILATION_SCRIPT to tell the components,
> through some script and code path to be named,
> whether to execute DA related code before starting the next forecast.

This code path already exists - when the DA component (script) changes the value of 
DATA_ASSIMILATION_$COMP it is updated in the drivers namelist and used to set the
pause_$comp variable.    Note that only the driver namelist is updated between DA cycles and so 
if a component needs to have some changed input from one cycle to the next they can only get it from the modified input or restart files or from infodata.

> If we want the resume_$comp variables to handle more than 1 ESP
> component, then they may need to be a non-logical variables,
> so that they can tell the components which specialized code to
> execute.

I was given to understand that multiple ESP components would not be allowed.

kdraeder:
I was thinking of one ESP or another, not more than one simultaneously.

Kevin

gold2718:
In both cases, I think the component can have an active namelist variable. For the runs with no DA, they will not get the resume signal so action will be taken.

kdraeder:
I believe at our meeting on 12/6 that Jim and Steve reached agreement
about how DA instructions would propagate through CIME into the components.
Is that coding finished?
And the next step is to make some component(s) (CAM, CLM, CICE)
use the signal to execute DA-specific code.  This will be done as separate
issues, however each component development process does it.  
What's the status of those?
Kevin

kdraeder:
@gold2718 @jedwards4b 
I'm available to test CAM and/or CIME code in the DART context,
if it's ready for that.
Thanks,
Kevin

kdraeder:
@gold2718 @jedwards4b 
I'm available to test CAM and/or CIME code in the DART context,
if it's ready for that.
Thanks,
Kevin

gold2718:
@kdraeder, 
Sorry for the radio silence. I've been fighting some other fires but will try to get to this within the next few days. I believe I will have time to make the necessary CIME change but I do not believe anyone has been assigned to bring in the CAM source mods.

kdraeder:
@gold2718 
I was able implement a functioning set of upgrades into CAM on 2017-12-1,
but we decided that the mechanism I used was not optimal.  If you outline
the mechanism that should be used, I might be able to implement a first draft
of it and save you/someone some time  (and maybe get these changes into
the CESM2 release).

gold2718:
@kdraeder,
I have a question about the `DATA_ASSIMILATION_XXX` variables. Currently, they are boolean so you set them to `.TRUE.` when you want the `XXX` component to do post-DA initialization.
However, the signal sent to the `XXX` components are strings so there is an opportunity to send more than yes or no if that would be useful. Can you think of anything?

gold2718:
@jedwards4b,
I think I am still missing (or more likely forgetting) something. In the DAE test, you set `DATA_ASSIMILATION_ATM` to `"TRUE"` and then run the model. However, in a real DA run, `DATA_ASSIMILATION_ATM` would be `FALSE` for the first run and `TRUE` after running the DA filter on the atmosphere state.
I think this means we need to regenerate the driver namelist between runs. Does someone have to call `preview_namelist`? Who should do this, loop over `data_assimilation_cycles` in `case_run`?
Something else?

jedwards4b:
@gold2718 yes, that code is already in place - the driver buildnml (only) is called at each data assimilation cycle.

gold2718:
@jedwards4b, sorry to be thick but could you tell me where that call is?

jedwards4b:
https://github.com/ESMCI/cime/blob/master/scripts/lib/CIME/case_run.py#L20

gold2718:
Ah, nice and buried. Thanks!

PR #2050, opened 2017-11-10
add optional COSTPES_PER_NODE

### LINKED ISSUE #1993: Figure out how to handle COST*
Concerns have been raised about the accuracy of the current COST_PES case setting. There are proposals to use a COST_PER_NODE concept instead.

jedwards4b:
Add a variable for COST_PER_NODE this variable is optional and can be used for machines that make that value difficult to compute automatically

Test suite: scripts_regression_tests.py PFS.f09_g17.B1850.yellowstone_intel 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1993 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -572,6 +572,7 @@
     <SUPPORTED_BY>cseg</SUPPORTED_BY>
     <MAX_TASKS_PER_NODE>256</MAX_TASKS_PER_NODE>
     <MAX_MPITASKS_PER_NODE>64</MAX_MPITASKS_PER_NODE>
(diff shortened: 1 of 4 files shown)


rljacob:
What exactly is this supposed to represent?

jedwards4b:
It is the number of pes per node used in the model cost computation in the timing file.
On yellowstone we typically use 15 of 16 pes per node and on cori-knl we use 64 of 68.

rljacob:
Ok.  A better name might be COSTPES_PER_NODE since its used to calculate the cost and not the cost itself.

PR #2051, opened 2017-11-10
locked files needs to handle BUILD_COMPLETE better

### LINKED ISSUE #1971: problem rebuilding if case.build fails after a successful first build
I am encountering the following failure.
1) I invoke case.build successfully
2) I realize I want to change a file - and I modify a file (say in CLM)
3) I run case.build again - and there is a compilation failure
4) I fix the file so that I expect it will now compile - and I rerun case.build. I then get the following:
```
Setting build complete to False
ERROR:
ERROR env_build HAS CHANGED
  A manual clean of your obj directories is required
  You should execute the following:
    ./case.build --clean-all
```
I know that I don't want to run case.build --clean-all since everything should work except for the file that I had changed. It turns out that to fix this I need to run the following two steps:
```
rm LockedFiles/env_build.xml
./xmlchange BUILD_STATUS=0
```
This is clearly a problem that users would run into when making changes to a case.

jedwards4b:
If BUILD_COMPLETE is the only variable changed in env_build.xml it should not flag in check_lockedfiles

Test suite: scripts_regression_tests.py, hand testing - see test in issue #1971
Test baseline: 
Test namelist changes: 

Fixes #1971 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -1330,8 +1330,6 @@ class Case(object):
         gfile = GenericXML(infile=xmlfile)
         ftype = gfile.get_id()
 
-        self.flush(flushall=True)
-
         logger.warning("setting case file to {}".format(xmlfile))
         components = self.get_value("COMP_CLASSES")
         new_env_file = None
--- a/scripts/lib/CIME/check_lockedfiles.py
+++ b/scripts/lib/CIME/check_lockedfiles.py
@@ -111,10 +111,12 @@ def check_lockedfiles(case):
                 continue
             diffs = f1obj.compare_xml(f2obj)
             if diffs:
+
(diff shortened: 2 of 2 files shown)


scripts/lib/CIME/case.py
```diff
@@ -1330,8 +1330,6 @@ def set_file(self, xmlfile):
         gfile = GenericXML(infile=xmlfile)
         ftype = gfile.get_id()
```
jedwards4b: This is an unrelated change - It looks like a merge error, the flushalll call is at line 1328.

gold2718: Gee wiz, I thought only subversion did stupid merge stuff like this. Sigh.

PR #2055, opened 2017-11-13
Fix redirection for run_sub to include stderr.

### LINKED ISSUE #2029: understand and better use python logging facility
Currently a script called as a subroutine from run_sub_or_cmd needs to add the following lines to
have logger output redirect to it's output file:
```
    ch = logging.StreamHandler(stream=sys.stdout)                               
    ch.setLevel(logging.DEBUG)                                                  
    logger.addHandler(ch)       
```
However this does not redirect exception output from the expect routine.    
This means that the error message ends up in the batch system output log and
not in the output stream set by the subroutine - how do we get this log output in the correct location?

jgfouca:
Fix redirection for run_sub to include stderr.

This way, both stdout and stderr will be captured in the log file
for run_sun routines.

Test suite: code_checker, cime_tiny
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2029 

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/utils.py
+++ b/scripts/lib/CIME/utils.py
@@ -2,9 +2,8 @@
 Common functions used by cime python scripts
 Warning: you cannot use CIME Classes in this module as it causes circular dependencies
 """
-import io, logging, gzip, sys, os, time, re, shutil, glob, string, random, imp, errno, signal
+import io, logging, gzip, sys, os, time, re, shutil, glob, string, random, imp, errno, signal, traceback, warnings
 import stat as statlib
-import warnings
 import six
 from contextlib import contextmanager
 #pylint: disable=import-error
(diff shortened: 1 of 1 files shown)


jgfouca:
@jedwards4b , I didn't test this much directly since ACME doesn't make much use of run_sub. That said, I'm pretty confident this change will be what you need to address #2029

jedwards4b:
@jgfouca I've created a branch to demonstrate the problem using the pio build script:
To github.com:jedwards4b/cime.git
 * [new branch]          run_sub_logging_demo -> run_sub_logging_demo
You can quickly get to the problem by building an X case.

jedwards4b:
Ideally the output from expect would go both to the terminal and to the file - I think this logging tool will do that - but I haven't figured it out.

jgfouca:
@jedwards4b , sorry, I made a stupid mistake. The exception handler causes us to leave the context manager. Try this latest version.

jedwards4b:
Works now, thanks!

PR #2069, opened 2017-11-15
Add bless logging capability

jgfouca:
Also, some minor cleanup in bless_test_results.

Test suite: scripts_regression_tests, extensive by-hand testing
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: New bless_log file in baseline dirs

### DIFF
--- a/scripts/lib/CIME/bless_test_results.py
+++ b/scripts/lib/CIME/bless_test_results.py
@@ -6,6 +6,7 @@ from CIME.hist_utils import generate_baseline, compare_baseline
 from CIME.case import Case
 
 import os, glob, time, six
+logger = logging.getLogger(__name__)
 
 ###############################################################################
 def bless_namelists(test_name, report_only, force, baseline_name, baseline_root):
@@ -15,7 +16,7 @@ def bless_namelists(test_name, report_only, force, baseline_name, baseline_root)
     # re-run create_test.
 
     # Update namelist files
-    print("Test '{}' had namelist diff".format(test_name))
+    logger.info("Test '{}' had namelist diff".format(test_name))
     if (not report_only and
         (force or six.moves.input("Update namelists (y/n)? ").upper() in ["Y", "YES"])):
         create_test_gen_args = " -g {} ".format(baseline_name if get_model() == "cesm" else " -g -b {} ".format(baseline_name))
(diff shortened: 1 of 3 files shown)


rljacob:
Where will this "bless_log" live?  How are multiple blesses on multiple machines handled?

jgfouca:
@billsacks I could turn off this feature for CESM if you guys want. I was thinking it may not make sense with your workflow. That may make more sense than trying to fix the problems you raised, both of which are legitimate.

@rljacob it will live in the baseline directory, so $BASELINE_ROOT/$COMPILER/$CASENAME/bless_log

Each blessing adds a line to this file.

billsacks:
Yeah, I probably won't use it in my own workflow. No strong opinion as to whether you turn it off or leave it on without much value. @jedwards4b do you have any feelings about this?

rljacob:
Can multiple people bless a test?  And if so can they each write to this file?

jgfouca:
@rljacob , yep! Sane maintenance already requires that the baseline area be group writable.

jedwards4b:
I think that the cam developers may use it in their workflow. @cacraigucar care to weigh in?

cacraigucar:
CAM's workflow is that we generate a new baseline directory with the CAM tag name each time we make a tag using bless_test_results.  Since we are naming the directory ourselves when we use bless_test_results and this naming is not error prone, I don't feel like we'd need the sha information as well.  So as long as this doesn't impact our baseline comparisons or running bless_test_results, I don't have an opinion on what is done with this new log.

jgfouca:
Sounds like this should be an ACME-only feature.

billsacks:
> Sounds like this should be an ACME-only feature.

Yes, probably, if the logic for making it acme-only is easy and limited. If it involves logic in a number of places, though, I'd be inclined to keep it active always and we'll just ignore it in cesm.

jgfouca:
@billsacks , nope, it's trivial.

billsacks:
Okay, sounds good, thanks.

jgfouca:
OK, this is ready.

billsacks:
Okay, I'm happy with this now, thanks.

jgfouca:
I'm going to go ahead and click the button for Jim, I have another PR that needs this.

Issue #2012, opened 2017-10-31
Make new STA test type


jgfouca:
make SystemTests/sta.py

it will test short-term archiving while running.

billsacks:
Can you describe how this will differ from other tests that cover the short-term archiver, like ERR and IRT?

mfdeakin-sandia:
First note this will be E3SM specific, since only E3SM supports running the archiver while the model is running.
The difference is we want to standardize testing the archiver while the E3SM model is running, with the --last-date option to ensure things don't move around before we're ready. I don't actually think we need a new test type for this though - adding a test to scripts_regression_tests.py for E3SM will likely be more straightforward and correct.

billsacks:
@mfdeakin-sandia okay thanks for the clarification.

rljacob:
Just because E3SM are the only ones who use the feature, that doesn't make the test "E3SM specific".  Its a feature anyone could use and this is a test of it.
Please call  it something like "AWR.py" for "Archive While Running".  "STA" sounds like a generic test of short term archiving.

jedwards4b:
What feature do you intend to test that IRT doesn't already cover?

mfdeakin-sandia:
@jedwards4b I created a prototype test in scripts_regression_tests in my mfdeakin-sandia/test_sta_last_date branch - essentially it just tests the last-date option.

PR #2071, opened 2017-11-16
Short term archiver last date option test

mfdeakin-sandia:
Adds a system test for the last-date option for the short term archiver. This verifies that the short term archiver correctly moves files to the archive directory.

Test suite: scripts_regression_tests (in progress)
Test baseline: 
Test namelist changes: 
Test status: should be BFB

Fixes ESMCI/cime#2012


config/config_tests.xml
```diff
@@ -348,6 +348,18 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
     <DOUT_S>FALSE</DOUT_S>
   </test>
 
+  <test NAME="LDSTA">
+    <DESC>Tests the short term archiver's last date functionality.</DESC>
+    <INFO_DBUG>1</INFO_DBUG>
+    <STOP_OPTION>nmonths</STOP_OPTION>
```
jedwards4b: 24 months is way too long in my opinion.  The test should be able to be run in just a few timesteps or at worst a few days.

mfdeakin-sandia: 24 months is very long, yes, but I need to test the functionality for different years. I think better would be to get a mock run file generator going, and not use this as a system test, and just a scripts_regression_tests test again

jedwards4b: What about just setting the start date to Dec 30th and running a few days?  How would a mock run file generator predict the run files from the various components?  Using config_archive.xml to anticipate file names perhaps?

mfdeakin-sandia: Making a mock file generator work correctly is one of my main concerns, I haven't put enough thought into how it would work. It's complicated by components (MPAS...) using non-standard file names.
Setting the start date to Dec 30th seems like the best immediate solution, thanks for pointing it out.

jgfouca:
@mfdeakin-sandia , I almost forgot. Do we want to add a LDSTA test to cime_developer?

rljacob:
I thought we were restricted to 3 letters for a test name.  Does this work with create_test?

jgfouca:
@rljacob , nope, no restriction. We have plenty of tests with longer names, like NODEFAIL.

mfdeakin-sandia:
Adding it to cime_developer seems like a good idea, I'll make a new branch and do that

rljacob:
Does the test take a long time to run?  JimE seemed concerned about the length.

mfdeakin-sandia:
Most of the run time is initialization I believe, it takes ~30 s. With 24 months it took ~2 minutes

jedwards4b:
It's okay for an A case, but we need this test for external components which by definition are not A cases and will take much longer to run.

mfdeakin-sandia:
Currently I'm only verifying the restart directories created are as expected (which is insufficient in the long term), so running this for other components shouldn't change the results yet

PR #2074, opened 2017-11-16
Add LDSTA to cime_developer test suite

mfdeakin-sandia:
See title

Test suite: cime_developer
Test baseline: 
Test namelist changes: 
Test status: BFB

### DIFF
--- a/scripts/lib/update_acme_tests.py
+++ b/scripts/lib/update_acme_tests.py
@@ -53,7 +53,8 @@ _TEST_SUITES = {
                              "SMS.T42_T42.S",
                              "PRE.f19_f19.ADESP",
                              "PRE.f19_f19.ADESP_TEST",
-                             "MCC_P1.f19_g16_rx1.A")
+                             "MCC_P1.f19_g16_rx1.A",
+                             "LDSTA.f45_g37_rx1.A")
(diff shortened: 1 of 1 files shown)


scripts/lib/update_acme_tests.py
```diff
@@ -53,7 +53,8 @@
                              "SMS.T42_T42.S",
                              "PRE.f19_f19.ADESP",
                              "PRE.f19_f19.ADESP_TEST",
-                             "MCC_P1.f19_g16_rx1.A")
+                             "MCC_P1.f19_g16_rx1.A",
+                             "LDSTA.ne4np4_oQU240.A")
```
jedwards4b: I don't think that this grid is defined for cesm

mfdeakin-sandia: Okay. What's the smallest grid e3sm and cesm support? Or does it affect A case runtimes and/or output file sizes?

jedwards4b: f19_g16_rx1 is typical for testing

rljacob: I think f19 is much higher res then ne4.  Try f45_g37_rx1

Issue #2083, opened 2017-11-18
LDSTA.f45_g37_rx1.A. test fails if run twice. 


jedwards4b:
scripts_regression_tests.py runs each test twice to make sure that tests have that capability, this one needs some work.

mfdeakin-sandia:
Would it make sense to just delete the archive directory, or might that mess up other tests?

PR #2084, opened 2017-11-18
Enable ldsta case.submit to be run twice

mfdeakin-sandia:
This deletes the archive directory tree used by LDSTA before running the test again.

Test status: In progress

Fixes ESMCI/cime#2083

### DIFF
--- a/scripts/lib/CIME/SystemTests/ldsta.py
+++ b/scripts/lib/CIME/SystemTests/ldsta.py
@@ -11,7 +11,9 @@ from CIME.case_st_archive import case_st_archive, get_file_date
 
 import datetime
 import glob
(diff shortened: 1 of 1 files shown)


mfdeakin-sandia:
I'm still testing this, so this shouldn't be merged yet

mfdeakin-sandia:
This appears to be working when tested by hand; scripts_regression_tests is still in progress but no failures yet

mfdeakin-sandia:
This passed the Z_FullSystemTest in scripts regression tests, so it should be ready

jedwards4b:
@jgfouca why is the melvin cdash test passing without this change?

PR #2086, opened 2017-11-20
Fix run_sub_or_cmd in the case where a cmd is run and log is None

### LINKED ISSUE #2085: run_sub_or_cmd logs output to file 'None'
optional logfile parameter is never checked, resulting in a  logfile 'None'

```
def run_sub_or_cmd(cmd, cmdargs, subname, subargs, logfile=None, case=None,
                   from_dir=None, combine_output=False):

```
...
```
  output = run_cmd_no_fail("{} 1> {} 2>&1".format(fullcmd, logfile),
                                 combine_output=combine_output,
                                 from_dir=from_dir)

```

`ERROR: Command: '/sandbox/sarich/ACME/components/clm/cime_config/buildnml /sandbox/sarich/ACME/cime/scripts/TESTWCYCLE 1> None 2>&1' failed with error '' from dir '/sandbox/sarich/ACME/cime/scripts/TESTWCYCLE'`

jgfouca:
Also, refactor to remove combine_output arg. It was always True and
the implementation made it so False wouldn't de-combine output anyway.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2085 

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/utils.py
+++ b/scripts/lib/CIME/utils.py
@@ -225,8 +225,7 @@ def _convert_to_fd(filearg, from_dir):
 
 _hack=object()
 
-def run_sub_or_cmd(cmd, cmdargs, subname, subargs, logfile=None, case=None,
-                   from_dir=None, combine_output=False):
+def run_sub_or_cmd(cmd, cmdargs, subname, subargs, logfile=None, case=None, from_dir=None):
 
     # This code will try to import and run each buildnml as a subroutine
     # if that fails it will run it as a program in a seperate shell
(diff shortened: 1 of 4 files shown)


scripts/lib/CIME/utils.py
```diff
@@ -225,8 +225,7 @@ def _convert_to_fd(filearg, from_dir):
 
 _hack=object()
 
-def run_sub_or_cmd(cmd, cmdargs, subname, subargs, logfile=None, case=None,
-                   from_dir=None, combine_output=False):
+def run_sub_or_cmd(cmd, cmdargs, subname, subargs, logfile=None, case=None, from_dir=None):
```
jedwards4b: Should we leave the combine_output flag in place and mark it deprecated so that we don't break any external users - I don't actually know that there are any but...

jgfouca: I think we can just remove it. I thought external users used CIME through it's script interface?

jedwards4b:
Not always - many of the buildlib and buildnml scripts are written in python and instantiate a case object, I don't think any use the run_sub_or_cmd but I think that we need to be careful when removing optional arguments.

jgfouca:
Oh right. I forget that because ACME hasn't converted any of that stuff to python yet. Can you grep your repo for run_sub_or_cmd calls?

PR #2125, opened 2017-12-08
Pio2 update

jedwards4b:
pnetcdf 1.9.0 and netcdf 4.5 will no longer use NC_MAX variables and set them to INT_MAX.  This can cause the compiler to generate medium memory model code.   

There are 4 major changes:

limit to 2GiB due to romio bug
dont use NC_MAX values
Fix for incorrect type of 'mpi_type' in pioc_support.c
added enable-logging option to autotools build

Test suite: external pio2 test suite using pnetcdf/1.9.0 candidate as well as pnetcdf/1.8.1 and netcdf 4.4.1.1
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/externals/pio2/configure.ac
+++ b/src/externals/pio2/configure.ac
@@ -26,8 +26,17 @@ AC_ARG_WITH([piobuffersize],
 AC_MSG_RESULT([$PIO_BUFFER_SIZE])
 AC_DEFINE_UNQUOTED([PIO_BUFFER_SIZE], [$PIO_BUFFER_SIZE], [buffer size for darray data.])
 
-# Need to allow user to set this.
-AC_DEFINE([PIO_ENABLE_LOGGING], [1], [log messages from library])
+# Does the user want to enable logging?
+AC_MSG_CHECKING([whether debug logging is enabled])
+AC_ARG_ENABLE([logging],
+              [AS_HELP_STRING([--enable-logging],
+                              [enable debug logging capability (will negatively impact performance). \
+			      This debugging feature is probably only of interest to PIO developers.])])
+test "x$enable_logging" = xyes || enable_logging=no
+AC_MSG_RESULT([$enable_logging])
+if test "x$enable_logging" = xyes; then
+   AC_DEFINE([PIO_ENABLE_LOGGING], 1, [If true, turn on logging.])
+fi
 
 # NetCDF (at least classic) is required for PIO to build.
 AC_DEFINE([_NETCDF], [1], [netCDF classic library available])
--- a/src/externals/pio2/ctest/runcdash-nwscla-intel.sh
+++ b/src/externals/pio2/ctest/runcdash-nwscla-intel.sh
@@ -14,14 +14,14 @@ module unload netcdf
 module swap intel intel/17.0.1
 module load cmake/3.7.2
 module load netcdf-mpi/4.4.1.1
-module load pnetcdf/1.8.0
-module switch mpt mpt/2.15
+module load pnetcdf/1.8.1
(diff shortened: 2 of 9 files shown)


dqwu:
`if (!(chunksizesp = malloc(ndims)))`
Where will this allocated memory be freed?

jedwards4b:
Good point - updated.

dqwu:
@jedwards4b 
The following patch seems OK for the ROMIO bug (the fix on ROMIO side will not be available in a short term). 
However, it reminds me that in pio_darray.c, there is a global variable PIO_Offset pio_buffer_size_limit (default value is  10 MB) which can be set by users with API PIOc_set_buffer_size_limit()
Should this user specified value be used to limit the IO buffer size here (in addition to the 2 GB limit due to the ROMIO bug)?
```
+    /* the limit of data_size < INT_MAX is due to a bug in ROMIO which limits
+       the size of contiguous data to INT_MAX, a fix has been proposed in
+       https://github.com/pmodels/mpich/pull/2888 */
+    io_data_size = (1 + wmb->num_arrays) * iodesc->maxiobuflen * vdesc->mpi_type_size;
+    if(io_data_size > INT_MAX)
+	needsflush = 2;
```

jedwards4b:
@dqwu that limits the size of the memory buffer pio uses, but doesn't have anything to do with the size of the contiguous array that pio will try to write so it's really orthogonal to this change.

dqwu:
@jedwards4b 
You might also need to move the patch for the ROMIO bug after the code block below. Otherwise, when PIO_USE_MALLOC is defined, the code will likely call "needsflush = 0", reverting "needsflush = 2" in the patch.

> #if PIO_USE_MALLOC
> ...
>         if ((realloc_data = realloc(wmb->data, data_size)))
>         {
>             needsflush = 0;
>             wmb->data = realloc_data;
>         }
>         else /* Failed to realloc, but wmb->data is still valid for a flush. */
>         {
>             needsflush = 1;
>         }
> ...
> #else
> ...
>     if (needsflush == 0)
>         needsflush = (maxfree <= 1.1 * (1 + wmb->num_arrays) * arraylen * vdesc->mpi_type_size);
> #endif

jedwards4b:
rebased to 9d67e96

dqwu:
@jedwards4b 
I ran AddressSanitizer and test_pioc failed. The following two lines have a wrong argument passed to malloc, which should be "ndims * sizeof(PIO_Offset)".
```
1098: if (!(chunksizesp = malloc(ndims)))
...
1524: if (!(chunksizesp = malloc(ndims))) 
```
PS, the fix "limit to 2GiB due to romio bug" should be moved to a new place as mentioned earlier (to avoid overwriting needsflush which has been set to 2).

dqwu:
@jedwards4b 
The content of this PR should also be updated. From the commits there are 4 major changes:
1. limit to 2GiB due to romio bug
2. dont use NC_MAX values
3. Partial fix for incorrect type of 'mpi_type' in pioc_support.c
4. added enable-logging option to autotools build

The title is "Pio2 update" but the content only mentions 2 (1, 3, 4 should be included):
> pnetcdf 1.9.0 and netcdf 4.5 will no longer use NC_MAX variables and set them to INT_MAX. This can cause the compiler to generate medium memory model code.

jedwards4b:
@dqwu Thanks for the address sanitizer, I have also moved the romio workaround and updated the PR as suggested.

jedwards4b:
@dqwu are you waiting for something else?

PR #2132, opened 2017-12-12
Address slow CIME performance (do NOT merge)

jgfouca:
Hi all,

I wanted to start a discussion about CIME's performance and it's use of XML as a database. I've been doing some unrelated work for ACME where I've been watching "top" while running create_test. Far too often, I see cases bottle-necked in python instead of building/running. Some recent testing on melvin (fast machine, local disk) consistently shows that it takes about 15 seconds for the first 3 phases (create, xml, setup) on the cime_tiny tests, with case_setup being the most expensive at about 10 seconds. Surely, we can do better than this?

I profiled case.setup and attached the profile to our slack. The thing that caught my eye is that we're spending 60% of our runtime in generic_xml.get_nodes, so this is obviously the area to focus  on. I do not have much experience with python XML libraries, but it seems like we're being very inefficient in the way we're doing queries.

In this PR, I did some prototyping. The goal was to fully encapsulate our use of XML (no direct calls to XML libs) and to try to use caching to speed things up. Our encapsulation was pretty good with most queries going through our classes, but direct access of XML elements was done in several places.

I had some modest success with this toy PR, reducing time spent in get_nodes by 50%, and reducing the cost of case.submit from 10 seconds down to 7 seconds on melvin. Logging showed about 10k cache "hits" and 5k "misses".

Anyway, I don't trust this PR and this was only intended as a discussion topic. Ideally, I'd like to see full encapsulation of our use of XML, and the addition of a high-performing internal representation as an in-memory intermediary between CIME and the XML file database.

### DIFF
--- a/scripts/lib/CIME/XML/archive.py
+++ b/scripts/lib/CIME/XML/archive.py
@@ -4,7 +4,7 @@ Interface to the archive.xml file.  This class inherits from GenericXML.py
 """
 
 from CIME.XML.standard_module_setup import *
-from CIME.XML.generic_xml import GenericXML
+from CIME.XML.generic_xml import GenericXML, ConstantElement
 from CIME.XML.files import Files
 from CIME.utils import expect, get_model
 
@@ -26,9 +26,8 @@ class Archive(GenericXML):
         if files is None:
             files = Files()
 
-        components_node = ET.Element("components")
+        components_node = ConstantElement(ET.Element("components"))
         components_node.set("version", "2.0")
-
 
         model = get_model()
         if 'drv' not in components:
--- a/scripts/lib/CIME/XML/batch.py
+++ b/scripts/lib/CIME/XML/batch.py
@@ -113,11 +113,11 @@ class Batch(GenericXML):
         jobs = []
         bnode = self.get_node("batch_jobs")
         for jnode in bnode:
-            if jnode.tag == "job":
+            if jnode.tag() == "job":
                 name = jnode.get("name")
                 jdict = {}
                 for child in jnode:
-                    jdict[child.tag] = child.text()
+                    jdict[child.tag()] = child.text()
 
             jobs.append((name, jdict))
 
--- a/scripts/lib/CIME/XML/batch.py
+++ b/scripts/lib/CIME/XML/batch.py
@@ -91,7 +91,7 @@ class Batch(GenericXML):
 
         node = self.get_optional_batch_node(name)
         if node is not None:
-            value = node.text
+            value = node.text()
 
         if value is None:
             # if all else fails
@@ -117,7 +117,7 @@ class Batch(GenericXML):
                 name = jnode.get("name")
                 jdict = {}
                 for child in jnode:
-                    jdict[child.tag] = child.text
+                    jdict[child.tag] = child.text()
 
             jobs.append((name, jdict))
 
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -9,6 +9,88 @@ import six
 
 logger = logging.getLogger(__name__)
 
+class ConstantElement(object):
+
+    def __init__(self, xml_element):
+        self._xml_element = xml_element
+        self._cache_find  = {}
+
+    #
+    # const methods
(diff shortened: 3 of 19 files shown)


billsacks:
This seems like a very worthwhile effort. I know very little about this part of cime, so I'll defer to others in the review. One question, though: Before we start doing our own optimizations: Does anyone know if we're using an optimal xml parser, or if there is a more optimized version out there? e.g., some googling suggests using `cElementTree` rather than `ElementTree`. 

I tried simply

```diff
diff --git a/scripts/lib/CIME/XML/standard_module_setup.py b/scripts/lib/CIME/XML/standard_module_setup.py
index 8ab608c5d..8f4cc5842 100644
--- a/scripts/lib/CIME/XML/standard_module_setup.py
+++ b/scripts/lib/CIME/XML/standard_module_setup.py
@@ -4,7 +4,7 @@
 import logging
 import os
 import sys
-import xml.etree.ElementTree as ET
+import xml.etree.cElementTree as ET
 import re
 LIB_DIR = os.path.dirname(os.path.dirname(os.path.dirname(os.path.abspath(__file__))))
 sys.path.append(LIB_DIR)
```

but apparently we'd need to do a bit more than simply that.

scripts/lib/CIME/XML/archive.py
```diff
@@ -26,10 +26,9 @@ def setup(self, env_archive, components, files=None):
         if files is None:
             files = Files()
 
-        components_node = ET.Element("components")
+        components_node = ConstantElement(ET.Element("components"))
```
jedwards4b: I think that the general philosophy has been that xml should not be referenced directly outside of the XML directory.   I consider this a bug.

scripts/lib/CIME/XML/batch.py
```diff
@@ -91,7 +91,7 @@ def get_value(self, name, attribute=None, resolved=True, subgroup=None):
 
         node = self.get_optional_batch_node(name)
         if node is not None:
-            value = node.text
+            value = node.text()
```
jedwards4b: Interesting - is this change just for good form?  It seems to function correctly as is.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -91,7 +187,7 @@ def get_node(self, nodename, attributes=None, root=None, xpath=None):
 
         nodes = self.get_nodes(nodename, attributes=attributes, root=root, xpath=xpath)
 
-        expect(len(nodes) == 1, "Incorrect number of matches, {:d}, for nodename '{}' and attrs '{}' in file '{}'".format(len(nodes), nodename, attributes, self.filename))
+        #expect(len(nodes) == 1, "Incorrect number of matches, {:d}, for nodename '{}' and attrs '{}' in file '{}'".format(len(nodes), nodename, attributes, self.filename))
```
jedwards4b: why remove these checks? Should they be enabled only for debug?   Should they be removed entirely?

jgfouca:
@billsacks , that's a good idea and one of the first things I tried was exactly what you did. It seems in more recent versions of python, cElementTree and ElementTree have identical performance.

scripts/lib/CIME/XML/archive.py
jgfouca: Yes, this is definitely a leak of XML encapsulation.

scripts/lib/CIME/XML/batch.py
jgfouca: I had to do this since nodes are ConstantElements in this PR.

scripts/lib/CIME/XML/generic_xml.py
jgfouca: If you look at the profile, a significant chunk of computation is taken by these expects since we call get_nodes so often. I was just trying to see how much speed-up I could achieve. They add some value, so would probably have to be left in.

jgfouca:
@jedwards4b yes, we can certainly do better in terms of design. The real solution would probably use GenericXML as the interface to the database as you suggested. Don't worry too much about the details of this PR, it will never be merged.

jgfouca:
Replaced with #2148 and #2205

PR #2135, opened 2017-12-13
Some progress moving to db-neutral API. (do not merge)

jgfouca:
Beginning to define solid APIs for acting on individual nodes and
entire files.

Starting to convert archive, compilers, entry_id

xml libs are no longer imported by all files. We only want generic_xml
using xml libs directly.

### DIFF
--- a/scripts/lib/CIME/XML/entry_id.py
+++ b/scripts/lib/CIME/XML/entry_id.py
@@ -408,8 +408,8 @@ class EntryID(GenericXML):
                             if f1val != f2val:
                                 xmldiffs[vid] = [f1val, f2val]
                         else:
-                            f1val = self.to_string(node)
-                            f2val = self.to_string(f2match)
+                            f1val = self.to_string(node, method="text")
+                            f2val = self.to_string(f2match, method="text")
                             if f2val != f1val:
                                 f1value_nodes = self.get_children("value", root=node)
                                 for valnode in f1value_nodes:
--- a/scripts/lib/CIME/XML/entry_id.py
+++ b/scripts/lib/CIME/XML/entry_id.py
@@ -190,13 +190,13 @@ class EntryID(GenericXML):
         return valid_values_list
 
     def set_valid_values(self, vid, new_valid_values):
-        node = self.get_optional_child("entry", {"id":vid})
+        node = self.scan_optional_child("entry", {"id":vid})
         if node is None:
             return None
         return self._set_valid_values(node, new_valid_values)
 
     def get_nodes_by_id(self, vid):
-        return self.get_children("entry", {"id":vid})
+        return self.scan_children("entry", {"id":vid})
 
     def _set_valid_values(self, node, new_valid_values):
         old_vv = self._get_valid_values(node)
--- a/scripts/lib/CIME/XML/entry_id.py
+++ b/scripts/lib/CIME/XML/entry_id.py
@@ -195,7 +195,7 @@ class EntryID(GenericXML):
             return None
         return self._set_valid_values(node, new_valid_values)
 
-    def get_children_by_id(self, vid):
+    def get_nodes_by_id(self, vid):
         return self.get_children("entry", {"id":vid})
 
     def _set_valid_values(self, node, new_valid_values):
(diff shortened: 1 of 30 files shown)


jgfouca:
The first step is defining APIs and making all classes use. This PR will show the progression of the APIs and the conversion of files to the new API.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -3,12 +3,18 @@
...
 """
 from CIME.XML.standard_module_setup import *
+import xml.etree.ElementTree as ET
 from distutils.spawn import find_executable
 import getpass
 import six
 
 logger = logging.getLogger(__name__)
 
+class _Element(object): # private class, don't want users constructing directly
```
jgfouca: Returning XML Elements as this class will prevent users from making their own Element or making calls directly on those elements (notice there are no methods on this class).

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -48,18 +52,114 @@ def read(self, infile, schema=None):
...
+
+    def make_child(self, name, text=None, root=None, attributes=None):
+        root = root if root else self.root
+        node = __Element(ET.SubElement(root, name, attrib=attributes))
+        if text:
+            self.set_text(node, text)
+
+        return node
+
+    def get_children(self, root=None, name=None, attributes=None):
```
jgfouca: I think a key to better performance will be to primary limit searches to direct children.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -82,41 +182,38 @@ def write(self, outfile=None):
             with open(outfile,'w') as xmlout:
                 xmlout.write(xmlstr)
 
-    def get_node(self, nodename, attributes=None, root=None, xpath=None):
+    def scan_node(self, nodename, attributes=None, root=None):
```
jgfouca: Doing broad xpath searches is expensive. Rename these and only use them sparingly.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -152,15 +249,6 @@ def get_nodes(self, nodename, attributes=None, root=None, xpath=None):
 
         return nodes
 
-    def add_child(self, node, root=None):
```
jgfouca: Moved up.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -250,13 +338,6 @@ def get_resolved_value(self, raw_value):
 
         return item_data
 
-    def add_sub_node(self, node, subnode_name, subnode_text):
```
jgfouca: Redundant.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -270,19 +351,6 @@ def validate_xml_file(self, filename, schema):
         else:
             logger.warning("xmllint not found, could not validate file {}".format(filename))
 
-    def get_element_text(self, element_name, attributes=None, root=None, xpath=None):
```
jgfouca: Moved up.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -82,41 +182,38 @@ def write(self, outfile=None):
...
         """
 
-        nodes = self.get_nodes(nodename, attributes=attributes, root=root, xpath=xpath)
+        nodes = self.get_nodes(nodename, attributes=attributes, root=root)
 
         expect(len(nodes) == 1, "Incorrect number of matches, {:d}, for nodename '{}' and attrs '{}' in file '{}'".format(len(nodes), nodename, attributes, self.filename))
         return nodes[0]
 
-    def get_optional_node(self, nodename, attributes=None, root=None, xpath=None):
+    def scan_optional_node(self, nodename, attributes=None, root=None):
```
jgfouca: I don't think we ever want users to use xpath as that leaks the fact that the underlying database is XML.

jedwards4b:
I think that you are taking this a step further than I had considered... I'm good with it if it works.

scripts/lib/CIME/XML/entry_id.py
```diff
@@ -319,19 +319,19 @@ def _get_value(self, node, attribute=None, resolved=True, subgroup=None):
...
-        nodes = self.get_nodes("entry")
+        nodes = self.get_children("entry")
         elements = []
         for node in nodes:
             content = self.get_element_text(childname, root=node)
-            expect(content is not None,"No childname {} for id {}".format(childname, node.get("id")))
+            expect(content is not None,"No childname {} for id {}".format(childname, self.get(node, "id")))
             if content == childcontent:
-                elements.append(deepcopy(node))
+                elements.append(node)
```
jedwards4b: why don't we need the deepcopy any longer?  I remember this causing problems because without deepcopy you got a pointer to the original instead of a copy.

jgfouca: @jedwards4b , I double-checked and add_elements_by_group was the only caller of this method and it was already making a deepcopy of these nodes.

PR #2139, opened 2017-12-14
Utf8 encoding

jedwards4b:
Use the encoding flag to open in py3 to set utf-8 

Test suite: scripts_regression_tests.py py2 and py3
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes cdash testing with py3

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -46,20 +46,13 @@ class GenericXML(object):
         Read and parse an xml file into the object
         """
         logger.debug("read: " + infile)
-        if six.PY3:
-            with open(infile, mode='r', encoding='utf-8') as fd:
-                if self.tree:
-                    self.root.append(ET.parse(fd).getroot())
-                else:
-                    self.tree = ET.parse(fd)
-                    self.root = self.tree.getroot()
-        else:
-            with open(infile, mode='r') as fd:
-                if self.tree:
-                    self.root.append(ET.parse(fd).getroot())
-                else:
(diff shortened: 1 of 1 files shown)


scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -46,12 +46,20 @@ def read(self, infile, schema=None):
         Read and parse an xml file into the object
         """
         logger.debug("read: " + infile)
-        with open(infile, 'r') as fd:
-            if self.tree:
-                self.root.append(ET.parse(fd).getroot())
-            else:
-                self.tree = ET.parse(fd)
-                self.root = self.tree.getroot()
+        if six.PY3:
```
jgfouca: Nice catch for this bug. Did you figure out why it worked from the command line? You were still using python3 on the command-line, right?

Anyway, I think we can reduce some code duplication here, even though py2 does not support encoding:
```
file_open = (lambda x: open(x, 'r', encoding='utf-8')) if six.PY3 else (lambda x: open(x, 'r'))
with file_open(infile) as fd:
    ...
```

jedwards4b: Thanks - I thought there must be a way to do that.  Do you want to push that change?

jedwards4b: I don't understand why it worked from the command line but not from cron.

jedwards4b:
Okay, i've pushed the suggested change

PR #2148, opened 2017-12-18
Fully encapsulate XML within generic_xml

jgfouca:
Fully encapsulate XML within generic_xml in order to facilitate future refactors.

I've confirmed that we are losing lots of performance in scan_children (formerly get_nodes). The ".//" xpath scan is very expensive since it recursively scans eveything under the root node. We can greatly speed-up CIME by replacing these searches with seaches of immediate children only.

Note, this change is not backwards compatible for scripts using the CIME/XML library directly. I had to make minor changes here and there to buildnml python scripts.

The big change is that get_children (formerly get_nodes) is no-longer recursive, it only looks for immediate children. The places that really needed recursion have been changed to use the scan_* methods. In the future, we should investigate how we can replace these scan calls with get_child calls.

In order to accommodate this change, we need to catch all the places where get_children and scan_children return different sets of elements. These are calls that either need to be restricted to a different root or need to be scans.

Test suite: scripts_regression_tests (melvin, skybridge coming soon)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: Significant python API changes

Code review: @jedwards4b


jgfouca:
Still testing.

jgfouca:
I will annotate these changes sometime tomorrow.

scripts/lib/CIME/XML/compilerblock.py
```diff
@@ -81,7 +81,7 @@ class CompilerBlock(object):
     matches_machine
     """
 
-    def __init__(self, writer, compiler_elem, machobj):
+    def __init__(self, writer, compiler_elem, machobj, db):
```
jgfouca: Need to pass-in a db (some GenericXML object) so that we can use the generic_xml API.

scripts/lib/CIME/XML/batch.py
```diff
@@ -111,13 +111,13 @@ def get_batch_jobs(self):
...
-            if jnode.tag == "job":
-                name = jnode.get("name")
+        bnode = self.get_child("batch_jobs")
+        for jnode in self.get_children(root=bnode, no_validate=True):
+            if self.name(jnode) == "job":
+                name = self.get(jnode, "name")
                 jdict = {}
-                for child in jnode:
-                    jdict[child.tag] = child.text
+                for child in self.get_children(root=jnode, no_validate=True):
```
jgfouca: Don't need to validate this get_children call since this is a new get_children call and not a replacement of an existing call.

scripts/lib/CIME/XML/compilerblock.py
```diff
@@ -172,19 +173,19 @@ def _handle_references(self, elem, set_up, tear_down, depends):
...
                 depends.add(var_name)
             else:
                 expect(False,
-                       "Unexpected tag "+child.tag+" encountered in "
+                       "Unexpected tag "+self._db.name(child)+" encountered in "
                        "config_build.xml. Check that the file is valid "
                        "according to the schema.")
-            if child.tail is not None:
-                output += child.tail
+            if child.xml_element.tail is not None:
```
jgfouca: Didn't want to think about a replacement for tail, so left in the raw XML call here for now.

scripts/lib/CIME/XML/compilers.py
```diff
@@ -191,12 +188,14 @@ def _write_macros_file_v2(self, build_system, output, xml=None):
         value_lists = dict()
         node_list = []
         if xml is None:
-            node_list = self.get_nodes("compiler")
+            node_list = self.get_children(name="compiler")
         else:
-            node_list = ET.parse(xml).findall("compiler")
+            gen_xml = GenericXML()
```
jgfouca: Some effort was required here to avoid the direct ET call.

scripts/lib/CIME/XML/compsets.py
```diff
@@ -16,14 +16,13 @@ def __init__(self, infile=None, files=None):
             files = Files()
         schema = files.get_schema("COMPSETS_SPEC_FILE")
         GenericXML.__init__(self, infile, schema=schema)
-        self.groups={}
```
jgfouca: unused.

scripts/lib/CIME/XML/compsets.py
```diff
@@ -45,33 +44,36 @@ def get_compset_var_settings(self, compset, grid):
         Variables can be set in config_compsets.xml in entry id settings with compset and grid attributes
         find and return id value pairs here
         '''
-        nodes = self.get_nodes("entry")
-        # Get an empty entryid obj to use
-        entryidobj = EntryID()
+        entries = self.get_optional_child("entries")
```
jgfouca: We can avoid the recursive search by grabbing the "entries" element and using that to find "entry" elements.

scripts/lib/CIME/XML/compsets.py
```diff
@@ -86,18 +88,3 @@ def print_values(self, arg_help=True):
         logger.info("       --------------------------------------")
         for key in sorted(compsets_text.keys()):
             logger.info("   {:20} : {}".format(key, compsets_text[key]))
-
-    def return_all_values(self):
```
jgfouca: Couldn't find anything in CIME making any call to return_all_values.

scripts/lib/CIME/XML/entry_id.py
```diff
@@ -71,32 +70,33 @@ def _get_value_match(self, node, attributes=None, exact_match=False):
         # if there is a <values> element - check to see if there is a match attribute
         # if there is NOT a match attribute, then set the default to "first"
         # this is different than the component class _get_value_match where the default is "last"
-        values_node = self.get_optional_node("values", root=node)
+        values_node = self.get_optional_child("values", root=node)
         if values_node is not None:
-            match_type = values_node.get("match", default="first")
+            match_type = self.get(values_node, "match", default="first")
+            node = values_node
```
jgfouca: Look for "value" nodes under "values" node if it exists.

scripts/lib/CIME/XML/entry_id.py
```diff
@@ -165,19 +165,19 @@ def get_description (self, node):
...
+            nodes = self.get_children("entry", attributes={"id":vid}, root=group)
             if nodes:
-                result.append(group.get("id"))
+                result.append(self.get(group, "id"))
 
         return result
 
     def get_valid_values(self, vid):
-        node = self.get_optional_node("entry", {"id":vid})
+        node = self.scan_optional_child("entry", {"id":vid})
```
jgfouca: In the future, get "group" children and then look for entries under each group.

scripts/lib/CIME/XML/entry_id.py
```diff
@@ -306,11 +306,13 @@ def _get_value(self, node, attribute=None, resolved=True, subgroup=None):
             logger.debug("No node")
             return val
 
-        logger.debug("Found node {} with attributes {}".format(node.tag , node.attrib))
+        logger.debug("Found node {} with attributes {}".format(self.name(node) , self.attrib(node)))
         if attribute:
+            vals = self.get_optional_child("values", root=node)
```
jgfouca: "value" node may be under "values" node. get_element_text is not a recursive search.

scripts/lib/CIME/XML/entry_id.py
```diff
@@ -321,19 +323,19 @@ def _get_value(self, node, attribute=None, resolved=True, subgroup=None):
...
     def get_elements_from_child_content(self, childname, childcontent):
-        nodes = self.get_nodes("entry")
+        nodes = self.get_children("entry")
         elements = []
         for node in nodes:
             content = self.get_element_text(childname, root=node)
-            expect(content is not None,"No childname {} for id {}".format(childname, node.get("id")))
+            expect(content is not None,"No childname {} for id {}".format(childname, self.get(node, "id")))
             if content == childcontent:
-                elements.append(deepcopy(node))
```
jgfouca: No need for a deepcopy, add_elements_by_group is already making a deepcopy of these nodes.

scripts/lib/CIME/XML/entry_id.py
```diff
@@ -345,41 +347,40 @@ def add_elements_by_group(self, srcobj, attributes=None, infile=None):
...
             if gname not in self.groups.keys():
-                newgroup = ET.Element("group")
-                newgroup.set("id",gname)
                 # initialize an empty list
+                newgroup = self.make_child(name="group", attributes={"id":gname})
                 self.groups[gname] = newgroup
-                self.add_child(newgroup)
 
             # Remove {<group>, <file>, <values>} from the entry element
-            node = self.cleanupnode(node)
```
jgfouca: No need for assignment. cleanupnode modifies node in-place.

scripts/lib/CIME/XML/env_base.py
```diff
@@ -22,21 +22,22 @@ def __init__(self, case_root, infile, schema=None):
...
 
     def set_components(self, components):
         if hasattr(self, '_components'):
             # pylint: disable=attribute-defined-outside-init
             self._components = components
 
-    def check_if_comp_var(self, vid, attribute=None):
-        nodes = self.get_nodes("entry", {"id" : vid})
-        node = None
+    def check_if_comp_var(self, vid, attribute=None, node=None):
```
jgfouca: Allow this method to take a node if we already know the node we need to check.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -95,51 +94,47 @@ def get_type_info(self, vid):
...
 
         return results
 
     def create_job_groups(self, batch_jobs):
         # Subtle: in order to support dynamic batch jobs, we need to remove the
         # job_submission group and replace with job-based groups
 
-        orig_group = self.get_optional_node("group", {"id":"job_submission"})
-        expect(orig_group, "Looks like job groups have already been created")
+        orig_group = self.get_child("group", {"id":"job_submission"},
```
jgfouca: even though it looks different, the semantics of this method was not changed at all.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -149,16 +144,16 @@ def set_batch_system(self, batchobj, batch_system_type=None):
             self.set_batch_system_type(batch_system_type)
 
         if batchobj.batch_system_node is not None and batchobj.machine_node is not None:
-            for node in batchobj.get_nodes('any', root=batchobj.machine_node, xpath='*'):
-                oldnode = batchobj.get_optional_node(node.tag, root=batchobj.batch_system_node)
-                if oldnode is not None and oldnode.tag != "directives":
-                    logger.debug( "Replacing {}".format(oldnode.tag))
-                    batchobj.batch_system_node.remove(oldnode)
+            for node in batchobj.get_children(root=batchobj.machine_node, no_validate=True):
```
jgfouca: The '*' xpath was only needed because we didn't support a None nodename as an argument to get_nodes.

scripts/lib/CIME/XML/env_mach_specific.py
```diff
@@ -26,44 +26,37 @@ def __init__(self, caseroot=None, infile="env_mach_specific.xml",
...
                     value = nodes[0].text
-                newnode.set("value", value)
-                newnode = self.add_sub_node(newnode, "type", "char")
-                if item == "run_exe":
-                    newnode = self.add_sub_node(newnode, "desc", "executable name")
-                else:
-                    newnode = self.add_sub_node(newnode, "desc", "redirect for job output")
 
-                self.add_child(newnode)
+                entity_node = self.make_child("entry", {"id":item, "value":value})
```
jgfouca: Some cleanup for conciseness, same semantics.

scripts/lib/CIME/XML/env_test.py
```diff
@@ -46,62 +46,57 @@ def set_test_parameter(self, name, value):
         otherwise create a node and initialize it to value
         """
         case = self.get_value("TESTCASE")
-        tnode = self.get_node("test",{"NAME":case})
-        # This xpath statement gets immediate children only (ignoring
-        # children of build and run blocks)
-        idnode = self.get_optional_node(name, root=tnode, xpath=name)
```
jgfouca: No need for xpath, looking at immediate children is exactly what get_child does now.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -3,20 +3,42 @@
...
 from CIME.XML.standard_module_setup import *
+import xml.etree.ElementTree as ET
 from distutils.spawn import find_executable
 import getpass
 import six
+from copy import deepcopy
 
 logger = logging.getLogger(__name__)
 
+class _Element(object): # private class, don't want users constructing directly or calling methods on it
```
jgfouca: Important: Return nodes to clients as this class means they can no longer make calls directly on nodes; it forces them to go through GenericXML for everything. This can be a little more tedious for clients:

node.text     vs.   db.text(node)

But this is the most-flexible API for us because it makes the fewest assumptions.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -35,31 +57,169 @@ def __init__(self, infile=None, schema=None):
             expect("$" not in infile,"File path not fully resolved {}".format(infile))
 
             self.filename = infile
-            root = ET.Element("xml")
-            self.root = ET.SubElement(root, "file")
-            self.root.set("id", os.path.basename(infile))
-            self.root.set("version", "2.0")
+            root = _Element(ET.Element("xml"))
+            if root_name_override:
+                self.root = self.make_child(root_name_override, root=root, attributes=root_attrib_override)
```
jgfouca: Some subclasses do not use the traditional file element as the root element.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -35,31 +57,169 @@ def __init__(self, infile=None, schema=None):
...
 
+    def read_fd(self, fd):
+        if self.tree:
+            self.add_child(_Element(ET.parse(fd).getroot()))
+        else:
+            self.tree = ET.parse(fd)
+            self.root = _Element(self.tree.getroot())
+
+    #
+    # API for individual node operations
```
jgfouca: Important, here is our API for node operations.

We support the concepts:
* Nodes
* Nodes can have a key:value dict of attributes
* Nodes have a name
* Node can have text
* Nodes can have child nodes.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -35,31 +57,169 @@ def __init__(self, infile=None, schema=None):
...
+            node = _Element(ET.SubElement(root.xml_element, name))
+        else:
+            node = _Element(ET.SubElement(root.xml_element, name, attrib=attributes))
+
+        if text:
+            self.set_text(node, text)
+
+        return node
+
+    def get_children(self, name=None, attributes=None, root=None, no_validate=False):
```
jgfouca: Important function. This will be the core code for searching for direct children of a node.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -35,31 +57,169 @@ def __init__(self, infile=None, schema=None):
...
+                                match = False
+                                break
+
+                    if not match:
+                        continue
+
+            children.append(_Element(child))
+
+        # Remove
+        if not no_validate:
```
jgfouca: Important: The global replacement of get_nodes with get_children means that all the calls that used to do recursive searches now only do direct-child searches. This validation code caught all the places that really needed the recursive scan and those places have since been changed to use scan_child or refactored so that the recursion is no longer needed.

This will incur a big performance hit. We should turn this off once we're confident in this big API refactor.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -82,84 +242,73 @@ def write(self, outfile=None):
             with open(outfile,'w') as xmlout:
                 xmlout.write(xmlstr)
 
-    def get_node(self, nodename, attributes=None, root=None, xpath=None):
+    def scan_child(self, nodename, attributes=None, root=None):
```
jgfouca: Remove support for xpath.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -82,84 +242,73 @@ def write(self, outfile=None):
...
 
         if root is None:
             root = self.root
         nodes = []
 
-        expect(attributes is None or xpath is None,
-               " Arguments attributes and xpath are exclusive")
-        if xpath is None:
-            xpath = ".//"+nodename
+        xpath = ".//" + (nodename if nodename else "")
```
jgfouca: Allow for nodename=None, meaning we don't want to use element tag as selection criteria.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -82,84 +242,73 @@ def write(self, outfile=None):
...
             # xml.etree has limited support for xpath and does not allow more than
             # one attribute in an xpath query so we query seperately for each attribute
             # and create a result with the intersection of those lists
 
             for key, value in attributes.items():
-                if value is not None:
-                    expect(isinstance(value, six.string_types),
-                           " Bad value passed for key {}".format(key))
+                if value is None:
+                    xpath = ".//{}[@{}]".format(nodename, key)
```
jgfouca: Add support for value=None, meaning we don't care what the value of the attribute is as long as the element has an attribute by that name.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -82,84 +242,73 @@ def write(self, outfile=None):
...
         else:
             logger.debug("xpath: {}".format(xpath))
-            nodes = root.findall(xpath)
+            nodes = root.xml_element.findall(xpath)
 
         logger.debug("Returning {} nodes ({})".format(len(nodes), nodes))
 
-        return nodes
-
-    def add_child(self, node, root=None):
```
jgfouca: Moved higher in file

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -250,13 +398,6 @@ def get_resolved_value(self, raw_value):
 
         return item_data
 
-    def add_sub_node(self, node, subnode_name, subnode_text):
-        expect(node is not None," Bad value passed")
-        subnode = ET.Element(subnode_name)
```
jgfouca: Redundant with make_child.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -270,32 +411,19 @@ def validate_xml_file(self, filename, schema):
         else:
             logger.warning("xmllint not found, could not validate file {}".format(filename))
 
-    def get_element_text(self, element_name, attributes=None, root=None, xpath=None):
-        element_node = self.get_optional_node(element_name, attributes, root, xpath)
-        if element_node is not None:
-            return element_node.text
-        return None
-
```
jgfouca: Moved up.

scripts/lib/CIME/XML/grids.py
```diff
@@ -700,15 +704,3 @@ def _get_all_values_v2(self):
                     grid_list.append({gridname:(grids_dict[alias], domains[gridname])})
 
         return default_comp_grids, grid_list
-
-    def return_all_values(self):
-        # parse grid data and return a dictionary
-        help_text = self.get_element_text("help")
-
-        # get a list of grid dictionaries
```
jgfouca: dead code

scripts/lib/CIME/XML/pes.py
```diff
@@ -69,18 +69,18 @@ def _find_matches(self, grid_nodes, grid, compset, machine, pesize_opts, overrid
         pesize_choice = None
         max_points = -1
         pes_ntasks, pes_nthrds, pes_rootpe, other_settings = {}, {}, {}, {}
-        pe_select = []
+        pe_select = None
```
jgfouca: empty list is not a valid value for this variable. it's used as a root=pe_select later on

scripts/lib/CIME/XML/standard_module_setup.py
```diff
@@ -1,11 +1,9 @@
 
 # pragma pylint: disable=unused-import
 
-import logging
-import os
-import sys
-import xml.etree.ElementTree as ET
```
jgfouca: We do not want anyone besides generic_xml making direct calls to ElementTree, so don't import it by default.

scripts/lib/CIME/XML/test_reporter.py
```diff
@@ -19,82 +19,43 @@ def __init__(self):
...
-        tlelem.append(elem)
-        elem   = ET.Element('baselinetag')
-        elem.text   = baseline
-        tlelem.append(elem)
-
-        self.root=tlelem
-
+        for name, text, attribs in [ ("tag_name"   , tagname , None),
+                                     ("mach"       , machine , None),
+                                     ("compiler"   , compiler, {"version":""}),
```
jgfouca: Refactor for conciseness.

scripts/lib/CIME/XML/test_reporter.py
```diff
@@ -19,82 +19,43 @@ def __init__(self):
...
-        elem.text= test_status['TPUTCOMP']
-        tlelem.append(elem)
-
-        self.root.append(tlelem)
+        tlelem = self.make_child("tests", {"testname":test_name})
 
+        for attrib_name, text in [ ("casestatus", None),
+                                   ("comment",    test_status["COMMENT"]),
+                                   ("compare",    test_status["BASELINE"]),
+                                   ("memcomp",    test_status["MEMCOMP"]),
```
jgfouca: More concisee.

scripts/lib/CIME/XML/testlist.py
```diff
@@ -112,38 +112,37 @@ def _get_testsv2(self, machine=None, category=None, compiler=None, compset=None,
...
                 this_test_node = {}
-                for key, value in tnode.items():
+                for key, value in self.attrib(tnode).items():
                     if key == "name":
                         this_test_node["testname"] = value
                     else:
                         this_test_node[key] = value
 
+                # Get options that apply to all machines/compilers for this test
+                optionnodes = self.scan_children("option", root=tnode)
```
jgfouca: Pull this out of the inner loop. It's independent of machnodes.

scripts/lib/CIME/XML/testspec.py
```diff
@@ -17,57 +17,34 @@ def __init__(self, infile):
...
-            elem.text = baselinetag
-            tlelem.append(elem)
-        if baselineroot is not None:
-            elem = ET.Element('baselineroot')
-            elem.text = baselineroot
-            tlelem.append(elem)
-        self.add_child(tlelem)
+        tlelem = self.make_child("testlist")
+
+        for name, text in [ ("testroot", testroot), ("machine", machine), ("testid", testid), ("baselinetag", baselinetag), ("baselineroot", baselineroot) ]:
```
jgfouca: Cleanup

scripts/lib/CIME/XML/testspec.py
```diff
@@ -17,57 +17,34 @@ def __init__(self, infile):
...
         self._testlist_node = tlelem
 
     def add_test(self, compiler, mpilib, testname):
         expect(testname not in self._testnodes, "Test {} already in testlist".format(testname))
-        telem = ET.Element("test")
-        telem.set("name", testname)
-        elem = ET.Element("compiler")
-        elem.text = compiler
-        telem.append(elem)
-        elem = ET.Element("mpilib")
```
jgfouca: cleanup

scripts/lib/CIME/XML/testspec.py
```diff
@@ -17,57 +17,34 @@ def __init__(self, infile):
...
         root = self._testnodes[testname]
-        pnode = self.get_optional_node("section", {"name":phase}, root=root)
+        pnode = self.get_optional_child("section", {"name":phase}, root=root)
         if pnode is not None:
-            pnode.set("status", status)
+            self.set(pnode, "status", status)
         else:
-            pnode = ET.Element("section")
-            pnode.set("name", phase)
-            pnode.set("status", status)
```
jgfouca: make_child API is powerful-enough to do all this in one call.

scripts/lib/CIME/nmlgen.py
```diff
@@ -123,11 +123,11 @@ def init_defaults(self, infiles, config, skip_groups=None, skip_entry_loop=None
             # over later settings).
             self._namelist.merge_nl(new_namelist)
 
-        if skip_entry_loop is None:
```
jgfouca: The old code was dangerous. skip_entry_loop=False would have passed this check.

scripts/lib/CIME/nmlgen.py
```diff
@@ -123,11 +123,11 @@ def init_defaults(self, infiles, config, skip_groups=None, skip_entry_loop=None
...
+        if not skip_entry_loop:
             for entry in entry_nodes:
-                self.add_default(entry.get("id"))
+                self.add_default(self._definition.get(entry, "id"))
         else:
-            return entry_nodes
+            return [self._definition.get(entry, "id") for entry in entry_nodes]
```
jgfouca: No client really needed the nodes; they just needed the ids.

src/drivers/mct/cime_config/config_component_acme.xml
```diff
@@ -629,49 +629,6 @@
...
-      <!-- estimated cost (used only for development purposes) -->
-      <value compset="_DATM.*_DLND.*_DICE.*_DOCN.*_DROF.*_SGLC_SWAV"		>-3</value> <!-- A -->
-      <value compset="_DATM.*_DICE.*_DOCN.*_WW3"				>-3</value> <!-- A_WAV -->
-      <value compset="_DATM.*_SLND.*_DICE.*_POP2_DROF.*_SGLC_SWAV"		>-1</value> <!-- C -->
-      <value compset="_DATM.*_SLND.*_DICE.*_POP2%ECO_DROF.*_SGLC_SWAV"		> 1</value> <!-- CECO -->
-      <value compset="_DATM.*_SLND.*_CICE.*_DOCN%SOM_DROF.*_SGLC_SWAV"		> 1</value> <!-- D -->
-      <value compset="_DATM.*_SLND.*_CICE.*_POP2_DROF.*_SGLC_SWAV"		>-1</value> <!-- G,H -->
-      <value compset="_DATM.*_SLND.*_CICE.*_POP2%ECO_DROF.*_SGLC_SWAV"		> 1</value> <!-- GECO -->
-      <value compset="_DATM.*_CLM.*_SICE_SOCN_RTM.*_SGLC_SWAV"			>-2</value> <!-- I -->
-      <value compset="_DATM.*_CLM.*_SICE_SOCN_MOSART.*_SGLC_SWAV"               >-2</value> <!-- IM -->
```
jgfouca: Unused.

jgfouca:
@jedwards4b , annotations complete. Also, tests should pass for you now.

jgfouca:
I know this PR looks huge, but 95% of it is just 1:1 mapping to a new, exactly equivalent API.

jedwards4b:
Line 129 of test_scheduler.py has
```
   self._machobj.set_value("PROJECT", project)
```
however PROJECT is not a direct child of _machobj root and so this fails.   However 
_machobj defines the parent in _machobj.machine_node so adding something like:
```
    def set_value(self, vid, value, subgroup=None, ignore_type=True, root=None):
        return(super(Machines, self).set_value(vid, value, subgroup=self.machine_node,
                                               ignore_type=ignore_type)
```
but this also requires a change in generic_xml.py line 329
```
      valnodes = self.get_children(vid, root=subgroup)
```
I'm not entirely comfortable with this change and so have created a new branch and pushed to it.
```
To git@github.com:jedwards4b/cime.git
 * [new branch]      jpe/hide_xml_interface -> jpe/hide_xml_interface
```

jgfouca:
@jedwards4b @fischer-ncar , thanks so much for the help with this huge PR! I've done an upstream merge to resolve conflicts. I think it's ready!

PR #2163, opened 2017-12-20
Jpe/hide xml interface

jedwards4b:

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


scripts/lib/CIME/test_scheduler.py
```diff
@@ -124,8 +124,6 @@ def __init__(self, test_names, test_data=None,
                 self._project = self._machobj.get_value("PROJECT")
         else:
             self._project = project
-
-            # Needed in case default root depends on PROJECT
```
jedwards4b: didn't mean to remove the comment

PR #2170, opened 2017-12-26
Testing/multiinst baselines

### LINKED ISSUE #2157: Unexpected test results in multiinst tests if instances are not identical
I recently set up a new POP test to make sure multi-instance runs can handle different MARBL configurations, but I think the test results are misleading. I'm running an `SMS_N2` test, and the current behavior is

1. I run the test and generate baselines

   1. Namelists from all instances get copied to the baseline directory
   1. History files from instance `0001` get copied to the baseline directory

1. I run the test again and compare to existing baselines

   1. Namelists from each instance are compared to the existing baseline from the same instance, so all `NLCOMP` tests pass
   1. History files from each instance are compared to the baseline from instance `0001`, so instances that differ from the first (i.e. instance `0002`) fail the baseline comparison

So by definition, multi-instance tests are assuming that all instances are the same. I don't like that, but if that's the way it's going to be then I think a better test set-up would be to compare instance `0002` against instance `0001` so that the failure isn't reported as a baseline comparison failure -- the test run generating the baseline should also report this failure!

(A better solution, IMO, would be to copy history files from ALL instances to baselines and remove the assumption that they are all the same.)

From conversations in slack, it turns out this was partially discussed by @billsacks and @jedwards4b in #465 but that predates my usage of multi-instance tests so I didn't chime in at the time.

edit: as @billsacks points out, I somehow managed to type 406 instead of 465 in the last paragraph

mnlevy1981:
When generating baselines for a multi-instance test, save history files
from all instances, not just the first. This allows users to create
testmod directories where two instances are not identical.

[Not ready to be accepted yet, see later comments]

Test suite: Ran an `SMS_N2_Ld3.T62_g37.C1850ECO.hobart_nag.pop-ecosys_multiinst_lsource_sink` test to generate baselines, verified history files from both instances were in baseline directory, then ran again to compare to baselines and verified that history files from both instances were compared to correct baseline files
Test baseline: 
Test namelist changes: N/A
Test status: bit for bit

Fixes #2157

User interface changes?: None

Code review:


mnlevy1981:
All I've done so far is change the `generate_baseline()` subroutine so that multi-instance cases are not treated in a special manner. It looks like the baseline comparison looks for an exact match to compare against, and then strips out the instance number if none is found and tries to compare against a generic name.

For example:
1. I run an `SMS_N2_Ld3.T62_g37.C1850ECO.hobart_nag.pop-ecosys_multiinst_lsource_sink` and that generates `SMS_N2_Ld3.T62_g37.C1850ECO.hobart_nag.pop-ecosys_multiinst_lsource_sink.C.20171226_111936_m39rvn.pop_0001.h.0001-01-03.nc`
2. Look for `pop_0001.h.0001-01-03.nc` in the baseline directory. If found, compare... otherwise look for `pop.h.0001-01-03.nc`

I think I should either remove that fallback (since it should never be triggered) or add an option to keep the previous behavior (by default, history files for all instances are kept in baselines but if a user does `____` to trigger this option only the first instance would generate baselines). I'm leaning towards the first option, but would be open to the second if that's what folks want.

After that last bit of clean-up I'll run [more thorough tests](https://github.com/ESMCI/cime/wiki/Testing)

mnlevy1981:
To be more explicit, in my previous comment I'm proposing cleaning up `_hists_match()` so that the result of

```
>>> hists1 = ['cam.h0.1850-01-08-00000.nc']
    >>> hists2 = ['cam_0001.h0.1850-01-08-00000.nc','cam_0002.h0.1850-01-08-00000.nc']
    >>> _hists_match('cam', hists1, hists2, '', '')
```
changes from

```
    ([], [], [('cam.h0.1850-01-08-00000.nc', 'cam_0001.h0.1850-01-08-00000.nc'), ('cam.h0.1850-01-08-00000.nc', 'cam_0002.h0.1850-01-08-00000.nc')])
```

to

```
    (['cam.h0.1850-01-08-00000.nc', 'cam.h0.1850-01-08-00000.nc'], ['cam_0001.h0.1850-01-08-00000.nc', 'cam_0002.h0.1850-01-08-00000.nc'], [])
```

jedwards4b:
@mnlevy1981 I'm no sure that I fully understand but we do have tests which compare a multiinstance to a single instance case - NCK - please be sure that it will still work as expected.

mnlevy1981:
@jedwards4b Thanks for the heads-up about the behavior of the `NCK` tests! As this code stands now in this pull request, those tests still work. My plan to change `_hists_match()` was a bad one, though, because it would lead to the following failure:

```
$ cat TestStatus | grep FAIL
FAIL NCK.T62_g37.C.hobart_nag.pop-default COMPARE_base_multiinst
```

So clearly I won't go with option (1) above. I could see an argument for adding an option that would only store the baseline from the first instance (behavior on the current `master`), but if we do that then we should also include `COMPARE_base_multiinst` to ensure that all other instances are identical to the first. I don't think that's necessary for this PR, and would prefer not to be the one to implement it, so I'll come talk to you or @billsacks tomorrow (Wednesday) to make sure I'm testing this correctly before assigning it to someone to merge back in.

mnlevy1981:
I ran `system_regression_tests.py` on `hobart_intel` and output looks good:

```
Ran 334 tests in 4880.373s

OK (skipped=8)
Detected failed test or user request no teardown
Leaving files:
Detected failed test or user request no teardown
Leaving files:
Detected failed test or user request no teardown
Leaving files:
Detected failed test or user request no teardown
Leaving files:
Detected failed test or user request no teardown
Leaving files:
Detected failed test or user request no teardown
Leaving files:
('All pass, removing directory:', '/scratch/cluster/mlevy/scripts_regression_test.20171227_114550')
```

PR #2173, opened 2017-12-27
Srt use cimeconfig

### LINKED ISSUE #2171: setting COMPILER in .cime/config breaks scripts_regression_tests.py
On hobart, my `.cime/config` file looks like

```
[main]
CIME_MODEL=cesm
COMPILER=nag
[create_test]
```

And when I run `scripts_regression_tests.py` I get a lot of errors that boil down to the system generating cases with `nag` but looking for case directories using `intel`:

```
FAIL: test_cime_case_xmlchange_append (__main__.K_TestCimeCase)
  File "./scripts_regression_tests.py", line 1552, in test_cime_case_xmlchange_append
    self.assertTrue(os.path.isdir(casedir), msg="Missing casedir '%s'" % casedir)
AssertionError: Missing casedir '/scratch/cluster/mlevy/scripts_regression_test.20171227_093212/TESTRUNPASS_P1x1.f19_g16_rx1.A.hobart_intel.fake_testing_only_20171227_093300'
```

(Note that `TESTRUNPASS_P1x1.f19_g16_rx1.A.hobart_intel.fake_testing_only_20171227_093300` does not exist but `TESTRUNPASS_P1x1.f19_g16_rx1.A.hobart_nag.fake_testing_only_20171227_093300` does; the only difference being `hobart_nag` instead of `hobart_intel`).

Commenting out the `COMPILER` line "fixes" the problem in the sense that the case is built using `intel` instead of `nag` so the directory match works.

jedwards4b:
scripts_regression_tests.py should use values in .cime/config if provided.  

Test suite: scripts_regression_tests.py with compiler defined as gnu in ./cime/config on cheyenne
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2171 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/tests/scripts_regression_tests.py
+++ b/scripts/tests/scripts_regression_tests.py
@@ -2480,21 +2480,25 @@ def write_provenance_info():
 
 def _main_func():
     global MACHINE
+    global NO_CMAKE
+    global FAST_ONLY
+    global NO_BATCH
+    global TEST_COMPILER
+    global TEST_MPILIB
+    global TEST_ROOT
+    global GLOBAL_TIMEOUT
     config = CIME.utils.get_cime_config()
 
     if "--fast" in sys.argv:
         sys.argv.remove("--fast")
-        global FAST_ONLY
         FAST_ONLY = True
 
     if "--no-batch" in sys.argv:
         sys.argv.remove("--no-batch")
-        global NO_BATCH
         NO_BATCH = True
 
     if "--no-cmake" in sys.argv:
         sys.argv.remove("--no-cmake")
-        global NO_CMAKE
         NO_CMAKE = True
 
     if "--machine" in sys.argv:
@@ -2515,7 +2519,6 @@ def _main_func():
         MACHINE = Machines()
 
     if "--compiler" in sys.argv:
-        global TEST_COMPILER
         midx = sys.argv.index("--compiler")
         TEST_COMPILER = sys.argv[midx + 1]
         del sys.argv[midx + 1]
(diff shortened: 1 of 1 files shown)


scripts/tests/scripts_regression_tests.py
```diff
@@ -2506,36 +2511,45 @@ def _main_func():
...
         MACHINE = Machines()
 
-
     if "--compiler" in sys.argv:
-        global TEST_COMPILER
         midx = sys.argv.index("--compiler")
         TEST_COMPILER = sys.argv[midx + 1]
         del sys.argv[midx + 1]
         del sys.argv[midx]
+    elif config.has_option("create_test", "COMPILER"):
```
gold2718: No environment variable here? You promised me a `CIME_COMPILER` environment variable! Sniff :(

> I lied - there is no CIME_COMPILER environment variable. Working on this now.

`elif "CIME_COMPILER" in os.environ:`

jedwards4b: Why?  I don't think that we need one.  When I said working on this I meant I was addressing the issue of the PR not that I was creating an environment variable.

gold2718: Fair enough -- I just misunderstood (but who doesn't love more environment variables or 16 ways to do the same thing? :)

jedwards4b:
@gold2718 For future reference - our workflow would have the assignee merge the PR once approved.

gold2718:
@jedwards4b, According to the [official workflow](https://github.com/ESMCI/cime/wiki/CIME-Git-Workflow):

> This merge can be done by either the assignee (reviewer) of the pull request or by the original developer. In the latter case, the developer should wait for an "ok" from the reviewer.

If you think something is missing or incorrect, please update the workflow document -- that is what I follow.

jedwards4b:
You are absolutely correct - am I to assume that you prefer not to do the merge when you are assignee?

gold2718:
No, I just wasn't sure everyone had weighed in and all was ready to go. If I had been paying closer attention, I should have noticed that all reviewers had approved and gone ahead and done the merge. However, the last time I checked in, Mike had not yet done his review.
So the rules are good IMHO in that they allow you to go ahead and do the merge when your assignee drops the ball (again).

Issue #2175, opened 2017-12-29
Need to pass CIMEROOT to CMake-based unit test builds


billsacks:
In order to support flexible directory structures (e.g., clm stuff appearing at the top level or under components/clm), we need to pass CIMEROOT to the CMake-based unit test builds.

PR incoming once scripts_regression_tests finishes.

PR #2176, opened 2017-12-31
Pass CIMEROOT to CMake unit test builds

billsacks:
This is needed to support flexible directory structures

Test suite: scripts_regression_tests on hobart
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#2175

User interface changes?: N

Code review:

### DIFF
--- a/scripts/fortran_unit_testing/run_tests.py
+++ b/scripts/fortran_unit_testing/run_tests.py
@@ -180,6 +180,7 @@ def cmake_stage(name, test_spec_dir, build_optimized, use_mpiserial, mpirun_comm
             "cmake",
             "-C Macros.cmake",
             test_spec_dir,
+            "-DCIMEROOT="+_CIMEROOT,
             "-DCIME_CMAKE_MODULE_DIRECTORY="+os.path.abspath(os.path.join(_CIMEROOT,"src","CMake")),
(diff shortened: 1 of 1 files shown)


scripts/fortran_unit_testing/run_tests.py
```diff
@@ -180,6 +180,7 @@ def cmake_stage(name, test_spec_dir, build_optimized, use_mpiserial, mpirun_comm
             "cmake",
             "-C Macros.cmake",
             test_spec_dir,
+            "-DCIMEROOT="+_CIMEROOT,
```
jedwards4b: Why use an underscore here?  If this is intended to have the same value as the xml variable then shouldn't it have the same name as well?

billsacks: I may be confused about what you're asking here. This `_CIMEROOT` variable is pre-existing in this code; it looks like you introduced it in 00f113e86, where you set:

```python
_CIMEROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "../..")
```

I assume you used a leading underscore to denote that it's private? Am I misunderstanding your question?

jedwards4b: No, that makes sense.  Thanks

Issue #2164, opened 2017-12-22
Bug in shr_cal_advDate when secIn is not 0


billsacks:
I'm adding some unit tests of shr_cal_mod as part of #2145 . In doing so, I came across a bug in `shr_cal_advDate` (the version that accepts real-valued secIN). Here is a unit test that shows the error (I'll be pushing this and some other unit tests to the branch in #2145 soon):

```Fortran
  @Test
  subroutine advDate_int_1dayPlus1sec(this)
    class(TestShrCal), intent(inout) :: this
    integer(i4) :: date_in, date_out
    real(r8) :: sec_in, sec_out

    date_in = 98760317
    sec_in = 100._r8
    call shr_cal_advDate(delta = 86401._r8, units = 'seconds', &
         dateIN = date_in, secIN = sec_in, &
         dateOUT = date_out, secOUT = sec_out, &
         calendar = 'noleap')

    @assertEqual(98760318, date_out)
    @assertEqual(101._r8, sec_out, tolerance=tol)
  end subroutine advDate_int_1dayPlus1sec
```

The above code shows the expected output. The actual result is date_out=98760317, sec_out=86301.

This change fixes this test (along with a similar test that uses a negative increment of -86401):

```diff
diff --git a/src/share/util/shr_cal_mod.F90 b/src/share/util/shr_cal_mod.F90
index eb8297d6a..b84f22140 100644
--- a/src/share/util/shr_cal_mod.F90
+++ b/src/share/util/shr_cal_mod.F90
@@ -786,7 +786,7 @@ subroutine shr_cal_advDate_int(delta,units,dateIN,secIN,dateOUT,secOUT,calendar)
    endif

    ! take secIn into account here since it's real
-   dSec = dSec - secIn
+   dSec = dSec + secIn

    ! i8 math, convert reals to nearest second
    i8dSec = nint(dSec,SHR_KIND_I8)
```

Looking back at the csm_share history in subversion, I see that dSec used to be computed as `dSec + secIn`, but @apcraig changed it to `dSec - secIn` in this tag (which included a major refactor of shr_cal_mod):

```
Originator: tcraig
Date: Feb 19 2012
Model: share
Version: share3_120219
One-line summary: remove use of shr_cal eday methods, replace them (not bit-for-bit)
  merge between calup02_share3_120123 and calup03_share3_120123

M       shr/shr_tInterp_mod.F90
M       shr/shr_stream_mod.F90
M       shr/shr_cal_mod.F90
[...]

Issue #858, opened 2016-11-30
Possible bug in shr_cal_mod


jedwards4b:
Keith Oleson 2013-04-04 14:54:20 MDT
In running a single-point (single gridcell) simulation with clm4_0_69 I found a possible bug in shr-cal_mod.F90 (subroutine shr_cal_advDate).
In this simulation the START_TOD is 72000 seconds.  In the initial call to this subroutine, the following statement is executed:

   ! take secIn into account here since it's real
   dSec = dSec - secIn

Here, dSec on the right side is 0. and secIn is 72000.  So dSec is -72000 and
eventually the datm looks for a time stamp in an atmospheric forcing file that is 72000 seconds before the first time step in the file and the model aborts.

I see that previous versions of this subroutine have:

   dSec = dSec + secIn

When I use this statement, the model runs as expected.  Not sure if this is a robust solution for all situations however.

jedwards4b:
Related issue from CESM_DEVELOPMENT issue 75:

   I found a bug in CESM 1.1.1 with CLM. I used GREGORIAN in drv_in.
   The bug is:

   if we use:
start_tod = 72000
start_ymd = 20120228
stop_option = 'date'
stop_tod = 18000
stop_ymd = 20120301

CLM can move the the stop time.

if we use:
start_tod = 72000
start_ymd = 20120228
stop_option = 'date'
stop_tod = 18000
stop_ymd = 20120229

CLM will get the wrong stop time at (stop_tod = 18000 and  stop_ymd = 20120228).

billsacks:
@olyson I stumbled across this same bug recently. I agree with your suggested fix. I'm implementing this fix, and it should be on cime master soon.

PR #2179, opened 2018-01-02
Fix shr_cal_advDate with non-zero secIN

billsacks:
The sign on the adjustment of secIN was wrong

Test suite: scripts_regression_tests on hobart
Test baseline: N/A
Test namelist changes: none
Test status: expected to be bit for bit, but I'm not positive

Fixes ESMCI/cime#2164
Fixes ESMCI/cime#858

User interface changes?: N

Code review:

### DIFF
--- a/src/share/test/unit/shr_cal_test/test_shr_cal.pf
+++ b/src/share/test/unit/shr_cal_test/test_shr_cal.pf
@@ -122,9 +122,7 @@ contains
   ! Tests of shr_cal_advDate
   ! ------------------------------------------------------------------------
 
-  ! BUG(wjs, 2017-12-22, https://github.com/ESMCI/cime/issues/2164) Skipping this test
-  ! until shr_cal_advDate is fixed
-  ! @Test
+  @Test
   subroutine advDate_int_1dayPlus1sec(this)
     class(TestShrCal), intent(inout) :: this
     integer(i4) :: date_in, date_out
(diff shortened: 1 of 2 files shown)


billsacks:
I have assigned a bunch of reviewers here. Not all of you need to review, but I'd welcome a review by any of you of the core changes here: https://github.com/ESMCI/cime/pull/2179/files#diff-e9c14120311dacc4251cea084a346a97

@apcraig I'd especially welcome your review, since it looks like you were involved with sign change (see #2164 ): can you confirm that this was a bug and not intentional?

@fischer-ncar could you run some tests of this to determine what, if anything, changes answers due to this bug fix? @fischer-ncar @mvertens @apcraig can any of you suggest tests that would be worth running to determine this?

fischer-ncar:
There's no answer changes with hobart prealpha testing.

billsacks:
@fischer-ncar okay thanks for doing that testing! I have no idea what configurations would be affected by this bug; I'm hoping @mvertens or @apcraig would have more insight.

mvertens:
@billsacks - I'm not sure what configurations would be effected. I really
had no part in this implementation. I also hope that @apcraig can provide
some feedback on this. Sorry for passing the buck.

apcraig:
I have been looking at the implementation, and I don't have any insight.  I don't know if there was a reason the sign was changed or if it's a bug.  If it seems to work properly now, and the interface is not really used in CESM this way, then I think you should go ahead and make the change.

billsacks:
Thanks, @mvertens and @apcraig . The interface is used in CESM in some (obscure) cases, as noted in #858 .

@mvertens @apcraig and @fischer-ncar : do you want to officially review and accept this if you're happy with the change, or else remove yourselves as reviewers?

jedwards4b:
I have rebased this branch and resolved the conflict, I can push it back with your ok.

billsacks:
@jedwards4b yes, it would be fine to push the rebased branch back - thank you for checking.

jedwards4b:
@rljacob waiting for your review

PR #2181, opened 2018-01-03
share code cleanup

jedwards4b:
Reduce the number of compiler warnings in building the csm_share library.  Make indentation consistent across files.    Large number of whitespace changes - sorry.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/components/xcpl_comps/xshare/dead_mct_mod.F90
+++ b/src/components/xcpl_comps/xshare/dead_mct_mod.F90
@@ -1,24 +1,24 @@
 module dead_mct_mod
 
   use shr_kind_mod,         only : IN =>SHR_KIND_IN, R8=>SHR_KIND_R8
-  use shr_log_mod,           only: logunit => shr_log_Unit
-  use shr_sys_mod		, only : shr_sys_abort, shr_sys_flush
-  use seq_flds_mod	, only : seq_flds_dom_coord, seq_flds_dom_other
-  use shr_file_mod		, only : shr_file_getlogunit, shr_file_getunit, shr_file_setio, shr_file_getloglevel, &
-                                         shr_file_setlogunit, shr_file_freeunit,shr_file_setloglevel
-  use esmf                      , only : esmf_clock
-  use seq_cdata_mod		, only : seq_cdata, seq_cdata_setptrs
-  use seq_infodata_mod		, only : seq_infodata_type, seq_infodata_getData, seq_infodata_PutData
-  use mct_mod			, only : mct_gsmap, mct_ggrid, mct_avect, mct_ggrid_init, mct_gsmap_lsize, mct_ggrid_lsize, &
-                                         mct_avect_lsize, MCT_AVECT_NRATTR, mct_avect_indexra,mct_avect_zero, &
-                                         mct_ggrid_importiattr, mct_ggrid_importrattr, mct_gsmap_init, mct_aVect_init, &
-                                         mct_gsmap_orderedpoints
-  use dead_data_mod		, only : dead_grid_lat, dead_grid_lon, dead_grid_area, dead_grid_mask, dead_grid_frac, dead_grid_index
-  use dead_mod			, only : dead_setnewgrid
-  use seq_comm_mct		, only : seq_comm_suffix, seq_comm_name,seq_comm_inst, seq_comm_get_ncomps
-  use shr_const_mod		, only : shr_const_pi
-  use seq_timemgr_mod		, only : seq_timemgr_EClockGetData
-  use shr_mpi_mod		, only : shr_mpi_bcast
+  use shr_log_mod,          only: logunit => shr_log_Unit
+  use shr_sys_mod,          only : shr_sys_abort, shr_sys_flush
+  use seq_flds_mod,         only : seq_flds_dom_coord, seq_flds_dom_other
+  use shr_file_mod,         only : shr_file_getlogunit, shr_file_getunit, shr_file_setio, shr_file_getloglevel, &
+       shr_file_setlogunit, shr_file_freeunit,shr_file_setloglevel
+  use esmf,                 only : esmf_clock
+  use seq_cdata_mod,        only : seq_cdata, seq_cdata_setptrs
+  use seq_infodata_mod,     only : seq_infodata_type, seq_infodata_getData, seq_infodata_PutData
+  use mct_mod,              only : mct_gsmap, mct_ggrid, mct_avect, mct_ggrid_init, mct_gsmap_lsize, mct_ggrid_lsize, &
+       mct_avect_lsize, MCT_AVECT_NRATTR, mct_avect_indexra,mct_avect_zero, &
+       mct_ggrid_importiattr, mct_ggrid_importrattr, mct_gsmap_init, mct_aVect_init, &
+       mct_gsmap_orderedpoints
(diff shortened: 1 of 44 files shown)


jedwards4b:
Add ?w=1 to the URL to see the diff with whitespace ignored.

jgfouca:
I'm unqualified to review (need to learn fortran one of these days), so I rubber-stamped.

billsacks:
@jedwards4b can you please document the indentation settings you used here? From a quick spot-check, I think they match the settings I use, which I'm pretty sure I originally based off of what seemed to be most common across CESM, but it would be good to confirm that before accepting this major whitespace change.

Here are my settings:

- Do/If/Type indent (also applies to select case, where, interface, etc.): 3

- Program indent (applies to program, module, subroutine and function): 2

- Continuation indent: 5

- Associate indent (may be irrelevant in cime): 0

Does that match what you used?

jedwards4b:
f90-associate-indent = 2
f90 continuation indent = 5
f90 critical indent = 2
f90 do indent = 3
f90 if indent = 3
f90 program indent = 2
f90 type indent = 3

src/components/xcpl_comps/xshare/dead_mct_mod.F90
```diff
@@ -447,15 +448,15 @@ subroutine dead_run_mct( model, EClock, cdata, x2d, d2x)
 
     lsize = mct_avect_lsize(x2d)
     nflds_d2x = mct_avect_nRattr(d2x)
-    nflds_x2d = mct_avect_nRattr(x2d)
+    !    nflds_x2d = mct_avect_nRattr(x2d)
```
billsacks: Why comment this out instead of just removing it?

src/drivers/mct/shr/seq_comm_mct.F90
```diff
@@ -653,7 +648,7 @@ subroutine seq_comm_setcomm(ID,pelist,nthreads,iname,inst,tinst)
 
     call shr_mpi_chkerr(ierr,subname//' mpi_comm_create mpigrp')
 
-    ntasks = ((pelist(2,1) - pelist(1,1)) / pelist(3,1)) + 1
+!    ntasks = ((pelist(2,1) - pelist(1,1)) / pelist(3,1)) + 1
```
billsacks: Again, why not just remove this if it's unneeded?

src/drivers/mct/shr/seq_comm_mct.F90
```diff
@@ -1129,14 +1124,15 @@ end subroutine seq_comm_setnthreads
...
     integer :: omp_get_num_threads
-    character(*),parameter :: subName =   '(seq_comm_getnthreads) '
 
     seq_comm_getnthreads = -1
-#ifdef _OPENMP
 !$OMP PARALLEL
     seq_comm_getnthreads = omp_get_num_threads()
 !$OMP END PARALLEL
+#else
+    seq_comm_getnthreads = -1
```
billsacks: Redundant with the initialization on line 1130 - you can remove one or the other

src/share/esmf_wrf_timemgr/ESMF_CalendarMod.F90
```diff
@@ -312,10 +312,10 @@ SUBROUTINE initdaym
   monthbdysleap(0)%Sn = 0
   monthbdysleap(0)%Sd = 0
   DO i = 1,MONTHS_PER_YEAR
-    DO j = 1,mdayleap(i)
-      daymleap(m) = i
-      m = m + 1
-    ENDDO
+!    DO j = 1,mdayleap(i)
```
billsacks: Again, why not remove this block of unused code?

src/share/esmf_wrf_timemgr/ESMF_TimeIntervalMod.F90
```diff
@@ -782,14 +779,14 @@ subroutine ESMFold_TimeIntervalGetString(timeinterval, TimeString, rc)
       negative = ( signnormtimeint( timeInterval ) == -1 )
       IF ( negative ) THEN
         iS = -timeinterval%basetime%S
-        iSn = -timeinterval%basetime%Sn
+!        iSn = -timeinterval%basetime%Sn
```
billsacks: This and the next few lines: more commented-out code that could be removed

src/share/streams/shr_dmodel_mod.F90
```diff
@@ -1576,7 +1552,6 @@ subroutine shr_dmodel_mapSet_dest(smatp,ggridS,gsmapS,nxgS,nygS,ggridD,gsmapD,nx
     type(mct_aVect) :: AVl
     type(mct_aVect) :: AVg
 
-    character(len=32) :: lstrategy
```
billsacks: Since it looks like the strategy optional argument is unused, it should be removed as an argument to this routine to avoid confusion.

src/share/streams/shr_strdata_mod.F90
```diff
@@ -191,7 +190,6 @@ subroutine shr_strdata_init(SDAT,mpicom,compid,name,scmmode,scmlon,scmlat, &
   integer(IN) :: nu,nv         ! u,v index
   integer(IN) :: my_task,npes  ! my task, total pes
   integer(IN),parameter :: master_task = 0
-  character(CS) :: lname       ! local name
```
billsacks: Since it looks like the name optional argument is unused, it should be removed as an argument to this routine to avoid confusion

jedwards4b:
I cannot remove unused optional arguments without assuring that no component calls the subroutine with that argument.  

In several cases there was commented code already in source, so for example an equation was commented out but a variable only used in that equation was left in.  I felt it important to leave the variable and the equation in place but comment out the variable as well.

There is certainly a lot more that could be done here, I am not attempting perfection, just an incremental improvement on the current code.     

I can submit a PR with whitespace changes only, then rebase this one so that only the code changes appear - is that acceptable?

billsacks:
@jedwards4b thanks for the reply.

1.

> I can submit a PR with whitespace changes only, then rebase this one so that only the code changes appear - is that acceptable?

Yes, that would be great – thank you.

2.

> In several cases there was commented code already in source, so for example an equation was commented out but a variable only used in that equation was left in. I felt it important to leave the variable and the equation in place but comment out the variable as well.

Okay, if these were left in for a reason, I'm okay with it.

3. 

> I cannot remove unused optional arguments without assuring that no component calls the subroutine with that argument.

Fair enough. But can you please add comments in those routines saying that these arguments are deprecated and should not be used?

4. The one remaining thing that you didn't address in your reply is my comment on `src/drivers/mct/shr/seq_comm_mct.F90`: there is now a redundant setting of `seq_comm_getnthreads` there.

src/drivers/mct/shr/seq_comm_mct.F90
jedwards4b: This is not redundant as the compiler will only see one of them.

billsacks: Oh, sorry, I misread this code. Really what I should have said is that line 1130 is unnecessary because it will get set on line 1132. But this isn't a big deal.

jedwards4b:
I'm not happy with the way this is rebasing, think I'll start over from master...

PR #2187, opened 2018-01-04
Fix some misuses of MAX_(MPI)TASKS...

### LINKED ISSUE #2151: use MAX_MPI_TASKS_PER_NODE for core limit in test_schedular
The current scheme fails to run tests on small systems that should be okay even though they overload the node.

jgfouca:
PET test needed to halve MAX_MPITASKS in addition to MAX_TASKS.

test_scheduler needed to use MAX_MPITASKS to set degree of task-level parallelism
and use MAX_TASKS to compute the proc/thread pool.

Test suite: code_checker, PET test by-hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2151 

User interface changes?: N

Update gh-pages html (Y/N)?:N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/SystemTests/pet.py
+++ b/scripts/lib/CIME/SystemTests/pet.py
@@ -46,6 +46,7 @@ class PET(SystemTestsCompareTwo):
         # to the batch submission, things break. Setting MAX_TASKS_PER_NODE to half of
         # it original value prevents this.
         self._case.set_value("MAX_TASKS_PER_NODE", int(self._case.get_value("MAX_TASKS_PER_NODE") / 2))
+        self._case.set_value("MAX_MPITASKS_PER_NODE", int(self._case.get_value("MAX_MPITASKS_PER_NODE") / 2))
 
         # Need to redo case_setup because we may have changed the number of threads
         case_setup(self._case, reset=True)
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/SystemTests/pet.py
```diff
@@ -46,6 +46,7 @@ def _case_two_setup(self):
         # to the batch submission, things break. Setting MAX_TASKS_PER_NODE to half of
         # it original value prevents this.
         self._case.set_value("MAX_TASKS_PER_NODE", int(self._case.get_value("MAX_TASKS_PER_NODE") / 2))
+        self._case.set_value("MAX_MPITASKS_PER_NODE", int(self._case.get_value("MAX_MPITASKS_PER_NODE") / 2))
```
amametjanov: PET tests are breaking on Cori-KNL and Theta after this change. I'm wondering if the first division (`MAX_TASKS_PER_NODE/2`) is still necessary. It looks like this was needed for skybridge. If it's still necessary, can there be a skybridge-specific change in config_machines.xml using a `$SHELL{...}` script? Skybridge has 16 cores, so the first 16x2 run would be on 2 nodes and the second 16x1 run would be on 1 node and mpiexec would not exceed procs-per-node?

jgfouca: @amametjanov , sure.

jgfouca: @amametjanov this will be fixed in the next CIME update.

PR #2192, opened 2018-01-04
reducing warnings in fortran

jedwards4b:
This reduces the number of unused local variables and possibly uninitialized variables in the code.
My plan is to do several of these code clean up pr's in the hope that smaller pieces will be more manageable.

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/share/streams/shr_stream_mod.F90
+++ b/src/share/streams/shr_stream_mod.F90
@@ -257,7 +257,11 @@ contains
     !--- read data ---
     read(nUnit,'(a)',END=999) str
     call shr_string_leftalign_and_convert_tabs(str)
-    strm%dataSource = str(1:SHR_KIND_CS)
+    if(len_trim(str)<= len(strm%dataSource)) then
+       strm%dataSource = str(1:len(strm%dataSource))
+    else
+       call shr_sys_abort('dataSource too long for variable ' // errMsg(sourcefile, __LINE__))
+    endif
     if (debug>0 .and. s_loglev>0) write(s_logunit,F00) '  * format = ', trim(strm%dataSource)
 
     close(nUnit)
@@ -492,7 +496,11 @@ contains
           call shr_stream_abort(subName//"ERROR: no time variable name")
        else
           call shr_string_listGetName (fldListFile,n,substr,rc)
-          strm%domTvarName = subStr(1:SHR_KIND_CS)
+          if(len_trim(substr) <= len(strm%domTVarName)) then
+             strm%domTvarName = subStr(1:len(strm%domTVarName))
+          else
+             call shr_stream_abort(subName//"ERROR: string too long for variable name")
+          endif
        endif
        !--- get longitude variable name ---
        n = shr_string_listGetIndexF(fldListModel,"lon")
@@ -502,7 +510,11 @@ contains
           call shr_stream_abort(subName//"ERROR: no lon variable name")
        else
           call shr_string_listGetName (fldListFile,n,substr,rc)
-          strm%domXvarName = subStr(1:SHR_KIND_CS)
+          if(len_trim(substr) <= len(strm%domXVarName)) then
+             strm%domXvarName = subStr(1:len(strm%domXVarName))
+          else
+             call shr_stream_abort(subName//"ERROR: string too long for variable name")
+          endif
        endif
        !--- get latitude variable name ---
        n = shr_string_listGetIndexF(fldListModel,"lat")
@@ -512,7 +524,11 @@ contains
           call shr_stream_abort(subName//"ERROR: no lat variable name")
        else
           call shr_string_listGetName (fldListFile,n,substr,rc)
-          strm%domYvarName = subStr(1:SHR_KIND_CS)
+          if(len_trim(substr) <= len(strm%domYVarName)) then
+             strm%domYvarName = subStr(1:len(strm%domYVarName))
+          else
+             call shr_stream_abort(subName//"ERROR: string too long for variable name")
+          endif
        endif
        !--- get vertical variable name ---
        n = shr_string_listGetIndexF(fldListModel,"hgt")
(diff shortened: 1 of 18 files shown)


src/share/util/shr_orb_mod.F90
```diff
@@ -433,7 +433,7 @@ SUBROUTINE shr_orb_params( iyear_AD , eccen  , obliq , mvelp     ,     &
     real   (SHR_KIND_R8) :: obsum   ! Obliquity series summation
     real   (SHR_KIND_R8) :: cossum  ! Cos series summation for eccentricity/fvelp
     real   (SHR_KIND_R8) :: sinsum  ! Sin series summation for eccentricity/fvelp
-    real   (SHR_KIND_R8) :: fvelp   ! Fixed vernal equinox long of perihelion
+    real   (SHR_KIND_R8) :: fvelp = 0.0_SHR_KIND_R8   ! Fixed vernal equinox long of perihelion
```
billsacks: This initialization should be done in the code, not in the variable declaration: doing this in the variable declaration implicitly gives it the `save` attribute, which is generally a bad thing.

So please move this initialization to just before this block of code:

```fortran
       if (abs(cossum) .le. 1.0E-8_SHR_KIND_R8) then
          if (sinsum .eq. 0.0_SHR_KIND_R8) then
             fvelp = 0.0_SHR_KIND_R8
          else if (sinsum .lt. 0.0_SHR_KIND_R8) then
             fvelp = 1.5_SHR_KIND_R8*pi
          else if (sinsum .gt. 0.0_SHR_KIND_R8) then
             fvelp = .5_SHR_KIND_R8*pi
          endif
       else if (cossum .lt. 0.0_SHR_KIND_R8) then
          fvelp = atan(sinsum/cossum) + pi
       else if (cossum .gt. 0.0_SHR_KIND_R8) then
          if (sinsum .lt. 0.0_SHR_KIND_R8) then
             fvelp = atan(sinsum/cossum) + 2.0_SHR_KIND_R8*pi
          else
             fvelp = atan(sinsum/cossum)
          endif
       endif
```

or (better in my mind) change this line:

```fortran
       else if (cossum .gt. 0.0_SHR_KIND_R8) then
```

to:

```fortran
       else  ! cossum > 1.e-8
```

(I'm pretty sure that the logic ends up being the same with that change, though you could double-check it. It makes it more clear to the reader - as well as the compiler- that one of those conditional blocks will always be executed.)

src/share/util/shr_flux_mod.F90
```diff
@@ -843,7 +849,7 @@ real(R8)    :: psix2   ! stability function at ztref reference height
             Hb = (Qdel/rcpocn)+(Fd*betaS/alphaT)
             Hb = min(Hb , 0.0_R8)
             lambdaV = lambdaC*(1.0_R8 + ( (0.0_R8-Hb)*16.0_R8*molvisc(tBulk(n))* &
-                 shr_const_g*alphaT*molPr(tBulk(n))**2/ustarw**4)**0.75)**(-1/3)
+                 shr_const_g*alphaT*molPr(tBulk(n))**2/ustarw**4)**0.75_R8)**(-1.0_R8/3.0_R8)
```
jedwards4b: These are in the diurnal code which is not currently exercised, do we really need another pr?

jedwards4b:
@billsacks I think I've addressed all comments, and moved bug fix to #2195

src/share/streams/shr_stream_mod.F90
```diff
@@ -257,7 +257,11 @@ subroutine shr_stream_init(strm,infoFile,yearFirst,yearLast,yearAlign,taxMode,rc
     !--- read data ---
     read(nUnit,'(a)',END=999) str
     call shr_string_leftalign_and_convert_tabs(str)
-    strm%dataSource = str
+    if(len_trim(str)<= len(strm%dataSource)) then
+       strm%dataSource = str(1:len(strm%dataSource))
+    else
+       call shr_sys_abort('dataSource too long for variable ' // errMsg(sourcefile, __LINE__))
```
billsacks: Sorry, my comment wasn't clear here. `Using errMsg(sourcefile, __LINE__)` will only work if you define sourcefile and add a use statement for errMsg. Why don't you just change this to match the others:

```Fortran
call shr_stream_abort(subName//" ERROR: dataSource too long for variable")
```

src/share/util/shr_flux_mod.F90
```diff
@@ -355,6 +358,9 @@ real(R8)    :: ugust  ! function: gustiness as a function of convective rainfall
            tstar = rh * delt
            qstar = re * delq
         enddo
+        if (iter < 1) then
+           call shr_sys_abort('No iterations performed ' //  // errMsg(sourcefile, __LINE__))
```
billsacks: Sorry, my comment wasn't explicit enough here. For this to work, you need to define sourcefile at the top of the file:

```fortran
  character(len=*), parameter, private :: sourcefile = &
       __FILE__
```

and also add a `use` statement:

```fortran
  use shr_log_mod  , only : errMsg => shr_log_errMsg
```
  
Also, I see I had a typo here: there are two instances of `//`

src/share/util/shr_flux_mod.F90
```diff
@@ -939,7 +945,9 @@ real(R8)    :: psix2   ! stability function at ztref reference height
             qstar = re * delq
 
          ENDDO   ! end iteration loop
-
+         if (iter < 1) then
+            call shr_sys_abort('No iterations performed ' //  // errMsg(sourcefile, __LINE__))
```
billsacks: As above, some more changes are needed for this to compile.

jedwards4b:
@billsacks I've made the changes, rerunning tests now.

jedwards4b:
testing complete - all tests pass

PR #2195, opened 2018-01-05
exp(1/3) is identically 1

jedwards4b:
Correct equations of the form **(1/3) to be **(1.0_R8/3.0_R8) to evaluate correctly.   

Test suite:  Not sure what to do here - this code is not currently used.
Test baseline: 
Test namelist changes: 
Test status: roundoff

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/share/util/shr_flux_mod.F90
+++ b/src/share/util/shr_flux_mod.F90
@@ -462,6 +462,12 @@ SUBROUTINE shr_flux_atmOcn(nMax  ,zbot  ,ubot  ,vbot  ,thbot ,  prec_gust, gust_
 
 END subroutine shr_flux_atmOcn
 
+real(R8) elemental function cuberoot(a)
+  real(R8), intent(in) :: a
+  real(R8), parameter :: one_third = 1._R8/3._R8
+  cuberoot = sign(abs(a)**one_third, a)
+end function cuberoot
+
 !===============================================================================
 ! !BOP =========================================================================
 !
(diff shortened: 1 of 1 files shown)


billsacks:
@mvertens What do you think we should do here: Should relevant scientists be notified? If so, who? Should we ask them to do some scientific tests of the impact of this bug before or after merging this to master?

jedwards4b:
In this equation:
            lambdaV = lambdaC*(1.0_R8 + ( (0.0_R8-Hb)*16.0_R8*molvisc(tBulk(n))* &
                 shr_const_g*alphaT*molPr(tBulk(n))**2/ustarw**4)**0.75)**(-1/3)

and this one:
             Kdiff = shr_const_karman*ustarw*shr_const_zsrflyr *(1.0_R8-7.0_R8*doL)**(1/3)


**(1/3) evaluates equivalently to 1.0 since integer arithmetic means 1/3 = 0.

I've tried some tests where I corrected this by replace 1/3 with: 
real(r8), parameter :: onethird = 1.0_R8/3.0_R8

but this results in the model blowing up (not locally but in rrtmg)

I'm not quite sure how to proceed here.   We aren't currently using this code by default, but I don't want to leave it there when we know it's buggy.

mvertens:
@billsacks - @jedwards4b has already notified the relevant scientists.
@jedwards4b - I concur with Bill - great catch. Thanks so much.

rljacob:
What option turns on this code?

jedwards4b:
@rljacob diurnal_flux=.true. in drv_in namelist.

billsacks:
@jedwards4b so my understanding is that I should hold off on merging this until we get some more input from others. Is that correct?

jedwards4b:
From Bill Large:  lambdaC= 6.0 and lambdaV is a correction which under physical conditions is nearly always lambdaV=6.5 so we should remove this calculation and simply use lambdaV = 6.5_R8

jedwards4b:
I finally have a formulation that doesn't blowup in the first 5 days. 
I ran SMS_D.f09_g17.B1850.cheyenne_intel.allactive-defaultio with 
flux_diurnal = .true.   in user_nl_cpl

This really needs a science review, but it may be better than what is on master now.

jedwards4b:
Rebased and changed comment to restart travis-ci testing.

PR #2196, opened 2018-01-05
Remove concept of RUN_WITH_SUBMIT

jgfouca:
It added a good deal of complexity and broke some usecases just for
the sake of protecting the user from themself. Instead, make batch
script files hidden from user so that they don't need to be
protected in the first place.

Test suite: scripts_regression_tests (on skybridge, batch machine)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: Y, renaming of batch scripts

Update gh-pages html (Y/N)?:N

### DIFF
--- a/scripts/lib/CIME/case_setup.py
+++ b/scripts/lib/CIME/case_setup.py
@@ -9,7 +9,7 @@ from CIME.preview_namelists import create_dirs, create_namelists
 from CIME.XML.env_mach_pes  import EnvMachPes
 from CIME.XML.machines      import Machines
 from CIME.BuildTools.configure import configure
-from CIME.utils             import get_cime_root, run_and_log_case_status, get_model
+from CIME.utils             import get_cime_root, run_and_log_case_status, get_model, get_batch_script_for_job
 from CIME.test_status       import *
 
 import shutil, six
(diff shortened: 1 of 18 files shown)


jgfouca:
@billsacks , the comments in the st_archive template file say this:

```
template to create a case short term archiving script. This should only ever be called                                                                                                                                                                                             
by case.submit when on batch system. This script only exists as a way of providing                                                                                                                                                                                                 
batch directives. Use case.submit from the command line to run your case. 
```

So, that indicated to me that it should not be called. Perhaps it's a copy/paste error?

As to 2, yes, that's correct. The only UI change is file names of stuff they should have been touching anyway.

scripts/lib/CIME/case_setup.py
```diff
@@ -83,21 +83,18 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False):
...
         if not test_mode:
             # rebuild the models (even on restart)
             case.set_value("BUILD_COMPLETE", False)
 
             # backup and then clean test script
-            if os.path.exists("case.test"):
-                os.remove("case.test")
-                logger.info("Successfully cleaned test script case.test")
-
-            if os.path.exists("case.testdriver"):
```
jgfouca: case.testdriver is ancient legacy stuff, not used anymore.

jedwards4b:
@billsacks you can run just the st_archive script through case.submit like
case.submit --job case.st_archive, you can also do this without submiting
case.submit --job case.st_archve --no-batch

billsacks:
I spoke with @bertinia and she said that there has been the expectation that you should be able to run `./case.st_archive` directly (i.e., the comment you pointed to, @jgfouca , was indeed probably a copy/paste error). @jedwards4b - I see your point about this alternative way to run it, but that sounds a lot harder and less intuitive for users than just running `./case.st_archive`.

So my feeling is that you should revert `case.st_archive` to its old name (without a leading `.`) - just hiding `case.run` and `case.test`.

jgfouca:
@billsacks , the only problem with that approach is that we would then be inconsistent in our naming of batch scripts. I'm leaning towards @jedwards4b and for making case.submit the one entry-point to running anything.

jgfouca:
An alternative would be to provide wrapper script, case.st_archive, which just calls case.submit --job case.st_archive.

jedwards4b:
We could add a short case.st_archive script that just does this:
case.submit --job case.st_archive --no-batch

jgfouca:
@jedwards4b , on it, i see we're on the same wavelength here.

jedwards4b:
hmm - does renaming these cause the --job argument to case.submit to break - @jgfouca did you happen to test this?

jgfouca:
@jedwards4b , I ran the full suite on a batch system. I think I caught all the spots where there was an assumption that the job name matched the script name.

billsacks:
@jgfouca @jedwards4b I'm happy with the approach you mention:

> We could add a short case.st_archive script that just does this:
case.submit --job case.st_archive --no-batch

jgfouca:
OK, I've added the wrapper script we talked about and tested it by-hand.

rljacob:
case.st_archive is run by hand all the time.

jgfouca:
@rljacob , that's OK; it's still supported.

PR #2202, opened 2018-01-06
Refactor seq_diag_mod.F90 step one

jedwards4b:
I think I need to do this in several PR's in this one I have renamed variable if -> nf and variables
da, di, dl, do, dg, dr -> ca_a, ca_i, ca_l, ca_o, ca_g, ca_r (cell area) I have tried to avoid whitespace changes in this one and will present another pr with only whitespace changes.  

Test suite: scripts_regression_tests.py, scripts_regression_tests.py --compiler nag on hobart
       compared coupler log of SMS_Ld3.f09_g17.B1850.cheyenne_intel with INFO_DBUG=2 
       against master, all comm_diag output was identical.
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/drivers/mct/main/seq_diag_mct.F90
+++ b/src/drivers/mct/main/seq_diag_mct.F90
@@ -27,7 +27,7 @@
 ! !INTERFACE: ------------------------------------------------------------------
 
 module seq_diag_mct
-! !USES:
+  ! !USES:
 
    use shr_kind_mod, only: r8 => shr_kind_r8, in=>shr_kind_in
    use shr_kind_mod, only: i8 => shr_kind_i8,  cl=>shr_kind_cl, cs=>shr_kind_cs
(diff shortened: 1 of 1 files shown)


billsacks:
Thank you for the renames of `if` and `do`! It does look like about half of the changes here are whitespace, but I'm okay with that since so much of this file is touched through the variable renames anyway.

My one question/request is: Have you done any tests where you compare against baselines to ensure bfb, to verify that these changes were all done correctly? (Actually, I'm not sure what could be affected by bugs in seq_diag_mod, so I'm not sure myself what tests would be most appropriate to test these changes.)

jedwards4b:
I plan to do that testing prior to merging but it's not done yet.  I'm trying to resolve the merge conflict and can't find a way to do it without introducing even more whitespace changes - I'm sorry but I think it may be unavoidable.

jedwards4b:
I'm not sure how I could avoid the whitespace changes, you can view diffs without whitespace changes by adding ?w=1 to the URL.

billsacks:
@jedwards4b - since @apcraig approves, I'll merge this once you've done final testing.

jedwards4b:
I believe that the correct test is a comparison of cpl logs with INFO_DBUG=2, I did this in
SMS_Ld3.f09_g17.B1850.cheyenne_intel and they are the same.

PR #2205, opened 2018-01-07
Use caching to speed-up retrieval of complaint entries

jgfouca:
get_children and scan_children for EnvBase and it's subclasses has been dramatically optimized. When the XML object is constructed, it loads all the groups and entries into python dictionaries (caches) so that compliant data lookups are much faster.

Based on my profiling of case.setup on my laptop, this reduces the time spent in scan_children from about 50% of the total runtime to 2%.

This reduces case.setup time on my laptop from 10s to about 5s. It also reduces total scripts_regression_tests time on melvin from 124 minutes to 104 minutes.

Also:

1) Remove no_validate from generic_xml.get_children. That was for debugging the XML APL consolidation.
2) Change case_clone to a much safer implementation via the new change_file method. We should  never be modifying data members of an object from outside
3) Restructure env_mach_specific so at least a couple entries are compliant.
4) Minor robustness improvements for environment loading 
5) Remove very expensive USERDEFINED_required_build check

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/XML/env_base.py
+++ b/scripts/lib/CIME/XML/env_base.py
@@ -19,10 +19,58 @@ class EnvBase(EntryID):
             fullpath = os.path.join(case_root, infile)
 
         EntryID.__init__(self, fullpath, schema=schema)
+
+        self._id_map = {}    # map id directly to nodes
+        self._group_map = {} # map group name to entry id dict
         if not os.path.isfile(fullpath):
             headerobj = Headers()
             headernode = headerobj.get_header_node(os.path.basename(fullpath))
             self.add_child(headernode)
+        else:
+            # Setup cache
+            group_elems = self.get_children("group")
+            for group_elem in group_elems:
+                group_name = self.get(group_elem, "id")
+                expect(group_name not in self._group_map, "Repeat group '{}'".format(group_name))
+                group_map = {}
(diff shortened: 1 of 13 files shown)


scripts/lib/CIME/XML/env_base.py
```diff
@@ -19,10 +19,69 @@ def __init__(self, case_root, infile, schema=None):
...
+        self._group_map = None
+
         if not os.path.isfile(fullpath):
             headerobj = Headers()
             headernode = headerobj.get_header_node(os.path.basename(fullpath))
             self.add_child(headernode)
+        else:
+            self._setup_cache()
+
+    def _setup_cache(self):
```
jgfouca: Important part here.

fischer-ncar:
I've started the prealpha tests on cheyenne.

billsacks:
@jgfouca thanks for taking on this performance optimization work!

fischer-ncar:
The prealpha tests passed, no unexpected failures.

jedwards4b:
@mvertens waiting for your review

jgfouca:
Thanks for the help, all!

PR #2212, opened 2018-01-09
Jgfouca/branch for acme split 2018 01 09

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b


jgfouca:
Tests still running.

@agsalin @mt5555 @rljacob , it appears the integration is working again.

scripts/lib/acme_cime_mgmt.py
```diff
@@ -174,7 +174,7 @@ def acme_cime_merge(resume):
         pr_branch = get_branch_from_tag(new_merge_tag)
 
     try:
-        run_cmd_no_fail("git push -u origin {}".format(pr_branch), verbose=True)
+        run_cmd_no_fail("git push -u origin {}".format(pr_branch))
```
jgfouca: This is not right.

src/drivers/mct/main/prep_ocn_mod.F90
```diff
@@ -48,6 +48,8 @@ module prep_ocn_mod
 
   public :: prep_ocn_get_x2oacc_ox
   public :: prep_ocn_get_x2oacc_ox_cnt
+  ! Sarat: Dummy variable added to workaround PGI compiler bug (PGI 17.9) as of Oct 23, 2017
+  public :: dummy_pgibugfix
```
jedwards4b: This is ugly - can we add 
#ifdef CPRPGI around it?

jgfouca:
These changes were mostly exactly what I expected, the vast majority being xml changes to config/acme.

src/drivers/mct/main/prep_ocn_mod.F90
jgfouca: @amametjanov , can you comment here?

amametjanov: Looks like it's more specific than CPRPGI, was needed for v17.9  on the SummitDev machine: see https://github.com/ACME-Climate/ACME/pull/1908.

jedwards4b: We can at least limit it to PGI builds with that cppdef.

amametjanov: There is a new `SUMMITDEV_PGI` macro in that PR, so `#ifdef SUMMITDEV_PGI` should be just the right scope.

src/drivers/mct/main/prep_ocn_mod.F90
```diff
@@ -48,6 +48,10 @@ module prep_ocn_mod
 
   public :: prep_ocn_get_x2oacc_ox
   public :: prep_ocn_get_x2oacc_ox_cnt
+#ifdef SUMMITDEV_PGI
```
jgfouca: OK, I added the preprocessor restriction.

amametjanov: Also need to add this to references in `cime/src/drivers/mct/main/seq_rest_mod.F90`.

jgfouca: @amametjanov I don't see any references to "dummy" or "pgi" in that file.

amametjanov: Oh, the latest cime merge took out those `dummy_pgibugfix = associated(x2oacc_ox)` workarounds from that PR. @sarats may want to re-check with the latest E3SM master.

jgfouca: @amametjanov you're right, it was lost in the recent ESMCI integration, I will put it back.

jgfouca: @amametjanov @sarats , to avoid this in the future, it might be best to make driver changes under ESMCI.

sarats: @jgfouca Understood, thanks for fixing.

PR #2216, opened 2018-01-10
Correct bld output

jedwards4b:
Fixes several issues in the build: stderr was suppressed for component and exe logs, uncompressed log files were left behind after compressing, library (mct, pio, csm_share) log files were not compressed.   

Test suite:  scripts_regression_test.py
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/buildlib.py
+++ b/scripts/lib/CIME/buildlib.py
@@ -82,4 +82,5 @@ def run_gmake(case, compclass, libroot, bldroot, libname="", user_cppdefs=""):
     if user_cppdefs:
         cmd = cmd + "USER_CPPDEFS='{}'".format(user_cppdefs )
 
-    run_cmd_no_fail(cmd, combine_output=True)
+    stat, out, _ = run_cmd(cmd, combine_output=True)
+    print(out)
--- a/scripts/lib/CIME/buildlib.py
+++ b/scripts/lib/CIME/buildlib.py
@@ -82,5 +82,5 @@ def run_gmake(case, compclass, libroot, bldroot, libname="", user_cppdefs=""):
     if user_cppdefs:
         cmd = cmd + "USER_CPPDEFS='{}'".format(user_cppdefs )
 
-    stat, out, _ = run_cmd(cmd, combine_output=True)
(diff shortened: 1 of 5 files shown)


scripts/lib/CIME/buildlib.py
```diff
@@ -82,4 +82,5 @@ def run_gmake(case, compclass, libroot, bldroot, libname="", user_cppdefs=""):
     if user_cppdefs:
         cmd = cmd + "USER_CPPDEFS='{}'".format(user_cppdefs )
 
-    run_cmd_no_fail(cmd, combine_output=True)
+    _, out, _ = run_cmd(cmd, combine_output=True)
```
jedwards4b: buildlib is called by run_cmd with stdout redirected to a file and stderr combined with stdout.  Calling run_cmd again inside that environment with no stderr or stdout args causes output to get lost.

src/drivers/mct/cime_config/buildexe
```diff
@@ -49,7 +49,7 @@ def _main_func():
     cmd = "%s exec_se -j %d EXEC_SE=%s MODEL=%s LIBROOT=%s -f %s "\
         % (gmake, gmake_j, exename, "driver", libroot, makefile)
 
-    rc, out, err = run_cmd(cmd)
+    rc, out, err = run_cmd(cmd, combine_output=True)
```
jgfouca: you can remove err here (replace with "_") since it will never contain anything due to combined output.

scripts/lib/CIME/buildlib.py
```diff
@@ -82,4 +82,5 @@ def run_gmake(case, compclass, libroot, bldroot, libname="", user_cppdefs=""):
...
+    print(out)
```
jgfouca: Why print instead of logger.info?

jedwards4b: I think print is appropriate here because we want this output regardless of logger level.

Issue #2222, opened 2018-01-10
grid in the XML schema need to be xs:string rather than xs:NCName to allow for CLM 1x1_* grids


ekluzek:
In both config_grids_v2.xml and testlist.xsd in config/xml_schemas the type for grid needs to be xs:string rather than xs:NCName, because NCName doesn't allow the name to start with a number. If I do xmllint on the CLM testlist it shows these as errors because of this problem.

ekluzek:
OK, config_grids_v2.xsd is fine since it uses NMTOKEN. It's only testlist.xsd that needs to be changed, and should be changed to NMTOKEN.

ekluzek:
PR#2232 fixed this issue

PR #2232, opened 2018-01-12
Change grid in testlist schema to NMTOKEN

ekluzek:
Change the grid in the testlist XSD schema to NMTOKEN from NCName, in order
to allow the CLM grids for single point such as 1x1_brazil. This allows
the CLM testlist to be upgraded to the version 2.0 format.

Test suite: regression test
Test status: bit for bit

Fixes: 2222


rljacob:
For issues to close automatically, it has to be "Fixes #2222"  You need the "#" sign and no ":".

PR #2234, opened 2018-01-16
Implement walltime stash to support more-accurate walltimes

jgfouca:
This is for ACME only, but I'd still like feedback on the merits of the idea.

The idea is to make a "stash" of walltimes in the baseline area under a hardcoded baseline name "walltimes". Each time a test is run, store the time it took in a special file. This file will be used for more-accurate walltime requests on the batch system whenever the test is run again.

Test suite: scripts_regression_tests (anvil)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -3,13 +3,14 @@ Base class for CIME system tests
 """
 from CIME.XML.standard_module_setup import *
 from CIME.XML.env_run import EnvRun
-from CIME.utils import append_testlog
+from CIME.utils import append_testlog, get_model
 from CIME.case_setup import case_setup
 from CIME.case_run import case_run
 from CIME.case_st_archive import case_st_archive
 from CIME.test_status import *
 from CIME.check_lockedfiles import *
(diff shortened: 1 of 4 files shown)


jgfouca:
@jedwards4b , i'm thinking of adding some complexity to map the requested runtime to its wiggle-room multiplier. You're idea is pretty good too and simpler. It may result in timeouts for really big tests though.

jgfouca:
OK, I'm pretty happy with this now.

jgfouca:
@jedwards4b I almost forgot, I'm also worried about multi-submit jobs, especially in the case where the second job is much shorter than the first.

jgfouca:
@jedwards4b , I believe this is ready to merge unless @rljacob or @mvertens want to chime in.

rljacob:
I did have a question about where the "walltimes" file lives.  Is it directly under BASELINE_ROOT?

jgfouca:
@rljacob , yes, it lives under $BASELINE_ROOT with a baseline name of "walltimes", so
$BASELINE_ROOT/walltimes/$CASE/walltimes

PR #2243, opened 2018-01-23
Cache file reads to avoid re-parsing XML files

jgfouca:
Use a class member in GenericXML to store a map of filepath to xml etree. This way, XML files will only ever by read once in a CIME process.

This will further increase the performance of CIME. In my profiling, this increase performance of case.setup by 13%.

Remove extremely unsafe "restore" method in check_lockedfiles.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/check_lockedfiles.py
+++ b/scripts/lib/CIME/check_lockedfiles.py
@@ -7,6 +7,7 @@ from CIME.XML.env_build import EnvBuild
 from CIME.XML.env_case import EnvCase
 from CIME.XML.env_mach_pes import EnvMachPes
 from CIME.XML.env_batch import EnvBatch
+from CIME.XML.generic_xml import GenericXML
 from CIME.utils import run_cmd_no_fail
 
 logger = logging.getLogger(__name__)
@@ -23,7 +24,14 @@ def lock_file(filename, caseroot=None, newname=None):
     if not os.path.exists(fulllockdir):
         os.mkdir(fulllockdir)
     logging.debug("Locking file {} to {}".format(filename, newname))
+
+    # JGF: It is extremely dangerous to alter our database (xml files) without
+    # going through the standard API. The copy below invalidates all existing
+    # GenericXML instances that represent this file and all caching that may
+    # have involved this file. We should probably seek a safer way of locking
+    # files.
     shutil.copyfile(os.path.join(caseroot, filename), os.path.join(fulllockdir, newname))
+    GenericXML.invalidate_file(os.path.join(fulllockdir, newname))
 
 def unlock_file(filename, caseroot=None):
     expect("/" not in filename, "Please just provide basename of locked file")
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/check_lockedfiles.py
```diff
@@ -37,18 +45,6 @@ def is_locked(filename, caseroot=None):
...
     return os.path.exists(os.path.join(caseroot, LOCKED_DIR, filename))
 
-def restore(filename, caseroot=None, newname=None):
-    """
-    Restore the locked version of filename into main case dir
-    """
-    expect("/" not in filename, "Please just provide basename of locked file")
-    caseroot = os.getcwd() if caseroot is None else caseroot
-    newname = filename if newname is None else newname
-    shutil.copyfile(os.path.join(caseroot, LOCKED_DIR, filename), os.path.join(caseroot, newname))
```
jgfouca: This alarmed me even though this function was not being used., discussion below.

scripts/lib/CIME/check_lockedfiles.py
```diff
@@ -23,7 +24,14 @@ def lock_file(filename, caseroot=None, newname=None):
     if not os.path.exists(fulllockdir):
         os.mkdir(fulllockdir)
     logging.debug("Locking file {} to {}".format(filename, newname))
+
+    # JGF: It is extremely dangerous to alter our database (xml files) without
+    # going through the standard API. The copy below invalidates all existing
+    # GenericXML instances that represent this file and all caching that may
+    # have involved this file. We should probably seek a safer way of locking
+    # files.
```
jedwards4b: Should we have a flush at this point to make sure that the file is up to date before the copy?

jgfouca:
@jedwards4b I think we need to firm-up our concept of env*.xml as a database. I think we are getting closer, but I'm still finding code that undermines our own system.

The Case class, which contains a handful of GenericXML objects which themselves wrap individual XML files in a case, is an in-memory database of entries for a case. In order for the system to be consistent, the underlying XML files cannot be changed without going through the standard set-value API.

Things that undermine this system:
* File copies/deletions that impact the XML files. These changes will not be seen in the in-memory db. The locking system (check_lockedfiles.py) seems to be the primary culprit here.
* Fixed with this PR: Multiple GenericXML objects wrapping the same XML file, changes to one were not being seen in the others.
* Forking subprocess commands that alter XML files. We've addressed this by forcing re-reads when we suspect something could have changes. This is an extremely slow and error-prone approach though.

jedwards4b:
I'm afraid that we can't avoid that flush / refresh of objects - external processes may call xmlchange to update a field, this will update the file and so objects will need to reread - also some externals are still using the perl interface to the xml.

jgfouca:
@jedwards4b , I agree, just something to think about longterm. I'm mostly happy with the current situation.

PR #2250, opened 2018-01-25
Revert file caching (pr #2243)

### LINKED ISSUE #2249: ERR test not setting DOUT_S
It should be TRUE but its currently false on master, git blame doesn't make any sense to me this must have just happened because ERR tests will fail when DOUT_S is FALSE.

jedwards4b:
This reverts PR #2243 which was causing ERR test to fail and generally reducing confidence in the system tests.   I think that the ERR failure may be a race condition because it doesn't seem to happen every time. 
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #2249 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -31,15 +31,6 @@ class _Element(object): # private class, don't want users constructing directly 
         return _Element(deepcopy(self.xml_element))
 
 class GenericXML(object):
-
-    _FILEMAP = {}
-    DISABLE_CACHING = False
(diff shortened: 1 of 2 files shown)


jedwards4b:
We have several tests (all compare_two based tests) and tools (create_clone) which interact with two case objects at the same time.  It appears that some how with the file caching the case objects are getting mixed up.  I think that the safest thing to do at this time is revert the change.

jedwards4b:
okay - new PR coming.

PR #2254, opened 2018-01-25
Fix for LockedFiles (issue #2252)

### LINKED ISSUE #2252: Check locked files function is broken.
When a user who doesn't have access to the default account (or repo) on a machine, they can try (at least) these 2 things:
a) use the -p option to create_test
b) go into case directory and xmlchange CHARGE_ACCOUNT=new_repo

Both of these methods show some issues.

a) Example:
./create_test -p m1759 SMS.ne4_ne4.FC5AV1C-L

Then looking at CHARGE_ACCOUNT:
cjyang@cori12:/global/cscratch1/sd/cjyang/acme_scratch/cori-knl/SMS.ne4_ne4.FC5AV1C-L.cori-knl_intel.20180125_112550_jdwdsv> ./xmlquery CHARGE_ACCOUNT
 
Results in group case.run
	CHARGE_ACCOUNT: /project/projectdirs/nstaff/cjyang
 
	CHARGE_ACCOUNT: /project/projectdirs/nstaff/cjyang
 
Results in group case.test
	CHARGE_ACCOUNT: /project/projectdirs/nstaff/cjyang

However, user job seems to be OK. ?

b) Somewhat similar issue.  And for both, it looks like what is reported from ./preview_run is a little confusing.

jedwards4b:
Fix for the locked files issue, needed to add locking for env_batch.xml and write a class specific compare_xml function also added a test to scripts_regression_tests.py

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #2252 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: jgfoucar

### DIFF
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -461,11 +461,3 @@ class GenericXML(object):
         if xmlid is not None:
             return xmlid
         return self.name(self.root)
-
-    def elements_equal(self, other):
-        if e1.tag != e2.tag: return False
-        if e1.text != e2.text: return False
-        if e1.tail != e2.tail: return False
-        if e1.attrib != e2.attrib: return False
-        if len(e1) != len(e2): return False
-        return all(elements_equal(c1, c2) for c1, c2 in zip(e1, e2))
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -461,3 +461,11 @@ class GenericXML(object):
         if xmlid is not None:
             return xmlid
         return self.name(self.root)
+
+    def elements_equal(self, other):
+        if e1.tag != e2.tag: return False
+        if e1.text != e2.text: return False
+        if e1.tail != e2.tail: return False
+        if e1.attrib != e2.attrib: return False
+        if len(e1) != len(e2): return False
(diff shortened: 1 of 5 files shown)


scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -461,3 +461,11 @@ def get_id(self):
         if xmlid is not None:
             return xmlid
         return self.name(self.root)
+
+    def elements_equal(self, other):
+        if e1.tag != e2.tag: return False
```
jgfouca: There's probably an easier way to do this. Do XML Elements have a built-in equality operator?

jgfouca: Looks like _Element in this file already supports equality operations.

jedwards4b: so I only need to remove the new one that I introduced?

jgfouca: Yes, just use elem1 == elem2

jgfouca:
@jedwards4b , one last thing, can you add a test to catch the original problem?

jedwards4b:
Yes, working on it.

jedwards4b:
Testing complete, test added, all pass.

jgfouca:
Nice!

PR #2255, opened 2018-01-25
Change PET to be a multisubmit job

jgfouca:
Changing all the threading makes it hacky to try to do both runs
under a single submission.

Test suite: by hand, melvin and skybridve
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Update gh-pages html (Y/N)?:N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/case_run.py
+++ b/scripts/lib/CIME/case_run.py
@@ -231,7 +231,7 @@ def resubmit_check(case):
         run_cmd(cmd, verbose=True)
 
     if resubmit:
-        if testcase is not None and testcase in ['ERR']:
+        if testcase is not None and testcase in ['ERR', 'PET']:
             job = "case.test"
         else:
             job = "case.run"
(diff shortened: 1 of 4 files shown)


scripts/lib/CIME/case_run.py
```diff
@@ -231,7 +231,7 @@ def resubmit_check(case):
         run_cmd(cmd, verbose=True)
 
     if resubmit:
-        if testcase is not None and testcase in ['ERR']:
+        if testcase is not None and testcase in ['ERR', 'PET']:
```
jgfouca: @jedwards4b , this seems pretty hack to me. I had to use pdb to step through the entire submit/run phase of PET to figure out why resubmit wasn't running case.test. I'm wondering why we do this in the first place?

jedwards4b: It's probably enough to have if testcase is not None:

jgfouca: OK, let me do that.

jgfouca:
Yep, running now.

jgfouca:
@jedwards4b , all tests passed.

PR #2260, opened 2018-01-26
Improve python interface to environment modules

### LINKED ISSUE #2091: CIME error with python >= version 2.7.14
Using python 2.7.14, I get a value error in case.submit on anvil with E3SM, the traceback follows
```
  File "/var/spool/torque/mom_priv/jobs/1788221.bmgt1.lcrc.anl.gov.SC", line 87, in <module>
    _main_func(__doc__)
  File "/var/spool/torque/mom_priv/jobs/1788221.bmgt1.lcrc.anl.gov.SC", line 82, in _main_func
    success = case_run(case, skip_pnl=skip_pnl)
  File "/blues/gpfs/home/lvanroe/acme_master/cime/scripts/Tools/../../scripts/lib/CIME/case_run.py", line 290, in case_run
    lid = run_model(case, lid, skip_pnl)
  File "/blues/gpfs/home/lvanroe/acme_master/cime/scripts/Tools/../../scripts/lib/CIME/case_run.py", line 147, in run_model
    return run_and_log_case_status(functor, "case.run", caseroot=case.get_value("CASEROOT"))
  File "/blues/gpfs/home/lvanroe/acme_master/cime/scripts/Tools/../../scripts/lib/CIME/utils.py", line 1359, in run_and_log_case_status
    rv = func()
  File "/blues/gpfs/home/lvanroe/acme_master/cime/scripts/Tools/../../scripts/lib/CIME/case_run.py", line 146, in <lambda>
    functor = lambda: _run_model_impl(case, lid, skip_pnl=skip_pnl)
  File "/blues/gpfs/home/lvanroe/acme_master/cime/scripts/Tools/../../scripts/lib/CIME/case_run.py", line 83, in _run_model_impl
    pre_run_check(case, lid, skip_pnl=skip_pnl)
  File "/blues/gpfs/home/lvanroe/acme_master/cime/scripts/Tools/../../scripts/lib/CIME/case_run.py", line 40, in pre_run_check
    case.load_env()
  File "/blues/gpfs/home/lvanroe/acme_master/cime/scripts/Tools/../../scripts/lib/CIME/case.py", line 1238, in load_env
    env_module.load_env(self)
  File "/blues/gpfs/home/lvanroe/acme_master/cime/scripts/Tools/../../scripts/lib/CIME/XML/env_mach_specific.py", line 82, in load_env
    self.load_modules(modules_to_load)
  File "/blues/gpfs/home/lvanroe/acme_master/cime/scripts/Tools/../../scripts/lib/CIME/XML/env_mach_specific.py", line 108, in load_modules
    self._load_modules_generic(modules_to_load)
  File "/blues/gpfs/home/lvanroe/acme_master/cime/scripts/Tools/../../scripts/lib/CIME/XML/env_mach_specific.py", line 309, in _load_modules_generic
    os.environ[key] = newenv[key]
  File "/lcrc/group/acme/lvanroe/conda/lib/python2.7/os.py", line 473, in __setitem__
    putenv(key, item)
ValueError: illegal environment variable name
```
case.submit appears to be failing when loading the ```LS_COLORS=.........```.  Indeed the python 2.7.14 release docs (https://raw.githubusercontent.com/python/cpython/c707893f9cee870bba8364b3a06eb9cfa3b80e58/Misc/NEWS) show that equal signs are now prohibited in environmental variables with putenv, and this is the problem here.

This should be reproducible on any platform with python 2.7.14.

jgfouca:
1) Get rid of module_lmod concept. All versions of module are capable of
python mode, they just need to be configured correctly in config_machines.xml
2) Fix module config for summitdev
3) Make internal methods load_modules and load_env private
4) Optimize module interation by bundling N commands doing the same action
into a single command with multiple arguments. This will greatly reduce the
number of subprocess forks and improve the speed of loading env.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2091 

User interface changes?: Yes, module_lmod no longer a support module type

Code review: @jedwards4b @vanroekel

### DIFF
--- a/scripts/lib/CIME/XML/env_mach_specific.py
+++ b/scripts/lib/CIME/XML/env_mach_specific.py
@@ -74,11 +74,11 @@ class EnvMachSpecific(EnvBase):
         # in the environment_variables block
         modules_to_load = self._get_modules_for_case(case)
         if (modules_to_load is not None):
-            self.load_modules(modules_to_load)
+            self._load_modules(modules_to_load)
 
         envs_to_set = self._get_envs_for_case(case)
         if (envs_to_set is not None):
-            self.load_envs(envs_to_set)
+            self._load_envs(envs_to_set)
 
         self._get_resources_for_case(case)
 
@@ -93,12 +93,10 @@ class EnvMachSpecific(EnvBase):
                 limits = (int(val), limits[1])
                 resource.setrlimit(attr, limits)
 
-    def load_modules(self, modules_to_load):
+    def _load_modules(self, modules_to_load):
         module_system = self.get_module_system_type()
         if (module_system == "module"):
             self._load_module_modules(modules_to_load)
-        elif (module_system == "module_lmod"):
-            self._load_modules_generic(modules_to_load)
         elif (module_system == "soft"):
             self._load_modules_generic(modules_to_load)
         elif (module_system == "generic"):
@@ -120,7 +118,7 @@ class EnvMachSpecific(EnvBase):
         else:
             source_cmd = ""
 
-        if (module_system in ["module", "module_lmod"]):
+        if (module_system in ["module"]):
             return run_cmd_no_fail("{}module list".format(source_cmd), combine_output=True)
         elif (module_system == "soft"):
             # Does soft really not provide this capability?
@@ -161,7 +159,7 @@ class EnvMachSpecific(EnvBase):
         with open(filename, "w") as fd:
             fd.write("\n".join(lines))
 
-    def load_envs(self, envs_to_set):
+    def _load_envs(self, envs_to_set):
         for env_name, env_value in envs_to_set:
             os.environ[env_name] = "" if env_value is None else env_value
 
@@ -231,10 +229,25 @@ class EnvMachSpecific(EnvBase):
         # Note this is independent of module system type
         mod_cmd = self.get_module_system_cmd_path(shell)
         cmds = []
+        last_action = None
+        last_cmd    = None
+
         for action, argument in modules_to_load:
             if argument is None:
                 argument = ""
-            cmds.append("{} {} {}".format(mod_cmd, action, "" if argument is None else argument))
+
+            if action == last_action:
+                last_cmd = "{} {}".format(last_cmd, argument)
+            else:
+                if last_cmd is not None:
+                    cmds.append(last_cmd)
+
+                last_cmd = "{} {} {}".format(mod_cmd, action, "" if argument is None else argument)
+                last_action = action
+
+        if last_cmd:
+            cmds.append(last_cmd)
+
         return cmds
 
     def _load_module_modules(self, modules_to_load):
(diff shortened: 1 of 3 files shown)


scripts/lib/CIME/XML/env_mach_specific.py
```diff
@@ -298,6 +311,8 @@ def _load_modules_generic(self, modules_to_load):
         for key in newenv:
             if key in os.environ and key not in newenv:
                 del(os.environ[key])
+            elif key in os.environ and os.environ[key] == newenv[key]:
```
jgfouca: @vanroekel , I believe this should fix your issue. It won't bother trying to set LS_COLORS. Please confirm.

vanroekel: @jgfouca I still have the same error with this line included.  It looks like the regular expression match above splits out the value after the last equal in the string, this gives keys in newenv with '=' in them.  A solution that works for me is commented above.

vanroekel: git won't let me comment above.  so I'll do so here.  If I change line 293 above from 
```
m=re.match(r'^(\S+)=(\S+)\s*;*\s*$',line)
```
to
```
m=re.match(r'^(\S+?)=(\S+)\s*;*\s*$',line)
```
This will stop at the first equal instead of the final one and my problem is fixed.

jgfouca:
@jedwards4b , @vanroekel , I've made big updates to this PR

scripts/tests/scripts_regression_tests.py
```diff
@@ -1758,6 +1758,49 @@ def test_create_test_longname(self):
...
+                os.environ.update(orig_env)
+
+                env_mach.load_env(case, force_method="generic")
+                generic_env = dict(os.environ)
+
+                os.environ.clear()
+                os.environ.update(orig_env)
+
+                problems = ""
+                for mkey, mval in module_env.items():
```
jgfouca: @jedwards4b , I discovered this test is very sensitive to user environment, we may need to add additional exceptions to the environment comparison. I do think a test like this is pretty important as it immediately caught a major bug in generic env loading.

jgfouca:
@vanroekel , can you confirm the fix?

scripts/lib/CIME/XML/env_mach_specific.py
```diff
@@ -270,37 +284,31 @@ def _load_modules_generic(self, modules_to_load):
...
-                val = string.Template(valunresolved).safe_substitute(newenv)
-                expect(val is not None,
-                       'string value {} unable to be resolved'.format(valunresolved))
-                newenv[key] = val
+        # resolve variables
+        for key, val in newenv.items():
+            newenv[key] = string.Template(val).safe_substitute(newenv)
 
         # Set environment with new or updated values
         for key in newenv:
```
jgfouca: There was a bug in the old code, environment variables were never being deleted. Made no sense to loop over newenv and then check if key not in newenv.

jgfouca:
@vanroekel , great! @jedwards4b , this is ready to merge.

PR #2261, opened 2018-01-26
fix compare of env_batch

jedwards4b:
Need to be more careful about matching xml fields in env_batch.xml

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


jedwards4b:
This PR is expected to resolve the CDASH test errors on the current master.

PR #2264, opened 2018-01-30
Jgfouca/branch for acme split 2018 01 30

jgfouca:
Shared change list:
* bless_test_results fix using REALUSER
* more robust stderr presentation for run_cmd_no_fail

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


jgfouca:
@billsacks , I used --squash for this split. Hope that resolved the history issues.

scripts/lib/CIME/bless_test_results.py
```diff
@@ -43,33 +43,35 @@ def bless_namelists(test_name, test_dir, report_only, force, baseline_name, base
 def bless_history(test_name, test_dir, baseline_name, baseline_root, report_only, force):
 ###############################################################################
     with Case(test_dir) as case:
-        if baseline_name is None:
-            baseline_name = case.get_value("BASELINE_NAME_CMP")
-            if not baseline_name:
-                baseline_name = CIME.utils.get_current_branch(repo=CIME.utils.get_cime_root())
+        real_user = case.get_value("REALUSER")
```
jgfouca: Bug fix for dynamic USER evaluation breaking bless_test_results.

scripts/lib/CIME/case.py
```diff
@@ -908,6 +908,8 @@ def configure(self, compset_name, grid_name, machine_name=None,
             else:
                 self._check_testlists(compset_alias, grid_name, files)
 
+        self.set_value("REALUSER", os.environ["USER"])
```
jedwards4b: this is kinda ugly - who is real user?  The current user, or the one that setup the case?

scripts/lib/CIME/utils.py
```diff
@@ -374,6 +410,17 @@ def run_cmd_no_fail(cmd, input_str=None, from_dir=None, verbose=None,
         # If command produced no errput, put output in the exception since we
         # have nothing else to go on.
         errput = output if not errput else errput
+        if errput is None:
```
jgfouca: Fix some cases where stderr reporting was causing problems.

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -509,7 +517,7 @@
   <entry id="BATCH_SYSTEM">
      <type>char</type>
      <default_value>none</default_value>
-     <valid_values>lc_slurm,moab,pbs,lsf,slurm,cobalt,cobalt_theta,none</valid_values>
+     <valid_values>nersc_slurm,lc_slurm,moab,pbs,lsf,slurm,cobalt,cobalt_theta,none</valid_values>
```
jedwards4b: Why so many exceptions?  All on the acme side...

scripts/lib/CIME/case.py
jgfouca: Yeah, it isn't great. REALUSER is the one who created the case. This fixes the bless_test_results issue we were discussing yesterday without needing additional flags to that tool.

src/drivers/mct/cime_config/config_component.xml
jgfouca: @jedwards4b , that's a good question. I was not involved in the creation of any of these. @ndkeen , do you know why a nersc_slurm batch system was created? It appears to be an exact copy of slurm.

jgfouca: Correction, it looks like nersc_slurm uses -q to select queue and slurm uses -p.

jedwards4b: We do that on the cesm side and still maintain a single slurm definition.

jgfouca: You're right, we need to clean our stuff up.

jgfouca:
@billsacks , I put the couple inline comments I made into the PR description. There were very few changes beyond ACME config changes.

As far as the file history, I'm at a complete loss on how to address that other than entirely suspending ACME/ESMCI integration.

billsacks:
@jgfouca ah right, now I remember: The plan for fixing the problem didn't have to do with how you created this branch, but has to do with how this branch is merged to master. As long as you merge this to master using `--squash` (on the command-line, or presumably you'd get the same thing with the "squash and merge" option on the "merge pull request" button), then history won't get messed up in this way. The downside is that this will look like a single commit with you as the author.

Will that plan work for you?

I'll go ahead and approve these changes, under the assumption that that's how you'll bring these changes to master.

jgfouca:
@billsacks , yes

jgfouca:
@billsacks , OK, I did a squash-and-merge. See what you think.

billsacks:
I spot-checked `scripts/lib/CIME/hist_utils.py` and at least that one looks good. So I think this squash-and-merge worked. Thank you!

Issue #2266, opened 2018-01-31
Components need post-DA resume signal


gold2718:
In order for components to do post-data assimilation state processing, they need a resume signal from the driver during their initialization phase. This 'resume' signal should be set for component `XXX` if the driver XML variable, `DATA_ASSIMILATION_XXX` is `TRUE`.

ekluzek:
Is this done now? This is required for DART to work in CESM right? And it's desired for CESM2.0 right?

gold2718:
That is weird. This issue was fixed by PR #2267 which contains the line:

> Fixes CIME Github issue #2266

Closing it manually now

PR #2267, opened 2018-01-31
Separate post-DA from pause/resume and add resume signal at init

gold2718:
- MCT driver now uses `DATA_ASSIMILATION_XXX` XML variables to send resume
signal to components at initialization time. Components will use this signal to do
post-data-assimilation processing after reading restart files. 
  - DAE test now passes resume signal after DA via DATA_ASSIMILATION_WAV.
  - DWAV model was modified to look for and log resume signal during initialization.
Fixed some issues in DWAV model. If DAE test is used with the DWAV model then the DAE test will make sure that the signal was seen.
- Added `PAUSE_ACTIVE_XXX` XML variables to the MCT driver config_component.xml
to support pause/resume functionality. 
  - Changed PRE test to use these variables.

Test suite: scripts_regression_tests.py on Cheyenne
DAE_D_Ld4.ww3a.ADWAV.hobart_nag for PASS and also to hand check that
resume signal was found, logged, and checked by both the data assimilation
script (da_no_data_mod.sh) and by the DAE test.
DAE_N2_D_Ld4.ww3a.ADWAV.hobart_nag and
DAE_C2_D_Ld4.ww3a.ADWAV.hobart_nag (as above)
Test baseline: NA
Test namelist changes: NA
Test status: bit for bit

Fixes CIME Github issue #2266

User interface changes?: New XML variables, `PAUSE_ACTIVE_XXX`.

Code review:

### DIFF
--- a/scripts/data_assimilation/da_no_data_mod.sh
+++ b/scripts/data_assimilation/da_no_data_mod.sh
@@ -17,6 +17,53 @@ else
   cycle=$2
   echo "caseroot: ${caseroot}"
   echo "cycle: ${cycle}"
+  cd ${caseroot}
+  res=$?
+  if [ $res -ne 0 ]; then
+    echo "ERROR: Unable to cd to caseroot, ${caseroot}"
+    errcode=$(( errcode + 1 ))
+  else
+    dval="`./xmlquery --value COMPSET 2> /dev/null | grep DWAV`"
+    ./xmlchange DATA_ASSIMILATION_WAV=TRUE
+    res=$?
+    if [ $res -ne 0 ]; then
+      echo "ERROR: Unable to change DATA_ASSIMILATION_WAV to TRUE"
+      errcode=$(( errcode + 1 ))
+    fi
+    if [ $cycle -gt 0 -a -n "${dval}" ]; then
+      rundir="`./xmlquery --value RUNDIR`"
+      if [ -n "${rundir}" -a -d "${rundir}" ]; then
+        cd ${rundir}
+        res=$?
+        if [ $res -ne 0 ]; then
+          echo "ERROR: Unable to cd to caseroot, ${caseroot}"
+          errcode=$(( errcode + 1 ))
+        else
+          # Check the latest log file for a resume signal
+          lfile="`ls -t wav.log.* | head -1`"
+          if [ -n "${lfile}" -a -f "${lfile}" ]; then
+            if [ -n "`echo ${lfile} | grep 'gz$'`" ]; then
+              sig="`zcat ${lfile} | grep 'Resume signal, TRUE' 2> /dev/null`"
+            else
+              sig="`cat ${lfile} | grep 'Resume signal, TRUE' 2> /dev/null`"
+            fi
+            if [ -n "${sig}" ]; then
+              echo "Post-DA resume signal found for cycle ${cycle}"
+            else
+              echo "ERROR: No post-DA resume signal found for cycle ${cycle}"
+              errcode=$(( errcode + 1 ))
+            fi
+          else
+            echo "ERROR: Unable to find wav log file in `pwd -P`"
+            errcode=$(( errcode + 1 ))
+          fi
+        fi
+      else
+        echo "ERROR: RUNDIR (${rundir}) is not a valid directory"
+        errcode=$(( errcode + 1 ))
+      fi
+    fi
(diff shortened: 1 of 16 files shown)


scripts/lib/CIME/SystemTests/dae.py
```diff
@@ -91,9 +90,12 @@ def run_phase(self): # pylint: disable=arguments-differ
                "ERROR: There were {:d} DA cycles in run but {:d} DA files were found".format(da_cycles, len(da_files) if da_files is not None else 0))
         da_files.sort()
         cycle_num = 0
+        compset = self._case.get_value("COMPSET")
+        is_dwav = '_DWAV' in compset
         for fname in da_files:
             found_caseroot = False
             found_cycle = False
+            found_signal = not is_dwav
```
gold2718: Can only do the checks if DWAV is in the compset. Currently, scripts_regression_tests uses an A compset for DAE testing.

scripts/data_assimilation/da_no_data_mod.sh
```diff
@@ -17,6 +17,53 @@ else
   cycle=$2
```
jedwards4b: why do this in sh?   Wouldn't python be a better example?

src/drivers/mct/cime_config/testdefs/testlist_drv.xml
```diff
@@ -614,12 +614,20 @@
...
+      <option name="wallclock"> 00:20:00 </option>
+    </options>
+  </test>
   <test name="PRE" grid="f09_f09" compset="ADESP">
     <machines>
       <machine name="hobart" compiler="nag" category="pauseresume"/>
     </machines>
     <options>
-      <option name="wallclock"> 00:20 </option>
+      <option name="wallclock"> 00:20:00 </option>
```
gold2718: Is this right?

jedwards4b: yes

gold2718: At one point, I thought we were standardizing times so they were platform independent. Is that not the case? I do not think I had an issue with this (the issue I was having turned out to be a different problem).
I'm happy to back this out if it is not an improvement.

jedwards4b: We are - yellowstone was the exception, all others are hh:mm:ss

scripts/data_assimilation/da_no_data_mod.sh
gold2718: I guess it depends on what the DA community is doing. The idea is to test the DA cycle in a way that is similar to active DA experiments. Currently, all the DA scripts I have seen are shell scripts.

gold2718: > I'm not really sure of what distinction you are trying to make here and why the original implementation was inadequate.

Are you talking about da_no_data_mod.sh (the question is in a weird place).
The idea here is to check to see if DWAV noted and logged the resume signal.

jedwards4b: This question was in reference to the entire PR.  However since I am not the primary reviewer I can let this go.

gold2718: Ah. The reason for this PR is that we need separate mechanisms to support CESM2 data assimilation and post-CESM2 pause/resume functionality (including data assimilation).
To support current DA usage, we need to send a resume signal to the appropriate components at initialization and only at initialization. This PR does that using the `DATA_ASSIMILATION_XXX` variables. Prior to this PR, there was no resume signal going to the components at initialization time.

Pause/resume also needs to use the resume signal but at run time (and not at initialization time). I need the `PAUSE_ACTIVE_XXX` variables to determine which components will participate in pause (i.e., will receive pause signals). In pause/resume, the resume signal is not controlled by the configuration but by the active ESP component.

Pause/resume is not tied to data assimilation so I felt I needed separate variables.

kdraeder: I confirm that all of the existing DART scripts use [[t]]c]sh.
I'm don't know enough about the CESM context to review
the PAUSE_ACTIVE_XXX changes, but it sounds like it's
still incorporating the functionality that DA needs, so I'll just
approve it.

PR #2269, opened 2018-02-01
Reinstate xml file caching, but only for read-only files

jgfouca:
Introduces the concept of read-only files. Files being read out of
the repo should never be modified, only EnvBase subclasses are
eligible for modification. This means we can cache the read-only
files without having to worry about cache-coherence issues.

Change list:
1) EnvArchive needs to be a subclass of EnvBase since it's an env case file
2) Caching only for read only files, ensure read-only prevents modifications
3) Try to minimize the differences between env_entryid and env_generic files in Case

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -350,7 +350,7 @@ class SystemTestsCommon(object):
         Compare env_run file to original and warn about differences
         """
         components = self._case.get_values("COMP_CLASSES")
-        f1obj = EnvRun(self._caseroot, "env_run.xml", components=components)
+        f1obj = self._case.get_env("run")
         f2obj = EnvRun(self._caseroot, os.path.join(LOCKED_DIR, "env_run.orig.xml"), components=components)
         diffs = f1obj.compare_xml(f2obj)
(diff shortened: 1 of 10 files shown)


scripts/lib/CIME/SystemTests/system_tests_common.py
```diff
@@ -350,7 +350,7 @@ def compare_env_run(self, expected=None):
         Compare env_run file to original and warn about differences
         """
         components = self._case.get_values("COMP_CLASSES")
-        f1obj = EnvRun(self._caseroot, "env_run.xml", components=components)
+        f1obj = self._case.get_env("run")
```
jgfouca: Dangerous to create extra Env objects. That can gets us intro trouble with coherence issues.

jgfouca:
I gave up on caching the modifiable (case Env) files. There were too many potential coherence problems.

jedwards4b:
@jgfouca have you had a chance to test this on a batch machine?

jgfouca:
@jedwards4b , yes, it worked.

PR #2272, opened 2018-02-01
allow merge of xml files without extra child level

### LINKED ISSUE #2221: userdefined_laptop_template files need to be changed to work
It appears that the required format for `config_machines.xml` and maybe `config_compilers.xml` has been changed; this should be updated in the templates in `userdefined_laptop_template`. These should no longer have the outer `<config_machines>` (and maybe `<config_compilers>` tags).

See the last couple of comments in #2215 for discussion.

jedwards4b:
Merging of xml files such as those in .cime directory was creating an extra parent level, this 
removes that layer if it exists. 

Test suite: scripts_regression_tests.py, SSP_Ld10.f19_g17.I1850Clm50Bgc.cheyenne_gnu.clm-rtmColdSSP 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2221 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -164,7 +164,12 @@ class GenericXML(object):
         """
         expect(not self.locked, "locked")
         root = root if root is not None else self.root
-        root.xml_element.append(node.xml_element)
+        logger.debug("add_child node name is {} root name is {}".format(node.xml_element.tag, self.name(root)))
+        if node.xml_element.tag == self.name(root):
+            for child in node.xml_element:
+                root.xml_element.append(child)
+        else:
+            root.xml_element.append(node.xml_element)
 
     def copy(self, node):
         return deepcopy(node)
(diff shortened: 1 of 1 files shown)


jedwards4b:
Yea, I could move it there.

jgfouca:
Maybe add a merge option to add_child and have the invocation from read_fd set to true.

billsacks:
@jedwards4b and @jgfouca With this change, will the outer-level tags be required or optional? (e.g., the outer-level `<config_machines>` tags in a config_machines.xml file in your .cime directory). If optional, will there be an official recommendation as to whether they should be included? I ask because I've started to change some files to remove these outer-level xml tags, and I'm wondering if I should change them back.

jgfouca:
@billsacks , they are required now I think.

jgfouca:
In fact, read_fd should probably error-out if top-level root tag doesn't match it's root tag.

jedwards4b:
No it's optional.   I did it this way because I've been asking folks to reformat but I never liked that, you can't for example check those files with xmllint.    But I didn't want to tell people that they have to undo what they just did.

jedwards4b:
Officially they should be included.

billsacks:
Okay, thanks @jedwards4b . Sounds good.

jedwards4b:
@jgfouca all tests pass.

PR #2277, opened 2018-02-05
st_archive counting too many cpl.r files when MULTI_… 

### LINKED ISSUE #2275: case_st_archive.py:_get_datenames needs to use MULTI_DRIVER
@gold2718 @jedwards4b
This issue stems from #2240.
CESM+DART cases will generally be built with the MULTI_DRIVER option,
since it saves a lot of core hours and wall clock.
_get_datenames creates a list of all of the cpl restart files in order to know
how many restart file sets need to be archived.  But  in the M_D cases
there are NINST cpl restart files for each date stamp: cpl_0001.r, ..., 
so the archiver tries to archive NINST*#_dates file sets.
The proposed fix is to pass "case" down to _get_datenames,
query MULTI_DRIVER, and assemble the appropriate list
of restart sets.  I've coded and tested this, and will issue a PR soon,
unless I shouldn't.

kdraeder:
When the MULTI_COUPLER option is used to build a case,
_get_datenames was not accounting for there being NINST cpl*.r
files at each restart date (cpl_0001.r., cpl_0002.r.,...).  
So st_archive was repeatedly trying to archive the same set.
_get_datenames now gets 'case' as an argument, 
which it queries for MULTI_COUPLER, and then counts
the correct subset of cpl.r files.

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes #2275

User interface changes?: no

Update gh-pages html (Y/N)?:N

Code review:

### DIFF
--- a/scripts/lib/CIME/case_st_archive.py
+++ b/scripts/lib/CIME/case_st_archive.py
@@ -30,16 +30,10 @@ def _get_datenames(rundir, casename, case):
     """
     expect(isdir(rundir), 'Cannot open directory {} '.format(rundir))
 
-    # KDR
     # The MULTI_DRIVER option requires a more careful search for dates.
     # When True, each date has MULTI_DRIVER cpl_####.r files.
     ndriver = case.get_value('MULTI_DRIVER')
     logger.debug("_get_datenames:  ndriver = {} ".format(ndriver))
-   
-    # glob uses shell style wildcards:
-    #   [!chars], instead of [^chars]).
-    #   ? matches a single char, instead of .
-    #   * matches everything
     if ndriver :
        files = sorted(glob.glob(os.path.join(rundir, casename + '.cpl_0001.r.*.nc')))
     else:
--- a/scripts/lib/CIME/case_st_archive.py
+++ b/scripts/lib/CIME/case_st_archive.py
@@ -21,7 +21,7 @@ def _get_archive_file_fn(copy_only):
     return shutil.copyfile if copy_only else shutil.move
 
 ###############################################################################
-def _get_datenames(rundir, casename):
+def _get_datenames(rundir, casename, case):
 ###############################################################################
     """
     Returns the datetime objects specifying the times of each file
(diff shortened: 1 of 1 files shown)


scripts/lib/CIME/case_st_archive.py
```diff
@@ -21,15 +21,25 @@ def _get_archive_file_fn(copy_only):
     return shutil.copyfile if copy_only else shutil.move
 
 ###############################################################################
-def _get_datenames(rundir, casename):
+def _get_datenames(rundir, casename, case):
```
gold2718: If you are passing in the case, why also pass in rundir and casename which have just been extracted from the case by the caller? I suggest just passing in case and move the case.get_value calls into `_get_datenames`.

scripts/lib/CIME/case_st_archive.py
```diff
@@ -545,7 +555,7 @@ def _archive_process(case, archive, last_date, archive_incomplete_logs, copy_onl
                        archive_incomplete_logs, archive_file_fn)
 
     # archive restarts and all necessary associated files (e.g. rpointer files)
-    datenames = _get_datenames(case.get_value("RUNDIR"), case.get_value("CASE"))
+    datenames = _get_datenames(case.get_value("RUNDIR"), case.get_value("CASE"), case)
```
jedwards4b: If you are going to pass case as an argument then just pass case and get "RUNDIR" and "CASE" inside, 
or since you are only really adding multi_driver, just change the argument list to 
case.get_value("RUNDIR"), case.get_value("CASE"), case.get_value("MULIT_DRIVER")

gold2718:
I am not competent enough on the ST archive infrastructure to be the assignee. Also, please add appropriate labels to this PR.

kdraeder:
Good points from both of you.  I had tunnel vision about making the
minimal changes to make it work.  I'll rewrite it more efficiently.

kdraeder:
I got the impression that Alice is too busy, and this seemed small and
straightforward.
Can you suggest someone else?
Thanks,
Kevin

scripts/lib/CIME/case_st_archive.py
```diff
@@ -21,15 +21,25 @@ def _get_archive_file_fn(copy_only):
...
     """
     expect(isdir(rundir), 'Cannot open directory {} '.format(rundir))
-    files = sorted(glob.glob(os.path.join(rundir, casename + '.cpl*.r*.nc')))
+
+    # The MULTI_DRIVER option requires a more careful search for dates.
+    # When True, each date has MULTI_DRIVER cpl_####.r files.
+    ndriver = case.get_value('MULTI_DRIVER')
+    logger.debug("_get_datenames:  ndriver = {} ".format(ndriver))
+    if ndriver :
+       files = sorted(glob.glob(os.path.join(rundir, casename + '.cpl_0001.r.*.nc')))
```
jedwards4b: Why are you only looking for one cpl file here?

jedwards4b: That is why isn't it cpl_\d+.r.*.nc

jedwards4b:
Also see https://github.com/ESMCI/cime/wiki/Testing

kdraeder:
I'm looking for 1 cpl.r file from each date.
I chose 0001 because if there are multiple cpl_####.r. files for a date,
then 0001 will be one of them.

cpl_\d+.r.*.nc  would find too many files: NINST (from _\d+)  x num_dates.
We just want num_dates here.

Also, I'm not sure the \d+ would work in the glob context;
it uses shell style wild-cards, not python.  I'd have to test it to know.

Kevin

jedwards4b:
We have a couple of problems on master that will prevent proper testing, I should have a fix this afternoon - once that is in you will need to rebase and test.

kdraeder:
I've opted to pass just 'case' to _get_datenames 
and extract data from it in _get_datenames.
I've written and tested the the code in st_archive,
but am waiting until the testing software is fixed
( @jedwards4b  2018-2-5 3pm) before rebasing, 
running the cime testing sofware,
and pushing the new version.

jedwards4b:
Okay, that's PR #2278 should be this afternoon.

jedwards4b:
@kdraeder you should be able to rebase and test now.

kdraeder:
This has been superceded by #2285.

PR #2278, opened 2018-02-05
Fix issues with read only xml 

### LINKED ISSUE #1470: Additional functionality desired for query_config
#1388 is improving query_config, partly to address #1190 . However, many aspects of https://github.com/ESMCI/cime/issues/1190#issuecomment-282928761 are not addressed in that PR.

I see the most important outstanding requests as:

1. For each compset, a listing of which grids are scientifically and test-supported for that compset

1. Allow subsetting compsets by the first letter of the compset, rather than cam/clm/etc., along with some help text saying "B means XXX, F means YYY", etc. This is important because it's not necessarily intuitive that you get I compsets by specifying `--compsets clm`.

### LINKED ISSUE #2276: query_config no longer works for CESM
query_config no longer works for either cime5.4.0-alpha.23 or on master.
Here is the error for cime5.4.0-alpha.23:

[aliceb@cheyenne1:scripts]>./query_config --components clm
  File "./query_config", line 343, in <module>
    _main_func(__doc__)
  File "./query_config", line 336, in _main_func
    query_component(args.components)
  File "./query_config", line 146, in query_component
    if not all_components and not config_exists:
UnboundLocalError: local variable 'config_exists' referenced before assignment
and here is the error from master:

[aliceb@cheyenne6:scripts]>./query_config --component clm
ERROR: locked

jedwards4b:
Fix issues associated with read-only xml objects, add a test to append config_machines.xml

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #2276  + CDASH

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/files.py
+++ b/scripts/lib/CIME/XML/files.py
@@ -25,12 +25,19 @@ class Files(EntryID):
         schema = os.path.join(cimeroot, "config", "xml_schemas", "entry_id.xsd")
         EntryID.__init__(self, infile, schema=schema)
         config_files_override = os.path.join(os.path.dirname(cimeroot),".config_files.xml")
+        # variables COMP_ROOT_DIR_{} are mutable, all other variables are read only
+        self.COMP_ROOTS = {}
+
         # .config_file.xml at the top level may overwrite COMP_ROOT_DIR_ nodes in config_files
+
         if os.path.isfile(config_files_override):
             self.read(config_files_override)
             self.overwrite_existing_entries()
 
     def get_value(self, vid, attribute=None, resolved=True, subgroup=None):
+        if "COMP_ROOT_DIR" in vid:
+            if vid in self.COMP_ROOTS:
+                return self.COMP_ROOTS[vid]
         value = super(Files, self).get_value(vid, attribute=attribute, resolved=False, subgroup=subgroup)
         if "COMP_ROOT_DIR" not in vid and value is not None and resolved and "COMP_ROOT_DIR" in value:
             m = re.search("(COMP_ROOT_DIR_[^/]+)/", value)
@@ -40,6 +47,13 @@ class Files(EntryID):
         if resolved and value is not None:
             value = self.get_resolved_value(value)
 
+        return value
+
+    def set_value(self, vid, value):
+        if "COMP_ROOT_DIR" in vid:
+            self.COMP_ROOTS[vid] = value
+        else:
+            expect(False, "Attempt to set a nonmutable variable {}".format(vid))
         return value
 
     def get_schema(self, nodename, attributes=None):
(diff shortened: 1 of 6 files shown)


scripts/lib/CIME/XML/files.py
```diff
@@ -32,11 +32,15 @@ def __init__(self):
 
     def get_value(self, vid, attribute=None, resolved=True, subgroup=None):
         value = super(Files, self).get_value(vid, attribute=attribute, resolved=False, subgroup=subgroup)
-        if "COMP_ROOT_DIR" not in vid and value is not None and resolved and "COMP_ROOT_DIR" in value:
-            m = re.search("(COMP_ROOT_DIR_[^/]+)/", value)
-            comp_root_dir_var_name = m.group(1)
-            comp_root_dir = self.get_value(comp_root_dir_var_name, attribute=attribute, resolved=False, subgroup=subgroup)
-            self.set_value(comp_root_dir_var_name, comp_root_dir)
+        if value is None and "COMP_ROOT_DIR" not in vid and attribute is not None:
+            value = super(Files, self).get_value(vid, attribute=None, resolved=False, subgroup=subgroup)
```
jgfouca: This is a repeat of an earlier line.

jgfouca:
Looks great other than the one comment.

rljacob:
Actual bug issue number is #2276.  I changed the PR description.

scripts/lib/CIME/XML/files.py
jedwards4b: No it's not, note the change in attribute.

jgfouca: Ah, didn't notice that.

jedwards4b:
I've redone this with a simpler, cleaner approach.  Have another look.

jedwards4b:
wait, it's not working yet..

scripts/lib/CIME/XML/machines.py
```diff
@@ -293,6 +293,7 @@ def get_suffix(self, suffix_type):
         return None
 
     def set_value(self, vid, value, subgroup=None, ignore_type=True):
```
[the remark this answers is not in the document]
jgfouca: ok

scripts/lib/CIME/XML/files.py
```diff
@@ -25,12 +25,19 @@ def __init__(self):
...
+        self.COMP_ROOTS = {}
+
         # .config_file.xml at the top level may overwrite COMP_ROOT_DIR_ nodes in config_files
+
         if os.path.isfile(config_files_override):
             self.read(config_files_override)
             self.overwrite_existing_entries()
 
     def get_value(self, vid, attribute=None, resolved=True, subgroup=None):
+        if "COMP_ROOT_DIR" in vid:
```
jgfouca: I like this approach.

jedwards4b:
@jgfouca Okay all tests are passing now.

kdraeder:
@jedwards4b 
So it looks like I should go ahead with regression testing in #2277,
after rebasing.

PR #2283, opened 2018-02-07
move orb year settings for 1850 out of user mods

### LINKED ISSUE #2282: CPL_USER_MODS match for b1850 is too general
c8c758224a introduced this in `config_component_cesm.xml`:

```xml
  <entry id="CPL_USER_MODS">
    <type>char</type>
    <valid_values></valid_values>
    <values match="last">
      <value compset="1850_">$SRCROOT/cime_config/usermods_dirs/b1850</value>
    </values>
    <group>run_component_cpl</group>
    <file>env_case.xml</file>
    <desc>User mods to apply to specific compset matches. </desc>
  </entry>
```

But matching on compset="1850_" is too general: This matches non-B compsets (e.g., I compsets). For a standalone ctsm checkout, this raises an error because `$SRCROOT/cime_config` doesn't exist. For other configurations, I'm not sure this is working as intended - e.g., is this supposed to be included in an I1850 compset?

@jedwards4b I'm not sure exactly what's intended here.

jedwards4b:
Setting the orb_year and orb_year_align for 1850 should not be done via user mods for 1850 cases. 
Here it is set in the namelist_definition_drv.xml file.   These will be set for any 1850 compset - is there any reason this wouldn't be the case for an I1850 or F1850?

Test suite: tests by hand of 1850 compsets
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2282

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/drivers/mct/cime_config/buildnml
+++ b/src/drivers/mct/cime_config/buildnml
@@ -32,6 +32,10 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
     #--------------------------------
     config = {}
     config['cime_model'] = get_model()
+
+    compset = case.get_value("COMPSET")
+    config['iyear'] = compset.split('_')[0]
+
     config['BGC_MODE'] = case.get_value("CCSM_BGC")
     config['CPL_I2O_PER_CAT'] = case.get_value('CPL_I2O_PER_CAT')
     config['COMP_RUN_BARRIERS'] = case.get_value('COMP_RUN_BARRIERS')
--- a/src/drivers/mct/cime_config/config_component_cesm.xml
+++ b/src/drivers/mct/cime_config/config_component_cesm.xml
@@ -152,7 +152,7 @@
     <valid_values></valid_values>
     <default_value></default_value>
     <values match="last">
-      <value compset="1850_">$SRCROOT/cime_config/usermods_dirs/b1850</value>
+      <value/>
     </values>
     <group>run_component_cpl</group>
     <file>env_case.xml</file>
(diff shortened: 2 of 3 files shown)


billsacks:
Thanks for this @jedwards4b . In addition to B compsets, this will change the orbital settings for some C, E, F, G, J, I and T compsets. If we go with the change as written, I feel that the relevant co-chairs / liaisons should be contacted before making this change, to make them aware of this behavior change. I can sign off on the J and T compsets myself, but I don't know about the others.

jedwards4b:
I have no problem with that - but it seems to me that this is a fundamental part of the definition of an 1850 compset.

billsacks:
> it seems to me that this is a fundamental part of the definition of an 1850 compset.

I agree with your intuition here, but I think it's worth double-checking in case the current behavior is intentional. So thanks for sending an email out.

I realized this will also affect behavior in ACME. So this probably needs a sign-off from @rljacob or others.

billsacks:
@jedwards4b I see you merged this without there being reviews from @mvertens or @rljacob . Because there are behavior changes here, I think it's important that this receive sign-off from others. Did you receive verbal confirmation from them that this is okay?

jedwards4b:
It seems that I double clicked when I intended to merge #2288, I did not intend to merge this.   I have received verbal okay from @mvertens and would like to leave this in for testing with alpha10a unless there is an objection in which case I will revert.

rljacob:
This would change answers for E3SM where, as far as I know, we are using 1990 orb params for all our 1850 runs.  So this should not be merged until after the release.

jedwards4b:
Noted - I will submit a new PR this afternoon that will change this so that e3sm is not affected.

rljacob:
That would also work.  Thanks!

PR #2287, opened 2018-02-08
Disable caching withing the main test_scheduler process.

jgfouca:
Subprocesses (shell calls) will still do caching.

For reasons I can't understand, caching breaks things when running
multiple tests in test_scheduler, even when I protect the cache
with a mutex.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/test_scheduler.py
+++ b/scripts/lib/CIME/test_scheduler.py
@@ -18,6 +18,7 @@ from update_acme_tests import get_recommended_test_time
 from CIME.utils import append_status, append_testlog, TESTS_FAILED_ERR_CODE, parse_test_name, get_full_test_name, get_model
 from CIME.test_status import *
 from CIME.XML.machines import Machines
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/test_scheduler.py
```diff
@@ -880,7 +881,9 @@ def run_tests(self, wait=False,
         # Setup cs files
         self._setup_cs_files()
 
+        GenericXML.DISABLE_CACHING = True
```
jedwards4b: Shouldn't this be done as a function call for good form?

jgfouca: Maybe, it should only be temporary though.

jedwards4b: okay

PR #2288, opened 2018-02-09
Test compset vs alias

### LINKED ISSUE #2258: Add a test for user_compset vs compset alias
Both methods should set up the same case, we need a test that assures this is happening.

jedwards4b:
Add a test that compares a case created using the compset alias to 
one created using the longname for the same compset.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #2258

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


jedwards4b:
Oops I created this from another PR instead of master.  Hang on while I clean it up.

jedwards4b:
Okay - good to go now.

PR #2290, opened 2018-02-12
redo orb_year_defaults for cesm only

jedwards4b:
For cesm only set orb_iyear to 1850 for 1850 compsets

Test suite: by hand B1850 for both cesm and acme models 
Test baseline: 
Test namelist changes: 
Test status: expected to change answers for non-B 1850 compsets in CESM
Fixes

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/drivers/mct/cime_config/buildnml
+++ b/src/drivers/mct/cime_config/buildnml
@@ -32,6 +32,7 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
     #--------------------------------
     config = {}
     config['cime_model'] = get_model()
+    config['iyear'] = case.get_value('COMPSET').split('_')[0]
     config['BGC_MODE'] = case.get_value("CCSM_BGC")
     config['CPL_I2O_PER_CAT'] = case.get_value('CPL_I2O_PER_CAT')
     config['COMP_RUN_BARRIERS'] = case.get_value('COMP_RUN_BARRIERS')
--- a/src/drivers/mct/cime_config/config_component_cesm.xml
+++ b/src/drivers/mct/cime_config/config_component_cesm.xml
@@ -152,7 +152,7 @@
     <valid_values></valid_values>
     <default_value></default_value>
     <values match="last">
-      <value compset="1850_">$SRCROOT/cime_config/usermods_dirs/b1850</value>
+      <value/>
     </values>
     <group>run_component_cpl</group>
     <file>env_case.xml</file>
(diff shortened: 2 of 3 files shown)


billsacks:
Note for future reference: this is a redone version of #2283

billsacks:
@jedwards4b - this is missing the changes to `src/drivers/mct/cime_config/buildnml` that were in #2283 . Are these not needed?

jedwards4b:
It's there at line 34, just in a slightly different form.

billsacks:
Okay, thanks.

One last question: Have you heard back from anyone in the atmosphere group (for F compsets), ocean group (for C & G compsets) and sea ice group (for E compsets, I think), to confirm that they're okay with this change? Dave Lawrence signed off on it for I compsets, and I sign off on it for J and T compsets, but it would be good to get definitive confirmation that other groups are okay with this change being made.

rljacob:
Was this not branched off of master after you merged 2283?   The only code you should need is the cime_model qualifiers.

rljacob:
Why not **always** make the iorb params consistent with iyear in the compset?

jedwards4b:
My mistake was in the creation of #2288 which I started from this branch instead of master as I should have, to correct the problem I reverted the changes on that branch, but then when I merged that branch it also marked the #2283 branch as merged and the original PR was lost.   So I've had to do the whole thing over again.

jedwards4b:
@sacks I feel that I've given component liaisons plenty of opportunity to review and provide feedback.

dabail10:
I am fine with this change.

billsacks:
Cheryl Craig, Cecile Hannay and Julio Bacmeister have approved this for F compsets. I'm just waiting to hear back from the ocean group to confirm they're okay with this change for C & G compsets.

billsacks:
@jedwards4b I was just looking at the PR description. You said bit for bit, but actually this is expected to change answers for non-B 1850 compsets in CESM, right?

jedwards4b:
yes, updated

billsacks:
@mnlevy1981 says:

> We discussed this at the Oceanography section meeting today, and nobody has issues with it.

billsacks:
This has now been approved scientifically, we just have to determine how to stage it for testing purposes. I'll talk to @fischer-ncar about that.

jedwards4b:
@billsacks What is the hold up here?

billsacks:
Sorry! I talked to @fischer-ncar , got approval from him to merge it, then must have gotten distracted on the way back to my office and forgot to hit merge. Merging now.

Issue #2207, opened 2018-01-08
Add a gitignore to files added to a cime/scripts/cases subdirectory so cases created under there can be automatically ignored


ekluzek:
Add a gitignore for a "cases" subdirectory under cime/scripts, so that if you create cases there they will be ignored. We talked about this at the CSEG meeting, and decided this change was acceptable. Making this the default or creating an empty subdirectory was decided to not be acceptable.

Issue #1749, opened 2017-07-17
Remove the generic GSW DATM_MODE in datm


ekluzek:
@rljacob @bishtgautam we'd like to remove the older generic "GSW" option for DATM_MODE in config_component.xml for DATM, and only allow the more explicit GSWP3v7 option. We don't believe this affects ACME, but wanted to check. If you want to use the GSWP3 data you should use the GSWP3v7 version.

bishtgautam:
Thanks for checking in, but removing this will not impact ACME as no one so far has been running a simulation with GSW.

rljacob:
Ok with me.

PR #2300, opened 2018-02-15
Catch more errors in the testlist schema, remove generic GSW DATM_MODE ignore scripts/cases

### LINKED ISSUE #2245: testlist XSD should disallow options under the "<machines>" header and only allow under "<test>"  or "<machine>"
I screwed up the CLM testlist because I thought options could go under the "<machines>" header. Based on how they are currently read you can either put a general set of options under a "<test>" or you can put it under the specific "<machine>". But, it will ignore options put under the "<machines>" header at the same level as specific machine options.

This simple fix to the XSD will check for that and not allow you to do it wrong.

```
Index: ../../../../cime/config/xml_schemas/testlist.xsd
--- ../../../../cime/config/xml_schemas/testlist.xsd	(revision 11033)
+++ ../../../../cime/config/xml_schemas/testlist.xsd	(working copy)
@@ -27,7 +27,6 @@
     <xs:complexType>
       <xs:sequence>
         <xs:element maxOccurs="unbounded" ref="machine"/>
-        <xs:element minOccurs="0" ref="options"/>
       </xs:sequence>
     </xs:complexType>
   </xs:element>
```

ekluzek:
Make testlist schema more strict so it will catch errors, instead of silently ignoring issues with how it was configured, especially for options.

Also remove generic GSW forcing option for DATM. Add scripts/cases to gitignore.

Test namelist changes: None
Test status: bit for bit
Also made sure query_testlist works for components in cesm2_0_alpha10a
Fixes #2245, #2207 #1749

User interface changes?: no

Update gh-pages html (Y/N)?: no

Code review:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -11,3 +11,8 @@ doc/build
 
 # ignore the JENKINS files created when make html is run
 scripts/Tools/JENKINS*
+
+# Ignore anything that are produced under scripts "cases" directory
+
+/scripts/cases/
+
--- a/config/xml_schemas/testlist.xsd
+++ b/config/xml_schemas/testlist.xsd
@@ -4,7 +4,7 @@
   <xs:element name="testlist">
     <xs:complexType>
       <xs:sequence>
-        <xs:element maxOccurs="unbounded" ref="test"/>
+        <xs:element minOccurs="1" maxOccurs="unbounded" ref="test"/>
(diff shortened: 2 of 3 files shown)


jedwards4b:
@fischer-ncar This does not require review from @bertinia

PR #2305, opened 2018-02-20
add support for provenance in cesm

### LINKED ISSUE #2289: MODEL_VERSION doesn't get set correctly for non CESM checkouts...
For cesm MODEL_VERSION gets set by reading in the ChangeLog at the top of SRCROOT as follows in case.py. This is with the latest master.

```
    def set_model_version(self, model):
        version = "unknown"
        srcroot = self.get_value("SRCROOT")
        if model == "cesm":
            changelog = os.path.join(srcroot,"ChangeLog")
            if os.path.isfile(changelog):
                for line in open(changelog, "r"):
                    m = re.search("Tag name: (cesm.*)$", line)
                    if m is not None:
                        version = m.group(1)
                        break
```
This logic requires that ChangeLog exist at the top of SRCROOT, and that the tag name start with cesm. Neither of these things are true for CLM checkouts. Nor would be true for a CAM checkout.

This problem doesn't happen for acme since they use get_current_commit from utils.py, which fills the value with the SHA1 using a git command (git rev-parse HEAD). For acme, they also save a bunch of provenance information including the value from "git describe".

Since, cesm has now moved over to git, could we have MODEL_VERSION for cesm set by "git describe" of SRCROOT rather than the ChangeLog file? 'git describe' is nice since it gets you the most recent tagname plus part of the SHA from any commits since that tag. For example, 'git describe' on current master of cesm gives you "cesm2_0_alpha09a".

@billsacks @bandre-ucar @gold2718 @cacraigucar

jedwards4b:
add provenance output to README.case by running the manage_externals/checkout_externals script with --verbose and --status - this gives the status of all cesm or related model externals

Test suite: by hand with cesm and ctsm
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2289 
User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -638,7 +638,7 @@ class Case(object):
                    "Config file {} for component {} not found.".format(comp_config_file, comp_name))
             compobj = Component(comp_config_file, comp_class)
             # For files following version 3 schema this also checks the compsetname validity
-            
+
             self._component_description[comp_class] = compobj.get_description(self._compsetname)
             expect(self._component_description[comp_class] is not None,"No description found in file {} for component {} in comp_class {}".format(comp_config_file, comp_name, comp_class))
             logger.info("{} component is {}".format(comp_class, self._component_description[comp_class]))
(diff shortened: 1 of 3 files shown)


jedwards4b:
Okay model version is now taken from git.

Issue #2177, opened 2018-01-01
Weird history from ACME merge a few months ago


billsacks:
Nothing needs to be done about this - indeed, I doubt anything reasonable could be done about it at this point. But I wanted to point it out so that we can hopefully identify the cause and prevent oddness like this in the future.

If I do a git blame of `scripts/fortran_unit_testing/run_tests.py` (I suspect I'd see the same for other files; I just noticed it for that one), I see most of the lines labeled as

```
^587cf5950 (Azamat Mametjanov 2017-05-20 01:34:50 +0000
```

The leading caret apparently means that there is no parent of this commit. That commit was merged to master in the big acme merge:

```
commit 0c1b306aa
Merge: 4f201fe2d 2df9636fa
Author: James Foucar <jgfouca@sandia.gov>
Date:   Mon Oct 23 11:39:42 2017 -0600

    Merge pull request #1970 from ESMCI/jgfouca/branch-for-acme-split-10-17-2017-pr

    Bring in E3SM cime changes for past 3 months
```

Looking at the history along the branch that resulted in that ACME merge, it looks like it's common for the commits to be parentless according to the git history (presumably because they're extracted out of the ACME history). But I'm quite sure that Azamat was not the original author of most of the lines in run_tests.py, as it now appears according to a `git blame`.

@jgfouca or others: Any ideas how we can maintain proper history of files in the future (via git blame, etc.)?

jgfouca:
@billsacks I think we are at the mercy of git subtree unfortunately.

billsacks:
@jgfouca Spot-checking a few other files, it looks like this problem is wide-spread. For example, `git blame src/drivers/mct/shr/seq_flds_mod.F90` makes it look like that file was written almost entirely by Pat Worley, and `git blame src/drivers/mct/main/cime_comp_mod.F90` makes it look like that file was written almost entirely by Steve Ghan. This makes `git blame` and similar history tools all but useless. It's hard for me to believe that this wide-spread history mess-up is an unavoidable result of using git subtree. If there is truly no way to avoid this with git subtree then I'd oppose further uses of git subtree to bring changes back into cime, since I feel that these history mess-ups are a pretty big deal.

If you can't figure out what aspect of ACME's subtree-based workflow is responsible for this – and thus, what could be changed to prevent it – then could you please describe this workflow in detail so that I or others can help you figure out if anything can be done to prevent this problem in the future?

jgfouca:
[...]

PR #2312, opened 2018-02-22
Jgfouca/branch for acme split 2018 02 22

jgfouca:
Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


scripts/lib/e3sm_cime_mgmt.py
```diff
@@ -76,7 +76,7 @@ def get_branch_from_tag(tag):
 def do_subtree_split(old_split_tag, new_split_tag, merge_tag):
 ###############################################################################
     subtree_branch = get_branch_from_tag(new_split_tag)
-    run_cmd_no_fail("git subtree split {}.. --prefix=cime --onto={} -b {}".\
+    run_cmd_no_fail("git subtree split {}.. --prefix=cime --squash --onto={} -b {}".\
```
jgfouca: The squash and the explicit commit list are going to be removed

jgfouca:
@billsacks , this is the first split with your recommended changes from https://github.com/ESMCI/cime/issues/2177#issuecomment-366413324 ( IE , no explicit commit list and no squash for the git subtree split ).

billsacks:
@jgfouca I spot-checked some files and didn't notice any history problems. It was a bit hard to tell for sure because many of the changed files here already had history issues from earlier ACME merges (so there were a lot of lines in git blame that already started with `^`), but from what I can tell from my spot-checks, this iteration hasn't introduced any problems.

Thank you for making this change!

billsacks:
So, just to be explicit about it: I'm comfortable with your merging this to master without `--squash` this time.

PR #2315, opened 2018-02-23
Add comment pes file

### LINKED ISSUE #2248: Request for the comment in config_pesall.xml to be included in the generated env_mach_pes.xml file
Would be nice if we included any comment about the PE layout that is in the master config_pesall.xml file into the generated env_mach_pes.xml file that is written in the case dir.

For example:
```
      <pes compset=".*CAM5.+CLM45.+MPASCICE.+MPASO.+MOSART.*" pesize="S">
        <comment>cori-knl ne120-wcycl on 106 nodes (hmod106)</comment>
        <MAX_MPITASKS_PER_NODE>33</MAX_MPITASKS_PER_NODE>
        <MAX_TASKS_PER_NODE>264</MAX_TASKS_PER_NODE>
```

jedwards4b:
Add the comment from config_pes.xml to the env_mach_pes.xml file

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2248 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/env_mach_pes.py
+++ b/scripts/lib/CIME/XML/env_mach_pes.py
@@ -20,7 +20,7 @@ class EnvMachPes(EnvBase):
 
     def add_comment(self, comment):
         if comment is not None:
-            tmp = GenericXML(root_name_override="comment",infile="stupid.file",read_only=False)
+            tmp = GenericXML(root_name_override="comment", infile="fake", read_only=False)
             tmp.set_text(tmp.root, comment)
             self.add_child(tmp.root, position=1)
 
--- a/scripts/lib/CIME/XML/env_mach_pes.py
+++ b/scripts/lib/CIME/XML/env_mach_pes.py
@@ -20,9 +20,11 @@ class EnvMachPes(EnvBase):
 
     def add_comment(self, comment):
         if comment is not None:
-            tmp = GenericXML(root_name_override="comment", infile="fake", read_only=False)
-            tmp.set_text(tmp.root, comment)
-            self.add_child(tmp.root, position=1)
+            node = self.make_child("comment", text=comment)
+            # make_child adds to the end of the file but we want it to follow the header
+            # so we need to remove it and add it in the correct position
+            self.remove_child(node)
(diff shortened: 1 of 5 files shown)


scripts/lib/CIME/XML/env_mach_pes.py
```diff
@@ -17,6 +18,12 @@ def __init__(self, case_root=None, infile="env_mach_pes.xml", components=None):
         schema = os.path.join(get_cime_root(), "config", "xml_schemas", "env_mach_pes.xsd")
         EnvBase.__init__(self, case_root, infile, schema=schema)
 
+    def add_comment(self, comment):
+        if comment is not None:
+            tmp = GenericXML(root_name_override="comment", infile="fake", read_only=False)
```
jedwards4b: This is a bit awkward, but I couldn't find a cleaner way.

jgfouca: @jedwards4b , I see now that make_child cannot support what you want since it attaches the child as a SubElement. Probably the easiest thing here would be:

```
comment = self.make_child("comment", text=comment)
self.remove_child(comment)
self.add_child(comment, position=1)
```

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -170,13 +166,16 @@ def name(self, node):
...
 
-    def add_child(self, node, root=None):
+    def add_child(self, node, root=None, position=0):
         """
         Add element node to self at root
         """
         expect(not self.locked and not self.read_only, "locked")
         root = root if root is not None else self.root
-        root.xml_element.append(node.xml_element)
+        if position > 0:
```
jgfouca: 0 is a valid position. I think we want position=None to be default and this if check should check against None.

jedwards4b:
Not sure that your suggestion is any less awkward than what I came up with?

jgfouca:
@jedwards4b , I think it's a bit better since it doesn't involve creating an entire fake file.

jedwards4b:
done

PR #2317, opened 2018-02-23
New allow_error field for module system

### LINKED ISSUE #1625: CIME assumes all output from "module" commands are errors
I realized this when I was loading a module on Summitdev at OLCF that always gives output, and ./case.setup failed. The details are below, and this is for commit 7822b6b8903feb26c4171160423b472351ddb035. ~~Also, I noticed that the cime/utils/python/CIME directory now only contains compiled python code. Am I missing something, or do we no longer have access to the CIME python source code?~~ Nevermind, I found the cime/scripts/lib/CIME directory now. Thanks.

```
[imn@summitdev-login1:~/ACME/cime/scripts/summit_test1] 8-) ./case.setup 
ERROR: module command module load DefApps failed with message:
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
There are messages associated with the following module(s):
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------

DefApps:
   Please report any missing or needed software versions to help@olcf.ornl.gov
---------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------------
```

Just to be sure about this, I tested a module ls command, and got the following:

```
[imn@summitdev-login1:~/ACME/cime/scripts/summit_test1] 8-) ./case.setup 
ERROR: module command module ls  failed with message:
Currently Loaded Modules:
  1) xl/20161123   2) spectrum_mpi/10.1.0.2-20161221   3) hsi/5.0.2.p5   4) xalt/0.7.5   5) DefApps   6) python/3.5.2
```

jgfouca:
This is necessary since implementations of environment modules are
highly variable and some systems produce stderr output even when
things are working fine.

Default this to false since we want to catch errors on standard
systems.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1625 

User interface changes?: Y, new allow_error attribute for module systems

Code review: @jedwards4b

### DIFF
--- a/config/e3sm/machines/config_machines.xml
+++ b/config/e3sm/machines/config_machines.xml
@@ -571,7 +571,7 @@
         <arg name="tasks_per_node"> --map-by ppr:{{ tasks_per_numa }}:socket:PE=$ENV{OMP_NUM_THREADS} --bind-to hwthread</arg>
       </arguments>
     </mpirun>
-    <module_system type="module">
+    <module_system type="module" allow_error="false">
       <init_path lang="python">/usr/share/Modules/init/python.py</init_path>
       <init_path lang="perl">/usr/share/Modules/init/perl.pm</init_path>
       <init_path lang="sh">/usr/share/Modules/init/sh</init_path>
--- a/config/xml_schemas/env_mach_specific.xsd
+++ b/config/xml_schemas/env_mach_specific.xsd
@@ -73,6 +73,7 @@
         <xs:element minOccurs="0" maxOccurs="unbounded" ref="modules"/>
       </xs:sequence>
       <xs:attribute name="type" use="required" type="xs:NCName"/>
+      <xs:attribute name="allow_error" type="xs:boolean"/>
     </xs:complexType>
   </xs:element>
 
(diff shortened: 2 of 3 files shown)


jedwards4b:
Fixed in #2654

lvankampenhout:
Thanks Jim

jgfouca:
@jedwards4b , does E3SM not do XSD checking on this file? We've had allow_error on for months.

rljacob:
Its probably version 1 vs version 2 behavior.

jedwards4b:
It seems like if you created a case on melvin, summit or summitdev it would have an error.  The test is on env_mach_specific.xml rather than on the config_machines.xml file.

jgfouca:
@jedwards4b , you'd think so, but i've created 10's of thousands of cases on melvin since this was introduced.

jedwards4b:
I think @rljacob may be right and it's the version , but then I don't understand why @lvankampenhout ran into the problem - is he using cesm or e3sm?

jgfouca:
@jedwards4b he may be the first CESM person to try to use it.

jedwards4b:
ah - that makes sense then.

lvankampenhout:
yes I'm using CESM

PR #2320, opened 2018-02-25
Auto-regen batch files in case_submit.

### LINKED ISSUE #2259: Changing *any* env_batch.xml variable requires rerunning case.setup
The recent locked file support improvement introduced a minor issue - 
changing JOB_QUEUE or JOB_WALLCLOCK_TIME now require rerunning 
case.setup.  This should be a minor inconvenience for most users and its 
kind of a better safe than sorry situation, because not all machines can be 
configured such that these changes do not require running case.setup.

jgfouca:
We continue to lock env_batch.xml at case.setup time. In case.submit, if env_batch.xml != locked/env_batch.xml, we regenerate batch scripts, otherwise, we accept them as-is.
This will allow the batch scripts to auto-regen if the user has made changes the "right" way via xmlchange. It will also allow and preserve hand-edits as long as the user had not made any xmlchanges to env_batch.

Test suite: scripts_regression_tests (on batch and non-batch)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2259 

User interface changes?: Yes, user edits to batch files are no longer supported

Code review: @jedwards4b @billsacks @jonbob

### DIFF
--- a/scripts/lib/CIME/case_submit.py
+++ b/scripts/lib/CIME/case_submit.py
@@ -48,10 +48,12 @@ def _submit(case, job=None, no_batch=False, prereq=None, resubmit=False,
     if not resubmit:
         if case.get_value("TEST"):
             case.set_value("IS_FIRST_RUN", True)
+
         if no_batch:
             batch_system = "none"
         else:
             batch_system = env_batch.get_batch_system_type()
+
         case.set_value("BATCH_SYSTEM", batch_system)
     else:
         if env_batch.get_batch_system_type() == "none":
@@ -60,6 +62,10 @@ def _submit(case, job=None, no_batch=False, prereq=None, resubmit=False,
         # This is a resubmission, do not reinitialize test values
         if case.get_value("TEST"):
             case.set_value("IS_FIRST_RUN", False)
+
+    if env_batch.get_batch_system_type() != "none":
+        # May need to regen batch files if user made walltime changes
+        env_batch.make_all_batch_files(case)
 
     #Load Modules
     case.load_env()
--- a/scripts/lib/CIME/case_submit.py
+++ b/scripts/lib/CIME/case_submit.py
@@ -9,7 +9,7 @@ import socket
 from CIME.XML.standard_module_setup import *
 from CIME.utils                     import expect, run_and_log_case_status, verbatim_success_msg
 from CIME.preview_namelists         import create_namelists
-from CIME.check_lockedfiles         import check_lockedfiles
+from CIME.check_lockedfiles         import check_lockedfiles, check_lockedfile, unlock_file
 from CIME.check_input_data          import check_all_input_data
 from CIME.test_status               import *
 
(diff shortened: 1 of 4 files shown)


jgfouca:
This is not ready to merge, but I wanted to start a discussion. It seems like we need to more-clearly define CIME's interface and start enforcing it. A mix of user hand edits and xmlchanges is not sustainable.

To sum-up, I don't see how we can continue to support hand-editing the batch files if we're going to regenerate batch files every time case.submit is called, because hand-edits are now guaranteed to be lost when the files are regenerated.

This raises an issue that has come up many times in CIME: what is the supported interface? To me, the **only** supported interface is xmlchange/xmlquery and the various case.X tools. However, I know @jonbob and others will push back on this because they are used to doing hand edits. I believe even the run_e3sm script does sed commands on the batch scripts as part of its normal operation. We either need to convince users to code to interfaces (and not implementations) or we need to rethink CIME's interface entirely.

jgfouca:
To sum up the problem:
```
% ./case.setup
% ./case.build
% ./xmlchange JOB_WALLCLOCK_TIME=X subgroup=case.test
% * edit case.test and change JOB_WALLCLOCK_TIME to Y
% ./case.submit
which walltime is used?
current CIME master would throw a lockfile error here, which users hate
```

gold2718:
@jgfouca,
Thanks for putting this together. I am a big fan of sticking with a definable user interface (although I hate the specific names of `xmlquery` and `xmlchange`).

I agree with your assessment of the user interface but would add `create_X`, `manage_X`, and `query_X`.

rljacob:
Maybe for advanced users case.submit could have an option to only output a user-editable run script.  Then its the advanced user's job to edit and submit it from there.

jedwards4b:
I have been trying to discourage hand editing of the batch scripts for quite some time and those who have heeded this and started using the tools are generally satisfied with the results.   I think that we need to continue to encourage users to learn the tools and the prefered methods for interface.  I don't think that @rljacob suggestion is a good one because often those who will use it are anything but "advanced".

scripts/lib/CIME/case_submit.py
```diff
@@ -61,6 +63,10 @@ def _submit(case, job=None, no_batch=False, prereq=None, resubmit=False,
         if case.get_value("TEST"):
             case.set_value("IS_FIRST_RUN", False)
 
+    if env_batch.get_batch_system_type() != "none":
+        # May need to regen batch files if user made walltime changes
```
jedwards4b: walltime is misleading here - the changes could be anything in env_batch: - job_queue, wallclock time, handling of script error and output files - really anything that could be changed in that file.

rljacob:
The submit script is a particularly bad case to exert total control because....its just a script.   Anything can be put before or after the mpirun command and we just can't anticipate all the possible scenarios.    We now have a system that allows basic changes to be done safely but we need to allow advanced users to do advanced things.

jedwards4b:
I submit that we allow much more than you are aware of and that anything you might want to put in that script can be done without hand editing.

gold2718:
Another reason to only support a well-defined user interface is the effect of two behaviors I have seen over the years.

1) Experts-only mode allows experts to craft their own solutions but also encourages a diverging set of solutions to common problems. When the underlying implementation changes the generated script, the 'expert' scripts break and we get tons of complaints for breaking their custom scripts.

2) When everyone uses the same user interface, experts complain early and often about the features they need which are not supported. The scripts and/or user interface is improved to support those needs and everyone benefits.

Sure, waiting for the second option can take a bit of time and can be frustrating but the long-term benefits for the whole community are significant, and in my opinion, worthwhile.

jgfouca:
How about this as a compromise for now:

1) We lock env_batch.xml at case.setup time
2) In case.submit, if env_batch.xml != locked/env_batch.xml, we regenerate batch scripts, otherwise, we accept them as-is.

This would allow the batch scripts to auto-regen if the user has made changes the "right" way via xmlchange. It would also allow and preserve hand-edits as long as the user had **not** made any xmlchanges to env_batch.

This approach is a little clunky, because it opens the door to hand-edits being lost if a user does both hand-edits and xmlchanges. It also allows hand-edits to be lost in the case of a re-run of case.setup (this is the case on current master too).

billsacks:
@jgfouca I like your last suggestion. I agree that it isn't perfect, but it seems like a reasonable compromise. Can you record that design information in some form of documentation?

As for the broader question of interface: This isn't really addressing the batch scripts specifically, but more the question of "should we allow users to edit xml files": Now that I'm experienced with the system and generally know what I want to change, I exclusively use xmlquery / xmlchange. But when I started using the system, I almost always edited xml files by hand - partly because it helped me explore the available options, and partly because then I could see what was happening (I think that, at first, I wasn't confident in my use of xmlchange or exactly what it accomplished). Anecdotally, my observations of other new-ish users have suggested a similar pattern. So, while I encourage people to use xmlquery/xmlchange, I'd be reluctant to dis-allow direct editing of the xml files. Long-term, if we put a GUI on top of the case xml files, this could be the best of all worlds - at which point we could support two interfaces - xmlquery/xmlchange or the GUI - and disallow direct editing of the xml files.

rljacob:
So does this now have the compromise you suggested?

jgfouca:
@rljacob , yes

jgfouca:
@jedwards4b , still testing. Will let you know when I'm satisfied.

jedwards4b:
Okay - I think that you may need to rebase in order for the travis-ci testing to work.

jgfouca:
@jedwards4b , all testing passed.

jgfouca:
@jedwards4b , I'm going to go ahead and merge this because I want to get this change over to e3sm.

PR #2326, opened 2018-02-27
Refactor handling of small jobs

### LINKED ISSUE #2209: Want ability to select default queue based on number of tasks, not just number of nodes
We used to have the ability for cime to select a default queue based on the number of tasks. With @jgfouca 's changes in #1918 (to fix #1913), this is no longer possible.

This is important for jobs that use less than a full node - such as single-point or other jobs that use a single task. CLM, CISM and possibly other components have used this capability in some configurations. Now these single-processor jobs are using a full node - so, on cheyenne, they are getting charged 36x what they should be.

Can we get back the capability to select queues based on the number of tasks?

@jedwards4b @ekluzek

jgfouca:
Refactor how "non-core" jobs, like case.st_archive are handled by our system.

1) Since we allow a non-core job to specify num_tasks, we should also let it specify walltime
2) Remove unused pedocumentation
3) Allow --job argument to preview run
4) Fix significant bug in env_batch (remove aliasing of elements)
5) user-requested walltime/queue ONLY applies to "core" job.

With these changes, assuming a system has a well-configured batch system, this should allow intelligent selection of a "small job" queue for non-core jobs.

Test suite: by-hand, code_checker
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2209 

User interface changes?: Yes

### DIFF
--- a/config/cesm/machines/config_batch.xml
+++ b/config/cesm/machines/config_batch.xml
@@ -556,6 +556,7 @@
     <job name="case.st_archive">
       <template>template.st_archive</template>
       <task_count>1</task_count>
+      <walltime>0:10:00</walltime>
       <!-- If DOUT_S is true and case.run (or case.test) exits successfully then run st_archive-->
       <dependency>case.run or case.test</dependency>
       <prereq>$DOUT_S</prereq>
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -112,7 +112,7 @@ class EnvBatch(EnvBase):
 
         childnodes = []
         for child in reversed(orig_group_children):
-            childnodes.append(self.copy(child))
+            childnodes.append(child)
 
         self.remove_child(orig_group)
 
@@ -124,7 +124,7 @@ class EnvBatch(EnvBase):
                 self.make_child("type", root=node, text="char")
 
             for child in childnodes:
-                self.add_child(child, root=new_job_group)
+                self.add_child(self.copy(child), root=new_job_group)
 
     def cleanupnode(self, node):
         if self.get(node, "id") == "batch_system":
@@ -192,7 +192,7 @@ class EnvBatch(EnvBase):
             force_queue = case.get_value("USER_REQUESTED_QUEUE", subgroup=job) if case.get_value("USER_REQUESTED_QUEUE", subgroup=job) else None
             logger.info("job is {} USER_REQUESTED_WALLTIME {} USER_REQUESTED_QUEUE {}".format(job, walltime, force_queue))
             task_count = jsect["task_count"] if "task_count" in jsect else None
-            walltime = jsect["walltime"] if ("walltime" in jsect and walltime is None) else None
+            walltime = jsect["walltime"] if ("walltime" in jsect and walltime is None) else walltime
             if task_count is None:
                 node_count = case.num_nodes
             else:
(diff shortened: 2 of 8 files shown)


scripts/lib/CIME/case.py
```diff
@@ -957,6 +952,12 @@ def configure(self, compset_name, grid_name, machine_name=None,
 
         env_batch.set_batch_system(batch, batch_system_type=batch_system_type)
         env_batch.create_job_groups(bjobs)
+
+        if walltime:
```
jgfouca: Important change. Limit user requests to specific job groups that only exist after create_job_groups.

config/cesm/machines/config_batch.xml
```diff
@@ -556,6 +556,7 @@
     <job name="case.st_archive">
       <template>template.st_archive</template>
       <task_count>1</task_count>
+      <walltime>0:10:00</walltime>
```
jgfouca: Is 10 minutes long enough for all possible case.st_archive runs?

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -124,7 +124,7 @@ def create_job_groups(self, batch_jobs):
                 self.make_child("type", root=node, text="char")
 
             for child in childnodes:
-                self.add_child(child, root=new_job_group)
+                self.add_child(self.copy(child), root=new_job_group)
```
jgfouca: We need separate copies for each job, otherwise we have to deal with aliasing.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -193,6 +192,7 @@ def set_job_defaults(self, batch_jobs, case):
             force_queue = case.get_value("USER_REQUESTED_QUEUE", subgroup=job) if case.get_value("USER_REQUESTED_QUEUE", subgroup=job) else None
             logger.info("job is {} USER_REQUESTED_WALLTIME {} USER_REQUESTED_QUEUE {}".format(job, walltime, force_queue))
             task_count = jsect["task_count"] if "task_count" in jsect else None
+            walltime = jsect["walltime"] if ("walltime" in jsect and walltime is None) else walltime
```
jgfouca: Here is where new walltime parameter is used.

config/cesm/machines/config_batch.xml
billsacks: I'd probably err on the side of making it longer - maybe 20 or 30 min?

scripts/lib/CIME/XML/env_batch.py
jedwards4b: didn't you just move this from line 115?

billsacks:
I don't think this actually addressed #2209 , since my request related to the core job, and my understanding is that this PR only addressed non-core jobs.

scripts/lib/CIME/XML/env_batch.py
jgfouca: Yes

jgfouca:
@billsacks , i thought this was to better-support the jobs that set num_tasks=1, which is always a non-core job? Maybe I misunderstood. Were you talking about core SMS_P1 jobs?

jgfouca:
@billsacks , I would assume, for any core job, if the underlying machine is configured correctly (defaulting to the smaller queue), then everything would be fine?

jedwards4b:
I think that the issue is that (case.run) jobs which only use a single task end up using a full node because we can't specify a special share queue for jobs that use < a node.

billsacks:
@jedwards4b stated this correctly. The issue is that we have a number of cases where the CESM case only uses a single processor, or possibly a few processors (the latter most often occurs in testing - e.g., multi-instance tests of cases that would otherwise use a single processor). We used to be able to specify that these should use a share queue, but that's no longer possible.

billsacks:
By the way: sorry for the unclear initial statement of the problem, and for not realizing until this was merged that the PR wasn't actually addressing the issue. I still think that the capabilities you put in place are useful, though.

PR #2327, opened 2018-02-28
Reinstate ability to select queue by num tasks

### LINKED ISSUE #2209: Want ability to select default queue based on number of tasks, not just number of nodes

jgfouca:
Test suite: scripts_regression_tests (on batch machine)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2209 

User interface changes?: N


jgfouca:
I'm unable to test this new feature without access to cheyenne.

jedwards4b:
I did suggest a way to test on nersc systems.  Edison and cori have a debug queue that should work for this purpose.

jgfouca:
@jedwards4b , i think the debug queue supports a few nodes. At least is does in the e3sm configuration. I suppose I could temporarily change the config to use num_tasks=1.

jgfouca:
@jedwards4b , my test:

change sandiatoss3 queue settings:
```
  <batch_system MACH="sandiatoss3" type="slurm" >
    <queues>
      <queue jobmin="1" jobmax="1" walltimemax="04:00:00" strict="true" default="true">short</queue>
      <queue nodemin="1" nodemax="64" walltimemax="24:00:00">batch</queue>
    </queues>
  </batch_system>
```

Then do:
```
% ./create_test ERS_P1.f19_g16_rx1.A --no-build

% ./create_test ERS_P2.f19_g16_rx1.A --no-build
```

I confirmed that the P1 case uses the "short" queue via preview_run. The P2 case uses "batch".

jedwards4b:
On this branch I add to config_batch.xml:
```      <queue walltimemax="06:00" jobmin="1" jobmax="18">share</queue>```
and tried a test SMS_P1.f09_g16.X
But preview run shows
```BATCH SUBMIT: (nodes 1)
    case.test -> qsub    -q regular -l walltime=12:00:00 -A P93300606 -m a .case.test ```

jgfouca:
@jedwards4b , the system will prefer the default queue if it's compatible.

jgfouca:
This is why, I generally prefer to make the "smaller" queue the default whenever possible.

jedwards4b:
Right, once I did that I get the correct queue for case.run and case.test, but the case.st_archive is still going to the other (regular) queue - is task_count being interpreted somewhere as node count?

jgfouca:
@jedwards4b , shoot! Let me try to fix that.

jgfouca:
@billsacks , i added comments.

jgfouca:
@jedwards4b , bug fixed.

billsacks:
@jgfouca looks good - thanks!

PR #2339, opened 2018-03-02
Env batch cleanup

jedwards4b:
Remove most usage of scan_children and replace with root discovery and get_children.
There are a couple of unrelated changes here that clean up README.case and move the cesm
version information to CaseStatus

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -156,16 +156,20 @@ class EnvBatch(EnvBase):
             self.set_batch_system_type(batch_system_type)
 
         if batchobj.batch_system_node is not None and batchobj.machine_node is not None:
-            for node in batchobj.get_children(root=batchobj.machine_node):
-                oldnode = batchobj.get_optional_child(self.name(node), root=batchobj.batch_system_node)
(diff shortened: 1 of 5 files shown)


scripts/lib/CIME/XML/env_batch.py
```diff
@@ -591,21 +608,31 @@ def get_queue_specs(self, queue):
...
-            node = self.get_optional_child("queue")
+        bs_nodes = self.get_children("batch_system")
+        for bsnode in bs_nodes:
+            qnodes = self.get_children("queues", root=bsnode)
+            for qnode in qnodes:
+                node = self.get_optional_child("queue", attributes={"default" : "true"}, root=qnode)
+                if node is None:
+                    node = self.get_optional_child("queue", root=qnode)
         expect(node is not None, "No queues found")
         return node
```
amametjanov: Need an initialization above: e.g. `node = None`.

jgfouca: @amametjanov , done (i pushed directly to master).

PR #2341, opened 2018-03-05
Mpirun depends on queue

jedwards4b:
Add support for environment and module to depend on job, reorganize env_mach_specific

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/env_mach_specific.py
+++ b/scripts/lib/CIME/XML/env_mach_specific.py
@@ -50,23 +50,6 @@ class EnvMachSpecific(EnvBase):
                 for node in nodes:
                     self.add_child(node)
 
-    def _get_modules_for_case(self, case, job=None):
-        module_nodes = self.get_children("modules", root=self.get_child("module_system"))
-        modules_to_load = None
-        if module_nodes is not None:
-            modules_to_load = self._compute_module_actions(module_nodes, case, job=job)
-
-        return modules_to_load
-
-    def _get_envs_for_case(self, case, job=None):
-        env_nodes = self.get_children("environment_variables")
-
-        envs_to_set = None
-        if env_nodes is not None:
-            envs_to_set = self._compute_env_actions(env_nodes, case, job=job)
-
-        return envs_to_set
-
     def load_env(self, case, force_method=None, job=None, verbose=False):
         """
         Should only be called by case.load_env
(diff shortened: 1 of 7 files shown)


jgfouca:
@jedwards4b , can you explain the env_mach_specific reorg briefly?

scripts/lib/CIME/XML/env_mach_specific.py
```diff
@@ -160,11 +119,156 @@ def make_env_mach_specific_file(self, shell, case):
...
+        exec_node = self.get_child("executable", root=the_match)
+        expect(exec_node is not None,"No executable found")
+        executable = self.text(exec_node)
+
+        return executable, args
+
+    def get_type_info(self, vid):
+        return "char"
 
     # Private API
```
jedwards4b: This comment did not separate the public from private parts of the class, now it does.

jedwards4b:
I've undone the rearrange and will redo it in another PR.

scripts/lib/CIME/XML/env_mach_specific.py
```diff
@@ -160,11 +160,16 @@ def make_env_mach_specific_file(self, shell, case):
...
+    def _load_envs(self, envs_to_set, verbose=False):
         for env_name, env_value in envs_to_set:
             os.environ[env_name] = "" if env_value is None else env_value
-
-    # Private API
+            if verbose:
+                if env_value is not None:
+                    logger.warning("Setting Environment {}={}".format(env_name, env_value))
+                else:
+                    logger.warning("Unsetting Environment {}".format(env_name))
```
jgfouca: This warning is not accurate. Setting an env var to "" is not the same as unsetting it. I'd prefer this block to match the block in load_module_modules:

```
            if verbose:
                logger.warning("Setting Environment {}={}".format(env_name, env_value))
            else:
                logger.debug("Setting Environment {}={}".format(env_name, env_value))
```

scripts/lib/CIME/XML/env_mach_specific.py
```diff
@@ -305,6 +313,8 @@ def _load_modules_generic(self, modules_to_load):
         # Use null terminated lines to give us something more definitive to split on.
         # Env vars can contain newlines, so splitting on newlines can be ambiguous
         cmd += " && env -0"
+        if verbose:
+            logger.warning("cmd: {}".format(cmd))
```
jgfouca: a debug print would be nice here too.

scripts/lib/CIME/case.py
```diff
@@ -136,6 +136,7 @@ def initialize_derived_attributes(self):
             "compiler" : self.get_value("COMPILER"),
             "mpilib"   : self.get_value("MPILIB"),
             "threaded" : self.get_build_threaded(),
+            "queue": self.get_value("JOB_QUEUE", subgroup="case.run")
```
jgfouca: This concerns me a lot as it appears to be a catch-22. Way down in the case creation code, it says that initialize_derived_attributes must be called before the env_batch file is properly set up. I think this is because queue choice depends on num_nodes. But it now looks like initialize_derived_attributes depends env_batch being set up already.

jedwards4b: We don't actually need this, I added it before tracking down the real problem.   Removed.

scripts/lib/CIME/XML/env_mach_specific.py
jedwards4b: done

scripts/lib/CIME/XML/env_mach_specific.py
jedwards4b: I'm now actually unsetting the env var if it exists so the statement is correct.

jedwards4b:
Rerunning tests now.

jedwards4b:
All tests pass, ready to merge.

PR #2344, opened 2018-03-09
Provide infrastructure to allow SCM to run with SE dycore 

bogensch:
Changes to two files to allow the single column model (SCM) to be run with Spectral Element (SE) dynamical core.  The changes are necessary because the current infrastructure assumes a 2d grid compatability when determining the nearest longitude/latitude point to the SCM column.  However, SE input grids are 1D.   

Test suite: acme_developer tests
Test baseline:  test Eulerian SCM to ensure b4b
Test namelist changes: 
Test status: [bit for bit]

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/share/util/shr_scam_mod.F90
+++ b/src/share/util/shr_scam_mod.F90
@@ -271,6 +271,10 @@ subroutine shr_scam_getCloseLatLonNC(ncid, targetLat,  targetLon, closeLat, clos
    if ( allocated(londimnames) ) deallocate(londimnames)
    if ( allocated(vars)        ) deallocate( vars )
 
+   if (nlat .eq. nlon) then !-- Enforce for SE grids
+     closelatidx = 1
+   endif
+
    return
 
 end subroutine shr_scam_getCloseLatLonNC
@@ -326,7 +330,7 @@ subroutine shr_scam_getCloseLatLonPIO(pioid, targetLat,  targetLon, closeLat, cl
    integer(IN)                      ::  ndimid
    integer(IN)                      ::  strt(nf90_max_var_dims),cnt(nf90_max_var_dims)
    integer(IN)                      ::  nlon = 0, nlat = 0
-   logical                          ::  lfound        ! local version of found
+   logical                          ::  lfound, islatitude        ! local version of found
    integer(IN), dimension(nf90_max_var_dims) :: dimids
    character(len=80), allocatable   ::  vars(:)
    character(len=80), allocatable   ::  latdimnames(:)
@@ -416,6 +420,14 @@ subroutine shr_scam_getCloseLatLonPIO(pioid, targetLat,  targetLon, closeLat, cl
    nlon=0
    nvarid=0
 
+   if (latlen .eq. 1 .and. lonlen .gt. 1) then
+     latlen=lonlen
+     islatitude=.false. ! if spectral element lat and lon 
+                        !   are on same array structure
+   else
+     islatitude=.true.
+   endif
+
    !--- Loop through all variables until we find lat and lon ---
 
    do while (nvarid < nvars .and.(nlon.eq.0 .or. nlat.eq.0))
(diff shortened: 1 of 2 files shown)


mvertens:
What tests were run to verify this PR? We need to be very careful that this
does not change any of the previous behavior - since CESM relies
extensively on the data models - and data model share code has been
modified. I see no summary of the testing here to ensure that bfb behavior
is maintained for previous configurations. In fact I see no testing summary
here.

src/share/util/shr_scam_mod.F90
```diff
@@ -271,6 +271,10 @@ subroutine shr_scam_getCloseLatLonNC(ncid, targetLat,  targetLon, closeLat, clos
    if ( allocated(londimnames) ) deallocate(londimnames)
    if ( allocated(vars)        ) deallocate( vars )
 
+   if (nlat .eq. nlon) then !-- Enforce for SE grids
```
jedwards4b: This doesn't seem like a reliable test of what grid you are on - can you do better?

src/share/util/shr_scam_mod.F90
```diff
@@ -935,8 +957,26 @@ subroutine get_close( targetlon, targetlat, nlon, lons, nlat, lats, closelonidx,
 
    !--- find index of value closest to 0 and set returned values ---
 
-   closelonidx=(MINLOC(abs(poslons-postargetlon),dim=1))
-   closelatidx=(MINLOC(abs(lats-targetlat),dim=1))
+   !--- if SCM uses Eulerian grids
+   if (nlat .ne. nlon) then
```
jedwards4b: Again I don't think this is the best test to use to determine what grid you are on.

mvertens:
@bogensch - sorry if my initial comment came out as snippy. I appreciate your taking this on. I just wanted to make sure that enough testing was done to the data models to ensure that everything is bfb where expected.

bogensch:
No worries @mvertens , sorry for not documenting my testing.  I ran the full suite of acme_developer tests and tested the standard Eulerian SCM.  Once I make the requested changes, I will run the scripts_regression_tests.py  test as requested by Jim.

mvertens:
@bogensch - thanks! I think it would be helpful if a few data model
configurations could be run - such as I, C, G, D compsets to make sure that
that they are not effected. We are happy to do this on our end as well
before the PR is accepted.

rljacob:
The acme_developer you ran is good enough to substitute for scripts_regression_tests since this PR only touches fortran code.

src/share/util/shr_scam_mod.F90
```diff
@@ -763,7 +781,8 @@ logical function is_latlon( var_name, latitude, varnotdim )
            trim(var_name) == trim(xyvar)    .or. trim(var_name) == 'lsmlat'    .or. &
            trim(var_name) == trim(gcvar)    .or.                                    &
            trim(var_name) == 'LAT'          .or. trim(var_name) == 'LATIXY'    .or. &
-           trim(var_name) == trim(Capxyvar) .or. trim(var_name) == 'LSMLAT' )  then
+           trim(var_name) == trim(Capxyvar) .or. trim(var_name) == 'LSMLAT'    .or. &
+           trim(var_name) == 'ncol')  then
```
mvertens: is this correct for lat/lon grids? ncol is typically not lat/lon

bogensch:
crap @jedwards4b , that was an accidental push before the changes were thoroughly tested.  Not all my tests passed with those changes.... will update once I have a better tested solution.

src/share/util/shr_scam_mod.F90
bogensch: it appears this call is checking to see if the variable is a lat/lon variable in the file and checking for appropriate dimensions for the lat/lon array.  Also, yes, I verified that all A tests in the acme_developer package were b4b with the latest code push.

bogensch:
@jedwards4b just pushed updated code that passes all acme_developer tests

PR #2347, opened 2018-03-12
Mfdeakin sandia/st archive/encapsulate datetime

### LINKED ISSUE #2162: use of python datetime in case_st_archive precludes dates beyond 9999
We don't need to use datetime here.

mfdeakin-sandia:
This removes the use of Python's datetime object from the short term archiver, enabling archiving of dates year >9999

Test baseline: 
Test namelist changes: 
Test status: BFB

Fixes  #2162

User interface changes?: N


mfdeakin-sandia:
I've run B_CheckCode and the LDSTA test, the rest of scripts_regression_tests are still in progress, but I expect no issues

scripts/lib/CIME/utils.py
```diff
@@ -1589,3 +1589,213 @@ def run_bld_cmd_ensure_logging(cmd, arg_logger, from_dir=None):
 
 def get_batch_script_for_job(job):
     return job if "st_archive" in job else "." + job
+
+class date:
```
billsacks: I feel like this class warrants its own file.

mfdeakin-sandia: I don't disagree, though I'm a little wary of cluttering the scripts directory with a utility class. If you're okay with it for this, I've moved it to a file just called `date.py`.

billsacks: Thanks.

PR #2355, opened 2018-03-13
Add ability to xmlquery derived attributes

### LINKED ISSUE #2351: Allow referencing case envs inside of shell commands
e.g.

```
$SHELL{echo $NUM_NODES * 6 | bc}
```

jgfouca:
Add ability to xmlquery derived attributes

Refactor xmlchange:
The code blocks that handled the two types of user input were diverging.
This commit encapsulates to core xmlchange operation. It also will cause
the user to be informed with a special error if they try to change a derived
attribute.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2351 

User interface changes?: xmlquery will now accept derived attributes

Code review: @jedwards4b

### DIFF
--- a/scripts/Tools/xmlquery
+++ b/scripts/Tools/xmlquery
@@ -190,10 +190,14 @@ formatter_class=argparse.RawDescriptionHelpFormatter)
         args.type, args.valid_values, args.partial_match, args.file
 
 def get_value_as_string(case, var, attribute=None, resolved=False, subgroup=None):
-    thistype = case.get_type_info(var)
-    value = case.get_value(var, attribute=attribute, resolved=resolved, subgroup=subgroup)
-    if value is not None and thistype:
-        value = convert_to_string(value, thistype, var)
+    if var in ["THREAD_COUNT", "TOTAL_TASKS", "TASKS_PER_NODE", "NUM_NODES", "SPARE_NODES", "TASKS_PER_NUMA", "CORES_PER_TASK"]:
+        value = str(getattr(case, var.lower()))
+    else:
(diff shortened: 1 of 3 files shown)


scripts/Tools/xmlquery
```diff
@@ -190,10 +190,14 @@ formatter_class=argparse.RawDescriptionHelpFormatter)
         args.type, args.valid_values, args.partial_match, args.file
 
 def get_value_as_string(case, var, attribute=None, resolved=False, subgroup=None):
-    thistype = case.get_type_info(var)
-    value = case.get_value(var, attribute=attribute, resolved=resolved, subgroup=subgroup)
-    if value is not None and thistype:
-        value = convert_to_string(value, thistype, var)
+    if var in ["THREAD_COUNT", "TOTAL_TASKS", "TASKS_PER_NODE", "NUM_NODES", "SPARE_NODES", "TASKS_PER_NUMA", "CORES_PER_TASK"]:
```
jedwards4b: I wonder if there should be a message attached to the output to the effect that these variables cannot be modified with xmlchange?

jgfouca: @jedwards4b , i'm not sure how to do that without cluttering the output in the case that the user wants the raw value.

jedwards4b: Can you use logger.info and suppress if --value is used?

jgfouca: @jedwards4b I have an idea.

jgfouca:
@jedwards4b , see what you think now.

PR #2357, opened 2018-03-13
Jgfouca/branch for acme split 2018 03 13

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b


scripts/lib/CIME/XML/pes.py
```diff
@@ -96,7 +96,7 @@ def _find_matches(self, grid_nodes, grid, compset, machine, pesize_opts, overrid
                             pesize_match = self.get(pes_node, "pesize")
                             compset_match = self.get(pes_node, "compset")
                             if (pesize_match == "any" or (pesize_opts is not None and \
-                                                          re.search(pesize_match, pesize_opts))) and \
+                                                          pesize_match == pesize_opts)) and \
```
jgfouca: This is a meaningful change.

scripts/lib/CIME/XML/env_mach_specific.py
```diff
@@ -246,7 +246,6 @@ def _get_module_commands(self, modules_to_load, shell):
         cmds = []
         last_action = None
         last_cmd    = None
-        actions_that_cannot_be_batched = ["swap"]
```
jedwards4b: seems like a fix is being undone here?

scripts/lib/CIME/XML/pes.py
jgfouca: @amametjanov explained this to me:

re.search(pattern, string) returns true if the regular expression pattern is a substring anywhere in the string. If there are 2 or more PE layouts with pesize=x and pesize=y, where x is a substring of y or y is a substring of x, then these PE layouts have the same match score and the resulting error is

    Errors were:
        mach_choice theta mach_match theta
        grid_choice a%ne120np4 grid_match a%ne120np4
        compset_choice .*CAM5.+CLM45.+MPASCICE.+MPASO.+MOSART.* compset_match .*CAM5.+CLM45.+MPASCICE.+MPASO.+MOSART.*
        pesize_choice L pesize_match XL  
        points = 13
        ERROR: More than one PE layout matches given PE specs
I think pecount/pesize naming conventions are

S - small
M/any/default - medium
L - large
XL - extra-large
XS - extra-small
T - threaded
So this PR tries to follow that convention with pesize=MT for medium-threaded and pesize=XLT for extra-large-threaded PE layouts. This requires change from regular expressions to exact equality, because L is a substring of XL.

jgfouca:
Annotations complete, only 1 line needs to be looked at.

scripts/lib/CIME/XML/pes.py
jedwards4b: I have been arguing that we had too many possibilities here and should just reduce it to S,M,L but you are going in the opposite direction.   As I write this I realize that it's not relevant to this change.

ndkeen: fwiw, I've been wanting more options here as well.  For one, I think having more sizes is always better (T, XL, X1, X2?)  And, for example, we are thinking of having different PE layouts based on the amount of high frequency output.  Say T for tiny and THF for tiny with high-frequency output.  OR, maybe not call out HF output, but something like favoring using less memory, ie fewer MPI's, more threads.  TDEBUG?  I don't know.

scripts/lib/CIME/XML/env_mach_specific.py
jgfouca: looks that way, but this is actually expected. Look below the comment and you'll see this variable is being set.

jedwards4b: i see, thanks

billsacks:
As @rljacob pointed out on Slack, there's something funny with the history here: If I'm viewing it right, it looks like the history of this branch contains commits back to more than a year ago:

```
git log --graph --abbrev-commit --decorate --format=format:'%C(yellow)%h%C(reset)%C(auto)%d%C(reset) - %C(green)(%ar)%C(reset) %C(black)%s%C(reset) %C(white)- %an%C(reset)' 3bdab2cad..575f0e6d4
```

This is different from the earlier problems we had with subtree, but still seems weird. Personally I don't care much about this, since I usually just look at first-parent history. But I wonder if there's still something non-standard in E3SM's use of subtrees. Alternatively, maybe it will just take a few iterations of the new subtree usage for things to sort themselves out.

rljacob:
We've had two merges to E3SM now with the revised process and both have giant commit histories even though the actual merge only touches the right number of files.

PR #2366, opened 2018-03-15
ignore files beginning in test_

jedwards4b:
Using mkSrcfiles to build the FMS libraries requires ignoring files with name pattern test_* 
I checked all of cime and all of the cesm component models to assure that this name pattern was never used.

Test suite: by hand, build new FMS library
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes
User interface changes?:  Source code files beginning in test_ will be ignored in build

Update gh-pages html (Y/N)?:

Code review:


billsacks:
Does this script have any mechanism to write output to the bld log file? If so, it might be a good idea to write out a message like, "Skipping file test_foo: Rename to not begin with 'test_' if this should be included in the build."

jedwards4b:
Print statement added and tested.

PR #2373, opened 2018-03-16
Fix skip-pnl for tests

jgfouca:
skip-pnl was being entirely ignored for case.test.

Also fixes argument passdown bug on some systems.

Test suite: scripts_regression_tests (both batch and desktop)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -433,7 +433,10 @@ class EnvBatch(EnvBase):
 
             function_name = job.replace(".", "_")
             if not dry_run:
-                locals()[function_name](case)
+                if "archive" not in function_name:
+                    locals()[function_name](case, skip_pnl=skip_pnl)
+                else:
+                    locals()[function_name](case)
 
             return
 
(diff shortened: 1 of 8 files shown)


scripts/lib/CIME/XML/env_batch.py
```diff
@@ -496,16 +499,23 @@ def _submit_single_job(self, case, job, dep_jobs=None, no_batch=False,
                "Unable to determine the correct command for batch submission.")
         batchredirect = self.get_value("batch_redirect", subgroup=None)
         submitcmd = ''
-        for string in (batchsubmit, submitargs, batchredirect, get_batch_script_for_job(job)):
-            if  string is not None:
-                submitcmd += string + " "
+        batch_env_flag = self.get_value("batch_env", subgroup=None)
```
jgfouca: Important: if the current machine passes arguments to the batch script via batch env flag, then this needs to appear in the command line before the batch script name, otherwise it needs to come after.

scripts/lib/CIME/case_run.py
```diff
@@ -124,11 +126,12 @@ def _run_model_impl(case, lid, skip_pnl=False, da_cycle=0):
 
                         case.set_value("CONTINUE_RUN",
                                        case.get_value("RESUBMIT_SETS_CONTINUE_RUN"))
-                        create_namelists(case)
```
jgfouca: Not sure how this ever worked since it came before the new lid?

PR #2374, opened 2018-03-16
Update MCT to 2.10rc2

rljacob:
Update MCT with changes from the 2.10-rc2 tag.

Only significant diff is slight change to workaround for the mpi libs
that have trouble when arguments in MPI_alltoallv are not allocated.
Also update strings in docs from 2.9 to 2.10

Test suite:  e3sm_developer
Test baseline:  BFB using a low res E3SM fully coupled case
Test namelist changes:  None
Test status: BFB

### DIFF
--- a/src/externals/mct/README
+++ b/src/externals/mct/README
@@ -1,6 +1,6 @@
 ######################################################################
 
- -- Mathematics + Computer Science Div. / Argonne National Laboratory
+ -- Argonne National Laboratory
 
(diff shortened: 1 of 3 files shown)


rljacob:
@jgfouca FYI wait for this be to be merged before next ESMCI to E3SM cime update.

jedwards4b:
@mvertens waiting for your review.

PR #2378, opened 2018-03-19
Compare two case setup

### LINKED ISSUE #2375: All two run tests failing - no namelists in case2
No namelist is produced for case2 of two case runs, causing a runtime error.

jedwards4b:
The PR #2373 causes some two test system test not to populate the case2 run directory with namelists.   This also occurs for the ERI test which clones ref1 and ref2 cases.   Fixed by explicitly running case_setup for both phases of comparetwo type tests and for each clone of the ERI test 

Test suite: cime_developer, 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2375 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -232,7 +232,7 @@ class SystemTestsCompareTwo(SystemTestsCommon):
             with self._case2:
                 logger.info('Doing second run: ' + self._run_two_description)
                 self._activate_case2()
-               # we need to make sure run2 is properly staged.
+                # we need to make sure run2 is properly staged.
                 if run_type != "startup":
                     check_case(self._case2)
 
@@ -457,7 +457,7 @@ class SystemTestsCompareTwo(SystemTestsCommon):
         # caller of this method).
         self._case.flush()
         # This assures that case one namelists are populated
-        # and creates the case.test script 
+        # and creates the case.test script
         case_setup(self._case, test_mode=False, reset=True)
 
         # Set up case 2
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -40,7 +40,6 @@ from CIME.case import Case
 from CIME.case_submit import check_case
 from CIME.case_st_archive import archive_last_restarts
 from CIME.utils import get_model
-from CIME.case_setup import case_setup
 
 import shutil, os, glob
 
@@ -458,7 +457,7 @@ class SystemTestsCompareTwo(SystemTestsCommon):
         self._case.flush()
         # This assures that case one namelists are populated
         # and creates the case.test script
-        case_setup(self._case, test_mode=False, reset=True)
+        self._case.case_setup(test_mode=False, reset=True)
 
         # Set up case 2
         self._activate_case2()
@@ -469,7 +468,7 @@ class SystemTestsCompareTwo(SystemTestsCommon):
         self._case.flush()
 
         # This assures that case two namelists are populated
-        case_setup(self._case, test_mode=True, reset=True)
+        self._case.case_setup(test_mode=True, reset=True)
 
         # Go back to case 1 to ensure that's where we are for any following code
         self._activate_case1()
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -40,6 +40,7 @@ from CIME.case import Case
 from CIME.case_submit import check_case
 from CIME.case_st_archive import archive_last_restarts
 from CIME.utils import get_model
+from CIME.case_setup import case_setup
 
 import shutil, os, glob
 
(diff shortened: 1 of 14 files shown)


scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -231,7 +232,7 @@ def run_phase(self, success_change=False):  # pylint: disable=arguments-differ
             with self._case2:
                 logger.info('Doing second run: ' + self._run_two_description)
                 self._activate_case2()
-                # we need to make sure run2 is properly staged.
+               # we need to make sure run2 is properly staged.
```
jgfouca: Intentional? Looks like a bad indent.

jgfouca:
We'd be able to delete even more lines if the base class implementation of case_N_setup was "pass" instead of NotImplemented error.

billsacks:
Thanks, @jedwards4b . This basically looks good, but I have one question before I approve: It looks like the previous code wasn't consistent about whether `test_mode` was set in the call to `case_setup`, either for case one or for case two. The new, shared code forces consistency in this respect. Is this going to cause any changes in behavior for any of the tests?

billsacks:
> We'd be able to delete even more lines if the base class implementation of case_N_setup was "pass" instead of NotImplemented error.

I'd object to that change. I think that forcing all of the tests to implement case_N_setup helps ensure that the API is being used correctly. That said, I don't feel super strongly about it, so I'd be open to being overruled.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
jedwards4b: fixed

jedwards4b:
unit tests are failing - I'll look into that tomorrow.

jedwards4b:
Oh, @billsacks I did note the inconsistent behavior and tested that the choice I made worked in all cases.

billsacks:
Thanks a lot, @jedwards4b !

billsacks:
@jedwards4b I think the problem here is that we can't call `case_setup` on a fake case object. The solution that would make sense to me would be similar to what we did a few months ago for `create_clone`: create a pass-through `case_setup` method in the `Case` class. Then have `case_fake.py` also implement `case_setup`, but there it will simply be implemented as `pass`. Then we'd change calls that look like `case_setup(case, ...)` to `case.case_setup(...)`. I'm not sure if I'm explaining that clearly. I can take a stab at implementing it if you'd like me to.

billsacks:
Looks good, thanks.

jedwards4b:
This change causes me to once again call into question the value of this style of test.   The test failure did nothing to elucidate any new problem in the code and caused me to inject a fake pass through function in the case class which does nothing to further code development.  All tests are now passing.

rljacob:
What style of test do you mean?

jedwards4b:
@rjacob test_unittests (__main__.A_RunUnitTests)

jedwards4b:
Maybe we should consider incorporating some methods like case_clone, case_setup etc into the case class with something like http://www.qtrac.eu/pyclassmulti.html ?   This would remove the need for these wrappers in the case class.

billsacks:
@jedwards4b I'll admit that, whenever you have run into a problem like this, I also question whether I made the right choice in how I set up these unit tests. I keep reaching the tentative conclusion that this is at least a good a choice as the alternative, but I'm not that confident in my own assessment. The alternative (I think) would be to use a real case for each of the unit tests. In addition to being slower (how much slower, I'm not sure, but I suspect significantly slower), I suspect that would come with its own maintenance problems. But I don't have a lot of intuition here.

I'll also add, though, that I personally feel that having the case object control the case_setup behavior is good design: I feel like, as a general rule, a given object should control its own behavior. I have found it unintuitive that case_setup lives entirely outside of the Case class. There's good reason for this, to prevent Case from growing too large, but I personally prefer the design where users of the case class call `case_setup` by calling a method on the case object, rather than by calling some totally separate function that lives outside of Case.

billsacks:
@jedwards4b If I understand http://www.qtrac.eu/pyclassmulti.html correctly, it seems to be advocating for using inheritance for a relationship that I would more intuitively view as composition.

I think there are really two related questions: (1) What is the "best" design for the Case class and various things that operate on the case; (2) Should we use a fake Case class for some unit tests? The following is largely an elaboration of my previous comment.

(1) What is the "best" design for the Case class and various things that operate on the case? I feel that it would be easier for code that uses case objects if there were a consistent rule that, when you're doing something to a case, you do so via method calls on the case object. We currently aren't consistent in this respect (`case_setup` being a case in point), which makes it harder to know where to look when you want to figure out how to do something. I acknowledge that this will grow Case's already-large interface even larger, but I'd argue that this is better than the current state of affairs, which makes it _appear_ like case_setup isn't really part of the interface of case, when in fact (for all intents and purposes) it is. 

I feel like a pretty common object-oriented design technique to deal with large classes like this is to then delegate particular behaviors to other classes. So, from the perspective of a user of a case object, they would only need to know about the interface of Case. But to keep the Case implementation manageable, it would delegate some of its implementation to other classes via pass-through methods. Sure, pass-through methods have a bit of overhead, but I personally see them as providing more benefits than costs in situations like this.

From this perspective, the need to introduce a `case_setup` pass-through method for the sake of the unit tests indicated an issue with the original design, and introducing this method actually improves the design - not just the testability.

(2) Should we use a fake Case class for some unit tests? If, for example, others don't agree with my assessments in (1), then it would be more challenging to use a fake Case class for unit tests. Is that a problem? This could be evaluated based on (a) the performance implications for testing, (b) the extra effort it takes to write unit tests using a real Case object, and (c) the extra effort it takes to maintain unit tests using a real Case object, compared to the effort needed to maintain the fake Case class.

PR #2382, opened 2018-03-20
make case tools part of the Case object

jedwards4b:
This makes the create_clone and case_setup subroutines part of the case class without
requiring that they reside in the same source file.   

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -68,6 +68,14 @@ class Case(object):
     """
     from CIME.case_setup import case_setup
     from CIME.case_clone import create_clone
+    from CIME.case_test  import case_test
+    from CIME.case_submit import check_DA_settings, check_case, submit
+    from CIME.case_st_archive import case_st_archive, restore_from_archive, \
+        archive_last_restarts
+    from CIME.case_run import case_run
+    from CIME.case_cmpgen_namelists import case_cmpgen_namelists
+    from CIME.check_lockedfiles import check_lockedfile, check_lockedfiles, check_pelayouts_require_rebuild
+    from CIME.preview_namelists import create_dirs, create_namelists
 
     def __init__(self, case_root=None, read_only=True):
 
--- a/scripts/lib/CIME/case.py
+++ b/scripts/lib/CIME/case.py
@@ -35,8 +35,6 @@ from CIME.XML.env_batch             import EnvBatch
 from CIME.XML.generic_xml           import GenericXML
 from CIME.user_mod_support          import apply_user_mods
 from CIME.aprun import get_aprun_cmd_for_case
-from CIME.case_clone import create_case_clone
-from CIME.case_setup import case_setup
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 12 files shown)


scripts/lib/CIME/case.py
```diff
@@ -65,7 +63,12 @@ class Case(object):
     the case object creates and manipulates the Case env classes
     by reading and interpreting the CIME config classes.
 
+    This class extends across multiple files, class members external to this file
+    are listed in the following imports
     """
+    from CIME.case_setup import case_setup
```
jedwards4b: By moving these imports into the class and defining them with self instead of case we incorporate them into the class.

jedwards4b:
I think that this is a place where violating the standard makes sense.  It's more clear that these are closely tied to the case.  I made a few more changes and am retesting now.

jedwards4b:
No - it has to be self - that's what makes it part of the class and I have carefully documented each of these files to make that clear.

jgfouca:
"self" is not semantically any different than any other name in python, it's just convention to use it as the name for the object in class methods. In this case I think it makes more sense to leave it called case. @billsacks , thoughts?

jedwards4b:
The point is that these are not "outside of the class"

jgfouca:
@jedwards4b , that's true. They just appear to be at first glance.

jgfouca:
I guess I'm OK with it as is.

billsacks:
I'm a bit late to comment on this already-merged PR, but looking at the result of "taking it to the limit", I find some aspects of this to be confusing:

1. At least in some instances (I'm looking at `case_st_archive` as an example), there are some functions in the module that are now part of the Case class, but some (both public and private) that are outside the Case class. There are some indications of this distinction (comments at the top of the file, use of `self` as the first argument), but I'm finding this aspect to be confusing. I'd find it easier to understand this if, for a given module, either (a) all functions are part of the Case class, and the module was written in an object-oriented way - i.e., the same way you'd write a function that's actually in case.py (with calls to `self.foo(...)` rather than calls to `foo(self, ...)`) or (b) no functions are part of the Case class.

2. At least some of the functions that are now part of the Case class (again, I'm looking at `case_st_archive` as an example) are still called directly from other modules. i.e., they have something like `from CIME.case_st_archive import case_st_archive` and then they call `case_st_archive` directly rather than calling `case.case_st_archive`. I thought the idea of this was that we'd start having everyone call `case.case_st_archive`. (If you are planning to do this, but wanted to do it incrementally, that's okay.)

3. It would help if there were a way to quickly identify which modules are part of the Case implementation. My preference would be if all such modules were in a `case` subdirectory - i.e., `lib/CIME/case`. If that's not desirable, then an alternative would be to make sure that all such modules are prefixed with `case_`. This is mostly true right now, but not entirely so.

I'll be honest: The more I look at this, the more I feel like I'd prefer a more "textbook" object-oriented approach, where Case delegates to other classes via small wrapper methods. My sense is that this would take some additional work up-front (maybe a few hours to write the wrapper methods), but probably wouldn't have _that large_ of a maintenance cost. The benefits in terms of understandability of the code would outweigh these costs in my mind. This is particularly due to point (1) above: The current situation is confusing in this respect. The solution of making (for example) all functions from `case_st_archive.py` be part of the Case class would be less confusing (for me, anyway), but would have the negative implication that all of the currently-private members of `case_st_archive.py` would now become members of the Case class; this would also mean that, whenever a private function is added in `case_st_archive.py`, it needs to be added to the import statement in the Case class. That's all to say: I don't love the current situation, and I also don't love the solution I propose to (1), and I'd prefer the more "textbook" object-oriented approach over either of these.

PR #2389, opened 2018-03-21
Rearrange case (testing is ongoing - do not merge yet)

jedwards4b:
In response to comments on PR #2382 I have created a case_members subdirectory and moved members of the case class into it.  I first attempted to call this directory case but that creates a conflict with case.py and python imports.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


jgfouca:
This looks like a nice change. My only regret is that we can no longer have free functions or utility functions in case_members files.

jedwards4b:
I just figured out how to use __init__.py to move the entire case object to the case subdirectory.

jedwards4b:
All testing completed - the T_TestRunRestart is failing, but it also fails on master so I will open an issue and deal with it separately.

jedwards4b:
If there are no more concerns this one is ready to merge @jgfouca

jedwards4b:
I have noticed that if you update to this PR and then try to back up to a previous tag - it will leave the lib/CIME/case directory in place and cause everything to fail - just manually remove that directory to solve the problem.

PR #2397, opened 2018-03-22
Update MPI message tags in coupler-component exchanges

amametjanov:
* Increase msgtag values to reduce likelihood of msgtag collisions
* Add a more informative error message in shr_flux_mod.F90

Test suite: high-res coupled runs on Theta and Cori-KNL
Test baseline: None
Test namelist changes: None
Test status: [bit for bit]

Fixes ESMCI/cime/issues/2363
Addresses ESMCI/cime/issues/2340

User interface changes?: None

Code review:

### DIFF
--- a/src/drivers/mct/main/component_mod.F90
+++ b/src/drivers/mct/main/component_mod.F90
@@ -402,7 +402,7 @@ contains
                    call shr_sys_flush(logunit)
                 end if
                 call seq_mctext_gGridInit(comp(1))
-                call seq_map_map_exchange(comp(1), flow='c2x', dom_flag=.true., msgtag=comp(1)%cplcompid*100+1*10+1)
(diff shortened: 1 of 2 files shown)


amametjanov:
No, I just tried it out on Theta and that error never re-occurred. Previous message tag values ranged from 600 to 3400 (600 is a relatively small value). This change lifts message tags to 60,000-340,000 range: lowest `cplcompid` is 6 for `CPLATM`, highest is 34 for `CPLESP`. Since it would be easy to mess up message tags just because of a pelayout, it might be better to use a fixed range? Git history shows that the previous tags have worked for a long time.

rljacob:
A message is made unique by a combo of its tag AND communicator.  Are these tag mismatches happening the same communicator?

PR #2405, opened 2018-03-23
Do not make unnecessary env_batch jobs or batch job scripts

### LINKED ISSUE #2393: Clean up env_batch.xml file
There's no reason to have a case.run block in a case that's a test case and vice versa.

jgfouca:
Do not make unnecessary env_batch jobs or batch job scripts, meaning test cases don't make case.run scripts/xmldata and vice versa.

This is a significant change because, in many places throughout CIME, case.run was considered to be the default job that we could rely on. This was very dangerous since it meant changes to case.test job configuration could potentially have been ignored/lost.

This PR establishes the concept of a "primary" job that will either be case.test or case.run depending on if the case is a test case.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2393 

User interface changes?: Sort-of?

Code review: @jedwards4b


scripts/lib/CIME/XML/env_batch.py
```diff
@@ -65,7 +65,7 @@ def set_value(self, item, value, subgroup=None, ignore_type=False):
         return val
 
     # pylint: disable=arguments-differ
-    def get_value(self, item, attribute=None, resolved=True, subgroup="case.run"):
+    def get_value(self, item, attribute=None, resolved=True, subgroup="PRIMARY"):
```
jgfouca: This is a little hacky. No case object is available to retrieve the primary job id.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -740,15 +756,11 @@ def compare_xml(self, other):
...
-    def make_all_batch_files(self, case, test_mode=False):
-        testcase = case.get_value("TESTCASE")
+    def make_all_batch_files(self, case):
+        is_test = case.get_value("TEST")
         machdir  = case.get_value("MACHDIR")
-        logger.info("Creating batch script case.run")
+        logger.info("Creating batch scripts")
         for job in self.get_jobs():
             input_batch_script = os.path.join(machdir,self.get_value('template', subgroup=job))
-            if job == "case.test" and testcase is not None and not test_mode:
```
jgfouca: Why were we not regenerating case.test in test_mode?

jgfouca:
Tests still running. Review only for now.

scripts/lib/CIME/XML/env_batch.py
jedwards4b: Because in some cases, and these may all be gone now that we do two case tests, we needed to change the run script mid test but keep the test script the same.

jgfouca:
@jedwards4b , I think this is ready.

PR #2406, opened 2018-03-27
Jgfouca/branch for acme split 2018 03 27

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b


scripts/lib/CIME/case/case_st_archive.py
```diff
@@ -384,7 +384,7 @@ def _archive_restarts_date_comp(case, archive, archive_entry,
         for i in range(ninst):
             restfiles = ""
             if compname.find("mpas") == 0:
-                pattern = compname + suffix + '_'.join(datename_str.rsplit('-', 1))
+                pattern = compname + r'\.' + suffix + r'\.' + '_'.join(datename_str.rsplit('-', 1))
```
jgfouca: Important fix that was needed on the E3SM side.

jgfouca:
Tests passed. I don't think there is much here of interest to @jedwards4b , so I think it's OK to merge and let him review at a later date.

PR #2412, opened 2018-03-30
allow envvars to be unresolved

### LINKED ISSUE #2411: Undefined env var error
env_mach_specific.xml may include environment variables which are only defined on compute nodes and so should not fail due to undefined environment variables on the login nodes.   As an example the machine hobart defines 
```
 <arg name="machine_file">--machinefile $ENV{PBS_NODEFILE}</arg>
```
which now results in a failure at case setup time:
CaseStatus:ERROR: Undefined env var 'PBS_NODEFILE'

jedwards4b:
Some env vars in env_mach_specific.xml may only be defined on compute nodes, this change allows that to happen while resolving all cime xml vars. 

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #2411 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -203,7 +203,7 @@ class EnvBatch(EnvBase):
         overrides["mpirun"] = case.get_mpirun_cmd(job=job)
         output_text = transform_vars(open(input_template,"r").read(), case=case, subgroup=job, overrides=overrides)
         output_name = get_batch_script_for_job(job)
-
+        logger.info("Creating file {}".format(output_name))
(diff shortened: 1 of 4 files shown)


jgfouca:
@jedwards4b , I think we want get_mpirun_cmd to fail if env vars are unresolved when it gets called from case_run.

The env_batch.make_batch_script and preview_run contexts are the ones where it's OK for env vars to remain unresolved. But this is not OK in the case_run context. Can you add a flag to get_mpirun_cmd too?

jedwards4b:
This is what you had in mind?

PR #2418, opened 2018-04-02
Remove the run_e3sm testing scripts from the CIME repo

mfdeakin-sandia:
This removes the run_e3sm test script from the CIME repository, as it has been put in a more appropriate repository.

Test suite: N/A


jgfouca:
@mfdeakin-sandia , ignore the failed check. That was from another PR.

PR #2419, opened 2018-04-03
Restore shell batch capability

jgfouca:
A previous PR to simplify env_batch.xml broke this capability. 

Test suite: by-hand, code-checker
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b


jedwards4b:
Should a test be added to assure that we don't break it again in the future?

jgfouca:
@jedwards4b , i would say no test is needed since this is e3sm-only and not an officially supported feature.

PR #2433, opened 2018-04-09
E3SM split

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b


scripts/lib/CIME/SystemTests/eri.py
```diff
@@ -145,6 +145,7 @@ def run_phase(self):
 
         _helper(dout_sr1, refdate_2, refsec_2, rundir2)
 
+        self._skip_pnl = False
```
jgfouca: Wondering why this is here.

scripts/lib/CIME/case/case.py
```diff
@@ -1259,7 +1259,7 @@ def get_mpirun_cmd(self, job=None, allow_unresolved_envvars=True):
         if executable is not None and "aprun" in executable and not "theta" in self.get_value("MACH"):
             aprun_args, num_nodes = get_aprun_cmd_for_case(self, run_exe)[0:2]
             expect( (num_nodes + self.spare_nodes) == self.num_nodes, "Not using optimized num nodes")
-            return executable + aprun_args + " " + run_misc_suffix
+            return self.get_resolved_value(executable + aprun_args + " " + run_misc_suffix, allow_unresolved_envvars=allow_unresolved_envvars)
```
jgfouca: Important bug fix made on the E3SM side.

scripts/lib/CIME/test_scheduler.py
```diff
@@ -430,11 +430,7 @@ def _create_newcase_phase(self, test):
 
         if test_mods is not None:
             files = Files()
-            if self._cime_model == "e3sm":
-                component = "allactive"
-                modspath = test_mods
-            else:
-                (component, modspath) = test_mods.split('/',1)
+            (component, modspath) = test_mods.split('/',1)
```
jgfouca: e3sm no longer uses a special test mod location.

jgfouca:
@jedwards4b , annotations complete. Only one change in eri has me scratching my head.

scripts/lib/CIME/SystemTests/eri.py
jedwards4b: Ah - because something seems to set skip_pnl back to true after line 106 and again before line 199.   I gave up trying to figure out what was doing it and just always make sure it's false when calling run_indv

jgfouca: Should I remove this change?

PR #2438, opened 2018-04-09
Put e3sm test definitions under e3sm config.

jgfouca:
Leave the functions in place under a better file name.

Remove stale update_e3sm_tests tool.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: File name changes, remove old tool

Code review: @rljacob @jedwards4b


jgfouca:
Things are still slightly awkward, with cesm code invoking get_e3sm_tests in one spot and a couple cime test suites defined in config/e3sm/e3sm_tests.py, but I don't think it's any worse than before.

jgfouca:
@rljacob , config/e3sm/e3sm_tests.py shouldn't be executable, it isn't in my repo.

rljacob:
Yes I see that now.

scripts/lib/get_e3sm_tests.py
```diff
@@ -1,193 +1,10 @@
...
-
-    "fates" : (None, None,
-                         ("ERS_Ld9.1x1_brazil.ICLM45ED",
-                          "ERS_D_Ld9.1x1_brazil.ICLM45ED",
-                          "SMS_D_Lm6.1x1_brazil.ICLM45ED")
-                         ),
-}
+import six, sys, os
+
+sys.path.append(os.path.join(get_cime_root(), "config", "e3sm"))
```
rljacob: In this line, can "e3sm" be replaced with $model?

rljacob:
We want scripts/lib to be generic functionality any model system could use.   So I'd like to change "get_e3sm_tests" to just "get_tests" or "get_testsuites".   And the config file should just be "test_suites.py" or similar.  Basically take "e3sm" out of the name.

scripts/lib/get_e3sm_tests.py
jgfouca: No, the cime test suites are needed for both models and cesm does not define their tests this way.

jgfouca:
@rljacob , the e3sm-specificness of get_e3sm_tests is why it lives in scripts/lib and not scripts/lib/CIME.

jgfouca:
Note the e3sm-specific lib for managing the subtree are its siblings.

rljacob:
Its not "e3sm" specific.  Its "define test suites with a python dict" specific.  e3sm happens to be the model that uses that method.  A 3rd model may want to use it as well.  We should name things by what they do, not what model uses them.

jgfouca:
@rljacob , ok I guess that makes sense.

rljacob:
Cool.   config/e3sm/e3sm_tests.py  should be renamed without the "e3sm".  Its already in the config/e3sm directory so we know its defining python-based test suites for e3sm.

jgfouca:
@rljacob , ok, see if you like it better now

jgfouca:
@jedwards4b , you may want to take a quick look at this even though this shouldn't impact cesm at all.

doc/source/users_guide/testing.rst
```diff
@@ -286,7 +286,7 @@ manage_testlists
 Adding tests
 =============
 
-Open the update_e3sm_tests.py file, you'll see a python dict at the top
+Open the e3sm_tests.py file, you'll see a python dict at the top
```
rljacob: Change to "tests.py".

rljacob:
Looks great!

PR #2439, opened 2018-04-10
Addgridftp

jedwards4b:
Adds inputdata servers for grid ftp, ftp, and wget.   

Test suite: lots of testing by hand, scripts_regression_tests.py 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/Servers/__init__.py
+++ b/scripts/lib/CIME/Servers/__init__.py
@@ -1,3 +1,4 @@
+#pylint: disable=import-error
 from distutils.spawn import find_executable
 has_gftp = find_executable("globus-url-copy")
 has_svn = find_executable("svn")


fischer-ncar:
Is there a script that keeps the different sources of inputdata synced?  What are the chances that gridftp might end up having a file that's not on the subversion or ftp server?  This could causing a missing file for someone who isn't using gridftp.

jedwards4b:
@fischer-ncar if a file isn't found using one protocol or server it will try the next one and continue that way until all servers have been exhausted.   Not all inputdata sources have to have all of the data - so for example a collaborator (like a university) could stand up their own server with a small set of cases that they support independent of our servers.   There is an assumption that a file of the same name is the same file on different servers - we can add md5 sum checks in a future upgrade that will help maintain consistancy.   
If a client doesn't have the proper software to connect using a given protocol, the next protocol will be attempted.

Issue #2377, opened 2018-03-19
Rename components/cism/glimmer-cism to source_cism


billsacks:
We want to rename the components/cism/glimmer-cism external to source_cism, because we have stopped referred to this model as glimmer. Because of some special handling that's needed for this sub-component, there are a few places in cime where this needs to be changed: `scripts/Tools/Makefile` and `scripts/lib/CIME/case.py`.

I'll make these changes, which need to be coordinated with a CISM update in CESM.

Issue #2376, opened 2018-03-19
Empty SourceMods directories are not protected from scrubbers


gold2718:
On Hobart, the overnight file scrubber will delete empty SourceMods directories for cases on /scratch/cluster. Any missing directory (e.g., `<case>/SourceMods/src.cam`) will prevent the case from building or running.
Creating an empty directory in the place of a missing directory can be problematic as the system has no way of knowing if the directories were previously empty or were scrubbed.
Suggested solution: Add a small README file in each SourceMods sub-directory. A current example is `<case>/SourceMods/src.cism/glimmer-cism`.

PR #2462, opened 2018-04-12
Support renamed glimmer-cism directory; add README files to SourceMods directories

billsacks:
This PR includes two changes that are conceptually unrelated, but I have
combined them because they touch the same code:

(1) Support CISM's rename of the glimmer-cism directory to
    source_cism. Because the source files in this subdirectory are built
    into a separate library, a few updates are needed in cime, too. Note
    that, for CESM, this change must be coordinated with
    cism2_1_51 (for e3sm, there are no dependencies).

(2) Add README files in each subdirectory of SourceMods. The main point
    of this is to prevent SourceMods directories from being scrubbed due
    to being empty, which happens on hobart and possibly other systems
    (and then this breaks the build).

Test suite: scripts_regression_tests on hobart; also, aux_glc test suite
  for cism2_1_51 (with these cime changes rebased onto cime5.4.0-alpha.24)
Test baseline: For aux_glc test suite: cism2_1_50
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#2377
Fixes ESMCI/cime#2376

User interface changes?: none

Code review:

### DIFF
--- a/scripts/Tools/Makefile
+++ b/scripts/Tools/Makefile
@@ -682,7 +682,7 @@ CMAKE_ENV_VARS += CC=$(CC) \
 .PHONY: $(GLC_DIR)/Makefile
 $(GLC_DIR)/Makefile:
 	cd $(GLC_DIR); \
-	$(CMAKE_ENV_VARS) cmake $(CMAKE_OPTS) $(GLCROOT)/glimmer-cism
+	$(CMAKE_ENV_VARS) cmake $(CMAKE_OPTS) $(GLCROOT)/source_cism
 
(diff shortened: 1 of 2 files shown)


billsacks:
@gold2718 in case you're interested

billsacks:
@fischer-ncar this needs to be coordinated with `cism2_1_51`, currently listed in alpha10d (but it can be moved to a later alpha tag if needed)

fischer-ncar:
@billsacks you can leave cism2_1_51 in alpha10d.

PR #2466, opened 2018-04-12
Refactor BUILD_THREADED case value.

jgfouca:
The old BUILD_THREADED value was a bit overloaded since it represented
both an override for the user and as indicator to external build/buildnml
scripts that CIME is being built threaded.

This refactor splits off the override capability to a separate setting.

Test suite: code-checker, scripts_regression_tests K_TestCimeCase
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: New FORCE_BUILD_THREADED

Code review: @jedwards4b


jgfouca:
@jedwards4b , this is to address this ticket on E3SM:

```
 ./create_newcase --case QQQ -compset A_WCYCL1950S_CMIP6_LR -res ne30_oECv3_ICG -mach titan -compiler pgi -project cli115
 cd QQQ
 ./case.setup
./xmlquery BUILD_THREADED
FALSE
```

```
 ./xmlchange -file env_mach_pes.xml -id NTHRDS -val 2
 ./case.setup --reset
./xmlquery BUILD_THREADED
TRUE
```

```
 ./xmlchange -file env_mach_pes.xml -id NTHRDS -val 1
 ./case.setup --reset
./xmlquery BUILD_THREADED
TRUE
```

^ Demonstrates the awkwardness of using BUILD_THREADED as an override and a matter of record.

scripts/lib/CIME/case/case.py
```diff
@@ -1294,7 +1294,7 @@ def get_build_threaded(self):
         """
         Returns True if current settings require a threaded build/run.
         """
-        force_threaded = self.get_value("BUILD_THREADED")
+        force_threaded = self.get_value("FORCE_BUILD_THREADED")
```
jedwards4b: Should be SMP_BUILD I think.

scripts/lib/CIME/case/case_setup.py
```diff
@@ -131,7 +131,8 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False):
                     else:
                         expect(False, "NINST_{} value {:d} greater than NTASKS_{} {:d}".format(comp, ninst, comp, ntasks))
                 case.set_value("NTASKS_PER_INST_{}".format(comp), int(ntasks / ninst))
-        if nthrds > 1:
+
+        if nthrds > 1 or case.get_value("FORCE_BUILD_THREADED"):
```
jedwards4b: SMP_BUILD

scripts/tests/scripts_regression_tests.py
```diff
@@ -1608,7 +1608,7 @@ def test_cime_case_build_threaded_1(self):
             build_threaded = case.get_build_threaded()
             self.assertFalse(build_threaded)
 
-            case.set_value("BUILD_THREADED", True)
+            case.set_value("FORCE_BUILD_THREADED", True)
```
jedwards4b: SMP_BUILD

jgfouca:
@jedwards4b  thanks, I was not aware of SMP_BUILD.

jgfouca:
@jedwards4b , between BUILD_THREADED and SMP_BUILD, which one is the override and which one is the matter of record?

amametjanov:
There is SMP_PRESENT case-var to indicate if the build was force-threaded or if it had NTHRDS greater than 1 for any component.

jedwards4b:
As @amametjanov it's a bit of a mess right now.   
In the Makefile there are two variables BUILD_THREADED and SMP and it has this logic:
```
ifeq ($(strip $(SMP)),TRUE)
   THREADDIR = threads
   compile_threaded = true
else
   ifeq ($(strip $(BUILD_THREADED)),TRUE)
      THREADDIR = threads
      compile_threaded = true
   else
      THREADDIR = nothreads
      compile_threaded = false
   endif
endif
```
So SMP is the override and BUILD_THREADED the record.

jedwards4b:
Makefile SMP is set based on SMP_BUILD in build.py

jedwards4b:
SMP_PRESENT seems to be a third use of the same functionality and I think it can be removed.

rljacob:
What are the actions you are trying to control and what are the states your are trying to record?

jedwards4b:
Two conditions: nthreads > 1 => BUILD_THREADED=True
nthreads == 1 but force threaded flag => SMP_BUILD=True

rljacob:
The build system needs just one command to make it build with threading.  But there are 2 different states:  built threaded AND threads >1, built threaded AND threads =1.

jgfouca:
@jedwards4b , it appears from the XML descriptions that BUILD_THREADED (now FORCE_BUILD_THREADED) was intended to be the override. The python code was consistent with this notion, but in buildnml and buildlib scripts is where is gets murky, where some of them were using BUILD_THREADED as a matter of record.

The description of SMP_BUILD is:
"Status: smp status of previous build, coded string." , so it seems to be something else.

jgfouca:
@jedwards4b , take another look and see if you like it now.

scripts/Tools/Makefile
```diff
@@ -32,9 +32,9 @@ complib: $(COMPLIB)  Depends
 
 CIME_MODEL ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) MODEL --value)
 EXEROOT ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) EXEROOT --value)
-BUILD_THREADED ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) BUILD_THREADED --value)
+BUILD_THREADED ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) SMP_PRESENT --value)
 SMP ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) SMP --value)
-LIBROOT ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) BUILD_THREADED --value)
+LIBROOT ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) SMP_PRESENT --value)
```
jgfouca: I found it very strange the LIBROOT was being set to BUILD_THREADED?

scripts/lib/CIME/case/case_setup.py
```diff
@@ -131,8 +129,6 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False):
                     else:
                         expect(False, "NINST_{} value {:d} greater than NTASKS_{} {:d}".format(comp, ninst, comp, ntasks))
                 case.set_value("NTASKS_PER_INST_{}".format(comp), int(ntasks / ninst))
-        if nthrds > 1:
```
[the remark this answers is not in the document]
jgfouca: because SMP_PRESENT is being set below based on case.get_build_threaded

config/config_tests.xml
```diff
@@ -247,7 +247,7 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>11</STOP_N>
     <DOUT_S>FALSE</DOUT_S>
-    <BUILD_THREADED>TRUE</BUILD_THREADED>
+    <FORCE_BUILD_THREADED>TRUE</FORCE_BUILD_THREADED>
```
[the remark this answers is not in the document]
jgfouca: @jedwards4b , not new , just a rename of BUILD_THREADED.

rljacob:
How is this being tested?  Not sure cime_developer really tests threading that well.

scripts/Tools/Makefile
```diff
@@ -32,9 +32,9 @@ complib: $(COMPLIB)  Depends
 
 CIME_MODEL ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) MODEL --value)
 EXEROOT ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) EXEROOT --value)
-BUILD_THREADED ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) BUILD_THREADED --value)
+BUILD_THREADED ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) SMP_PRESENT --value)
```
jgfouca: Wondering why we need BUILD_THREADED and SMP? Why not just BUILD_THREADED?

jedwards4b:
Ugh - this is a frigging mess, thanks for taking it on.   But I agree with Rob - scripts_regression_tests.py isn't enough for this change.   Whats the difference between SMP_BUILD and SMP_VALUE?

jgfouca:
@jedwards4b , both SMP_BUILD and SMP_VALUE don't appear to be used outside of build.py

jgfouca:
It looks like SMP_BUILD and SMP_VALUE encode the threadedness of each component into a string. BUILD is the most recent, VALUE is the current. There's a check in build.py to make sure they are the same.

jgfouca:
@jedwards4b , I think it's OK now. FORCE_BUILD_THREADED is now clearly the override, SMP_PRESENT is the matter of record, and SMP_BUILD and SMP_CURRENT are just used internally by build.py.

rljacob:
Does "SMP" = "THREADED"?  Maybe we should be more specific since we're talking about building with OpenMP, right?.   FORCE_BUILD_OMP,  OMP_PRESENT.

jgfouca:
@rljacob , yes, SMP_PRESENT = threaded

rljacob:
If they're the same then don't use two different words throughout.  FORCE_BUILD_SMP,  SMP_PRESENT, etc.   or FORCE_BUILD_THREADED,  THREAD_PRESENT, etc.

jgfouca:
@rljacob , that seems reasonable.

PR #2475, opened 2018-04-16
Transition CIME to using a custom util for copying files.

jgfouca:
The file-copy operation is very common in CIME and we were using
a smattering of shutil.copyfile, shutil.copy, and shutil.copy2.
This PR replaces these with safe_copy, a flexble and fault-tolerant
copying function. This should make it much easier for users to share
case directories and other common directories.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b


jedwards4b:
Nice cleanup - thanks.

PR #2491, opened 2018-04-20
Make st_archive handle DART's restart history files.

kdraeder:
case_st_archive.py:
   Enhance the 'restfiles' pattern in _archive_restarts_date_comp
   to find files which have a 'string' following the 'suffix'
   ($case.$comp.rh.$string.$date.nc).

   Improve the pattern used to find 'files' to be more selective
   (search for '\.' instead of '.' between $case and $comp) and to not
   search for '.*' at the end.

config/cesm/config_archive.xml:
   Add test file names into the dart section,
   including rh files.

Test suite: case.st_archive --test
Test baseline: 
Test namelist changes: none
Test status: bit for bit

Fixes  #2470

User interface changes?: none

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/config_archive.xml
+++ b/config/cesm/config_archive.xml
@@ -156,6 +156,7 @@
 
     <comp_archive_spec compclass="esp" compname="dart">
       <rest_file_extension>r</rest_file_extension>
+      <rest_file_extension>rh\d?</rest_file_extension>
       <hist_file_extension>[ei]</hist_file_extension>
       <rest_history_varname>restart_hist</rest_history_varname>
       <rpointer>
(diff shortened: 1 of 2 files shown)


mfdeakin-sandia:
Should the config_archive changes also be made to the E3SM version? I don't know if we have the same version of DART

kdraeder:
@mfdeakin-sandia 
Thanks for bringing that up.  I'm not aware of anyone using DART with e3sm.
So it seems like it wouldn't hurt to put the same config_archive.xml changes
into that version if we want to keep it all consistent.  Then it will be ready
if/when someone wants it.

PR #2499, opened 2018-04-24
Match entire string for XML attribute selectors for env_mach_specific

### LINKED ISSUE #2472: The string in `--compiler=` is being matched unexpectedly
We have compilers such as: "intel,intel18,gnu,gnu7".
And currently, using `--compiler=gnu` does what is expected, but `--compiler=gnu7` will match BOTH "gnu" and "gnu7", which means the module command for both of these will be executed.  Is this a bug or feature?

For example, when using `--compiler=intel8` on edison, I see the following module commands (as shown in the csh file).  Note that the intel17 module commands are executed as well as the intel18 commands.  This still works (intel18 will be used), but only because of the order in config_machines.xml.

```
source /opt/modules/default/init/csh
module rm PrgEnv-intel PrgEnv-cray PrgEnv-gnu intel cce gcc cray-parallel-netcdf cray-parallel-hdf5 pmi cray-libsci cray-mpich2 cray-mpich cray-netcdf cray-hdf5 cray-netcdf-hdf5parallel craype-san
dybridge craype-ivybridge craype papi cray-petsc esmf craype
module load craype/2.5.12.3 craype-ivybridge
module rm pmi
module load pmi/5.0.12
module rm cray-mpich
module load cray-mpich/7.6.0 PrgEnv-intel/6.0.4
module rm intel
module load intel/17.0.2.174
module rm cray-libsci
module load PrgEnv-intel/6.0.4
module rm intel
module load intel/18.0.2.199
```

jgfouca:
If a machine has compilers "intel" and "intel18", we don't want to
match "intel" selectors when the user asks for "intel18" compiler.
If the user truly wants to express the concept "match any intel compiler",
they can use "intel.*" as the selector.

Looking over config_machines.xml for both e3sm and cesm, I don't think any machine was intentionally making-use of partial-string matching, except for maybe yellowstone.

Test suite: scripts_regression_tests, by-hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2472 

User interface changes?: Yes, changes to matching behavior in env_mach_specific

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/XML/env_mach_specific.py
+++ b/scripts/lib/CIME/XML/env_mach_specific.py
@@ -245,12 +245,12 @@ class EnvMachSpecific(EnvBase):
 
     def _match(self, my_value, xml_value):
         if xml_value.startswith("!"):
-            result = re.match(xml_value[1:],str(my_value)) is None
+            result = re.match(xml_value[1:] + "$",str(my_value)) is None
         elif isinstance(my_value, bool):
             if my_value: result = xml_value == "TRUE"
             else: result = xml_value == "FALSE"
         else:
-            result = re.match(xml_value,str(my_value)) is not None
+            result = re.match(xml_value + "$",str(my_value)) is not None
 
         logger.debug("(env_mach_specific) _match {} {} {}".format(my_value, xml_value, result))
         return result


rljacob:
Does "match any" pick the first one found and just use that?  The problem in #2472 was that the union of "intel" AND "intel18" was being applied.  I don't think a user ever wants that.

jgfouca:
@rljacob , the core of the problem on edison:

```
    <modules compiler="intel">
      <command name="load">PrgEnv-intel</command>
      <command name="rm">intel</command>
      <command name="load">intel/17.0.2.174</command>
      <command name="rm">cray-libsci</command>
    </modules>

    <modules compiler="intel18">
      <command name="load">PrgEnv-intel</command>
      <command name="rm">intel</command>
      <command name="load">intel/18.0.1.163</command>
      <command name="rm">cray-libsci</command>
    </modules>
```

Let's assume compiler=intel18. What CIME was doing is, for the first block, the "intel" selector was being turned into a regex and the matched against "intel18". This match succeeded since, in python, regex.match doesn't have to match the entire string, it only needs to be valid for some prefix of the string being matched. The second block was also being matched, since "intel18" obviously matches itself.

rljacob:
Okay.  That doesn't help me understand what "match any intel compiler" will wind up doing.  The intent of this system is that one and only one compiler setting be applied.

PR #2511, opened 2018-04-30
Enables submitting resubmits early and for prerequisites to fail when the user requests it

### LINKED ISSUE #2510: Enable submitting resubmits before the model finishes
Also enable the dependent resubmit to run even if the model crashes

### LINKED ISSUE #2547: Running case.st_archive with RESUBMIT > 0 submits case.run jobs
One of our users complained about this, and it seems very unintuitive from a UI perspective for people who run case.st_archive from the login node.
My thought on a solution to this is to add an argument to case.st_archive which enables resubmitting, and have case.submit set that option for the archiving jobs.

mfdeakin-sandia:
This implements two features:
1) Resubmitting regardless of the success of the preceding job(s). Note that this can be dangerous and waste significant computer time if done with restarts which are too far apart or if systematic errors cause the model to crash. This feature is used by running `case.submit` with the `--prereq-allow-failure` option.
2) Submitting the resubmitted jobs all at once, potentially significantly reducing the time spent in the queue on some HPC systems. This feature is used by running `case.submit` with the `--resubmit-immediate` option.

These options can (and are intended to) be used together; doing so will submit `$RESUBMIT` jobs which have a dependency on the job submitted immediately before, but does not require it to have completed successfully, only to have at least run.

Test status: ok

Fixes #2510, Fixes #2547

User interface changes?: Adds the `--prereq-allow-failure` and the `--resubmit-immediate` option to case.submit.

### DIFF
--- a/scripts/Tools/case.submit
+++ b/scripts/Tools/case.submit
@@ -39,8 +39,14 @@ OR
                         help="Do not submit jobs to batch system, run locally.")
 
     parser.add_argument("--prereq",
-                        help="Specify a prerequiset job id, this job will not start until the "
+                        help="Specify a prerequisite job id, this job will not start until the "
                         "job with this id is completed (batch mode only)")
+
+    parser.add_argument("--prereq-allow-failure", action="store_true",
+                        help="Allows starting the run even if the prerequisite fails."
+                        "This also allows resubmits to run if the original failed and the "
+                        "resubmit was submitted to the queue with the orginal as a dependency, "
+                        "as in the case of --resubmit-immediate")
 
     parser.add_argument("--resubmit", action="store_true",
                         help="Used with tests only, to continue rather than restart a test. ")
@@ -57,14 +63,14 @@ OR
 
     CIME.utils.resolve_mail_type_args(args)
 
-    return args.caseroot, args.job, args.no_batch, args.prereq, \
+    return args.caseroot, args.job, args.no_batch, args.prereq, args.prereq_allow_failure, \
         args.resubmit, args.skip_preview_namelist, args.mail_user, args.mail_type, \
         args.batch_args
 
 ###############################################################################
 def _main_func(description):
 ###############################################################################
-    caseroot, job, no_batch, prereq, resubmit, skip_pnl, \
+    caseroot, job, no_batch, prereq, allow_fail, resubmit, skip_pnl, \
         mail_user, mail_type, batch_args = parse_command_line(sys.argv, description)
 
     # save these options to a hidden file for use during resubmit
@@ -86,9 +92,9 @@ def _main_func(description):
         os.remove(config_file)
 
     with Case(caseroot, read_only=False) as case:
-        case.submit(job=job, no_batch=no_batch, prereq=prereq, resubmit=resubmit,
-               skip_pnl=skip_pnl, mail_user=mail_user, mail_type=mail_type,
-               batch_args=batch_args)
+        case.submit(job=job, no_batch=no_batch, prereq=prereq, allow_fail=allow_fail,
+                    resubmit=resubmit, skip_pnl=skip_pnl, mail_user=mail_user, mail_type=mail_type,
+                    batch_args=batch_args)
 
 if __name__ == "__main__":
     _main_func(__doc__)
--- a/scripts/Tools/case.submit
+++ b/scripts/Tools/case.submit
@@ -51,6 +51,11 @@ OR
     parser.add_argument("--resubmit", action="store_true",
                         help="Used with tests only, to continue rather than restart a test. ")
 
+    parser.add_argument("--resubmit-immediate", action="store_true",
+                        help="This queues all of the resubmissions immediately after "
+                        "the first job is queued. These rely on the queue system to "
+                        "handle dependencies.")
+
     parser.add_argument("--skip-preview-namelist", action="store_true",
                         help="Skip calling preview-namelist during case.run")
 
@@ -63,14 +68,14 @@ OR
 
     CIME.utils.resolve_mail_type_args(args)
 
-    return args.caseroot, args.job, args.no_batch, args.prereq, args.prereq_allow_failure, \
-        args.resubmit, args.skip_preview_namelist, args.mail_user, args.mail_type, \
-        args.batch_args
+    return (args.caseroot, args.job, args.no_batch, args.prereq, args.prereq_allow_failure,
+            args.resubmit, args.resubmit_immediate, args.skip_preview_namelist, args.mail_user,
+            args.mail_type, args.batch_args)
 
 ###############################################################################
 def _main_func(description):
 ###############################################################################
-    caseroot, job, no_batch, prereq, allow_fail, resubmit, skip_pnl, \
+    caseroot, job, no_batch, prereq, allow_fail, resubmit, resubmit_immediate, skip_pnl, \
         mail_user, mail_type, batch_args = parse_command_line(sys.argv, description)
 
     # save these options to a hidden file for use during resubmit
@@ -93,8 +98,8 @@ def _main_func(description):
 
     with Case(caseroot, read_only=False) as case:
         case.submit(job=job, no_batch=no_batch, prereq=prereq, allow_fail=allow_fail,
-                    resubmit=resubmit, skip_pnl=skip_pnl, mail_user=mail_user, mail_type=mail_type,
-                    batch_args=batch_args)
+                    resubmit=resubmit, resubmit_immediate=resubmit_immediate, skip_pnl=skip_pnl,
+                    mail_user=mail_user, mail_type=mail_type, batch_args=batch_args)
 
 if __name__ == "__main__":
     _main_func(__doc__)
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -507,8 +507,8 @@ class EnvBatch(EnvBase):
             if allow_fail:
                 dep_string = self.get_value("depend_allow_string", subgroup=None)
                 if dep_string is None:
-                    logger.warn("'depend_allow_string' is not defined for this batch system, " +
-                                "falling back to the 'depend_string'")
+                    logger.warning("'depend_allow_string' is not defined for this batch system, " +
+                                   "falling back to the 'depend_string'")
                     dep_string = self.get_value("depend_string", subgroup=None)
             else:
                 dep_string = self.get_value("depend_string", subgroup=None)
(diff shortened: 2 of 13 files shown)


scripts/Tools/case.submit
```diff
@@ -38,9 +38,12 @@ OR
                         help="Do not submit jobs to batch system, run locally.")
 
     parser.add_argument("--prereq",
-                        help="Specify a prerequiset job id, this job will not start until the "
+                        help="Specify a prerequisite job id, this job will not start until the "
                         "job with this id is completed (batch mode only)")
 
+    parser.add_argument("--prereq-allow-failure", action="store_true",
```
jedwards4b: there should be an additional option here --resubmit-in-queue or --resumbit-immediate to indicate that all jobs should be submitted at once with prereqs

scripts/lib/CIME/case/case_run.py
```diff
@@ -250,6 +251,33 @@ def case_run(self, skip_pnl=False):
     # set up the LID
     lid = new_lid()
 
+    resubmit_at_start = False
+
+    if self.get_value("RESUBMIT_AT_START"):
```
jedwards4b: This should use the RESUBMIT arg to queue all the jobs that will need to be run - that is there should be a loop here from 1 to RESUBMIT and the RESUBMIT variable should be decremented for each job submitted.

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -473,6 +473,37 @@
     </desc>
   </entry>
 
+  <entry id="RESUBMIT_AT_START">
```
jedwards4b: If DOUT_S is TRUE than the new job should depend on the completion of case.st_archive, not on case.run.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -495,9 +496,19 @@ def _submit_single_job(self, case, job, dep_jobs=None, no_batch=False,
 
         if dep_jobs is not None and len(dep_jobs) > 0:
             logger.info("dependencies: {}".format(dep_jobs))
-            dep_string = self.get_value("depend_string", subgroup=None)
+            if allow_fail:
+                dep_string = self.get_value("depend_allow_string", subgroup=None)
+                if dep_string is None:
+                    logger.warn("'depend_allow_string' is not defined for this batch system, " +
```
jedwards4b: logger.warn is deprecated - should be logger.warning

scripts/lib/CIME/case/case_run.py
jedwards4b: Yes this should be in case_submit.py.   I'm not sure I understand your comment about CONTINUE_RUN, this change shouldn't have any affect on that variable.

mfdeakin-sandia: I guess my mental model of what happens isn't right?
I thought CONTINUE_RUN was used to set a namelist variable for determining if restarts are to be used; and that namelists are generated in case.run? So setting it in case_submit.py would cause the namelists to be wrong for the first run.
OTOH, the bash script to queue resubmits doesn't have this problem, so I guess I've misunderstood something.

mfdeakin-sandia:
I've implemented this without the xml variables to resubmit from the login node, though I still don't understand why the change to `CONTINUE_RUN` isn't picked up by the namelist generator in `case.run`, so I can't guarantee the initial namelist will have the correct `run_type`.
I still need to add more `depend_allow_string` configurations and write a unit test, though it should be fairly simple for this implementation.

mfdeakin-sandia:
Ah. My job made it through the queue and it's definitely crashing in what I believe to be the restart reading code, which is what I'd expect for an incorrect `run_type` in the initial run.

jedwards4b:
Are the subsequent submits modifying the run_type in the namelist before the first job starts?   It's important here to make sure that the namelist update for RESUBMIT=1 does not modify the namelist in case.submit - that should only happen after the new job starts.   I think that you may need the skip_pnl option in case.submit, but then make sure not to skip_pnl in case.run.

mfdeakin-sandia:
Yes, they do modify `CONTINUE_RUN` before the first run, but it's not clear how else to set it from `case.submit`? The simplest thing I could do would be another xml variable telling `case.run` to set it after its finished the run.
`case.run` always generates namelists for the driver, regardless of `skip_pnl`, so I don't think that would help.

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -485,6 +485,19 @@
     </desc>
   </entry>
 
+  <entry id="COMPLETION_SETS_CONTINUE_RUN">
```
jedwards4b: I don't think we need another variable to do this.   You just need to make sure that the namelists are not updated too early.

mfdeakin-sandia: The other idea I had was to submit another job which would run `xmlchange CONTINUE_RUN=TRUE`, and make the resubmits dependent on that, but that seems worse. Otherwise I don't know of any way of doing this, since case.submit submits all the jobs, but can't set `CONTINUE_RUN` because of the namelist generation in `case.run`, and AFAIK, has no other portable way to tell case.run to set `CONTINUE_RUN` after the namelists are generated.

cameronsmith1: Submitting an extra job that only changes continue_run to true is what I did when I create a script to queue up runs.

jedwards4b:
I don't believe that you need anything extra for CONTINUE_RUN - if you get this set up correctly CONTINUE_RUN will take care of itself.   If you don't then the CONTINUE_RUN is only a symptom of a bigger problem - you can not prepare the namelist for run2 before run1 has started.

mfdeakin-sandia:
I don't understand - how should I ensure that variables which cannot be set before the first preview namelist from case.run is run from case.submit? They're completely asynchronous; any xml change I do after the queue submission can occur at any point during the run script.
In any case, I've implemented this without an xml variable, and instead just another argument to case.run. Is this better?
It's still testing, of course

mfdeakin-sandia:
Are you suggesting I should instead create a postrun script and tell case.run to do that?

jedwards4b:
It is the problem that since it doesn't do it here it has to do it someplace else.

mfdeakin-sandia:
Z_FullSystemTest passes, so this seems to be working; though I still need to add a unit test for the resubmission

scripts/lib/CIME/case/case_run.py
```diff
@@ -283,6 +283,9 @@ def case_run(self, skip_pnl=False):
         self.read_xml()
         _save_logs(self, lid)
 
+    if set_continue_run:
```
jedwards4b: This should be case.get_value("RESUBMIT_SETS_CONTINUE_RUN") and not simply True

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -439,53 +439,118 @@ def submit_jobs(self, case, no_batch=False, job=None, user_prereq=None,
...
+                    deps = dependency.split()
+                else:
+                    deps = []
+                dep_jobs = []
+                if user_prereq is not None:
+                    dep_jobs.append(user_prereq)
+                for dep in deps:
+                    if dep in depid.keys() and depid[dep] is not None:
+                        dep_jobs.append(str(depid[dep]))
+                if prev_job is not None:
```
jedwards4b: I think that this is wrong - prev_job is only an explicit dependency if this is the first job in the current submit, otherwise dependancies are defined in env_batch.xml

mfdeakin-sandia: Isn't it the opposite? ie, prev_job is an explicit dependency of jobs which were submitted after the first one, which only explicitly depends on what's in env_batch.xml?
I've also realized I'm not handling the case where short term archiving jobs are submitted after the jobs complete, and need to think about how to deal with that and similar cases

jedwards4b: Dependancies between jobs in the variable "jobs" are defined in env_batch.xml
the only additional dependancy we need is between the last job in submit 1 and the first one in submit 2.

mfdeakin-sandia:
This passes the new tests, the rest of K_TestCimeCase, B_CheckCode, and Z_FullSystemTest.
Are there any other changes which should be made?

mfdeakin-sandia:
Fair point about the possibility of setting RESUBMIT to a positive value, I'll think about what I can do about that

mfdeakin-sandia:
This has the fix for the st_archive submitting in here now, and passes B_CheckCode, K_TestCimeCase, and most of Z_FullSystemTest (all be ERIO have passed, ERIO is pending).
I'll look into adding a test to ensure case.st_archive doesn't perform resubmits by default now.

jedwards4b:
@mfdeakin-sandia Our primary Cheyenne is down today, but I'm getting failures on the hobart cluster:
```
FAIL: test_cime_case_allow_failed_prereq (__main__.K_TestCimeCase)
  File "./scripts_regression_tests.py", line 1614, in test_cime_case_allow_failed_prereq
    self.assertTrue(len(batch_commands) == 1, "case.submit_jobs did not return any job submission strings")
AssertionError: case.submit_jobs did not return any job submission strings

FAIL: test_cime_case_resubmit_immediate (__main__.K_TestCimeCase)
  File "./scripts_regression_tests.py", line 1632, in test_cime_case_resubmit_immediate
    self.assertTrue(len(batch_commands) == num_submissions, "case.submit_jobs did not return {} submitted jobs".format(num_submissions))
AssertionError: case.submit_jobs did not return 6 submitted jobs

```

jedwards4b:
@jgfouca I think that this is ready, can you give it another once over?

mfdeakin-sandia:
Could someone verify the test I added in the latest commit is reasonable? The main thing that gives me pause is the second check, where it's submitting jobs which fail.

mfdeakin-sandia:
Yes, that's about right. I'll add it to the PR message

rljacob:
What does "--prereq-allow-failure" do?

mfdeakin-sandia:
Sorry for the delay. I've updated the PR comment which I think explains what both options are and how they are used

rljacob:
Can I mix workflows?  That is can I still run just the initial run by hand and then use --resubmit-immediate for the restart runs?

jedwards4b:
Yes.   You need to take care though not to run case.submit again while you still have jobs running or in the queue.   This is a necessary precaution with or without the current PR.

jedwards4b:
I would like to merge this sooner rather than later so that if there are any problems we can sort them out - any objections?

jgfouca:
None from me. @mfdeakin-sandia ?

rljacob:
You want this in your release, @jedwards4b ?

jedwards4b:
Yes, I think that it's a pretty good option to have.

jedwards4b:
The --resubmit-immediate that is, not too sure about the value of the other one.

rljacob:
Fine by me.

mfdeakin-sandia:
Sorry, I was thinking of making the last test I committed better, but am fine with it as is

jedwards4b:
You can always submit a new pr with improvements...

PR #2516, opened 2018-05-01
refine cheyenne retry method

jedwards4b:
Refining the retry code for cheyenne, it turns out that mpiexec may not return a non-zero stat on failure.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case/case_run.py
+++ b/scripts/lib/CIME/case/case_run.py
@@ -129,9 +129,9 @@ def _run_model_impl(case, lid, skip_pnl=False, da_cycle=0):
 
                     case.spare_nodes -= num_fails
 
-                if stat != 0 and not loop:
-                    # We failed and we're not restarting
-                    expect(False, "RUN FAIL: Command '{}' failed\nSee log file for details: {}".format(cmd, model_logfile))
+        if stat != 0 and not loop:
+            # We failed and we're not restarting
+            expect(False, "RUN FAIL: Command '{}' failed\nSee log file for details: {}".format(cmd, model_logfile))
 
     logger.info("{} MODEL EXECUTION HAS FINISHED".format(time.strftime("%Y-%m-%d %H:%M:%S")))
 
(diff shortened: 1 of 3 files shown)


jedwards4b:
@jgfouca please test this on titan and confirm that it still works correctly there.

scripts/lib/CIME/case/case_run.py
```diff
@@ -98,45 +98,40 @@ def _run_model_impl(case, lid, skip_pnl=False, da_cycle=0):
...
-                        case.create_namelists()
+                    case.create_namelists()
 
-                        case.spare_nodes -= num_fails
+                    case.spare_nodes -= num_fails
 
-            if not loop:
-                # We failed and we're not restarting
-                expect(False, "RUN FAIL: Command '{}' failed\nSee log file for details: {}".format(cmd, model_logfile))
+                if stat != 0 and not loop:
```
jgfouca: I think the indentation is definitely wrong here. This check used to be under the 'if stat != 0' block. It now needs to be under the while loop since you removed that check.

scripts/lib/CIME/case/case_run.py
```diff
@@ -98,45 +98,40 @@ def _run_model_impl(case, lid, skip_pnl=False, da_cycle=0):
...
-                        lid = new_lid()
-                        loop = True
+                    lid = new_lid()
+                    loop = True
 
-                        case.create_namelists()
+                    case.create_namelists()
 
-                        case.spare_nodes -= num_fails
+                    case.spare_nodes -= num_fails
```
jgfouca: You'll need a 'continue' here to avoid the expect(False, ...) below once the indentation of that check is corrected.

jedwards4b: I don't think you do since loop = True

jgfouca: You're right.

PR #2519, opened 2018-05-01
mali is an mpas model and needs special handling

### LINKED ISSUE #2522: case.st_archive --test  doesn't work
```
% ./create_test SMS.f45_g37_rx1.A --no-build
...
% cd $case
% ./case.st_archive --test
...
ERROR: Failed to move file mali.hist.am.globalStats.1976-01-01.nc to archive
```

jedwards4b:
Fix st archive for e3sm mali component.

Test suite: case.st_archive --test for e3sm
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #2522 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/e3sm/config_archive.xml
+++ b/config/e3sm/config_archive.xml
@@ -113,7 +113,7 @@
   <comp_archive_spec compname="mali" compclass="glc">
     <rest_file_extension>rst</rest_file_extension>
     <rest_file_extension>rst.am.timeSeriesStatsMonthly</rest_file_extension>
-    <hist_file_extension>hist</hist_file_extension>
+    <hist_file_extension>hist.am</hist_file_extension>
     <rest_history_varname>unset</rest_history_varname>
     <rpointer>
       <rpointer_file>rpointer.glc$NINST_STRING</rpointer_file>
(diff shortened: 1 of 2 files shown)


rljacob:
Why does mali need special handling?  cc @matthewhoffman 

We are modifying mpas filenames to fit CIME expectations so could fix this.

jedwards4b:
mali needs the same handling as other mpas models - we still have a hardcoded section for that in case_st_archive.py  which is looking for an mpas prefix in component name.  As far as I can tell you haven't modified any mpas filenames yet.

jonbob:
we have a PR and it should be in soon

rljacob:
By the time this change makes it back in the E3SM, we will have updated the filenames so that this is no longer necessary.

jedwards4b:
this is causing the cime_regression_tests.py to fail now though

rljacob:
Oh ok then.   @jgfouca wait for https://github.com/E3SM-Project/E3SM/pull/2308 to be merged before doing an E3SM -> ESMCI split.

jedwards4b:
@jgfouca I would like to get the cdash testing all passing again - this PR is needed for that - can we merge it?

jgfouca:
@jedwards4b , I think we'll just have to undo this change when the PR @rljacob mentioned comes into CIME. Maybe we can just disable the test until then or tolerate the fail?

PR #2524, opened 2018-05-01
(WIP) work toward immediate resubmit

jedwards4b:
Submitted for discussion with @mfdeakin-sandia  this is untested code in which I've modified the loop in env_batch.py to do immediate resubmits.

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/Tools/case.submit
+++ b/scripts/Tools/case.submit
@@ -6,7 +6,6 @@ case.submit - Submit a cesm workflow to the queueing system or run it if there i
 
 from standard_script_setup import *
 from CIME.case        import Case
-from six.moves import configparser
 
 ###############################################################################
 def parse_command_line(args, description):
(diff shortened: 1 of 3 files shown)


mfdeakin-sandia:
Sorry, I'll test this later, but I don't see how it solves the issue with `CONTINUE_RUN`? In this case, it doesn't look like it's set at all for runs after the first one.

jedwards4b:
This code is incomplete, I'm just trying to illustrate the way I think the RESUBMIT loop should work.

mfdeakin-sandia:
Alright, so if I pull this into the branch I have, would that be acceptable? Once I fix all the other issues, that is.
I was considering cleaning things up more, maybe sharing the code which sets `CONTINUE_RUN` with the resubmit version.

jedwards4b:
Sure you can pull this into your branch and we'll look at what you have after you clean it up.  I think we'll need to iterate a few more times before we get to where we want to go.

jedwards4b:
Incorporated in #2511

PR #2526, opened 2018-05-01
rename NODE_FAIL RETRY_MPIRUN

jedwards4b:
Renaming some variables to clarify functionality NODE_FAIL was too narrow in scope since there
are other reasons you might want to reattempt the mpirun command

Test suite: scripts_regression_tests, hand testing of retry code on cheyenne (in progress)
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case/case_run.py
+++ b/scripts/lib/CIME/case/case_run.py
@@ -131,17 +131,18 @@ def _run_model_impl(case, lid, skip_pnl=False, da_cycle=0):
                     logger.warning("Detected model run failed due to node failure, restarting")
                     case.spare_nodes -= num_node_fails
                     loop = True
+                    case.set_value("CONTINUE_RUN",
+                                   case.get_value("RESUBMIT_SETS_CONTINUE_RUN"))
                 elif num_retry_fails > 0 and retry_count >= num_retry_fails:
                     logger.warning("Detected model run failed, restarting")
-                    retry_count -= num_retry_fails
+                    retry_count -= 1
                     loop = True
                 if loop:
                     # Archive the last consistent set of restart files and restore them
                     if case.get_value("DOUT_S"):
                         case.case_st_archive(no_resubmit=True)
                         case.restore_from_archive()
-                    case.set_value("CONTINUE_RUN",
-                                   case.get_value("RESUBMIT_SETS_CONTINUE_RUN"))
+
                     lid = new_lid()
                     case.create_namelists()
 
--- a/scripts/Tools/case.submit
+++ b/scripts/Tools/case.submit
@@ -73,7 +73,7 @@ def _main_func(description):
         config = configparser.SafeConfigParser()
         config.add_section("SubmitOptions")
         if skip_pnl:
-            config.set("SubmitOptions", "skip_pnl", "True")
+            config.set("SubmitOptions", "skip_pnl", b"True")
         if mail_user:
             config.set("SubmitOptions", "mail_user", mail_user)
         if mail_type:


scripts/lib/CIME/case/case_run.py
```diff
@@ -106,11 +110,13 @@ def _run_model_impl(case, lid, skip_pnl=False, da_cycle=0):
         stat = run_and_log_case_status(run_func, "model execution", caseroot=case.get_value("CASEROOT"))
         model_logfile = os.path.join(rundir, model + ".log." + lid)
         # Determine if failure was due to a failed node, if so, try to restart
-        if node_fail_re:
```
jgfouca: We can have two different "counters" here, one for node fails and one for mpirun retries.

scripts/lib/CIME/case/case_submit.py
```diff
@@ -123,8 +123,6 @@ def submit(self, job=None, no_batch=False, prereq=None, resubmit=False,
     if resubmit and os.path.exists(submit_options):
         config = configparser.SafeConfigParser()
         config.read(submit_options)
-        if not skip_pnl and config.has_option('SubmitOptions','skip_pnl'):
```
jgfouca: What's the reason for this change?

jedwards4b: the skip_pnl is a logical and needs to be converted to text to write, I'm not sure why this didn't fail in the original PR but it's failing now.

jedwards4b:
I really don't think that we need two REGEX variables. I can add another counter and avoid overloading case.spare_nodes.  As far as variable naming, I'm a little salty since I ran those name by you before I made the change, but yea I can change them again.

jgfouca:
Sorry, I guess I spaced it out or didn't notice the order flip flop.

jgfouca:
I do think we need two REGEX though... what if a machine needs both node-fail handling and mpi retry?

jedwards4b:
You could handle both conditions in the same regex.   I'm not saying that's the greatest idea but I am leery of a proliferation of variables to handle every new case that comes along.

jgfouca:
@jedwards4b , i'm fine with staying with your current implementation until the need arises for different regexs.

jedwards4b:
ha - too late

jedwards4b:
@jgfouca reworked, testing in progress, please review

scripts/lib/CIME/case/case_run.py
```diff
@@ -106,29 +115,36 @@ def _run_model_impl(case, lid, skip_pnl=False, da_cycle=0):
...
+                elif num_retry_fails > 0 and retry_count >= num_retry_fails:
+                    logger.warning("Detected model run failed, restarting")
+                    retry_count -= num_retry_fails
+                    loop = True
+                if loop:
                     # Archive the last consistent set of restart files and restore them
-                    case.case_st_archive(no_resubmit=True)
-                    case.restore_from_archive()
-
+                    if case.get_value("DOUT_S"):
```
jgfouca: So, if DOUT_S is off, then how will these retry attempts work at all?

scripts/lib/CIME/case/case_run.py
```diff
@@ -106,29 +115,36 @@ def _run_model_impl(case, lid, skip_pnl=False, da_cycle=0):
...
+                    num_retry_fails = len(retry_run_regex.findall(open(model_logfile, 'r').read()))
+                logger.debug ("RETRY: num_retry_fails {} spare_nodes {} retry_count {}".
+                              format(num_retry_fails, case.spare_nodes, retry_count))
+                if num_node_fails > 0 and case.spare_nodes >= num_node_fails:
                         # We failed due to node failure!
                     logger.warning("Detected model run failed due to node failure, restarting")
-
+                    case.spare_nodes -= num_node_fails
+                    loop = True
+                elif num_retry_fails > 0 and retry_count >= num_retry_fails:
```
jgfouca: Do you want mpirun-retry attempts to decrease by num_retry_fails or 1? I would think 1. The reason we decrease by num_node_fails above is because we assume multiple occurrences of the regex means multiple nodes failed. I wouldn't think you'd want mpi retry to work that way, since it could be printing the regex message NP times.

jedwards4b: In the current cheyenne case num_retry_fails will always be 1.   I could change this in anticipation of some other issue that is reported NP times or the user could just be clever with the REGEX.

scripts/lib/CIME/case/case_run.py
jedwards4b: If DOUT_S is off the restart files remain in the run directory - in most cases DOUT_S is true.   In cases where DOUT_S is False the archive directory was never created and so you would get an error trying to write there.

scripts/lib/CIME/case/case_run.py
jgfouca: Is there any case where we wouldn't want it to be 1?

scripts/lib/CIME/case/case_run.py
jgfouca: OK, so I guess this is the best we're going to do here.

scripts/lib/CIME/case/case_run.py
jedwards4b: I can't think of a case - I will make this change.

jedwards4b:
@jgfouca I've made the suggested change and one further change for CONTINUE_RUN.   I am retesting but our system is misbehaving today so it could take some time.

Issue #2504, opened 2018-04-26
Cannot write cpl aux hist files due to bug in format statement


billsacks:
When trying to write a coupler auxiliary history file, I got this death:

```
1:forrtl: severe (27): too many records in I/O statement, unit -5, file Internal Formatted Write
1:Image              PC                Routine            Line        Source
1:cesm.exe           000000000196C2EC  Unknown               Unknown  Unknown
1:cesm.exe           00000000019C4D0D  Unknown               Unknown  Unknown
1:cesm.exe           00000000004BFF0D  seq_hist_mod_mp_s        1153  seq_hist_mod.F90
1:cesm.exe           000000000040F671  cime_comp_mod_mp_        3761  cime_comp_mod.F90
1:cesm.exe           0000000000424942  MAIN__                    103  cime_driver.F90
```

The problem is that, in the following, `a6` should be `6a`:

https://github.com/ESMCI/cime/blob/2e0ba3b318e9abd1ddc00dfc223b3306368a489f/src/drivers/mct/main/seq_hist_mod.F90#L1150-L1151

I'll fix this in the course of fixing #2446 .

cc @jedwards4b

bishtgautam:
Yes, ELM is using cpl history files for offline BGC spinup.

Issue #2446, opened 2018-04-11
Need to fix T compsets


billsacks:
Both the generation and use of forcing data for T compsets (glc-only, typically forced by cplhist output) have been somewhat broken for a long time. CESM land ice scientists want this fixed for CESM2.

PR #2530, opened 2018-05-02
Fix generation and use of forcing data for T compsets

billsacks:
Both the generation and use of forcing data for T compsets (glc-only,
typically forced by cplhist output) have been somewhat broken for a long
time. This PR fixes both the generation and use of this cplhist dlnd
forcing data.

I have confirmed that this is working properly by running companion I
and T compset cases, where I generate data from the I compset and then
use it in the T compset (making sure that the processor count is the
same for the T compset as for the I compset):

- Generating data from a 3-year run of I1850Clm50Sp at resolution
  f10_f10_musgs with histaux_l2x1yrg = .true. and
  histaux_double_precision = .true., then running 3 years of T1850G with
  `./xmlchange CISM_EVOLVE=FALSE`, I get bit-for-bit results.

- When I instead used I1850Clm50SpG (which has ice evolution turned on),
  and ran the T1850G compset with `glc_renormalize_smb = 'on'`, I got
  bit-for-bit results for the first 2 years, but then roundoff-level
  differences in qice forcing in many grid cells in year 3 (and thus
  small differences in many CISM fields in year 3). Since the
  differences were only roundoff-level, I did not pursue this further.

Test suite:
- scripts_regression_tests on hobart (ran from cdd168651)
- aux_cime_baselines from cesm2_0_alpha10e (compared this cime branch
  with the version of master on which it was based: 974839b)
  - Tests pass except for two that failed in the baseline:
    ERP_D_Ld10.f10_f10_musgs.IHistClm50BgcCrop.cheyenne_intel.clm-ciso_decStart
    and SMS_N3_PM3_Ld2.f19_g17.BHISTWs.cheyenne_gnu.allactive-defaultio
  - NLCOMP failures as expected
  - BASELINE comparisons bit-for-bit except for
    SMS_Ly2.f09_g16_gl20.T1850G.cheyenne_intel, as expected
- SMS_Ly3.f09_g17_gl4.T1850G.hobart_intel from ESCOMP/cism-wrapper@bc15238
- SMS_D_Ld3.f09_f09_mg17.A1850DLND.hobart_intel
  - Note: This out-of-the-box A1850DLND case passes. However, if I set
    GLC_NEC to 0 (as suggested by @mvertens), this fails (despite
    passing on master). I think this is because runs with GLC_NEC=0 now
    have no fields listed in the streams file. This seems okay to me:
    running A1850DLND with a manual setting of GLC_NEC=0 seems like a
    non-sensical thing to do in practice. (This was more important in
    the past when GLC_NEC=0 was the default; now GLC_NEC=10 is the
    default.)

Test baseline: N/A
Test namelist changes: Namelists will differ for cases with dlnd
Test status: climate changing for cases with dlnd (T compsets), should
  otherwise be bit-for-bit

Fixes ESMCI/cime#2446
Fixes ESMCI/cime#2504

User interface changes?:
- Renamed drv namelist variable histaux_l2x1yr to histaux_l2x1yrg to
  denote that it's just for glc fields (following the naming convention
  used for histaux_a2x3hrp)
- Added drv namelist variable histaux_double_precision: if true, use
  double-precision rather than single-precision for coupler auxiliary
  history files

Code review:


src/drivers/mct/main/cime_comp_mod.F90
```diff
@@ -2094,6 +2095,10 @@ subroutine cime_run()
     logical            :: drv_pause  ! Driver writes pause restart file
     character(len=CL)  :: drv_resume ! Driver resets state from restart file
 
+    type(ESMF_Time)    :: etime_curr            ! Current model time
```
billsacks: I introduced local variables in this subroutine. I notice that, for the most part, this module uses module-level variables rather than subroutine-local variables. I couldn't bring myself to follow that convention, but could change this if you feel within-module consistency is more important than following typical Fortran coding practices.

src/drivers/mct/main/cime_comp_mod.F90
```diff
@@ -3716,15 +3724,54 @@ subroutine cime_run()
...
+                write(logunit,*) 'ERROR: histaux_l2x1yrg requested;'
+                write(logunit,*) 'lnd2glc fields were averaged this time step, but it is not the year boundary.'
+                write(logunit,*) '(It only works to request histaux_l2x1yrg if GLC_AVG_PERIOD is yearly.)'
+                call shr_sys_abort(subname// &
+                     ' do_hist_l2x1yrg and lnd2glc_averaged_now are true, but t1yr_alarm is false')
+             end if
+
+             if (t1yr_alarm) then
+                call seq_timemgr_EClockGetData( EClock_d, ECurrTime = etime_curr)
+                call shr_cal_ymds2rday_offset(etime=etime_curr, &
```
billsacks: I struggled with what would be the best design for getting tbnds1_offset... I'm not sure I did this in the best way

billsacks:
I want to run CESM's aux_cime_baselines test suite to ensure that my data model-related changes did not inadvertently break other configurations. I haven't been able to do so yet due to the flakiness of cheyenne, but plan to do so as soon as the machine stabilizes.

billsacks:
@mvertens - I'm done commenting on this; it is ready for review.

billsacks:
I have run aux_cime_baselines. Results were as expected. I have updated the main comment at the top of the PR with details.

src/drivers/mct/main/cime_comp_mod.F90
billsacks: @mvertens suggests - put a comment here saying: normally we'd have rdays_offset = -365, but to be more general (e.g., for gregorian calendar) we use this function

PR #2531, opened 2018-05-02
add more checks for time settings

### LINKED ISSUE #1324: Check on coupling frequencies
_From @quantheory on July 30, 2015 0:37_

Tony [proposed some time ago](http://bugs.cgd.ucar.edu/show_bug.cgi?id=1740) that the following checks be added to the coupler, to ensure that the various NCPL variables are set in a consistent way:

ATM_NCPL >= ROF_NCPL
ATM_NCPL >= OCN_NCPL

_Copied from original issue: CESM-Development/cime#81_

### LINKED ISSUE #1707: Coupling/run length check to prevent restarts from failing
When a run length is specified that is shorter than the longest coupling interval, necessary restart files are not written.  A check should be put into the driver to abort if this is the case.

jedwards4b:
Add checks that stop_n is at least as long as the longest coupling interval and that atm_ncpl >= ocn_ncpl and atm_ncpl >= rof_ncpl

Test baseline: 
Test namelist changes: 
Test status: bit for bit,

Fixes #1324 
Fixes #1707 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/drivers/mct/cime_config/config_component_cesm.xml
+++ b/src/drivers/mct/cime_config/config_component_cesm.xml
@@ -277,9 +277,6 @@
       <value compset="_POP2" grid="oi%gx1v7">24</value>
       <value compset="_DATM%NYF.*_SLND.*_DICE.*_DOCN.*_SWAV">1</value>
       <value compset="_DATM%NYF.*_DLND.*_DICE.*_DOCN.*_DWAV">1</value>
-      <value compset="_XATM.*_XLND.*_XICE.*_XOCN">1</value>
-      <value compset="_DATM.*_CLM.*_SICE.*_SOCN">1</value>
-      <value compset="_SATM.*_SLND.*_SICE.*_SOCN">1</value>
       <value compset="_DLND.*_CISM\d">1</value>
     </values>
     <group>run_coupling</group>
--- a/src/drivers/mct/cime_config/config_component_cesm.xml
+++ b/src/drivers/mct/cime_config/config_component_cesm.xml
@@ -278,6 +278,8 @@
       <value compset="_DATM%NYF.*_SLND.*_DICE.*_DOCN.*_SWAV">1</value>
       <value compset="_DATM%NYF.*_DLND.*_DICE.*_DOCN.*_DWAV">1</value>
       <value compset="_DLND.*_CISM\d">1</value>
+      <value compset="_XOCN">$ATM_NCPL</value>
+      <value compset="_SOCN">$ATM_NCPL</value>
     </values>
     <group>run_coupling</group>
     <file>env_run.xml</file>
@@ -288,9 +290,11 @@
 
   <entry id="GLC_NCPL">
     <type>integer</type>
-    <default_value>$ATM_NCPL</default_value>
+    <default_value>1</default_value>
     <values match="last">
       <value compset="_DLND.*_CISM\d">1</value>
+      <value compset="_SGLC">$ATM_NCPL</value>
+      <value compset="_XGLC">$ATM_NCPL</value>
     </values>
     <group>run_coupling</group>
     <file>env_run.xml</file>
@@ -320,7 +324,7 @@
 
   <entry id="ROF_NCPL">
     <type>integer</type>
-    <default_value>$ATM_NCPL</default_value>
+    <default_value>8</default_value>
     <values match="last">
       <value compset="_DATM.*_POP2.*_DROF">$ATM_NCPL</value>
       <value compset="_DATM.*_DOCN%SOM">$ATM_NCPL</value>
@@ -328,6 +332,8 @@
       <value compset="_DATM%CPLHIST.+POP\d">8</value>
       <value compset="_XATM.*_XLND.*_XICE.*_XOCN">$ATM_NCPL</value>
       <value compset="_DLND.*_CISM\d">1</value>
+      <value compset="_SROF">$ATM_NCPL</value>
+      <value compset="_XROF">$ATM_NCPL</value>
     </values>
     <group>run_coupling</group>
     <file>env_run.xml</file>
--- a/src/drivers/mct/cime_config/config_component_cesm.xml
+++ b/src/drivers/mct/cime_config/config_component_cesm.xml
@@ -33,7 +33,7 @@
     <group>run_metadata</group>
     <file>env_case.xml</file>
     <desc>run DOI</desc>
-  </entry> 
+  </entry>
 
   <entry id="DRV_THREADING">
     <type>logical</type>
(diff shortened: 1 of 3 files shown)


jedwards4b:
@billsacks I think that I need to ignore GLC_NCPL in this or set GLC_NCPL=ATM_NCPL when GLC is XGLC?

billsacks:
@jedwards4b do you mean SGLC? Good point. More generally, it seems like this check shouldn't be done for any stub component. e.g., if you're running with SROF with a rof coupling frequency of 3 hours, it's fine to restart more frequently than 3 hours, right?

jedwards4b:
Sorry - yes I mean SGLC

jedwards4b:
The issue is that I don't have information about the type of component in seq_time_manager where I need to do these checks, so it's easier to set the stub component intervals to $NCPL_ATM instead.

billsacks:
> The issue is that I don't have information about the type of component in seq_time_manager where I need to do these checks, so it's easier to set the stub component intervals to $NCPL_ATM instead.

Okay, I guess that's fine... I assume the performance impact would be miniscule. Actually, maybe this is an issue for XGLC, too, because it also uses a daily coupling interval (which would just be an issue if you tried to run an X compset restart test with restart interval less than daily). In addition to GLC, it looks like it may be an issue for OCN and ROF: I'm thinking that you'd need to set the coupling interval to ATM_NCPL when running with stub versions of those components?

jedwards4b:
@billsacks 
This test: ERS_Ln48.f10_f10_musgs.I1850Clm50Sp.cheyenne_intel
Passes on this branch.

src/drivers/mct/cime_config/config_component_cesm.xml
```diff
@@ -291,7 +288,7 @@
 
   <entry id="GLC_NCPL">
     <type>integer</type>
-    <default_value>1</default_value>
+    <default_value>$ATM_NCPL</default_value>
```
billsacks: This will change coupling frequency for cases with active land and CISM, which we don't want to do.

src/drivers/mct/cime_config/config_component_cesm.xml
```diff
@@ -323,7 +320,7 @@
 
   <entry id="ROF_NCPL">
     <type>integer</type>
-    <default_value>8</default_value>
+    <default_value>$ATM_NCPL</default_value>
```
billsacks: This changes the coupling frequency for I compsets, and maybe others, which we don't want to do.

billsacks:
> @billsacks
> This test: ERS_Ln48.f10_f10_musgs.I1850Clm50Sp.cheyenne_intel
> Passes on this branch.

That's because you have changed coupling frequency for ROF. We don't want that. See the review comment that I just posted.

billsacks:
@jedwards4b and @mvertens - Based on the iterations we're going through here, it appears that it's challenging to resolve https://github.com/ESMCI/cime/issues/1707 without introducing other issues. I'm starting to think that, so close to release, we should leave https://github.com/ESMCI/cime/issues/1707 unresolved rather than risk introducing some new problems due to unforeseen side-effects that might be more serious than the original.

jedwards4b:
I think that it should have had these defaults all along, but you are right it's too close to the release for this change I can take a more conservative approach, I don't think we need to completely abandon.

billsacks:
Okay, I'm still willing to entertain a more conservative approach.

jedwards4b:
@billsacks I've added code that correctly calls out the error in ERS_Ln48.f10_f10_musgs.I1850Clm50Sp
However when I rerun that case with REST_N=30 which should be a suitable interval for the ROF component, it still doesn't write an ROF restart.   I'm confused by the ROF restart alarm as printed in the cpl log - why is it different than the other components?   
For ROF:
```
(seq_timemgr_clockPrint)   Alarm =  4  seq_timemgr_alarm_restart
(seq_timemgr_clockPrint)     Prev Time   =     00010101   00000
(seq_timemgr_clockPrint)     Next Time   =     00010104   64800
(seq_timemgr_clockPrint)     Intervl yms =        0       0      324000
```
While for ATM:
```
(seq_timemgr_clockPrint)   Alarm =  4  seq_timemgr_alarm_restart
(seq_timemgr_clockPrint)     Prev Time   =     00010101   00000
(seq_timemgr_clockPrint)     Next Time   =     00010101   54000
(seq_timemgr_clockPrint)     Intervl yms =        0       0       54000
```

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -1000,6 +1017,22 @@ subroutine seq_timemgr_clockInit(SyncClock, nmlfile, restart, restart_file, pioi
           write(logunit,*) subname,' ERROR: offset not multiple',n,dtime(seq_timemgr_nclock_drv),offset(n)
           call shr_sys_abort()
        endif
+
+       call ESMF_TimeIntervalSet( TimeStep, s=dtime(n), rc=rc )
+       if(CurrTime + TimeStep > minStopTime ) then
+          call shr_sys_abort(subname//"ERROR: Stop time too short, increase STOP_N")
```
billsacks: Please remove this check on stop time. For testing purposes, I can see value in running the model for just a few time steps - and as long as you're not trying to write restart intervals I can't see any reason why we should stop people from doing this.

(Then you can also remove the minStopTime variable.)

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -907,6 +912,11 @@ subroutine seq_timemgr_clockInit(SyncClock, nmlfile, restart, restart_file, pioi
             RefTime = CurrTime,            &
             alarmname = trim(seq_timemgr_alarm_restart))
 
+       call seq_timemgr_alarmGet(SyncClock%EAlarm(n,seq_timemgr_nalarm_restart), IntSec=RestInterval)
+       if(n==1 .or. RestInterval < minRestInterval) then
```
billsacks: Since Fortran doesn't use short-circuit boolean operators, I believe some compilers will complain about uninitialized variables when executing this in debug mode. It should be reworked to look like what you did for setting minStopTime.

billsacks:
> However when I rerun that case with REST_N=30 which should be a suitable interval for the ROF component, it still doesn't write an ROF restart. I'm confused by the ROF restart alarm as printed in the cpl log - why is it different than the other components?

No idea; maybe we can try to puzzle it out together; I'll stop by shortly.

jedwards4b:
Opened issue #2537 to deal with the problem this PR has reveled.   Change STOP_N too short message to a warning.

jedwards4b:
Yes - I am rerunning tests and will let you know when that completes.

jedwards4b:
@gold2718 there is some weirdness going on with the PRE test we need to talk about.

jedwards4b:
@billsacks This is ready now, needed to ignore esp component in restart check.

PR #2540, opened 2018-05-03
Fix restart alarm calc

### LINKED ISSUE #2537: component Restart time interval calculation incorrect when REST_OPTION=nsteps
All components should have restart alarms at the same time but when using nsteps to compute
the restart alarm components are using their own timestep instead of the coupler timestep to compute nsteps and so alarms are not consistent across components.    Fix this and add a check that all component restart alarms are consistent.

jedwards4b:
When intervals are specified in terms of steps , stop and restart alarms for components need to be calculated based on the coupler step not the component step.   Add a check that restart alarm is consistant across components except esp. 

Test baseline: 
Test namelist changes: 

Fixes #2537 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/drivers/mct/shr/seq_timemgr_mod.F90
+++ b/src/drivers/mct/shr/seq_timemgr_mod.F90
@@ -314,7 +314,7 @@ contains
     character(SHR_KIND_CS)  :: pause_option          ! Pause option units
     integer(SHR_KIND_IN)    :: pause_n               ! Number between pause intervals
     integer(SHR_KIND_IN)    :: RestInterval              ! Component Restart Interval
-    integer(SHR_KIND_IN)    :: minRestInterval        ! Driver Restart Interval
+    integer(SHR_KIND_IN)    :: drvrestinterval        ! Driver Restart Interval
 
     logical :: pause_active_atm
     logical :: pause_active_cpl
@@ -875,12 +875,10 @@ contains
     end if
 
     ! --- Initialize component and driver clocks and alarms common to components and driver clocks ---
-
+    call ESMF_TimeIntervalSet( CplTimeStep, s=dtime(seq_timemgr_nclock_drv), rc=rc )
     do n = 1,max_clocks
        call ESMF_TimeIntervalSet( TimeStep, s=dtime(n), rc=rc )
-       if (n == seq_timemgr_nclock_drv) then
-          CplTimeStep = TimeStep
-       endif
+
        call seq_timemgr_ESMFCodeCheck( rc, subname//': error ESMF_TimeIntervalSet' )
 
        call seq_timemgr_EClockInit( TimeStep, StartTime, RefTime, CurrTime, SyncClock%ECP(n)%EClock)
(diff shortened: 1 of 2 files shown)


jedwards4b:
@billsacks ready for your review - this also has a fix for a test in scripts_regression_tests.py

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -910,6 +916,7 @@ subroutine seq_timemgr_clockInit(SyncClock, nmlfile, restart, restart_file, pioi
             opt_n   = restart_n,           &
             opt_ymd = restart_ymd,         &
             RefTime = CurrTime,            &
+            cplTimeStep = cplTimeStep, &
             alarmname = trim(seq_timemgr_alarm_restart))
 
        call seq_timemgr_alarmGet(SyncClock%EAlarm(n,seq_timemgr_nalarm_restart), IntSec=RestInterval)
```
billsacks: It seems like the following 6 lines are no longer needed: from my read of the code, minRestInterval is set again, in a different way, before it is ever needed.

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -312,8 +313,9 @@ subroutine seq_timemgr_clockInit(SyncClock, nmlfile, restart, restart_file, pioi
     integer(SHR_KIND_IN)    :: restart_ymd           ! Restart date (YYYYMMDD)
     character(SHR_KIND_CS)  :: pause_option          ! Pause option units
     integer(SHR_KIND_IN)    :: pause_n               ! Number between pause intervals
-    integer(shr_kind_in)    :: restinterval          ! restart interval seconds
-    integer(shr_kind_in)    :: minrestinterval          ! restart interval seconds
+    integer(SHR_KIND_IN)    :: RestInterval              ! Component Restart Interval
+    integer(SHR_KIND_IN)    :: minRestInterval        ! Driver Restart Interval
```
billsacks: A better name for this now would be drvRestInterval

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -1019,26 +1029,23 @@ subroutine seq_timemgr_clockInit(SyncClock, nmlfile, restart, restart_file, pioi
...
              if (RestInterval .ne. minRestInterval) then
                 write(logunit,*) subname, 'RestInterval=',RestInterval,&
                      ' minRestInterval=',minRestInterval
                 call shr_sys_abort("Component RestInterval inconsistant with driver")
              endif
-             if (mod(minRestInterval, dtime(n)) .ne. 0) then
-                write(logunit,*) subname, 'RestInterval=',minRestInterval,&
-                     ' Dtime=',dtime(n)
-                call shr_sys_abort("Restart interval not compatible with dtime")
+             if (RestInterval /= minRestInterval) then
```
billsacks: This check seems to now be duplicating the check a few lines up. Unless I'm missing something, one of these two blocks should be deleted.

Also, spelling fix: inconsistant should be inconsistent.

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -1538,13 +1546,21 @@ subroutine seq_timemgr_alarmInit( EClock, EAlarm, option, opt_n, opt_ymd, opt_to
        call ESMF_TimeSet( NextAlarm, yy=cyy, mm=cmm, dd=opt_n, s=0, calendar=seq_timemgr_cal, rc=rc )
 
     case (seq_timemgr_optNSteps)
-       call ESMF_ClockGet(EClock, TimeStep=AlarmInterval, rc=rc)
+       if (present(CplTimeStep)) then
+          AlarmInterval = CplTimeStep
+       else
+          call ESMF_ClockGet(EClock, TimeStep=AlarmInterval, rc=rc)
```
billsacks: Should this abort (with a message like, "for optNSteps, must provide coupler time step") rather than falling back to the incorrect (as I understand it) use of the component's time step?

src/drivers/mct/shr/seq_timemgr_mod.F90
```diff
@@ -1538,13 +1546,21 @@ subroutine seq_timemgr_alarmInit( EClock, EAlarm, option, opt_n, opt_ymd, opt_to
...
        if (.not.present(opt_n)) call shr_sys_abort(subname//trim(option)//' requires opt_n')
        if (opt_n <= 0)  call shr_sys_abort(subname//trim(option)//' invalid opt_n')
        AlarmInterval = AlarmInterval * opt_n
 
     case (seq_timemgr_optNStep)
-       call ESMF_ClockGet(EClock, TimeStep=AlarmInterval, rc=rc)
+       if (present(CplTimeStep)) then
+          AlarmInterval = CplTimeStep
+       else
+          call ESMF_ClockGet(EClock, TimeStep=AlarmInterval, rc=rc)
```
billsacks: See comment a few lines up.

PR #2549, opened 2018-05-04
Convert e3sm grids

jgfouca:
New tool for converting v1 to v2 grids.
Convert e3sm/config_grids.xml to v2.

Test suite: scripts_regression_tests, cime_developer with baseline comparison
Test baseline: 
Test namelist changes: 
Test status: bit for bi

Fixes [CIME Github issue #]

User interface changes?: Yes, e3sm grids are now v2

Code review: @jedwards4b @rljacob

### DIFF
--- a/config/e3sm/config_grids.xml
+++ b/config/e3sm/config_grids.xml
@@ -1,1262 +1,1611 @@
 <?xml version="1.0"?>
 
-<grid_data>
+<grid_data version="2.0">
 
   <help>
     =========================================
     GRID naming convention
     =========================================
-    The long grid name has the order atm,lnd,ocn/ice,river,mask,glc
-    The following shortname is used
-    a% => atm, l% => lnd, oi% => ocn/ice, r% => river, m% => mask, g% => glc, w% => wav
-    The notation is
+    The notation for the grid longname is
     a%name_l%name_oi%name_r%name_m%mask_g%name_w%name
-    Each grid is associated with three required elements and one optional element
-    longname      (a%name_l%name_oi%name_r%name_m%mask_g%name_w%name)
-    shortname     (for backwards compatibility)
-    alias         (even shorter notation than shortname) (optional)
-    support       (Description of the support level for this grid) (optional)
-    Each grid can also be associated with the following optional attributes
-    compset       (Regular expression for compsets that are required to match for this grid) (optional)
+    where
+    a% => atm, l% => lnd, oi% => ocn/ice, r% => river, m% => mask, g% => glc, w% => wav
+
+    Supported out of the box grid configurations are given via alias specification in
+    the file "config_grids.xml". Each grid alias can also be associated  with the
+    following optional attributes
+
+    compset       (Regular expression for compset matches that are required for this grid)
+    not_compset   (Regular expression for compset matches that are not permitted this grid)
+
+    Using the alias and the optional "compset" and "not_compset" attributes a grid longname is created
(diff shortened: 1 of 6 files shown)


scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -278,14 +278,17 @@ def write(self, outfile=None):
         if outfile is None:
             outfile = self.filename
 
-        logger.debug("write: " + outfile)
+        # logger.debug("write: " + outfile)
```
jgfouca: May throw these changes out.

scripts/Tools/xmlconvertors/convert-grid-v1-to-v2
```diff
@@ -0,0 +1,217 @@
...
+
+        for child in v1file_obj.get_children(root=grid_block):
+            expect(v1file_obj.name(child) in ["lname", "sname", "alias", "support"],
+                   "Unhandled child of grid '{}'".format(v1file_obj.name(child)))
+
+        grid_data.append((compset, lname, sname, alias, support))
+
+    v2grids = v2file_obj.make_child("grids")
+
+    # TODO: How to leverage model_grid_defaults
```
jgfouca: I had no idea how to use model_grid_defaults in this conversion.

config/e3sm/config_grids.xml
```diff
@@ -1,1262 +1,1611 @@
...
-      <sname>hcru_hcru</sname>
-      <lname>a%360x720cru_l%360x720cru_oi%360x720cru_r%r05_m%360x720cru_g%null_w%null</lname>
-    </grid>
-
-    <grid compset="(DATM_SATM).+CLM">
-      <sname>NLDAS_NLDAS</sname>
-      <lname>a%NLDAS_l%NLDAS_oi%NLDAS_r%NLDAS_m%NLDAS_g%null_w%null</lname>
-    </grid>
+
+    <model_grid_defaults>
```
jgfouca: model_grid_defauls was just copied verbatim from cesm/config_grids.xml

config/e3sm/config_grids.xml
```diff
@@ -1266,934 +1615,939 @@
...
 
   </domains>
 
-  <!-- ======================================================== -->
-  <!-- Mapping -->
-  <!-- ======================================================== -->
+  <!-- The following are the required grid maps that must not be idmap if the   -->
+  <!-- attributes grid1 and grid2 are not equal -->
+
+  <required_gridmaps>
```
jgfouca: required_gridmaps was copied verbatim from cesm/config_grids.xml

jgfouca:
TODO: Remove v1 legacy code

scripts/Tools/xmlconvertors/convert-grid-v1-to-v2
jedwards4b: These are features that were not in v1.  model_grid_defaults is the grid that component will use if it is not explicitly called out in a grid definition.   So CISM2 will always use gland4 and rtm will always use r05 unless they are explicit in the grid name.  You may want to add some defaults for your mpas based grids.

config/e3sm/config_grids.xml
jedwards4b: Required_gridmaps calls out the grid maps that will be required for interactions between grid1 and grid2.  There are three possible map types, flux, vector, scalar.  All three grids may not be required for interactions between a given pair of components.  I don't expect that there would be any difference here between cesm and e3sm.

PR #2550, opened 2018-05-04
st_archive interactive resubmit fix

### LINKED ISSUE #2547: Running case.st_archive with RESUBMIT > 0 submits case.run jobs

mfdeakin-sandia:
This detects whether the short term archiver is being run from the queue or not, and if not, does not perform any resubmits.
This still needs a unit test, but that should be easy

Test status: in progress

Fixes #2547 

User interface changes?: N

### DIFF
--- a/config/xml_schemas/config_batch.xsd
+++ b/config/xml_schemas/config_batch.xsd
@@ -19,6 +19,7 @@
   <xs:element name="depend_string" type="xs:string"/>
   <xs:element name="depend_separator" type="xs:string"/>
   <xs:element name="walltime_format" type="xs:string"/>
+  <xs:element name="batch_jobid_env_var" type="xs:string"/>
   <xs:element name="batch_mail_flag" type="xs:string"/>
   <xs:element name="batch_mail_type_flag" type="xs:string"/>
   <xs:element name="batch_mail_type" type="xs:string"/>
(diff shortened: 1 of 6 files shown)


mfdeakin-sandia:
I got a failure from `PRE.f19_f19.ADESP.sandiatoss3_intel.fake_testing_only` in `Z_FullSystemTest` that looks like restart files were moved before they should have been? Can I ignore that? All the others in the full system test passed

config/xml_schemas/config_batch.xsd
```diff
@@ -78,6 +79,9 @@
   <!-- walltime_format: time format expected by batch system for the wall clock time field -->
         <xs:element minOccurs="0" ref="walltime_format"/>
 
+  <!-- batch_jobid_env_var: environment variable set by the batch system indicating the
+       job id for the submitted job -->
+        <xs:element minOccurs="0" ref="batch_jobid_env_var"/>
```
jedwards4b: you have minOccurs = 0 but it looks like this is now a required field?

mfdeakin-sandia: I check and verify it exists in `EnvBatch.is_submission_job`, so it shouldn't be not required

jedwards4b: okay - what about the case in which you are not using a batch system at all?

mfdeakin-sandia: TBC, I do prefer the idea of adding a command line argument to case.st_archive; no new xml and doesn't have this issue

mfdeakin-sandia: @jedwards4b Another reason to prefer the command line option - it solves the issue of RESUBMIT being set by the user during a run in the other PR I have. If it's okay, I'll just add that method to that branch, and close this PR?

Issue #2448, opened 2018-04-11
Possible race condition regarding env_build.xml during build


mnlevy1981:
This is everyone's favorite type of issue: an intermittent error that seems to be related to doing things in parallel (and so far has only been seen on cheyenne though there's no indication that the machine itself is causing problems)

**The Issue**

When running `create_test` from a compute node (with `--parallel-jobs 4 --proc-pool 36`), I will occasionally get build errors that trace back to ww3 not being able to build because of issues reading `env_build.xml`. These occur during `C` or `G` compset tests. and @klindsay28 has also run into this problem (although I am unsure of his settings for `parallel-jobs` and `proc-pool`).

@billsacks and @sherimickelson have gotten the same error, but via a different mechanism. Both of them report that trying to build on the cheyenne head-node (which is frowned upon by CISL and typically killed before completion) and then trying to build again from a compute node without running `./case.build --clean-all` in between will then generate this.

**The error**

The `wav.bldlog` will contain something like

```
ERROR: Command: '/usr/bin/xmllint --noout --schema /glade/p/work/mlevy/codes/CESM/cesm2_0_beta09+pop2_trunk/cime/config/xml_schemas/env_entry_id.xsd /glade2/scratch2/mlevy/tests/aux_pop_MARBL/cime5.4.0-alpha.23-cice5_20180123-cesm_pop_2_1_20180406/ERS.T62_g37.G1850ECO.cheyenne_intel.pop-cice_ciso.G.20180410_144957_n54xs4/env_build.xml' failed with error '/glade2/scratch2/mlevy/tests/aux_pop_MARBL/cime5.4.0-alpha.23-cice5_20180123-cesm_pop_2_1_20180406/ERS.T62_g37.G1850ECO.cheyenne_intel.pop-cice_ciso.G.20180410_144957_n54xs4/env_build.xml:1: parser error : Document is empty
```

or

```
[...]

PR #2567, opened 2018-05-09
Smarter xmlchange 

jedwards4b:
Move the flag that indicates an xml file needs to be rewritten from the case object to the individual file objects and before writing a value confirm that that value is not already the same as the file value.

Test suite: scripts_regression_tests.py (running full cesm prealpha test suite in progress)
Test baseline: cesm2_0_alpha10f
Test namelist changes: none
Test status: bit for bit
Addresses #2448 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jgfouca

### DIFF
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -77,6 +77,8 @@ class GenericXML(object):
         """
         if not self.DISABLE_CACHING and infile in self._FILEMAP and self.read_only:
             logger.debug("read (cached): " + infile)
+            expect(not self.needsrewrite, "Reading into object marked for rewrite, file {}"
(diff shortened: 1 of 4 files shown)


jedwards4b:
@jgfouca cheyenne crashed again this morning delaying the complete testing we would like to do on this PR.   However I was able to do a pretty good subset of that testing last night and I'm pretty satisfied that it's ready to go.  Your careful review would be very much appreciated.

jgfouca:
Testing complete?

jedwards4b:
One of the new changes causes FAIL: test_a_createnewcase (__main__.J_TestCreateNewcase)
I haven't figured out why yet.

jedwards4b:
Okay that problem is solved.  All testing complete.

PR #2575, opened 2018-05-11
define T85_T85_mg17

jedwards4b:
Adds T85_T85_mg17 grid

Test suite:  SMS_Ln9.T85z60_T85_mg17.FHS94.cheyenne_intel.cam-outfrq9s, SMS_Ln9.T85_T85_mg17.FHS94.cam-outfrq9s
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/config_grids.xml
+++ b/config/cesm/config_grids.xml
@@ -179,6 +179,13 @@
       <mask>gx1v7</mask>
     </model_grid>
 
+    <model_grid alias="T85_T85_mg17" not_compset="_POP">
+      <grid name="atm">T85</grid>
+      <grid name="lnd">T85</grid>
+      <grid name="ocnice">T85</grid>
+      <mask>gx1v7</mask>
+    </model_grid>
+
     <model_grid alias="T42_T42_mg16" not_compset="_POP">
       <grid name="atm">T42</grid>
       <grid name="lnd">T42</grid>


config/cesm/config_grids.xml
```diff
@@ -179,18 +172,25 @@
...
       <grid name="ocnice">T85</grid>
-      <mask>usgs</mask>
+      <mask>gx1v7</mask>
+    </model_grid>
+
+    <model_grid alias="T42_T42_mg16" not_compset="_POP">
+      <grid name="atm">T42</grid>
+      <grid name="lnd">T42</grid>
+      <grid name="ocnice">T42</grid>
+      <mask>gx1v6</mask>
```
gold2718: Is there a reason this is down here instead of with the other T42_T42 grids? When working on this file, it is very helpful (to me at least) to have these in some semblance of order.

ekluzek:
My understanding is that T42 is currently required by SCAM. I support a surface dataset at that resolution for this reason. If SCAM were no longer to need it, I'd drop support of that file in CTSM as well.

jedwards4b:
@gold2718 @cacraigucar I think that you two should look through this file and make additional changes if you think them appropriate.

cacraigucar:
SCAM does need T42 (we are currently using that in our tests), so it needs to remain

config/cesm/config_grids.xml
cacraigucar: SCAM does need T42 (we are currently using that in our tests), so it needs to remain

cacraigucar: SCAM does need T42 (we are currently using that in our tests), so it needs to remain

cacraigucar:
No other changes seen by me.  I'm happy with mods.

jedwards4b:
@gold2718 please reorder in the way that you think makes sense.

Issue #1668, opened 2017-06-09
Review and edit help documentation for user-facing scripts


billsacks:
The help documentation (what you get with `-h`) of many of our user-facing scripts could use some minor editing for grammar and whitespace, and possibly some heavier editing for general understandability.

One example is a lack of whitespace between words which arises from having python code like this:

```python
    parser.add_argument("--user-grid", action="store_true",
                        help="If set, then the -grid argument is treated as a user specified grid."
                        "As such, the additional argument --gridfile must also be specified"
                        "and all of the grid settings in the --res argument name (which must be a grid longname"
                        "have been defined in the grid file pointed to by --gridfile")
```

– note the lack of spaces before the closing quotes on each line, leading to help text that looks like this:

```
  --user-grid           If set, then the -grid argument is treated as a user
                        specified grid.As such, the additional argument
                        --gridfile must also be specifiedand all of the grid
                        settings in the --res argument name (which must be a
                        grid longnamehave been defined in the grid file
                        pointed to by --gridfile (default: False)
```

Here is a list of what I see as the most critical user-facing scripts. I suggest that we check each one off as someone goes through and edits it (or determines that it doesn't need editing):

Top-level scripts:

- [x] create_clone
- [x] create_newcase
- [x] create_test
- [x] manage_pes (edit: we're planning to remove this)
- [x] manage_testlists (edit: this is gone now)
- [x] query_config
- [x] query_testlists

Scripts created (linked) in caseroot:

- [x] archive_metadata.sh (edit: no longer part of cime)
- [x] case.build
- [x] case.cmpgen_namelists
- [x] case.qstatus
- [x] case.setup
- [x] case.submit
- [x] check_case
- [x] check_input_data (see https://github.com/ESMCI/cime/issues/1668#issuecomment-388194634)
- [x] pelayout (deferred to #2573 )
- [x] preview_namelists
- [x] preview_run
- [x] xmlchange
- [x] xmlquery

This does not include scripts in scripts/Tools/ or elsewhere; given our limited time, I'd consider those to be lower priority.

billsacks:
cc @bertinia @mvertens

mvertens:
@billsacks - thanks for pointing this out and listing out the files that
need to be reviewed.
@bertinia - which files are you willing to take on?
@billsacks - which files are you willing to take on?
I'll do the rest.

billsacks:
I took care of query_testlists via #1669 . I'm happy to take a few others - I don't care which ones.

[...]

PR #2585, opened 2018-05-13
Documentation updates, including major updates to --help for scripts

billsacks:
This joint PR between @mvertens and myself involves documentation
updates - both to the rst-based documentation and to the documentation
you get from running --help. We have reviewed and edited scripts in the
scripts/ directory and scripts that are linked in your case directory,
but this doesn't include all scripts (e.g., there are some in
scripts/Tools that we haven't reviewed / edited).

Two general changes that we made for most/all of these scripts were:
- Letting argparse generate its default usage message
- Using RawTextHelpFormatter so that we can control line breaks,
  generally using longer lines in the help output

In addition, where necessary, we have added more documentation to both
individual options and the top-level documentation message for scripts.

Test suite: scripts_regression_tests.py on cheyenne, plus code_checker
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#1668

User interface changes?: Better help text

Code review:


billsacks:
Of the scripts for which I changed the help, xmlchange could particularly use a review. e.g., do others agree with my documentation of --id, --val and --file?

scripts/Tools/xmlchange
```diff
@@ -1,13 +1,49 @@
...
- setup/build process should generally use -noecho.
-
+Allows changing variables in env_*xml files via a command-line interface.
+
+This provides two main benefits over editing the xml files by hand:
+  - Settings are checked immediately for validity
+  - Settings are echoed to the CaseStatus file, providing a "paper trail" of
+    changes made by the user.
+
+Examples:
```
billsacks: I tried to make the xmlchange help roughly parallel the help for xmlquery, though I didn't go so far as to make the help text for these two tools completely consistent (mainly because I ran out of time and energy).

scripts/Tools/xmlchange
```diff
@@ -24,64 +60,72 @@ logger = logging.getLogger("xmlchange")
...
+                        help="Ignore typing checks and store value.")
+
+    parser.add_argument("-loglevel",
+                        help="Ignored, only for backwards compatibility.")
 
     args = CIME.utils.parse_args_and_handle_standard_logging_options(args, parser)
 
     listofsettings = []
     if( len(args.listofsettings )):
+        expect(args.id is None, "Cannot specify both listofsettings and --id")
```
billsacks: Most of the changes in this PR just involve documentation, but the next few lines involve some new logic to do error-checking on arguments. It would be great if someone could double check this.

PR #2596, opened 2018-05-15
better fix for xmlquery

### LINKED ISSUE #2593: With change in #2588, xmlquery sometimes doesn't give you an error when it should
This is a minor issue, so I don't feel like it needs to be fixed for the CESM2 release, but I feel like it would eventually be good to fix this to work more robustly:

With the fix that @jedwards4b put in in #2588 , a new issue appears: If you xmlquery a non-existent variable with flag `--file`, you no longer get an error: e.g.:

```
$ ./xmlquery STOP_FOO --file env_run.xml
setting case file to env_run.xml
$ echo $?
0
```

(This still works fine - giving an error as expected - if you don't include the --file option.)

Since specifying --file in this single-variable usage of xmlquery seems uncommon, this doesn't feel critical to fix immediately.

From a quick look through the logic, I'm wondering if a better solution to #2580 would be to revert the logic on these lines to the pre-PR #2588 logic:

https://github.com/ESMCI/cime/blob/6c4b1524cd0cdab8257dc98ea6c7e6024a1ac9d9/scripts/Tools/xmlquery#L255-L258

and instead adding an option to `case.get_record_fields`, as called on this line:

https://github.com/ESMCI/cime/blob/6c4b1524cd0cdab8257dc98ea6c7e6024a1ac9d9/scripts/Tools/xmlquery#L286

to pass in an optional file; if specified, it would only query that file, rather than querying all files.

My understanding of the current --listall logic when --file is given is that it gets variables from all files, then builds up a result list containing just those variables in the given file; with the above suggestion, the subsetting on file would happen up-front. I'm not sure if this would work right, though.

jedwards4b:
Rework the --listall --file fix

Test baseline: 
Test namelist changes: 

Fixes #2593 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/Tools/xmlquery
+++ b/scripts/Tools/xmlquery
@@ -94,7 +94,7 @@ There are two usage modes:
      - You can tailor the query by adding ONE of the following possible qualifier arguments:
        [--full --fileonly --raw --description --get-group --type --valid-values]
 
-     - The env_mach_specific.xml env_archive.xml files are not supported by this tool.
+     - The env_mach_specific.xml and env_archive.xml files are not supported by this tool.
 """
 
 from standard_script_setup import *
@@ -291,7 +291,6 @@ def _main_func(description):
     with Case(caseroot) as case:
         if listall or partial_match:
             if xmlfile:
-                comp_classes = case.get_values("COMP_CLASSES")
                 case.set_file(xmlfile)
             all_variables = sorted(case.get_record_fields(None, "varid"))
             logger.debug("all_variables: {}".format(all_variables))
--- a/scripts/Tools/xmlquery
+++ b/scripts/Tools/xmlquery
@@ -94,6 +94,7 @@ There are two usage modes:
      - You can tailor the query by adding ONE of the following possible qualifier arguments:
        [--full --fileonly --raw --description --get-group --type --valid-values]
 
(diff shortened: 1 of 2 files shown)


scripts/Tools/xmlquery
```diff
@@ -280,10 +283,18 @@ def _main_func(description):
...
 
+    expect(xmlfile not in unsupported_files,
+           "XML file {} is unsupported by this tool."
+           .format(xmlfile))
+
     # Initialize case ; read in all xml files from caseroot
     with Case(caseroot) as case:
         if listall or partial_match:
+            if xmlfile:
+                comp_classes = case.get_values("COMP_CLASSES")
```
billsacks: This comp_classes variable is unused (and is causing pylint check to fail).

scripts/Tools/xmlquery
```diff
@@ -94,6 +94,7 @@ There are two usage modes:
      - You can tailor the query by adding ONE of the following possible qualifier arguments:
        [--full --fileonly --raw --description --get-group --type --valid-values]
 
+     - The env_mach_specific.xml env_archive.xml files are not supported by this tool.
```
billsacks: Please add an "and" here, changing it to:

```
     - The env_mach_specific.xml and env_archive.xml files are not supported by this tool.
```

Issue #2599, opened 2018-05-15
system_tests_compare_two needs to fail if phase2 doesn't run


jgfouca:

billsacks:
@jgfouca see also #399 which I just reopened - but I'm fine with a different solution to the problem than the solution proposed there.

PR #2600, opened 2018-05-15
multisubmit tests were not resubmitting.

### LINKED ISSUE #2597: PET test (and maybe others) are acting like smoke tests
From cime5.6.6, a PET test (e.g., PET_P36x2_D.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_gnu.clm-default but I expect you'd see this with any PET test) is acting like a smoke test: It only runs the first case, doesn't run the second, and doesn't do any comparison between the cases. You can see this by seeing the empty rundir from case2, the lack of a run in case2's CaseStatus, and the lack of a compare phase in TestStatus.

My suspicion is that something is broken with the multisubmit mode of system_tests_compare_two.

jedwards4b:
Tests with the multisubmit flag were not submitting the second phase.   Issue #2599 will be addressed separately. 

Test suite: scripts_regression_tests.py , PET_P36x2_Ln9.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_gnu.clm-default 
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #2597 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case/case_st_archive.py
+++ b/scripts/lib/CIME/case/case_st_archive.py
@@ -630,10 +630,7 @@ def case_st_archive(self, last_date_str=None, archive_incomplete_logs=True, copy
 
     # resubmit case if appropriate
     resubmit_cnt = self.get_value("RESUBMIT")
-    logger.info("resubmit_cnt {} resubmit {}".format(resubmit_cnt, resubmit))
-    if not resubmit:
-        import traceback
-        traceback.print_stack()
+    logger.debug("resubmit_cnt {} resubmit {}".format(resubmit_cnt, resubmit))
     if resubmit_cnt > 0 and resubmit:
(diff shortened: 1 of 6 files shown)


scripts/lib/CIME/case/case_st_archive.py
```diff
@@ -630,6 +630,10 @@ def case_st_archive(self, last_date_str=None, archive_incomplete_logs=True, copy
 
     # resubmit case if appropriate
     resubmit_cnt = self.get_value("RESUBMIT")
+    logger.info("resubmit_cnt {} resubmit {}".format(resubmit_cnt, resubmit))
+    if not resubmit:
```
jgfouca: Is this needed?

Issue #2598, opened 2018-05-15
datm doesn't restart properly with variable year orbit


billsacks:
In a case that crosses the year boundary with variable year orbit, datm doesn't restart properly.

See also https://github.com/ESCOMP/ctsm/issues/379

I'm working on a fix, with help from @mvertens

billsacks:
This is now fixed on the 5.6 maintenance branch, but not yet on master.

PR #2602, opened 2018-05-16
In datm, set orbital parameters in run loop

billsacks:
This is needed for datm to restart properly when using variable-year
orbital parameters.

Test suite:
- These tests now pass in CTSM tag clm5.0.dev010 if I put in a hack in
  CTSM to fix a problem there:
  - ERS_D_P36x1.f10_f10_musgs.IHistClm50Bgc.cheyenne_intel.clm-decStart
  - ERP_D_P48x1.f10_f10_musgs.IHistClm50Bgc.hobart_nag.clm-decStart
- These tests are bit-for-bit with out-of-the-box clm5.0.dev010
  (confirming that these changes haven't completely broken datm; note
  that this includes a selection of 1850, 2000 and HIST tests, but
  deliberately does not include HIST tests that cross the year boundary):
  - ERP_D_P36x2_Ld3.f10_f10_musgs.I1850Clm50BgcCrop.cheyenne_intel.clm-default1850
  - ERP_D_Ld3_P36x2.f10_f10_musgs.I2000Clm50BgcCruGs.cheyenne_intel.clm-default
  - ERP_D_Ld5.f10_f10_musgs.I2000Clm50Sp.cheyenne_intel.clm-decStart
  - ERS_D_Ld10.f10_f10_musgs.IHistClm50SpG.cheyenne_intel.clm-glcMEC_decrease

Test baseline: clm5.0.dev010
Test namelist changes: none
Test status: potentially climate-changing for I compsets that use
variable_year orbital parameters (currently HIST compsets), and any
other compsets using datm with solar zenith angle-based interpolation (I
*think* that's the criteria determining if these changes change
answers): Previously, datm looked at orbital parameters at the start of
the run, and used those orbital parameters for its zenith angle-based
interpolations throughout the entire run (though re-checking this on
each restart). Now, datm re-checks the orbital parameters every time
step, so the zenith angle-based interpolations are kept in sync with the
current orbital parameters.

Fixes ESMCI/cime#2598

User interface changes?: none

Code review:

### DIFF
--- a/src/components/data_comps/datm/datm_comp_mod.F90
+++ b/src/components/data_comps/datm/datm_comp_mod.F90
@@ -312,9 +312,6 @@ CONTAINS
        k = mct_aVect_indexRA(SDATM%grid%data,'frac')
        SDATM%grid%data%rAttr(k,:) = 1.0_R8
 
-       !--- set data needed for cosz t-interp method ---
-       call shr_strdata_setOrbs(SDATM,orbEccen,orbMvelpp,orbLambm0,orbObliqr,idt)
-
        if (my_task == master_task) then
           call shr_strdata_print(SDATM,'ATM data')
        endif
@@ -551,9 +548,24 @@ CONTAINS
     !----------------------------------------------------------------------------
 
     call t_adj_detailf(+2)
-    call datm_comp_run(EClock, x2a, a2x, &
-         SDATM, gsmap, ggrid, mpicom, compid, my_task, master_task, &
-         inst_suffix, logunit, nextsw_cday)
+    call datm_comp_run( &
+         EClock = EClock, &
+         x2a = x2a, &
+         a2x = a2x, &
+         SDATM = SDATM, &
+         gsmap = gsmap, &
+         ggrid = ggrid, &
+         mpicom = mpicom, &
+         compid = compid, &
+         my_task = my_task, &
+         master_task = master_task, &
+         inst_suffix = inst_suffix, &
+         logunit = logunit, &
+         orbEccen = orbEccen, &
+         orbMvelpp = orbMvelpp, &
+         orbLambm0 = orbLambm0, &
+         orbObliqr = orbObliqr, &
+         nextsw_cday = nextsw_cday)
     call t_adj_detailf(-2)
 
     call t_stopf('DATM_INIT')
@@ -563,7 +575,9 @@ CONTAINS
   !===============================================================================
   subroutine datm_comp_run(EClock, x2a, a2x, &
        SDATM, gsmap, ggrid, mpicom, compid, my_task, master_task, &
-       inst_suffix, logunit, nextsw_cday, case_name)
+       inst_suffix, logunit, &
+       orbEccen, orbMvelpp, orbLambm0, orbObliqr, &
+       nextsw_cday, case_name)
 
     ! !DESCRIPTION: run method for datm model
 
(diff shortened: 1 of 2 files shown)


billsacks:
@rljacob - I'm adding you as a reviewer in case you want to look at this, because I think it might affect e3sm, too.

@mvertens - In addition to reviewing the diffs, can you also confirm my statement in the top comment about when this will change answers?

@ekluzek also bringing you into the loop

rljacob:
Will it change answers for **any** case using datm?

src/components/data_comps/datm/datm_comp_mod.F90
```diff
@@ -312,9 +312,6 @@ subroutine datm_comp_init(Eclock, x2a, a2x, &
        k = mct_aVect_indexRA(SDATM%grid%data,'frac')
        SDATM%grid%data%rAttr(k,:) = 1.0_R8
 
-       !--- set data needed for cosz t-interp method ---
-       call shr_strdata_setOrbs(SDATM,orbEccen,orbMvelpp,orbLambm0,orbObliqr,idt)
```
billsacks: Note that I removed the setOrbs call from datm_comp_init. There is now a setOrbs call in datm_comp_run, and datm_comp_run is called from near the end of datm_comp_init, so this seemed like a safe thing to do. But I wasn't 100% confident in this being safe, which was part of my motivation for running a handful of tests to confirm I hadn't changed answers.

If others feel this removal is potentially problematic, I can put it back in.

billsacks:
> Will it change answers for any case using datm?

For many cases, it will not change answers. First, it should only change answers for cases with variable_year orbital parameters. Second, my understanding (though I'd like confirmation from @mvertens ) is that only select datm modes have logic that relies on the orbital parameters (primarily the modes used for forcing a land model, I think - not the modes for forcing an ocean model, at least in CESM).

@rljacob and @mvertens do you feel that we should do more testing to more definitively characterize when this changes answers?

billsacks:
@rljacob - a bit more about when/why it changes answers (I'll also add this in the big comment at the top): Previously, datm looked at orbital parameters at the start of the run, and used those orbital parameters for its zenith angle-based interpolations throughout the entire run (though re-checking this on each restart). Now, datm re-checks the orbital parameters every time step, so the zenith angle-based interpolations are kept in sync with the current orbital parameters.

billsacks:
@rljacob - I think you're right about this:

> I don't think we have any tests with varying orbital params.

I missed the fact that #2561 only affects cesm.

> Do you want this in the release? I think it needs to be merged to your new maintenance branch.

Yes - this should come to master and the maintenance branch. @jedwards4b - I put the maint-5.6_bugfix tag on this; let me know if there's anything else that needs to be done for this to happen.

billsacks:
Here are two tests that give answers that differ from baseline, as expected:

```
ERP_Ly3_P72x2.f10_f10_musgs.IHistClm50BgcCrop.cheyenne_intel.clm-cropMonthOutput
SMS_Ld5.f19_g17.IHistClm50Bgc.cheyenne_intel.clm-decStart
```

Note that both of these have orb_mode = variable_year and cross a year boundary (the decStart test starts Dec 30 and ends in early January).

rljacob:
In that case the target for this PR should be the maintenance branch.

billsacks:
Okay, thanks - just changed the base to the maintenance branch.

rljacob:
Just merge the things you want in the release to the maint-5.6 branch and when you're done, merge the maint branch to master.

Never merge master to maint.

rljacob:
Maybe the maint branch was made too early?  You don't really need that branch until there's a PR for master that you *don't* want in the release.

jedwards4b:
@jgfouca wanted to bring in an e3sm merge, which is why the maint branch was made.

rljacob:
But the PR for that isn't even up yet.

jedwards4b:
True but I don't see any point in undoing what's done.

rljacob:
Sure its just more hassle.   There has been at least one commit to master after the maint-5.6 banch point so its also important that feature branches intended for maint-5.6 start either at the branch point or from the maint-5.6 branch itself.

rljacob:
Hang on:  it looks like the maint-5.6 branch does not start from master.  That's a problem.  Lets take this to Slack.

billsacks:
@mvertens I can hit merge if you feel this is good to go... but can you first confirm that the statement I made about when answers will change is correct?

mvertens:
@billsacks - your statement about when answers will change is correct.

PR #2633, opened 2018-05-30
add test_release category as tested in cesm

jedwards4b:
Add test_release as a tested category for cesm 

Test suite: ./create_newcase --case foo --compset FDABIP04 --res f09_f09_mg17
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


jedwards4b:
It was determined that this isn't needed.

PR #2635, opened 2018-05-31
Improve robustness of compare-two tests

### LINKED ISSUE #399: Want clearer indication of what aspects of a test have not yet finished
This issue began discussion in the comments for https://github.com/ESMCI/cime/pull/392 with my proposal for a requirement:

**If there is a TestStatus file in the test directory, and if the test has not completed successfully, then there must be at least one line in the TestStatus file with a status other than PASS**

@mvertens agrees with this proposal, and we both like the solution proposed by @gold2718 

> When a test is created, it should know all the phases that are part of the test.... At the end of the init, I would like to see the self.phase_statuses be filled out with each phase marked INCOMPLETE (with INCOMPLETE added to the list of valid test statuses.

(See discussion in https://github.com/ESMCI/cime/pull/392 for details.)

### LINKED ISSUE #2599: system_tests_compare_two needs to fail if phase2 doesn't run

jgfouca:
1. Preemptively add a PEND for the comparison between the two runs. This should prevent relapsing into a situation where PET silently turns into SMS.
2. Multi-submit tests were broken on non-batching machines. Made changes so that no-batch runs will never be multisubmit runs.

Test suite: scripts_regression_tests (on both batch and non-batch machines)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2599 
Fixes #399 

User interface changes?: N

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -208,6 +208,8 @@ class SystemTestsCommon(object):
         stop_option = self._case.get_value("STOP_OPTION")
         run_type    = self._case.get_value("RUN_TYPE")
         rundir      = self._case.get_value("RUNDIR")
+        is_batch    = self._case.get_value("BATCH_SYSTEM") != "none"
+
         # remove any cprnc output leftover from previous runs
         for compout in glob.iglob(os.path.join(rundir,"*.cprnc.out")):
             os.remove(compout)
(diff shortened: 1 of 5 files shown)


jgfouca:
Still testing, but this is ready to review.

jgfouca:
Thanks @billsacks . I'm going to go ahead and merge this since we really want this in E3SM and I'm about to integrate. @jedwards4b , review at your leisure.

jedwards4b:
Should we add this to the maint-5.6 branch?

jgfouca:
@jedwards4b , this will mostly protect us from future regressions. As long as we know PET etc are working on the branch, I don't think we need to add this.

jgfouca:
Although the fix for non-batch systems might be nice to have.

PR #2656, opened 2018-06-07
Archive test refinement

jedwards4b:
This improves the st_archiver testing by creating a test that works for the specific case at hand instead of the general test that was there before 

Test suite: scripts_regression_tests.py (currently in progress)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: --test becomes --test-all and --test-case is added

Update gh-pages html (Y/N)?:

Code review:


jgfouca:
@jedwards4b , so we went every system test to always do an st-archiver test?

jedwards4b:
yes, it's very fast and assures that the archiver will work for that compset.

PR #2663, opened 2018-06-12
E3sm st archive

### LINKED ISSUE #2659: PR #2655 has caused some cime_developer tests to fail on anlworkstation during phase SETUP
The errors can be reproduced on anlworkstation by running a single cime_developer test:
`./create_test SMS.T42_T42.S --machine=anlworkstation --test-root=****`

Errors were:
`ERROR: The file config/e3sm/machines/config_compilers.xml has bad $VAR references. Check for circular references or variables that are used in a $VAR but not actually defined.`

This error is from the following append line:
`<append> $SHELL{$NETCDF_PATH/bin/nf-config --flibs} -lblas -llapack</append>`
If we use $(NETCDF_PATH) instead of $NETCDF_PATH this error is gone.

Not sure if this is only reproducible for anlworkstation. Some other machines (e.g. anvil, blues, melvin) in config_compilers.xml use exactly the same append line, so they might have similar issues during phase setup.

jedwards4b:
The ESP component is not used by e3sm and should not be tested in the case.st_archive tests
I also needed to update the e3sm modules for machine edison so that I could complete testing.

Test baseline: 
Test namelist changes: 
Test status: roundoff 

Fixes #2659 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/e3sm/machines/config_machines.xml
+++ b/config/e3sm/machines/config_machines.xml
@@ -133,51 +133,45 @@
       <command name="load">cray-mpich/7.6.0</command>
     </modules>
 
-    <modules compiler="intel">
-      <command name="load">PrgEnv-intel/6.0.4</command>
-      <command name="rm">intel</command>
-      <command name="load">intel/17.0.2.174</command>
-      <command name="rm">cray-libsci</command>
-    </modules>
-
-    <modules compiler="intel18">
-      <command name="load">PrgEnv-intel/6.0.4</command>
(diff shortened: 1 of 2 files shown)


jedwards4b:
I'm backing out the e3sm/machines changes from this PR - I'm just making the changes locally so that I can test the st_archive fix.  Please note that the e3sm port to machine edison needs to be updated.

PR #2674, opened 2018-06-15
Cmake build

jedwards4b:
I've handled some of the hardcoded paths here and am working on building case SMS.f19_g16.X
I'll place further comments inline.  I am able to build a SMS.f19_g17.S test with these changes. 
I now have X components building.   There are some problems in the current Macros.cmake file that we need to work on.  For example the proper flags for the gptl library are there but they are only used if an env variable MODEL is set to gptl - this works for the make style build, but for cmake we need to figure out another way to do this.   We also need to figure out how to support SourceMods in the case directory.


scripts/cmake/includedCMakeLists.txt
```diff
@@ -3,21 +3,17 @@
 # the top level of CIME or the top level of ACME. The required
 # CIME_REL_DIR variable is set in CIME to be "." and in ACME
 # to be "cime".  Except for this misdirection, this is meant
-# to be the full-functioning top-level CMakeLists.txt file 
+# to be the full-functioning top-level CMakeLists.txt file
```
jedwards4b: Now using this file as a template and copying it to the case bld directory in the case.setup phase.

scripts/cmake/includedCMakeLists.txt
```diff
@@ -3,21 +3,17 @@
...
 
-
-message("Hardwiring include path -- must automate")
-include(/home/asteyer/cime/src/CMake/CIME_initial_setup.cmake)
-
-#include(/home/agsalin/buildmacrocime/Macros.cmake)
-
 PROJECT(E3SM Fortran C)
-message("XXX ${NetCDF_PATH}")
+message("XXX ${NETCDF_PATH}")
```
jedwards4b: Including the Macros.cmake file gives many of the variables you are hardcoding here.

jedwards4b:
NOW passing all cime_developer tests.

jedwards4b:
@asteyer I am concerned that you are pushing conflicting changes to your branch without reviewing or commenting on this PR.

PR #2680, opened 2018-06-19
Use ConfigParser rather than SafeConfigParser

### LINKED ISSUE #2679: case.submit fails when some arguments have '%[A-Za-z]' in them
case.submit previously allowed substituting things like the job id into strings, such as the batch job id. This now causes a failure.
For an example, see E3SM-Project/E3SM#2395

mfdeakin-sandia:
This enables submitting jobs with batch arguments which substitute things like the job id into their strings. For an example, see E3SM-Project/E3SM#2395

Test suite: scripts_regression_tests B_CheckCode and Z_FullSystemTest
Test status: BFB

Fixes #2679

Code review: @jedwards4b

### DIFF
--- a/scripts/Tools/case.submit
+++ b/scripts/Tools/case.submit
@@ -75,7 +75,7 @@ def parse_command_line(args, description):
             args.mail_type, args.batch_args)
 
 ###############################################################################
-def _main_func(description):
+def _main_func(description, test_args=False):
 ###############################################################################
     caseroot, job, no_batch, prereq, allow_fail, resubmit, resubmit_immediate, skip_pnl, \
         mail_user, mail_type, batch_args = parse_command_line(sys.argv, description)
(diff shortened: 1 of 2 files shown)


jedwards4b:
yes but is that really a problem - we have a flag on several systems

rljacob:
I'm fine with the edison default which, in E3SM, is to put the batch output in $CASE.run.$JOBID in the casedir.  The user  wants to change the prefix, remove the "run", and put it in a sub-directory.  That shouldn't be forced on all users.

jedwards4b:
okay they can do that by changing it in their case env_batch.xml file.   I guess I can live with this change if you insist, but it really doesn't seem necessary

rljacob:
Its necessary because it restores a functionality that used to work and people were using.

mfdeakin-sandia:
@jedwards4b I think this is ready now

jgfouca:
@mfdeakin-sandia , I'm ready to do a merge back to e3sm. Do you want to merge this first?

mfdeakin-sandia:
Yes; I'm also bringing the hacky LSF fix in if you don't mind waiting a bit

PR #2683, opened 2018-06-21
fix issue with e3sm mali file archiving

jedwards4b:
Fix the e3sm case.st_archive --test-all option by adding component mali to the list of mpas style 
non-conforming file names.   

Test suite: byhand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


jgfouca:
Thx!

PR #2693, opened 2018-06-25
Make docs more consistent

billsacks:
Three small changes to create_newcase --help and sphinx-based
documentation:

(1) Remove square brackets around command-line arguments in sphinx-based
    documentation of create_newcase command.

   Square brackets usually suggest that an argument is optional. Here the
   argument is required. Removing the square brackets makes the
   documentation more consistent with the output from
   `./create_newcase --help`.

(2) Make the help for create_newcase use GRID instead of RES

   This makes the help output consistent with our documentation

(3) Refer to the --case argument as CASEROOT rather than CASE

   Elsewhere in the documentation, we distinguish between CASEROOT and
   CASE. So it is confusing that we were labeling the --case argument as
   CASE, when really this was giving CASEROOT.

   Also add a bit more documentation about this argument.

Test suite: code_checker on create_newcase; manually checked
   create_newcase --help
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

Fixes none

User interface changes?: none

   I can do this update if this PR is approved

Code review:

### DIFF
--- a/doc/source/users_guide/create-a-case.rst
+++ b/doc/source/users_guide/create-a-case.rst
@@ -38,7 +38,7 @@ CIME supports out-of-the-box *component sets*, *model grids* and *hardware platf
 
 ``CASEROOT`` can be a full path specifying where the case should be created; the name of
 the case will be the last component of the path. If not a full path, then the case is
-created under the current working directory.
+created in a location relative to the current working directory.
 
 ======================================
 Results of calling **create_newcase**
--- a/doc/source/users_guide/create-a-case.rst
+++ b/doc/source/users_guide/create-a-case.rst
@@ -26,19 +26,23 @@ See the options for `create_newcase  <../Tools_user/create_newcase.html>`_ in th
 
 The only required arguments to `create_newcase  <../Tools_user/create_newcase.html>`_ are::
 
-  > create_newcase --case CASEROOT --compset COMPSET --res GRID
+  > create_newcase --case CASENAME --compset COMPSET --res GRID
 
 Creating a CIME experiment or *case* requires, at a minimum, specifying a compset and a model grid and a case directory.
 CIME supports out-of-the-box *component sets*, *model grids* and *hardware platforms* (machines).
 
 .. warning::
-   The ``CASEROOT`` argument must be a string and may not contain any of the following special characters
+   The ``--case`` argument must be a string and may not contain any of the following special characters
    ::
       > + * ? < > { } [ ] ~ ` @ :
 
-``CASEROOT`` can be a full path specifying where the case should be created; the name of
-the case will be the last component of the path. If not a full path, then the case is
-created in a location relative to the current working directory.
+The ``--case`` argument is used to define the name of your case, a very important piece of
+metadata that will be used in filenames, internal metadata and directory paths. The
+``CASEROOT`` is a directory create_newcase will create with the same name as the
+``CASENAME``. If ``CASENAME`` is simply a name (not a path), ``CASEROOT`` is created in
+the directory where you execute create_newcase. If ``CASENAME`` is a relative or absolute
+path, ``CASEROOT`` is created there, and the name of the case will be the last component
+of the path.
 
 ======================================
 Results of calling **create_newcase**
@@ -53,8 +57,8 @@ Here, $CIMEROOT is the full pathname of the root directory of the CIME distribut
 In the example, the command creates a ``$CASEROOT`` directory: ``~/cime/example1``.
 If that directory already exists, a warning is printed and the command aborts.
 
-In the argument to ``--case``, the directory path is ignored and only the string after the
-last slash is used as the case name --- so here the case name is ``example1``.
+In the argument to ``--case``, the case name is taken from the string after the last slash
+--- so here the case name is ``example1``.
 
 The output from create_newcase includes information such as.
 
--- a/doc/source/users_guide/create-a-case.rst
+++ b/doc/source/users_guide/create-a-case.rst
@@ -26,13 +26,13 @@ See the options for `create_newcase  <../Tools_user/create_newcase.html>`_ in th
 
 The only required arguments to `create_newcase  <../Tools_user/create_newcase.html>`_ are::
 
-  > create_newcase --case [CASE] --compset [COMPSET] --res [GRID]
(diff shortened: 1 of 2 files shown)


doc/source/users_guide/create-a-case.rst
```diff
@@ -26,15 +26,19 @@ See the options for `create_newcase  <../Tools_user/create_newcase.html>`_ in th
...
 .. warning::
-   The [CASE] argument must be a string and may not contain any of the following special characters
+   The ``CASEROOT`` argument must be a string and may not contain any of the following special characters
    ::
-      > + * ? < > / { } [ ] ~ ` @ :
+      > + * ? < > { } [ ] ~ ` @ :
+
+``CASEROOT`` can be a full path specifying where the case should be created; the name of
+the case will be the last component of the path. If not a full path, then the case is
+created under the current working directory.
```
jedwards4b: If not a full path than the case is created in a location relative to the current working directory.

rljacob: I think this argument should be CASENAME.  You don't have to specify a directory path at all and CIME will form the CASEROOT from the CASENAME.   So the truly required part of this argument is the CASENAME.

rljacob: Also the --output-root argument lets you set CASEROOT to something that doesn't even have the CASENAME in it.  So CASENAME is the argument we are asking for.

billsacks: @rljacob - starting with your second comment:

> Also the --output-root argument lets you set CASEROOT to something that doesn't even have the CASENAME in it. So CASENAME is the argument we are asking for.

`--output-root` is something different: that specifies the location where the bld and run directories will go (in `$CIME_OUTPUT_ROOT/$CASE/bld` and `$CIME_OUTPUT_ROOT/$CASE/run`, where the `--output-root` argument sets the value of `$CIME_OUTPUT_ROOT`).

Now back to your first comment:

> I think this argument should be CASENAME. You don't have to specify a directory path at all and CIME will form the CASEROOT from the CASENAME. So the truly required part of this argument is the CASENAME.

While I like `CASENAME` better than `CASE`, I think this would add more confusion than it removes: Both `CASE` and `CASEROOT` are xml variables that can be queried; `CASENAME` is not. So in trying to explain the relationships between these different variables, we'd need to explain that we're using `CASENAME` here, but that becomes equivalent to `CASE` as used elsewhere. I think that's confusing, and we should only change the documentation to `CASENAME` if we also change the xml variable from `CASE` to `CASENAME`.

As for why I changed this from `CASE` to `CASEROOT`: I'll admit I felt torn here: I felt a tension between what's simple and easy to understand (`CASE`) vs. what's complete and precise (`CASEROOT`). The real trouble comes if you try to explain the relationships between (1) the `--case` argument to `create_newcase`, (2) `CASE` and (3) `CASEROOT`. We try to explain that relationship in the CESM Quick Start (http://esmci.github.io/cime/users_guide/create-a-case.html – though I'm reworking the details). I can't see a sane way to explain that relationship if you pretend that the argument to `--case` is `CASE` (or `CASENAME`).

One solution would be to simplify the documentation in the CESM Quick Start, pretending for the sake of that documentation that you always want to create the case in the current working directory, and so pretending that the `--case` argument is just the case name. I'd be okay doing that in the quick start guide, but I'm less comfortable pretending that in the documentation for `create_newcase` - where I think completeness and precision is more important... and if we use `CASE` in the quick start but `CASEROOT` in `create_newcase --help`, then the two would be inconsistent, which I don't like.

rljacob: @billsacks Sorry I was thinking of the --script-root option which doesn't seem to be in the --help.

Yes its unfortunate that CASE in the xml is not called CASENAME.   In the documentation, we use "case" to refer to the general concept of what you get when you combine a specific compset, resolution, machine and compiler (see 1.2).  So it would also be confusing to have "case" and "CASE".

But the CASEROOT, in the xml, may actually be different from what you supply to --case if you don't use a full path.  So its just not accurate to say --case *is* the CASEROOT.  It is used to create the CASEROOT.  And maybe not if you use --script-root.

I like your idea that the quick-start only talks about --case as the casename.

What's wrong with this:
"The --case argument is used to define the name of your case, a very important piece of metadata that will be used in filenames, internal metadata and directory paths.  The CASEROOT is a directory create_newcase will create with the same name as the CASENAME.  By default, this directory is placed in the  directory where you execute create_newcase.  If you prepend the casename with a path, the CASEROOT will be placed there.   If you don't want the CASENAME in the CASEROOT, use the --script-root option to specify the name of the CASEROOT."

billsacks: @rljacob Thanks – I'll try to rework things as per your suggestion, both in this cime documentation and in the CESM quick start. I do have a couple of minor quibbles with some of your rationale, though:

First, it looks like it's intentional that `--script-root` isn't in the help: there's a comment in the code saying that this option has been deprecated (presumably because you can get the same effect by giving a full path in your `--case` argument?).

Second, we already have a couple of references to CASE in the CIME documentation (in running-a-case.rst and troubleshooting.rst).

Third, assuming you don't use the deprecated `--script-root` (which I figure we don't have to acknowledge because it's deprecated), CASEROOT is simply given by os.path.abspath(casename) – which to me says that those two are effectively the same.

All that said, I'm fine documenting this as you suggest, and I'll give that a shot.

billsacks:
I just pushed a commit that follows @rljacob 's suggestions. @rljacob - I didn't use your exact wording, but tried for something similar. (I found the wording, "if you prepend the casename with a path" to be confusing, so I reworded that a bit.)

@rljacob please let me know if this looks good to you.

Others may want to look back at this, too, since it's pretty completely rewritten from my last edits.

billsacks:
@mvertens do you want to look at this still, or should I go ahead and merge it given the 4 other sets of eyes on it?

billsacks:
@mvertens just gave her approval via email, so I'm merging it.

billsacks:
I just updated the html by following the instructions in the README and https://github.com/ESMCI/cime/wiki/Working-with-Sphinx-and-reStructuredText. I don't think I've done that before, so let me know if you notice anything funny.

PR #2703, opened 2018-07-06
More fixes and tests for V2 compiler macro system

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Code review: @jedwards4b (pre-approved)

### DIFF
--- a/scripts/lib/CIME/XML/compilerblock.py
+++ b/scripts/lib/CIME/XML/compilerblock.py
@@ -190,7 +190,7 @@ class CompilerBlock(object):
         # Create the setting object.
         append = self._db.name(elem) == "append" or (self._db.name(elem) == "base" and self._compiler and self._db.compiler != self._compiler)
         setting = ValueSetting(value_text, append,
-                               conditions, set_up, tear_down)
+                               conditions, set_up, tear_down, force_no_append=self._db.name(elem) == "base")
         return (setting, depends)
 
     def _add_elem_to_lists(self, name, elem, value_lists):
--- a/scripts/lib/CIME/XML/compilerblock.py
+++ b/scripts/lib/CIME/XML/compilerblock.py
@@ -191,7 +191,6 @@ class CompilerBlock(object):
         append = self._db.name(elem) == "append" or (self._db.name(elem) == "base" and self._compiler and self._db.compiler != self._compiler)
         setting = ValueSetting(value_text, append,
                                conditions, set_up, tear_down)
-
         return (setting, depends)
 
     def _add_elem_to_lists(self, name, elem, value_lists):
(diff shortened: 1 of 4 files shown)


quantheory:
These changes seem fine to me, although I'm not sure I entirely understand the reason why they are necessary. This code has changed significantly (and my memory has not improved) since the last time I interacted with it.

scripts/lib/CIME/XML/compilerblock.py
```diff
@@ -190,8 +190,7 @@ def _elem_to_setting(self, elem):
         # Create the setting object.
         append = self._db.name(elem) == "append" or (self._db.name(elem) == "base" and self._compiler and self._db.compiler != self._compiler)
```
quantheory: I know that this `force_no_append` stuff was already merged, but I've decided that I really don't like it. I think that the real problem is this line, which makes no sense at all. As far as I can tell, this line is trying to get some `<base>` tags to be treated as `<append>` tags in order to trick an ambiguity checker down the line. But when they were ending up actually being appended we were forced to put in this `force_no_append` bit to work around that.

The right answer would be to change this line to simply:

    append = self._db.name(elem) == "append"

Then remove the `force_no_append` flag, and if that causes problems with the ambiguity checker, then that code should be patched properly!

scripts/lib/CIME/XML/compilerblock.py
```diff
@@ -208,8 +207,8 @@ def _add_elem_to_lists(self, name, elem, value_lists):
             value_lists[name] = PossibleValues(name, setting,
                                                self._specificity, depends)
         else:
-            value_lists[name].add_setting(setting, self._specificity,
-                                          depends)
+            specificity = 0 if len(elem.xml_element.attrib) else self._specificity
```
quantheory: Again, this was already merged, but I'm not liking this at all upon second glance. First, this is not a bug fix, but a new behavior (and one that I don't even understand the purpose of). It should probably have a test associated with it. Secondly, the order in which settings are specified in config_compilers.xml is not supposed to matter. But because this line is inside this conditional, it will never affect the first time a particular setting is specified in a block.

quantheory: I take back what I said about the testing, since I notice that the new test actually does cover this. But I will note that this XML can now fail with an ambiguity error:

```
<compiler COMPILER="intel">
  <MPIFC MPILIB="openmpi">mpifc</MPIFC>
</compiler>
<compiler MACH="mymach" COMPILER="intel">
  <MPIFC MPILIB="openmpi">mpif90</MPIFC>
</compiler>
```

Also, my comment about the order of the blocks mattering stands.

jedwards4b:
@quantheory I don't like this code either, you are certainly welcome to contribute code that produces the correct behavior and passes all tests, however I don't really appreciate vague comments about how things should be without providing solutions to the problems that caused the changes to be necessary.

PR #2710, opened 2018-07-11
Rework machine overrides in config_compilers.

quantheory:
In the original design, machine-specific settings would completely
override all non-machine-specific settings. This led to two unintuitive
behaviors in particular:
 - A machine-specific setting that does *not* apply could override one
   that does (e.g. because the machine-specific setting only applies to
   cases with a particular MPI library, whereas the non-specific setting
   applies to all cases).
 - A machine-specific `<base>` tag causes all non-machine-specific
   `<append>` tags to be ignored.

This commit changes both behaviors. Now machine-specific settings will
only override non-specific settings if they actually apply to the
current case. Furthermore, `<append>` tags can never be overridden,
regardless of the `<base>` tags that are involved.

While this is expected to have minimal effect on existing
config_compilers.xml settings, it is quite possible, even likely, that a
small number of changes will have to be made for some machines.

Test suite: scripts_regression_tests, Python doctests for valuesetting and possiblevalues
Test baseline: N/A
Test namelist changes: None
Test status: N/A, but no reason this shouldn't be bit for bit

This fixes some bugs that @jgfouca encountered and others that I noticed after #2703, though there's no open bug report for those.

User interface changes?: N

Code review: 
@jedwards4b when he gets back, if he wants?

### DIFF
--- a/scripts/lib/CIME/BuildTools/macroconditiontree.py
+++ b/scripts/lib/CIME/BuildTools/macroconditiontree.py
@@ -53,7 +53,7 @@ class MacroConditionTree(object): # pylint: disable=too-many-instance-attributes
                         "found after the ambiguity check was complete, " \
                         "or there is a mixture of appending and initial " \
                         "settings in the condition tree."
-                self._assignments.append((name, setting.value, setting.force_no_append))
+                self._assignments.append((name, setting.value))
                 self._set_up += setting.set_up
                 self._tear_down += setting.tear_down
         else:
@@ -138,8 +138,8 @@ class MacroConditionTree(object): # pylint: disable=too-many-instance-attributes
         if self._is_leaf:
             for line in self._set_up:
                 writer.write_line(line)
-            for (name, value, force_no_append) in self._assignments:
-                if self._do_append and not force_no_append:
+            for (name, value) in self._assignments:
+                if self._do_append:
                     writer.append_variable(name, value)
                 else:
                     writer.set_variable(name, value)
--- a/scripts/lib/CIME/BuildTools/possiblevalues.py
+++ b/scripts/lib/CIME/BuildTools/possiblevalues.py
@@ -22,12 +22,11 @@ class PossibleValues(object):
     append_settings - A dictionary of lists of possible appending settings
                       for the variable, with the specificity of each list
                       as the associated dictionary key.
-    depends - The current list of variables that this variable depends on
-              to get its value.
 
     Public methods:
     add_setting
     ambiguity_check
+    dependencies
     to_cond_trees
     """
 
@@ -38,17 +37,20 @@ class PossibleValues(object):
         arguments are the same as for append_match.
         """
         self.name = name
-        self.depends = depends
         # If this is an appending setting, its specificity can't cause it
-        # to overwrite other settings, but we want to keep track of it.
+        # to overwrite other settings.
         if setting.do_append:
             self.settings = []
-            self.append_settings = {specificity: [setting]}
-            self._specificity = 0
+            self.append_settings = [setting]
+            self._specificities = []
+            self._depends = []
+            self._append_depends = depends
         else:
             self.settings = [setting]
-            self.append_settings = {}
-            self._specificity = specificity
+            self.append_settings = []
+            self._specificities = [specificity]
+            self._depends = [depends]
+            self._append_depends = set()
 
     def add_setting(self, setting, specificity, depends):
         """Add a possible value for a variable.
@@ -56,48 +58,53 @@ class PossibleValues(object):
         Arguments:
         setting - A ValueSetting to start the list.
         specificity - An integer representing how specific the setting is.
-                      Only the initial settings with the highest
-                      specificity and appending settings with at least that
-                      specificity will actually be kept in the list. The
-                      lowest allowed specificity is 0.
+                      Low-specificity settings that will never be used will be
+                      dropped from the list. The lowest allowed specificity is
+                      0.
         depends - A set of variable names, specifying the variables that
                   have to be set before this setting can be used (e.g. if
                   SLIBS refers to NETCDF_PATH, then NETCDF_PATH has to be
                   set first).
 
+        >>> from CIME.BuildTools.valuesetting import ValueSetting
         >>> a = ValueSetting('foo', False, dict(), [], [])
         >>> b = ValueSetting('bar', False, dict(), [], [])
         >>> vals = PossibleValues('var', a, 0, {'dep1'})
         >>> vals.add_setting(b, 1, {'dep2'})
         >>> a not in vals.settings and b in vals.settings
         True
-        >>> 'dep1' not in vals.depends and 'dep2' in vals.depends
+        >>> 'dep1' not in vals.dependencies() and 'dep2' in vals.dependencies()
         True
         >>> vals.add_setting(a, 1, {'dep1'})
         >>> a in vals.settings and b in vals.settings
         True
-        >>> 'dep1' in vals.depends and 'dep2' in vals.depends
+        >>> 'dep1' in vals.dependencies() and 'dep2' in vals.dependencies()
         True
         """
         if setting.do_append:
-            # Appending settings with at least the current level of
-            # specificity should be kept.
-            if specificity >= self._specificity:
-                if specificity not in self.append_settings:
-                    self.append_settings[specificity] = []
-                self.append_settings[specificity].append(setting)
-                self.depends |= depends
+            self.append_settings.append(setting)
+            self._append_depends |= depends
         else:
-            # Add equally specific settings to the list.
-            if specificity == self._specificity:
-                self.settings.append(setting)
-                self.depends |= depends
-            # Replace the list if the setting is more specific.
-            elif specificity > self._specificity:
-                self.settings = [setting]
-                self._specificity = specificity
-                self.depends = depends
-        # Do nothing if the setting is less specific.
+            mark_deletion = []
+            for i in range(len(self.settings)):
+                other_setting = self.settings[i]
+                other_specificity = self._specificities[i]
+                # Ignore this case if it's less specific than one we have.
+                if other_specificity > specificity:
+                    if other_setting.has_special_case(setting):
+                        return
+                # Override cases that are less specific than this one.
+                elif other_specificity < specificity:
+                    if setting.has_special_case(other_setting):
+                        mark_deletion.append(i)
+            mark_deletion.reverse()
+            for i in mark_deletion:
+                del self.settings[i]
+                del self._specificities[i]
+                del self._depends[i]
+            self.settings.append(setting)
+            self._specificities.append(specificity)
+            self._depends.append(depends)
 
     def ambiguity_check(self):
         """Check the current list of settings for ambiguity.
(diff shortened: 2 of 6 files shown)


jgfouca:
@quantheory , thanks so much for doing this. Let me test in on theta.

jgfouca:
Worked!

jgfouca:
@quantheory , did you run ./scripts_regression_tests.py H_TestMakeMacros ? I'm seeing some fails:

```
FAIL: test_appends_not_overriden (__main__.H_TestMakeMacros)
Test that machine-specific base value changes don't interfere with appends.
  File "./scripts_regression_tests.py", line 2719, in test_appends_not_overriden
    tester.assert_variable_equals("FFLAGS", "-base2", env={"COMPILER": self.test_compiler})
  File "./scripts_regression_tests.py", line 2205, in assert_variable_equals
    self.parent.assertEqual(self.query_var(var_name, env, var), value)
AssertionError: '-base1' != '-base2'

FAIL: test_mach_other_compiler (__main__.H_TestMakeMacros)
The macro writer compiler-specific logic works as expected.
  File "./scripts_regression_tests.py", line 2425, in test_mach_other_compiler
    tester.assert_variable_equals("CFLAGS", "x y z",env={"COMPILER":self.test_compiler})
  File "./scripts_regression_tests.py", line 2205, in assert_variable_equals
    self.parent.assertEqual(self.query_var(var_name, env, var), value)
AssertionError: 'a b c' != 'x y z'
```

quantheory:
I did run those tests, and they worked for me. I also checked that they pass with both Python 2 and 3, since that trips me up sometimes. Honestly not sure what's different between how I run it and how you do. I also remember that you had trouble reproducing the issue on theta using the regression tests, but I didn't have any trouble, so something must be different between how I'm using the tests and how you are?

jgfouca:
@quantheory , wow, this gets stranger by the minute. Are you running in CESM mode?

jgfouca:
Looks like CESM/E3SM mode made no difference.

jgfouca:
So, I just re-cloned my repo, and logged-out and back-in. This would have fixed any problems with stale pyc files or rogue entry in PYTHONPATH causing problems. I'm still seeing the test failure:

```
./scripts_regression_tests.py H_TestMakeMacros.test_mach_other_compiler
FAIL: test_mach_other_compiler (__main__.H_TestMakeMacros)
The macro writer compiler-specific logic works as expected.
  File "./scripts_regression_tests.py", line 2425, in test_mach_other_compiler
    tester.assert_variable_equals("CFLAGS", "x y z",env={"COMPILER":self.test_compiler})
  File "./scripts_regression_tests.py", line 2205, in assert_variable_equals
    self.parent.assertEqual(self.query_var(var_name, env, var), value)
AssertionError: 'a b c' != 'x y z'
```

jgfouca:
I'm seeing the error on anlworkstation too.

jgfouca:
Error on blues too.

rljacob:
What was the part that "Worked" above?

jgfouca:
@rljacob , the thing that worked on theta was the DEBUG FFlags that have been causing problems for us on e3sm.

quantheory:
I am increasingly confused as well.

```
[15:00:42]santos36@quartz188:~/Code/cime >git status
On branch quantheory/redo_machine_specific_compilers
Your branch is up-to-date with 'origin/quantheory/redo_machine_specific_compilers'.
nothing to commit, working directory clean
[15:00:45]santos36@quartz188:~/Code/cime >git log | head -n3
commit 1169bf496b688a828287225e676c52e2158a4b86
Author: Sean Patrick Santos <SeanPatrickSantos@gmail.com>
Date:   Wed Jul 11 11:16:44 2018 -0700
```

Also:

```
[15:02:29]santos36@quartz188:~/Code/cime/scripts/tests >python scripts_regression_tests.py H_TestMakeMacros

Testing commit 1169bf496b688a828287225e676c52e2158a4b86
Using cime_model = e3sm
Testing machine = quartz
Test root: /p/lscratchh/santos36/scripts_regression_test.20180711_150304

test_append_flags (__main__.H_TestMakeMacros)
Test appending flags to a list. ... ok
test_append_flags_without_base (__main__.H_TestMakeMacros)
Test appending flags to a value set before Macros is included. ... ok
test_appends_not_overriden (__main__.H_TestMakeMacros)
Test that machine-specific base value changes don't interfere with appends. ... ok
test_base_flags (__main__.H_TestMakeMacros)
Test that we get "base" compiler flags. ... ok
test_build_time_append_flags (__main__.H_TestMakeMacros)
Test build_time selection of compiler flags. ... ok
test_build_time_attribute (__main__.H_TestMakeMacros)
The macro writer writes conditionals for build-time choices. ... ok
test_build_time_base_flags (__main__.H_TestMakeMacros)
Test selection of base flags based on build-time attributes. ... ok
test_build_time_base_flags_same_parent (__main__.H_TestMakeMacros)
Test selection of base flags in the same parent element. ... ok
test_compiler_changeable_at_build_time (__main__.H_TestMakeMacros)
The macro writer writes information for multiple compilers. ... ok
test_config_reject_cyclical_references (__main__.H_TestMakeMacros)
Test that cyclical $VAR references are rejected. ... ok
test_config_reject_self_references (__main__.H_TestMakeMacros)
Test that $VAR self-references are rejected. ... ok
test_config_variable_insertion (__main__.H_TestMakeMacros)
Test that $VAR insert variables from config_compilers. ... ok
test_env_and_shell_command (__main__.H_TestMakeMacros)
Test that $ENV works inside $SHELL elements. ... ok
test_environment_variable_insertion (__main__.H_TestMakeMacros)
Test that ENV{..} inserts environment variables. ... ok
test_generic_item (__main__.H_TestMakeMacros)
The macro writer can write out a single generic item. ... ok
test_ignore_non_match (__main__.H_TestMakeMacros)
The macro writer ignores an entry with the wrong machine name. ... ok
test_mach_and_os_beats_mach (__main__.H_TestMakeMacros)
The macro writer chooses the most-specific match possible. ... ok
test_mach_beats_os (__main__.H_TestMakeMacros)
The macro writer chooses machine-specific over os-specific matches. ... ok
test_mach_other_compiler (__main__.H_TestMakeMacros)
The macro writer compiler-specific logic works as expected. ... ok
test_machine_specific_append_flags (__main__.H_TestMakeMacros)
Test appending flags that are either more or less machine-specific. ... ok
test_machine_specific_base_and_append_flags (__main__.H_TestMakeMacros)
Test that machine-specific base flags coexist with machine-specific append flags. ... ok
test_machine_specific_base_flags (__main__.H_TestMakeMacros)
Test selection among base compiler flag sets based on machine. ... ok
test_machine_specific_base_keeps_append_flags (__main__.H_TestMakeMacros)
Test that machine-specific base flags don't override default append flags. ... ok
test_machine_specific_item (__main__.H_TestMakeMacros)
The macro writer can pick out a machine-specific item. ... ok
test_multilevel_specificity (__main__.H_TestMakeMacros)
Check that settings with multiple levels of machine-specificity can be resolved. ... ok
test_multiple_shell_commands (__main__.H_TestMakeMacros)
Test that more than one $SHELL element can be used. ... ok
test_os_specific_item (__main__.H_TestMakeMacros)
The macro writer can pick out an OS-specific item. ... ok
test_override_with_machine_and_new_attributes (__main__.H_TestMakeMacros)
Test that overrides with machine-specific settings with added attributes work correctly. ... ok
test_override_with_machine_and_same_attributes (__main__.H_TestMakeMacros)
Test that machine-specific conditional overrides with the same attribute work correctly. ... ok
test_reject_ambiguous (__main__.H_TestMakeMacros)
The macro writer dies if given an ambiguous set of matches. ... ok
test_reject_duplicate_defaults (__main__.H_TestMakeMacros)
The macro writer dies if given many defaults. ... ok
test_reject_duplicates (__main__.H_TestMakeMacros)
The macro writer dies if given many matches for a given configuration. ... ok
test_remove_dependency_issues (__main__.H_TestMakeMacros)
Check that overridden settings don't cause inter-variable dependencies. ... ok
test_shell_command_insertion (__main__.H_TestMakeMacros)
Test that $SHELL insert shell command output. ... ok
test_variable_insertion_with_machine_specific_setting (__main__.H_TestMakeMacros)
Test that machine-specific $VAR dependencies are correct. ... ok

Ran 35 tests in 0.752s

OK
('All pass, removing directory:', '/p/lscratchh/santos36/scripts_regression_test.20180711_150304')
[15:03:09]santos36@quartz188:~/Code/cime/scripts/tests >python --version
Python 2.7.14
```

Got the essentially the same thing on both quartz (LC machine) and edison.

quantheory:
Note also that my copy of CIME on edison was a brand new clone. Are we testing the exact same commit? Is there something we should be checking in our respective .bashrc files or something? Not sure how we can so consistently get different results on different machines.

jgfouca:
@quantheory , OK edison is the first machine where this works for me. This is with python-2.7.13

jgfouca:
This may be a problem with older python 2.7's

quantheory:
OK, that's a start. What versions are failing?

jgfouca:
2.7.6 (anlworkstation), 2.7.9 (melvin), 3.5.2(melvin)

rljacob:
python 2.7.3 on blues failed for me. Same 2 fails Jim saw above.

jgfouca:
OK, it's not the python version. I install 2.7.15 on melvin and it still breaks.

rljacob:
Oh joy.  python 2.7.5 on bebop works.

jgfouca:
Wow. We really need to get to the source of this divergence in behavior and eliminate it.

quantheory:
Question: Does the same test pass in I_TestCMakeMacros? Those tests are identical to those in H_TestMakeMacros, except that the intermediate file produced is in CMake rather than Makefile format.

quantheory:
Found it! At least partly. Remember that line that sets the `append` flag weird? That's screwing things up somehow. I'm not entirely sure that I understand why it affects some machines and not others though. (Or why it even exists.)

jgfouca:
I_TestCMakeMacros has the same fails.

quantheory:
OK, try this!

jgfouca:
> OK, try this!

GOAAAAAALLLLLLLLLLLLLLLLLLLLLLL

jgfouca:
Running full test suite now, will merge when it passes.

quantheory:
Cool! I do want to warn that this collection of changes might require config_compilers.xml to be changed as well for some machines. I didn't see any big red flags, but there are some entries in the E3SM config_compilers that look like copy+pasted versions of other entries. Most of the time that's just going to result in duplicated options (e.g. specifying `-g` 3-4 times for DEBUG cases), but there could be something more annoying that arises. CESM's config_compilers looks a bit better-curated, but then again I'm looking at the ESMCI versions of both, which are probably both out of date.

mvertens:
@santos - I have not had a chance to look at this carefully, but I do know
that @jedwards4b was working quite a bit on config_compilers.xml before he
left for vacation. Could we wait until he returns on Monday before we
officially accept a PR?

quantheory:
@mvertens Works for me either way. I was only working on this because #2703 kind of broke some things, and @jgfouca seemed to be having a rough time trying to deal with that yesterday. Though my handle here is @quantheory. I don't know who SaNToS is on here.

mvertens:
@quantheory - okay that makes sense.  Thanks.

jgfouca:
@mvertens , I believe @jedwards4b had wrapped-up his work, and I desperately need this on the E3SM side to complete the V2 upgrade, so I'm going to have to merge this.

mvertens:
@jfouca - thanks for the clarification. All sounds good.

jedwards4b:
@quantheory Thanks for the fix and the additional tests.

PR #2721, opened 2018-07-17
Fix pylint error

jgfouca:
Test suite: code-checker (py3)
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/batch.py
+++ b/scripts/lib/CIME/XML/batch.py
@@ -93,10 +93,6 @@ class Batch(GenericXML):
         if node is not None:
             value = self.text(node)
 
-        if value is None:
-            # if all else fails
-            value = GenericXML.get_value(self, name, attribute, resolved, subgroup)
-
         if resolved:
             if value is not None:
                 value = self.get_resolved_value(value)
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -453,7 +453,7 @@ class GenericXML(object):
 
             # The overridden versions of this method do not simply return None
             # so the pylint should not be flagging this
-            ref = self.get_value(var) # pylint disable=assignment-from-none
+            ref = self.get_value(var) # pylint: disable=assignment-from-none
 
             if ref is not None:
                 logger.debug("resolve: " + str(ref))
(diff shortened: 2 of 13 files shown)


jgfouca:
@jedwards4b , please re-review. I originally opened this PR solely for triggering a travis CI check. Their machine must have gotten some updates and the new pylint for py3 is a lot pickier. It found at least a dozen issues, some significant. I think I've fixed them.

scripts/lib/CIME/XML/batch.py
```diff
@@ -93,10 +93,6 @@ def get_value(self, name, attribute=None, resolved=True, subgroup=None):
         if node is not None:
             value = self.text(node)
 
-        if value is None:
```
jgfouca: This pattern was broken. The GenericXML.get_value is a placeholder that just always returns None.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -400,6 +400,8 @@ def set_value(self, vid, value, subgroup=None, ignore_type=True): # pylint: disa
         for node in valnodes:
             self.set_text(node, value)
 
+        return value if valnodes else None
```
jgfouca: This was a serious error. All of the subclasses of GenericXML return something if the set_value was successful, but this one wasn't doing that. Some of the subclasses were invoking this version and so this was a bug.

jedwards4b:
There still appears to be a failure.

jgfouca:
@jedwards4b , yes, a typo in one of my disables

jgfouca:
As a side note, I want to thank the folks who pushed for py3. It looks like py3 and py3 libraries have moved, and are continuing to move, ahead of the py2 versions.

mfdeakin-sandia:
Thanks again for actually implementing python 3 @jedwards4b !
As a side note, when I set the scream project up to use codacity for static analysis code reviews, it picked up a lot of potential issues in cime that you all might be interested in

jedwards4b:
@mfdeakin-sandia Sure we would like to see them if you have an easy way to share.

mfdeakin-sandia:
I did have them easily available, but don't know how to get to them since I excluded the CIME directory from the repo. Codacy is free for open source projects and trivial to set up with github, so you should be able to set it up for CIME itself.

PR #2426, opened 2018-04-05
more fixes for archiver and tests

jedwards4b:
Problems with fully coupled tests - archiving wasn't correct.
Test suite: scripts_regression_tests.py
Test baseline: 
Test namelist changes: 
Test status: bit for bit,

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

PR #2728, opened 2018-07-24
Hist utils use archive xml

### LINKED ISSUE #2695: baseline compare and generate should use config_archive.xml
Currently the parsing for history files used by component_compare_* is independent of the parsing used in config_archive.xml - they should use the same parsing so that modifications need only be made in one place.

### LINKED ISSUE #2735: cpl avg hist and aux hist files not archived?
I haven't confirmed this is an issue, but from looking at https://github.com/ESMCI/cime/blob/4f41034b08cf9922224bb412b0e9ad0b9f5e675a/src/drivers/mct/cime_config/config_archive.xml#L4

it looks like we currently aren't archiving cpl avg hist (.ha) or aux hist files.

(With the changes in #2728 , this means those also wouldn't be picked up in system test comparisons, if we had any system tests that produced those files, which we probably don't yet but may in the future.)

jedwards4b:
Use information from env_archive.xml to determine history files for baselines and compares.
This removes the need for changes in two places to support different history file name formats

Test suite: scripts_regression_tests.py, cesm prealpha tests on hobart
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #2695 
Fixes #2735 
User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: sacks

### DIFF
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -6,7 +6,7 @@ from CIME.XML.standard_module_setup import *
 from CIME.test_status import TEST_NO_BASELINES_COMMENT
 from CIME.utils import get_current_commit, get_timestamp, get_model, safe_copy
 
-import logging, glob, os, re, stat, filecmp
+import logging, os, re, stat, filecmp
 logger = logging.getLogger(__name__)
 
 BLESS_LOG_NAME = "bless_log"
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -17,7 +17,7 @@ def _iter_model_file_substrs(case):
     for model in models:
         yield model
 
-def _get_all_hist_files(testcase, model, from_dir, suffix="", file_extensions=None):
+def _get_all_hist_files(model, from_dir, suffix="", file_extensions=None, ref_case=None):
     suffix = (".{}".format(suffix) if suffix else "")
 
     test_hists = []
@@ -32,13 +32,18 @@ def _get_all_hist_files(testcase, model, from_dir, suffix="", file_extensions=No
         pfile = re.compile(string)
         test_hists.extend([os.path.join(from_dir,f) for f in os.listdir(from_dir) if pfile.search(f)])
 
+    if ref_case:
+        for test in test_hists:
+            if ref_case in os.path.basename(test):
+                test_hists.remove(test)
+
     test_hists = list(set(test_hists))
     test_hists.sort()
     logger.debug("_get_all_hist_files returns {} for model {}".format(test_hists, model))
     return test_hists
 
-def _get_latest_hist_files(testcase, model, from_dir, suffix="", file_extensions=None):
-    test_hists = _get_all_hist_files(testcase, model, from_dir, suffix, file_extensions)
+def _get_latest_hist_files(testcase, model, from_dir, suffix="", file_extensions=None, ref_case=None):
+    test_hists = _get_all_hist_files(model, from_dir, suffix=suffix, file_extensions=file_extensions, ref_case=ref_case)
     latest_files = {}
     histlist = []
     for hist in test_hists:
@@ -60,7 +65,7 @@ def copy(case, suffix):
     """
     rundir   = case.get_value("RUNDIR")
     testcase = case.get_value("CASE")
-
+    ref_case = case.get_value("RUN_REFCASE")
     # Loop over models
     archive = case.get_env("archive")
     comments = "Copying hist files to suffix '{}'\n".format(suffix)
@@ -71,7 +76,7 @@ def copy(case, suffix):
             file_extensions = archive.get_hist_file_extensions(archive.get_entry('drv'))
         else:
             file_extensions = archive.get_hist_file_extensions(archive.get_entry(model))
-        test_hists = _get_latest_hist_files(testcase, model, rundir, file_extensions=file_extensions)
+        test_hists = _get_latest_hist_files(testcase, model, rundir, file_extensions=file_extensions, ref_case=ref_case)
         num_copied += len(test_hists)
         for test_hist in test_hists:
             new_file = "{}.{}".format(test_hist, suffix)
@@ -190,6 +195,7 @@ def _compare_hists(case, from_dir1, from_dir2, suffix1="", suffix2="", outfile_s
     comments = "Comparing hists for case '{}' dir1='{}', suffix1='{}',  dir2='{}' suffix2='{}'\n".format(testcase, from_dir1, suffix1, from_dir2, suffix2)
     multiinst_driver_compare = False
     archive = case.get_env('archive')
+    ref_case = case.get_value("RUN_REFCASE")
     for model in _iter_model_file_substrs(case):
         if model == 'cpl' and suffix2 == 'multiinst':
             multiinst_driver_compare = True
@@ -198,8 +204,8 @@ def _compare_hists(case, from_dir1, from_dir2, suffix1="", suffix2="", outfile_s
             file_extensions = archive.get_hist_file_extensions(archive.get_entry('drv'))
         else:
             file_extensions = archive.get_hist_file_extensions(archive.get_entry(model))
-        hists1 = _get_latest_hist_files(testcase, model, from_dir1, suffix1, file_extensions)
-        hists2 = _get_latest_hist_files(testcase, model, from_dir2, suffix2, file_extensions)
+        hists1 = _get_latest_hist_files(testcase, model, from_dir1, suffix1, file_extensions, ref_case)
+        hists2 = _get_latest_hist_files(testcase, model, from_dir2, suffix2, file_extensions, ref_case)
         if len(hists1) == 0 and len(hists2) == 0:
             comments += "    no hist files found for model {}\n".format(model)
             continue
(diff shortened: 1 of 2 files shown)


jedwards4b:
Still testing this - it would be good if you could run e3sm tests against a baseline.

billsacks:
I haven't looked at this in detail yet, but I realized one likely issue with this change: For CISM, we do NOT want the test system to pick up the history files with extension 'initial_hist' (which is listed as an entry in config_archive.xml): see http://bugs.cgd.ucar.edu/show_bug.cgi?id=2355. My initial thought is: Can we have an optional attribute in the xml attached to the `hist_file_extension` tag, like "ignore_for_test_compares"? I'm imagining something like this:

```xml
    <hist_file_extension ignore_for_test_compares="true">initial_hist</hist_file_extension>
```

though I'm certainly open to other implementations.

rljacob:
Maybe config_archive.xml should also be renamed to something more general like config_filenames.xml.   I can see a need to describe to CIME the complete output of a model.  Some of that info is used for archiving, some for baseline compare, some for a future use.

jedwards4b:
@rljacob  In order to be consistent in our conventions it would mean renaming the associated python files as well.  Since archiving is the primary functionality I don't see a strong reason for renaming it at this time.

scripts/lib/CIME/hist_utils.py
```diff
@@ -352,11 +357,22 @@ def get_extension(model, filepath):
...
     >>> get_extension("pop","PFS.f09_g16.B1850.cheyenne_intel.allactive-default.GC.c2_0_b1f2_int.pop.h.ecosys.nday1.0001-01-02.nc")
     'h'
+    >>> get_extension("mom", "ga0xnw.mom6.frc._0001_001.nc")
+    'frc'
+    >>> get_extension("mom", "ga0xnw.mom6.sfc.day._0001_001.nc")
+    'sfc.day'
     """
     basename = os.path.basename(filepath)
-    ext_regex = re.compile(r'.*%s[^_]*_?([0-9]{4})?[.](h.?)([.].*[^.])?[.]nc' % model)
+    if model == "mom":
```
billsacks: Does the need to list component-specific rules here kind of defeat the purpose of these changes? Is there a way to encode and extract this information in env_archive.xml?

billsacks:
I talked with Bill Lipscomb, and he said it _is_ important to archive CISM's initial_hist file. I haven't tried it, but I believe that a test that includes CISM will currently fail; here's a doctest that shows how I expect it to fail:

```diff
diff --git a/scripts/lib/CIME/hist_utils.py b/scripts/lib/CIME/hist_utils.py
index 6d5808464..a9a8677cf 100644
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -361,6 +361,8 @@ def get_extension(model, filepath):
     'frc'
     >>> get_extension("mom", "ga0xnw.mom6.sfc.day._0001_001.nc")
     'sfc.day'
+    >>> get_extension("cism", "SMS_D_Ly1.f09_g17_gl4.T1850G.cheyenne_gnu.GC.0708_1208_g.cism.initial_hist.0001-01-01-00000.nc")
+    'initial_hist'
     """
     basename = os.path.basename(filepath)
     if model == "mom":
```

If you add that doc_test, you'll see that it currently fails with:

```
    SystemExit: ERROR: Failed to get extension for file 'SMS_D_Ly1.f09_g17_gl4.T1850G.cheyenne_gnu.GC.0708_1208_g.cism.initial_hist.0001-01-01-00000.nc'
```

So I think it's important to have a way for the initial_hist file to be picked up for archiving but not for system test comparison purposes. (The problem is that, for tests that do a restart, it's present in the initial run but not the restart run, which at least in the past caused the test to fail.)

jgfouca:
One thing I can do is kick off a new round of sync-ing after this PR is merged.

jedwards4b:
Okay this now passes all cesm cheyenne prealpha tests.

scripts/lib/CIME/hist_utils.py
```diff
@@ -18,30 +17,33 @@ def _iter_model_file_substrs(case):
...
-        test_hists.extend(glob.glob("{}/{}*.h?.nc".format(from_dir, model)))
-        test_hists.extend(glob.glob("{}/{}*.h.*.nc".format(from_dir, model)))
-        test_hists.extend(glob.glob("{}/{}*.h?.*.nc".format(from_dir, model)))
-
+    for extension in file_extensions:
+        if 'initial' in extension:
+            continue
+        if extension.endswith('$'):
+            extension = extension[:-1]
+        string = model+r'\d?_?\d*\.'+extension+suffix+'$'
```
billsacks: This will only work if `<hist_file_extension>` is a regex that goes all the way up through the `.nc`. That is mostly true for CESM, with the exception of:

A few components have this for DART (which according to @jedwards4b may be okay):
```xml
<hist_file_extension>e</hist_file_extension>
```

POP has this; I'm not sure if this one should be captured by this match or not:
```xml
    <hist_file_extension>d[dovt]</hist_file_extension>
```

CISM has this, though in this case it's okay that it isn't picked up (we skip this 'initial_hist' file anyway):
```xml
    <hist_file_extension>initial_hist</hist_file_extension>
```

ESP has this... it may be okay that this isn't picked up:
```xml
      <hist_file_extension>[ei]</hist_file_extension>
```

**However, e3sm has a lot of instances of hist_file_extension not going up through the .nc.**

At the very least, the e3sm config_archive file needs to be changed. Even so, this leaves me a little worried about the asymmetry between how `hist_file_extension` is listed compared with `rest_file_extension`, as well as inconsistencies within `hist_file_extension` itself: We seem to now have a rule that `hist_file_extension` should be a regex going all the way up through the `.nc` _if_ you want these history files to be picked up by the test system. If that is the intended rule, can it be documented somewhere?

scripts/lib/CIME/hist_utils.py
```diff
@@ -342,21 +354,26 @@ def get_extension(model, filepath):
...
     >>> get_extension("pop","PFS.f09_g16.B1850.cheyenne_intel.allactive-default.GC.c2_0_b1f2_int.pop.h.ecosys.nday1.0001-01-02.nc")
     'h'
+    >>> get_extension("mom", "ga0xnw.mom6.frc._0001_001.nc")
+    'frc'
+    >>> get_extension("mom", "ga0xnw.mom6.sfc.day._0001_001.nc")
+    'sfc'
     """
     basename = os.path.basename(filepath)
-    ext_regex = re.compile(r'.*%s[^_]*_?([0-9]{4})?[.](h.?)([.].*[^.])?[.]nc' % model)
+    regex = model+r'\d?_?(\d{4})?\.(\w+)[-\w\.]*\.nc\.?'
```
billsacks: The `\d{4}` here is inconsistent with `\d*` in `_get_all_hist_files`. It seems these two should be consistent to avoid subtle bugs.

billsacks:
Adding some more thoughts regarding my comment above; I don't feel this needs to be addressed now, but I want to record these notes for the future:

> As we discussed, I was also a bit concerned about the fact that the regex in `get_extension` (`regex = model+r'\d?_?(\d{4})?\.(\w+)[-\w\.]*\.nc\.?'`) implies that we only accept `hist_file_extensions` made up of word characters (alphanumeric plus `_`). After discussing this with you, I feel this is okay for now; I'll add some thoughts on this in a separate comment below, but those thoughts don't need to be addressed now.

It seems like we should be able to get the extension in `_get_all_hist_files`, rather than relying on a hard-coded regex in `get_extension`: If we add a capture group in the regex in `_get_all_hist_files`, we can extract this portion of the match to get the extension. (Actually, we'll need two capture groups, like there are in `get_extension`, with the first capture group giving the optional multi-instance piece of the file name.) Then `_get_all_hist_files` would return a list of tuples, (filename, extension).

I can see two ways to do this:

(1) Require there to be a capture group in the `<hist_file_extension>` regular expression in `config_archive.xml`. For example, we'd have `<hist_file_extension>(h\d*).*\.nc$</hist_file_extension>`. Then "string" in `_get_all_hist_files` would look like:

```python
string = model+'\d?_?(\d{4})?\.'+extension+suffix+'$'
```

@jedwards4b suggested: we could make the capture group optional in the xml; if we see that there is no capture group there, then we could fall back on something like the current logic in `get_extension`.

(2) An alternative to this is that <hist_file_extension> could just look like `<hist_file_extension>h\d*</hist_file_extension>` and "string" would look like:

```python
string = model+'\d?_?(\d{4})?\.'+'('+extension+')'+'.*\.nc'+suffix+'$'
```

I like that alternative better because (a) it keeps hist_file_extension looking more like the other entries in config_archive, and (b) it doesn't require people who edit config_archive.xml to know about the need for a capture group. However, it only works if all history files end with .nc. But actually, would it work to drop the '.*\.nc', so that we have?:

```xml
<hist_file_extension>h\d*</hist_file_extension>
```

```python
string = model+'\d?_?(\d{4})?\.'+'('+extension+')'+'.*'+suffix+'$'
```

The problem with (2), though, is that it might bring back whatever problems were fixed in https://github.com/ESMCI/cime/pull/2426 (@jedwards4b doesn't remember exactly what problems were encountered that needed to be fixed with #2426).

scripts/lib/CIME/hist_utils.py
```diff
@@ -18,30 +17,33 @@ def _iter_model_file_substrs(case):
     for model in models:
         yield model
 
-def _get_all_hist_files(testcase, model, from_dir, suffix=""):
+def _get_all_hist_files(model, from_dir, suffix="", file_extensions=None, ref_case=None):
```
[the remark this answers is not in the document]
jedwards4b: done

scripts/lib/CIME/hist_utils.py
jedwards4b: ok

jedwards4b:
@jgfouca can we merge this?

jgfouca:
@jedwards4b , what about the pylint errors?

jedwards4b:
Those are already fixed on master.

PR #2733, opened 2018-07-29
ESMF logging fix

### LINKED ISSUE #2729: ESMF builds do not respect the setting of ESMF_LOGFILE_KIND
Given this setting:
```
./xmlquery ESMF_LOGFILE_KIND
        ESMF_LOGFILE_KIND: ESMF_LOGKIND_NONE
```
a run should produce no ESMF files. However, runs build with ESMF are producing per-PE log files as if the setting was `ESMF_LOGKIND_NONE`.

Reported by a user who just upgraded to CESM 2.0

gold2718:
Add a namelist variable and code to read and use the ESMF logging option XML variable, `ESMF_LOGFILE_KIND`, in the driver's config_component.xml file.
The new namelist, `esmf_inparm`, is read in `cime_pre_init1` so that the logging option is available for the ESMF_initialize call.
Also added a new logging option, `ESMF_LOGKIND_MULTI_ON_ERROR`.

Note, there are existing pylint failures in scripts/Tools/archive_metadata
Test baseline: NA
Test namelist changes: NA
Test status: bit for bit

Fixes #2729 

User interface changes?: New driver namelist and variable, 

Code review:


gold2718:
It is a bug because users have supposedly been able to control logging with the existing XML variable. This fix merely makes proper use of an _existing_ XML variable.
It is also a bug reported by a CESM 2.0 user as they set an existing XML variable and got the incorrect behavior.

jedwards4b:
@mvertens what do you think?

mvertens:
@jedwards4b - I do think its a bug - since this is a supported feature of
ESMF that we are not able to access. That said I do not think it needs to
go into 2.0.1 but should go into 2.1.

gold2718:
@jedwards4b, does that mean we just delay this PR until after 2.0.1 is tagged? Something else?

jedwards4b:
If it's a bug fix then it will go into 2.0 and may be able to go into 2.0 if Erik doesn't have the clm tag done yet.   @fischer-ncar?

ekluzek:
@jedwards4b the clm tag for the 2.0 release update was done yesterday. I can't say where @fischer-ncar is at with testing with it though.

fischer-ncar:
I started the testing Sunday afternoon.   I can kill it and restart it tomorrow with this fix, if wanted.

gold2718:
@fischer-ncar, no, I think the message from @mvertens (see above in this PR thread) is that this fix is not required for 2.0.1.

PR #2740, opened 2018-08-07
For inputdata servers make it so key members are always defined.

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/Servers/ftp.py
+++ b/scripts/lib/CIME/Servers/ftp.py
@@ -15,6 +15,8 @@ class FTP(GenericServer):
         logger.info("server address {} root path {}".format(ftp_server, root_address))
         self.ftp = FTPpy(ftp_server)
 
+        self._ftp_server = address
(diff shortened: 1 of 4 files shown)


jgfouca:
@jedwards4b , I wasn't entirely sure what the right fix is here, but something is needed. Under the old implementation, if the connection in the constructor failed, a key data member was not getting set, resulting in ugly exceptions every time a download attempt was made.

Under this implementation, the constructor connection attempt is mostly useless except for providing some error info to the user.

I'd also be open to setting a flag in the constructor if the connection fails. If the flag is set, all download attempts are skipped with errors to the user that it's a bad server. That may be too strict though.

jedwards4b:
Can you look into the py3 error before we merge?

PR #2742, opened 2018-08-08
updates for compatibility with XML to html tools

bertinia:
These are updates to cime based classes in support of the XML to HTML 
auto-generation tools in https://github.com/NCAR/CESM_xml2html.
Also minor update to DOCN config_component.xml to correct invalid ASCII character.

Test suite: scripts_regression_tests.py B_CheckCode
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

Fixes [CIME Github issue #] 

User interface changes?: None

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/component.py
+++ b/scripts/lib/CIME/XML/component.py
@@ -282,3 +282,46 @@ class Component(EntryID):
             name = self.get(entry, "id")
             text = self.text(self.get_child("desc", root=entry))
             logger.info("   {:20s} : {}".format(name, text.encode('utf-8')))
+
+
+    def return_values(self):
+        """
+        return a list of hashes from target config_component.xml file
+        """
+        entry_dict = dict()
+        items = list()
+        try:
+            helpnode = self.get_child("help")
+            helptext = self.text(helpnode)
+        except:
+            helptext = ''
+            
+        entries = self.get_children("entry")
+        for entry in entries:
+            item = dict()
+            name = self.get(entry, "id")
+            datatype = self.text(self.get_child("type", root=entry))
+            try:
+                valid_values = self.text(self.get_child("valid_values", root=entry))
+            except:
+                valid_values = ''
+
+            try:
+                value = self.get_value(name)
+            except:
+                value = ''
+            
+            group = self.text(self.get_child("group", root=entry))
+            filename = self.text(self.get_child("file", root=entry))
+            text = self.text(self.get_child("desc", root=entry))
+            item = { "name":name,
+                     "datatype":datatype,
+                     "valid_values":valid_values,
+                     "value":value,
+                     "group":group,
+                     "filename":filename,
+                     "desc":text.encode('utf-8')}
+            items.append(item)
+        
+        entry_dict = { "items" : items }
+        return helptext, entry_dict
--- a/scripts/lib/CIME/XML/component.py
+++ b/scripts/lib/CIME/XML/component.py
@@ -303,12 +303,10 @@ class Component(EntryID):
                 valid_values = self.text(self.get_child("valid_values", root=entry))
             except:
                 valid_values = ''
-
             try:
                 value = self.get_value(name)
             except:
                 value = ''
-            
             group = self.text(self.get_child("group", root=entry))
             filename = self.text(self.get_child("file", root=entry))
             text = self.text(self.get_child("desc", root=entry))
@@ -320,6 +318,6 @@ class Component(EntryID):
                      "filename":filename,
                      "desc":text.encode('utf-8')}
             items.append(item)
-        
         entry_dict = { "items" : items }
+
         return helptext, entry_dict
--- a/scripts/lib/CIME/XML/component.py
+++ b/scripts/lib/CIME/XML/component.py
@@ -289,19 +289,21 @@ class Component(EntryID):
         """
         entry_dict = dict()
         items = list()
-        try:
+        exists = self.child_exists("help")
+        if exists:
             helpnode = self.get_child("help")
             helptext = self.text(helpnode)
-        except:
+        else:
             helptext = ''
         entries = self.get_children("entry")
         for entry in entries:
             item = dict()
             name = self.get(entry, "id")
             datatype = self.text(self.get_child("type", root=entry))
-            try:
+            exists = self.child_exists("valid_values", root=entry)
+            if exists:
                 valid_values = self.text(self.get_child("valid_values", root=entry))
-            except:
+            else:
                 valid_values = ''
             try:
                 value = self.get_value(name)
--- a/scripts/lib/CIME/XML/component.py
+++ b/scripts/lib/CIME/XML/component.py
@@ -300,23 +300,15 @@ class Component(EntryID):
             item = dict()
             name = self.get(entry, "id")
             datatype = self.text(self.get_child("type", root=entry))
-            exists = self.get_optional_child("valid_values", root=entry)
-            if exists:
-                valid_values = self.text(self.get_child("valid_values", root=entry))
-            else:
-                valid_values = ''
-            exists = self.get_optional_child("default_value", root=entry)
-            if exists:
-                value = self.text(self.get_child("default_value", root=entry))
-            else:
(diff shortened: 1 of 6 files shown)


scripts/lib/CIME/XML/component.py
```diff
@@ -282,3 +282,46 @@ def print_values(self):
...
+    def return_values(self):
+        """
+        return a list of hashes from target config_component.xml file
+        """
+        entry_dict = dict()
+        items = list()
+        try:
+            helpnode = self.get_child("help")
+            helptext = self.text(helpnode)
+        except:
```
jedwards4b: I think that you need to specify an exception here, py3 does not allow general exceptions.

scripts/lib/CIME/XML/machines.py
```diff
@@ -320,3 +320,30 @@ def print_values(self):
...
                 print("      max_tasks/node ",self.text(max_tasks_per_node))
+
+    def return_values(self):
+        # return a dictionary of machine info
+        machines = self.get_children("machine")
+        mach_dict = dict()
+        logger.debug("Machines return values")
+        for machine in machines:
+            name = self.get(machine, "MACH")
+
```
jedwards4b: remove this line

scripts/lib/CIME/XML/component.py
```diff
@@ -282,3 +282,46 @@ def print_values(self):
             name = self.get(entry, "id")
             text = self.text(self.get_child("desc", root=entry))
             logger.info("   {:20s} : {}".format(name, text.encode('utf-8')))
+
```
jedwards4b: remove this blank line

scripts/lib/CIME/XML/component.py
```diff
@@ -282,3 +282,46 @@ def print_values(self):
...
+        return a list of hashes from target config_component.xml file
+        """
+        entry_dict = dict()
+        items = list()
+        try:
+            helpnode = self.get_child("help")
+            helptext = self.text(helpnode)
+        except:
+            helptext = ''
+
```
jedwards4b: remove extra whitespace

scripts/lib/CIME/XML/component.py
bertinia: What would be the exception returned if the self.text(helpnode) does not exist?

scripts/lib/CIME/XML/machines.py
bertinia: which line do you want me to remove? line 331?

jedwards4b: yes

scripts/lib/CIME/XML/component.py
```diff
@@ -282,3 +282,46 @@ def print_values(self):
...
+            try:
+                valid_values = self.text(self.get_child("valid_values", root=entry))
+            except:
+                valid_values = ''
+
+            try:
+                value = self.get_value(name)
+            except:
+                value = ''
+
```
jedwards4b: "try and except" is the wrong construct here,  I think you want to check if node valid_values exists and only call self.text if it does.  If it exists and has no text valid_values will be ''.   Same with the other one.

bertinia:
There's still the need for a try and except block around the call to  component.get_value due to this 
error:
ERROR: Multiple matches for nodename 'entry' and attrs '{'id': 'SUPPORTED_BY'}' in file '/home/aliceb/cesm/cime/src/drivers/mct/cime_config/config_component.xml'
I don't know how to get around this issue where the XML id attributes are not unique in 
a given config_component.xml file.

jedwards4b:
That is an error in the config_component.xml file - there shouldn't be two entries for SUPPORTED_BY

bertinia:
What do you want the two different "SUPPORTED_BY" entries to be? What are the repercussions
for the case control system scripts?  Do you want me to include the change in this PR?

jedwards4b:
NO I think that the correct thing to do was to remove one of them, it's just a duplication.  It doesn't matter which you remove.

bertinia:
ok - perhaps, @jedwards4b you could be a bit more specific instead of just replying with "NO". 
There are the 2 different SUPPORTED_BY entries in this config_component.xml file:

 <entry id="SUPPORTED_BY">
    <type>char</type>
    <default_value>UNSET</default_value>
    <group>case_def</group>
    <file>env_case.xml</file>
    <desc>current machine name support contact</desc>
  </entry>

and

  <entry id="SUPPORTED_BY">
    <type>char</type>
    <valid_values></valid_values>
    <group>build_def</group>
    <file>env_build.xml</file>
    <desc>email address of person (or group) that supports the build and port for this machine (do not edit)></desc>
  </entry>

They are defined with different groups, files and descs, so which one should I delete?

mvertens:
@bertinia - your are raising a valid question. Which config_component is this in?

jedwards4b:
I really didn't mean to shout NO - I was in a hurry and didn't correct the typo.   However there really is only one variable, just pick one of them and throw the other one away.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -250,6 +250,13 @@ def get_children(self, name=None, attributes=None, root=None):
 
         return children
```
jedwards4b: We do not need this additional function - use get_optional_child or get_children instead.

scripts/lib/CIME/XML/machines.py
```diff
@@ -320,3 +320,25 @@ def print_values(self):
...
+        for machine in machines:
+            name = self.get(machine, "MACH")
+            desc = self.get_child("DESC", root=machine)
+            mach_dict[(name,"description")] = self.text(desc)
+            os_  = self.get_child("OS", root=machine)
+            mach_dict[(name,"os")] = self.text(os_)
+            compilers = self.get_child("COMPILERS", root=machine)
+            mach_dict[(name,"compilers")] = self.text(compilers)
+            max_tasks_per_node = self.get_child("MAX_TASKS_PER_NODE", root=machine)
+            if max_tasks_per_node is not None:
```
jedwards4b: You should not need these if statements since this is a required machine entry.

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -136,7 +136,7 @@
     <desc>Model grid - DO NOT EDIT (for experts only)</desc>
   </entry>
 
-  <entry id="SUPPORTED_BY">
+  <entry id="MACHINE_SUPPORTED_BY">
```
jedwards4b: There is only one SUPPORTED_BY field.  Choose one of these and delete the other.

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -1910,6 +1911,7 @@
 
   <entry id="NTASKS">
     <type>integer</type>
+    <default_value>1</default_value>
```
jedwards4b: default_value is a deprecated field.   The default, if defined, is the value below without attribute.

scripts/lib/CIME/XML/generic_xml.py
bertinia: This return is from function get_children. I don't understand why you want to delete it. Please clarify.

scripts/lib/CIME/XML/component.py
```diff
@@ -289,7 +289,7 @@ def return_values(self):
         """
         entry_dict = dict()
         items = list()
-        exists = self.child_exists("help")
+        exists = self.get_optional_child("help")
         if exists:
             helpnode = self.get_child("help")
```
jedwards4b: I'm not sure I understand your approach here.  Why not:
```
helpnode = self.get_optional_child('help')
if helpnode:
  helptext = self.text(helpnode)
```

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -1911,6 +1903,7 @@
   <entry id="NTASKS">
     <type>integer</type>
     <values>
+      <value>1</value>
```
jedwards4b: we should remove these since they didn't work

PR #2744, opened 2018-08-08
Add BFAIL in more contexts when doing baseline comparisons

### LINKED ISSUE #789: If history files are missing from the baseline directory, BFAIL should be printed
(This was tested with cime5.2.0-alpha.7, but as far as I can tell, this behavior hasn't been changed since then.)

If I run a test with baseline comparisons, and the baseline directory is present but contains no history files, this should be reported as a BFAIL via the test status comment. Instead, no comment is given on the BASELINE line of TestStatus.

A BFAIL is correctly reported if the baseline directory is missing entirely, but in the case where the baseline directory is present but has no history files, I see, in TestStatus:

```
FAIL ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default BASELINE
```

despite having this in TestStatus.log:

```
2016-11-08 22:33:21: Comparing hists for case 'ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default.GC.1108-1358.40.p' dir1='/glade/scratch/sacks/ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default.GC.1108-1358.40.p/run', suffix1='',  dir2='/glade/p/cesmdata/cseg/ccsm_baselines/clm4_5_12_r201/ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default' suffix2=''
  comparing model 'datm'
    no hist files found for model datm
  comparing model 'clm'
    File '/glade/scratch/sacks/ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default.GC.1108-1358.40.p/run/ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default.GC.1108-1358.40.p.clm2.h0.0003-01-09-00000.nc' had no counterpart in '/glade/p/cesmdata/cseg/ccsm_baselines/clm4_5_12_r201/ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default' with suffix ''
    File '/glade/scratch/sacks/ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default.GC.1108-1358.40.p/run/ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default.GC.1108-1358.40.p.clm2.h1.0003-01-09-00000.nc' had no counterpart in '/glade/p/cesmdata/cseg/ccsm_baselines/clm4_5_12_r201/ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default' with suffix ''
  comparing model 'sice'
    no hist files found for model sice
  comparing model 'socn'
    no hist files found for model socn
  comparing model 'rtm'
    no hist files found for model rtm
  comparing model 'sglc'
    no hist files found for model sglc
  comparing model 'swav'
    no hist files found for model swav
  comparing model 'cpl'
    File '/glade/scratch/sacks/ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default.GC.1108-1358.40.p/run/ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default.GC.1108-1358.40.p.cpl.hi.0003-01-09-00000.nc' had no counterpart in '/glade/p/cesmdata/cseg/ccsm_baselines/clm4_5_12_r201/ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default' with suffix ''
Did not compare any hist files! Missing baselines?
```

What I want is something in TestStatus like what's printed if the baseline directory is missing entirely:

```
FAIL ERP_Ly5_Mmpi-serial.1x1_numaIA.ICRUCLM50BGCCROP.hobart_nag.clm-monthly BASELINE ERROR BFAIL baseline directory '/fs/cgd/csm/ccsm_baselines/clm4_5_12_r201/ERP_Ly5_Mmpi-serial.1x1_numaIA.ICRUCLM50BGCCROP.hobart_nag.clm-monthly' does not exist
```

but now it should say something like:

```
FAIL ERI_D_Ld9.f19_g16.I1850CN.yellowstone_pgi.clm-40default BASELINE ERROR BFAIL Did not compare any hist files! Missing baselines?
```

However, I'd say that this BFAIL should _only_ appear if _no_ history files were found in the baseline directory. Particularly important is to _not_ have a BFAIL if some history files differed and others were missing - so that a `grep -v BFAIL` wouldn't filter out those results.

My reason for wanting this is that these missing baselines generally indicate a very different problem from a standard BASELINE failure: namely, something went wrong in the previous creation of the baselines.

jgfouca:
The original implementation only produce BFAIL if a baseline directory
was entirely missing. Users are asking for BFAILs for individual baseline
files.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #789 

User interface changes?: N

Update gh-pages html (Y/N)?:N

### DIFF
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -183,7 +183,7 @@ def _hists_match(model, hists1, hists2, suffix1="", suffix2=""):
 
     return one_not_two, two_not_one, match_ups
 
-def _compare_hists(case, from_dir1, from_dir2, suffix1="", suffix2="", outfile_suffix="", add_bfail=False):
+def _compare_hists(case, from_dir1, from_dir2, suffix1="", suffix2="", outfile_suffix=""):
     if from_dir1 == from_dir2:
         expect(suffix1 != suffix2, "Comparing files to themselves?")
 
@@ -211,12 +211,11 @@ def _compare_hists(case, from_dir1, from_dir2, suffix1="", suffix2="", outfile_s
 
         one_not_two, two_not_one, match_ups = _hists_match(model, hists1, hists2, suffix1, suffix2)
         for item in one_not_two:
-            comments += "    {}File '{}' had no counterpart in '{}' with suffix '{}'\n".format(
-                (TEST_NO_BASELINES_COMMENT + " ") if add_bfail else "", item, from_dir2, suffix2)
+            comments += "    File '{}' had no compare counterpart in '{}' with suffix '{}'\n".format(item, from_dir2, suffix2)
             all_success = False
 
         for item in two_not_one:
-            comments += "    File '{}' had no counterpart in '{}' with suffix '{}'\n".format(item, from_dir1, suffix1)
+            comments += "    File '{}' had no original counterpart in '{}' with suffix '{}'\n".format(item, from_dir1, suffix1)
             all_success = False
 
         num_compared += len(match_ups)
@@ -336,7 +335,7 @@ def compare_baseline(case, baseline_dir=None, outfile_suffix=""):
         if not os.path.isdir(bdir):
             return False, "ERROR {} baseline directory '{}' does not exist".format(TEST_NO_BASELINES_COMMENT,bdir)
 
-    success, comments = _compare_hists(case, rundir, basecmp_dir, outfile_suffix=outfile_suffix, add_bfail=True)
+    success, comments = _compare_hists(case, rundir, basecmp_dir, outfile_suffix=outfile_suffix)
     if get_model() == "e3sm":
         bless_log = os.path.join(basecmp_dir, BLESS_LOG_NAME)
         if os.path.exists(bless_log):
@@ -457,3 +456,41 @@ def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
                                                get_timestamp(timestamp_format="%Y-%m-%d_%H:%M:%S")))
 
     return True, comments
+
+def get_ts_synopsis(comments):
+    r"""
+    Reduce case diff comments down to a single line synopsis so that we can put
+    something in the TestStatus file
+
+    >>> get_ts_synopsis('')
+    ''
+    >>> get_ts_synopsis('big error')
+    'big error'
+    >>> get_ts_synopsis('big error\n')
+    'big error'
+    >>> get_ts_synopsis('stuff\n    File foo had no compare counterpart in bar with suffix baz\nPass\n')
+    'ERROR BFAIL some baseline files were missing'
+    >>> get_ts_synopsis('stuff\n    File foo did NOT match bar with suffix baz\nPass\n')
+    'DIFF'
+    >>> get_ts_synopsis('stuff\n    File foo did NOT match bar with suffix baz\n    File foo had no compare counterpart in bar with suffix baz\nPass\n')
+    'DIFF'
+    """
+    if not comments:
+        return ""
+    elif "\n" not in comments.strip():
+        return comments.strip()
+    else:
+        has_bfails = False
+        has_real_fails = False
+        for line in comments.splitlines():
+            if "no compare counterpart" in line:
+                has_bfails = True
+            elif "did NOT match" in line:
+                has_real_fails = True
+
+        if has_real_fails:
+            return "DIFF"
+        elif has_bfails:
+            return "ERROR {} some baseline files were missing".format(TEST_NO_BASELINES_COMMENT)
+        else:
+            return ""
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -463,7 +463,8 @@ def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
 def get_ts_synopsis(comments):
     r"""
     Reduce case diff comments down to a single line synopsis so that we can put
-    something in the TestStatus file
+    something in the TestStatus file. It's expected that the comments provided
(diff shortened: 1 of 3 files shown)


jgfouca:
@billsacks , sorry for the slow response time.

rljacob:
If a test against baselines is done and some files show non-BFB results while other files are missing, what is the result of the test?

jgfouca:
@rljacob , DIFF

jgfouca:
This will impact some of the output in TestStatus.log. It won't impact TestStatus or dashboards.

billsacks:
@jgfouca Thanks for taking this on. I'm confused, though, about what effect this PR has on the TestStatus file (which is what I care about most). Can you tell me what the BASELINE line of the TestStatus file will now look like in these three cases?:

1. The given baseline directory exists, but there are no history files there. (What I'd like: BFAIL on the BASELINE line.)

2. The given baseline directory exists. There are some history files there, but not all that there should be. The history files there are bit-for-bit with the new test. (What I'd like: BFAIL on the BASELINE line.)

3. The given baseline directory exists. There are some history files there, but not all that there should be. At least one history file is _not_ bit-for-bit with the new test. (What I'd like: FAIL, but _no_ BFAIL on the BASELINE line.)

(See https://github.com/ESMCI/cime/issues/789#issue-188320518 for rationale.)

jgfouca:
@billsacks , as far as I know, BFAIL was never being added to TestStatus, it was only ever being added to TestStatus.log.

jgfouca:
I suppose that would be nice, wouldn't it (having it there).

billsacks:
@jgfouca okay, so BFAIL still doesn't appear in TestStatus? That's still an improvement, but it doesn't actually fix #789 , so we should keep that issue open.

jgfouca:
I don't think it's hard to do, just a moment.

jgfouca:
I was wrong, BFAIL is produced in TestStatus if the baseline dir is entirely missing.

billsacks:
For what it's worth: The behavior I was asking for in #789 (and https://github.com/ESMCI/cime/pull/2744#issuecomment-412208219) is what used to happen (I'm pretty sure) in cime4, with `component_compgen_baseline.sh`.

I'm okay with either this PR being merged as is but #789 staying open (because at least it's a step in the right direction), or this PR staying open until it adds the BFAIL status to the TestStatus file.

jgfouca:
@billsacks , I've already made the changes to make it appear in TestStatus. Testing now.

jgfouca:
@billsacks , ok, this should do what you want now.

jgfouca:
I guess the question that remains is if this functionality is important enough to add regression tests for it.

billsacks:
@jgfouca thanks. Referring back to my list from earlier:

> 1. The given baseline directory exists, but there are no history files there. (What I'd like: BFAIL on the BASELINE line.)
> 
> 2. The given baseline directory exists. There are some history files there, but not all that there should be. The history files there are bit-for-bit with the new test. (What I'd like: BFAIL on the BASELINE line.)
> 
> 3. The given baseline directory exists. There are some history files there, but not all that there should be. At least one history file is _not_ bit-for-bit with the new test. (What I'd like: FAIL, but _no_ BFAIL on the BASELINE line.)
> 

My impression from reading the code is that (1) and (2) are correct now, but not (3). I have taken a stab at reworking this to satisfy (3): see https://github.com/ESMCI/cime/pull/2748 . Note that the changes there are not yet tested (manually or via unit tests), but I think they're at least close to what we'd want. (I'm tired, though, so it's quite possible that I made some mistakes.) I'm curious what you think of that.

jgfouca:
@billsacks , your impression is correct. Let me implement that.

billsacks:
@jgfouca Thanks for this. I'm happy with the general idea here, but I'm concerned about the tight coupling between `get_ts_synopsis` and the exact format of the strings in `_compare_hists`: someone might make a slight tweak to the wording in `_compare_hists` which would break `get_ts_synopsis`. I'd be more comfortable if there were one (or both) of these:

1. String constants that were put into the comment string, with `get_ts_synopsis` looking for these constants rather than a hard-coded string. I think this would make it more obvious that the two were coupled in this way. e.g., if you want to look for `no compare counterpart`, then that should be a string constant. Even better, though, would be to have some all-caps code that would be extremely unlikely to appear by chance, like putting `TEST_NO_BASELINES_COMMENT` in the comment string.

2. A unit test that runs the combination of `_compare_hists` and `get_ts_synopsis` to ensure that they continue to work together as intended.

jgfouca:
@jedwards4b , this PR has changed a bit since your review.

scripts/lib/CIME/hist_utils.py
```diff
@@ -455,3 +459,41 @@ def generate_baseline(case, baseline_dir=None, allow_baseline_overwrite=False):
                                                get_timestamp(timestamp_format="%Y-%m-%d_%H:%M:%S")))
 
     return True, comments
+
+def get_ts_synopsis(comments):
```
billsacks: Can you make it clear, through the function name and/or documentation, that this is meant to be called on the comments returned from `compare_baseline`? (What about the comments returned from `compare_test`? My sense is that this would work there, too, but that's not its intended use.)

jgfouca:
@billsacks , I added a comment. Since get_ts_synopsis adds things like BFAIL, it makes no sense to pass in commands from compare_test.

I tested this PR for 4 cases: baseline pass, baseline diff, baseline missing his, no baseline.

PR #2745, opened 2018-08-09
Do a better job of handling exceptions in sub-phases of run

### LINKED ISSUE #2706: Failure in _st_archive_case_test reported as RUN FAIL
At least for some types of failures in the `_st_archive_case_test` that is now run as part of every system test, the failure is incorrectly reported as `FAIL... RUN` rather than `FAIL... SHORT_TERM_ARCHIVER`, and the `SHORT_TERM_ARCHIVER` phase does not even appear in the TestStatus file. This is confusing when trying to understand why a test has failed.

You can reproduce this by:

1. checking out cime5.6.9

2. Introducing the following diffs:

```diff
diff --git a/src/drivers/mct/cime_config/config_archive.xml b/src/drivers/mct/cime_config/config_archive.xml
index c871741..5563695 100644
--- a/src/drivers/mct/cime_config/config_archive.xml
+++ b/src/drivers/mct/cime_config/config_archive.xml
@@ -1,19 +1,2 @@
 <components version="2.0">
-  <comp_archive_spec compname="drv" compclass="cpl">
-    <rest_file_extension>r</rest_file_extension>
-    <hist_file_extension>hi?\d*.*\.nc$</hist_file_extension>
-    <rest_history_varname>unset</rest_history_varname>
-    <rpointer>
-      <rpointer_file>rpointer$NINST_STRING.drv</rpointer_file>
-      <rpointer_content >$CASE.cpl$NINST_STRING.r.$DATENAME.nc</rpointer_content>
-    </rpointer>
-    <test_file_names>
-      <tfile disposition="move">cpl_0001.log.5548574.chadmin1.180228-124723.gz</tfile>
-      <tfile disposition="copy">casename.cpl.r.1976-01-01-00000.nc</tfile>
-      <tfile disposition="copy">rpointer.drv_0001</tfile>
-      <tfile disposition="copy">rpointer.drv</tfile>
-      <tfile disposition="ignore">casenamenot.cpl.r.1976-01-01-00000.nc</tfile>
-    </test_file_names>
-  </comp_archive_spec>
-
 </components>
```

3. Running `SMS.f19_g17.X.cheyenne_intel` (or a similar test)

The resulting TestStatus file shows:

```
PASS SMS.f19_g17.X.cheyenne_intel CREATE_NEWCASE
PASS SMS.f19_g17.X.cheyenne_intel XML
PASS SMS.f19_g17.X.cheyenne_intel SETUP
PASS SMS.f19_g17.X.cheyenne_intel SHAREDLIB_BUILD time=120
PASS SMS.f19_g17.X.cheyenne_intel MODEL_BUILD time=21
PASS SMS.f19_g17.X.cheyenne_intel SUBMIT
FAIL SMS.f19_g17.X.cheyenne_intel RUN time=42
PASS SMS.f19_g17.X.cheyenne_intel MEMLEAK
```

and TestStatus.log shows:

```
2018-07-07 06:57:03: Exception during run:
ERROR: No archive entry found for components: ['CPL']
  File "/glade/p/work/sacks/cime/scripts/Tools/../../scripts/lib/CIME/SystemTests/system_tests_common.py", line 155, in run
    self._st_archive_case_test()
  File "/glade/p/work/sacks/cime/scripts/Tools/../../scripts/lib/CIME/SystemTests/system_tests_common.py", line 284, in _st_archive_case_test
    result = self._case.test_env_archive()
  File "/glade/p/work/sacks/cime/scripts/Tools/../../scripts/lib/CIME/case/case_st_archive.py", line 749, in test_env_archive
    expect(not components, "No archive entry found for components: {}".format(components))
  File "/glade/p/work/sacks/cime/scripts/Tools/../../scripts/lib/CIME/utils.py", line 130, in expect
    raise exc_type(msg)
SystemExit: ERROR: No archive entry found for components: ['CPL']
```

I forget the design of TestStatus in this respect. Is the solution to move the `with self._test_status` line in `_st_archive_case_test` to the top of that method??? If so, should the same be done in `_check_for_memleak`???

jgfouca:
Exceptions coming from sub-phases of the run phase (phases that happen during
SystemTestCommon.run other than that RUN phase) were causing RUN to be
reported as FAIL instead of the sub-phase.

This PR introduces a pattern of handling such phases with more robust
exception handling.

Also, adds a test for this.

Also, splits the 3 phases that were happening under _compare_baseline to
their own functions.

This implementation allows subclasses of SystemTestCommon to override
these sub-phase functions without having to worry about exception
handling.

Test baseline: 
Test namelist changes: 

Fixes #2706 

User interface changes?: N

### DIFF
--- a/config/config_tests.xml
+++ b/config/config_tests.xml
@@ -186,6 +186,10 @@ TESTRUNFAIL       Insta-fail run step. Used to confirm that model run
 TESTRUNFAILEXC    Insta-fail run step via exception. Used to test correct
                   correct behavior when exceptions are generated.
 
+TESTRUNSTARCFAIL  Insta-fail st_archive step via exception. Used to test correct
+                  correct behavior when exceptions are generated within in run_phase
+                  for phases other than RUN.
+
 TESTRUNPASS       Insta-pass run step. Used to test that run that work
                   are reported correctly.
 
@@ -391,6 +395,15 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
 
   <test NAME="TESTRUNFAIL" INFRASTRUCTURE_TEST="TRUE">
     <DESC>For testing infra only. Insta-fail run step.</DESC>
+    <INFO_DBUG>1</INFO_DBUG>
+    <STOP_OPTION>ndays</STOP_OPTION>
+    <STOP_N>11</STOP_N>
+    <CHECK_TIMING>FALSE</CHECK_TIMING>
+    <DOUT_S>FALSE</DOUT_S>
+  </test>
+
+  <test NAME="TESTRUNSTARCFAIL" INFRASTRUCTURE_TEST="TRUE">
+    <DESC>For testing infra only. Insta-fail st archive test with exception.</DESC>
     <INFO_DBUG>1</INFO_DBUG>
     <STOP_OPTION>ndays</STOP_OPTION>
     <STOP_N>11</STOP_N>
--- a/config/config_tests.xml
+++ b/config/config_tests.xml
@@ -187,7 +187,7 @@ TESTRUNFAILEXC    Insta-fail run step via exception. Used to test correct
                   correct behavior when exceptions are generated.
 
 TESTRUNSTARCFAIL  Insta-fail st_archive step via exception. Used to test correct
-                  correct behavior when exceptions are generated within in run_phase
+                  behavior when exceptions are generated within in run_phase
                   for phases other than RUN.
 
 TESTRUNPASS       Insta-pass run step. Used to test that run that work
--- a/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -145,15 +145,16 @@ class SystemTestsCommon(object):
             self.run_phase()
 
             if self._case.get_value("GENERATE_BASELINE"):
-                self._generate_baseline()
+                self._phase_modifying_call(GENERATE_PHASE, self._generate_baseline)
 
             if self._case.get_value("COMPARE_BASELINE"):
-                self._compare_baseline()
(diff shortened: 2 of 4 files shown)


jgfouca:
@jedwards4b , @billsacks , I think this PR gets us into a pretty good state regarding the interaction between exceptions and the TestStatus file.

My remaining concern is with st_archiving. The phase is being set by st_archive_case_test, which is a bit confusing as I would expect to be set by the case_st_archive. Also, exceptions coming from case_st_archive will continue to appear as RUN FAIL since it's tightly coupled within an individual run.

jedwards4b:
I'm open to any changes you want to make here.  I put it in without a very good grasp of how the phases work.   It could be a prerun phase like NLCOMP is.

config/config_tests.xml
```diff
@@ -186,6 +186,10 @@ TESTRUNFAIL       Insta-fail run step. Used to confirm that model run
 TESTRUNFAILEXC    Insta-fail run step via exception. Used to test correct
                   correct behavior when exceptions are generated.
 
+TESTRUNSTARCFAIL  Insta-fail st_archive step via exception. Used to test correct
+                  correct behavior when exceptions are generated within in run_phase
```
billsacks: duplicate "correct"

jgfouca: Fixed, thx.

billsacks:
@jgfouca - Thanks a lot for this: it's good to have this more robust and consistent mechanism in place. I'm confused, though, based on https://github.com/ESMCI/cime/pull/2745#issuecomment-411594828 - does this actually fix #2706 or will the st_archive phase still be incorrectly reported as a run failure?

jgfouca:
@billsacks , yes, I used TESTRUNSTARCFAIL to confirm that fix.

jgfouca:
Misclicked "Close and comment"

billsacks:
Thanks for the clarification.

PR #2750, opened 2018-08-13
Disable problematic task-node output by default

jgfouca:
This code doesn't compile within the unit testing framework
for mct (cime/scripts/fortran_unit_testing/run_tests.py).

Test suite: scripts_regression_tests N_TestUnitTest
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @worleyph


jgfouca:
@worleyph , it wouldn't let me assign you as a reviewer, but I'd like you to sign off on this.

worleyph:
@jgfouca , why does it fail? Don't know why a TIMING ifdef makes sense for enabling/disabling it.

worleyph:
Is the problem solely with the calls to t_startf/t_stopf?

jedwards4b:
@worleyph Because the unit tests don't include the perf_mod or gptl and don't know about the new shr_taskmap_write file you added.

worleyph:
So, I don't know what the goal is if these are not being tested, but if we always build the model with TIMING enabled, this is fine by me. @jgfouca , you might add a comment within the ifdef describing why these are ifdef'ed.

rljacob:
What does the unit testing framework test within seq_comm_mct?

jedwards4b:
@worleyph the unit tests aren't there to test the timing code, however if you would like to add that to the unit test framework that's another approach that would work.

jedwards4b:
@rljacob I'm not sure but I think that seq_comm_mct is not tested but needed for other code that is.

worleyph:
@jedwards4b , thanks. I'm asking questions without any understanding of what is going on here. If TIMING is excluded in the testing, this makes sense. The shr_taskmap_write call is not timing, but if irrelevant to what you are trying to test, then including it with the TIMING exclusion is fine with me.

jgfouca:
@jedwards4b , just grepping around, it's pretty clear that pio build uses -DTIMING, but it's not as clear to me that mct does.

jedwards4b:
I am confused about the status of testing this PR - @jgfouca can you clarify.

jgfouca:
@jedwards4b , I tested N_TestUnitTest

jgfouca:
I did not test if this code is enabled during a normal run, I'll do that.

worleyph:
@jgfouca , output is near the top of the e3sm.log (pretty much first model output after the PIO options description, though some systems output system environment variable settings first). An F case would be sufficient as well.

Look for something that looks like

48 pes participating in computation of coupled model
--------------------------------------------------------------
GLOBAL communicator : 2 nodes, 48 MPI tasks
COMMUNICATOR NODE # [NODE NAME] : (# OF MPI TASKS) TASK # LIST
GLOBAL NODE 0 [ b634 ] : ( 36 MPI TASKS ) ...
GLOBAL NODE 1 [ b635 ] : ( 12 MPI TASKS ) ...

jgfouca:
@worleyph , i'm testing within cime standalone, so I only have the basic compsets available (A, X, etc).

jedwards4b:
I don't see any reason you wouldn't see it with an A case.

jgfouca:
OK, I've pushed a commit that appears to work. N_TestUnitTest doesn't have a build failure and I've confirmed that a simple test case still has the task-node map output.

@jedwards4b , please re-review. I had to touch the main Makefile. The unit tests use CMake so they don't see this flag.

jedwards4b:
I don't understand why you did this instead of ifdef timing?

jgfouca:
@jedwards4b , I checked, mct doesn't use TIMING

jgfouca:
And I wasn't sure I wanted to turn on TIMING for everything because that could impact pio.

jedwards4b:
TIMING is on by default but is currently only used as a macro in PIO.   I guess that the unit tests didn't incorporate any perf_mod until now.   It would work to use ifdef TIMING here but you would then need to add it to the Makefile for the csm_share build.   I still think that may be better than creating a new ifdef term in the Makefile.

jgfouca:
@jedwards4b , OK , let me try that.

jgfouca:
@jedwards4b , done. Thanks for the help, I was a bit lost in our build systems.

src/drivers/mct/shr/seq_comm_mct.F90
```diff
@@ -20,8 +20,10 @@ module seq_comm_mct
   use shr_sys_mod    , only : shr_sys_abort, shr_sys_flush
   use shr_mpi_mod    , only : shr_mpi_chkerr, shr_mpi_bcast, shr_mpi_max
   use shr_file_mod   , only : shr_file_getUnit, shr_file_freeUnit
+#ifdef TIMING
```
rljacob: Add a comment.  This is for unit testing.  TIMING is on by defaut when compiling cime normally.  Unit tests for Fortran don't need gptl AND have TIMING undefined so add this ifdef for unit tests.

PR #2755, opened 2018-08-16
Extract code for creating a test suite-specific cs.status script

billsacks:
This will let us write a command-line utility that leverages this to
create custom versions of this script. I'll post an example of how I'm
planning to use this in a follow-up comment.

Test suite: scripts_regression_tests on cheyenne, B_CheckCode on my mac,
   and manual testing to ensure that the name and contents of the
   cs.status file are the same as before, with `./create_test --no-build
   SMS.f19_g16.X.roo2_gnu --test-root $scratch/test1 --test-id abc123`
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes none

User interface changes?: none

Code review:

### DIFF
--- a/scripts/lib/CIME/cs_status_creator.py
+++ b/scripts/lib/CIME/cs_status_creator.py
@@ -1,0 +1,40 @@
+"""
+Creates a test suite-specific cs.status file from a template
+"""
+
+from CIME.XML.standard_module_setup import *
+import CIME.utils
+import os
+import stat
+
+def create_cs_status(test_root, test_id, filename=None):
+    """Create a test suite-specific cs.status file from the template
+
+    Arguments:
+    test_root (string): path to test root; the file will be put here. If
+        this directory doesn't exist, it is created.
+    test_id (string): test id for this test suite. This can contain
+        shell wildcards if you want this one cs.status file to work
+        across multiple test suites. However, be careful not to make
+        this too general: for example, ending this with '*' will pick up
+        the *.ref1 directories for ERI and other tests, which is NOT
+        what you want.
+    filename (string): name of the generated cs.status file. If not
+        given, this will be built from the test_id.
+    """
+    python_libs_root = CIME.utils.get_python_libs_root()
+    cime_root = CIME.utils.get_cime_root()
+    template_file = os.path.join(python_libs_root, "cs.status.template")
+    template = open(template_file, "r").read()
+    template = template.replace("<PATH>",
+                                os.path.join(cime_root,"scripts","Tools")).replace\
+                                ("<TESTID>", test_id).replace\
+                                ("<TESTROOT>", test_root)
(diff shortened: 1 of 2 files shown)


billsacks:
This is part of my work to streamline the process of running and checking results from the CTSM test suite, which could be applicable to other large test suites as well. One issue we have is that we run separate create_test jobs for each compiler on a given machine, which results in separate cs.status files. This means that you need to run and parse the output from each cs.status file. What I'd like is to have a single cs.status file that aggregates across all of the test suites you just ran, and this refactor gives us a mechanism to create a custom cs.status file that does that. I considered trying to come up with a more general solution like we used to have with cimeteststatus, but that looked tricky. The solution here will involve CTSM's test-running scripts creating a custom cs.status file, which is possible because those test-running scripts know what test suites were just run. (The individual cs.status.* scripts will still be there, too - but we can add an extra, custom one.)

Here is an example showing what I envision:

```
[roo2:~/cime/scripts]$ ./create_test --no-build SMS.f19_g16.X.roo2_gnu --test-root $scratch/test3 --test-id def456_g
[roo2:~/cime/scripts]$ ./create_test --no-build SMS.f19_g16.X.roo2_gnu --test-root $scratch/test3 --test-id def456_i
```

Then running this script:

```python
#!/usr/bin/env python
import sys

sys.path.append('/Users/sacks/cime/scripts/lib')

from CIME.cs_status_creator import create_cs_status

create_cs_status(test_root='/Users/sacks/projects/scratch/test3',
                 test_id='def456_?',
                 filename='cs.status')
```

This gives a file called cs.status, with contents:

```
[roo2:~/projects/scratch/test3]$ cat cs.status
#! /bin/bash
/Users/sacks/cime/scripts/Tools/cs.status "$@" /Users/sacks/projects/scratch/test3/*.def456_?/TestStatus
```

When I run it I get:

```
[roo2:~/projects/scratch/test3]$ ./cs.status
def456_g
  SMS.f19_g16.X.roo2_gnu (Overall: PEND) details:
    PASS SMS.f19_g16.X.roo2_gnu CREATE_NEWCASE
    PASS SMS.f19_g16.X.roo2_gnu XML
    PASS SMS.f19_g16.X.roo2_gnu SETUP
    PEND SMS.f19_g16.X.roo2_gnu SHAREDLIB_BUILD
def456_i
  SMS.f19_g16.X.roo2_gnu (Overall: PEND) details:
    PASS SMS.f19_g16.X.roo2_gnu CREATE_NEWCASE
    PASS SMS.f19_g16.X.roo2_gnu XML
    PASS SMS.f19_g16.X.roo2_gnu SETUP
    PEND SMS.f19_g16.X.roo2_gnu SHAREDLIB_BUILD
```

From the perspective of a user of the CTSM test suite, all they need to know is to run the single 
 `cs.status` file in their testroot directory.

PR #2761, opened 2018-08-20
Allow TestStatus file to be copied into baselines

### LINKED ISSUE #2760: Copy TestStatus file to the baseline directory
I often find myself wanting to know if a given test passed or failed (entirely or some specific phase) in a previous run of the test suite, for which we have baselines. For example, this would be a quick way of seeing if a test failed the last time around, or it can help indicate if / why there might be problems with the baselines for a given test.

So it would be helpful if the TestStatus file were copied to the baseline directory when running a test suite with the `--generate` flag.

There are a few things I'm not sure of in terms of how this should be implemented:

1. If we copy the TestStatus file in the generate phase (which seems like a logical time to do it), it would be missing some information, since generate comes before some other phases. Would it be acceptable to move the generate phase to be the very last phase, or would that have some negative repercussions?

2. Alternatively, we could copy the TestStatus file as a separate step (probably not even a named phase) at the very end of system_tests_common:run. Would that be reasonable?

3. Is there a way to ensure that this copying of the TestStatus file happens even if there is a failure in the run or some other phase? e.g., would putting this copying in a `finally` block at the end of system_tests_common:run ensure this?

I'm happy to make and test this change myself, but I'd like to get input from @jgfouca , @jedwards4b or others if you have any immediate thoughts. My inclination would be to have a `finally` block at the end of system_tests_common:run that does this copying, so I'd probably try that approach first unless you have other feelings.

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2760 

User interface changes?: N

### DIFF
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -397,7 +397,7 @@ def generate_teststatus(testdir, baseline_dir):
             if not os.path.isdir(baseline_dir):
                 os.makedirs(baseline_dir)
 
-            safe_copy(os.path.join(test_dir, TEST_STATUS_FILENAME), baseline_dir)
+            safe_copy(os.path.join(testdir, TEST_STATUS_FILENAME), baseline_dir)
         except Exception, e:
             logger.warning("Could not copy {} to baselines, {}".format(os.path.join(testdir, TEST_STATUS_FILENAME), str(e)))
 
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -3,7 +3,7 @@ Functions for actions pertaining to history files.
 """
 
 from CIME.XML.standard_module_setup import *
-from CIME.test_status import TEST_NO_BASELINES_COMMENT
+from CIME.test_status import TEST_NO_BASELINES_COMMENT, TEST_STATUS_FILENAME
 from CIME.utils import get_current_commit, get_timestamp, get_model, safe_copy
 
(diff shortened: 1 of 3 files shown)


jgfouca:
@billsacks , I don't have access to a cesm machine, so you might want to try this out.

scripts/lib/CIME/test_scheduler.py
```diff
@@ -921,6 +921,15 @@ def run_tests(self, wait=False,
 
         expect(threading.active_count() == 1, "Leftover threads?")
 
+        # Copy TestStatus files to baselines for tests that have already failed.
```
billsacks: For my own edification: When does this code get invoked in the whole test process?

scripts/lib/CIME/hist_utils.py
```diff
@@ -387,6 +387,20 @@ def get_extension(model, filepath):
 
     return result
 
+def generate_teststatus(testdir, baseline_dir):
```
billsacks: This function seems kind of out of place in hist_utils, but I'm willing to let it go if you want it here.

jgfouca:
@billsacks , I guess I was making assumptions about the workflow here that were not accurate. I assumed you wanted this as part of a procedural run, not individual tests.

I would expect build failures to appear unless you did the build by hand with case.build.

scripts/lib/CIME/test_scheduler.py
jgfouca: At the end of create_test. So, on a batch machine, once all the tests are built and submitted.

PR #2769, opened 2018-08-23
Maint 5.6

jedwards4b:
Merge maint-5.6 to master

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


fischer-ncar:
@jedwards4b looks good except for the merge conflicts.

jedwards4b:
Right, I have a branch with them resolved that I can push when ready.

scripts/lib/CIME/XML/component.py
```diff
@@ -282,3 +282,36 @@ def print_values(self):
             name = self.get(entry, "id")
             text = self.text(self.get_child("desc", root=entry))
             logger.info("   {:20s} : {}".format(name, text.encode('utf-8')))
+
+    def return_values(self):
```
jgfouca: Hey @jedwards4b , i think I remember deleting these "return_values" methods because they were not being used. Was I wrong? git grep confirms nothing in cime is using them.

jgfouca:
I approve pending question.

scripts/lib/CIME/XML/component.py
jedwards4b: I think that they are used by some web interface code outside of cime: @bertinia?

jgfouca: OK, we should add a comment somewhere since checkers and manual inspection will flag these methods as dead code.

Issue #2770, opened 2018-08-23
Should we remove the cimeteststatus script?


billsacks:
As far as I can tell, scripts/Tools/cimeteststatus has been broken for a long time (maybe ever since the move to cime5). It used to be a very nice script, and it may still serve some purpose as a reference for some features we may want to bring back at some point, but it looks like it would need an overhaul to work with the much improved test_status design of cime5.

Does anyone object to its removal?

jgfouca:
I don't think we (e3sm) were ever aware of or users of this script.

billsacks:
@fischer-ncar do you know of (or use) this script?

fischer-ncar:
I do know of this script, and have no objections to it's removal.  I believe if was a effort by Jay to combine
cs.status and testreporter into one script.

billsacks:
Okay, I'll remove it.

PR #2775, opened 2018-08-24
Add options to cs.status to allow just summarizing failures

### LINKED ISSUE #2749: Should we support one or more scripts for quickly parsing output from large test suites?
We're trying to get more CTSM developers to run the full aux_clm test suite. One aspect of making this easier is providing a robust, easy-to-use mechanism for parsing the results of a large test suite to determine what has failed.

I wrote a simple script a few years ago that I use all the time for this purpose (https://github.com/billsacks/bin/blob/master/filter_cs_status). My workflow is (1) Run the appropriate cs.status script(s) for the test suite I ran; (2) Run `filter_cs_status` on the output from each. (I like having these two steps separated, because (1) can take a while on large test suites, and I often rerun (2) a few times for each test suite, excluding various things each time.) [Update 2018-08-14: cs.status seems to run fast enough that these don't really need to be separated. And for the sake of new users, it would be nice if this were built into cs.status.]

It's basically a glorified `grep -v`, with the added benefits that (1) There are a few hard-coded patterns that it always ignores (lines with 'Overall...' and 'PASS'; I also generally ignore TPUTCOMP and MEMCOMP, but that wouldn't need to be the case); (2) It prints counts of the number of lines matching each ignore pattern (so, for example, you can see if _everything_ had a TPUTCOMP failure, or just a few tests).

I'm wondering if I should put this script in cime. (I'd be happy to rewrite it in python.) Would others find this useful? If not, is there another, similar script that is already supported or should be supported for this purpose? There's the old cimeteststatus script, but my sense is that this no longer works with cime5. (Should we delete the old cimeteststatus script?)

billsacks:
This PR adds some functionality to cs.status (as well as the TestStatus
class that it relies on) to allow users to just see a brief summary of
test failures. Specifically, it adds three new arguments:

(1) -f / --fails-only: With this argument, PASS lines (as well as the
    "Overall" line for each test) are suppressed, allowing you to
    quickly focus in on the failures.

(2) -c PHASE / --count-fails PHASE: This argument can be specified
    multiple times. For each phase specified in this way, cs.status does
    not give line-by-line output of either passes or fails, but instead
    it keeps a running count of all non-passes and reports this count at
    the end.

(3) --count-performance-fails: This is simply a short-hand for
    '--count-fails TPUTCOMP --count-fails MEMCOMP', which is a common
    option that we want in CTSM testing.

In addition, this PR makes the following changes:

(4) Regardless of what options are specified, cs.status now outputs a
    count of the total number of tests for each test id. This is
    especially useful with the --fails-only option (to get a quick
    confirmation that it's summarizing the number of tests you expect),
    but it seems like a generally-useful addition, so I'm including it
    whether or not you're running with --fails-only.

(5) Adds unit tests for the cs_status implementation (and thus for some
    pieces of TestStatus as well).

(6) Cleans up help output from cs.status to make it more consistent with
    the help output from other scripts. (Note: This does not add the
    common logging options, because there is a collision with the
    current '-s' argument to cs.status. I don't use the '-s' argument
    myself, but I didn't want to break other people's usage of this
    argument.)

(7) Modifies cs_status_creator to allow custom cs.status scripts to add
    extra arguments. For example, for CTSM, we'll create a custom script
    in the testroot directory that adds '-f --count-performance-fails'.

(8) Removes the unused and broken cimeteststatus script (see #2770)

Note that, other than some additional optional arguments and the minor
change in (4), the operation of cs.status is unchanged from before.

Here is an example demonstrating the new cs.status functionality:

```
[cheyenne5:/glade/scratch_old/sacks/tests_0820_1540c]$ /glade/work/sacks/cime/scripts/Tools/cs.status *.0820_1540c_i/TestStatus -f --count-performance-fails
0820_1540c_i: 156 tests
    FAIL ERP_D_Ld10_P36x2.f10_f10_musgs.IHistClm50BgcCrop.cheyenne_intel.clm-ciso_decStart RUN time=108
    FAIL ERP_D_Ld5.f09_g17.I2000Clm50Vic.cheyenne_intel.clm-vrtlay RUN time=23


========================================================================
Non-PASS results for select phases:
TPUTCOMP non-passes: 3
MEMCOMP non-passes: 22
```

Test suite: scripts_regression_tests on cheyenne, B_CheckCode on my mac
(using pylint v. 2.1.1), plus manual testing. Fails Z_FullSystemTest due to a
failure in DAE.f19_f19.A.cheyenne_intel, but this is failing on master, too.
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes #2749

User interface changes?: Adds some optional arguments to cs.status, but
no other changes

Code review:


billsacks:
This failed Z_FullSystemTest due to a failure in DAE.f19_f19.A.cheyenne_intel, but this is failing on master, too, according to CDash.

billsacks:
@fischer-ncar I didn't add you as a reviewer, but feel free to look at this and review it if you'd like.

billsacks:
I see that there are some Codacy failures, due to methods that could be functions. This was a conscious design choice: I prefer to make things methods in order to (a) keep related code grouped together, and (b) aid future maintenance / refactoring, in case the method starts to need self for something. But I don't feel strongly about this, and am fine with making these functions if others agree with Codacy's critiques.

scripts/Tools/cs.status
```diff
@@ -1,80 +1,100 @@
...
+    ./cs.status /path/to/testroot/*.testid/TestStatus
+
+Returns True if no errors occured (not based on test statuses).
 """
 
 from standard_script_setup import *
 import argparse, sys, os, logging, glob
-from CIME.test_status import *
+from CIME.utils import expect
+from CIME.cs_status import cs_status
```
jedwards4b: I seem to recall there was a reason not to give files and classes the same name, but don't remember what it was - do you?

fischer-ncar:
Thanks for doing this @billsacks.  I have no issues with the changes.

scripts/Tools/cs.status
billsacks: I can't remember, but I can rename either the module or the function if you think I should.

jedwards4b: After I made the comment I realized that it's a function and not a class so I think this is okay.

billsacks:
I just pushed a commit changing the offending methods in Codacy to staticmethod. (I think that's more appropriate than classmethod here.) I reran the unit tests in that module and reran code_checker on that module.

scripts/lib/CIME/tests/test_cs_status.py
```diff
@@ -0,0 +1,315 @@
...
+        with test_status.TestStatus(test_dir=test_dir_path,
+                                    test_name=test_name) as ts:
+            ts.set_status(phase, status)
+
+    # ------------------------------------------------------------------------
+    # Begin actual tests
+    # ------------------------------------------------------------------------
+
+    def test_single_test(self):
+        """cs_status for a single test should include some minimal expected output"""
```
jgfouca: I see you added a bunch of tests but they don't appear to be getting called from scripts_regression_tests.

billsacks: These tests are picked up via python's unittest discovery mechanism. I have verified that they are run by inserting this change:

```diff
diff --git a/scripts/lib/CIME/tests/test_cs_status.py b/scripts/lib/CIME/tests/test_cs_status.py
index 7d3fba7e9..dea2176b2 100644
--- a/scripts/lib/CIME/tests/test_cs_status.py
+++ b/scripts/lib/CIME/tests/test_cs_status.py
@@ -208,6 +208,7 @@ def test_single_test(self):
         cs_status([os.path.join(test_dir_path, 'TestStatus')],
                   out=self._output)
         self.assert_core_phases(self._output.getvalue(), test_name, fails=[])
+        self.fail()

     def test_two_tests(self):
         """cs_status for two tests (one with a FAIL) should include some minimal expected output"""
```

and verifying that `./scripts_regression_tests.py A_RunUnitTests` leads to a failure.

jgfouca: Great. I thought that might be the case.

PR #2780, opened 2018-08-24
Make a new grid to ensure ADWAV works with f19_f19

### LINKED ISSUE #2779: DAE.f19_f19.ADWAV fails during phase SETUP on anlworkstation
PR #2777 has changed DAE.f19_f19.A in cime_developer list to DAE.f19_f19.ADWAV

However, DAE.f19_f19.ADWAV fails on anlworkstation:
```
        ERROR: WAV_GRID cannot be null

FAIL DAE.f19_f19.ADWAV.anlworkstation_gnu (phase SETUP)
```
Could you please check if WAV_GRID needs to be set for DAE.f19_f19.ADWAV on anlworkstation (not sure if other machines have this issue)?

jgfouca:
This grid is no different than f19_f19 for cesm, but it's still necessary because the special grid is needed for e3sm.

Test suite: by-hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2779 

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/config/e3sm/config_grids.xml
+++ b/config/e3sm/config_grids.xml
@@ -481,16 +481,6 @@
       <mask>gx1v6</mask>
     </model_grid>
 
-    <model_grid alias="f19_f19_ww3a" compset="(DOCN|XOCN|SOCN|AQP1)">
-      <grid name="atm">1.9x2.5</grid>
-      <grid name="lnd">1.9x2.5</grid>
-      <grid name="ocnice">1.9x2.5</grid>
-      <grid name="rof">r05</grid>
-      <grid name="glc">null</grid>
-      <grid name="wav">ww3a</grid>
-      <mask>gx1v6</mask>
-    </model_grid>
-
     <model_grid alias="f25_f25" compset="(DOCN|XOCN|SOCN|AQP1)">
       <grid name="atm">2.5x3.33</grid>
       <grid name="lnd">2.5x3.33</grid>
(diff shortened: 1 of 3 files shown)


config/e3sm/config_grids.xml
```diff
@@ -481,6 +481,16 @@
...
+    <model_grid alias="f19_f19_ww3a" compset="(DOCN|XOCN|SOCN|AQP1)">
+      <grid name="atm">1.9x2.5</grid>
+      <grid name="lnd">1.9x2.5</grid>
+      <grid name="ocnice">1.9x2.5</grid>
+      <grid name="rof">r05</grid>
+      <grid name="glc">null</grid>
+      <grid name="wav">ww3a</grid>
+      <mask>gx1v6</mask>
+    </model_grid>
+
```
gold2718: My recommendation is:
If you are going to add a new grid, how about the one I posted (the ww3a grid)? The only non-stub component in the ADWAV compset is DWAV so ww3 is the most appropriate grid for an ADWAV test.
If that is rejected:
Have you tried this new grid with AQP1?
The whole point of f19_f19 in the first place was that represented a lowest common denominator between models. Given that E3SM does not use the FV dycore (as far as I know), is there any reason to just not remove the null wav entry?
Given the issues we have run into with this test, at a minimum, I think tests should be run on both cesm and e3sm.

jgfouca: @gold2718 , thanks for the thoughtful response. I think your first option is a good one.

PR #2789, opened 2018-08-29
Do not report BFAIL for a test that is missing a file

### LINKED ISSUE #2788: BFAIL incorrectly generated if a test is missing files
If a test has both (a) a missing file in the baseline that is present in the test, and (b) a missing file in the test that is present in the baseline, this is currently reported as a BFAIL rather than a true failure.

I have a fix incoming.

billsacks:
If a test had both (a) a missing file in the baseline that was present
in the test, and (b) a missing file in the test that was present in the
baseline, this was previously reported as a BFAIL rather than a true
failure. This commit fixes this problem, reporting it as a true failure.

Test suite: scripts_regression_tests - just A_RunUnitTests and B_CheckCode
   Full scripts_regression_tests underway
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes #2788 

User interface changes?: no

Code review:

### DIFF
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -10,9 +10,17 @@ import logging, os, re, stat, filecmp
 logger = logging.getLogger(__name__)
 
 BLESS_LOG_NAME = "bless_log"
+
 NO_COMPARE     = "had no compare counterpart"
 NO_ORIGINAL    = "had no original counterpart"
 DIFF_COMMENT   = "did NOT match"
+# COMPARISON_COMMENT_OPTIONS should include all of the above: these are any of the special
+# comment strings that describe the reason for a comparison failure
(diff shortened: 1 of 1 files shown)


scripts/lib/CIME/hist_utils.py
```diff
@@ -492,6 +500,8 @@ def get_ts_synopsis(comments):
     'DIFF'
     >>> get_ts_synopsis('stuff\n    File foo did NOT match bar with suffix baz\n    File foo had no compare counterpart in bar with suffix baz\nPass\n')
     'DIFF'
+    >>> get_ts_synopsis('File foo had no compare counterpart in bar with suffix baz\n File foo had no original counterpart in bar with suffix baz\n')
```
billsacks: This new doctest fails without the other changes in this PR

jgfouca:
Looks good, thx!

PR #2800, opened 2018-09-11
Jgfouca/branch for acme split 2018 09 10

jgfouca:
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


scripts/lib/CIME/build.py
```diff
@@ -215,7 +215,7 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
         if (not os.path.exists(shared_item)):
             os.makedirs(shared_item)
     mpilib = case.get_value("MPILIB")
-    libs = ["gptl", comp_interface, "pio", "csm_share"]
+    libs = ["gptl", "mct", "pio", "csm_share"]
```
jgfouca: This seems odd. @sarich ?

sarich: Perhaps there more nuance here than I first realized. 

if --driver is used on the create_newcase command line, then COMP_INTERFACE is set to that value. In case.py:
```
self.set_lookup_value("COMP_INTERFACE", driver)
```
That value is used in build.py to build the driver:
```
        config_dir = os.path.join(cimeroot, "src", "drivers", comp_interface, "cime_config")
        run_cmd("{}/buildexe {} {} {} "
                       .format(config_dir, caseroot, libroot, bldroot), 
```

But it is also used to build external library src/externals/mct  in _build_libraries():
```
    libs = ["gptl", comp_interface, "pio", "csm_share"]
```

Now there's a confusion because the external library and the driver have the same name.
Does comp_interface refer to the driver component or the underlying library for communications or are they assumed to be the same? 

The moab driver, in it's fledgling current state, relies on the mct external library. This may get completely phased out.  Do I need to add another library src/externals/moab that is a copy of the mct library?

Link for reference: https://github.com/E3SM-Project/E3SM/pull/2513

jedwards4b:
@jgfouca here is a minor issue from travis-ci - I'll try to do a full review today or tomorrow.
scripts/lib/CIME/SystemTests/mvk.py:54:12: W1505: Using deprecated method warn() (deprecated-method)

scripts/lib/CIME/build.py
jgfouca: @jedwards4b , can you comment here?

jedwards4b: What is the error that you get when you leave the comp_interface variable here?  We do build the mct external library with the nuopc driver and the line as it was without this change.

sarich: Does the nuopc fork have a buildlib.nuopc file that builds the mct lib?

jedwards4b: Oops _ i just looked at this code in the nuopc branch and we have the same solution - this is okay.   Sorry for the confusion.

scripts/lib/CIME/SystemTests/mvk.py
```diff
@@ -0,0 +1,211 @@
...
+
+        # =================================================================
+        # Run-time settings.
+        # Do this already in build_phase so that we can check the xml and
+        # namelist files before job starts running.
+        # =================================================================
+
+        # namelist specifications for each instance
+        for iinst in range(1, ninst+1):
+            with open('user_nl_cam_{:04d}'.format(iinst), 'w') as nl_atm_file:
```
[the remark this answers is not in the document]
jgfouca: @jhkennedy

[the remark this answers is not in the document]
jhkennedy: @jedwards4b are you talking about specify the testmods on the command line using the [testname syntax](http://esmci.github.io/cime/users_guide/testing.html?highlight=testmod#testname-syntax)? If so, from what I can tell in the documentation, that's not an appropriate method for this test -- these modification are required (not optional) for any run of the test and should be transparent to the user. 

At the time, I didn't see any method for generating these name-list files and was using the pattern suggested by other CMDV-SM team members. There is a set of (completely unused) [user_nl_utils](https://github.com/ESMCI/cime/blob/master/scripts/lib/CIME/SystemTests/test_utils/user_nl_utils.py) that also follow a similar pattern. I suppose I should have abstracted this  *new method* into `user_nl_utils`, but a `with open` block in python is a pretty trivial piece of code so I didn't think about it. 

I can do that or do you have a better suggestion?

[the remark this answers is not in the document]
jedwards4b: No it's fine.  Just wanted to make sure you were aware of the available methods and had a good reason to reject them.  Thanks

PR #2803, opened 2018-09-11
Fix for ESMF_LIB versions before 8, fixed #2801

### LINKED ISSUE #2801: Problem building with ESMF_LIB
This is with cime5.7.3

I'm getting a build fail when I build using the ESMF library.

So using the following settings...

./xmlchange --force USE_ESMF_LIB=TRUE,ATM_NCPL=288,CALENDAR=GREGORIAN

WACCM-X now requires building with the ESMF library so this is a requirement for cime now.

```
Building cesm with output to /glade/scratch/erik/ERS_D_Ln9_P480x3.f19_g16.I2000Clm50SpGs.cheyenne_intel.clm-waccmx_offline.GC.ctsm10dev011chintela/bld/cesm.bldlog.180911-112617 
/gpfs/fs1/work/erik/ctsm_SSREsnowmip/cime/src/drivers/mct/main/cime_driver.F90(31): error #6580: Name in only-list does not exist.   [ESMF_LOGKIND_MULTI_ON_ERROR]

/gpfs/fs1/work/erik/ctsm_SSREsnowmip/cime/src/drivers/mct/main/cime_driver.F90(77): error #6404: This name does not have a type, and must have an explicit type.   [ESMF_LOGKIND_MULTI_ON_ERROR]

ERROR: BUILD FAIL: buildexe failed, cat /glade/scratch/erik/ERS_D_Ln9_P480x3.f19_g16.I2000Clm50SpGs.cheyenne_intel.clm-waccmx_offline.GC.ctsm10dev011chintela/bld/cesm.bldlog.180911-112617
```

ekluzek:
Changes so that cime can build with an ESMF library before version 8

This uses the pre-processor to see if an ESMF library is being used and if so, if the version
is before version 8. If it is it won't allow ESMF_LOGFILE_KIND to be set to ESMF_LOGKIND_MULTI_ON_ERROR, and will abort with an error, otherwise it will run.

Test baseline: Test a couple cases make sure it works with USE_ESMF_LIB on and off
   as well as default and
./xmlchange ESMF_LOGFILE_KIND='ESMF_LOGKIND_MULTI_ON_ERROR'
Test namelist changes: N
Test status: bit for bit
Fixes #2801

User interface changes?: N

Code review:

### DIFF
--- a/src/drivers/mct/main/cime_driver.F90
+++ b/src/drivers/mct/main/cime_driver.F90
@@ -28,12 +28,13 @@ program cime_driver
   use ESMF,          only : ESMF_Initialize, ESMF_Finalize
   use ESMF,          only : ESMF_LogKind_Flag, ESMF_LOGKIND_NONE
   use ESMF,          only : ESMF_LOGKIND_SINGLE, ESMF_LOGKIND_MULTI
-  use ESMF,          only : ESMF_LOGKIND_MULTI_ON_ERROR
+  use ESMF
   use cime_comp_mod, only : cime_pre_init1
   use cime_comp_mod, only : cime_pre_init2
   use cime_comp_mod, only : cime_init
   use cime_comp_mod, only : cime_run
   use cime_comp_mod, only : cime_final
+  use seq_comm_mct,  only : logunit
 
   implicit none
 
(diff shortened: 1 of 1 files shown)


src/drivers/mct/main/cime_driver.F90
```diff
@@ -28,12 +28,13 @@ program cime_driver
   use ESMF,          only : ESMF_Initialize, ESMF_Finalize
   use ESMF,          only : ESMF_LogKind_Flag, ESMF_LOGKIND_NONE
   use ESMF,          only : ESMF_LOGKIND_SINGLE, ESMF_LOGKIND_MULTI
-  use ESMF,          only : ESMF_LOGKIND_MULTI_ON_ERROR
+  use ESMF
```
jedwards4b: Please don't remove the only clause.  I think you need the cpp macro here.

jedwards4b:
@ekluzek Did you test this with USE_ESMF_LIB=TRUE?
I get error #6404: This name does not have a type, and must have an explicit type.   [ESMF_VERSION_MAJOR]
and I think that the variable ESMF_VERSION_MAJOR is not defined in esmf.mod

ekluzek:
Hmmm. I did do a test with USE_ESMF_LIB=TRUE on cheyenne, although I was using whatever the default version was at the time 7.0 maybe? As coded I'm expecting ESMF_VERSION_MAJOR to be a CPP token, and it's something that's set in the ESMF build. I don't expect it in esmf.mod. The ESMF.mk file sets it. We might need to make sure that file is included in the cime build.

jedwards4b:
I added a fix for this in #2811 - I'm not sure how it could have passed tests as is.

Issue #2363, opened 2018-03-14
Possible hang on slow processors


amametjanov:
Getting these stack-traces repeatably (twice) on abort after a wall-clock timeout on Theta (KNL nodes):
```
[Task A]
Source:Line                  Call
shr_mpi_mod.F90:568          call MPI_BCAST(vec,lsize,MPI_LOGICAL,lpebcast,comm,ierr)
seq_infodata_mod.F90:2586    call shr_mpi_bcast(infodata%atm_present,        mpicom, pebcast=cplpe)
cime_comp_mod.F90:1300       if (iamin_CPLALLOCNID) call seq_infodata_exchange(infodata,CPLALLOCNID,'cpl2ocn_init')
cime_driver.F90:92           call cime_init()

[Task B]
m_Rearranger.F90:1218        call MPI_WAITANY(RecvRout%nprocs,recv_ireqs,proc,recv_istatus,ier)
cplcomp_exchange_mod.F90:273 call mct_rearr_rearrange(dom_s%data, dom_d%data, mapper%rearr, tag=ltag, VECTOR=mct_usevector, ALLTOALL=mct_usealltoall)
component_mod.F90:405        call seq_map_map_exchange(comp(1), flow='c2x', dom_flag=.true., msgtag=comp(1)%cplcompid*100+1*10+1)
cime_comp_mod.F90:1225       call component_init_cx(ocn, infodata)
cime_driver.F90:92

[Task C]
cplcomp_exchange_mod.F90:544 call mpi_allreduce(mranko,mrankog,1,MPI_INTEGER,MPI_MAX,mpicomo,ierr)
cplcomp_exchange_mod.F90:332 call seq_mctext_gsmapExtend(gsmap_old     , mpicom_old  , gsmap_old_join, mpicom_join, ID_join)
component_mod.F90:376        call seq_mctext_gsmapInit(comp(1))
cime_comp_mod.F90:1226       call component_init_cx(ice, infodata)
cime_driver.F90:92 
```

with this pe-layout on 863 nodes, 64 tasks/node:
```
> ./pelayout 
Comp  NTASKS  NTHRDS  ROOTPE
CPL :  43200/     1;      0
ATM :  43200/     1;      0
LND :   4800/     1;  38400
ICE :  24000/     1;      0
OCN :  12000/     1;  43200
ROF :   4800/     1;  33600
GLC :      1/     1;      0
WAV :      1/     1;      0
ESP :      1/     1;      0
```

[...]

Issue #2295, opened 2018-02-14
multi instance problem on titan when  ./xmlchange MULTI_DRIVER=TRUE


kevans32:
I am trying out this multi instance capability in titan. Very cool!
I can run when ./xmlchange MULTI_DRIVER=FALSE but when 
 ./xmlchange MULTI_DRIVER=TRUE it seems to not know how to use the second set of tasks 

run command : aprun -S 8 -n 1280 -N 16 -d 1 /ccs/home/4ue/acme_scratch/cli115/FC5multi2/bld/acme.exe    >> acme.log.$LID 2>&1

But in output it does this:
(seq_comm_setcomm)  initialize ID (  1 GLOBAL          ) pelist   =     0   639     1 ( npes =   640) ( nthreads =  1)( suffix =)
Rank 526 [Wed Feb 14 12:04:40 2018] [c4-1c2s5n2] Fatal error in PMPI_Group_range_incl: Invalid argument, error stack:
PMPI_Group_range_incl(195)........: MPI_Group_range_incl(group=0x88000000, n=1, ranges=0x5b495c0, new_group=0x7fffffff3d44) failed
MPIR_Group_check_valid_ranges(323): **The 0th element of a range array ends at 1279 but must be nonnegative and less than 640**

**Here are my bash commands for a simple test run with 2 instances**
casename=FC5multi2
casedirectory=$PROJWORK/cli115/4ue                                             
machine=titan
projectname=cli115
compset=FC5
grid=ne30_ne30

module load python/2.7.9
code_dir=$HOME
code_tag=ACME
cd ${code_dir}/${code_tag}/cime/scripts

$code_dir/$code_tag/cime/scripts/create_newcase -case $casedirectory/$casename -mach $machine -project $projectname -compset $compset -res $grid

cd $casedirectory/$casename

  ./xmlchange NINST_ATM=2
  ./xmlchange NINST_LND=2
  ./xmlchange NINST_ICE=2
  ./xmlchange NINST_ROF=2
  ./xmlchange NINST_GLC=2
  ./xmlchange NINST_OCN=2
  ./xmlchange NINST_WAV=2
  ./xmlchange NINST_ESP=2

  ./xmlchange MULTI_DRIVER=TRUE (or FALSE)

  ./case.setup

cat <<EOF >> user_nl_cam_0001
 zmconv_c0_lnd = 0.007
EOF
cat <<EOF >> user_nl_cam_0002
 zmconv_c0_lnd = 0.008
EOF

[...]

PR #2804, opened 2018-09-13
Don't exceed maximum MPI tag value

jhkennedy:
The MPI standard specifies that the MPI tag upper bound (MPI_TAG_UB) must
be at least 32767 and many implementations greatly exceed this number
(eg. cray-mpich/7.4.4 has MPI_TAG_UB == 2^21-1). However, in
multi-instance runs, this tag is quickly exceed (~30 instances).

This PR adds a check for tags larger than MPI_TAG_UB and reduces
msgtag so that it will take ~300 instances to exceed MPI_TAG_UB on cray
machines while still protecting against tag collisions out to ~100
instances (see ESMCI/cime#2397 and ESMCI/cime#2363).

This should be sufficient for any multi-instance runs not using
multi-driver (where this isn't a problem), and runs larger should be
using multi-driver.

Test suite: scripts_regression_tests.py on edison; MVK SystemTest on Titan (E3SM)
Test baseline: N/A
Test namelist changes: none
Test status: [bit for bit]

Fixes E3SM-Project/E3SM#2487

User interface changes?: N

Code review:

### DIFF
--- a/src/drivers/mct/main/component_mod.F90
+++ b/src/drivers/mct/main/component_mod.F90
@@ -332,6 +332,7 @@ contains
     type(mct_gGrid) :: dom_tmp   ! temporary
     character(*), parameter :: subname = '(component_init_cx)'
     character(*), parameter :: F0I = "('"//subname//" : ', A, 2i8 )"
+    character(*), parameter :: F1I = "('"//subname//" : ', A, I10, A, I10 )"
     !---------------------------------------------------------------
 
     ! Initialize driver rearrangers and AVs on driver
@@ -390,7 +391,7 @@ contains
 
                 mpi_tag = comp(1)%cplcompid*1000+1*10+1
                 if (mpi_tag > max_mpi_tag) then
-                   write(logunit, F0I) 'MPI message tag ',mpi_tag,' exceeds max ',max_mpi_tag
+                   write(logunit, F1I) 'MPI message tag ',mpi_tag,' exceeds max ',max_mpi_tag
                    call shr_sys_abort(subname//' ERROR: Max mpi tag exceeded')
                 end if
                 call seq_map_map_exchange(comp(1), flow='c2x', dom_flag=.true., msgtag=mpi_tag)
@@ -569,7 +570,7 @@ contains
     integer :: eci, num_inst
     integer :: mpi_tag
     character(*), parameter :: subname = '(component_init_areacor)'
-    character(*), parameter :: F0I = "('"//subname//" : ', A, 2i8 )"
+    character(*), parameter :: F1I = "('"//subname//" : ', A, I10, A, I10 )"
     !---------------------------------------------------------------
 
     num_inst = size(comp)
@@ -581,7 +582,7 @@ contains
           ! Map component domain from coupler to component processes
           mpi_tag = comp(eci)%cplcompid*1000+eci*10+5
           if (mpi_tag > max_mpi_tag) then
-             write(logunit, F0I) 'MPI message tag ',mpi_tag,' exceeds max ',max_mpi_tag
+             write(logunit, F1I) 'MPI message tag ',mpi_tag,' exceeds max ',max_mpi_tag
              call shr_sys_abort(subname//' ERROR: Max mpi tag exceeded')
           end if
           call seq_map_map(comp(eci)%mapper_Cx2c, comp(eci)%dom_cx%data, comp(eci)%dom_cc%data, msgtag=mpi_tag)
@@ -604,7 +605,7 @@ contains
           ! Map corrected initial component AVs from component to coupler pes
           mpi_tag = comp(eci)%cplcompid*1000+eci*10+7
           if (mpi_tag > max_mpi_tag) then
-             write(logunit, F0I) 'MPI message tag ',mpi_tag,' exceeds max ',max_mpi_tag
+             write(logunit, F1I) 'MPI message tag ',mpi_tag,' exceeds max ',max_mpi_tag
              call shr_sys_abort(subname//' ERROR: Max mpi tag exceeded')
           end if
           call seq_map_map(comp(eci)%mapper_cc2x, comp(eci)%c2x_cc, comp(eci)%c2x_cx, msgtag=mpi_tag)
@@ -843,7 +844,7 @@ contains
     integer :: ierr
     integer :: mpi_tag
     character(*), parameter :: subname = '(component_exch)'
-    character(*), parameter :: F0I = "('"//subname//" : ', A, 2i8 )" 
+    character(*), parameter :: F1I = "('"//subname//" : ', A, I10, A, I10 )"
     !---------------------------------------------------------------
 
     if (present(timer_barrier))  then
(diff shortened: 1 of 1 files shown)


src/drivers/mct/main/component_mod.F90
```diff
@@ -63,6 +63,8 @@ module component_mod
   integer  :: mpicom_GLOID, mpicom_CPLID           ! GLOID, CPLID mpi communicator
   integer  :: nthreads_GLOID, nthreads_CPLID
   logical  :: drv_threading
+  integer (kind=MPI_ADDRESS_KIND) :: max_mpi_tag
```
jedwards4b: Are you sure that this is the correct type for this variable - from the mpich documentation:
The attribute_val in Fortran is a pointer to a Fortran integer, not a pointer to a void *.
Which I think implies that what is in max_mpi_tag is the address of the value and not the value itself?

jhkennedy: No I'm not a 100% sure. From my understanding though, this is returning the actual value not an address of the value. 

[This mpi-forum page](https://www.mpi-forum.org/docs/mpi-2.2/mpi22-report/node369.htm) explains it and provides some examples, but here is the relevant text: 

>MPI supports two types of attributes: address-valued (pointer) attributes, and integer valued attributes. C and C++ attribute functions put and get address valued attributes. Fortran attribute functions put and get integer valued attributes. When an integer valued attribute is accessed from C or C++, then MPI_xxx_get_attr will return the address of (a pointer to) the integer valued attribute, which is a pointer to MPI_Aint if the attribute was stored with Fortran MPI_xxx_SET_ATTR, and a pointer to int if it was stored with the deprecated Fortran MPI_ATTR_PUT. When an address valued attribute is accessed from Fortran, then MPI_xxx_GET_ATTR will convert the address into an integer and return the result of this conversion. This conversion is lossless if new style attribute functions are used, and an integer of kind MPI_ADDRESS_KIND is returned. The conversion may cause truncation if deprecated attribute functions are used. In C, the deprecated routines MPI_Attr_put and MPI_Attr_get behave identical to MPI_Comm_set_attr and MPI_Comm_get_attr.

And that same behavior is implied here as well:

https://www.open-mpi.org/doc/v3.0/man3/MPI_Comm_set_attr.3.php

jhkennedy:
@amametjanov I think you're right, it's probably best to keep the message tags unchanged (Note: this will limit multi-instance to ~20 instances without using multi-driver). I'm in the process of converting some of my multi-instance runs to use multi-driver. Once those pass I'll bump it back up and update the multi-instance docs to discuss this limitation.

src/drivers/mct/main/component_mod.F90
jedwards4b: thanks

jedwards4b:
@jhkennedy is this ready to merge?

jhkennedy:
@jedwards4b I think we need to rethink these tags first.

To recap,  ESMCI/cime#2397 and ESMCI/cime#2363 caused the MPI tag values in the mct driver to change from: 

```fortran
 mpi_tag = comp(eci)%cplcompid*100+eci*10+N
```
to 
```fortran
 mpi_tag = comp(eci)%cplcompid*10000+eci*10+N
```

where `N` is in the range of `[1,9]`, because there were some message tag collisions causing hangups on single instance high rez runs. 

This change causes multi-instance runs to exceed the maximum MPI tag value at ~20 instances on titan (see E3SM-Project/E3SM#2487) and at ~100 instances on edison (see ESCOMP/ctsm#515 ; pinging @ckoven). 

Multi-driver, should alleviate this issue, but I've run into ESMCI/cime#2295 when trying to change my multi-instance runs over to multi-driver (seems to be titan specific? trying at NERSC), and multi-driver isn't useful for  ESCOMP/ctsm#515 because of how long it takes to build the DATM namelists in their case. 

So the proposed solution was to leave a (MPI implementation dependent) limit on multi-instance runs and encourage multi-driver use, while also picking a middle ground in the MPI tag size (`mpi_tag = comp(eci)%cplcompid*1000+eci*10+5`; 1000 instead of 100 (old) or 10000 (new)).

But, due to the aforementioned multi-driver issues and @amametjanov feeling that 1000 still leaves too much potential for tag collision, I don't think this solution is the right one. 

So, that leaves us with **is there a different and better way to assign the tags?**

For example, it seems the `comp(eci)%cplcompid` part of the tag varies the greatest so could we reorder the tag to something like: 

```fortran
 mpi_tag = eci*10000+comp(eci)%cplcompid*10+5
```
?

However, I'm not sure I understand how the collisions are occurring in single instance runs -- @amametjanov what are these tags colliding with?

amametjanov:
@jhkennedy how about this patch to use old tags for multi-instance runs: https://gist.github.com/amametjanov/790661841af3a49977b1b584fd509668

ckoven:
Hi All,
Just wanted to add that the --multi-dirver method doesn't seem to be working for my use-case described in https://github.com/ESCOMP/ctsm/issues/515, I get DATM errors as described in https://github.com/ESCOMP/ctsm/issues/515#issuecomment-425493507.  So I guess I would just like to advocate maintaining/strengthening the ability to do single-instance DATM, multi-instance land simulations with large numbers of land instances if at all possible.  Thanks!

rljacob:
What is the status of this PR?

jhkennedy:
@rljacob I'm testing @amametjanov patch -- so far looks good; I'll know more tomorrow.

jhkennedy:
@jedwards4b this is ready to merge. 

@ckoven this should allow you to run your multi-instance runs.

src/drivers/mct/main/component_mod.F90
```diff
@@ -102,6 +106,9 @@ subroutine component_init_pre(comp, compid, cplcompid, cplallcompid, &
     call seq_comm_getinfo(CPLID, mpicom=mpicom_CPLID, iamroot=iamroot_CPLID, nthreads=nthreads_CPLID)
     iamin_CPLID = seq_comm_iamin(CPLID)
 
+    ! MPI spec. garuntees at least 32768 tags available, but most implementations will have more.
```
jedwards4b: spelling: guarantees

src/drivers/mct/main/component_mod.F90
```diff
@@ -377,9 +386,20 @@ subroutine component_init_cx(comp, infodata)
                 if (iamroot_CPLID) then
                    write(logunit,F0I) 'creating dom_cx'
                    call shr_sys_flush(logunit)
-                end if
+                end igit cof
```
jedwards4b: this should be end if

src/drivers/mct/main/component_mod.F90
```diff
@@ -377,9 +386,20 @@ subroutine component_init_cx(comp, infodata)
...
                 call seq_mctext_gGridInit(comp(1))
-                call seq_map_map_exchange(comp(1), flow='c2x', dom_flag=.true., msgtag=comp(1)%cplcompid*10000+1*10+1)
+
+                if (size(comp) > 1) then
+                    mpi_tag = comp(eci)%cplcompid*100+eci*10+1
```
jedwards4b: Why two cases here?  Why not just use line 393 regardless of size(comp)?

src/drivers/mct/main/component_mod.F90
jhkennedy: Woops, fixed.

src/drivers/mct/main/component_mod.F90
jhkennedy: To summarize the thread, that was the suggestion by @amametjanov to multi-instance runs with >20 instance to run without exceeding the MPI tag limit, but still prevent tag collision on single instance high res runs. Multi-driver is a possible alternative, but there were challenges around setting that up for people in this thread and others I've talked to about this issue. It seems if multi-instance is going to be an option it should be able to handle at least a normal ensemble size. 

I'd be welcome to a suggestion on how to assign these tag values entirely differently, but I don't know them, or with what they are colliding, well enough. I suggested re-ordering the way the tags were written, but that didn't get any traction from anyone in this thread. So this at least fixes the immediate issue.

PR #2823, opened 2018-10-03
fix inputdata issue for multi_instance cases

### LINKED ISSUE #2822: Automated input data download does not work for multi-instance runs
I recently found out that scripts to download missing input data do not work for multi-instance runs. In my multi-instance run, I had a different `ncdata` file for each instance. Launching `check_input_data` downloads only the `ncdata` file for the last instance. I looked into the script and it looks like `check_input_data` script download files mentioned in  `*.input_data_list` (e.g. `cam.input_data_list`) and `*.input_data_list`  file captures only the file from the last instance. 

I am not sure what would be the best way to solve this issue. One solution that I can think of is to add a "ninst" (total number of instances) loop which creates `*.input_data_list` files for each instance and then `check_input_data` scripts go through all these files to download missing data.

jedwards4b:
For multi instance cases append the input_data_list from each instance so that all required files will be downloaded.  This is a backward compatible change but requires a change in each external component to only delete the input_data_list file prior to the first instance and append with each instance.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2822

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/Tools/archive_metadata
+++ b/scripts/Tools/archive_metadata
@@ -79,6 +79,7 @@ class PasswordPromptAction(argparse.Action):
 # -------------------------------------------------------------------------------
     """ SVN developer's password class handler
     """
+    # pylint: disable=redefined-builtin
     def __init__(self,
                  option_strings=None,
                  dest=None,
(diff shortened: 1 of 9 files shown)


jgfouca:
Thanks, @jedwards4b

singhbalwinder:
Thanks @jedwards4b !

jgfouca:
@singhbalwinder , I assume you approve?

PR #2825, opened 2018-10-05
Improve continue run checks in case_submit

### LINKED ISSUE #213: Abort if user tries to do CONTINUE_RUN in the first iteration of a case
This isn't critical, but it has bitten two people in the last month, so at today's CSEG meeting we decided we should open an issue on it: Two people recently have tried to set up branch runs with the same case as the run they're branching from, and then set CONTINUE_RUN to TRUE in the first iteration of the branch run (thinking that that was what they were supposed to do, or at least that it was allowed). The scripts allowed this to happen, and then the model aborted in CISM, whose time manager gets confused by this case. (Apparently other components handle this more correctly.)

The conclusion from the CSEG meeting is that this case shouldn't be allowed: you should never set CONTINUE_RUN to TRUE in the first iteration of a run, whether it's a branch run or something else. Ideally, the scripts would catch this.

A simple solution would be to have a counter of the number of times a case has completed, and fail with an error message if you try to set CONTINUE_RUN before the case has successfully completed. Others pointed out that there are a number of cases where this wouldn't work, but in most / all of these cases, the problem would just be that the error message wouldn't be triggered (i.e., you wouldn't typically get false error messages).

jgfouca:
If there are no netcdf files, then the case was not run before
and CONTINUE_RUN=TRUE doesn't make sense.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #213 

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/case/case_submit.py
+++ b/scripts/lib/CIME/case/case_submit.py
@@ -6,12 +6,13 @@ if there is no queueing system.  A cesm workflow may include multiple
 jobs.
 submit, check_case and check_da_settings are members of class Case in file case.py
 """
-import socket
-from six.moves                   import configparser
+from six.moves                      import configparser
 from CIME.XML.standard_module_setup import *
 from CIME.utils                     import expect, run_and_log_case_status, verbatim_success_msg
 from CIME.locked_files              import unlock_file, lock_file
 from CIME.test_status               import *
+
+import socket, glob
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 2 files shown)


billsacks:
I'm okay with this error check, but I don't think it robustly addresses #213 . As I note in https://github.com/ESMCI/cime/issues/213#issuecomment-296291393 , in the two cases that inspired us to open this issue, users copied the contents of the restart directories into their run directories, so I don't think your new error check would have been triggered.

What do you think about putting in place a counter of the number of times the case has run, as was originally suggested when we opened this issue?

jgfouca:
@billsacks , that seems too complicated, but I'll go with the other solution I was thinking of.

billsacks:
Okay. You have me on the edge of my seat, wondering what that solution will be :-)

jgfouca:
@billsacks , behold my genius solution!

PR #2829, opened 2018-10-05
xmlchange: Give user guidance on what to do next

### LINKED ISSUE #815: Feature request: xmlchange tells you what steps need to be rerun for the change to take effect
This is a feature request that popped to mind, which is for the sake of relatively inexperienced users: It would be nice if xmlchange noticed when you change a setting that requires you to re-do some step, and gave you the commands you need to run for the change to take effect.

For example, if you change a setting in env_build.xml after you have already done the build, then it will tell you something like:

```
For this change to take effect, run:
./case.build --clean-all
./case.build
```

(I'm actually not sure if those are the commands you need to run when changing env_build.xml... maybe this feature request is as much for me as for "inexperienced users", or maybe I myself fall into the "inexperienced user" category in this respect....)

I'm imagining some pretty simple rules like: If a setting in env_build is changed after the build is complete, then you need to redo the build; if a setting in env_mach_pes is changed after case.setup is complete, then you need to redo case.setup. That may be all that's really needed.

What made me think of this was thinking of how things have changed since CESM1: In CESM1, you needed to explicitly specify the file to change via xmlchange. Now that is determined automatically. That automatic determination is very nice, but the downside is that it lets you change files without realizing that (for example) you're changing a build-time setting that requires a rebuild.

jgfouca:
Test suite: by-hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #815 

User interface changes?: N

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/Tools/xmlchange
+++ b/scripts/Tools/xmlchange
@@ -143,8 +143,19 @@ def xmlchange_single_value(case, xmlid, xmlval, subgroup, append, force, dryrun)
         xmlval = convert_to_type(xmlval, type_str, xmlid)
 
     if not dryrun:
-        newval = case.set_value(xmlid, xmlval, subgroup, ignore_type=force)
-        expect(newval is not None,"No variable \"%s\" found"%xmlid)
+        result = case.set_value(xmlid, xmlval, subgroup, ignore_type=force, return_file=True)
+        expect(result is not None,"No variable \"%s\" found"%xmlid)
+        filename = result[1]
+
+        if filename.endswith("env_batch.xml"):
+            logger.info("case.submit will auto-regen your batch files, no further user action is needed")
+        elif filename.endswith("env_build.xml"):
+            logger.info("You should consider cleaning and rebuilding your case")
+        elif filename.endswith("env_mach_pes.xml"):
+            logger.info("Changing mach_pes can have a big impact on your case. Please re-run case.setup, and clean/rebuild your case")
+        elif filename.endswith("env_case.xml"):
+            logger.info("You are changing a fundamental property of your case. Please re-run case.setup, and clean/rebuild your case")
+
     else:
         logger.warning("'%s' = '%s'" , xmlid , xmlval )
 
--- a/scripts/Tools/xmlchange
+++ b/scripts/Tools/xmlchange
@@ -48,7 +48,7 @@ Examples:
 
 from standard_script_setup import *
 
-from CIME.utils import expect, convert_to_type, append_case_status
+from CIME.utils import expect, convert_to_type, append_case_status, get_batch_script_for_job
 from CIME.case import Case
 
 import re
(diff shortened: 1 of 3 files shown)


jgfouca:
@billsacks , approve?

scripts/lib/CIME/case/case.py
```diff
@@ -368,11 +368,11 @@ def get_resolved_value(self, item, recurse=0, allow_unresolved_envvars=False):
 
         return item
 
-    def set_value(self, item, value, subgroup=None, ignore_type=False, allow_undefined=False):
+    def set_value(self, item, value, subgroup=None, ignore_type=False, allow_undefined=False, return_file=False):
         """
         If a file has been defined, and the variable is in the file,
-        then that value will be set in the file object and the file
-        name is returned
+        then that value will be set in the file object and the resovled value
```
billsacks: Please document the return value(s) - i.e., if return_file is False, then just the resolved value is returned; if True, then a tuple (value, filename) is returned. If the variable isn't found, then returns None.

scripts/Tools/xmlchange
```diff
@@ -143,8 +143,19 @@ def xmlchange_single_value(case, xmlid, xmlval, subgroup, append, force, dryrun)
...
+        expect(result is not None,"No variable \"%s\" found"%xmlid)
+        filename = result[1]
+
+        if filename.endswith("env_batch.xml"):
+            logger.info("case.submit will auto-regen your batch files, no further user action is needed")
+        elif filename.endswith("env_build.xml"):
+            logger.info("You should consider cleaning and rebuilding your case")
+        elif filename.endswith("env_mach_pes.xml"):
+            logger.info("Changing mach_pes can have a big impact on your case. Please re-run case.setup, and clean/rebuild your case")
+        elif filename.endswith("env_case.xml"):
```
billsacks: I thought you're not allowed to change something in env_case, ever. Is it really safe to change these values if you rerun case.setup and rebuild?

jgfouca: I was wondering that myself. @jedwards4b ?

jgfouca:
@billsacks , your "out of order" is a great idea. Let me work in this PR a bit more.

scripts/Tools/xmlchange
jedwards4b: Right - you should never change anything in env_case.xml -  I think there is a way to mark that file as read only.    There are a few special circumstances where you may want to change something in this directory but I'm not sure it'll work - say for example you need to move a case directory.

jgfouca:
@billsacks , see latest changes.

rljacob:
I like the change to only print something if its out of order.

jedwards4b:
```
./xmlchange ROOTPE_CPL=0
For your changes to take effect, run:
./case.setup --reset
./case.build --clean-all
./case.build
```
Is overkill - the --clean-all is not needed and makes the rebuild quite slow

billsacks:
@jedwards4b so what do you suggest instead?

jedwards4b:
```
For your changes to take effect, run:
./case.setup --reset
./case.build
```

jedwards4b:
The exception is if there is a change from no threading to threading

PR #2832, opened 2018-10-06
hr2x at endofrun

### LINKED ISSUE #2831: Coupler history file hr2x problems for forecasts < 24 hours.
The cpl.hr2x files have averaged fields in them, despite the names 
and (previous) documentation (CIME issue #2773, ESCOMP/cesm#71).
They are written out only daily, but that leads to incorrect results 
when the forecast is less than a day long, as during data assimilation 
runs with, for example, 6 hour forecasts.  Then the files are written
only at hour 24, not hours 6, 12, 18.
And the files contain data averaged over hours 18-24, not over the 
full day implied by the file names (and frequency): 
case.cpl.hr2x.yyyy-mm-dd-00000.nc
TOD was added to the filenames in an earlier commit
(CIME #2754).

kdraeder:
Write cpl.hr2x files at ENDOFRUN when forecast is < 24 hours. 
Write them every 24 hours for longer forecasts.

Test suite: scripts_regression_tests.py and cesm2_2 CAM forecasts from hours 12-18 and 18-24
Test baseline: 
Test namelist changes: none
Test status: bit for bit in the model run.  Answer changing in the cpl.hr2x files

Fixes #2831 

User interface changes?: None

Update gh-pages html (Y/N)?: possibly; to alert users to the actual contents of cpl.hr2x files

Code review:

### DIFF
--- a/src/drivers/mct/main/cime_comp_mod.F90
+++ b/src/drivers/mct/main/cime_comp_mod.F90
@@ -269,6 +269,7 @@ module cime_comp_mod
   logical  :: t24hr_alarm            ! alarm every twentyfour hours
   logical  :: t1yr_alarm             ! alarm every year, at start of year
   logical  :: pause_alarm            ! pause alarm
+  logical  :: write_hist_alarm       ! seq_timemgr index for driver
   integer  :: drv_index              ! seq_timemgr index for driver
 
   real(r8) :: days_per_year = 365.0  ! days per year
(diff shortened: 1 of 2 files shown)


src/drivers/mct/main/cime_comp_mod.F90
```diff
@@ -269,6 +269,7 @@ module cime_comp_mod
   logical  :: t24hr_alarm            ! alarm every twentyfour hours
   logical  :: t1yr_alarm             ! alarm every year, at start of year
   logical  :: pause_alarm            ! pause alarm
+  logical  :: write_hist_alarm       ! seq_timemgr index for driver
```
billsacks: Please fix the comment on this variable

kdraeder:
@billsacks 
Thanks for catching the mistaken description in the code,
and for suggesting the documentation.  The variable name
and description are fairly general, in case it can be used
for a similar purpose in a different section of the code.

kdraeder:
@bertinia @billsacks 
Alice, I named you as a reviewer since it has to do with history file naming
and timing of writes. If you'd rather not review it, that's fine, but it's a simple fix
which might take you less than 5 minutes.
Thanks,
Kevin

PR #2833, opened 2018-10-08
Propagate logging settings to submitted jobs

### LINKED ISSUE #1130: allow --debug option to propagate to queue dependent jobs
If you submit a case using "case.submit --debug" and DOUT_S=TRUE, it would be
helpful to propagate the debug argument to the case.st_archive call. Likewise,
if DOUT_L_MS=TRUE then the --debug argument should be applied to the case.lt_archive arg list.

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #1130

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -4,7 +4,7 @@ Interface to the env_batch.xml file.  This class inherits from EnvBase
 
 from CIME.XML.standard_module_setup import *
 from CIME.XML.env_base import EnvBase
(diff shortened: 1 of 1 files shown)


jgfouca:
Self-assigned since tests area still running.

jgfouca:
I tested by-hand with case.submit --debug, where st archiving was enabled. Both the case.test and case.st_archive jobs had the --debug flag.

PR #2837, opened 2018-10-09
change default for flux_convergence for all but 1 degree case

jedwards4b:
Changes the atm-ocn flux convergence criteria for all except the 1 degree case,  this changes increases performance considerably for f19_g17

Test suite:  PFS.f19_g17.B1850.cheyenne_intel.allactive-default I also made sure that
                    PFS.f09_g17.B1850.cheyenne_intel.allactive-default was not affected.  
                  This change is for cesm only.
Test baseline: 
Test namelist changes: 
Test status:  roundoff

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


billsacks:
What is the rationale for applying it to all resolutions except 0.9x1.25? Is that simply because we don't want to change answers for the 1 deg case at this point? It would be great if this rationale could be documented in the xml file and/or in this PR and/or via a new issue saying what we should do long-term.

jedwards4b:
@billsacks I've added a comment.

billsacks:
Thanks!

PR #2845, opened 2018-10-10
Improve idmap test

### LINKED ISSUE #1908: more robust method of determining when rof2ocn_fmapname (and maybe others) can be set to idmap
Currently, the setting of `rof2ocn_fmapname` needs to be set to a non "idmap" value if the "flood_present" flag is on. The flood_present flag is actually set by the ocean component at run time - so this is a bit of a chicken and egg problem. The flood_present flag is used in `prep_ocn_mod.F90` and then expects a valid value of `rof2ocn_fmapname`. For a data ocean model - flood_present can always be assumed to be false and so `rof2ocn_fmapname` can be "idmap". However, this is not necessarily the case for a prognostic ocean. A more robust method needs to be determined in order to determine if setting `rof2ocn_fmapname` = "idmap" is problematic.

@billsacks provided the following: Ideally, we'd have some data-driven list of mapping pairs that are optional (i.e., needed in some configuration but not in others).

jedwards4b:
Add a default value of idmap_ignore for certain mapping files - it can only be determined at runtime whether these files should or should not exists.   This PR doesn't really change anything except to move the list of files from hardcoded in the python to the xml file.    

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #1908 arguably 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/nmlgen.py
+++ b/scripts/lib/CIME/nmlgen.py
@@ -572,7 +572,7 @@ class NamelistGenerator(object):
                         continue
                     file_path = character_literal_to_string(literal)
                     # NOTE - these are hard-coded here and a better way is to make these extensible
-                    if file_path == 'UNSET' or file_path == 'idmap':
+                    if file_path == 'UNSET' or file_path == 'idmap' or file_path == 'idmap_ignore':
                         continue
                     if file_path == 'null':
                         continue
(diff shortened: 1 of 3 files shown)


rljacob:
Clarify from telecon:  "idmap" is used when two models are on the same grid OR when two components don't need a map (like runoff to data ocean).

Issue #2686, opened 2018-06-22
Improve config_machines.xsd


jedwards4b:
The current schema allows the < mpirun > and < module_system > sections to be missing, but they are both required for cime usage.

rljacob:
Telecon:  create_newcase --machine name  should give an error if the machine entry is missing those sections.

jhkennedy:
@rljacob do you want an error only if the `--machine` option is given, or anytime a machine is instantiated that doesn't have those values?

jedwards4b:
@jhkennedy It should be anytime those values are not provided.

PR #2854, opened 2018-10-15
Require < mpirun > and < module_system > in machine config

jhkennedy:
`< mpirun >` and `< module_system >` were required for CIME usage, but not required in the `config_machines.xsd` XML schema.

This changes the schema to require those fields and adds an expectation in CIME `Case.configure` for those elements. 

Test suite: ./scripts_regression_tests.py
Test baseline: N/A
Test namelist changes: None
Test status: B4B

Fixes ESMCI/cime#2686

User interface changes?: None

Update gh-pages html (Y/N)?: N -- already describes these as required

Code review:


scripts/lib/CIME/case/case.py
```diff
@@ -826,6 +826,11 @@ def configure(self, compset_name, grid_name, machine_name=None,
             logger.info("Machine is {}".format(machine_name))
 
         nodenames = machobj.get_node_names()
+        expect('mpirun' in nodenames,
```
jedwards4b: We really don't need this since the xmllint will pick up the error in the file before we arrive at this point.  Have you tried intentionally leaving out the mpirun or module_system statements as part of your testing?

jedwards4b:
I see the issue now, the e3sm/config_files.xml does not specify a schema for config_machines.xml
cesm does - 
https://github.com/ESMCI/cime/blob/2a2c3f00130dac4c53c0795c2c08fb19f0d53aaf/config/cesm/config_files.xml#L42

I would recommend adding this schema entry in e3sm but it may require cleaning up your config_machines.xml since it isn't regularly tested.  Meanwhile I will test this PR with CIME_MODEL=cesm.

jhkennedy:
Ah! Okay, will do.

jhkennedy:
@jgfouca , in order to fix ESMCI/cime#2686, I've turned on `xmllint`ing of `config/e3sm/machines/config_machines.xml`. This mostly required me to re-order the `config_machines.xml` file, but I also needed to add these elements to the schema: https://github.com/ESMCI/cime/blob/27f68b66d5cd8235aefc011aab944368988820e4/config/xml_schemas/config_machines.xsd#L58-L68

Which of those are actually used/required by E3SM, and which should we eliminate from `config_machines.xml` specs? 

---

Also, the generic mac and linux machine config, as well as  a few machines that seem to be personal machines (i.e., paths defined for a specific user), would likely be better defined in the CIME user config directory (`$HOME/.cime/`). Do you think we should take any of these out of `config_machines.xml`?

https://github.com/ESMCI/cime/blob/27f68b66d5cd8235aefc011aab944368988820e4/config/e3sm/machines/config_machines.xml#L597-L625

https://github.com/ESMCI/cime/blob/27f68b66d5cd8235aefc011aab944368988820e4/config/e3sm/machines/config_machines.xml#L627-L657

https://github.com/ESMCI/cime/blob/27f68b66d5cd8235aefc011aab944368988820e4/config/e3sm/machines/config_machines.xml#L2474-L2515

https://github.com/ESMCI/cime/blob/27f68b66d5cd8235aefc011aab944368988820e4/config/e3sm/machines/config_machines.xml#L2517-L2557

https://github.com/ESMCI/cime/blob/27f68b66d5cd8235aefc011aab944368988820e4/config/e3sm/machines/config_machines.xml#L2687-L2717

jedwards4b:
Please verify that the changes in cime/config/xml_schemas/config_machines.xsd are still compatible with cime/config/cesm/machines/config_machines.xml

jgfouca:
@jhkennedy , I would hold off removing old machines for now.

jhkennedy:
@jedwards4b both `cime/config/cesm/machines/config_machines.xml` and `cime/config/e3sm/machines/config_machines.xml` validate with `xmllint`

jhkennedy:
> RUNDIR, EXEROOT -> These are core CIME values, I'm very surprised these weren't already in the schema

These aren't defined anywhere in the CESM `config_machines.xml` file (or anywhere under `config/cesm` for that matter), but they are ref. like they are environment variables. So I suspect CESM setting those as environment variables outside of CIME somehow. 

> CESMSCRATCHROOT -> Definitely obsolete and can be deleted

Done!

> DOUT_L_HTAR, DOUT_L_MSROOT -> Long-term archiving is no-longer supported by CIME, these can be removed

Done and done!

> PIO_CONFIG_OPTS, PIO_BUFFER_SIZE_LIMIT -> These should be migrated to config/e3sm/config_pio.xml

I moved them to `config_pio.xml`, but is there anything that needs to be done with `PIO_BUILD_TIMING` in CIME for those options to be recognized from the `config_pio.xml` file instead of `config_machines.xml`? I had to create a new entry for `PIO_BUILD_TIMING` in `config_pio.xml` instead of just uncommenting an existing section. 


> TEST_TPUT_TOLERANCE, MAX_GB_OLD_TEST_DATA -> These need to be added to the XSD

Done and done! 

BATCHQUERY, BATCHSUBMIT -> These can be deleted.

Done and done!

--- 

@jedwards4b the regression test script passed on titan, but you might want to run it yourself for `CIME_MODEL=cesm` since I messed with the schema. `xmllint` passes for both CESM and E3SM still.

jgfouca:
> I moved them to `config_pio.xml`, but is there anything that needs to be done with `PIO_BUILD_TIMING` in CIME for those options to be recognized from the `config_pio.xml` file instead of `config_machines.xml`? I had to create a new entry for `PIO_BUILD_TIMING` in `config_pio.xml` instead of just uncommenting an existing section.

It looks to me like PIO_BUILD_TIMING is a cmake value for PIO, not a CIME value.

jhkennedy:
> It looks to me like PIO_BUILD_TIMING is a cmake value for PIO, not a CIME case value.

@jgfouca ah, woops; should I move it back? Or should it go some place else entirely?

jgfouca:
@jhkennedy , just delete it. It doesn't need to be handled by CIME at all.

jhkennedy:
@jedwards4b this should be good to go now.

jedwards4b:
@jhkennedy Conflicts in config_machines.xml

PR #2858, opened 2018-10-16
Jgfouca/branch for acme split 2018 10 16

jgfouca:
Summary

- Add two new tests (TSC and PGN) for testing non-BFB changes to the model. These tests can be launched using create_test script. PGN has the following python dependencies:
numpy
scipy
netCDF4
math
sklearn.metrics

These python libraries should be present on the system and be available on compute nodes as well. TSC test currently does not have any post-processing built in.

- PIO FindNetcdf.cmake:  Currently, FindNetCDF.cmake only uses ${NetCDF_${NCDFcomp}PATHS} if ${MPI${NCDFcomp}_FOUND} is true. There's no need for that. We should always allow the user to pass hints to find_package for where to find the libraries (and some users may have those libraries installed in different folders).

- Fix test status:  Fix test_status overall status for certain state layout.  The internal test status log showed this layout caused an incorrect FAIL to be reported to the dashboard. Fix that bug and add a unit test for this layout.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:  @jedwards4b


jedwards4b:
There are some pylint issues to be fixed before merge.

PR #2868, opened 2018-10-22
Remove all references to BATCHQUERY and BATCHSUBMIT

jgfouca:
Test suite: ./scripts_regression_tests --fast
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/case/case_run.py
+++ b/scripts/lib/CIME/case/case_run.py
@@ -22,7 +22,6 @@ def _pre_run_check(case, lid, skip_pnl=False, da_cycle=0):
 
     caseroot = case.get_value("CASEROOT")
     din_loc_root = case.get_value("DIN_LOC_ROOT")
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/case/case_run.py
```diff
@@ -42,17 +41,6 @@ def _pre_run_check(case, lid, skip_pnl=False, da_cycle=0):
     # load the module environment...
     case.load_env(reset=True)
 
-    # set environment variables
-
-    if batchsubmit is None or len(batchsubmit) == 0:
```
jgfouca: Any idea what these environment variables were for?

jedwards4b: not at the moment

Issue #2871, opened 2018-10-23
strm_datdir not set for datm mode CLM_QIAN_WISO


billsacks:
@mathewvrothstein ran into an issue trying to run datm with mode CLM_QIAN_WISO: the file path for stream CLM_QIAN_WISO.TPQW remained set to null. I have tracked this down to a typo, which I'll fix in an upcoming PR.

PR #2872, opened 2018-10-23
Fix typo in setting strm_datdir for CLM_QIAN_WISO.TPQW

billsacks:
Test suite: Just ran preview_namelists on a case with the new setting,
with DATM_MODE=CLM_QIAN_WISO
Test baseline: n/a
Test namelist changes: Fixes stream settings for cases with DATM_MODE=CLM_QIAN_WISO
Test status: bit for bit

Fixes ESMCI/cime#2871

User interface changes?: none

Code review:


billsacks:
Note that the forcing files for this datm_mode are not checked in to the inputdata repository. @mathewvrothstein is going to commit them after doing some more testing.

PR #2877, opened 2018-10-25
Sarich/update e3sm config driver

### LINKED ISSUE #2861: E3SM config files do not take driver into account
Now that multiple drivers/couples are available, E3SM config files need to take the driver of the current case into account instead of assuming mct. The file that needs to be modified is:

```
config/e3sm/config_files.xml
```

As an example, for CESM, the value of CONFIG_CPL_FILE is:

```
$COMP_ROOT_DIR_CPL/cime_config/config_component.xml
```

and for E3SM the value is:

```
$CIMEROOT/src/drivers/mct/cime_config/config_component.xml
```

Looking at the two values, you can see that E3SM is hardcoded to mct while CESM is flexible. There are many entries that will need to be made flexible like this in E3SM.

Once this is done, the test:

```
 TestCreateNewcase.test_m_createnewcase_alternate_drivers
```

... should PASS.

jedwards4b:
Allow nuopc driver to run with model e3sm
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2861 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/e3sm/config_files.xml
+++ b/config/e3sm/config_files.xml
@@ -96,7 +96,7 @@
     <default_value>unset</default_value>
     <values>
       <value component="allactive">$CIMEROOT/config/$MODEL/allactive/config_compsets.xml</value>
-      <value component="drv"      >$CIMEROOT/src/drivers/mct/cime_config/config_compsets.xml</value>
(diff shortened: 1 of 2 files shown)


jgfouca:
Thanks, guys!

PR #2782, opened 2018-08-25
Attach the instance number to 'cpl' in cpl history file names,

kdraeder:
instead of attaching to the file type (.ha2x, etc.).
Also incidentally includes a fix in case.py
for a problem with fetching GRIDS_SPEC_FILE for
user defined grids (#1798).

Test suite: scripts_regression_tests.py: passes all
Test baseline: 
Test namelist changes: none
Test status: bit for bit

Fixes #2754

User interface changes?: None

Update gh-pages html (Y/N)?:  No

Code review:

PR #2879, opened 2018-10-26
Strawman fix of #2832, which broke the non-r2x files.

### LINKED ISSUE #2831: Coupler history file hr2x problems for forecasts < 24 hours.

kdraeder:
For runs shorter than 24 hours, the names of all of the
coupler history files need to use the ymd and tod of the end
of the run, instead of the current time.  And the correct
number of time slots in each file must be used.
The proposed fixes are not complete: they ignore
> other flows (l2x, x2a, ...)
> other forecast lengths that could divide the day; 1,2,3,4,12 hours
They also are coded in an inflexible way.

Test suite: Single instance test ran correctly for hours 6,12,18,0 of a day,
in a CAM6 build from cesm2.1-alphabranch
Test baseline: 
Test namelist changes: none
Test status: bit-level changes to cpl aux hist files due to averaging span changes.

Fixes #2831 (again)

User interface changes?: If they need to use cpl hist files from runs shorter than 24 hours.

Update gh-pages html (Y/N)?:probably, but not done

Code review: none yet

### DIFF
--- a/src/drivers/mct/main/cime_comp_mod.F90
+++ b/src/drivers/mct/main/cime_comp_mod.F90
@@ -3765,14 +3765,15 @@ contains
           if (do_hist_a2x24hr) then
              do eai = 1,num_inst_atm
                 inst_suffix = component_get_suffix(atm(eai))
+                write_hist_alarm = t24hr_alarm .or. stop_alarm
                 if (trim(hist_a2x24hr_flds) == 'all') then
                    call seq_hist_writeaux(infodata, EClock_d, atm(eai), flow='c2x', &
                         aname='a2x1d',dname='doma',inst_suffix=trim(inst_suffix),  &
-                        nx=atm_nx, ny=atm_ny, nt=1, write_now=t24hr_alarm)
+                        nx=atm_nx, ny=atm_ny, nt=1, write_now=write_hist_alarm)
                 else
                    call seq_hist_writeaux(infodata, EClock_d, atm(eai), flow='c2x', &
                         aname='a2x1d',dname='doma',inst_suffix=trim(inst_suffix),  &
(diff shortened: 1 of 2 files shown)


billsacks:
I think the title of this PR is correct. My sense is that the problem actually originated with #2782 -- is that your understanding, too?

billsacks:
> I think the title of this PR is correct. My sense is that the problem actually originated with #2782 -- is that your understanding, too?

I should add that I have no idea how #2782 would have broken this, but based on your description of the problem, that seems like a more likely culprit than #2832 .

billsacks:
Oh, now I maybe understand: My new understanding is that neither #2832 nor #2782 actually broke anything new, but the problem is that #2832 wasn't a complete fix... is that the correct understanding?

kdraeder:
@billsacks 
Yes, 2832 wasn't a complete fix.  It fixed the hr2x problem,
partly by adding the TOD to the file names.  but it ignored
the effects on the other coupler history files, which were
created, but filled with 0s.

billsacks:
Closing as per our discussion (@kdraeder will make some one-off fixes on a branch, but we won't try to fix this on master or maint-5.6)

Issue #2886, opened 2018-10-29
CHARGE_ACCOUNT no longer allows for sub-projects


jhkennedy:
When using the `-p/--project` option in CIME, it is assumed that the account to charge for compute time and project directory are the same. However, on machines like Titan and Summit, time often needs to be charged to a subproject like `cli106ice`, but the project directory doesn't reflect the subproject name (`/lustre/atlas/proj-shared/cli106ice/` is not a valid path, and instead should be `/lustre/atlas/proj-shared/cli106/`). This is discussed in issue E3SM-Project/E3SM#1588

PR E3SM-Project/E3SM#1718 added an optional `CHARGE_ACCOUNT` which allow for charging sub-projects while using the `PROJECT` name for directory paths. 

However, issue ESMCI/cime#2676 thought that `CHARGE_ACCOUNT` and `PROJECT` should always be the same, and PR ESMCI/cime#2763 removed the ability to use subprojects. 

Because Summit will be a supported machine, and Titan is a machine currently used,  ESMCI/cime#2763 should be reverted. 

@billsacks , @sarich

billsacks:
@jhkennedy - I may be remembering/understanding wrong, but I don't think that simply reverting #2763 is the right answer here. According to the `create_newcase` documentation:

```
  --project PROJECT, -project PROJECT
                        Specify a project id as used in batch system accounting.
```

yet without #2763 setting `--project` has no effect on the account code used, at least on some machines. Can you state your suggested user interface for setting both project and charge accounts? I'd argue (fairly strongly) that any change should preserve the ability for a simple setting of the `--project` argument to set all necessary xml variables correctly on machines for which there is no distinction between project and charge_account (in order to maintain backwards compatibility with users' scripts that just set `--project`).

jhkennedy:
@billsacks you're definitely right about the lack of documentation of how these two options interplay -- I'll update that for sure for whatever fix we come up with.

billsacks:
@jhkennedy I'm pretty sure that, in my case, I had project set in some way (either in my environment or via a .cesm_proj file), but _not_ CHARGE_ACCOUNT. Prior to #2763 , the logic for CHARGE_ACCOUNT didn't work exactly as it was documented: The actual operation of step (4) in the hierarchy should have been stated (more precisely) as: "default to value given by get_project". Note that `get_project` does not itself know anything about the `--project` command-line flag.

[...]

PR #2888, opened 2018-10-29
Allow CHARGE_ACCOUNT to use subprojects and default to command line given project

### LINKED ISSUE #2676: CHARGE_ACCOUNT doesn't listen to --project flag
I think this is a bug, but I don't really understand `CHARGE_ACCOUNT`, so I may be wrong:

If I do:

```
./create_test --no-build --project foo --test-root . SMS.f19_g16.X.cheyenne_intel
```

Then this is correct:

```
 ./xmlquery PROJECT

	PROJECT: foo

Results in group case.test
	PROJECT: foo
```

but this is wrong:

```
./xmlquery CHARGE_ACCOUNT

	CHARGE_ACCOUNT: P93300601

Results in group case.test
	CHARGE_ACCOUNT: P93300601
```

I would expect `CHARGE_ACCOUNT` to also be `foo` in this case.

It looks like the `--project` flag on the command-line never makes it into the logic used to determine `charge_account` (the `get_charge_account` call from `case.py`).

jhkennedy:
CHARGE_ACCOUNT was not falling back to the project value given on the command line in  #2676 and was fixed in #2763 . However, #2763 no longer allowed for subprojects to be charged on OLCF machines. This reverts #2763 and fixes both #2676 and #2886. 

Test suite: scripts_regression_tests.py on Titan
Test baseline: N/A
Test namelist changes: N/A
Test status: [bit for bit]

Fixes #2676 and #2886

User interface changes?: None

Code review:

### DIFF
--- a/doc/source/users_guide/cime-config.rst
+++ b/doc/source/users_guide/cime-config.rst
@@ -16,15 +16,15 @@ CIME recognizes a user-created custom configuration directory, ``$HOME/.cime``. 
 
   * ``PROJECT=<account number>``
 
-    This is your project account code for batch submission and/or directory priveleges
+    Used to specify a project id for compute accounting and directory permissions when on a batch system.
 
   * ``CHARGE_ACCOUNT=<account number>``
 
-    An alternative to PROJECT for batch charging>
+    Used to override the accounting (only) aspect of PROJECT
 
   * ``MAIL_USER=<email address>``
 
-    Used request a non-default email for batch summary output
+    Used to request a non-default email for batch summary output
 
   * ``MAIL_TYPE=[never,all,begin,fail,end]``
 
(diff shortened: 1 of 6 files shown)


jhkennedy:
@billsacks I updated a little of the documentation; what do you think of those updates?

billsacks:
If you want this to auto-close #2886 you need to list things a little differently in the PR comment:
Fixes #2676 
Fixes #2886

billsacks:
(i.e., you need a separate "Fixes" for each issue.) (My comment isn't enough to auto-close those: you'll need to edit your initial comment.)

jhkennedy:
@billsacks Woops! You're right. Fixed my fixes.

PR #2892, opened 2018-10-30
Remove redundant check of mpi-tag values

### LINKED ISSUE #2890: PR #2804 broke builds with mpi-serial
#2804 added, among other things, a reference to `MPI_ADDRESS_KIND`. Builds with mpi-serial are now failing (example: `SMS_Mmpi-serial_P1x1.f09_g16.S.cheyenne_gnu`) with:

```
/gpfs/fs1/work/sacks/ctsm_code/current_branch1/cime/src/drivers/mct/main/component_mod.F90:66:32:

   integer (kind=MPI_ADDRESS_KIND) :: max_mpi_tag
                                1
Error: Symbol ‘mpi_address_kind’ at (1) has no IMPLICIT type
```

I see MPI_ADDRESS_KIND defined in `src/externals/mct/mpi-serial/mpif.F90`, but I'm thinking that the issue is that component_mod has:

```
#include <mpif.h>
```

but NOT

```
use mpi
```

Should this be changed to `use mpi`, or should mpi-serial be changed to move MPI_ADDRESS_KIND to the header file?

I'm labeling this as critical, because it's blocking our upgrade to a more recent cime version in CTSM.

@jhkennedy @jedwards4b @ekluzek @fischer-ncar

jhkennedy:
Because MPI-serial doesn't have the `MPI_attr_get` function,  PR #2804 broke MP-serial builds as discussed in #2890 . This removes a redundant exception that uses `MPI_attr_get` to determine the max MPI tag value, and report it and when it is exceed. MPI will still throw an exception when it is exceeded, but doesn't report the maximum value. I also added an MPI-serial test to cime_developer test suite. 

Test suite: ./scripts_regression_tests on titan, ./create_test cime_developer (with new SMS_D_Ln9_Mmpi-serial.f19_g16_rx1.A test) on anvil
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2890 

User interface changes?: N

Update gh-pages html (Y/N)?: N/A

Code review:

### DIFF
--- a/scripts/lib/get_tests.py
+++ b/scripts/lib/get_tests.py
@@ -64,7 +64,6 @@ _CIME_TESTS = {
                              "IRT_N2.f19_g16_rx1.A",
                              "ERR.f45_g37_rx1.A",
                              "ERP.f45_g37_rx1.A",
-                             "SMS_D_Ln9.f19_g16_rx1.A",
                              "SMS_D_Ln9_Mmpi-serial.f19_g16_rx1.A",
                              "DAE.ww3a.ADWAV",
                              "PET_P4.f19_f19.A",
--- a/scripts/lib/get_tests.py
+++ b/scripts/lib/get_tests.py
@@ -65,6 +65,7 @@ _CIME_TESTS = {
                              "ERR.f45_g37_rx1.A",
                              "ERP.f45_g37_rx1.A",
                              "SMS_D_Ln9.f19_g16_rx1.A",
+                             "SMS_D_Ln9_Mmpi-serial.f19_g16_rx1.A",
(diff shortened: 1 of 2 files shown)


scripts/lib/get_tests.py
```diff
@@ -65,6 +65,7 @@
                              "ERR.f45_g37_rx1.A",
                              "ERP.f45_g37_rx1.A",
                              "SMS_D_Ln9.f19_g16_rx1.A",
```
jedwards4b: I think that you should just replace the existing one rather than add a new one.

jhkennedy: Done!

Issue #2896, opened 2018-10-31
PR #2811 accidentally reverted a regex change in hist_utils.py


billsacks:
PR #2811 (the big nuopc merge) accidentally reverted a regex change in hist_utils.py:get_extension that was made in #2728 . This causes failures like this (this just shows up in a single test in the CTSM test suite; I don't expect this is a common problem):

```
SystemExit: ERROR: Failed to get extension for file '/glade/scratch/sacks/ERS_Ly3.f10_f10_musgs.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic.GC.1030-140553ch_int/run/ERS_Ly3.f10_f10_musgs.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic.GC.1030-140553ch_int.cpl.hl2x1yr_glc.0001-01-01-00000.nc'
```

I am reconciling the changes in #2811 with those in #2728 to come up with new regex logic for this function. I'll submit a PR with a fix soon.

PR #2811, opened 2018-09-21
Jedwards/merge cmeps

jedwards4b:
This is the merge of the nuopc development branch.   Really sorry for the size of this PR, we should have merged much sooner.

Test suite: scripts_regression_tests.py (all pass) 
                   create_test --xml-testlist ./testlist_cmeps.xml --xml-machine cheyenne --xml-category prealpha  --compare sept19 --baseline-root /glade/scratch/dunlap/BASELINES
                   I also ran cime_developer and compared baselines against cime master 
                   All tests pass with the exception of 
    FAIL ERI.f09_g16.X.cheyenne_intel BASELINE cimemaster0921: DIFF
    FAIL ERIO.f09_g16.X.cheyenne_intel BASELINE cimemaster0921: DIFF
    FAIL IRT_N2.f19_g16_rx1.A.cheyenne_intel BASELINE cimemaster0921:
    The x cases are inconsequential and the A case was failing on master and works on the branch.

Test baseline: cime master 3effd6
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

PR #2897, opened 2018-10-31
Fix regex matches in hist_utils: get_extension

billsacks:
PR ESMCI/cime#2811 (the big nuopc merge) accidentally reverted a regex
change in hist_utils.py:get_extension that was made in
ESMCI/cime#2728. This PR brings back the intended, more general regex,
and consolidates some near-duplicate regular expressions into a single
expression with a variable.

Also ran the test that caused the failure noted in #2896
(ERS_Ly3.f10_f10_musgs.I1850Clm50BgcCropCmip6.cheyenne_intel.clm-basic
from a CTSM branch)
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#2896

User interface changes?: N

Code review:

### DIFF
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -434,7 +434,7 @@ def get_extension(model, filepath):
     ext_regexes.append(r'\w+')
 
     for ext_regex in ext_regexes:
-        full_regex_str = model+r'\d?_?(\d{4})?\.('+ext_regex+')[-\w\.]*\.nc\.?'
+        full_regex_str = model+r'\d?_?(\d{4})?\.('+ext_regex+r')[-\w\.]*\.nc\.?'
         full_regex = re.compile(full_regex_str)
         m = full_regex.search(basename)
         if m is not None:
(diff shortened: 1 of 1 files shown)


scripts/lib/CIME/hist_utils.py
```diff
@@ -421,21 +421,24 @@ def get_extension(model, filepath):
...
+    if model == "mom":
+        # Need to check 'sfc.day' specially: the embedded '.' messes up the
+        # general-purpose regex
+        ext_regexes.append(r'sfc\.day')
+
+    # Now add the general-purpose extension regex
+    ext_regexes.append(r'\w+')
+
+    for ext_regex in ext_regexes:
+        full_regex_str = model+r'\d?_?(\d{4})?\.('+ext_regex+r')[-\w\.]*\.nc\.?'
```
billsacks: This regex is the same as what was used prior to #2811 except that I am now using the variable ext_regex in place of '\w+', so that we can match different extensions with this same regex.

PR #2915, opened 2018-11-19
ocn glc coupling (NOT ready for merge)

alperaltuntas:
This PR is for review only!

@billsacks, I think I am mostly done with standalone ocn (CMOMu) and ocn2glc (CMOMg) changes in CIME. (I am able to run MOM6+CISM compsets and pass fields from MOM6 to CISM) Before working on glc2ocn changes to finalize MISOMIP and ocn<->glc infrastructure, I wanted to touch base with you, and perhaps @mvertens too, at this intermediate stage to give you a chance to review my changes before there are even more commits. Let me know if you'd prefer to review all at once instead.

Some specific questions:

1. It looks like most subroutines in prep_glc_mod.F90 were coded with only lnd <-> glc coupling in mind e.g., in prep_glc_merge subroutine. So I added new subroutines specifically for ocn <-> glc coupling, e.g., prep_glc_merge_o2g, so as to reduce my chances of breaking lnd <-> glc coupling, and since there are substantial differences between those two types of coupling. However, let me know if it would be better to combine such subroutines to avoid any code duplication (which isn't much).

2. Any other issues in prep_glc_mod.F90 that you spotted?

3. Which test suite would be best to find out whether I broke any other glc compsets?

* there are debug print statements here and there which will be removed in the actual PR.

Also mentioning @whlipscomb, @Katetc, @gunterl, and @gustavo-marques if they'd like to follow this.

Test suite: n/a
Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes: n/a

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


rljacob:
Also mentioning @jonbob , @stephenprice and @matthewhoffman

mvertens:
What tests did you run for this PR? Can you please document this in the PR? Given that these changes might effect other compsets it would be important to run at least all of the prealpha tests to verify that bfb is still maintained.

alperaltuntas:
Thanks, @mvertens. Closing this for now. Will re-open once testing is complete and the issues I identified are resolved.

PR #2923, opened 2018-11-27
Update hobart to PGI 18.10

### LINKED ISSUE #2922: Update hobart PGI compiler
PGI 18.10 is available on hobart, and it would be great if `config/cesm/machines/config_machines.xml` was updated to use it (we want to remove a PGI bug workaround from MARBL, but can't do that if CESM is still using a buggy version of the compiler).

I'm happy to put together a PR if someone lets me know the target branch / branches.

mnlevy1981:
Test suite: Single case on hobart with PGI compiler
Test baseline: Generated from original code
Test namelist changes: None
Test status: round-off for hobart_pgi tests

Fixes #2922

User interface changes?: No

Code review:


jedwards4b:
See #2924

PR #2926, opened 2018-11-27
Jgfouca/branch for acme split 2018 11 27

jgfouca:
Changes:
* Update domain masks, mapping and init files for ne240
* introduce preqx_kokkos in HOMME, and as optional target for CAM
* Add model-specific logging capability
* cime_bisect: needs to handle submodules
* Move timing library customizations from perf_mod to GPTL
* Add data iceberg support
* Fix for cime's configure tool
* Add e3sm support for latband

Conflicts (need extra review):
* config/e3sm/machines/config_machines.xml
* scripts/Tools/Makefile
* scripts/lib/CIME/SystemTests/funit.py
* scripts/lib/CIME/SystemTests/system_tests_common.py
* scripts/lib/CIME/case/case_run.py
* scripts/lib/CIME/case/case_setup.py
* scripts/lib/CIME/hist_utils.py
* scripts/lib/CIME/utils.py
* src/components/data_comps/datm/cime_config/namelist_definition_datm.xml
* src/drivers/mct/main/cime_comp_mod.F90

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


scripts/lib/CIME/get_timing.py
```diff
@@ -65,13 +65,13 @@ def gettime2(self, heading_padded):
 
         heading = '"' + heading_padded.strip() + '"'
         for line in self.finlines:
-            m = re.match(r'\s*{}\s*(\d+)\s*\d+\s*(\S+)'.format(heading), line)
+            m = re.match(r'\s*{}\s+\S\s+(\d+)\s*\d+\s*(\S+)'.format(heading), line)
```
jgfouca: @jedwards4b , please take a look at this.

src/components/data_comps/datm/cime_config/namelist_definition_datm.xml
```diff
@@ -294,8 +294,9 @@
       <value stream="CORE_IAF_JRA.NCEP.U_10">domain.atm.TL319.151007.nc</value>
       <value stream="CORE_IAF_JRA.NCEP.V_10">domain.atm.TL319.151007.nc</value>
       <value stream="CORE_IAF_JRA.CORE2.ArcFactor">CORE2.t_10.ArcFactor.T62.1997-2004.nc</value>
+      <value stream="co2tseries.20tr" cime_model="e3sm">fco2_datm_1765-2007_c100614.nc</value>
```
jgfouca: This file was heavily conflicted.

src/components/data_comps/drof/cime_config/config_component.xml
```diff
@@ -13,12 +13,15 @@
   -->
 
   <description modifier_mode="1">
-    <desc rof="DROF[%NULL][%NYF][%IAF][%CPLHIST][%JRA]">Data runoff model</desc>
-    <desc option="NULL">NULL mode</desc>
-    <desc option="NYF" >COREv2 normal year forcing:</desc>
-    <desc option="IAF" >COREv2 interannual year forcing:</desc>
-    <desc option="CPLHIST">CPLHIST mode:</desc>
-    <desc option="JRA">JRA55 interannual forcing</desc>
+    <desc rof="DROF[%NULL][%NYF][%IAF][%IAFAIS00][%IAFAIS45][%IAFAIS55][%CPLHIST][%JRA]">Data runoff model</desc>
```
jgfouca: Potential cesm impacts?

src/components/data_comps/drof/cime_config/namelist_definition_drof.xml
```diff
@@ -49,11 +49,17 @@
     <group>streams_file</group>
     <desc>List of streams used for the given drof_mode.</desc>
     <values>
-      <value drof_mode="NULL"           >NULL</value>
-      <value drof_mode="CPLHIST"        >rof.cplhist</value>
-      <value drof_mode="DIATREN_ANN_RX1">rof.diatren_ann_rx1</value>
-      <value drof_mode="DIATREN_IAF_RX1">rof.diatren_iaf_rx1</value>
-      <value drof_mode="IAF_JRA"        >rof.iaf_jra</value>
+      <value drof_mode="NULL"                 >NULL</value>
```
jgfouca: Potential cesm impacts?

src/drivers/mct/main/seq_diag_mct.F90
```diff
@@ -1402,6 +1404,8 @@ subroutine seq_diag_ocn_mct( ocn, xao_o, frac_o, infodata, do_o2x, do_x2o, do_xa
        if (first_time) then
           index_x2o_Fioi_melth  = mct_aVect_indexRA(x2o_o,'Fioi_melth')
           index_x2o_Fioi_meltw  = mct_aVect_indexRA(x2o_o,'Fioi_meltw')
+          index_x2o_Fioi_bergh  = mct_aVect_indexRA(x2o_o,'PFioi_bergh')
```
jgfouca: needs review.

src/drivers/mct/shr/seq_flds_mod.F90
```diff
@@ -1412,6 +1412,24 @@ subroutine seq_flds_set(nmlfile, ID, infodata)
     attname  = 'Fioi_meltw'
     call metadata_set(attname, longname, stdname, units)
 
+    ! Heat flux from melting icebergs
+    call seq_flds_add(i2x_fluxes,"PFioi_bergh")
+    call seq_flds_add(x2o_fluxes,"PFioi_bergh")
+    longname = 'Heat flux from melting icebergs'
+    stdname  = 'surface_iceberg_melt_heat_flux'
```
jgfouca: needs review.

scripts/lib/CIME/case/case_run.py
```diff
@@ -50,14 +50,14 @@ def _pre_run_check(case, lid, skip_pnl=False, da_cycle=0):
     # This needs to be done everytime the LID changes in order for log files to be set up correctly
     # The following also needs to be called in case a user changes a user_nl_xxx file OR an env_run.xml
     # variable while the job is in the queue
-    logger.debug("{} NAMELIST CREATION BEGINS HERE".format(time.strftime("%Y-%m-%d %H:%M:%S")))
+    model_log("e3sm", logger, "{} NAMELIST CREATION BEGINS HERE".format(time.strftime("%Y-%m-%d %H:%M:%S")))
```
jedwards4b: I don't think this should be hardcoded to e3sm.

scripts/lib/CIME/get_timing.py
jedwards4b: Was there a change in the file format?  This seems to match a different string than before - any documentation on this change?

scripts/lib/CIME/utils.py
```diff
@@ -1787,3 +1787,7 @@ def string_in_list(_string, _list):
         if _string.lower() == x.lower():
             return x
     return None
+
+def model_log(model, arg_logger, msg):
```
jedwards4b: Okay - now I see why e3sm exclusively above - but this breaks our use of logger.debug to get these messages.  
I think that you may be able to fix it with
else:
  arg_logger.debug(msg) 
but that assumes that all of the messages replaced were debug messages.

src/components/data_comps/drof/cime_config/config_component.xml
jedwards4b: I think it's okay as long as IAF and IAF* are not confused anywhere.

src/components/data_comps/drof/cime_config/config_component.xml
```diff
@@ -13,12 +13,15 @@
...
+    <desc option="NULL"     >NULL mode</desc>
+    <desc option="NYF"      >COREv2 normal year forcing:</desc>
+    <desc option="IAF"      >COREv2 interannual year forcing:</desc>
+    <desc option="IAFAIS00" >COREv2 interannual year forcing:</desc>
```
jedwards4b: Can we fix the descriptions to distinguish these?

src/drivers/mct/main/seq_diag_mct.F90
jedwards4b: @billsacks can you have a look at these

src/share/timing/gptl.c
```diff
@@ -5363,4 +5889,15 @@ static void print_threadmapping (FILE *fp)
...
+**
+** Return value: true (1) or false (0)
+*/
+
+static int serial_region ()
+{
+  return 1;
+}
+
 #endif
```
jedwards4b: I guess my question about the gptl changes is - are we getting closer to the gptl git code or are we diverging further?

scripts/lib/CIME/case/case_run.py
jgfouca: If I recall, you guys (cesm) did not want this output as logger.info, so I made it logger.debug. That messed-up folks on E3SM since they had come to expect it, so this was the best compromise.

scripts/lib/CIME/get_timing.py
jgfouca: @worleyph

jgfouca: https://github.com/E3SM-Project/E3SM/pull/2537

scripts/lib/CIME/utils.py
jgfouca: Yes, all the replaced calls were debug messages. I will see if I can restore these for you.

scripts/lib/CIME/case/case_run.py
jedwards4b: RIght but see my note below - now we don't get it even with logger set to debug

scripts/lib/CIME/get_timing.py
worleyph: Is the pull request pointer sufficient to explain this? It makes the per process output file and global statistics file format consistent by adding an "on" column to the global statistics file.

jedwards4b: @worleyph yes once I found it.

src/drivers/mct/main/seq_diag_mct.F90
billsacks: I am not very familiar with the seq_diag code. @alperaltuntas are you familiar with this? (I'm asking you because it seems like this code is mainly referenced by ocean folks... if you're not familiar with it either, do you have suggestions for who might be able to review it?)

That said: My initial concern is what happens in this code if (as will be the case for CESM) these coupling indices and/or the data stored within them are undefined.

billsacks:
@alperaltuntas requesting a review from you just for the sake of the seq_diag changes... feel free to pass along to someone else if you don't feel able to review this either and/or you have a sense of who would be better able to review those changes.

jedwards4b:
I am running cesm prealpha tests against a cesm2_2_alpha01d baseline.

jedwards4b:
Failing two tests in scripts_regression_tests.py: 
H_TestMakeMacros.test_compiler_changeable_at_build_time
I_TestCMakeMacros.test_compiler_changeable_at_build_time

jedwards4b:
The tests appear to be failing on master too but only on cheyenne.  cdash shows that this test passed 11 hours ago.

src/share/timing/gptl.c
jgfouca: @worleyph ?

jedwards4b:
Okay the failure I was encountering was related to a module error on the cheyenne system and not to this PR

src/drivers/mct/main/seq_diag_mct.F90
billsacks: From @mvertens on telecon: Below, we should check if indices of these new variables are 0; if so, use the old calculation, if non-zero, use the new calculation.

@alperaltuntas no need for you to review further unless you want to.

src/share/timing/gptl.c
worleyph: Diverging further, but talk to @rljacob . At Rob's request, I went through and added all of the timing library customizations back into the February 2018 version of GPTL . (I couldn't build the most recent version, despite some suggestions from Jim Rosinski, but there were almost no source changes since February - just moving to autoconf and reorganizing the directory structure.) I won't be able to submit PRs to the GPTL git repository until someone can hold my hand and show me how to build the most recent version. I also do not know how to modify CIME to use the public release of GPTL, which is a requirement for me to test this there. I also do not know if Jim Rosinski will be willing to accept the modifications (though they are almost all just new options, and are disabled by default unless using perf_mod.F90). Jim R. doesn't know either, since he hasn't seen any PRs yet :-).

rljacob: @jedwards4b these are closer enough.

jedwards4b: I'll accept it.   I'll try to make a cime branch with the latest gptl build for @worleyph

scripts/lib/CIME/case/case_run.py
jgfouca: Done.

src/drivers/mct/main/seq_diag_mct.F90
jgfouca: @mvertens , @billsacks , can you confirm my change below?

billsacks: It looks mostly right, but elsewhere where we have code like this, we use `perrWith='quiet'` in the call to mct_aVect_indexRA. See prep_rof_mod for an example. I think @mvertens suggested this to me, so I'll defer to her as to whether that's needed... I think it may be needed to avoid having mct die when you try to get a non-existent index?

Also, minor quibble: can you change `.eq.` to `==`? I don't think we're totally consistent within cime, but we tend to prefer `==` over `.eq.`, and at least try to stay consistent within a module.

jgfouca: @billsacks , done

billsacks:
Looks good to me now - thanks!

PR #2932, opened 2018-12-03
Big improvement to robustness of CIME's XML handling

### LINKED ISSUE #2161: Current state of CIME data files
Hi all,

This is a follow-on discussion to the XML API discussion work that has been completed (not yet merged though as of this typing). Using the new API, I've confirmed that scanning for children is killing our performance. Note the 46% of runtime spent in scan_children (it would be a far higher % except for the fact that loading environment modules is crazy slow on this machine):

screen shot 2017-12-19 at 3 46 42 pm

And the reason that we need to do so many scans is because it's hard to utilize assumptions about file structure to do vastly-more-efficient direct-child searches because our file formats are so inconsistent.

To help inform the discussion, here's a class diagram of the python classes in XML:

screen shot 2017-12-19 at 3 50 00 pm

Black lines denote inheritance, red lines denote a "has-a" relationship with the critical Case class. Note that the case class splits it's has-a relationships to XML files between "entry-id" and "generic" files.

Diving deeper, here's an analysis of the files themselves:

```
Archive(GenericXML):
  * CIME/config/$MODEL/config_archive.xml
  * Format: custom

Batch(GenericXML):
  * CIME/config/$MODEL/machines/config_archive.xml
  * Format: custom

Compilers(GenericXML):
  * CIME/config/$MODEL/machines/config_compilers.xml
  * Format: custom, ACME and CESM have different formats (1.0 vs. 2.0)

Compsets(GenericXML):
  * CIME/config/acme/allactive/config_compsets.xml
  * CIME/src/drivers/mct/cime_config/config_compsets.xml
  * components can add more
  * Format: custom

EnvArchive(GenericXML):
  * $CASE/env_archive.xml : Note, this is the ONLY direct subclass of GenericXML that lives in the CASE directory
  * Format: custom, similar to Archive

Grids(GenericXML):
  * CIME/config/$MODEL/config_grids.xml
  * Format: custom, ACME and CESM have different formats (1.0 vs. 2.0)

Machines(GenericXML):
  * CIME/config/$MODEL/machines/config_machines.xml
  * Format: custom

Pes(GenericXML):
  * CIME/config/acme/allactive/config_pesall.xml
  * Format: custom

TestReporter(GenericXML):
  * ? (CESM only)
  * Format: custom

Tests(GenericXML):
  * CIME/config/config_tests.xml
  * Format: custom

TestSpec(GenericXML):
  * ? (CESM only)
  * Format: custom



Component(EntryId):
  * CIME/src/drivers/mct/cime_config/config_component.xml + config_component_$MODEL.xml
  * Format: entry_id->entry->(various entry info)

Headers(EntryId):
  * CIME/config/config_headers.xml
  * Format: custom, why is this a subclass of EntryId?

Files(EntryId):
  * CIME/config/config_headers.xml
  * Format: entry_id->entry->(various entry info)

NamelistDefinitions(EntryId):
  * Lots
  * Format: entry_id->entry->(various entry info)

PIO(EntryId):
  * CIME/config/$MODEL/machines/config_pio.xml
  * Format: entries->entry->(values) , why different root element tagname?



EnvBatch(EnvBase):
  * $CASE/env_batch.xml
  * Format: file->group(job)->entry + noncompliant batch_system block

EnvBuild(EnvBase):
  * $CASE/env_build.xml
  * Format: file->group->entry fully compliant file!

EnvCase(EnvBase):
  * $CASE/env_case.xml
  * Format: file->group->entry fully compliant file!

EnvMachPes(EnvBase):
  * $CASE/env_mach_pes.xml
  * Format: file->group->entry fully compliant file!

EnvMachSpecific(EnvBase):
  * $CASE/env_mach_specific.xml
  * Format: custom

EnvRun(EnvBase):
  * $CASE/env_run.xml
  * Format: file->group->entry fully compliant file!

EnvTest(EnvBase):
  * $CASE/env_test.xml
  * Format: file->group(job)->entry + noncompliant TEST block
```

The things that stick out to me:
* EnvArchive is strange. It's the only direct subclass of GenericXML that lives in the case directory, but it's totally non compliant with our standard env file formatting.
* Headers seems like is should be a direct subclass of GenericXML, not an EntryId, since it has no entries.
* The root element of PIO is inconsistent with its siblings
* EnvBatch is almost compliant except for one block
* EnvTest is almost compliant except for one block
* EnvMachPes makes no effort at all to be compliant

Short-term path forward:
* Re-work class hierarchy to differentiate compliant files from noncompliant
* The performance of all compliant Env*.xml files can be drastically increased by leveraging the consistent file structure they all share when retrieving entry id values
* This would most-likely involve a cached entry_name -> (group or entry node) map

Long-term path forward:
* It sure would be nice if there was more consistency all around
* break backwards compatibility by forcing more Env* files to be compliant
* more consistency between the direct subclasses of GenericXML would be nice (config_machines, config_grid), but is not a necessity since create_newcase is already pretty fast and the penalties for parsing these files is only incurred once (at create_newcase).

### LINKED ISSUE #2850: XML Caching needs to be more robust
User should never have to disable caching. Propose adding sha1 to FILEMAP to check that file was not changed. Maybe need addition class-level data in GenericXML to ensure that we never have multiple writable GenericXML objects open to the same file at the same time.

jgfouca:
Change list:
* Propagate Case read_only to its XML objects
* Cache all files, but use file mod-time to detect changes.
* Try to pass Case, not caseroot, to run_sub_or_cmd
* Add a CIME performance test
* Add tests for CIME XML handling

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2161 
Fixes #2850 

User interface changes?: GenericXML will now throw an error if XML files are changed behind its back without a re-read/invalidate.

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -38,6 +38,11 @@ class GenericXML(object):
     _FILEMAP = {}
     DISABLE_CACHING = False
 
+    @classmethod
+    def invalidate(cls, filename):
+        if filename in cls._FILEMAP:
+            del cls._FILEMAP[filename]
+
     def __init__(self, infile=None, schema=None, root_name_override=None, root_attrib_override=None, read_only=True):
         """
         Initialize an object
@@ -71,17 +76,25 @@ class GenericXML(object):
 
             self.tree = ET.ElementTree(root)
 
+            self._FILEMAP[infile] = (self.tree, self.root, 0.0)
+
     def read(self, infile, schema=None):
         """
         Read and parse an xml file into the object
         """
-        if not self.DISABLE_CACHING and infile in self._FILEMAP and self.read_only:
-            logger.debug("read (cached): " + infile)
-            expect(self.read_only or not self.filename or not self.needsrewrite, "Reading into object marked for rewrite, file {}"
-                   .format(self.filename))
-            self.tree, self.root = self._FILEMAP[infile]
-        else:
-            logger.debug("read: " + infile)
+        cached_read = False
+        if not self.DISABLE_CACHING and infile in self._FILEMAP:
+            timestamp_cache = self._FILEMAP[infile][2]
+            timestamp_file  = os.path.getmtime(infile)
+            if timestamp_file == timestamp_cache:
+                logger.debug("read (cached): {}".format(infile))
+                expect(self.read_only or not self.filename or not self.needsrewrite, "Reading into object marked for rewrite, file {}"
+                       .format(self.filename))
+                self.tree, self.root, _ = self._FILEMAP[infile]
+                cached_read = True
+
+        if not cached_read:
+            logger.debug("read: {}".format(infile))
             file_open = (lambda x: open(x, 'r', encoding='utf-8')) if six.PY3 else (lambda x: open(x, 'r'))
             with file_open(infile) as fd:
                 self.read_fd(fd)
@@ -91,7 +104,7 @@ class GenericXML(object):
 
             logger.debug("File version is {}".format(str(self.get_version())))
 
-            self._FILEMAP[infile] = (self.tree, self.root)
+            self._FILEMAP[infile] = (self.tree, self.root, os.path.getmtime(infile))
 
     def read_fd(self, fd):
         expect(self.read_only or not self.filename or not self.needsrewrite, "Reading into object marked for rewrite, file {}"               .format(self.filename))
@@ -289,6 +302,12 @@ class GenericXML(object):
         """
         Write an xml file from data in self
         """
+        timestamp_cache = self._FILEMAP[self.filename][2]
+        if timestamp_cache != 0.0:
+            timestamp_file  = os.path.getmtime(self.filename)
+            expect(timestamp_file == timestamp_cache,
+                   "File {} appears to have changed without a corresponding invalidation, modtimes {:0.2f} != {:0.2f}".format(self.filename, timestamp_cache, timestamp_file))
+
         if not (self.needsrewrite or force_write):
             return
 
@@ -310,6 +329,9 @@ class GenericXML(object):
         else:
             with open(outfile,'w') as xmlout:
                 xmlout.write(xmlstr)
+
+        self._FILEMAP[self.filename] = (self.tree, self.root, os.path.getmtime(self.filename))
+
         self.needsrewrite = False
 
     def scan_child(self, nodename, attributes=None, root=None):
--- a/scripts/lib/CIME/case/case_run.py
+++ b/scripts/lib/CIME/case/case_run.py
@@ -248,7 +248,7 @@ def _do_external(script_name, caseroot, rundir, lid, prefix):
     filename = "{}.external.log.{}".format(prefix, lid)
     outfile = os.path.join(rundir, filename)
     append_status("Starting script {}".format(script_name), "CaseStatus")
-    run_sub_or_cmd(script_name, [caseroot], (os.path.basename(script_name).split('.',1))[0], [caseroot], logfile=outfile)
+    run_sub_or_cmd(script_name, [caseroot], (os.path.basename(script_name).split('.',1))[0], [caseroot], logfile=outfile) # For sub, use case?
     append_status("Completed script {}".format(script_name), "CaseStatus")
 
 ###############################################################################
@@ -257,7 +257,7 @@ def _do_data_assimilation(da_script, caseroot, cycle, lid, rundir):
     expect(os.path.isfile(da_script), "Data Assimilation script {} not found".format(da_script))
     filename = "da.log.{}".format(lid)
     outfile = os.path.join(rundir, filename)
-    run_sub_or_cmd(da_script, [caseroot, cycle], os.path.basename(da_script), [caseroot, cycle], logfile=outfile)
+    run_sub_or_cmd(da_script, [caseroot, cycle], os.path.basename(da_script), [caseroot, cycle], logfile=outfile) # For sub, use case?
 
 ###############################################################################
 def case_run(self, skip_pnl=False, set_continue_run=False, submit_resubmits=False):
--- a/scripts/lib/CIME/case/check_lockedfiles.py
+++ b/scripts/lib/CIME/case/check_lockedfiles.py
@@ -53,16 +53,16 @@ def check_lockedfile(self, filebase):
         objname = filebase.split('.')[0]
         if objname == "env_build":
             f1obj = self.get_env('build')
-            f2obj = EnvBuild(caseroot, lfile)
+            f2obj = EnvBuild(caseroot, lfile, read_only=True)
         elif objname == "env_mach_pes":
             f1obj = self.get_env('mach_pes')
-            f2obj = EnvMachPes(caseroot, lfile, components=components)
+            f2obj = EnvMachPes(caseroot, lfile, components=components, read_only=True)
         elif objname == "env_case":
             f1obj = self.get_env('case')
-            f2obj = EnvCase(caseroot, lfile)
+            f2obj = EnvCase(caseroot, lfile, read_only=True)
         elif objname == "env_batch":
             f1obj = self.get_env('batch')
-            f2obj = EnvBatch(caseroot, lfile)
+            f2obj = EnvBatch(caseroot, lfile, read_only=True)
         else:
             logging.warning("Locked XML file '{}' is not current being handled".format(filebase))
             return
@@ -92,7 +92,7 @@ def check_lockedfile(self, filebase):
                                          "case.build --clean-all and rebuilding")
                     else:
                         self.set_value("BUILD_STATUS", 1)
-                    f2obj.set_value("BUILD_COMPLETE", False)
+
             elif objname == "env_batch":
                 expect(False, "Batch configuration has changed, please run case.setup --reset")
             else:
--- a/scripts/lib/CIME/XML/env_archive.py
+++ b/scripts/lib/CIME/XML/env_archive.py
@@ -9,12 +9,12 @@ logger = logging.getLogger(__name__)
 # pylint: disable=super-init-not-called
 class EnvArchive(ArchiveBase,EnvBase):
 
-    def __init__(self, case_root=None, infile="env_archive.xml"):
+    def __init__(self, case_root=None, infile="env_archive.xml", read_only=False):
         """
         initialize an object interface to file env_archive.xml in the case directory
         """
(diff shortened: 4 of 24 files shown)


scripts/lib/CIME/case/case_run.py
```diff
@@ -248,7 +248,7 @@ def _do_external(script_name, caseroot, rundir, lid, prefix):
     filename = "{}.external.log.{}".format(prefix, lid)
     outfile = os.path.join(rundir, filename)
     append_status("Starting script {}".format(script_name), "CaseStatus")
-    run_sub_or_cmd(script_name, [caseroot], (os.path.basename(script_name).split('.',1))[0], [caseroot], logfile=outfile)
+    run_sub_or_cmd(script_name, [caseroot], (os.path.basename(script_name).split('.',1))[0], [caseroot], logfile=outfile) # For sub, use case?
```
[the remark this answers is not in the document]
jedwards4b: wouldn't it just call them as externals until the interface was updated?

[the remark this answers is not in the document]
jgfouca: You mean as shell commands? That would depend on if they are python programs or not.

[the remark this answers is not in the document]
jgfouca: Pretty sure. Different number of arguments would be an AttributeError. But, since python is dynamically typed, it will be happy to pass along an argument of a type that the receiving function is not expecting.

scripts/lib/CIME/case/check_lockedfiles.py
```diff
@@ -92,7 +92,7 @@ def check_lockedfile(self, filebase):
                                          "case.build --clean-all and rebuilding")
                     else:
                         self.set_value("BUILD_STATUS", 1)
-                    f2obj.set_value("BUILD_COMPLETE", False)
+
```
jgfouca: This is an example of an unflushed write error. We are still vulnerable to this kind of problem.

billsacks:
I see that you changed all of the cime buildlib files, removing 'with Case'. Will we also need to change active components similarly? In CESM, I see:

```
[roo2:~/cesm_code/cesm/components]$ rgrep 'with Case' .
./pop/cime_config/buildnml:    with Case(caseroot) as case:
./pop/cime_config/phys_cycle_prerun:    with Case(caseroot) as case:
./pop/cime_config/phys_cycle_postrun:    with Case(caseroot) as case:
./pop/cime_config/buildcpp:    with Case(caseroot) as case:
./pop/cime_config/buildlib:    with Case(caseroot) as case:
./ww3/cime_config/buildnml:    with Case(caseroot) as case:
./cice/cime_config/buildnml:    with Case(caseroot) as case:
./cice/cime_config/buildcpp:    with Case(caseroot) as case:
./cice/cime_config/buildlib:    with Case(caseroot) as case:
./cam/cime_config/buildnml:    with Case(caseroot) as case:
./cam/cime_config/buildcpp:    with Case(caseroot) as case:
./cam/cime_config/buildlib:    with Case(caseroot) as case:
./mosart/cime_config/buildnml:    with Case(caseroot) as case:
./mosart/cime_config/buildlib:    with Case(caseroot) as case:
./rtm/cime_config/buildnml:    with Case(caseroot) as case:
./rtm/cime_config/buildlib:    with Case(caseroot) as case:
./clm/cime_config/buildnml:    with Case(caseroot) as case:
./clm/cime_config/buildlib:    with Case(caseroot) as case:
```

Should all of these be changed? If so, is that a requirement for compatibility with this PR, or just something that we could do whenever we get around to it?

scripts/lib/CIME/case/case_run.py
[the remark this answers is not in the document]
billsacks: Sorry, just looking at this now.

Can you explain the purpose of this comment? Is this a reminder of something that needs to be changed in the future? If so, adding a little more to this comment so it's understandable by others would help... also, consider adding a TODO in the comment to call it out as such.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -71,17 +76,25 @@ def __init__(self, infile=None, schema=None, root_name_override=None, root_attri
 
             self.tree = ET.ElementTree(root)
 
+            self._FILEMAP[infile] = (self.tree, self.root, 0.0)
```
billsacks: Using something like a namedtuple for the elements of _FILEMAP would be more clear here: then each of the 3 elements would have a logical name rather than just being accessible via indices 0, 1 and 2 and requiring the reader to know (for example) that element 2 is the modification time.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -91,7 +104,7 @@ def read(self, infile, schema=None):
 
             logger.debug("File version is {}".format(str(self.get_version())))
 
-            self._FILEMAP[infile] = (self.tree, self.root)
+            self._FILEMAP[infile] = (self.tree, self.root, os.path.getmtime(infile))
```
billsacks: It seems like there's a possible race condition due to the timing of when you call getmtime to set the time of the cached data. Consider this scenario:

- Starting point: the mtime of the given file is 100.

- The current process needs to read from file. It opens the file and reads it.

- Meanwhile, another process writes to this file, resulting in an updated mtime (say, 110).

- Then the current process calls getmtime, getting the value 110. At this point, the current process would think it has data from mtime=110, whereas in fact it just has the data that were current as of mtime=100.

- Later, another read is done (without any more writes happening). It would use the cached data (because it incorrectly thinks it has the data from mtime=110), whereas in fact it should reread the file.

I think that moving the call to getmtime to _before_ the file_open would fix this problem. You could have the opposite problem – the current process thinks the data are out-of-date when in fact they're up-to-date – but that problem is less severe (just leading to a loss of efficiency, not incorrect data).

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -289,6 +302,12 @@ def write(self, outfile=None, force_write=False):
         """
         Write an xml file from data in self
         """
+        timestamp_cache = self._FILEMAP[self.filename][2]
+        if timestamp_cache != 0.0:
+            timestamp_file  = os.path.getmtime(self.filename)
+            expect(timestamp_file == timestamp_cache,
```
billsacks: I haven't thought about how, if at all, this error-check interacts with my suggestions about changing where mtime is called.

scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -310,6 +329,9 @@ def write(self, outfile=None, force_write=False):
         else:
             with open(outfile,'w') as xmlout:
                 xmlout.write(xmlstr)
+
+        self._FILEMAP[self.filename] = (self.tree, self.root, os.path.getmtime(self.filename))
```
billsacks: In theory, it seems like this could be vulnerable to the same race condition described above, but I don't know how to solve this one... and hopefully the system would prevent two processes from trying to write to the same file at the same time anyway, so the risk here seems small enough.

jgfouca:
> I see that you changed all of the cime buildlib files, removing 'with Case'. Will we also need to change active components similarly?

Yes, any buildlib that is implemented in python will need to change.

scripts/lib/CIME/case/case_run.py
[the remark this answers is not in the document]
jgfouca: I'm not sure we want to do it or not, just something to think about. It could be hard to track down all the scripts users are using for pre/post processing.

scripts/lib/CIME/XML/generic_xml.py
jgfouca: Will do.

scripts/lib/CIME/XML/generic_xml.py
jgfouca: @billsacks you are exactly right and this is the point I was trying to make in the discussion ticket that we will never achieve 100% robustness unless we access entry values through some kind of database system.

The reason my approach works (I think) is that I've made the simplifying assumption that CIME will never be used in the way you described. We've already discussed that the bulk of CIME is not thread safe, there's far too much shared state. Due to the situation with the XML files, it is also not safe to run multiple CIME processes in the same case concurrently. Note that all the CIME subprocess calls are synchronous calls (the parent spawns the child and then waits for it to finish before doing anything else), so no true concurrency is happening. The subprocesses can modify XML and, as long as the parent refreshes the cache when the subprocess is done, everything is OK.

jedwards4b:
This discussion should factor into the recurring discussion of changing data file formats - would it be better to implement something more like an sql database for our data file interface?

billsacks:
> Yes, any buildlib that is implemented in python will need to change.

So what will happen until that change is made? Will the build fail, for example?

billsacks:
And: I forget what our latest thinking is on versioning (are we trying to use some flavor of semantic versioning?), but it seems like this change would then warrant some big bump in version number to indicate backwards-incompatibility.

scripts/lib/CIME/XML/generic_xml.py
billsacks: Okay, fair enough - thanks for the explanation.

jgfouca:
> This discussion should factor into the recurring discussion of changing data file formats - would it be better to implement something more like an sql database for our data file interface?

Yes, it will crash with a python type error.

jgfouca:
> This discussion should factor into the recurring discussion of changing data file formats - would it be better to implement something more like an sql database for our data file interface?

That's a good question, the answer will depend on the direction that CIME goes in the future. I think the best possible outcome is that all the components' CIME-related scripts (buildnml, buildlib) are in Python and so we always stay within a single process. Most of CIME is inherently sequential, each phase cannot start until the previous phase is done, so thread concurrency is most not needed. The one exception is component builds, which happen in parallel. We will have to set some constraints on what a component buildlib script can do with its Case object; for example, by passing them a read-only Case object.

If we can achieve all this, it should be fine to stick with the current system which is must simpler than using a "real" database like SQL.

billsacks:
> Yes, it will crash with a python type error.

Thanks. Given this, I'd ask:

1. I see that you updated the top-level comment to point this out a bit more, but can you please be even more explicit about this - i.e., basically giving a recipe for what components need to do in order to (a) find the places that need to be changed and (b) change them appropriately?

2. We should bump the version number when this is brought to master to give some indication of backwards incompatibility.

billsacks:
@jgfouca I'm still trying to figure out what, if anything, we need to change in the active components in CESM. Looking more closely at their buildlib scripts, I see that none of them actually define a buildlib function. Does that imply that we're fine for now? (I'm wondering if this means that all of the active components' buildlibs are being called as external programs rather than as python functions?)

jedwards4b:
Yes - that is what it means.

PR #2938, opened 2018-12-06
do not allow trailing / in filename

### LINKED ISSUE #2937: create_newcase and trailing '/' on case name
Hi all,

  This came up during some testing the other day - apparently creating a case with a trailing slash works fine (during create_newcase), but running a ./case.build on the created case will fail with the error:

ERROR: No default value found for case_name.

  Example:
create_newcase --case /home/foo/testcase_fails/ --compset B1850
...
./case.setup
...
./case.build
...
ERROR: No default value found for case_name._

  Is it worth adding a check to create_newcase for trailing slashes? 

  Thanks,
  - Brian

jedwards4b:
Disallow trailing / character in filename
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #2937 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/utils.py
+++ b/scripts/lib/CIME/utils.py
@@ -151,11 +151,15 @@ def check_name(fullname, additional_chars=None, fullpath=False):
     False
     >>> check_name("mycase*mods")
     False
+    >>> check_name("/some/full/path/name/")
+    False
     """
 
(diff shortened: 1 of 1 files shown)


jedwards4b:
@fischer-ncar I can't figure out what codacy is complaining about - I think this is ready to merge.

PR #2943, opened 2018-12-10
Update fv1.9x2.5,gx1v7 grids

fischer-ncar:
Older versions of the fv1.9x2.5,gx1v7 files were generated with ESMF 7.0.0, which had a bug.  The
new mappings files were generated with the ESMF7.1.0r, which fixed the bug.

Test suite: scripts_regression_tests.py,ERS_Ld5.f19_g17.BWma1850
Test baseline: 
Test namelist changes: 
Test status: answer changing for g19_g17 grids

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


billsacks:
There are other mapping files for fv1.9 besides those that have been replaced here - including mapping files to rof, wav and glc grids. Should those be regenerated, too?

jedwards4b:
@billsacks A pretty careful check of all of the mapping files was conducted after this problem was discovered, these are the only ones that were found to have issues.

billsacks:
Okay, thanks!

PR #2950, opened 2018-12-12
make sure optional arg exists before using

### LINKED ISSUE #2945: docn_comp_mod illegal use of optional argument
In docn_comp_mod subroutine docn_comp_run in both the nuopc and mct interfaces the variable case_name is declared as optional to the subroutine but is then used without checking for it's presence.   docn_comp_init calls docn_comp_run without this optional argument.

jedwards4b:
Need to make sure optional variable is provided before it is used.

Test suite: 
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #2945
User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/components/data_comps/docn/nuopc/docn_comp_mod.F90
+++ b/src/components/data_comps/docn/nuopc/docn_comp_mod.F90
@@ -595,7 +595,7 @@ contains
     integer           :: nu     ! unit number
     character(len=18) :: date_str
     character(len=CS) :: fldname
-
+    character(len=CS) :: local_case_name
     real(R8), parameter :: &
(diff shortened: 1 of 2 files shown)


src/components/data_comps/docn/nuopc/docn_comp_mod.F90
```diff
@@ -595,7 +595,7 @@ subroutine docn_comp_run(x2o, o2x, &
     integer           :: nu     ! unit number
     character(len=18) :: date_str
     character(len=CS) :: fldname
-
+    character(len=CS) :: local_case_name
```
billsacks: CS isn't long enough for a case name. I'd use at least CL (as you do for the mct version).

PR #2955, opened 2018-12-19
Jedwards/pio2 merge

jedwards4b:
Update to the latest pio2 master from https://github.com/NCAR/ParallelIO
pio hash is 317963f

Test suite: scripts_regression_tests.py with PIO_VERSION=2 in src/drivers/mct/cime_config/config_component.xml
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


scripts/tests/user_mods_test3/shell_commands
```diff
@@ -1 +1,7 @@
-./xmlchange PIO_VERSION=2
\ No newline at end of file
+pioversion=`./xmlquery --value PIO_VERSION`
```
jayeshkrishna: Is this script to make sure that we test both versions of PIO (if the default is PIO1 test PIO2 and vice versa)?

jayeshkrishna:
It would be useful to add the hash of PIO2 master being merged in the merge commit msg

scripts/tests/user_mods_test3/shell_commands
jedwards4b: No, currently we are not testing pio2 - this script is to test an aspect of create_clone - (that you cannot use the same executable if you change an xml variable that has a build dependency).

PR #2962, opened 2019-01-03
Add expected failure annotations to cs.status output

billsacks:
This PR adds expected failure annotations to the output of cs.status,
via adding expected failure functionality to
TestStatus.phase_statuses_dump.

To use this functionality: First, create an xml file listing the
expected failures for a given test suite (it's fine for this to include
tests that are not in the current test suite - so you can use a single
expected failures xml file for multiple different test suites). Then
point to this file using the new '-x'/'--expected-fails-file' argument
to the cs.status command. (I developed this primarily for use by
CTSM. CTSM now has a wrapper script to create_test that, among other
things, creates custom cs.status scripts in TESTROOT. So these custom
cs.status scripts can now automatically point to CTSM's expected failure
file, removing the need for people running the CTSM test suite to know
where to find this file.)

Here is sample output from a test suite where the only failures were
expected failures, run with --fails-only and the new
--expected-fails-file argument. Note that I'm showing expected failures
even with --fails-only, because I feel that it's worthwhile to be
constantly reminded of these expected failures, but I'm not tied to that
choice (i.e., I'd be open to having expected failures hidden if you run
with --fails-only if others feel that would be better):

```
1228-065617ch_gnu: 38 tests
    FAIL ERS_Lm20_Mmpi-serial.1x1_smallvilleIA.I2000Clm50BgcCropGs.cheyenne_gnu.clm-monthly RUN time=130 (EXPECTED FAILURE)


1228-065617ch_int: 162 tests
    FAIL ERP_D_Ld10_P36x2.f10_f10_musgs.IHistClm50BgcCrop.cheyenne_intel.clm-ciso_decStart RUN time=104 (EXPECTED FAILURE)
    PEND ERP_D_Ld10_P36x2.f10_f10_musgs.IHistClm50BgcCrop.cheyenne_intel.clm-ciso_decStart COMPARE_base_rest (EXPECTED FAILURE)
    FAIL ERS_D_Ln9_P480x3.f19_g16.I2000Clm50SpGs.cheyenne_intel.clm-waccmx_offline COMPARE_base_rest (EXPECTED FAILURE)
    FAIL PFS_Ld20.f09_g17.I2000Clm50BgcCrop.cheyenne_intel GENERATE exception (EXPECTED FAILURE)
```

If the PFS GENERATE started to pass, we would instead see:

```
    PASS PFS_Ld20.f09_g17.I2000Clm50BgcCrop.cheyenne_intel GENERATE exception (UNEXPECTED: expected FAIL)
```

This "UNEXPECTED" comment is printed for tests that appear in the
expected fail list but whose status doesn't match the expectation. (It
is NOT shown for failing tests that are absent from the expected fails
list.) The reason for showing unexpected results, including a PASS
result that was expected to fail, is that this usually indicates the
need to remove the test from the expected fails list.

Test suite: ./scripts_regression_tests on cheyenne,
  plus a bit of manual testing
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes none

User interface changes?: Adds '-x'/'--expected-fails-file' argument to
cs.status; also adds shorthand '-p' argument to cs.status, as a
shorthand for '--count-performance-fails'.

Code review:
I'd like reviews from @jgfouca @jedwards4b @ekluzek 
Also, @fischer-ncar I wanted to bring this to your attention, and you should feel free to review it if you want


scripts/lib/CIME/XML/expected_fails_file.py
```diff
@@ -0,0 +1,65 @@
...
+from CIME.expected_fails import ExpectedFails
+
+logger = logging.getLogger(__name__)
+
+class ExpectedFailsFile(GenericXML):
+
+    def __init__(self, infile, files=None):
+        if files is None:
+            files = Files()
+        # FIXME(wjs, 2019-01-02) Get schema, replacing the following line
```
billsacks: I still plan to add an xml schema definition. I'd like to get approval of my proposed schema, as per the example at the top of this file, before I write this xsd file.

billsacks:
While I welcome comments on any aspects of this PR, I would especially invite comments on:

(1) The format of the expected fails xml file, as shown by example at the top of the new file, `expected_fails_file.py`. I don't have much experience with designing xml schemas, so welcome thoughts on how this could be improved. Note that I imagine this file containing optional `<issue>` and `<comment>` elements; currently, the code doesn't do anything with these elements, but they are useful for human consumption of this file.

(2) The appearance of the output for expected / unexpected fails (as shown by example in the initial PR comment).

scripts/lib/CIME/XML/expected_fails_file.py
billsacks: (This is why I have labeled this PR as WIP.)

billsacks:
See https://github.com/ESCOMP/ctsm/pull/602 for how I propose to incorporate this in CTSM - including both the new expected fails xml file and how I automatically point to this file in the custom cs.status scripts created by CTSM's run_sys_tests command.

billsacks:
@jedwards4b sorry, I'm having a bit of trouble understanding what you're envisioning. First I want to point out, in case it wasn't apparent, that the `get_expected_fails` function returns a dictionary of ExpectedFails objects (as defined in scripts/lib/CIME/expected_fails.py). So the values of the returned dictionary are meant to be treated as opaque objects, and I envisioned that additional functionality could be added to the ExpectedFails class in the future without requiring any changes in client code.

But maybe that was already apparent and you were still envisioning something more/different? If so, could you please either explain what you're thinking in a bit more detail, and/or point me to an example of existing code that illustrates what you're thinking?

jedwards4b:
I guess I don't see why there are two seperate files - could all of this functionality be in 
scripts/lib/CIME/XML/expected_fails_file.py and scripts/lib/CIME/expected_fails.py be eliminated?

billsacks:
@jedwards4b I'll explain my rationale, but I'm open to changing the design if you think this doesn't make sense:

`TestStatus.phase_statuses_dump` requires information about a single test's expected failures. This motivated the creation of the ExpectedFails class, which holds information about a single test's expected failures. This function (and the entire TestStatus class) doesn't need to know whether this expected failure information came from an xml file or from somewhere else (e.g., for the unit tests, an ExpectedFails object is created directly in code), which was part of the motivation of having a separate ExpectedFails class rather than having `phase_statuses_dump` depend on the XML-related ExpectedFailsFile class.

I thought about putting the definition of the ExpectedFails class in the same file as ExpectedFailsFile, because the two are closely related. But I tend to default to putting each class in its own file, because I find it easier to find what I'm looking for that way.

I'm definitely open to removing expected_fails.py and just moving the definition of ExpectedFails into expected_fails_file.py if you think that would be better. I'm more hesitant about removing the ExpectedFails class entirely (having `phase_statuses_dump` accept an argument of type ExpectedFailsFile), partly because I think the design and dependencies are messier that way, and partly because it will complicate the unit testing of `phase_statuses_dump`... but I'm open to reconsidering that if you feel it would be much preferable.

jedwards4b:
what you've done is fine, I just have a general nit about exposing data structures like dicts that may make it hard to maintain.   In this case it appears you have considered the consequences and limited the exposure.

billsacks:
Okay, thanks. I do agree with your general nit and appreciate your thoughts.

billsacks:
@jedwards4b I just added an xsd file for the expected fails xml file, and have tested it with both a unit test and some manual tests. I'd welcome a critical eye on this, since I've never written an xsd file before. (I mostly used testlist.xsd as an example.)

Other than that:
- I'm kicking off scripts_regression_tests on cheyenne now
- I'll ping @ekluzek to add his review, if he wants to

fischer-ncar:
Thanks for doing this @billsacks.  I plan on created an ExpectedTestFails.xml for the allactive tests.
Also, it looks good.

config/xml_schemas/expected_fails_file.xsd
```diff
@@ -0,0 +1,42 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
jedwards4b: I don't see any problems here.

gold2718:
Is there a sample ExpectedTestFails.xml file? It is hard for me to evaluate an XSD file without knowing intended usage.

scripts/lib/CIME/XML/expected_fails_file.py
```diff
@@ -0,0 +1,76 @@
+"""Interface to an expected failure xml file
+
+Here is an example:
+
+<?xml version= "1.0"?>
```
jedwards4b: @gold2718

config/xml_schemas/expected_fails_file.xsd
```diff
@@ -0,0 +1,42 @@
...
+        <xs:element minOccurs="0" maxOccurs="unbounded" ref="test"/>
+      </xs:sequence>
+      <xs:attribute name="version" use="required" type="xs:decimal"/>
+    </xs:complexType>
+  </xs:element>
+
+  <xs:element name="test">
+    <xs:complexType>
+      <xs:sequence>
+        <xs:element minOccurs="0" maxOccurs="1" ref="category"/>
```
gold2718: What is the point of this element? Can a test fail in one category but pass in a different category?

config/xml_schemas/expected_fails_file.xsd
```diff
@@ -0,0 +1,42 @@
...
+        <xs:element minOccurs="1" maxOccurs="unbounded" ref="phase"/>
+      </xs:sequence>
+      <xs:attribute name="name" use="required" type="xs:NCName"/>
+    </xs:complexType>
+  </xs:element>
+
+  <xs:element name="phase">
```
gold2718: I would normally define an element (such as phase below) before I use it (as in test above) but this is purely a style issue.

config/xml_schemas/expected_fails_file.xsd
billsacks: @gold2718 - @ekluzek requested that I add this in his review of the companion PR in CTSM (where I changed CTSM's expected fails xml file to this new format): https://github.com/ESCOMP/ctsm/pull/602. My comment from that PR was similar to yours:

> Note that different test suites are now delineated by comments, not by xml sections. This is because, in the machine parsing of this file, we don't have an easy way to determine which category is appropriate, and in any case, it's not really important: if a test from one category is supposed to fail, then it will fail regardless of how we're running the test – including if we're running individual tests, not part of any test suite at all.

@ekluzek do you want to try to justify your thoughts here? Or, based on @gold2718 's query, do you feel okay dropping this element?

config/xml_schemas/expected_fails_file.xsd
billsacks: That makes sense, and I like the idea of that consistency. I have just pushed a commit where I made that change.

billsacks:
I ran scripts_regression_tests on cheyenne and all tests pass. This is now just waiting on approval from @ekluzek , as well as his reply to @gold2718 's query about the `category` element.

ekluzek:
@gold2718 <category> is one that I asked Bill to add. It refers to what test-list "-xml-category" was used to run the test. Since, most of our testing is run this way, I think it's useful to have. In our prototypes of expected fails it's something that we divided up our lists in terms of. We have a pretty extensive "fates" test list that's separate from our standard "aux_clm" test list. We also have a "clm_short" test list. And in the past we had a different test list for clm40 than clm45/50. And of course you can note that we have prealpha and prebeta test lists in addition to the standard ones for components. So I think it's helpful to know which test list this is part of. And I thought it should be part of the XML, rather than just a comment, since comments can often be wrong.

This is still optional additional information that's not required. The test isn't going to act differently if it's run from one xml-category or another. It'll FAIL in any test list it's run in.

jedwards4b:
@ekluzek I'm not sure that this makes any sense, presumably a test that appears in multiple categories will fail in all of them - which one would you list here and why?   If you want to know what category a test is in you use the query_testlists script.

gold2718:
@ekluzek,

My concern is that many tests are included in more than one test list. For the `category` element to be useful, you would need to have an entry for every such test list. Duplicating expected test fail information sounds like a maintenance nightmare but maybe I do not understand the expected use cases. My instinct would be to have one expected fails file per model which could be included in config_files (e.g., something like `  <entry id="EXPECTED_FAILS_FILE">`) but since it appears that you need to manually specify a filename to `cs.status` (is that right?), maybe you plan to use this differently.
@billsacks, how do you envision these files being managed by a model-development team?

ekluzek:
Hmmm. @gold2718 and @jedwards4b I see your point. I think the problem I'm really trying to solve resolves around the test list management. I find query_test to be inadequate to manage the test lists. What I think is useful to see is how many expected fails are from a specific test list. Then I can start thinking about if maybe the thing that needs to be done is to change the test list. So I guess I'm OK with Bill going back to putting this in comments rather than XML. So sorry about that Bill. Although now I wonder about having query_test to be able to read the expected fails list as an option. I'm also not quite sure how query_test needs to improve. But, that would be a separate issue.

billsacks:
Okay, removed `category`.

@gold2718 responding to your question: I also envision a single expected test fails file per component. For CTSM, we now have a wrapper to create_test that does some other stuff to make running and later parsing the test suite easier for our (or at least my) workflow; one thing this now does is to add the `--expected-fails-file` argument to a custom cs.status script (I was creating a custom cs.status script for other reasons anyway). I didn't think of having this file picked up automatically via config_files, but that could make sense.

billsacks:
I think the config_files piece will take some thought, because it isn't clear (to me, anyway) how you will decide which component's expected fails file to look in. (Note that a given test suite can aggregate the test lists from multiple components.) So I'd request that the current PR be merged without that, and then others can give that some thought if they'd like to.

gold2718:
I am fine with going ahead with the merge.

While, in general, I would expect expected fail entries to be in the model where the test is located,
I can see some complication with where to put expected fail test files. For instance, if I am working on a new CAM feature which will cause some tests to break until a corresponding change is made in CLM where is the best place to record that? If I put it in the CAM test directory, it will not automatically get removed when the CLM 'fix' goes in. Putting it in CLM is weird because the test as well as the change that broke it are in CAM. So I agree with @billsacks that we should merge now and do the hard thinking later since this feature is an improvement even with the current (low) level of integration.

Issue #2963, opened 2019-01-03
Remove support for building clm4.0


billsacks:
In CESM, we're removing CLM4.0. This allows cleaning up some of the build logic in cime, since the clm4.0 library is handled differently from the modern clm library (the latter is done using a sharedlibbuild). I'll do this cleanup.

@rljacob @jgfouca @bishtgautam  I think the amount of cleanup I can do depends on whether you still use clm4.0 in e3sm. Can you please let me know if you do? I'm especially thinking of this block in the Makefile:

https://github.com/ESMCI/cime/blob/37d7afbe9e12bf8d281af7283756720faedb2aa4/scripts/Tools/Makefile#L816-L834

For cesm, we'll now _always_ have `USE_SHARED_CLM=TRUE`, so the `ifeq ($(USE_SHARED_CLM),FALSE)` block will just apply to e3sm. What I'm wondering is if you still need this block:

```Makefile
  ifeq ($(CLMVER),$(null))
    LNDLIB := liblnd.a
```

or if you can always use the clm4_5/clm5_0 version,

```Makefile
  else
    LNDLIB := libclm.a
```

It would really clean things up if I can get rid of the use of `CLM_CONFIG_OPTS` entirely, but I'll keep it in place if need be.

rljacob:
We have removed clm40.  @jgfouca had some other reason why we couldn't build the shared clm lib.

jedwards4b:
If I remember correctly it was because the e3sm version of clm still has build time dependancies while these have all been removed from ctsm.

billsacks:
Thanks, @rljacob . I'll move ahead with the more aggressive removal of all things clm4.0-related.

billsacks:
Hmmm, before I move ahead with this, I want to double check: I do see this on e3sm master:

```python
    if "clm4_0" in clm_config_opts:
        complib = os.path.join(libroot,"liblnd.a")
```

so I want to confirm: is that still present but unused?

billsacks:
Ah, it looks like clm4.0 source code has been removed in E3SM, so I guess the remnants of references to clm4_0 there truly aren't exercised.

PR #2968, opened 2019-01-09
Remove support for CLM4.0

billsacks:
CLM4.0 is being dropped in CESM, and has already been dropped in
E3SM. This commit removes some of the complexity related to supporting
that old version of CLM.

This commit completely removes support for CLM4.0 for CESM. For E3SM,
support is removed in files that are shared between CESM and E3SM, but I
have not made any changes to E3SM-specific files.

A key change was in the build: we no longer need to distinguish between
CLM4.0 vs. later versions of CLM in setting some build options (but I
think that we *do* still need to determine whether we're running with
CLM or some other land model, including dlnd/slnd/xlnd, so some
CLM-specific logic still remains in the Makefile).

This change is backwards-incompatible with respect to CLM4.0 support,
but should be backwards compatible for cases using later versions of
CLM, as well as cases using other land components.

Test suite: `scripts_regression_tests` on cheyenne
Also ran most of the `aux_clm` test suite in the context of
https://github.com/ESCOMP/ctsm/pull/609 (note that that used a version
of this cime branch rebased onto cime5.7.5)
Test baseline: For `aux_clm` test suite, used previous ctsm tag
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#2963

User interface changes?: Drops support for CLM4.0

Code review:

### DIFF
--- a/config/cesm/machines/config_pio.xml
+++ b/config/cesm/machines/config_pio.xml
@@ -109,13 +109,6 @@
     </values>
   </entry>
   -->
-<!--  <entry id="LND_PIO_REARRANGER">
-    <values>
-      <value compset="_CLM40" >2</value>
-      <value compset="_CLM45" >1</value>
-      <value compset="_CLM50" >1</value>
-    </values>
-  </entry> -->
 
   <!--- uncomment and fill in relevant sections
   <entry id="LND_PIO_STRIDE">
--- a/scripts/Tools/Makefile
+++ b/scripts/Tools/Makefile
@@ -813,9 +813,8 @@ test_esmf: test_esmf.o
 #-------------------------------------------------------------------------------
 # create list of component libraries - hard-wired for current ccsm components
 #-------------------------------------------------------------------------------
-CLMVER = $(filter $(CLM_CONFIG_OPTS), clm5_0 clm4_5)
 ifeq ($(CIME_MODEL),cesm)
-  ifneq ($(CLMVER),$(null))
+  ifeq ($(COMP_LND),clm)
     USE_SHARED_CLM=TRUE
   else
     USE_SHARED_CLM=FALSE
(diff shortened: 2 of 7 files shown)


billsacks:
@jgfouca Most of these changes only apply to CESM. However, the Makefile change potentially impacts E3SM as well. I'm pretty sure that I've done this correctly to avoid breaking E3SM, but I'm assigning you as a reviewer in case you'd like to double-check that. As per the discussion in #2963 , this drops support for building CLM4.0 in E3SM as well as CESM. (It appears that CLM uses a different library name for CLM4.0 vs. later versions of the code, which is what's captured in the Makefile logic.)

scripts/Tools/Makefile
```diff
@@ -813,9 +813,8 @@ test_esmf: test_esmf.o
 #-------------------------------------------------------------------------------
 # create list of component libraries - hard-wired for current ccsm components
 #-------------------------------------------------------------------------------
-CLMVER = $(filter $(CLM_CONFIG_OPTS), clm5_0 clm4_5)
 ifeq ($(CIME_MODEL),cesm)
-  ifneq ($(CLMVER),$(null))
+  ifeq ($(COMP_LND),clm)
```
billsacks: I *think* we still need some logic in the makefile so that the correct thing is done for dlnd/slnd/xlnd. @jedwards - Does this change to the Makefile, together with the one below, look correct? I have tested cases both with CLM and with DLND in CESM and confirmed that it looks like the behavior is the same as before in both cases, but I'd be happy to get a critical set of eyes on the changes in this file.

config/cesm/machines/config_pio.xml
```diff
@@ -109,13 +109,6 @@
     </values>
   </entry>
   -->
-<!--  <entry id="LND_PIO_REARRANGER">
```
billsacks: I have no idea what this commented-out block of xml code was for, but I thought I'd go ahead and just remove it rather than trying to figure it out. @jedwards4b (or @ekluzek ) feel free to chime in if you feel this is important to maintain. (For what it's worth, I see this block uncommented-out on the e3sm side.)

ekluzek: I'm pretty sure this was something that was required at one point with PIO2. Since, we no longer use PIO2 with CLM it doesn't come into play. But, may if we even do start using it, unless PIO2 now has a default.

PR #2970, opened 2019-01-09
docn_comp_mod needs fix for optional variable

jedwards4b:
A couple of fixes for cesm testing.   

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case/case_clone.py
+++ b/scripts/lib/CIME/case/case_clone.py
@@ -10,7 +10,7 @@ from CIME.simple_compare            import compare_files
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


scripts/lib/CIME/case/case_clone.py
```diff
@@ -10,7 +10,7 @@
 
 logger = logging.getLogger(__name__)
 
-def create_clone(self, newcase, keepexe=False, mach_dir=None, project=None,
+def create_clone(self, newcaseroot, keepexe=False, mach_dir=None, project=None,
```
jgfouca: Thanks, this confused me also.

PR #2972, opened 2019-01-09
Fix Case context manager enforcement

jgfouca:
This is a significant change and needs a close review. The code will be easier to examine if whitespace changes are ignored.

A bug slipped-in some time ago that allowed cases to be
modified outside a context manager. This was allowing errors
to slip in, such as unflushed writes in SystemTestCompareTwo.

This PR also adds tests to ensure this doesn't happen again.

Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -173,33 +173,37 @@ class SystemTestsCompareTwo(SystemTestsCommon):
     # ========================================================================
 
     def build_phase(self, sharedlib_only=False, model_only=False):
-        if self._separate_builds:
-            self._activate_case1()
-            self.build_indv(sharedlib_only=sharedlib_only, model_only=model_only)
-            self._activate_case2()
-            # Although we're doing separate builds, it still makes sense
-            # to share the sharedlibroot area with case1 so we can reuse
-            # pieces of the build from there.
-            if get_model() != "e3sm":
-                # We need to turn off this change for E3SM because it breaks
-                # the MPAS build system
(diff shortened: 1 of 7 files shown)


jgfouca:
Testing is ongoing.

jgfouca:
I've confirmed the ERP PIO_TYPENAME bug is also fixed in this PR.

scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -173,33 +173,37 @@ def _case_two_custom_postrun_action(self):
...
-
-            # The following is needed when _case_two_setup has a case_setup call
-            # despite sharing the build (e.g., to change NTHRDS)
-            self._case2.set_value("BUILD_COMPLETE",True)
-            self._case2.flush()
+        # Subtle issue: case1 is already in a writeable state since it tends to be opened
+        # with a with statement in all the API entrances in CIME. case2 was created via clone,
+        # not a with statement, so it's not in a writeable state, so we need to use a with
+        # statement here to put it in a writeable state.
+        with self._case2:
```
billsacks: I would have a slight preference to moving this `with` statement down, so that it just surrounds the code that actually needs it, even if that means we need two or three separate `with` blocks in this function. For example, I'd think that the `build_indv` call on case1 doesn't need to be within a `with self._case2` block. For those of us who don't fully understand where this `with` statement is needed, I think that would help us in reading / modifying the code later - i.e., if only the only things in `with` blocks are things that _need_ to be in `with` blocks.

But I don't feel strongly about this, so if you prefer it out at this level, that's okay with me.

PR #2983, opened 2019-01-16
Better handling of file permissions when copying files

### LINKED ISSUE #2982: Baseline namelists for failed tests are not world-readable
For most tests, the generated namelists (i.e., in the baseline directory) have permissions `-rw-rw-r--` (at least on cheyenne). However, for at least some tests that fail, the generated namelists have permissions `-rw-rw----`. This leads to incorrect NLCOMP failures when someone tries to do baseline comparisons but isn't in the appropriate group.

This is an issue for tests that are expected to fail some later phase (say, the RUN phase), but for which namelists should still have been generated / compared. It is also an issue for the FUNIT test.

jgfouca:
Technique is to make better-use of SharedArea context manager
and copying files such that the user's umask will determine
the permissions of the copy, not the original file's meta data.

distutils.file_util|dir_util allows for more control of file
perms than shutil, so use that instead.

Add preserve_metadata flag to our safe_copy.

Include special exe handling in our safe_copy.

Copies occurring in SharedAreas should almost always have
preserve_metadata set to False.

Add a test to confirm all the above.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2982 

User interface changes?: N

### DIFF
--- a/scripts/tests/scripts_regression_tests.py
+++ b/scripts/tests/scripts_regression_tests.py
@@ -18,6 +18,7 @@ import subprocess, argparse
 subprocess.call('/bin/rm -f $(find . -name "*.pyc")', shell=True, cwd=LIB_DIR)
 import six
 from six import assertRaisesRegex
+import stat as osstat
 
 import collections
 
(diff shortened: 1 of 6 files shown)


jgfouca:
@jedwards4b , @billsacks , permissions issues have dogged CIME for a long time but I think, with this PR, the situation will be much-improved. One thing, I considered adding a copy routine to SharedArea for users to use within a SharedArea context; this would mean they wouldn't have to remember to add preverve_metadata=False to all their copies.

scripts/tests/scripts_regression_tests.py
```diff
@@ -2336,7 +2336,7 @@ def simple_test(self, manual_timing=False):
             provenance_dirs = glob.glob(os.path.join(timing_dir, "performance_archive", getpass.getuser(), casename, lids[0] + "*"))
             self.assertEqual(len(provenance_dirs), 1, msg="provenance dirs were missing")
 
-        self.verify_perms(self, timing_dir)
+        verify_perms(self, timing_dir)
```
jedwards4b: This was to avoid a pylint error?  Another way to do it is with the 
@staticmethod

jgfouca: No, multiple test classes (both BlessTests and SaveTimings) needed this capability, so I made it a free function.

PR #2985, opened 2019-01-18
Add a test for tools/configure

### LINKED ISSUE #2911: Need test for configure tool
The configure tool (scripts/tools/configure) has been broken for a while and we didn't notice it. I've fixed it in E3SM. After the next split, we should add a test for it so we catch problems in the future.

jgfouca:
Test suite: the test that was added
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2911 

User interface changes?: N

Update gh-pages html (Y/N)?:

Code review: @jedwards4b

### DIFF
--- a/scripts/tests/scripts_regression_tests.py
+++ b/scripts/tests/scripts_regression_tests.py
@@ -21,7 +21,7 @@ from six import assertRaisesRegex
 
 import collections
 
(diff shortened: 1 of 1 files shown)


ekluzek:
Yay. Really glad this is being done. This is important for tools that rely on configure. So this helps us in CTSM world. Thanks for doing this.

Issue #2994, opened 2019-02-08
cheyenne batch directives incorrect when using an unknown queue


billsacks:
This came up in the CTSM tutorial this week, in which participants were directed to use a special queue on cheyenne. I imagine it could impact other, future tutorials, as well as other use cases where users use some special queue for their work.

When you set a cheyenne queue to something unknown to cime, using something like:

```
./xmlchange JOB_QUEUE=R4231261 --subgroup case.run --force
```

you get batch directives that are appropriate for a single-node run rather than a multi-node run.

Here is an example: For `./create_newcase --case $scratch/test_queue_0208a --compset A --res f45_g37 --run-unsupported` followed by `./xmlchange NTASKS=144`: with the default queue, I have the following in `.case.run`:

```
 #PBS  -l select=4:ncpus=36:mpiprocs=36:ompthreads=1
```

But if I use the above `./xmlchange JOB_QUEUE` command, I instead get:

```
 #PBS  -l select=1:mpiprocs=144:ompthreads=1
```

which seems to put the whole job on a single node, leading the job to hang.

A couple of possible solutions come to mind, but I'm not sure which is the right one to do here:

(1) In CESM's config_batch.xml, for cheyenne, have a default directives block that applies to any queue not explicitly specified:

```xml
  <batch_system MACH="cheyenne" type="pbs">
    <directives>
      <directive default="/bin/bash" > -S {{ shell }}  </directive>
      <directive> -l select={{ num_nodes }}:ncpus={{ max_tasks_per_node }}:mpiprocs={{ tasks_per_node }}:ompthreads={{ thread_count }}</directive>
    </directives>
```

(Would this let us remove the duplicate directives blocks that are currently specified for regular, premium and economy, or would those still be needed?)

(2) It seems like there is some logic that picks up the last queue specified in the directives blocks for an unknown queue; if that were either changed to pick up the first, or if we reorder those blocks for cheyenne, then maybe this would work correctly. (Not yet tested.)

(This seems somewhat related to issue #2551 , which @jedwards4b resolved last May.)

jedwards4b:
I would prefer option (1) - do you know if it works in the current cime?

billsacks:
@jedwards4b - from a bit of playing around, it looks like (1) does _not_ currently work.

I'll give more details, but first I want to correct what I suggested for (2) in my original comment: it turns out that an unknown queue is getting values from the share queue because the share queue is listed as the default (`default="true"`) here:

[...]

PR #2995, opened 2019-02-11
Allow specifying directives to use for an unknown queue

billsacks:
Previously, when using an unknown queue, the batch directives would come
from the default queue for the machine. This is a problem on cheyenne,
for which the default queue is the share queue. It looks like we need to
keep the share queue as the default queue for other purposes, though (in
particular: if the default is the regular queue, then the share queue
won't get selected for small - e.g., single-processor - jobs, since
small jobs also satisfy the requirements of the regular queue), so the
solution is to provide a different mechanism for specifying the queue to
use for the sake of choosing directives.

For machines that don't use this new mechanism, we maintain the previous
behavior of using the directives for the machine's default queue.

Note: I chose this mechanism of pointing to a different queue's
directives section rather than explicitly listing the directives for an
unknown queue, partly because this was easy to implement and partly
because I thought the latter would likely lead to the two getting
accidentally out-of-sync (because people may not think to update the
directives for an unknown queue, and this would likely go untested and
undiscovered for a long time).

Test suite: scripts_regression_tests on cheyenne;
also manual testing on a version off of master,
and on this version off of maint-5.6 (all tests were
with compset A, res f45_g37 on cheyenne, with NTASKS changed to 144):
- Checked .case.run with regular queue: same as before
- Checked .case.run with share queue: same as before
- Checked .case.run with an unknown queue: now same as regular queue
- Checked .case.run with an unknown queue, but with the new
  <unknown_queue_directives> block temporarily removed: same as share
  queue, as before (demonstrating that this change is backwards
  compatible)

Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#2994

User interface changes?: no

Code review:

### DIFF
--- a/config/cesm/machines/config_batch.xml
+++ b/config/cesm/machines/config_batch.xml
@@ -201,6 +201,10 @@
       <directive default="/bin/bash" > -S {{ shell }}  </directive>
       <directive> -l select=1:mpiprocs={{ total_tasks }}:ompthreads={{ thread_count }}</directive>
     </directives>
+
+    <!-- Unknown queues use the batch directives for the regular queue -->
+    <unknown_queue_directives>regular</unknown_queue_directives>
+
     <queues>
       <queue walltimemax="12:00" nodemin="1" nodemax="4032">regular</queue>
       <queue walltimemax="12:00" nodemin="1" nodemax="4032">premium</queue>
--- a/config/xml_schemas/config_batch.xsd
+++ b/config/xml_schemas/config_batch.xsd
@@ -29,6 +29,7 @@
   <xs:element name="walltime" type="xs:string"/>
   <xs:element name="dependency" type="xs:string"/>
   <xs:element name="prereq" type="xs:string"/>
+  <xs:element name="unknown_queue_directives" type="xs:string"/>
 
   <!-- complex elements -->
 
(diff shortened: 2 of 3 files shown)


billsacks:
A few notes for review:
- I am requesting a merge into maint-5.6 so that this fix can appear in tutorials using the CESM2.1.x code base
- If this PR is approved, I'll redo my manual tests with this rebased branch (I have only run testing off of my original branch, which was off of master)
- If this PR is approved, I'll also run scripts_regression_tests (I haven't done this yet)

config/cesm/machines/config_batch.xml
```diff
@@ -201,6 +201,10 @@
       <directive default="/bin/bash" > -S {{ shell }}  </directive>
       <directive> -l select=1:mpiprocs={{ total_tasks }}:ompthreads={{ thread_count }}</directive>
     </directives>
+
+    <!-- Unknown queues use the batch directives for the regular queue -->
+    <unknown_queue_directives>regular</unknown_queue_directives>
```
billsacks: I don't love this name. I'm very open to suggestions.

billsacks:
Thanks for the quick turnaround, @jedwards4b . The rest of my day is totally booked with the LMWG meeting, so it might be a day or so before I get a chance to redo my testing on this version of the code. I'll let you know when that's done and this is ready to merge.

billsacks:
Sorry for the delay here. I have now run my manual tests and scripts_regression_tests (on cheyenne) on this version destined for maint-5.6, so I think this is ready to go.

PR #3016, opened 2019-02-23
Add Machine Coeus to xml config files

colleenRooney:
Add changes necessary to run cesm on the machine coeus to xml config files.

Test namelist changes: 
Test status: [2 failures](https://bb.cgd.ucar.edu/test-compiler-changeable-build-time)
Fixes N/A

User interface changes?: N

Code review:

### DIFF
--- a/config/cesm/machines/config_compilers.xml
+++ b/config/cesm/machines/config_compilers.xml
@@ -601,6 +601,60 @@ using a fortran linker.
   </SLIBS>
 </compiler>
 
+<compiler MACH="coeus" COMPILER="gnu">
+  <CFLAGS>
+    <base> -std=gnu99 </base>
+    <append compile_threaded="true"> -fopenmp </append>
+    <append DEBUG="TRUE"> -g -Wall -Og -fbacktrace -ffpe-trap=invalid,zero,overflow -fcheck=bounds </append>
+    <append DEBUG="FALSE"> -O </append>
+  </CFLAGS>
+  <CPPDEFS>
+    <!-- http://gcc.gnu.org/onlinedocs/gfortran/ -->
+    <append> -DFORTRANUNDERSCORE -DNO_R16 -DCPRGNU</append>
+  </CPPDEFS>
+  <CXX_LINKER>FORTRAN</CXX_LINKER>
+  <FC_AUTO_R8>
+    <base> -fdefault-real-8 </base>
+  </FC_AUTO_R8>
+  <FFLAGS>
+    <!-- -ffree-line-length-none and -ffixed-line-length-none need to be in FFLAGS rather than in FIXEDFLAGS/FREEFLAGS
+       so that these are passed to cmake builds (cmake builds don't use FIXEDFLAGS and FREEFLAGS). -->
+    <base>  -fconvert=big-endian -ffree-line-length-none -ffixed-line-length-none </base>
+    <append compile_threaded="true"> -fopenmp </append>
+    <!-- Ideally, we would also have 'invalid' in the ffpe-trap list. But at
+         least with some versions of gfortran (confirmed with 5.4.0, 6.3.0 and
+         7.1.0), gfortran's isnan (which is called in cime via the
+         CPRGNU-specific shr_infnan_isnan) causes a floating point exception
+         when called on a signaling NaN. -->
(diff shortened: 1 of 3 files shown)


config/cesm/machines/config_compilers.xml
```diff
@@ -601,6 +601,60 @@ using a fortran linker.
   </SLIBS>
 </compiler>
 
+<compiler MACH="coeus" COMPILER="gnu">
```
jedwards4b: The idea in config_compilers.xml is to define a base set of flags for the gnu compiler (above in the file) and then only modify those settings for a particular machine as necessary.   Can you reduce this xml clause to be only the flags that are changed for coeus from the default gnu settings?

config/cesm/machines/config_machines.xml
```diff
@@ -433,6 +433,63 @@ This allows using a different mpirun command to launch unit tests
...
+      Portland State University Coeus Cluster CentOS 7
+    </DESC>
+    <NODENAME_REGEX>(login[1,2].cluster|compute[0-9]*.cluster)</NODENAME_REGEX>
+    <OS>LINUX</OS>
+    <COMPILERS>gnu</COMPILERS>
+    <MPILIBS>mvapich2</MPILIBS>
+    <PROJECT>none</PROJECT>
+    <CIME_OUTPUT_ROOT>$ENV{CESMDATAROOT}/$USER</CIME_OUTPUT_ROOT>
+    <DIN_LOC_ROOT>$ENV{CESMDATAROOT}/inputdata</DIN_LOC_ROOT>
+    <DOUT_S_ROOT>$ENV{CESMDATAROOT}/archive/$CASE</DOUT_S_ROOT>
```
jedwards4b: Usually you will want the DOUT_S_ROOT to be $USER specific like
$CIME_OUTPUT_ROOT/archive/$CASE

config/cesm/machines/config_machines.xml
```diff
@@ -433,6 +433,63 @@ This allows using a different mpirun command to launch unit tests
...
+    <BASELINE_ROOT>$ENV{CESMDATAROOT}/cesm_baselines</BASELINE_ROOT>
+    <CCSM_CPRNC>/vol/apps/hpc/src/cesm-2.1.0/cime/tools/cprnc/cprnc</CCSM_CPRNC>
+    <GMAKE>make</GMAKE>
+    <GMAKE_J>8</GMAKE_J>
+    <BATCH_SYSTEM>slurm</BATCH_SYSTEM>
+    <SUPPORTED_BY>oit-rc-groups@pdx.edu</SUPPORTED_BY>
+    <MAX_TASKS_PER_NODE>19</MAX_TASKS_PER_NODE>
+    <MAX_MPITASKS_PER_NODE>19</MAX_MPITASKS_PER_NODE>
```
jedwards4b: 19 is a very strange number for tasks per node - what type of machine is this?

colleenRooney: Thank you for all your feedback. Our compute nodes are Dual Intel Xeon E2630 v4, 10 cores each. I have heard mixed advice on if this should be kept 1 below the number of cores, should I bump this up to 20?

config/cesm/machines/config_machines.xml
colleenRooney: That make sense I'll fix this now

config/cesm/machines/config_compilers.xml
colleenRooney: I'll work on seeing what the minimum settings to still have the regression tests pass to the same level are.

config/cesm/machines/config_machines.xml
jedwards4b: There is occasionally good reason to leave a core available for the OS but we would typically set
MAX_MPI_TASKS_PER_NODE=20 and MAX_TASKS_PER_NODE=40 - (this allows for smt threads which may improve performance in some cases) Try it with a case or two and see what works best for you.  The PFS test is designed to measure performance of a given compset.

colleenRooney:
@jedwards4b I have made all the suggested changes. All the tests in the regression tests script pass now. I have not used the PFS test to test performance of different max task settings yet, but I plan to tonight or tomorrow.

PR #3018, opened 2019-02-26
Checksum inputdata

### LINKED ISSUE #2910: Zero size inputdata file created when file isn't availble from the data repo.
I've run into a problem where an inputdata file of 0 size is created.  This happens when the file needs to be
downloaded from the data repo, but it isn't available.  

Reproduce the problem using cesm2_1_alpha01e.  Change your DIN_LOC_ROOT, then.
./create_test ERS_Ld5.f09_g17.B1PCTcmip6.cheyenne_intel.allactive-defaultio_min

inputdata/atm/waccm/lb/LBC_CMIP6_1pctCO2_y1-165_GlobAnnAvg_0p5degLat_c180929.nc
Will be created with a size of 0.

`   Model cam missing file flbc_file = '/glade/scratch/fischer/inputdata/atm/waccm/lb/LBC_CMIP6_1pctCO2_y1-165_GlobAnnAvg_0p5degLat_c180929.nc'
Trying to download file: 'atm/waccm/lb/LBC_CMIP6_1pctCO2_y1-165_GlobAnnAvg_0p5degLat_c180929.nc' to path '/glade/scratch/fischer/inputdata/atm/waccm/lb/LBC_CMIP6_1pctCO2_y1-165_GlobAnnAvg_0p5degLat_c180929.nc'
  File "./check_input_data", line 72, in <module>
    _main_func(__doc__)
  File "./check_input_data", line 67, in _main_func
    download=download) else 1)
  File "/gpfs/u/home/fischer/code/cesm2_1_alpha01e/cime/scripts/Tools/../../scripts/lib/CIME/case/check_input_data.py", line 50, in check_all_input_data
    success = _downloadfromserver(self, input_data_root, data_list_dir)
  File "/gpfs/u/home/fischer/code/cesm2_1_alpha01e/cime/scripts/Tools/../../scripts/lib/CIME/case/check_input_data.py", line 65, in _downloadfromserver
    input_data_root=input_data_root, data_list_dir=data_list_dir)
  File "/gpfs/u/home/fischer/code/cesm2_1_alpha01e/cime/scripts/Tools/../../scripts/lib/CIME/case/check_input_data.py", line 209, in check_input_data
    isdirectory=rel_path.endswith(os.sep))
  File "/gpfs/u/home/fischer/code/cesm2_1_alpha01e/cime/scripts/Tools/../../scripts/lib/CIME/case/check_input_data.py", line 37, in _download_if_in_repo
    return server.getfile(rel_path, full_path)
  File "/gpfs/u/home/fischer/code/cesm2_1_alpha01e/cime/scripts/Tools/../../scripts/lib/CIME/Servers/wget.py", line 35, in getfile
    logging.warning("wget failed with output: {} and errput {}\n".format(output, errput))
UnicodeEncodeError: 'ascii' codec can't encode character u'\u2018' in position 147: ordinal not in range(128)
`

jedwards4b:
Add checksum support for inputdata:  When files are downloaded the data is checksummed against the file inputdata_chksum.dat which has a format:
```  
 df9a88f7dd33ea7451307e2b3b68cea5             1445  eaton            1661657404  Feb  20   2010 inputdata/atm/cam/chem/1850-2000_emis/IPCC_emissions_aircraft_BC_1850-2000_1.9x2.5.c090729.nc
```
the first and last fields read into a hash and as files are downloaded the file check sum is compared to the hash in the file.  

Test suite: by hand, scripts_regression_tests.py
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #2910 

User interface changes?: adds a --chksum option to check_input_data, files downloaded are always checked, the option is to check files already on local disk for a given case

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case/check_input_data.py
+++ b/scripts/lib/CIME/case/check_input_data.py
@@ -6,7 +6,7 @@ from CIME.utils import SharedArea, find_files, safe_copy, expect
 from CIME.XML.inputdata import Inputdata
 import CIME.Servers
 
-import glob, hashlib
+import glob, hashlib, time
 
 logger = logging.getLogger(__name__)
 # The inputdata_checksum.dat file will be read into this hash if it's available
@@ -121,6 +121,9 @@ def _downloadfromserver(case, input_data_root, data_list_dir):
     success = False
     protocol = 'svn'
     inputdata = Inputdata()
+    if not input_data_root:
+        input_data_root = case.get_value('DIN_LOC_ROOT')
+    lockfile = _check_permissions_and_lock_inputdata(input_data_root, case.get_value('USER'))
     while not success and protocol is not None:
         protocol, address, user, passwd, chksum_file = inputdata.get_next_server()
         logger.info("Checking server {} with protocol {}".format(address, protocol))
@@ -128,6 +131,7 @@ def _downloadfromserver(case, input_data_root, data_list_dir):
                                         input_data_root=input_data_root,
                                         data_list_dir=data_list_dir,
                                         user=user, passwd=passwd, chksum_file=chksum_file)
+    os.remove(lockfile)
     return success
 
 def stage_refcase(self, input_data_root=None, data_list_dir=None):
@@ -203,7 +207,6 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
     # Fill in defaults as needed
     input_data_root = case.get_value("DIN_LOC_ROOT") if input_data_root is None else input_data_root
 
-    expect(os.path.isdir(input_data_root), "Invalid input_data_root directory: '{}'".format(input_data_root))
     expect(os.path.isdir(data_list_dir), "Invalid data_list_dir directory: '{}'".format(data_list_dir))
 
     data_list_files = find_files(data_list_dir, "*.input_data_list")
@@ -292,6 +295,22 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
 
     return no_files_missing
 
+def _check_permissions_and_lock_inputdata(input_data_root, user):
+    expect(os.path.isdir(input_data_root), "Invalid input_data_root directory: '{}'".format(input_data_root))
+    # check for existing lock
+    lock_file = os.path.join(input_data_root,"inputdata.lock")
+    # if the file exists print it's contents and error out
+    if os.path.isfile(lock_file):
+        with open(lock_file) as f:
+            line = f.readline()
+        expect(False,line)
+    # if it doesn't exist try to create it
+    try:
+        with open(lock_file, "w") as f:
+            f.write("Directory {} locked by user {} on {} at {}".format(input_data_root, user, time.strftime("%y%m%d"), time.strftime("%H%M%S")))
+    except IOError as x:
+        expect(False, "Write to directory {} failed with error {}: {}".format(input_data_root, x.errno, x.strerror))
+    return lock_file
 
 def verify_chksum(input_data_root, checksumfile, filename):
     if not chksum_hash:
--- a/scripts/lib/CIME/case/check_input_data.py
+++ b/scripts/lib/CIME/case/check_input_data.py
@@ -217,13 +217,13 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
     expect(data_list_files, "No .input_data_list files found in dir '{}'".format(data_list_dir))
 
     no_files_missing = True
-    user = case.get_value('USER')
+    local_user = case.get_value('USER')
     if download:
         chksum_hash.clear()
         if protocol not in vars(CIME.Servers):
             logger.warning("Client protocol {} not enabled".format(protocol))
             return False
-
+        logger.info("Using protocol {} with user {} and passwd {}".format(protocol, user, passwd))
         if protocol == "svn":
             server = CIME.Servers.SVN(address, user, passwd)
         elif protocol == "gftp":
@@ -281,12 +281,12 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
                             if (download):
                                 if chksum_file and firstdownload:
                                     # Get the md5 checksum file
-                                    got_chksum = _download_checksum_file(server, input_data_root, chksum_file, user)
+                                    got_chksum = _download_checksum_file(server, input_data_root, chksum_file, local_user)
                                     firstdownload = False
                                 isdirectory=rel_path.endswith(os.sep)
                                 no_files_missing = _download_if_in_repo(server,
                                                                         input_data_root, rel_path.strip(os.sep),
-                                                                        user, isdirectory=isdirectory)
+                                                                        local_user, isdirectory=isdirectory)
                                 if got_chksum and no_files_missing and not isdirectory:
                                     verify_chksum(input_data_root,chksum_file, rel_path.strip(os.sep))
                         else:
--- a/scripts/lib/CIME/case/check_input_data.py
+++ b/scripts/lib/CIME/case/check_input_data.py
@@ -6,24 +6,22 @@ from CIME.utils import SharedArea, find_files, safe_copy, expect
 from CIME.XML.inputdata import Inputdata
 import CIME.Servers
 
-import glob, hashlib
+import glob, hashlib, fileinput
 
 logger = logging.getLogger(__name__)
 # The inputdata_checksum.dat file will be read into this hash if it's available
 chksum_hash = dict()
+local_chksum_file = 'inputdata_checksum.dat'
 
-
-def _download_checksum_file(server, input_data_root):
+def _download_checksum_file(server, input_data_root, chksum_file):
     """
     Return True if successfully downloaded
     """
-    rel_path = "../inputdata_checksum.dat"
-    full_path = os.path.join(input_data_root, "inputdata_checksum.dat")
+    success = False
+    rel_path = chksum_file
+    full_path = os.path.join(input_data_root, local_chksum_file)
     protocol = type(server).__name__
-    if protocol == 'svn':
-        # svn server does not have this file
-        return False
-    logging.info("Trying to download file: '{}' to path '{}' using {} protocal.".format(rel_path, full_path, protocol))
+    logging.info("Trying to download file: '{}' to path '{}' using {} protocol.".format(rel_path, full_path, protocol))
     tmpfile = None
     if os.path.isfile(full_path):
         tmpfile = full_path+".tmp"
@@ -34,16 +32,28 @@ def _download_checksum_file(server, input_data_root):
         success = server.getfile(rel_path, full_path)
         if success:
             if tmpfile:
-                os.remove(tmpfile)
+                _merge_chksum_files(full_path, tmpfile)
+                chksum_hash = dict()
         else:
-            if os.path.isfile(tmpfile):
+            if tmpfile and os.path.isfile(tmpfile):
                 os.rename(tmpfile, full_path)
                 logger.warning("Could not automatically download file "+full_path+
                            " Restoring existing version.")
             else:
                 logger.warning("Could not automatically download file "+full_path+
-                           "download from ftp:ftp.cgd.ucar.edu/cesm/inputdata_chksum.dat")
-        return success
+                           " download from ftp:ftp.cgd.ucar.edu/cesm/inputdata_checksum.dat")
+    return success
+
+def _merge_chksum_files(new_file, old_file):
+    with open(old_file) as fin:
+        lines = fin.readlines()
+    with open(new_file) as fin:
+        lines += fin.readlines()
+    lines = set(lines)
+    with open(new_file, "w") as fout:
+        fout.write("".join(lines))
+    os.remove(old_file)
+
 
 
 def _download_if_in_repo(server, input_data_root, rel_path, isdirectory=False):
@@ -54,7 +64,7 @@ def _download_if_in_repo(server, input_data_root, rel_path, isdirectory=False):
         return False
 
     full_path = os.path.join(input_data_root, rel_path)
-    logging.info("Trying to download file: '{}' to path '{}' using {} protocal.".format(rel_path, full_path, type(server).__name__))
+    logging.info("Trying to download file: '{}' to path '{}' using {} protocol.".format(rel_path, full_path, type(server).__name__))
     # Make sure local path exists, create if it does not
     if isdirectory or full_path.endswith(os.sep):
         if not os.path.exists(full_path):
@@ -96,10 +106,12 @@ def _downloadfromserver(case, input_data_root, data_list_dir):
     protocol = 'svn'
     inputdata = Inputdata()
     while not success and protocol is not None:
-        protocol, address, user, passwd = inputdata.get_next_server()
+        protocol, address, user, passwd, chksum_file = inputdata.get_next_server()
         logger.info("Checking server {} with protocol {}".format(address, protocol))
         success = case.check_input_data(protocol=protocol, address=address, download=True,
-        input_data_root=input_data_root, data_list_dir=data_list_dir, user=user, passwd=passwd)
+                                        input_data_root=input_data_root,
+                                        data_list_dir=data_list_dir,
+                                        user=user, passwd=passwd, chksum_file=chksum_file)
     return success
 
 def stage_refcase(self, input_data_root=None, data_list_dir=None):
@@ -167,7 +179,7 @@ def stage_refcase(self, input_data_root=None, data_list_dir=None):
     return True
 
 def check_input_data(case, protocol="svn", address=None, input_data_root=None, data_list_dir="Buildconf",
-                     download=False, chksum=False, user=None, passwd=None):
+                     download=False, chksum=False, user=None, passwd=None, chksum_file=None):
     """
     Return True if no files missing
     """
@@ -184,6 +196,7 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
     no_files_missing = True
 
     if download:
+        chksum_hash = dict()
         if protocol not in vars(CIME.Servers):
             logger.warning("Client protocol {} not enabled".format(protocol))
             return False
@@ -195,8 +208,7 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
         elif protocol == "ftp":
             server = CIME.Servers.FTP(address, user, passwd)
         elif protocol == "wget":
-#            server = CIME.Servers.WGET(address, user, passwd)
-            server = CIME.Servers.WGET(address)
+            server = CIME.Servers.WGET(address, user, passwd)
         else:
             expect(False, "Unsupported inputdata protocol: {}".format(protocol))
 
@@ -244,23 +256,20 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
                             no_files_missing = False
 
                             if (download):
-                                if firstdownload:
+                                if chksum_file and firstdownload:
                                     # Get the md5 checksum file
-                                    got_chksum = _download_checksum_file(server, input_data_root)
+                                    got_chksum = _download_checksum_file(server, input_data_root, chksum_file)
                                     firstdownload = False
                                 isdirectory=rel_path.endswith(os.sep)
                                 no_files_missing = _download_if_in_repo(server, input_data_root, rel_path.strip(os.sep),
                                                                         isdirectory=isdirectory)
                                 if got_chksum and no_files_missing and not isdirectory:
-                                    verify_chksum(input_data_root,"inputdata_checksum.dat", rel_path.strip(os.sep))
+                                    verify_chksum(input_data_root,chksum_file, rel_path.strip(os.sep))
                         else:
                             if chksum:
-                                verify_chksum(input_data_root,"inputdata_checksum.dat", rel_path.strip(os.sep))
+                                verify_chksum(input_data_root,chksum_file, rel_path.strip(os.sep))
                                 logger.info("Chksum passed for file {}".format(os.path.join(input_data_root,rel_path)))
                             logging.debug("  Already had input file: '{}'".format(full_path))
-
-
-
                 else:
                     model = os.path.basename(data_list_file).split('.')[0]
                     logging.warning("Model {} no file specified for {}".format(model, description))
@@ -270,10 +279,9 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
 
 def verify_chksum(input_data_root, checksumfile, filename):
     if not chksum_hash:
-        hashfile = os.path.join(input_data_root,checksumfile)
+        hashfile = os.path.join(input_data_root, local_chksum_file)
         if not os.path.isfile(hashfile):
-            expect(False, "Failed to find or download file "+hashfile+
-                   " download from ftp:ftp.cgd.ucar.edu/cesm/inputdata_chksum.dat")
+            expect(False, "Failed to find or download file {}".format(hashfile))
 
         with open(hashfile) as fd:
             lines = fd.readlines()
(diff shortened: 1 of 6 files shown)


ekluzek:
This would also help in finding files that have zero size or aren't available on svn inputdata, so it wouldn't create a local file of zero size. This would help prevent problems like:

https://github.com/ESCOMP/ctsm/issues/646

where a zero size file is created, because it's not in svn.

jedwards4b:
@ekluzek if a file is not on the server an error is printed and a zero length file is no longer produced.
running ./check_input_data --chksum should revile any relevant files in DIN_LOC_ROOT which do not
match the chksum file.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -6,9 +6,45 @@
...
 logger = logging.getLogger(__name__)
+# The inputdata_checksum.dat file will be read into this hash if it's available
+chksum_hash = dict()
+
+
+def _download_checksum_file(server, input_data_root):
+    """
+    Return True if successfully downloaded
+    """
+    rel_path = "../inputdata_checksum.dat"
```
billsacks: Is it a problem that there is only one checksum file despite potentially having multiple inputdata servers? I would imagine having a separate checksum file for each server, with different file names distinguishing them. But maybe I'm missing something here.

Maybe this isn't a huge deal, because (if I understand the operation of this code correctly), a given call to `check_input_data` only works with a single remote server, and so would create a new `inputdata_checksum.dat` for that server. But it seems like this could still create some confusion, especially if there are any problems.

I don't feel strongly that this needs to be fixed, but wanted to run this by you.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -6,9 +6,45 @@
...
+    full_path = os.path.join(input_data_root, "inputdata_checksum.dat")
+    protocol = type(server).__name__
+    if protocol == 'svn':
```
billsacks: It seems like this logic is specific to our current svn inputdata server, and should be encoded in the xml defining the different servers rather than generally skipping this for any svn server.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -6,9 +6,45 @@
...
+        # svn server does not have this file
+        return False
+    logging.info("Trying to download file: '{}' to path '{}' using {} protocal.".format(rel_path, full_path, protocol))
```
billsacks: Spelling: "protocal" should be "protocol"

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -6,9 +6,45 @@
...
+    # have +s, then everything should work.
+    with SharedArea():
+        success = server.getfile(rel_path, full_path)
+        if success:
+            if tmpfile:
+                os.remove(tmpfile)
+        else:
+            if os.path.isfile(tmpfile):
+                os.rename(tmpfile, full_path)
+                logger.warning("Could not automatically download file "+full_path+
```
billsacks: Connected with some of my above comments: it seems like this warning is misleading for two reasons:

(1) Based on what you told me in person, it sounds like it's acceptable for some servers to not have this file, so maybe this shouldn't generate a warning? I wonder if each server's xml entry should say whether that file has a checksum or not; then we'll skip trying to download it if it doesn't have one, and just try to download it if it is supposed to have one (generating this warning if it isn't found).

(2) The comment about "Restoring existing version" is a bit misleading because it may be restoring a file for a completely different server right now.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -6,9 +6,45 @@
...
+                           " Restoring existing version.")
+            else:
+                logger.warning("Could not automatically download file "+full_path+
+                           "download from ftp:ftp.cgd.ucar.edu/cesm/inputdata_chksum.dat")
```
billsacks: Shouldn't this hard-coded path be replaced with the actual remote path for this particular server?

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -37,15 +73,16 @@ def _download_if_in_repo(server, input_data_root, rel_path, isdirectory=False):
             return server.getfile(rel_path, full_path)
 
 ###############################################################################
-def check_all_input_data(self, protocol=None, address=None, input_data_root=None, data_list_dir="Buildconf", download=True):
+def check_all_input_data(self, protocol=None, address=None, input_data_root=None, data_list_dir="Buildconf",
+                         download=True, chksum=False):
```
billsacks: Please add documentation for the new chksum argument. (Ideally there would be documentation for all arguments.)

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -129,7 +166,8 @@ def stage_refcase(self, input_data_root=None, data_list_dir=None):
             os.remove(os.path.join("Buildconf","refcase.input_data_list"))
     return True
 
-def check_input_data(case, protocol="svn", address=None, input_data_root=None, data_list_dir="Buildconf", download=False):
+def check_input_data(case, protocol="svn", address=None, input_data_root=None, data_list_dir="Buildconf",
+                     download=False, chksum=False, user=None, passwd=None):
```
billsacks: Please add documentation for the new chksum argument. (Ideally there would be documentation for all arguments.)

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -205,13 +244,57 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
...
+
                 else:
                     model = os.path.basename(data_list_file).split('.')[0]
                     logging.warning("Model {} no file specified for {}".format(model, description))
 
     return no_files_missing
+
+
+def verify_chksum(input_data_root, checksumfile, filename):
+    if not chksum_hash:
```
billsacks: Similar to my above comment about only having one checksum file for all servers: I think it's a problem that you only have a single chksum_hash variable for all servers. My understanding of the logic here is that you'll read the checksum file for the first server, but then for each subsequent server, you won't reset chksum_hash, but instead will use the variable from the first server. If my understanding is right, then this is a problem.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -205,13 +244,57 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
...
+            for line in lines:
+                lsplit = line.split()
+                if len(lsplit) < 8:
+                    continue
+                fname = (lsplit[7])[10:]
+                chksum_hash[fname] = lsplit[0]
+
+    chksum = md5(os.path.join(input_data_root, filename))
+    if chksum_hash:
+        expect(filename in chksum_hash.keys() and chksum == chksum_hash[filename],
```
billsacks: It seems like there are a few problems with expecting the current file to be in chksum_hash:

1. Tied in to my above comments: let's say we just downloaded a file from server B, but chksum_hash was set based on server A; then this new file would be missing from chksum_hash.

2. When running `check_input_data` with the new `--chksum` argument: My impression is that you'll end up trying to do this checksum for each server, but the file only exists on one server, so wouldn't the expect fail for all of the other servers? This case also has the problem of (1). In addition, due to the problem of using the same file name for all servers, this case has the problem that the checksum file on disk may never be for the server this file resides on (if no files need to be downloaded from that server), because it looks like existing files don't force a download of the checksum file from the server.

3. Can we assume that the checksum file on the server is updated near-instantaneously whenever a file is added to the server? If not, we could have a situation where you upload a file from one machine, then a few minutes later go to download it on a different machine, and get an error because the new file is not yet in the server's checksum file.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -205,13 +244,57 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
...
 
     return no_files_missing
+
+
+def verify_chksum(input_data_root, checksumfile, filename):
+    if not chksum_hash:
+        hashfile = os.path.join(input_data_root,checksumfile)
+        if not os.path.isfile(hashfile):
+            expect(False, "Failed to find or download file "+hashfile+
+                   " download from ftp:ftp.cgd.ucar.edu/cesm/inputdata_chksum.dat")
```
billsacks: Again, shouldn't this path be dynamic?

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -205,13 +244,57 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
...
+
+        with open(hashfile) as fd:
+            lines = fd.readlines()
+            for line in lines:
+                lsplit = line.split()
+                if len(lsplit) < 8:
+                    continue
+                fname = (lsplit[7])[10:]
```
billsacks: I'm uncomfortable with the hard-coded [10:]. Can you instead just strip the leading 'inputdata/' from this string?

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -205,13 +244,57 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
...
+        if not os.path.isfile(hashfile):
+            expect(False, "Failed to find or download file "+hashfile+
+                   " download from ftp:ftp.cgd.ucar.edu/cesm/inputdata_chksum.dat")
+
+        with open(hashfile) as fd:
+            lines = fd.readlines()
+            for line in lines:
+                lsplit = line.split()
+                if len(lsplit) < 8:
+                    continue
```
billsacks: Would it be feasible to require a header line on the checksum file, so that you could confirm that indeed the 1st column is the checksum and the 8th column is the file name?

I don't feel this is necessary, but it could help avoid some hard-to-debug errors in the future.

scripts/lib/CIME/case/check_input_data.py
billsacks: In addition, if you do keep the check of filename in chksum_hash.keys(), you should make this a separate check, rather than checking both conditions at once: These two issues (missing file name in the chksum file and mismatching chksum) are fundamentally different problems, and it's confusing to have them reported with the same error.

billsacks:
In addition to my review comments: I'm uncomfortable with this coming to maint-5.6 so close to the CESM2.1.1 release. I'd prefer if we had more time to exercise this functionality before it made it into a release - either by waiting to bring this to maint-5.6 until after this branch is tagged for CESM2.1.1, or by bringing it into cime master first.

billsacks:
> or by bringing it into cime master first.

However, if you do rebase this to be off of master, I'd find it easiest if you addressed my review comments in this PR before closing the PR and opening a new one.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -6,9 +6,41 @@
...
+chksum_hash = dict()
+
+
+def _download_checksum_file(server, input_data_root):
+    """
+    Return True if successfully downloaded
+    """
+    rel_path = "../inputdata_checksum.dat"
+    full_path = os.path.join(input_data_root, "inputdata_checksum.dat")
+    logging.info("Trying to download file: '{}' to path '{}' using {} protocal.".format(rel_path, full_path, type(server).__name__))
```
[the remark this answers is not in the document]
jedwards4b: fixed these

scripts/lib/CIME/case/check_input_data.py
jedwards4b: you are correct, i've added the file name to the config_inputdata.xml

jedwards4b:
@billsacks I think that I've addressed your concerns - my understanding is that the 2.1.1 release won't be until at least June so plenty of time to put this on maint-5.6

scripts/lib/CIME/case/check_input_data.py
jedwards4b: I agree that this could be a problem, I've dealt with it as follows:  The server checksum file is now listed in config_inputdata.xml and the client checksum file is a unique merge of server checksum files.   If the same file has different checksums on different servers an error will result.

jedwards4b:
I've also added a lockfile for DIN_LOC_ROOT to assure that more than one case or user is not trying to update at the same time.   It also gives a better error if permissions do not allow the user to write there.

billsacks:
@jedwards4b Thanks for addressing my comments. I haven't looked closely yet, but I'll try to do so soon (not sure if I'll have time today, though).

I wanted to reply to the file locking piece first, though: I hadn't thought of this issue, but you raise a good point about the problem if multiple cases are trying to do this at once. However, I'm not sure I like the solution of adding this top-level lock on the check_input_data call: it seems like this could very easily bite people in practice. I'm especially thinking of test suites, where multiple cases are working simultaneously... and when you multiply that by multiple users on a system, it seems almost inevitable that cases will fail sometimes due to this lock.

Can we think a little more about what purpose(s) this lock serves, and whether this can be done in a different way that would reduce lock-induced failures to a bare minimum? I'm thinking:

1. Due to the changes in this PR, we need a lock on the local checksum file, to ensure that two processes aren't trying to write this checksum file at the same time. But: is there actually much value in having the checksum file reside permanently in this inputdata location? Would things work just as well if the checksum file was written as a temporary file (e.g., via https://docs.python.org/2/library/tempfile.html)? You could save the name of the temporary file in a module-level variable so that a given process would only need to download the checksum file once per server, but this would avoid clashes between different processes. However, I forget the details of the logic, so let me know if this wouldn't work because you rely on the persistence of this checksum file for some reason.

2. Were you also thinking about the need for locking if two cases are trying to download the same file from the server at the same time? (This is no more of an issue in this PR than it would have been previously, but maybe you were thinking of this here, too?) Or would the OS already prevent this? If the OS wouldn't already prevent this, then can we do a per-file locking mechanism to just lock what we absolutely need to lock? i.e., in the directory where we're about to download a file, putting a FILENAME.lock file, then removing that file as soon as that file is downloaded.

jedwards4b:
@billsacks the --chksum option assumes that the inputdata_checksum.dat file is present so it does need to be there for this to work.   This option allows a user to chksum all of the existing local input files for validity and doesn't require the user to have write permission to inputdata.   

Most of the time we do not need to download new data for testing or running cases, but it is true that if  multiple cases in a test suite need to download this could cause failures.   I'll look into changing this to a per file lock as you suggest.

billsacks:
@jedwards4b thanks for changing this to a per-file lock. This addresses point (2) from my earlier comment. But my impression is that now there is no lock on the checksum files. It really feels to me like the best thing to do with the checksum files is to download them to a temporary location, even when running with `--chksum`. This doesn't have to be in the inputdata space (since, as you point out, the user may not have write access there): e.g., I seem to remember that python's tempfile module will create a temporary file/directory in some special, OS-specific temporary location on disk where you have write permission; or perhaps this checksum file could be put in the case or run directories.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -6,19 +6,84 @@
...
 
-def _download_if_in_repo(server, input_data_root, rel_path, isdirectory=False):
+def _download_checksum_file(server, input_data_root, chksum_file, user):
+    """
+    Return True if successfully downloaded
+    """
+    success = False
+    rel_path = chksum_file
+    full_path = os.path.join(input_data_root, local_chksum_file)
+    lockfile = _check_permissions_and_lock_inputdata_file(input_data_root, user)
```
jedwards4b: @billsacks here is the lock on the checksum file.

jedwards4b: The checksum file on local disk may be a combination of checksums from several servers and 
is available to any user once a user with write access creates it.  The --chksum option may currently be used without any access to the inputdata server or servers, it seems useful to me that you would be able to use this option when servers are not available.

billsacks: Thanks for pointing that out. But I still stand by my feeling that this checksum file should be in a case-specific location, rather than having the same file shared between all cases and users on the system. The more I think about this, the more strongly I feel about it: I feel that having a single file that is potentially written by all cases, for all users, is asking for trouble.

That said, if you still disagree, can we get a 3rd opinion here - from @jgfouca or someone else?

billsacks: (My response was meant in response to your first comment here, not your second. I'll think about your second comment for a few minutes.)

billsacks: > The checksum file on local disk may be a combination of checksums from several servers and
> is available to any user once a user with write access creates it. The --chksum option may currently be used without any access to the inputdata server or servers, it seems useful to me that you would be able to use this option when servers are not available.

I see your point, but I feel that the value of being able to run this when servers are unavailable does not justify the race condition issues caused by sharing a single file amongst all users and cases.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -205,13 +279,80 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
...
+        hashfile = os.path.join(input_data_root, local_chksum_file)
+        if not os.path.isfile(hashfile):
+            expect(False, "Failed to find or download file {}".format(hashfile))
+
+        with open(hashfile) as fd:
+            lines = fd.readlines()
+            for line in lines:
+                fchksum, fname = line.split()
+                if fname in chksum_hash:
+                    expect(chksum_hash[fname] == fchksum, " Inconsistant hashes in chksum for file {}".format(fname))
```
billsacks: This is a good error-check - thanks.

Minor spelling error: Inconsistant should be Inconsistent.

scripts/lib/CIME/case/check_input_data.py
billsacks: Your solution seems good - thanks.

scripts/lib/CIME/case/check_input_data.py
billsacks: This comment and my other similar comment about argument documentation are still unresolved: this and other functions could use more documentation of their arguments.

billsacks:
Regarding my main remaining concern: One thing I've been wondering is: Why is there a single file per server with the checksums for all files on that server, as opposed to having a separate checksum file for each file on the server? It seems like the latter would minimize the potential for race conditions, and I think would simplify this PR, but maybe there's some reason for the current design that I'm not seeing.

gold2718:
@billsacks, I have been thinking about this question. I think the main reason for a single checksum file (as opposed to the usual practice of a checksum file for every downloadable file) is disk management. On Cheyenne, there are currently over 23K NetCDF files in inputdata. I think that many small (e.g., checksum) files is a real drain on a large glade system. @jedwards4b, does that match your reasoning?
For managed HPC centers like NWSC it would make sense to me to have a process to create and maintain a central file which is fairly stable over time while individual users should not be held up.
A temporary download of checksums to the case directory avoids deadlocks. An attempt can be made to add new checksums to the master file but if it is locked, just give up -- it will eventually get added.

jgfouca:
@billsacks , @jedwards4b , I'm taking another look at this PR.

My understanding of the checksumming algorithm is:
1. User requests a checksum
2. Retrieve the name of the checksum file for that server
3. A lock file is created in the input data area with the name from (2) + .lock
4. The file is downloaded, which contains a map of filename to checksum for every file on that server
4.A If the file already exists, it is renamed so it can be merged later
5. The file is looked-up and the checksum is validated

The problem I see is that steps 3, 4, and 4.A all touch shared files but none of the steps are atomic, so a wide variety of races is possible. I could maybe be inclined to tolerate some risk here, but I also think the current solution is more complicated than it needs to be, so I'm going to recommend a refactor.

I'm thinking why do we need to maintain a checksum dictionary locally at all? I'm sure we could come up with a solution where a shell command could be used to directly request the needed checksum without touching the file system. Something like:
```
wget $remote_checksum_dict_file | grep filename | cut ...
```

scripts/lib/CIME/case/check_input_data.py
jedwards4b: It turns out that I should remove this because the the file is now reformatted after download

scripts/lib/CIME/case/check_input_data.py
jedwards4b: I am working with an existing chksum file that Mark Moore generates - I didn't really want to make any request that he change the format of the file on his server.

jedwards4b:
So there are two different workflows here:

1. check_input_data determines that not all input_data for a case is available in the local inputdata directory.  In this case if the chksum file is available on the server it is downloaded and merged with any preexisting chksum file.  This merged file is then used to checksum each file from the server as it is downloaded.   Each time a file is downloaded a filename.LOCK file is written prior to the download and deleted when the download is complete.  This assures that two check_input_data processes don't attempt to download the same file at the same time. 
  
2.  A user requests to checksum all of the inputdata for their case using the --chksum option.  In this case the existing chksum file in input_data_root is used and each local file is checksummed.  No access to any server occurs.  There is no write of any file.    

I've added documentation throughout the check_input_data.py

jedwards4b:
@goldy my actual reasoning was that the cgd server has a single existing checksum file with a given format, I chose to work with that file in that format rather than ask the data steward to change his practice.

jgfouca:
@jedwards4b , OK, that makes sense, but I think I'd still prefer if download_checksum returned a string of all the checksum data instead of interacting with the file system. For your case 2, I assume it's not a problem if we access the server? The download of the checksum data should be real quick and I don't think it's worth it to do complex things in CIME to avoid this download. The lockfile solution reduces the race window, but I promise you it's still there. It also introduces another potential bug of leaving the lock file around if the CIME process crashes/raises exception at the wrong time.

jedwards4b:
okay - I think that I can rework it a bit so that the checksum file is downloaded to the case rundir for each case - is that acceptable as a way to avoid race conditions?

jgfouca:
@jedwards4b , yes, but why even store it in a file?

jgfouca:
Oh, I suppose it allows you to use Server.get_file. I'm fine with that.

billsacks:
> okay - I think that I can rework it a bit so that the checksum file is downloaded to the case rundir for each case - is that acceptable as a way to avoid race conditions?

Yes, that sounds good - thanks a lot.

You have addressed my earlier, specific points, so this rework of the location of the checksum file is the only thing left from my review.

jedwards4b:
@jgfouca @billsacks Please have another look.  I've moved the chksum file into the case directory and download it each time instead of trying to maintain it locally.

jgfouca:
@jedwards4b , now that you're using the case's rundir as the download location, can't you get rid of all the lockfile stuff?

jedwards4b:
I'm still downloading data to the inputdata directory the lock file stuff is an improvement somewhat independent of the chksum.

jgfouca:
@jedwards4b , instead of locking the entire inputdata directory, I would recommend using the atomic write pattern using file renaming for the individual downloads:

1. User needs to download file x/y/z.nc
2. Download file to x/y/z.nc.$unique
3. rename x/y/z.nc.$unique to x/y/z.nc

Where $unique is some string to ensure the filename is unique. It could be process-id.

jedwards4b:
I am not locking the entire inputdata directory - the lock is on individual files.

jgfouca:
@jedwards4b , oh, that's better, but I still think atomic write is best.

jedwards4b:
atomic write patterns are already happening implicitly - all of the download methods do this internally - this lock is to prevent a second download of the same file while the first is underway.

jgfouca:
> atomic write patterns are already happening implicitly

Can you elaborate? I don't see where this is happening.

If we are already doing atomic writes, then no locking is necessary. The worst thing that could happen is a bit of wasted bandwidth.

jedwards4b:
We do need locking - if two concurrently running cases need the same file and see that the file is not in inputdata they will independent try to download it.  When a case starts to download a file it creates this lock so that another case will know not to.  We occasionally run our tests with the DIN_LOC_ROOT pointed to an empty directory - this assures that all of the needed input data is in the repository - but there are all sorts of conflicts that occur because multiple tests are attempting to simultaneously download the same file.  This is a stab at better addressing that problem.  It should rarely occur in practice because most of the data is already on the local file system.

jedwards4b:
> Can you elaborate?
I'm not doing it explicitly, but the clients (gftp, ftp, svn, wget) do it internally.

jedwards4b:
testing complete - this is ready @billsacks  do you want to review again?

PR #3020, opened 2019-02-28
update NYF CORE2 DROF data stream file.

alperaltuntas:
Updates core2 NYF data runoff stream file. The area and runoff units of the old file (*090225.nc) were incorrect and were leading to very large area correction factors within mct driver. This also fixes the erroneous drof fluxes in noupc driver for NYF-forced cases.

Test suite: aux_pop
Test baseline: cesm2.2_alpha01d-pop_2_1_20190118
Test namelist changes: n/a
Test status: roundoff differences and namelist changes in NYF forced cases for mct. 

Fixes [CIME Github issue #] n/a

User interface changes?: n/a

Code review:

### DIFF
--- a/src/components/data_comps/drof/cime_config/namelist_definition_drof.xml
+++ b/src/components/data_comps/drof/cime_config/namelist_definition_drof.xml
@@ -89,10 +89,10 @@
     <group>streams_file</group>
     <desc>Stream domain file path(s).</desc>
     <values>
-      <value stream="rof.diatren_ann_rx1"      >runoff.daitren.annual.090225.nc</value>
-      <value stream="rof.diatren_ann_ais00_rx1">runoff.daitren.annual.090225.nc</value>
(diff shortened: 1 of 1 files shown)


rljacob:
@jonbob does the ERS_Ld5.T62_oQU120.CMPASO-NYF test use this dataset?

jonbob:
@rljacob - yes, I think it does

mvertens:
@rljacob - the issue is that if you do not apply the area correction factor
in the cpl - you will be off by several orders of magnitude. This does not
occur with the CIAF (inter annual forcing compset).

mvertens:
I think that would be good. Could we please check with Gokhan before hand
just to make sure.

rljacob:
@mvertens so should this PR also turn off the area correction factors in the coupler?  It only updated the filename.

mvertens:
@rljacob - NO we should never turn off the area correction factors in the
coupler. Its just that if you did - you get non-sensical results with the
old DROF dataset. For nuopc we are not using area correction factors and we
hit on this problem.

PR #3021, opened 2019-02-28
Jgfouca/branch for acme split 2019 02 28

jgfouca:
Apologies for the big PR, I will annotate.

Change list:
* create_test --single-exe capability
* bless_test_results upgrades (use Case instead of xmlquery, autodetect newest test batch)
* change test-suite definition format
* add kokkos as a buildable submodule of e3sm
* make sure to pick up all doctests throughout scripts/lib

Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


scripts/Tools/Makefile
```diff
@@ -30,20 +30,6 @@ CP    := cp
 exec_se: $(EXEC_SE)  Depends
 complib: $(COMPLIB)  Depends
 
-
-CIME_MODEL ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) MODEL --value)
```
jgfouca: This is worth a close look. I don't know who added these, but I don't think we want xmlquery calls in the Makefile since they are slow and we already have a mechanism (environment) for passing values to Make from build.py.

scripts/lib/CIME/bless_test_results.py
```diff
@@ -20,16 +20,6 @@ def bless_namelists(test_name, test_dir, report_only, force, baseline_name, base
     if (not report_only and
         (force or six.moves.input("Update namelists (y/n)? ").upper() in ["Y", "YES"])):
 
-        if baseline_name is None:
```
jgfouca: I don't think any of the changes in this file will impact the CESM workflow.

scripts/lib/CIME/build.py
```diff
@@ -457,10 +465,15 @@ def _case_build_impl(caseroot, case, sharedlib_only, model_only, buildlist,
     os.environ["CISM_USE_TRILINOS"]    = stringify_bool(cism_use_trilinos)
     os.environ["MALI_USE_ALBANY"]      = stringify_bool(mali_use_albany)
     os.environ["USE_MOAB"]             = stringify_bool(use_moab)
+    os.environ["PIO_VERSION"]          = str(case.get_value("PIO_VERSION"))
```
jgfouca: Making sure everything the Makefile might need is set in the environment so it doesn't have to use xmlquery.

scripts/lib/CIME/test_scheduler.py
```diff
@@ -641,6 +646,16 @@ def _xml_phase(self, test):
             case.set_value("TEST", True)
             case.set_value("SAVE_TIMING", self._save_timing)
 
+            # handle single-exe here, all cases will use the EXEROOT from
+            # the first case.
```
jgfouca: Change in this file were to support --single-exe.

scripts/lib/get_tests.py
```diff
@@ -14,124 +14,195 @@
...
-                             "DAE.ww3a.ADWAV",
-                             "PET_P4.f19_f19.A",
-                             "PEM_P4.f19_f19.A",
-                             "SMS.T42_T42.S",
-                             "PRE.f19_f19.ADESP",
-                             "PRE.f19_f19.ADESP_TEST",
-                             "MCC_P1.f19_g16_rx1.A",
-                             "LDSTA.f45_g37_rx1.A")
-                            ),
+    "cime_tiny" : {
```
jgfouca: Changing this format was a significant change but should have no impacts to CESM.

scripts/tests/scripts_regression_tests.py
```diff
@@ -138,13 +138,20 @@ def test_unittests(self):
...
-    def test_CIME_doctests(self):
-        # Find and run all the doctests in the CIME directory tree
-        run_cmd_assert_result(self, "PYTHONPATH=%s:$PYTHONPATH python -m doctest *.py 2>&1" % LIB_DIR, from_dir=os.path.join(LIB_DIR,"CIME"))
-
-    def test_CIMEXML_doctests(self):
-        # Find and run all the doctests in the XML directory tree
-        run_cmd_assert_result(self, "PYTHONPATH=%s:$PYTHONPATH python -m doctest *.py 2>&1" % LIB_DIR, from_dir=os.path.join(LIB_DIR,"CIME","XML"))
+    def test_lib_doctests(self):
+        # Find and run all the doctests in the lib directory tree
+        skip_list = ["six.py", "CIME/SystemTests/mvk.py", "CIME/SystemTests/pgn.py"]
```
jgfouca: This was a very important change. The old implementation was missing lots of doctests since it was only searching two directories.

billsacks:
@jgfouca can you please add a brief summary of the changes here in the top-level PR comment? It has been helpful when you have done that in the past.

scripts/Tools/Makefile
jedwards4b: I put them in and in general they are not used because the variable is already defined.  Was there a performance issue that caused you to remove them?
?= says only do this if the left hand side is not defined.

scripts/Tools/Makefile
```diff
@@ -528,7 +505,7 @@ INCLDIR +=	-I$(INSTALL_SHAREDPATH)/include
 # Use the MCT dir for the cache for all configure calls because it is the first one
 #
 CFLAGS+=$(CPPDEFS)
-CXXFLAGS := $(CFLAGS)
+CXXFLAGS += $(CFLAGS)
```
jedwards4b: this looks wrong both before and after the change.

scripts/Tools/Makefile
```diff
@@ -640,6 +613,9 @@ ifeq ($(USE_CXX), true)
   endif
 endif
 
+# Remove arch flag if it exists
```
jedwards4b: why? seems pretty obscure

scripts/lib/CIME/BuildTools/macrowriterbase.py
```diff
@@ -132,7 +132,7 @@ def write_line(self, line):
 
         A trailing newline is added, whether or not the input has one.
         """
-        self.output.write(str(self.indent_string() + line + "\n"))
+        self.output.write(u"{}{}\n".format(self.indent_string(), line))
```
jedwards4b: In cmakemacroswriter.py you went from u"{}" to str({}) here you are going the other way around - which should it be?

scripts/lib/CIME/build.py
jedwards4b: It doesn't in general use xmlquery and it doesn't need to use environment variables - most builds use command line arguments to make eg:
gmake PIO_VERSION=2

jedwards4b: I made a big effort to not rely on the environment for Makefile arguments - I really don't want to see that reversed.

src/build_scripts/buildlib.kokkos
```diff
@@ -0,0 +1,72 @@
...
+
+    # We want to get the compiler and kokkos_options from Macros.make
+    # (generated from config_compilers.xml), but we want to otherwise
+    # let kokkos control flags
+    kokkos_options = run_cmd_no_fail("make -f Macros.make -p | grep KOKKOS_OPTIONS").split(":=")[-1].strip()
+    cxx            = run_cmd_no_fail("make -f Macros.make -p | grep SCXX").split(":=")[-1].strip()
+
+    gmake_cmd = case.get_value("GMAKE")
+    gmake_j = case.get_value("GMAKE_J")
+
```
jedwards4b: If you get the Makefile options and pass them to gmake here you can avoid both environment variables and xmlquery calls.

scripts/lib/CIME/BuildTools/macrowriterbase.py
jgfouca: Both py2 and py3 expect this value to be a unicode str. This wasn't caught due to doctests not being run.

jgfouca:
@billsacks , I added a change list.

scripts/Tools/Makefile
jgfouca: it breaks the linking of the model (e3sm.exe).

billsacks:
> @billsacks , I added a change list.

Thanks!

scripts/Tools/Makefile
jedwards4b: oh I see now - this is okay

PR #3038, opened 2019-03-11
Chksum fixes maint 5.6

### LINKED ISSUE #3033: cime_developer tests failed on anlworkstation
The errors can be reproduced on anlworkstation by running a single cime_developer test:
`./create_test DAE.ww3a.ADWAV --machine=anlworkstation --test-root=****`

Errors were:
```
          File "./case.submit", line 108, in <module>
            _main_func(__doc__)
          File "./case.submit", line 105, in _main_func
            mail_user=mail_user, mail_type=mail_type, batch_args=batch_args)
          File "scripts/lib/CIME/case/case_submit.py", line 191, in submit
            custom_success_msg_functor=verbatim_success_msg)
          File "scripts/lib/CIME/utils.py", line 1702, in run_and_log_case_status
            rv = func()
          File "scripts/lib/CIME/case/case_submit.py", line 189, in <lambda>
            batch_args=batch_args)
          File "scripts/lib/CIME/case/case_submit.py", line 120, in _submit
            case.check_case()
          File "scripts/lib/CIME/case/case_submit.py", line 205, in check_case
            self.check_all_input_data()
          File "scripts/lib/CIME/case/check_input_data.py", line 159, in check_all_input_data
            chksum_file)
          File "scripts/lib/CIME/case/check_input_data.py", line 37, in _download_checksum_file
            success = server.getfile(rel_path, new_file)
          File "scripts/lib/CIME/Servers/wget.py", line 31, in getfile
            full_url = os.path.join(self._server_loc, rel_path)
          File "/usr/lib/python2.7/posixpath.py", line 75, in join
            if b.startswith('/'):
        AttributeError: 'NoneType' object has no attribute 'startswith'
```
This test passed before. Is this caused by missing input files requierd by latest CIME?
Looks like CIME is not handling the case where there is no file to download (None)?

### LINKED ISSUE #3034: Possible issues with inputdata checksum downloads
I have noticed two possible issues with the inputdata checksum implementation. I'm not sure if these are actually issues, but wanted to get @jedwards4b 's thoughts on these:

(1) I noticed that my tests were downloading the `inputdata_checksum.dat` file every time I ran `./case.submit`. I would have thought this download from the inputdata server would only happen if (a) you are actually downloading any files (e.g., the first time it goes to download a file, it downloads the checksum file from that server, too), or (b) you run `check_input_data` with the `--chksum` argument.

(2) In this loop:

https://github.com/ESMCI/cime/blob/e1779e35e9d62f2d1a4f66b13aa8a71d6e37cb45/scripts/lib/CIME/case/check_input_data.py#L140-L159

If I'm understanding the logic correctly, the loop will exit as soon as it downloads the first checksum file successfully - rather than, say, downloading the checksum file from each server. Is that actually what's intended here? (Or am I misunderstanding the logic?)

### LINKED ISSUE #3036: Having long wait time on case.submit
This is with cime5.6.12

I'm having long wait times on running case.submit on cheyenne today, with it eventually failing as we see below. It seems to be because of some type of network issue getting to svn inputdata from cheyenne. I'm wondering if we need another --skip option to skip checking the connection to inputdata?

```
   Calling /gpfs/fs1/scratch/erik/release-clm5.0.19/cime/src/components/stub_comps/swav/cime_config/buildnml
   Calling /gpfs/fs1/scratch/erik/release-clm5.0.19/cime/src/components/stub_comps/sesp/cime_config/buildnml
   Calling /gpfs/fs1/scratch/erik/release-clm5.0.19/cime/src/drivers/mct/cime_config/buildnml
Finished creating component namelists
Checking that inputdata is available as part of case submission
Client protocol gftp not enabled
Using protocol wget with user anonymous and passwd user@example.edu
ERROR: Could not connect to repo '{0}'
This is most likely either a proxy, or network issue .
```

jedwards4b:
Fix chksum issues - only download chksum file(s) if requested, look on all servers in config_inputdata.xml for chksum files and merge those found.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3033
Fixes #3034
Fixes #3036

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case/check_input_data.py
+++ b/scripts/lib/CIME/case/check_input_data.py
@@ -320,7 +320,6 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
                             no_files_missing = False
                         else:
                             logging.debug("  Found input file: '{}'".format(full_path))
-
                     else:
                         # There are some special values of rel_path that
                         # we need to ignore - some of the component models
--- a/scripts/lib/CIME/case/check_input_data.py
+++ b/scripts/lib/CIME/case/check_input_data.py
@@ -93,7 +93,7 @@ def _download_if_in_repo(server, input_data_root, rel_path, isdirectory=False):
     user is the user name of the person running the script
     isdirectory indicates that this is a directory download rather than a single file
     """
-    if not server.fileexists(rel_path):
+    if not (rel_path or server.fileexists(rel_path)):
         return False
 
     full_path = os.path.join(input_data_root, rel_path)
@@ -137,7 +137,8 @@ def check_all_input_data(self, protocol=None, address=None, input_data_root=None
             inputdata = Inputdata()
             protocol = "svn"
             success = False
-            while not success and protocol is not None:
+            # download and merge all available chksum files.
+            while protocol is not None:
                 protocol, address, user, passwd, chksum_file = inputdata.get_next_server()
                 if protocol not in vars(CIME.Servers):
                     logger.warning("Client protocol {} not enabled".format(protocol))
--- a/scripts/lib/CIME/case/check_input_data.py
+++ b/scripts/lib/CIME/case/check_input_data.py
@@ -13,7 +13,7 @@ logger = logging.getLogger(__name__)
 chksum_hash = dict()
 local_chksum_file = 'inputdata_checksum.dat'
 
-def _download_checksum_file(server, rundir, chksum_file):
+def _download_checksum_file(rundir):
     """
     Return True if successfully downloaded
     server is an object handle of type CIME.Servers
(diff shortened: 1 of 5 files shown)


jgfouca:
@jedwards4b , I'd like to have this fix in CIME master before I merge back to E3SM. Will you be doing a maint merge any time soon?

jedwards4b:
Yes we can do it as soon as this is merged.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -137,7 +137,8 @@ def check_all_input_data(self, protocol=None, address=None, input_data_root=None
             inputdata = Inputdata()
             protocol = "svn"
             success = False
-            while not success and protocol is not None:
+            # download and merge all available chksum files.
+            while protocol is not None:
```
billsacks: Since you no longer use the `success` value returned from the call to `_download_checksum_file`, you should probably change that to `_` to signal that the return value isn't used.

jedwards4b: fixed

billsacks: Thanks. I'll wait to approve until you answer my more general question.

jedwards4b:
@billsacks - fixed now.

jedwards4b:
@billsacks you are correct, I did miss a case.  Fixed now

PR #3042, opened 2019-03-13
fix issue with PET and ERP tests

### LINKED ISSUE #3041: compile_threaded issue on anlworkstation
In config/e3sm/machines/config_compilers.xml, -fopenmp is appended only when  compile_threaded
is set to true:
```
<compiler COMPILER="gnu">
  <CFLAGS>
    ...
    <append compile_threaded="true"> -fopenmp </append>
    ...
  </CFLAGS>
  ...
  <FFLAGS>
    ...
    <append compile_threaded="true"> -fopenmp </append>
    ...
  </FFLAGS>
  ...
  <LDFLAGS>
    <append compile_threaded="true"> -fopenmp </append>
  </LDFLAGS>
  ...
</compiler>
```
When cime_developer test suite is run on anlworkstation,
compile_threaded seems to be false by default. This has
caused two tests to fail during MODEL_BUILD: 
ERP.f45_g37_rx1.A
PET_P4.f19_f19.A

Errors were:
```
        perf_mod.F90:(.text+0x7): undefined reference to `omp_get_max_threads_'
    
        ...
        
        gptl.c:(.text+0x455): undefined reference to `omp_get_thread_num'

        ...
```
Should CIME set compile_threaded to true by default?
CIME should also fix the link errors even when compile_threaded is set to false

jedwards4b:
PET and ERP tests were not setting compile_threaded correctly.

Test suite: ERP.f45_g37_rx1.A.centos7-linux_gnu ,  PET_P4.f19_f19.A.centos7-linux_gnu 
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #3041 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -18,6 +18,7 @@ def get_standard_makefile_args(case):
                  "OCN_SUBMODEL", "CISM_USE_TRILINOS", "USE_ALBANY", "USE_PETSC"]
 
     make_args = "CIME_MODEL={} ".format(case.get_value("MODEL"))
+    make_args += " compile_threaded={} ".format(case.get_build_threaded())
     for var in variables:
         make_args+=xml_to_make_variable(case, var)
 
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -1328,7 +1328,7 @@ directory, NOT in this subdirectory."""
         Returns True if current settings require a threaded build/run.
         """
         force_threaded = self.get_value("FORCE_BUILD_SMP")
-        smp_present = bool(force_threaded) or self.thread_count > 1
+        smp_present = force_threaded or self.thread_count > 1
         return smp_present
 
     def _check_testlists(self, compset_alias, grid_name, files):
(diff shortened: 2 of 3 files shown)


jgfouca:
Thanks Jim. I'll copy this over to E3SM. Is there any way we can test this so that it doesn't break again?

jedwards4b:
It's tested in cdash nightly.  Other than that I could have been a little more careful about testing this tag on multiple platforms.

jgfouca:
> It's tested in cdash nightly.

OK, that's probably good enough. I guess we just didn't check the dashboard?

jedwards4b:
I think that @dqwu just got to it first.

dqwu:
@jedwards4b 
The fix looks OK, the two tests can pass now.

For src/build_scripts/buildlib.gptl, the follwing code is removed:
```
    debug = case.get_value("DEBUG")
    ...
    if debug:
        gmake_opts += " DEBUG=True"
```
This change seems irrelevant to this fix?

jedwards4b:
The DEBUG flag was added twice, this just removes one of them.

PR #3054, opened 2019-03-29
Jgfouca/branch for acme split 2019 03 29

jgfouca:
Change list:

Features:
* Implements remaining changes for gustiness feature for EAMv2
* Support multiqueueing for sandiatoss3
* Add a new 'B' test option to indicate a build-only test

Bug fixes:
* Remove dependence on "mask" variables in the mapping files that are input to the gen_domain tool.
* Fix some issues with downloading input files when checksum files are not set up
* Fix kokkos build, Make logic needs to happen after macro include or else kokkos flags are lost
* CIME fixes for PIO2
* **Use consistent casing (all-caps) for build_threaded value** 

Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


jgfouca:
Tests ongoing.

jgfouca:
Testing passed on melvin.

billsacks:
> * Add a new 'B' test option to indicate a build-only test

This is intriguing. Can you give an example illustrating what you mean by this?

jedwards4b:
@mnlevy1981 Could you please review changes to the gen_domain tools?

rljacob:
@mnlevy1981 the original PR for the gen_domain changes is https://github.com/E3SM-Project/E3SM/pull/2633

tools/mapping/gen_domain_files/src/gen_domain.F90
```diff
@@ -38,20 +38,26 @@ program fmain
   character(LEN=512) :: usercomment ! user comment
   character(LEN= 8)  :: cdate       ! wall clock date
   character(LEN=10)  :: ctime       ! wall clock time
+  real(r8) :: fminval = 0.001_r8    ! min allowable land fraction; frac set to 0 if frac < fminval
```
mnlevy1981: It seems like the convention is to declare variables here and set default values below (e.g. how `set_fv_pole_yc` is handled)

tools/mapping/gen_domain_files/src/gen_domain.F90
```diff
@@ -38,20 +38,26 @@ program fmain
...
+  real(r8) :: fmaxval = 1.000_r8    ! max allowable land fraction; frac set to 1 if frac > fmaxval
```
mnlevy1981: Do we have access to the global parameter `c1` here?

tools/mapping/gen_domain_files/src/gen_domain.F90
```diff
@@ -38,20 +38,26 @@ program fmain
...
+  logical :: set_omask = .false.    ! set ocn mask if not present in input mapping file
   !----------------------------------------------------
 
+  ! Initialize options before parsing command line arguments
```
mnlevy1981: Let's initialize the new options here

tools/mapping/gen_domain_files/src/gen_domain.F90
```diff
@@ -470,14 +514,19 @@ end subroutine gen_domain
 
 !===========================================================================
 
-  subroutine check_ret(ret)
+  subroutine check_ret(ret, fatal)
     ! Check return status from netcdf call
     implicit none
     integer, intent(in) :: ret
+    logical, intent(in), optional :: fatal
+    logical :: fatal_local = .true.
```
mnlevy1981: As above, I think you want to declare `fatal_local` and set it to `.true.` in two calls. In this case, my understanding is that setting the variable in the declaration like this implicitly adds the `save` attribute, so in the block

```
call check_ret(ret1, fatal=.false.)
call check_ret(ret2)
```

The second call is treated as `fatal = .false.` which is not the desired behavior.

billsacks:
@jgfouca cool, thanks - this seems like something we could use for cesm, too.

jgfouca:
@brhillman , could you reply to @mnlevy1981 regarding the gen_domain changes?

tools/mapping/gen_domain_files/src/gen_domain.F90
brhillman: I can change this. I think I put these here because originally they were hard-coded as magic numbers elsewhere in the code and my first attempt was to just change them to parameter values here.

tools/mapping/gen_domain_files/src/gen_domain.F90
brhillman: `c1` is local to subroutine `gen_domain`, so no. Regardless, if we move this to the body of the program as you suggest above, I'd be inclined to set this as `fmaxval = 1` (no decimal or kind postfix) which will set `fmaxval` to the full precision allowed by the type `r8`. That is, I think using `c1` is unnecessary here.

tools/mapping/gen_domain_files/src/gen_domain.F90
brhillman: Good catch. This should be fixed.

tools/mapping/gen_domain_files/src/gen_domain.F90
brhillman: This should be fixed (also moved `set_omask` initialization for consistency).

tools/mapping/gen_domain_files/src/gen_domain.F90
brhillman: Done.

tools/mapping/gen_domain_files/src/gen_domain.F90
brhillman: Fixed.

Issue #119, opened 2016-03-26
compset matches in driver's config_component should be made more general


billsacks:
Moving issue CESM-Development/cime#402 to here, in the comments that follow.

billsacks:
Many of the compset matches in driver_cpl/cime_config/config_component.xml seem fragile, especially if we want this same file to work for a variety of active components. Examples are the various NCPL settings and CPL_SEQ_OPTION.

It feels like there references here to specific component names should mostly be eliminated (although it looks like they may be needed in some cases - e.g., to distinguish between the desired coupling frequency of POP vs. MPAS). I'm not sure what would be the best way to solve this problem, but one idea is to have variables like LND_COMP_TYPE saying whether (for example) the land model is active, data, stub or coupler_test - and same for OCN, ATM, GLC, ICE, etc. Then buggy(?) rules like this (for OCN_NCPL):

      <value compset="_DATM.*_CLM4.*_SICE.*_SOCN">1</value>

could be written as

ATM_COMP_TYPE="data" LND_COMP_TYPE="active" ICE_COMP_TYPE="stub" OCN_COMP_TYPE="stub"

billsacks:
Things like this also concern me (for CCSM_BGC):

```
      <value compset="HIST.*_DATM%(QIA|CRU)">CO2A</value>
      <value compset="RCP.*_DATM%(QIA|CRU)" >CO2A</value>
```

What if someone introduces a new DATM mode? It's so easy to miss the fact that you need to update the driver's config_component for things like this.

I often wish that we didn't have the ability to have regular expression matches against compset at all in these xml files, but instead required that components set various pieces of metadata that could be queried for these purposes.

billsacks:
In talking this through with @mvertens, we feel that there are a few things going on here that should be addressed at some point. This isn't critical, but will be important to address if we want the driver to be more component-independent:

- We should get rid of regular expression matching in the xml files, since this is error-prone. Instead, the compset long name should be split into pieces, and then you can match against those individual pieces (e.g., ATM_NAME, ATM_OPTIONS, etc.).

- We should introduce general variables like ATM_TYPE, LND_TYPE, etc., which would have the possible values active, stub, data, coupler_test. Where possible, we should match on those general types. If (for example) you want to do some setting when running with active land, the current workaround is to have a regex match like ^(DLND|SLND|XLND), but that is error-prone. (I'm not at all confident that I just wrote that regex correctly.)

[...]

PR #3055, opened 2019-04-01
Introduces a new stub IAC

bishtgautam:
Adds a stub for Integrated Assessment Component (IAC).

Test suite: scripts_regression_tests.py for e3sm and cesm
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: jedwards

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -544,32 +544,30 @@ class Case(object):
         ('2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV_SESP', ['2000', 'DATM%NYF', 'SLND', 'DICE%SSMI', 'DOCN%DOM', 'DROF%NYF', 'SGLC', 'SWAV', 'SESP'])
         >>> Case(read_only=False)._valid_compset_impl('2000_DICE%SSMI_DOCN%DOM_DATM%NYF_DROF%NYF', ['CPL', 'ATM', 'LND', 'ICE', 'OCN', 'ROF', 'GLC', 'WAV', 'ESP'], {'datm':1,'satm':1,'dlnd':2,'slnd':2,'dice':3,'sice':3,'docn':4,'socn':4,'drof':5,'srof':5,'sglc':6,'swav':7,'ww3':7,'sesp':8})
         ('2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV_SESP', ['2000', 'DATM%NYF', 'SLND', 'DICE%SSMI', 'DOCN%DOM', 'DROF%NYF', 'SGLC', 'SWAV', 'SESP'])
-        >>> Case(read_only=False)._valid_compset_impl('1850_CAM60_CLM50%BGC-CROP_CICE_POP2%ECO%ABIO-DIC_MOSART_CISM2%NOEVOLVE_WW3', ['CPL', 'ATM', 'LND', 'ICE', 'OCN', 'ROF', 'GLC', 'WAV', 'ESP'], {'datm':1,'satm':1, 'cam':1,'dlnd':2,'clm':2,'slnd':2,'cice':3,'dice':3,'sice':3,'pop':4,'docn':4,'socn':4,'mosart':5,'drof':5,'srof':5,'cism':6,'sglc':6,'ww':7,'swav':7,'ww3':7,'sesp':8})
-        ('1850_CAM60_CLM50%BGC-CROP_CICE_POP2%ECO%ABIO-DIC_MOSART_CISM2%NOEVOLVE_WW3_SESP', ['1850', 'CAM60', 'CLM50%BGC-CROP', 'CICE', 'POP2%ECO%ABIO-DIC', 'MOSART', 'CISM2%NOEVOLVE', 'WW3', 'SESP'])
+        >>> Case(read_only=False)._valid_compset_impl('2000_DICE%SSMI_DOCN%DOM_DATM%NYF_DROF%NYF_TEST', ['CPL', 'ATM', 'LND', 'ICE', 'OCN', 'ROF', 'GLC', 'WAV', 'ESP'], {'datm':1,'satm':1,'dlnd':2,'slnd':2,'dice':3,'sice':3,'docn':4,'socn':4,'drof':5,'srof':5,'sglc':6,'swav':7,'ww3':7,'sesp':8})
+        ('2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV_SESP_TEST', ['2000', 'DATM%NYF', 'SLND', 'DICE%SSMI', 'DOCN%DOM', 'DROF%NYF', 'SGLC', 'SWAV', 'SESP'])
+        >>> Case(read_only=False)._valid_compset_impl('1850_CAM60_CLM50%BGC-CROP_CICE_POP2%ECO%ABIO-DIC_MOSART_CISM2%NOEVOLVE_WW3_BGC%BDRD', ['CPL', 'ATM', 'LND', 'ICE', 'OCN', 'ROF', 'GLC', 'WAV', 'ESP'], {'datm':1,'satm':1, 'cam':1,'dlnd':2,'clm':2,'slnd':2,'cice':3,'dice':3,'sice':3,'pop':4,'docn':4,'socn':4,'mosart':5,'drof':5,'srof':5,'cism':6,'sglc':6,'ww':7,'swav':7,'ww3':7,'sesp':8})
+        ('1850_CAM60_CLM50%BGC-CROP_CICE_POP2%ECO%ABIO-DIC_MOSART_CISM2%NOEVOLVE_WW3_SESP_BGC%BDRD', ['1850', 'CAM60', 'CLM50%BGC-CROP', 'CICE', 'POP2%ECO%ABIO-DIC', 'MOSART', 'CISM2%NOEVOLVE', 'WW3', 'SESP'])
+        >>> Case(read_only=False)._valid_compset_impl('1850_CAM60_CLM50%BGC-CROP_CICE_POP2%ECO%ABIO-DIC_MOSART_CISM2%NOEVOLVE_WW3_BGC%BDRD_TEST', ['CPL', 'ATM', 'LND', 'ICE', 'OCN', 'ROF', 'GLC', 'WAV', 'IAC', 'ESP'], {'datm':1,'satm':1, 'cam':1,'dlnd':2,'clm':2,'slnd':2,'cice':3,'dice':3,'sice':3,'pop':4,'docn':4,'socn':4,'mosart':5,'drof':5,'srof':5,'cism':6,'sglc':6,'ww':7,'swav':7,'ww3':7,'sesp':8})
+        ('1850_CAM60_CLM50%BGC-CROP_CICE_POP2%ECO%ABIO-DIC_MOSART_CISM2%NOEVOLVE_WW3_SIAC_SESP_BGC%BDRD_TEST', ['1850', 'CAM60', 'CLM50%BGC-CROP', 'CICE', 'POP2%ECO%ABIO-DIC', 'MOSART', 'CISM2%NOEVOLVE', 'WW3', 'SIAC', 'SESP'])
         """
         # Find the models declared in the compset
         model_set = [None]*len(comp_classes)
         components = compset_name.split('_')
         model_set[0] = components[0]
-        # Check for noncomponent appends (BGC & TEST)
         noncomps = []
-        for noncomp in ('BGC', 'TEST'):
-            if components[-1][0:len(noncomp)] == noncomp:
-                noncomps.append(components[-1])
-                last_ind = len(components) - 1
-            else:
-                last_ind = len(components)
-
-
-
-        for model in components[1:last_ind]:
+        for model in components[1:]:
             match = Case.__mod_match_re__.match(model.lower())
             expect(match is not None, "No model match for {}".format(model))
             mod_match = match.group(1)
-            expect(mod_match in comp_hash,
-                   "Unknown model type, {}".format(model))
-            comp_ind = comp_hash[mod_match]
-            model_set[comp_ind] = model
+            # Check for noncomponent appends (BGC & TEST)
+            if mod_match in ('bgc', 'test'):
+                noncomps.append(model)
+            else:
+                expect(mod_match in comp_hash,
+                       "Unknown model type, {}".format(model))
+                comp_ind = comp_hash[mod_match]
+                model_set[comp_ind] = model
 
         # Fill in missing components with stubs
         for comp_ind in range(1, len(model_set)):
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -544,8 +544,8 @@ class Case(object):
         ('2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV_SESP', ['2000', 'DATM%NYF', 'SLND', 'DICE%SSMI', 'DOCN%DOM', 'DROF%NYF', 'SGLC', 'SWAV', 'SESP'])
         >>> Case(read_only=False)._valid_compset_impl('2000_DICE%SSMI_DOCN%DOM_DATM%NYF_DROF%NYF', ['CPL', 'ATM', 'LND', 'ICE', 'OCN', 'ROF', 'GLC', 'WAV', 'ESP'], {'datm':1,'satm':1,'dlnd':2,'slnd':2,'dice':3,'sice':3,'docn':4,'socn':4,'drof':5,'srof':5,'sglc':6,'swav':7,'ww3':7,'sesp':8})
         ('2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV_SESP', ['2000', 'DATM%NYF', 'SLND', 'DICE%SSMI', 'DOCN%DOM', 'DROF%NYF', 'SGLC', 'SWAV', 'SESP'])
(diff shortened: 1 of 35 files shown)


bishtgautam:
I have opened this PR to start the discussion. Presently, this PR brings all CIME related changes that were made in E3SM to introduce the stub IAC, but it required adding `_SIAC` to all compset definitions. I will next try to make changes in python script so `_SIAC` is an optional descriptor.

ekluzek:
Making IAC optional sounds great. Thanks for working on that. We in cesm will really need it as optional.

rljacob:
Why, exactly, does it have to be optional in CESM?

gold2718:
Optional components allow us to introduce new component types without having to go back and edit every other component set in every branch of every model.
There is currently one optional component, the ESP component type.
The ESP component type is run by the CIME driver but does not interact through the coupler (ESP models uses files to modify the state of other models in the system). If that describes the IAC system, it may be easier to use the infrastructure already built to support ESP models.

As for the optional bit, since there is currently only one optional component, the logic to implement it is overly simple (see lines 636 and 637 in case.py). Adding a new component type (i.e., if ESP does not meet the needs of the IAC component) will require a bit more logic but should be possible.

bishtgautam:
Here is the pseudo-code that I was considering adding in [case.py#L791](https://github.com/ESMCI/cime/blob/master/scripts/lib/CIME/case/case.py#L791)

```
        compset_alias, science_support = self._set_compset(compset_name, files, driver)

        # New code:
        if get_model() == "e3sm":
            If self._compsetname does not include {'_SIAC' or '_GCAM'}:
               self._compsetname = self._compsetname + '_SIAC'

```

To go with the above change, I was wondering if the code implemented in `_get_description_match` ([component.py#L209](https://github.com/ESMCI/cime/blob/master/scripts/lib/CIME/XML/component.py#L209)) could be implemented such that it doesn't require a component object and I can check if the compsetname includes `'_SIAC'` or `'_GCAM'`

rljacob:
Note:  GCAM is the real model we will eventually use in place of the stub for some compsets.

bishtgautam:
IAC will interact with the atmosphere and land components. Unlike the iESM project, we (e3sm) now wants IAC to interact with other components via coupler.

Btw, can't we consider if the long name of a composet does not include a component, it is a stub component? If we agree on it, the proposed additional check `if get_model() == "e3sm":` can also be omitted.

src/drivers/mct/cime_config/config_compsets.xml
```diff
@@ -24,6 +24,7 @@
     ROF  = [DROF, SROF, XROF]
     GLC  = [      SGLC      ]
     WAV  = [DWAV, SWAV      ]
+    IAC  = [DWAV, SIAC      ]
```
jedwards4b: typo  here

ekluzek:
@rljacob @gold2718 gave the primary reason for this, requiring changes to all compsets, on all branches. The other aspect of that is that you then need to test all those compsets, to make sure a mistake wasn't made. So that's a pretty huge requirement to the project as a whole. 

I'll let @mvertens and @billsacks also weigh in here. They also want to talk about this more at the Wednesday cime telecon. If we can introduce changes in a way that are backwards compatible, it often can save work over the entire project. Changes that aren't backwards compatible need more discussion and involvement of everyone across the project (and both cesm and e3sm so both projects).

src/drivers/mct/cime_config/config_compsets.xml
bishtgautam: fixed

rljacob:
You would only have to do "--no-setup" testing.  Its a BFB change.  Just have to make sure every compset was indeed edited correctly.

gold2718:
Here is an idea for making IAC an optional component along with ESP. Replace the current lines 636 and 637 with:
```
        opt_components = ['IAC', 'ESP']
        if len(self._component_classes) > len(self._components):
            # Find which optional components are missing
            for comp in opt_components:
                opt_models = files.get_components('COMP_ROOT_DIR_' + comp)
                comp_ok = False
                for model in opt_models:
                    if model in self._components:
                        comp_ok = True
                        break

                if not comp_ok:
                    self._components.append('s{}'.format(comp.lower()))
                    self._compsetname = self._compsetname + '_S{}'.format(comp)
```
This should probably be cleaned up to preserve the ordering of components but I at least wanted some feedback on the general idea.

bishtgautam:
@gold2718, I tried you suggestion but the value of `opt_models` in `opt_models = files.get_components('COMP_ROOT_DIR_' + comp)` is `None`. Is `opt_models` expected to contain names of the active/data/stub/etc model for the optional component?

gold2718:
@bishtgautam, I think that is because you did not add a `CONFIG_IAC_FILE` section to `<root>/config/e3sm/config_files.xml`

rljacob:
Note:  See meeting notes for discussion on how we will add this feature:  https://github.com/ESMCI/cime/wiki/Meeting-Notes#apr-3-2019

shippert:
I may be a little late to the discussion, but this is how I was implementing the mods to case.py to add an optional siac component:

```
        optional_stubs = ['sesp', 'siac']
        nmiss=len(self._component_classes) - len(self._components) - 1
        if nmiss > 0: self._components += optional_stubs[-nmiss:]
```
This is a little more obfuscated than the above solution, and it probably needs another upper limit check on nmiss to be completely robust, but it does enforce the component order, and adding a future optional stub would be simply a matter of adding it to the list.

gold2718:
@klindsay28 points out that the proposed solution (along with the solution from @shippert) has a flaw in that it will cause some pattern matching to break. For example, in config_compsets.xml, there are items like:
```
    <entry id="RUN_REFDATE">
      <values match="first">
        <value grid="a%0.9x1.25_l%0.9x1.25_oi%gx1v7_r%r05_g%gland4_w%ww3a_m%gx1v7"      compset="1850_CAM60_CLM50%BGC-CROP.*_CICE.*_POP2%ECO.*_MOSART_CISM2%NOEVOLVE_WW3_BGC%BDRD">0134-01-01</value>
[ . . . ]
      </values>
    </entry>

```
If stub components are inserted into the compset, the search pattern will fail because there will be one or more stub components between `WW3` and `BGC%BDRD`.
This is not happening now due to a flaw in the current logic for adding SESP (same counting problem that is in the @shippert code above).

Solution will have to be expanded to deal with this issue (or require modifying a lot of existing patterns which is not backwards compatible).

billsacks:
Good point @klindsay28 . I know this is not helpful in the short-term, but I'll take this opportunity to complain again about our fragile uses of compset regex matching all over the place, and point out #119 .

gold2718:
I appreciate @billsacks comment but I do not currently have a 'short-term' solution. The only hack I can think of is to remove 'optional' stub components from the compset name before attempting to match. This could be done by deciding on the 'non-optional' set (i.e., SATM, SLND, SOCN, SICE, SGLC, SROF, SWAV) and eliminating any other component beginning with S. I would just remove all the stub components from the compset but both CESM and E3SM have hundreds of compset match patterns which explicitly mention stub components.

rljacob:
So with the merge of https://github.com/ESMCI/cime/pull/3068 to master (thanks @gold2718 !) can this PR now be merged?  Does it require any more changes?

src/drivers/mct/cime_config/config_compsets.xml
```diff
@@ -40,32 +41,32 @@
 
   <compset>
     <alias>A</alias>
-    <lname>2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV</lname>
+    <lname>2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV_SIAC</lname>
```
gold2718: These stub components are no longer necessary, remove them?

src/drivers/mct/cime_config/config_compsets.xml
```diff
@@ -40,32 +41,32 @@
...
   </compset>
 
   <compset>
     <alias>ADSOM</alias>
-    <lname>2000_DATM%NYF_SLND_DICE%SSMI_DOCN%SOM_DROF%NYF_SGLC_SWAV_TEST</lname>
+    <lname>2000_DATM%NYF_SLND_DICE%SSMI_DOCN%SOM_DROF%NYF_SGLC_SWAV_SIAC_TEST</lname>
```
gold2718: Does anyone know what `_TEST` does in the compset name?
Any reason we shouldn't just get rid of it? @jedwards4b? @jgfouca?

jedwards4b:
I think you would need to rebase or merge to master and then run tests - are the changes here consistent with what was done in #3068 @gold2718?  Note that the initial comment in this PR says nothing about what testing was done here - @bishtgautam please update your comment to indicate what testing has been done.

src/drivers/mct/cime_config/config_compsets.xml
billsacks: In some cases (I'm not sure about this one), `_TEST` reduces the required input data volume (e.g., using just one or a few years rather than the full time series).

gold2718: I was wondering about this because I cannot find any usage of the item. If it is used, I need to fix my latest PR (I did not know about this and testing did not catch it). If it is not used, I think we should get rid of cases that use it.

src/drivers/mct/cime_config/config_compsets.xml
rljacob: I think at least one of the shared compsets should have it.  Just so we know its there.

src/drivers/mct/cime_config/config_compsets.xml
jedwards4b: @billsacks Is thinking of modifiers to the compset name like DTEST - that's different than a component called _TEST - as far as I know this isn't used anywhere in CIME.

bishtgautam:
I'm going to rebase this branch and address the comments for `src/drivers/mct/cime_config/config_compsets.xml`

billsacks:
Here are uses of `_TEST` in compset long names in CESM (I have removed a few results from the grep that were different uses of TEST:

```
[~/cesm_code/cesm]$ rgrep -n --include='config_compsets.xml' TEST .
./cime_config/config_compsets.xml:210:    <alias>ETEST</alias>
./cime_config/config_compsets.xml:211:    <lname>2000_CAM60_CLM50%SP_CICE_DOCN%SOM_MOSART_SGLC_SWAV_TEST</lname>
./cime_config/config_compsets.xml:215:    <alias>E1850L45TEST</alias>
./cime_config/config_compsets.xml:216:    <lname>1850_CAM60_CLM45%SP_CICE_DOCN%SOM_MOSART_SGLC_SWAV_TEST</lname>
./components/cice/cime_config/config_compsets.xml:52:    <alias>DTEST</alias>
./components/cice/cime_config/config_compsets.xml:53:    <lname>2000_DATM%NYF_SLND_CICE_DOCN%SOM_DROF%NYF_SGLC_SWAV_TEST</lname>
./cime/config/e3sm/allactive/config_compsets.xml:227:  <alias>ETEST</alias>
./cime/config/e3sm/allactive/config_compsets.xml:228:  <lname>2000_CAM4_CLM40%SP_CICE_DOCN%SOM_MOSART_SGLC_SWAV_TEST</lname>
./cime/config/e3sm/allactive/config_compsets.xml:237:  <alias>E1850C5TEST</alias>
./cime/config/e3sm/allactive/config_compsets.xml:238:  <lname>1850_CAM5_CLM40%SP_CICE_DOCN%SOM_MOSART_SGLC_SWAV_TEST</lname>
./cime/config/e3sm/allactive/config_compsets.xml:252:  <alias>E1850C5CNTEST</alias>
./cime/config/e3sm/allactive/config_compsets.xml:253:  <lname>1850_CAM5_CLM40%CN_CICE_DOCN%SOM_MOSART_SGLC_SWAV_TEST </lname>
./cime/src/drivers/mct/cime_config/config_compsets.xml:48:    <lname>2000_DATM%NYF_SLND_DICE%SSMI_DOCN%SOM_DROF%NYF_SGLC_SWAV_TEST</lname>
```

@dabail10 can you speak to these?

There are also these in CLM, but I'm pretty sure they're no longer used:

```
./components/clm/cime_config/config_compsets.xml:399:	<value compset="CLM[45].*_CISM.*_TEST"	  >1980-01-15</value>
./components/clm/cime_config/config_compsets.xml:400:	<value compset="CLM[45]%[^_]*CNDV.*_TEST" >1997-12-31</value>
```

gold2718:
> Why are you adding this to config/cesm? I though only config/e3sm needs it.

Because, currently, `COMP_CLASSES` is in the shared config_component.xml. The new `valid_compset` code is going to go looking for supported IAC models for both CESM and E3SM. An alternative (mentioned in #3068) is to move `COMP_CLASSES` to config_component_e3sm.xml and to config_component_cesm.xml so that the supported component classes can be different.

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -11,7 +11,7 @@
 
   <entry id="COMP_CLASSES">
     <type>char</type>
-    <default_value>CPL,ATM,LND,ICE,OCN,ROF,GLC,WAV,ESP</default_value>
+    <default_value>CPL,ATM,LND,ICE,OCN,ROF,GLC,WAV,IAC,ESP</default_value>
```
jedwards4b: Please move this field to config_component_e3sm.xml and config_component_cesm.xml and remove the changes from config/cesm/*

rljacob: Are you sure that is the right thing to do?  The system will have to build a stub IAC and link against it because of the coupler so might as well maintain a common COMP_CLASSES.

jedwards4b:
@bishtgautam The test  test_m_createnewcase_alternate_drivers (__main__.J_TestCreateNewcase) is failing with ERROR: No variable IAC_NX found in case - in this case we are not using the mct driver and so should not need an IAC stub.   I have a fix, can I push it here?  You will need to give me permission in your repo

jedwards4b:
It seems that iac%present is true, it should be false for a stub compset.  Does this case:
SMS_Ln9_P1.f19_g16_rx1.A run for e3sm?

bishtgautam:
I will run the SMS_Ln9_P1.f19_g16_rx1.A with e3sm and report back.

bishtgautam:
@jedwards4b, The reason why SMS_Ln9_P1.f19_g16_rx1.A is failing is:

- In [cime_comp_mod.F90#L1321](https://github.com/ESMCI/cime/pull/3055/files#diff-87e885f8a49b11dfabed6af2b99c02f5R1321), the IAC component is initialized as 
`call component_init_pre(iac, IACID, CPLIACID, CPLALLIACID, infodata, ntype='iac') ` .
- Thus, the `comp(1)%oneletterid` for IAC is `i`, which is **same as for the ICE component**.
- The code in [component_mod.F90#L170](https://github.com/ESMCI/cime/pull/3055/files#diff-1dcc630d470ba94b028798d74365031bR170) was expecting IAC's `oneletterid` to be `z`.

A kludge to avoid the error is to set `oneletterid` for IAC differently:
```
       select case(ntype)
       case ('atm','cpl','ocn','wav','glc','ice','rof','lnd','esp')
          comp(eci)%oneletterid =  ntype(1:1)
       case ('iac')
          comp(eci)%oneletterid = 'z'
       case default
          call shr_sys_abort(subname//': component_ntype, "'//component_ntype//'" not recognized"')
       end select
```
I acknowledge that the kludge is unsatisfactory as the `oneletterid` should be the first letter of `ntype` and I'm open to alternate fixes which could be:
1. Renaming IAC to start with a letter that is not taken
2. A better solution to the above-mentioned kludge
3. Other fixes

cc: @kvcalvin, @shippert

scripts/lib/CIME/case/case.py
```diff
@@ -577,8 +580,8 @@ def _valid_compset_impl(self, compset_name, comp_classes, comp_hash):
                 model_set[comp_ind] = stub
 
         # Return the completed compset
-        if bgc is not None:
-            model_set.append(bgc)
+        for noncomp in noncomps:
+            model_set.append(noncomp)
```
gold2718: I realize that this follows my previous methodology but I am not sure that BGC or TEST belong in the model set.
I think they should just be appended to the compset string.
Also, I thought there was no use of the TEST appendage, why not just eliminate it from all compsets?

jedwards4b: There is a usage for DTEST and other compsets which want to minimize inputdata requirements for testing.  I am open to a better long term solution if you have one.

src/drivers/mct/cime_config/config_component.xml
bishtgautam: We agreed that IAC field should remain in the shared coupler XML file.

jedwards4b:
@bishtgautam I am okay with the 'z' designation for iac @gold2718 @billsacks what do you think?

shippert:
@bishtgautam At this point IAC and 'z' are pretty well embedded in the changes I made, so it would be at least a minor nightmare to hunt them down change them to something else.  Your kludge may be inelegant, but it's also pretty unintrusive.

gold2718:
I am not sure how 'inelegant' it is but I would say that something like it was inevitable with only 26 letters vs. at least 17576 possible component class types.

rljacob:
Isn't the one letter id internal to CIME?  It doesn't need to make sense to a general user.  Just needs to be documented internally.  "z" is fine.

scripts/lib/CIME/case/case.py
gold2718: I think we should take out these two lines (but see below).

scripts/lib/CIME/case/case.py
```diff
@@ -577,8 +580,8 @@ def _valid_compset_impl(self, compset_name, comp_classes, comp_hash):
...
 
         compsetname = '_'.join(model_set)
```
gold2718: Add the noncomps directly to the compset name.
```
        for noncomp in noncomps:
            compsetname = compsetname + '_' + noncomp
```

billsacks:
> @bishtgautam I am okay with the 'z' designation for iac @gold2718 @billsacks what do you think?

No strong opinions. I'm okay with it if others are.

billsacks:
> @bishtgautam I am okay with the 'z' designation for iac @gold2718 @billsacks what do you think?

No strong opinions. I'm okay with it if others are.

> Isn't the one letter id internal to CIME? It doesn't need to make sense to a general user. Just needs to be documented internally. "z" is fine.

At a glance, it looks like one of the uses of this is to set prefixes for timer output, which is something that's visible to users.

jedwards4b:
I am still investigating a failing test PET_P4.f19_f19.A.cheyenne_intel

jedwards4b:
With my latest update I think I have all tests passing, rerunning now.  @gold2718 can you approve this now?

gold2718:
@jedwards4b, why did you take out the BGC tests?

gold2718:
I am going to push a fix in a bit.

jedwards4b:
@bishtgautam ready to merge?

bishtgautam:
Yes, it is ready to go from my end.

bishtgautam:
Thanks to @jedwards4b for testing and fixing this PR; @gold2718 for the PR-#3068 to avoid renaming all existing compsets; and everyone else who participated in telecon discussions and commented here. 

Now we an example of how to add a new **optional** stub component in CIME, let everyone go wild and bring their new components within CESM and E3SM!!!

kvcalvin:
Thanks everyone!

PR #3060, opened 2019-04-03
CMake for Jim Foucar

asteyer:
This is the CIME CMake based on an old stale CIME branch for Jim Foucar who has happily volunteered to work on the CMake build/test system.


jgfouca:
@asteyer , you need to change the target from master to that common-ancestor branch.

PR #3063, opened 2019-04-04
Add NLDAS grid for CTSM and MOSART

billsacks:
Adds a high-resolution regional grid over the U.S. for use in land-only
simulations.

(I haven't added the necessary files to the inputdata repository yet,
but will do so once this and the corresponding CTSM PRs are accepted.)

Test suite: SMS test using this new grid in the context of a CTSM branch
   with the necessary changes
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes none

User interface changes?: N

Code review:

### DIFF
--- a/config/cesm/config_grids.xml
+++ b/config/cesm/config_grids.xml
@@ -122,11 +122,11 @@
 
     <!-- Regional NLDAS-2 grid over the U.S. (0.125 degree resolution;
          25-53N, 235-293E), with mask from NLDAS-2 atmospheric drivers -->
-    <model_grid alias="nldas2_mnldas2" compset="DATM.+CLM">
+    <model_grid alias="nldas2_rnldas2_mnldas2" compset="DATM.+CLM">
       <grid name="atm">0.125nldas2</grid>
       <grid name="lnd">0.125nldas2</grid>
       <grid name="ocnice">0.125nldas2</grid>
-      <grid name="rof">null</grid>
+      <grid name="rof">0.125nldas2</grid>
       <mask>nldas2</mask>
     </model_grid>
 
@@ -1103,6 +1103,7 @@
       <desc>5x5 Amazon regional case -- only valid for DATM/CLM compset</desc>
     </domain>
 
+    <!-- This grid is also used by ROF -->
     <domain name="0.125nldas2">
       <nx>464</nx> <ny>224</ny>
       <file grid="atm|lnd" mask="nldas2">$DIN_LOC_ROOT/share/domains/domain.clm/domain.lnd.0.125nldas2_0.125nldas2.190410.nc</file>
--- a/config/cesm/config_grids.xml
+++ b/config/cesm/config_grids.xml
@@ -121,7 +121,7 @@
     </model_grid>
 
     <!-- Regional, 12km NLDAS grid over the U.S., with the standard NLDAS mask -->
-    <model_grid alias="224x464_nldas_mnldas" compset="DATM.+CLM">
+    <model_grid alias="nldas_mnldas" compset="DATM.+CLM">
       <grid name="atm">224x464_nldas</grid>
       <grid name="lnd">224x464_nldas</grid>
       <grid name="ocnice">224x464_nldas</grid>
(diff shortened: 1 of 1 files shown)


config/cesm/config_grids.xml
```diff
@@ -120,6 +120,15 @@
       <grid name="rof">null</grid>
     </model_grid>
 
+    <!-- Regional NLDAS grid over the U.S. (roughly 12km resolution), with the standard NLDAS mask -->
+    <model_grid alias="nldas_mnldas" compset="DATM.+CLM">
```
billsacks: Does this seem like a reasonable name for the alias? I struggled a bit, because the mask has the same name as the grid: we're using the standard NLDAS mask with the NLDAS grid.

For many grid aliases, we also give the ocean resolution, but that seems irrelevant here, and would lead to even more redundancy in the name (`nldas_nldas_mnldas`).

But I'm very open to suggestions here to keep this consistent with any grid naming conventions.

config/cesm/config_grids.xml
```diff
@@ -120,6 +120,15 @@
...
+      <grid name="atm">224x464_nldas</grid>
+      <grid name="lnd">224x464_nldas</grid>
+      <grid name="ocnice">224x464_nldas</grid>
+      <grid name="rof">null</grid>
```
billsacks: I'm using the same mechanism we have used elsewhere to turn off ROF for this regional grid: setting the rof grid to null.

config/cesm/config_grids.xml
jedwards4b: why nldas instead of 12km or nl12?

billsacks: Good idea, @jedwards4b . Though a couple of thoughts on this are:
- I prefer spelling things out a bit more, so would probably go with `nldas12_mnldas` or `nldas12km_mnldas`
- Do you think that including `12` or `12km` in the name gives the impression that the grid is at exactly 12 km (which isn't true)? This grid is actually at 1/8 deg – so roughly, but not exactly, 12 km.

@barlage I'd also welcome any opinions you have on this.

config/cesm/config_grids.xml
```diff
@@ -120,6 +120,15 @@
       <grid name="rof">null</grid>
     </model_grid>
 
+    <!-- Regional NLDAS grid over the U.S. (roughly 12km resolution), with the standard NLDAS mask -->
```
barlage: Suggestions: 

Regional NLDAS-2 grid over the U.S. (0.125 degree resolution; 25-53N, 235-293E), with the standard NLDAS-2 mask

config/cesm/config_grids.xml
barlage: @billsacks My preference would be to keep this as `nldas_mnldas` and leave the details in the grid definition below; NLDAS is a well-defined grid but another consideration would be to add an NLDAS version here (e.g. nldas2) since I believe NLDAS v3 may have a different, higher resolution grid when released. Is there any need to explicitly declare this as regional in the model_grid?

config/cesm/config_grids.xml
```diff
@@ -120,6 +120,15 @@
...
+    <model_grid alias="nldas_mnldas" compset="DATM.+CLM">
+      <grid name="atm">224x464_nldas</grid>
```
barlage: is it too much to add the resolution here, e.g., `224x464_0.125_nldas2`?

billsacks:
Based on comments today, I need to rework this. I'll comment when it's ready for review.

billsacks:
I have updated this PR with new grid and mask names based on discussions with @barlage and @mvertens . I plan to make files using this new naming convention later today, so please speak up soon if you have problems with these names.

After that, I'll still need to:
- Make new domain files using the new mask definition that we discussed on Thursday (using the mask of the NLDAS atmospheric forcing data, so it will include lakes)
- Add grid and mapping files for a corresponding 1/8 deg regional ROF grid

billsacks:
Okay, this is ready for final review. @mvertens @jedwards4b @barlage do you want to look at this one last time, with particular focus on the grid names (full grid name and alias)?

Note that the ROF grid is identical to the LND grid, so no mapping files are needed.

billsacks:
@mvertens gave her approval by email.

PR #3068, opened 2019-04-05
Support optional components

### LINKED ISSUE #3067: Optional stub components not supported in compset name
While the ESP component type is optional, it is not handled in a uniform manner. In addition, there is no support for new optional components.
Stub components should be optional in a compset long name.

gold2718:
Support optional components by filling in stub models for any missing component class

Test suite: scripts_regression_tests.py + many tests of running `create_newcase` with mangled compset names (both on command line and by modifying `<lname>` field in `<compset>` definition). E.g.:
```
2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV
2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV_SESP
2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC
2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SWAV
2000_DATM%NYF_DICE%SSMI_DOCN%DOM_DROF%NYF
2000_DICE%SSMI_DOCN%DOM_DATM%NYF_DROF%NYF
2000_DROF%NYF_DICE%SSMI_DATM%NYF_DOCN%DOM
```
Test baseline:  NA
Test namelist changes: NA
Test status: bit for bit

Fixes #3067 

User interface changes?: Stub components are now optional in compset long names. Also there is less order dependency.

Code review:

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -533,35 +533,57 @@ class Case(object):
         return primary_component
 
 
+    # RE to match component type name without optional piece (stuff after %).
+    # Drop any tailing digits (e.g., the 60 in CAM60) to ensure match
+    # Note, this will also drop trailing digits such as in ww3 but since it
+    # is handled consistenly, this should not affect functionality.
     __mod_match_re__ = re.compile(r"([^%]*[^0-9%]+)")
-    def _valid_compset(self, files, compset_name):
-        "Fill in missing stub models, return full compset name"
-        drv_config_file = files.get_value("CONFIG_CPL_FILE")
-        drv_comp = Component(drv_config_file, "CPL")
-        comp_classes = drv_comp.get_valid_model_components()
-        model_hash = {} # Hash model name to component class index
-        model_set = [None]*len(comp_classes)
-        # First, create hash of model names
-        # A note about indexing. Relevant component classes start at 1
-        # because we ignore CPL for finding model components.
-        # Model components would normally start at zero but since we are
-        # dealing with a compset, 0 is reserved for the time field
-        for comp_ind in range(1, len(comp_classes)):
-            comp = comp_classes[comp_ind]
-            # Find list of models for component class
-            # List can be in different locations, check CONFIG_XXX_FILE first
-            node_name = 'CONFIG_{}_FILE'.format(comp)
-            models = files.get_components(node_name)
-            if (models is None) or (None in models):
-                # Backup, check COMP_ROOT_DIR_XXX
-                node_name = 'COMP_ROOT_DIR_' + comp
+    def _valid_compset(self, files, compset_name, model_hash=None):
+        """Fill in missing stub models, return full compset name
+        >>> Case(read_only=False)._valid_compset(None, '2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV', model_hash={'datm':1,'satm':1,'dlnd':2,'slnd':2,'dice':3,'sice':3,'docn':4,'socn':4,'drof':5,'srof':5,'sglc':6,'swav':7,'ww3':7,'sesp':8})
+        '2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV_SESP'
+        >>> Case(read_only=False)._valid_compset(None, '2000_DATM%NYF_DICE%SSMI_DOCN%DOM_DROF%NYF', model_hash={'datm':1,'satm':1,'dlnd':2,'slnd':2,'dice':3,'sice':3,'docn':4,'socn':4,'drof':5,'srof':5,'sglc':6,'swav':7,'ww3':7,'sesp':8})
+        '2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV_SESP'
+        >>> Case(read_only=False)._valid_compset(None, '2000_DICE%SSMI_DOCN%DOM_DATM%NYF_DROF%NYF', model_hash={'datm':1,'satm':1,'dlnd':2,'slnd':2,'dice':3,'sice':3,'docn':4,'socn':4,'drof':5,'srof':5,'sglc':6,'swav':7,'ww3':7,'sesp':8})
+        '2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV_SESP'
+        >>> Case(read_only=False)._valid_compset(None, '1850_CAM60_CLM50%BGC-CROP_CICE_POP2%ECO%ABIO-DIC_MOSART_CISM2%NOEVOLVE_WW3_BGC%BDRD', model_hash={'datm':1,'satm':1, 'cam':1,'dlnd':2,'clm':2,'slnd':2,'cice':3,'dice':3,'sice':3,'pop':4,'docn':4,'socn':4,'mosart':5,'drof':5,'srof':5,'cism':6,'sglc':6,'ww':7,'swav':7,'ww3':7,'sesp':8})
+        '1850_CAM60_CLM50%BGC-CROP_CICE_POP2%ECO%ABIO-DIC_MOSART_CISM2%NOEVOLVE_WW3_SESP_BGC%BDRD'
+        """
+        if model_hash is None:
+            # First, create hash of model names
+            # A note about indexing. Relevant component classes start at 1
+            # because we ignore CPL for finding model components.
+            # Model components would normally start at zero but since we are
+            # dealing with a compset, 0 is reserved for the time field
+            drv_config_file = files.get_value("CONFIG_CPL_FILE")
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/case/case.py
```diff
@@ -530,6 +533,71 @@ def _find_primary_component(self):
...
+        # A note about indexing. Relevant component classes start at 1
+        # because we ignore CPL for finding model components.
+        # Model components would normally start at zero but since we are
+        # dealing with a compset, 0 is reserved for the time field
+        for comp_ind in xrange(1, len(comp_classes)):
+            comp = comp_classes[comp_ind]
+            # Find list of models for component class
+            # List can be in different locations, check CONFIG_XXX_FILE first
+            node_name = 'CONFIG_{}_FILE'.format(comp)
+            models = files.get_components(node_name)
```
gold2718: First, look for supported components where E3SM lists them.

scripts/lib/CIME/case/case.py
```diff
@@ -530,6 +533,71 @@ def _find_primary_component(self):
...
+            if (models is None) or (None in models):
+                # Backup, check COMP_ROOT_DIR_XXX
+                node_name = 'COMP_ROOT_DIR_' + comp
+                models = files.get_components(node_name)
```
gold2718: Backup, look where CESM lists supported components.

scripts/lib/CIME/case/case.py
```diff
@@ -530,6 +533,71 @@ def _find_primary_component(self):
         return primary_component
 
 
+    __mod_match_re__ = re.compile(r"([^%]*[^0-9%]+)")
```
gold2718: To always match a model name, peel off any `%` stuff plus any tailing digits (e.g., `CAM60`). Since this is done consistently, no special cases for ww3 or fv3 are required.

src/components/data_comps/desp/cime_config/config_component.xml
```diff
@@ -14,7 +14,7 @@
   -->
 
   <description modifier_mode="1">
-    <desc esp="DESP[%NOOP][%TEST]">Data External System Processor (DESP)</desc>
+    <desc esp="DESP[%NOOP][%TEST]">Data External System Processor (DESP) </desc>
```
gold2718: Unrelated typo fix.

jedwards4b:
Error in py3 testing:
```
FAIL: test_pylint_scripts_lib_CIME_case_case_py (__main__.B_CheckCode)
  File "./scripts_regression_tests.py", line 3204, in test
    self.assertTrue(result == "", msg=result)
AssertionError: False is not true : ************* Module CIME.case.case
scripts/lib/CIME/case/case.py:549:24: E0602: Undefined variable 'xrange' (undefined-variable)
scripts/lib/CIME/case/case.py:585:24: E0602: Undefined variable 'xrange' (undefined-variable)
```

scripts/lib/CIME/case/case.py
billsacks: Would it help to have a bit of error checking here - e.g., abort if models is still None (and/or maybe if (None in models)??? I don't totally understand what is an error condition and what's okay here)

scripts/lib/CIME/case/case.py
billsacks: Can you add a comment here illustrating why this regular expression looks like it does? In particular, I'm confused by the initial `[^%]*`. A few examples would help here.

gold2718:
@billsacks: Is this any better? I had to move the code around to make it easy to test without a real case.

scripts/lib/CIME/case/case.py
billsacks: Sorry, I'm still confused about why you have `[^%]*` in this regex, as opposed to just `[^0-9%]+`. Can you provide an example where the `[^%]*` is needed? Sorry to be harping on this, but I find that it's very hard to modify regexes later without sufficient examples illustrating why they look the way they look.

scripts/lib/CIME/case/case.py
```diff
@@ -530,6 +533,93 @@ def _find_primary_component(self):
...
+                       "Unable to find list of supported components")
+
+                for model in models:
+                    mod_match = Case.__mod_match_re__.match(model.lower()).group(1)
+                    model_hash[mod_match] = comp_ind
+
+        else:
+            # Note: this case is only used for testing, normal code should
+            #       not call with model_hash
+            comp_classes = ['CPL', 'ATM', 'LND', 'ICE', 'OCN', 'ROF', 'GLC', 'WAV', 'ESP']
```
billsacks: I don't love having code like this in the production code. I can live with it, but can you at least add a comment at the top of the routine documenting that model_hash should be None in production code, non-None for testing?

To clarify my initial thought on making this testable: I was imagining that you could take everything below this point and extract it into its own function, then write the doctests on that function. Then the code would look something like this:

```python
def _valid_compset(self, files, compset_name):
    # Existing code for setting comp_classes and model_hash goes here; no need for a
    # conditional on model_hash

    return _valid_compset_impl(self, compset_name, comp_classes, model_hash)

def _valid_compset_impl(self, compset_name, comp_classes, model_hash):
    # Doctests here

    # Implementation, starting with:
    # Find the models declared in the compset
    model_set = [None]*len(comp_classes)
```

I still feel that would be easier to understand and maintain, but I'm okay with what you have here as long as you add the documentation of `model_hash` mentioned above.

scripts/lib/CIME/case/case.py
gold2718: `fv3gfs`
Your version would only pick up `fv`. Using greedy nature of regex processing, my version only chops trailing digits. Unless I goofed?

billsacks: Ah, thanks, that's exactly the example I was looking for. So a comment that would help me would be something like:

This regex picks up characters up to the first '%', but leaving off any digits immediately before the '%'. So, for example, 'CAM60%FOO' will pick up 'CAM', and 'FV3GFS%BAR' will pick up 'FV3GFS'.

If that is correct, can you please add a comment along those lines?

gold2718: The current comment states "Drop any trailing digits" (I fixed my typo). Is that not explicit enough? I don't really want to write a regex tutorial in the comment block. Currently, I have added this at the end:
```
# Note: interstitial digits are included (e.g., in FV3GFS).
```
Sufficient?

billsacks: Yes, your suggested addition to the comment is sufficient - thanks.

bishtgautam:
@gold2718, Thanks for this PR.

bishtgautam:
@gold2718, Using latest CIME master, I'm getting an error when I add `_SIAC` in one of the compset name for e3sm. Here are the detail instructions of how to reproduce the error https://gist.github.com/bishtgautam/421bf5419ad893e99379306e072a66aa. Any idea what is going wrong? During last week's call, it was pointed out that one namelist setting (can't recall which one) is not defined in E3SM but was present in CESM. Could that be an issue here?

gold2718:
@bishtgautam,
I keep mentioning that you need to add the component to config_files.xml (see my review of #3055 . I do not see that this has been done in the version used in the gist. Can you please try that? Also, because the supported component classes (`COMP_CLASSES`) is currently in src/drivers/mct/cime_config/config_component.xml (as opposed to config_component_e3sm.xml), the new class will have to be added to both the cesm version and the e3sm version.
I think it would be easier if `COMP_CLASSES` was moved into the config_component_e3sm.xml and config_component_cesm.xml, any thoughts on this from @mvertens? @jedwards4b?

bishtgautam:
>I keep mentioning that you need to add the component to config_files.xml (see my review of #3055 . I do not see that this has been done in the version used in the gist. 

I'm not testing my branch within CIME, but am using CIME master. With CIME master, do I still need to add `CONFIG_IAC_FILE` in `config/e3sm/config_files.xml`? Also, I don't see an entry for `CONFIG_IAC_FILE` in `config/cesm/config_files.xml`. Does a CESM composet with a long name that explicitly includes `_SIAC` works with latest CIME master?

gold2718:
No. No one but you is working with the IAC component class.
The change I made was only the infrastructure, you need to add the IAC component class in the correct places for it to be recognized. In addition to `COMP_CLASSES`, you need to add at least one component model (e.g., `SIAC`) to the correct places in config_files.xml  The models listed in `CONFIG_IAC_FILE` (for E3SM) or `COMP_ROOT_DIR_IAC` (for CESM) are the way the new compset validator function recognizes valid components.

rljacob:
@bishtgautam you have to edit config/e3sm/config_files.xml.   Look at the block of code for  CONFIG_WAV_FILE and add something similar for CONFIG_IAC_FILE.  You'll only have two entries so far: one for saic and one for xaic.   You can later add a gcam entry.

bishtgautam:
@gold2718, @rljacob, thanks for the explanation and suggestions.

PR #3079, opened 2019-04-15
Jgfouca/branch for acme split 2019 04 15

jgfouca:
Changes:
* Improve robustness of wget and err msgs
* Restore Aquaplanet Capability
* Making sure that we print out PIO logs only from rank 0 of each component.
* modified log message in shr_strdata_mod

Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


scripts/lib/CIME/Servers/wget.py
```diff
@@ -15,9 +15,9 @@ def __init__(self, address, user='', passwd=''):
             self._args += "--password {} ".format(passwd)
         self._server_loc = address
 
-        err = run_cmd("wget {} --spider {}".format(self._args, address))[0]
-        expect(err == 0,"Could not connect to repo '{0}'\nThis is most likely either a proxy, or network issue .")
-
+        cmd = "wget {} --no-check-certificate --spider {}".format(self._args, address)
```
jedwards4b: Is this addition wise?  Maybe it should be an option in the xml that is implemented on a per server basis?

jgfouca:
@jedwards4b , that's a good question and it crossed my mind as well.

I'm no security expert, but I think the risk here is extremely low. I was also bypassing certificates when we were using an svn repo for input data.

PR #3084, opened 2019-04-18
Allow specific tests to ignore field list differences

billsacks:
Due to recent changes to cprnc, differences in field lists (i.e., if one
case has some diagnostic fields that are missing from the other case)
are now treated as differences. However, this is a problem for some new
tests that are wanted, such as ensuring that turning on carbon isotopes
doesn't change answers.

This PR allows a particular test type to dictate that field list
differences should be ignored. For tests that use SystemTestsCompareTwo,
this is as simple as adding an argument to the SystemTestsCompareTwo
constructor. For other tests, a new argument can be added to the call to
self._component_compare_test. (Differences in the values in shared
fields are still treated as differences.)

Test suite: scripts_regression_tests on cheyenne; also manual testing as
    noted below
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Manual testing:

- Ran a PEM test where I put in place a different copy of the '.base'
  file as the test is running, which has an extra field: reported as
  DIFFER, as expected.

- Redid that test with this new flag: now reported as IDENTICAL; done
  via the following diffs:

```diff
diff --git a/scripts/lib/CIME/SystemTests/pem.py b/scripts/lib/CIME/SystemTests/pem.py
index 34bb67a..abe2546 100644
--- a/scripts/lib/CIME/SystemTests/pem.py
+++ b/scripts/lib/CIME/SystemTests/pem.py
@@ -20,7 +20,8 @@ class PEM(SystemTestsCompareTwo):
                                        separate_builds = True,
                                        run_two_suffix = 'modpes',
                                        run_one_description = 'default pe counts',
-                                       run_two_description = 'halved pe counts')
+                                       run_two_description = 'halved pe counts',
+                                       ignore_fieldlist_diffs = True)

     def _case_one_setup(self):
         pass
```

- If I instead put in place a file with diffs in one field along with
  the added field, this is still reported as DIFFER (i.e., I didn't
  break the reporting of true differences when adding
  ignore_fieldlist_diffs)

Fixes: none

User interface changes?: N

Thanks to @klindsay28 for the inspiration for this change. Also cc @ekluzek .

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -52,7 +52,8 @@ class SystemTestsCompareTwo(SystemTestsCommon):
                  run_two_suffix = 'test',
                  run_one_description = '',
                  run_two_description = '',
-                 multisubmit = False):
+                 multisubmit = False,
+                 ignore_fieldlist_diffs = False):
         """
         Initialize a SystemTestsCompareTwo object. Individual test cases that
         inherit from SystemTestsCompareTwo MUST call this __init__ method.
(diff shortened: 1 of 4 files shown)


scripts/lib/CIME/SystemTests/system_tests_compare_two.py
```diff
@@ -52,7 +52,8 @@ def __init__(self,
                  run_two_suffix = 'test',
                  run_one_description = '',
                  run_two_description = '',
-                 multisubmit = False):
+                 multisubmit = False,
+                 ignore_fieldlist_diffs = False):
```
billsacks: This new constructor argument is the "user interface" to this new change. i.e., this is how you can leverage this feature in system tests that extend SystemTestsCompareTwo.

scripts/lib/CIME/hist_utils.py
```diff
@@ -376,8 +388,11 @@ def cprnc(model, file1, file2, case, rundir, multiinst_driver_compare=False, out
             elif "the two files seem to be DIFFERENT" in out:
                 files_match = False
             elif "the two files DIFFER only in their field lists" in out:
-                files_match = False
-                comment = CPRNC_FIELDLISTS_DIFFER
+                if ignore_fieldlist_diffs:
```
billsacks: This is the main substantive change in this PR.

Issue #3082, opened 2019-04-17
Don't set OMP_NUM_THREADS in MPI-only runs


amametjanov:
There is no need for cases configured without threading to set `OMP_NUM_THREADS=1`. Presence/absence would make it easier to differentiate between MPI-only and threaded+thread_count=1 runs: e.g. by just looking at output of `preview_run`.

This would be helpful:
```
diff --git a/cime/scripts/lib/CIME/case/case.py b/cime/scripts/lib/CIME/case/case.py
index 220a554..3d736d0 100644
--- a/cime/scripts/lib/CIME/case/case.py
+++ b/cime/scripts/lib/CIME/case/case.py
@@ -177,7 +177,8 @@ class Case(object):
         threads_per_core = 1 if (threads_per_node <= max_mpitasks_per_node) else smt_factor
         self.cores_per_task = self.thread_count / threads_per_core

-        os.environ["OMP_NUM_THREADS"] = str(self.thread_count)
+        if self.get_build_threaded():
+            os.environ["OMP_NUM_THREADS"] = str(self.thread_count)

         self.srun_binding = smt_factor*max_mpitasks_per_node / self.tasks_per_node

@@ -1411,7 +1412,8 @@ directory, NOT in this subdirectory."""
         if not self._is_env_loaded or reset:
             if job is None:
                 job = self.get_primary_job()
-            os.environ["OMP_NUM_THREADS"] = str(self.thread_count)
+            if self.get_build_threaded():
+                os.environ["OMP_NUM_THREADS"] = str(self.thread_count)
             env_module = self.get_env("mach_specific")
             env_module.load_env(self, job=job, verbose=verbose)
             self._is_env_loaded = True
diff --git a/cime/scripts/lib/CIME/case/case_run.py b/cime/scripts/lib/CIME/case/case_run.py
index f7fc66a..05f9b4b 100644
--- a/cime/scripts/lib/CIME/case/case_run.py
+++ b/cime/scripts/lib/CIME/case/case_run.py
@@ -76,7 +76,8 @@ def _run_model_impl(case, lid, skip_pnl=False, da_cycle=0):
     model = case.get_value("MODEL")

     # Set OMP_NUM_THREADS
-    os.environ["OMP_NUM_THREADS"] = str(case.thread_count)
+    if case.get_build_threaded():
+        os.environ["OMP_NUM_THREADS"] = str(case.thread_count)

     # Run the model
[...]

PR #3085, opened 2019-04-19
Omit OMP_NUM_THREADS in MPI-only runs

amametjanov:
Omit OMP_NUM_THREADS in MPI-only runs

Test suite: SMS.f19_g16.X
Test baseline: mpi-only vs. threaded
Test namelist changes: none
Test status: [bit for bit]

Fixes ESMCI/cime#3082

User interface changes?: none

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -177,7 +177,8 @@ class Case(object):
         threads_per_core = 1 if (threads_per_node <= max_mpitasks_per_node) else smt_factor
         self.cores_per_task = self.thread_count / threads_per_core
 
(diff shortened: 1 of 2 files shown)


jedwards4b:
@jgfouca any reason not to merge this one?

jgfouca:
Didn't realize this was waiting on me, sry.

jgfouca:
@amametjanov , we're going to have to revert this PR. Both CESM and E3SM have many machines where the mpirun command is configured to use OMP_NUM_THREADS from the environment. If it's not set, then CIME will crash.

amametjanov:
Okay

PR #3088, opened 2019-04-22
properly link nag f90

jedwards4b:
Fix an issue in mct configure by providing the correct link flags for nag f90.

Test suite: SMS_D.f45_g37.A.izumi_nag
Test baseline: 
Test namelist changes: 

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/machines/config_batch.xml
+++ b/config/cesm/machines/config_batch.xml
@@ -402,6 +402,7 @@
 
   <batch_system type="pbs" MACH="izumi" >
     <batch_submit>ssh izumi cd $CASEROOT ; qsub</batch_submit>
+    <jobid_pattern>(\d+.izumi.unified.ucar.edu)$</jobid_pattern>
     <directives>
       <directive>-l nodes={{ num_nodes }}:ppn={{ tasks_per_node }}</directive>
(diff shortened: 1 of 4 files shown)


fischer-ncar:
For hobart, these links are specified in cime/scripts/Tools/Makefile.  Which is bad, because it makes
it hard to find, as I found out.  We should remove the code for hobart from the Makefile and put
it in config_compiler.xml where it belongs.

Code in Makefile

      ifeq ($(strip $(MACH)),hobart)
       # GCC needs to be able to link to
       # nagfor runtime to get autoconf
       # tests to work.
         CFLAGS += -Wl,--as-needed,--allow-shlib-undefined
         SLIBS += -L$(COMPILER_PATH)/lib/NAG_Fortran -lf62rts
      endif

jedwards4b:
agreed

jedwards4b:
@fischer-ncar I'll add it to this PR

jedwards4b:
Rerunning tests

jedwards4b:
scripts_regression_tests.py --compiler=nag now passes on both izumi and hobart.

PR #3093, opened 2019-04-26
Change DATM_CLMNCEP_YR_ALIGN for present-day compsets

billsacks:
For present-day compsets, the CESM land group wants
DATM_CLMNCEP_YR_ALIGN to match DATM_CLMNCEP_YR_START for present-day
compsets. This will make the model year align with the forcing year. For
example, if RUN_STARTDATE is set to 2000-01-01, then the model will use
year-2000 atmospheric forcing data in the first year, etc.

Test suite:
- Ran `SMS_D_Ld3.f10_f10_musgs.I2000Clm50SpGs.cheyenne_gnu`
  - Confirmed that DATM_CLMNCEP_YR_ALIGN is set correctly
  - With RUN_STARTDATE=2000-01-01, confirmed that it uses year-2000
    atmospheric forcing data for the first few days of the run
Test baseline: n/a
Test namelist changes: Yes - changes settings in datm_in
Test status: climate changing: changes behavior significantly for
  present-day I compsets. Climatological averages should be the same,
  but year alignment changes.

Fixes none

User interface changes?: N

Code review:

### DIFF
--- a/src/components/data_comps/datm/cime_config/config_component.xml
+++ b/src/components/data_comps/datm/cime_config/config_component.xml
@@ -187,30 +187,31 @@
     <valid_values></valid_values>
     <default_value>1</default_value>
     <values match="last">
-      <value compset="2000.*_DATM%1PT">1</value>
+      <value compset="2000.*_DATM%1PT">1972</value>
       <value compset="1850.*_DATM%QIA">1</value>
       <value compset="1850.*_DATM%CRU">1</value>
       <value compset="1850.*_DATM%GSW">1</value>
-      <value compset="2000.*_DATM%QIA">1</value>
+      <value compset="2000.*_DATM%WISOQIA">2000</value>
+      <value compset="2000.*_DATM%QIA">1972</value>
       <value compset="HIST.*_DATM%QIA">1895</value>
       <value compset="HIST.*_DATM%CRU">1901</value>
       <value compset="HIST.*_DATM%GSW">1901</value>
       <value compset="20TR.*_DATM%QIA">1895</value>
       <value compset="20TR.*_DATM%CRU">1901</value>
       <value compset="20TR.*_DATM%GSW">1901</value>
-      <value compset="4804.*_DATM%QIA">1</value>
+      <value compset="4804.*_DATM%QIA">1948</value>
       <value compset="RCP.*_DATM%QIA">2004</value>
       <value compset="RCP.*_DATM%CRU">2005</value>
       <value compset="RCP.*_DATM%GSW">2005</value>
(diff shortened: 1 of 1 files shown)


billsacks:
@rljacob As written, this PR will change behavior for present-day I compsets in E3SM as well as CESM. Can you please pass this along to the relevant person/people to review this on the E3SM side? It may be that they want this change, too: The CESM land group found the old behavior problematic, and much prefer this change in order to have the model years in a present-day I compset use atmospheric forcing from the appropriate year, for the sake of comparisons with observations. If E3SM doesn't want this change, then we can come up with a way to make this behavior model-specific.

I have looked at what present-day E3SM compsets will be affected by this change. From what I can find in E3SM:
- ELM has lots of 2000 compsets: this will change behavior for them
- I don't see any uses of 2010 compsets
- 2003 compsets seem to just be old ones inherited from CESM
- one 4804 compset that just seems to be an old one inherited from CESM

rljacob:
@bishtgautam please take a look at this and the above comment.

bishtgautam:
@jgfouca, FYI, After this PR is merged to master and the update of E3SM's cime will not be BFB.

jgfouca:
@bishtgautam , thanks for the heads-up.

billsacks:
@rljacob do you want to review this yourself, or should I remove you as a reviewer since @bishtgautam gave his okay?

rljacob:
@billsacks what was the behavior before this PR?   An I case that started in year 2000 would use what for forcing?

billsacks:
> @billsacks what was the behavior before this PR? An I case that started in year 2000 would use what for forcing?

It's hard to tell. Before this PR, yr_align was set to 1 for year-2000 datm forcing. As per http://esmci.github.io/cime/data_models/input-streams.html#details-on-year-align that means that simulation year 1 used the first year of datm forcing data. So the year of forcing data in year 2000 would depend on something like 1999 modulo (number of years of forcing data).

I'll point out, though, that at least in CESM, until recently, present-day I compsets had started at year 1 by default, not year 2000. If that was true in e3sm, too, then the relevant question may be what year of forcing data is used in year 1. This cime PR is designed to work well with a present-day configuration actually starting with a real year, like year 2000, rather than year 1.

rljacob:
We had a few cases that use 2000_DATM%NYF and these did NOT show any diffs from this change.  Does that make sense?

billsacks:
@rljacob yes that makes sense: I'm pretty sure NYF mode is distinct from CLMNCEP mode, so doesn't use the settings affected by this change.

PR #3095, opened 2019-04-26
better error reporting for refcase errors

### LINKED ISSUE #3094: GET_REFCASE error reporting
If GET_REFCASE=TRUE but the refcase directory is invalid no error is reported until runtime.

jedwards4b:
Improve reporting of invalid settings in REFCASE when GET_REFCASE=TRUE
Test suite: scripts_regression_tests.py, by hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #3094 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case/check_input_data.py
+++ b/scripts/lib/CIME/case/check_input_data.py
@@ -208,6 +208,8 @@ def stage_refcase(self, input_data_root=None, data_list_dir=None):
 
         if os.path.isabs(run_refdir):
             refdir = run_refdir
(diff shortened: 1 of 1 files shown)


jedwards4b:
I'm happy to change the content of the error message but I don't find your suggested wording especially helpful and keep in mind that this error message may also occur when a refcase is in inputdata.

Katetc:
I think that's a good error. With the full path, you'll be able to see if you haven't pointed to the place that you need to. And requiring the final directory to contain the rpointer files is the correct way to fix the error that I had this morning. Thanks!

PR #3118, opened 2019-05-22
Use truncated division operator in SEQ SystemTest for py2+3 compatibility

jhkennedy:
When using python 3, the SEQ test will fail to set `NTASKS_*` because the XML `set_value` expects an integer, but a float is produced. This is because of the change from the `/` operator using truncated division for integers in python 2 to using real division always in python 3. Simply replacing `/` with the truncated division operator `//` fixes the issue (available since python 2.2).

Test suite: ./scripts_regression_tests.py on cori-knl
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes E3SM-Project/E3SM#2940
User interface changes?: N


jhkennedy:
Thanks @jedwards4b

PR #3123, opened 2019-05-28
Jgfouca/branch for acme split 2019 05 28

jgfouca:
Change List:
* **Add first stage of ocean/ice shelf coupling**
* Bring COMPOSE semi-Lagrangian tracer transport code into HOMME and E3SM (required minor changes to build.py)
* Add checkpoint of timing data immediately after initialization
* bless_test_results: Be able to handle build-only tests

Test baseline: 
Test namelist changes: 
Test status: BFB

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


scripts/lib/CIME/build.py
```diff
@@ -233,7 +233,7 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
     if mpilib == "mpi-serial":
         libs.insert(0, mpilib)
 
-    if cam_target == "preqx_kokkos":
+    if cam_target in ("preqx_kokkos", "theta-l"):
```
jedwards4b: Can we combine this and line 499 so that it only appears in one place?

src/drivers/mct/main/cime_comp_mod.F90
```diff
@@ -4467,4 +4542,42 @@ subroutine cime_run_write_restart(drv_pause, write_restart, drv_resume)
...
+    character(len=*), intent(in)    :: ckpt_filename
+    integer, intent(in)             :: ckpt_mpicom
+
+103 format( 5A )
+104 format( A, i10.8, i8)
+
+    call t_adj_detailf(+1)
+
+    call t_startf("sync1_tprf")
+    call mpi_barrier(ckpt_mpicom,ierr)
```
jedwards4b: This code introduces an mpi_barrier to the cime_run loop - it could have serious performance implications.  In other cases timing barriers are inside ``` if (barrier_alarm)  ```  code blocks.  It seems like this should be the same.

src/drivers/mct/shr/seq_flds_mod.F90
```diff
@@ -2399,6 +2410,155 @@ subroutine seq_flds_set(nmlfile, ID, infodata)
     call set_glc_elevclass_field(name, attname, longname, stdname, units, x2l_fluxes_from_glc, &
          additional_list = .true.)
 
+    if (trim(cime_model) == 'e3sm') then
```
jedwards4b: Is this really the appropriate test here or should it be something about active glc model?

jedwards4b:
@billsacks is on vacation this week, I've asked @Katetc to review the glc changes.

rljacob:
@Katetc FYI, Bill looked carefully at the glc-ocean coupling code in its E3SM PR here: https://github.com/E3SM-Project/E3SM/pull/2726

src/drivers/mct/main/cime_comp_mod.F90
rljacob: @matthewhoffman was that barrier needed?

matthewhoffman: @rljacob , I'm not familiar with this code - it is not part of the PR from me.

src/drivers/mct/shr/seq_flds_mod.F90
matthewhoffman: @jedwards4b , originally @billsacks had wanted much of the OCN-GLC functionality enabled only in E3SM.  In the end we settled on an implementation that avoids needing to know which model is being used and instead functionality is enabled by compset definition via the OCN init routine.  As such, this test could potentially be removed if you and Bill prefer.

src/drivers/mct/main/cime_comp_mod.F90
jgfouca: @worleyph made this change I think (7b1e7442 in E3SM)

scripts/lib/CIME/build.py
jgfouca: Good idea, will do

src/drivers/mct/main/cime_comp_mod.F90
worleyph: This same logic existed in-line previously. I just encapsulated it in a subroutine to cut down on duplicate code. It is called in one new location, just before the start of the run loop (but outside), immediately preceding the existing RUN_LOOP_BSTART barrier (which is annoying to have both, but seemed preferable to the other options I considered). 
It is executed immediately after initialization (so once - a new feature, and the point the PR), after the end of the first day (also once, has been this way for a long time), and whenever tprof_alarm is true (also for a long time). The barriers also occur during cime_final.

rljacob: The communicator on the barrier changed. It was call mpi_barrier(mpicom_GLOID,ierr)

jedwards4b: @worleyph thanks for the explanation, that makes sense - but since mpicom_GLOID is global anyway can we remove it as a function argument and just use the global variable?  I think that
renaming the variable creates confusion.

worleyph: There has always been the possibility that components would write out there own checkpoints (or even do their timing independently). We have never supported that, but the design has always been in the back of my head, thus the use of a formal parameter and not hardwiring mpicom_GLOID. This is a subroutine, and the communicator is clearly a passed parameter. I don't find this confusing :-). You just have to check the call sites? So... my opinion is to leave it as is. However, do what you feel that you need to do.

worleyph: Actually, I'd strongly prefer that this remain a formal parameter. There are a lot of comunicators floating around in cime_comp_mod.F90. It seems a strange design to hardwire GLOID in a subroutine that is otherwise pretty generic.

rljacob: I see why we might want to change it back to GLOID.  If we want to search the code for everywhere a barrier is called on GLOID, it would be missed.  Its not obvious from the name of the function that a barrier is going to called inside it.

worleyph: t_drvstartf, t_drvstopf also has the (optional) parameter to specify communicator to be used in a barrier, though it is t_barrierf and so not normally enabled. It also never uses GLOID (currently), but rather CPLID and omp(eci)%mpicom_compid, and mpicom_barrier (as called from component_exch, so even more once removed).

seq_domain_check_grid also has a barrier (and an allreduce) using an mpicom formal parameter. 

Could make this an optional parameter that defaults to GLOID if the parameter is not specified, so grep would pick it up? But, aince all of the calls (all two of them) are local to the file cime_comp_mod.F90 at the moment, would not be hard to hunt down.

In general, if there is a need to identify all barrier (and allreduce?) calls using GLOID, I wouldn't depend on grep without further investigation of all barrier and allreduce calls, unless we eliminate all communicator parameters. So, my preference is the same, to leave as it, but will make changes if people request it (or you can make the changes yourself).

jgfouca: Telecon: we are fine with this as is.

src/drivers/mct/shr/seq_flds_mod.F90
billsacks: (I'm sort of back from vacation today, but still taking most of the day off, so may not respond to further discussion until tomorrow.)

As @matthewhoffman says, I requested this change after consultation with @mvertens . We wanted to ensure that these new fields don't impact CESM performance. I suspect that the new flag that @matthewhoffman is referring to isn't available at the time of `seq_flds_set`, so I don't think it will work to use that logical here (but I may be wrong).

billsacks:
@Katetc There is no need for you to review these code changes unless you want to, since (as @rljacob says) they have already undergone a pretty careful review by myself and others. However, before this comes to master, can you please run testing to ensure that CESM tests pass and are bit-for-bit. I think we discussed this before, but I forget what testing we came up with. I'm now thinking aux_glc and prealpha; prebeta is probably unnecessary (unless we said differently before).

src/drivers/mct/shr/seq_flds_mod.F90
matthewhoffman: @billsacks is correct that the needed new flags are not available when `seq_flds_set` is called.  `seq_flds_set` is called from `cime_pre_init2`, which occurs before `cime_init` which is the routine from which the new flags are set (in `ocn_init`).  So I think the current implementation here is the best we can do to prevent these new coupler fields from affecting CESM without significant restructuring.

jgfouca:
@billsacks , @jedwards4b , did anyone kick off tests? I'm getting some pressure to do a CIME merge to E3SM soon and it can't happen until this PR is merged.

jedwards4b:
@fischer-ncar can you please kick off tests with this PR merged to master?

fischer-ncar:
I'm getting the following error for all of my tests.

ERROR: Namelist default for variable glc2ocn_fmapname refers to unknown XML variable GLC2OCN_FMAPNAME.'

billsacks:
@jgfouca @Katetc - The plan I worked out with @fischer-ncar is: I am doing our `aux_glc` testing today. Once I get that all passing, @fischer-ncar will run CESM prealpha tests.

I think I have fixed the problem @fischer-ncar ran into. I have pushed this and some other minor changes (fixing whitespace / indentation) to a branch on my fork (https://github.com/billsacks/cime/tree/jgfouca/branch-for-acme-split-2019-05-28). Once I get tests passing there, I'll hand off these changes.

@jgfouca how do you prefer that I hand these changes back to you? Should I simply merge these changes back to your branch here, or do you prefer something different?

jgfouca:
@billsacks , yes, just merge them into this PR.

billsacks:
Okay, I have pushed two commits. The first just cleans up whitespace. The second fixes the problem @fischer-ncar ran into, by moving a bunch of xml variables from the e3sm-specific file to the shared file.

This now passes all `aux_glc` tests on cheyenne, and results are bit-for-bit. (Also, performance doesn't seem to be affected significantly from a few spot-checks.) This test suite covers I and T compsets. @fischer-ncar can you go ahead and retry prealpha tests with this branch?

fischer-ncar:
The prealpha testing found

ERROR: Need to provide valid mapping file between glc and ocn in xml variable glc2ocn_fmapname 

For B, F, and X compsets.

billsacks:
@fischer-ncar - I just pushed a fix for your issue

jgfouca:
@billsacks , @fischer-ncar , how are we looking?

billsacks:
@jgfouca Chris's testing is still underway. Cheyenne (the machine, not the whole town, AFAIK) just went down due to weather problems, so it may be a while before full test results are in. But it looks like we're past the major issues at this point. @fischer-ncar and I agree that, if you're wanting to get this to master, you can go ahead and do so, and we'll follow up with additional issues/PRs if Chris runs into any more problems in his testing.

jgfouca:
@billsacks , thanks.

PR #3125, opened 2019-05-29
Add workflow xml 

jedwards4b:
Move the batch_jobs portion of config_batch.xml to config_workflow.xml 

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: Adds a --workflow-case option to create_newcase

Update gh-pages html (Y/N)?:

Code review:


config/e3sm/machines/config_batch.xml
```diff
@@ -541,30 +541,4 @@
       <queue walltimemax="01:00:00" default="true">lr3</queue>
     </queues>
    </batch_system>
-
-  <batch_jobs>
```
jgfouca: These should have reappeared in a workflow XML file for E3SM, right?

jedwards4b: Yes, it's there now.

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -2607,64 +2607,6 @@
     <desc>External script to be run after model completion</desc>
   </entry>
 
- <!-- Job Submission stuff -->
```
jgfouca: Since both e3sm and cesm are using env_workflow.xml now, shouldn't we be able to continue to share these entries?

jedwards4b: yes, I started out with the idea that I would only change cesm but then decided that there was no reason not to change e3sm as well.

jedwards4b:
This is now functionally ready and tested but we want to wait until after the next cesm release is ready to merge it, probably sometime the week of 6/3

jedwards4b:
Oops - closed unintentionally

jedwards4b:
@fischer-ncar Now that cesm2.1.1 is out the door this PR is ready to review and merge.

PR #3139, opened 2019-06-10
Jgfouca/branch for acme split 2019 06 10

jgfouca:
Changes:
* Fix bug in PET timestamp (need to use build as a library, not a subprocess) (https://github.com/E3SM-Project/E3SM/pull/2949)
* Fix bug in generating baselines (Need to use the base case id when parsing test opts.) (https://github.com/E3SM-Project/E3SM/pull/2950)
* Change FFLAGS_NOPT to be an append to the normal set of flags. This reduces duplication with FFLAGS (https://github.com/E3SM-Project/E3SM/pull/2922)
* Add ability to ask for CMake args (get_standard_cmake_args) (https://github.com/E3SM-Project/E3SM/pull/2922)
* CMake macros should NOT be looking at environment, just CMake vars (https://github.com/E3SM-Project/E3SM/pull/2922)
* case_setup will make both macro types (Make and CMake) (https://github.com/E3SM-Project/E3SM/pull/2922)
* Import a FindPetsc CMake module (https://github.com/E3SM-Project/E3SM/pull/2922)
* Implement a more-advanced way of sharing builds within test suites (Implements a concept of build groups within test_scheduler.) (https://github.com/E3SM-Project/E3SM/pull/2961)

Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


config/cesm/machines/config_compilers.xml
```diff
@@ -511,6 +503,38 @@ using a fortran linker.
   <TRILINOS_PATH MPILIB="mpich2">$ENV{TRILINOS_PATH}</TRILINOS_PATH>
 </compiler>
 
+<compiler MACH="athena">
```
jgfouca: These additions are suspect.

jedwards4b: seem to be a repeat of lines starting at 474

jgfouca: OK, i will remove

billsacks:
@jgfouca with this change

> Change FFLAGS_NOPT to be an append to the normal set of flags. This reduces duplication with FFLAGS

What is the purpose of FFLAGS_NOOPT as compared with `<append DEBUG="TRUE">`? It seems that the two are now accomplishing the same thing, so it seems confusing to have both of them.

Also, have you or someone else verified that the resulting flags for a debug build of cesm are unchanged for all compilers for which we have FFLAGS_NOOPT defined? If not yet, I'd like to request that that comparison be done before this PR is merged to master, to avoid any surprises.

billsacks:
(Adding myself as a reviewer so that I can mark when I feel my above comment has been addressed.)

jgfouca:
> What is the purpose of FFLAGS_NOOPT as compared with <append DEBUG="TRUE">? It seems that the two are now accomplishing the same thing, so it seems confusing to have both of them.

@billsacks , that's a good point, they are very similar now. I think the only difference is that FFLAGS_NOOPT guarantees that the flags come later in the compile line than the normal FFLAGS.

I did confirm that the builds we as expected for us in cam (using CMake). I used the bld_diff tool to validate.

jgfouca:
@billsacks , as an example of a NOOPT file:

```
rrtmg_lw_k_g.o: rrtmg_lw_k_g.f90
        $(FC) -c $(FPPFLAGS) $(INCLDIR) $(INCS) $(FREEFLAGS) $(FFLAGS) $(FFLAGS_NOOPT) $<

rrtmg_sw_k_g.o: rrtmg_sw_k_g.f90
        $(FC) -c $(FPPFLAGS) $(INCLDIR) $(INCS) $(FREEFLAGS) $(FLLAGS) $(FFLAGS_NOOPT) $<
```

billsacks:
@jgfouca Thanks a lot for the bulleted summary of the major changes in here. In the future, for cime splits like this that include a lot of different changes that potentially impact CESM as well as E3SM (i.e., a lot more than just e3sm-specific config changes), it would help if you could also include pointers to the E3SM PR associated with each bulleted item, to make it easier for us to review particular changes separately. I have done that this time by editing your initial comment (hope you don't mind).

scripts/Tools/Makefile
```diff
@@ -563,10 +563,10 @@ ifeq ($(MODEL),cam)
   # Until mods are made to read the data from files, just remove optimization from
   # their compilation.
 rrtmg_lw_k_g.o: rrtmg_lw_k_g.f90
-	$(FC) -c $(FPPFLAGS) $(INCLDIR) $(INCS) $(FREEFLAGS) $(FFLAGS_NOOPT) $<
+	$(FC) -c $(FPPFLAGS) $(INCLDIR) $(INCS) $(FREEFLAGS) $(FFLAGS) $(FFLAGS_NOOPT) $<
 
 rrtmg_sw_k_g.o: rrtmg_sw_k_g.f90
-	$(FC) -c $(FPPFLAGS) $(INCLDIR) $(INCS) $(FREEFLAGS) $(FFLAGS_NOOPT) $<
+	$(FC) -c $(FPPFLAGS) $(INCLDIR) $(INCS) $(FREEFLAGS) $(FLLAGS) $(FFLAGS_NOOPT) $<
```
billsacks: Typo: FLLAGS should be FFLAGS

jgfouca: Thanks! Not sure how that made it through my verification...

jgfouca: Oh, I wasn't using the Makefile to build those (they are in cam, which switched to CMake). And the baselines were generated with master.

jgfouca:
@billsacks , you're right, I neglected to update some of the Make Depends files. Doing that now.

jgfouca:
OK, all tests are passing for me now, so I'll merge once you guys approve.

billsacks:
@jgfouca thanks for these fixes. This looks fine to me, now, but I'd still like @jedwards4b to confirm the change I called out in https://github.com/ESMCI/cime/pull/3139#pullrequestreview-248425224

jedwards4b:
@billsacks I don't think that it will be a problem, overall I think that the cleanup is worth the risk.

jgfouca:
Thanks, @billsacks !

PR #3147, opened 2019-06-13
if CIME_DRIVER=nuopc add pylint check for nuopc files

jedwards4b:
Add the external files in the nuopc directory to code_checker if the CIME_DRIVER=nuopc 

Test suite: CIME_DRIVER=nuopc scripts_regression_tests.py
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/code_checker.py
+++ b/scripts/lib/CIME/code_checker.py
@@ -96,7 +96,6 @@ def check_code(files, num_procs=10, interactive=False):
     files_to_check = []
     if files:
         repo_files = get_all_checkable_files()
-#        repo_files = run_cmd_no_fail('git ls-files', from_dir=get_cime_root(), verbose=False).splitlines()
         for filearg in files:
             if os.path.exists(filearg):
                 files_to_check.append(os.path.abspath(filearg))
--- a/scripts/lib/CIME/code_checker.py
+++ b/scripts/lib/CIME/code_checker.py
@@ -4,7 +4,7 @@ Libraries for checking python code with pylint
 
 from CIME.XML.standard_module_setup import *
 
-from CIME.utils import run_cmd, run_cmd_no_fail, expect, get_cime_root, is_python_executable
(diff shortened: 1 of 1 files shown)


scripts/lib/CIME/code_checker.py
```diff
@@ -92,7 +95,8 @@ def check_code(files, num_procs=10, interactive=False):
     # is a valid file, if not, we search the repo for a file with similar name.
     files_to_check = []
     if files:
-        repo_files = run_cmd_no_fail('git ls-files', from_dir=get_cime_root(), verbose=False).splitlines()
+        repo_files = get_all_checkable_files()
+#        repo_files = run_cmd_no_fail('git ls-files', from_dir=get_cime_root(), verbose=False).splitlines()
```
jedwards4b: @jgfouca looking for feedback on this change - if it's okay I'll remove the commented code.

scripts/lib/CIME/code_checker.py
```diff
@@ -75,9 +75,12 @@ def get_all_checkable_files():
 ###############################################################################
     cimeroot = get_cime_root()
     all_git_files = run_cmd_no_fail("git ls-files", from_dir=cimeroot, verbose=False).splitlines()
-
+    if get_cime_default_driver() == "nuopc":
```
jgfouca: Curious as to why this was needed. all_git_files are computed from cimeroot, so wouldn't that include the nuopc subdirectory?

jedwards4b: Because the nuopc subdirectory is no longer part of the same git repo.

jgfouca: Ah, ok.

scripts/lib/CIME/code_checker.py
jgfouca: It looks ok. I can't remember why I used ls-files instead of checkable_files.

PR #3148, opened 2019-06-14
Pio2 update

jedwards4b:
Update to the latest pio2 development version 2.4.4, set pio2 as default (thought I had already done this)

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


src/drivers/mct/cime_config/config_component.xml
```diff
@@ -2313,7 +2313,7 @@
 
   <entry id="PIO_VERSION">
     <type>integer</type>
-    <default_value>1</default_value>
+    <default_value>2</default_value>
```
jedwards4b: This is the only change outside of the pio2 directory

rljacob:
This sets every case to using pio2?

jedwards4b:
It sets the default as pio2, you can override in config_pio.xml

billsacks:
@jedwards4b at a CSEG meeting you said:

> One thing needed for this is: pio2 doesn't allow data conversions at the low level, so the components (ctsm and cam) need to convert double to single precision themselves

As far as I know, there haven't been any changes made to CTSM in this respect (I don't know about CAM). Or have you updated pio2 to be able to do the data conversions?

Also, what testing have you done of recent versions of pio2, for basic runability, performance and answer changes? If you haven't done much recently, I'm thinking it would be good if you or @fischer-ncar ran some testing on this.

src/drivers/mct/cime_config/config_component.xml
jedwards4b: Actually I also ended up with a change in the Makefile to get mpi-serial working.

jedwards4b:
Along with prealpha testing (which is still running) I ran  SMS.f09_g17.B1850.cheyenne_intel.allactive-defaultio  with all history files set to 4byte rather than 8byte and with pIo_rearranger=1 and pio_rearranger=2.  It passed in each case.

billsacks:
Thanks a lot @jedwards4b ! So does that mean that nothing is needed in ctsm and cam?

jgfouca:
This may impact E3SM, but I can worry about that when I do the next merge.

jedwards4b:
There were other problems int testing, I've reset the default PIO_VERSION=1 so that this PR can be accepted.

PR #3153, opened 2019-06-25
Peano/nemo

daniele-peano:
* Update config files for Athena
* Add NEMO grids
* Add NEMO as recognized ocean component
* Add resubmit option for LSF machine
* Add short term archive case for NEMO

### DIFF
--- a/config/cesm/machines/config_compilers.xml
+++ b/config/cesm/machines/config_compilers.xml
@@ -460,6 +460,36 @@ using a fortran linker.
   </FFLAGS>
 </compiler>
 
+<compiler MACH="athena">
+  <CPPDEFS>
+    <!-- these flags enable nano timers -->
+    <append MODEL="gptl"> -DHAVE_NANOTIME -DBIT64 -DHAVE_VPRINTF -DHAVE_BACKTRACE -DHAVE_SLASHPROC -DHAVE_COMM_F2C -DHAVE_TIMES -DHAVE_GETTIMEOFDAY </append>
+  </CPPDEFS>
+  <NETCDF_PATH>$ENV{NETCDF}</NETCDF_PATH>
+  <PIO_FILESYSTEM_HINTS>gpfs</PIO_FILESYSTEM_HINTS>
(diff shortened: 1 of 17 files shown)


jedwards4b:
@daniele-peano I pushed a change to fix the pylint errors.

daniele-peano:
@jedwards4b I tested the changes that you pushed. Nemo run and resubmit successfully on Athena.

config/cesm/machines/config_compilers.xml
```diff
@@ -74,6 +74,13 @@ using a fortran linker.
   <SUPPORTS_CXX>FALSE</SUPPORTS_CXX>
 </compiler>
 
+<compiler>
```
jedwards4b: I think that this can go into nemo/cime_config/buildlib instead of here.

scripts/lib/CIME/case/case.py
```diff
@@ -1020,13 +1020,21 @@ def _create_caseroot_tools(self):
...
 
-        # set up utility files in caseroot/Tools/
-        toolfiles = [os.path.join(toolsdir, "check_lockedfiles"),
-                     os.path.join(toolsdir, "getTiming"),
-                     os.path.join(toolsdir, "save_provenance"),
-                     os.path.join(toolsdir,"Makefile"),
-                     os.path.join(toolsdir,"mkSrcfiles"),
-                     os.path.join(toolsdir,"mkDepends")]
+        if "NEMO" in self._compsetname:
+            nemodir = os.path.join(self.get_value("CIMEROOT"),"..","components","nemo","NEMO36","bld")
```
jedwards4b: Why do you have a different mkDepends?  Do we really need it?

config/cesm/machines/config_compilers.xml
daniele-peano: I moved this command in nemo/cime_config/buildlib

scripts/lib/CIME/case/case.py
daniele-peano: I update the mkDepend to account for NEMO case

PR #3157, opened 2019-07-01
Caseinsensitive user_nl

jedwards4b:
We were downcasing all user_nl and namelist variables but not all inputs handled by this function are case insensitive - for example ESMF runconfig files used in nuopc.   So the input from namelist_definition.xml must be accepted in whatever case it is in the file and input from user_nl_* must match that input.   The solution is to use a case insensitive dict type in namelist_definition.xml 

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


jgfouca:
One thought: you may want to provide a citation for that case-insensitive dict implementation. I assume you got it off the web somewhere.

PR #3160, opened 2019-07-01
Workflow changes

jedwards4b:
Changes to continue workflow development 

Test suite: scripts_regression_tests.pu on izumi 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/Tools/preview_run
+++ b/scripts/Tools/preview_run
@@ -64,19 +64,24 @@ def _main_func(description):
 
         set_logger_indent("      ")
         job_id_to_cmd = case.submit_jobs(dry_run=True, job=job)
+        env_batch = case.get_env('batch')
         for job_id, cmd in job_id_to_cmd:
+            overrides = env_batch.get_job_overrides(job_id, case)
             print("  FOR JOB: {}".format(job_id))
             print("    ENV:")
             case.load_env(job=job_id, reset=True, verbose=True)
             if "OMP_NUM_THREADS" in os.environ:
                 print("      Setting Environment OMP_NUM_THREADS={}".format(os.environ["OMP_NUM_THREADS"]))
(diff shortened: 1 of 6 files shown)


scripts/Tools/preview_run
```diff
@@ -74,9 +74,8 @@ def _main_func(description):
             print("      {}".format(case.get_resolved_value(cmd)))
             print("")
 
-        if job == case.get_primary_job():
-            print("MPIRUN:")
-            print ("  {}".format(case.get_mpirun_cmd()))
+            print("MPIRUN (job={}):".format(job))
```
jedwards4b: MPIRUN should be printed for each job that uses it.

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -167,26 +167,34 @@ def set_batch_system(self, batchobj, batch_system_type=None):
     def make_batch_script(self, input_template, job, case, outfile=None):
         expect(os.path.exists(input_template), "input file '{}' does not exist".format(input_template))
         env_workflow = case.get_env('workflow')
-        task_count = env_workflow.get_value("task_count", subgroup=job)
-        tasks_per_node = env_workflow.get_value("tasks_per_node", subgroup=job)
-        if not tasks_per_node:
-            tasks_per_node = case.tasks_per_node
+        total_tasks, num_nodes, tasks_per_node, thread_count = env_workflow.get_job_specs(job)
```
jedwards4b: allow jobs other than case.run and case.test to use mpirun, this was neglected in the past because the only other job in the original workflow case.st_archive does not use mpirun.

PR #3169, opened 2019-07-15
Ensemble clone

jedwards4b:
Adds a --ensemble option to create_clone allowing the user to generate an ensemble of cases cloned from --clone 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: Adds --ensemble to create_clone, to use this option the original case must end in one or more integers, so for example if the case is named case.001 and --ensemble 4 is used you will get case.002, case.003, case.004 created.  

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/create_clone
+++ b/scripts/create_clone
@@ -5,7 +5,7 @@ from Tools.standard_script_setup import *
 from CIME.utils import expect
 from CIME.case  import Case
 from argparse   import RawTextHelpFormatter
-
+import re
 logger = logging.getLogger(__name__)
 
 ###############################################################################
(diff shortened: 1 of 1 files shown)


scripts/create_clone
```diff
@@ -23,6 +23,9 @@ def parse_command_line(args):
                         help="(required) Specify a case to be cloned. If not a full pathname, "
                         "\nthe case to be cloned is assumed to be under then current working directory.")
 
+    parser.add_argument("--ensemble", default=1,
+                        help="clone an ensemble of cases, the case name argument must end in an integer.")
```
billsacks: Can you add a few more details here, such as:

The specified number of new cases will be created, each ending with a unique integer. The numbering will start with the value in the case name argument.

billsacks: Actually, it might be even better to give an example here instead of my suggestion. I like the example that you gave in the PR text, so maybe just add that to the help text here?

jedwards4b: I've updated the help section with your suggestion @billsacks

billsacks: Thanks, this looks good now.

PR #3201, opened 2019-08-14
Compset aliascheck

### LINKED ISSUE #3182: Recent cime versions let you create a case with a non-existent compset
If you try to create a case with a compset alias, where that alias isn't actually defined anywhere, cime should (and I'm pretty sure did, in the past) abort with an error saying that that alias cannot be found. But now the scripts happily proceed, treating the alias as the date specifier and using all stub components: e.g., from a standalone cime checkout, at the current master:

```
./create_newcase --compset I2000Ctsm50NwpBgcCropGswpGs --res f10_f10_musgs --case $scratch/test_err_0722a --run-unsupported
```

This happily creates the case; according to README.case, the generated compset long name is:

```
I2000Ctsm50NwpBgcCropGswpGs_SATM_SLND_SICE_SOCN_SROF_SGLC_SWAV_SIAC_SESP
```

My initial hunch is that this problem came in when code was added to allow stub components to be absent in the specified compset name (#3068 from @gold2718 ), but I haven't looked closely to confirm this hunch.

(This problem was first picked up by @slevisconsulting.)

jedwards4b:
If all components are stubs and this is not a test, then throw an error - this is not a valid compset name or alias. 

Test baseline: 
Test namelist changes: 

Fixes #3182 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -439,7 +439,7 @@ class Case(object):
             if result is not None:
                 del self.lookups[key]
 
-    def _set_compset(self, compset_name, files, driver="mct"):
+    def _set_compset(self, compset_name, files, driver="mct", test=False):
         """
         Loop through all the compset files and find the compset
         specifation file that matches either the input 'compset_name'.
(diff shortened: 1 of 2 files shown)


billsacks:
My understanding from this morning's discussion in #3182 is that this will probably be closed in favor of a solution from @gold2718 . Am I understanding that correctly? (Just wondering whether I should actually go ahead and review this pr.)

jedwards4b:
Yes, that's my understanding as well.

PR #3203, opened 2019-08-14
Fix bad compset

### LINKED ISSUE #3182: Recent cime versions let you create a case with a non-existent compset

gold2718:
Added a check for an invalid compset name. If a compset long name has all stub components, it must originate in an alias.
Added doctest tests and a scripts_regression_test to ensure bad compsets are trapped during `create_newcase`.

Note that the MCC_P1.f19_g16_rx1.A test timed out. Some timesteps were taking 3-6 times as long as expected.
Test baseline: NA
Test namelist changes: No
Test status: bit for bit

Fixes #3182 

User interface changes?: No

Code review:

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -483,7 +483,7 @@ class Case(object):
             self._compsetname = compset_name
 
         # Fill in compset name
-        self._compsetname, self._components = self.valid_compset(self._compsetname, files)
+        self._compsetname, self._components = self.valid_compset(self._compsetname, compset_alias, files)
         # if this is a valiid compset longname there will be at least 7 components.
         components = self.get_compset_components()
         expect(len(components) > 6, "No compset alias {} found and this does not appear to be a compset longname.".format(compset_name))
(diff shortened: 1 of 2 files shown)


jgfouca:
Thanks for adding a test!

PR #3227, opened 2019-09-04
Srt update

jedwards4b:
Clean up some issues in scripts_regression_tests.py, it wasn't removing directories after completing and it was printing some confusing messages about removing files when there were no files to remove.

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


jedwards4b:
@jgfouca good point, I've changed it so that it only removes an empty directory.

PR #3231, opened 2019-09-05
Changes to support Radiative Convective Equilibrium

cacraigucar:
Add aquaplanet aqp11 to support running radiative convective equlibrium (RCE)

Add namelist settings:
   - rce_sst (Sea surface temperature for RCE runs
   - seq_flux_mct_albdif and seq_flux_mct_albdir - The albdif and albdir variables which were assigned
       constant values in seq_flux_mct.F90.  The default values are what they were previously.  
       They are assigned different values for RCE runs

Add a namelist for seq_flux_mct.F90 to handle the new variables

Add an optional argument to shr_orb_cosz in shr_orb_mod.F90 to return values based on a
       uniform angle.

Test suite: aux_cam and cime_developer test
Test baseline:  cam6_1_031
Test namelist changes: Ran test with RCE on (to verify all namelists are being brought in)
                                     Ran tests without RCE on (all runs BFB)
Test status: [bit for bit, roundoff, climate changing] BFB

Fixes [CIME Github issue #]
N/A

User interface changes?: 
new namelist settings described above

Code review:


cacraigucar:
It should be noted that these changes need to go into both the release branch and into master.

billsacks:
In talking with @cacraigucar about this: It seems like, in principle, this change could change answers, because values that used to be double precision 0.06 and 0.07 are now being read from the namelist as single precision 0.06 and 0.07. It is unclear why this didn't result in answer changes in @fischer-ncar 's recent testing. We have decided that she will change these namelist values to be explicitly double precision on master, because that feels like the right thing to do long-term, but won't change them on maint-5.6 (to avoid adding one more change in to the cesm2.1.2 release testing).

PR #3237, opened 2019-09-13
Merge maint-5.6

fischer-ncar:
- Add support for running Radiative Convective Equilibrium.
- Add support to archive metadata for lens experiment type.
- Make the streams file legitiment xml files so that we can parse them.
- Add support for CMIP6 OMIP (datm-pop-cice) compsets.
- Update JRA forcing files for datm.
- Update JRA025->gx1v7, JRA025->tx0.1v3 ROF2OCN mapping files.
- Update CCSM_CO2_PPMV to CMIP6 1850 value (284.317)
  in compsets with datm, POP, and POP ecosys.

Test suite:  scripts_regression_tests
Test baseline: 
Test namelist changes: 
Test status: climate changing in C & G compsets

Fixes [CIME Github issue #]

User interface changes?: 
Add a namelist for seq_flux_mct.F90 to handle the new variables.
Add an optional argument to shr_orb_cosz in shr_orb_mod.F90 to return values based on a
uniform angle.

Update gh-pages html (Y/N)?:

Code review:

PR #3241, opened 2019-09-19
Merge maint-5.6

fischer-ncar:
- Add support for running Radiative Convective Equilibrium.
- Add support to archive metadata for lens experiment type.
- Make the streams file legitiment xml files so that we can parse them.

Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:
Add a namelist for seq_flux_mct.F90 to handle the new variables.
Add an optional argument to shr_orb_cosz in shr_orb_mod.F90 to return values based on a
uniform angle.

Update gh-pages html (Y/N)?:

Code review:


fischer-ncar:
This is the same as PR #3237.  But with answer changes that affect E3SM removed.

jgfouca:
@fischer-ncar , just to be clear, it's OK to put answer-changing stuff in; you just have to use the equivalent of 
```
if (model == 'cesm')
    * do answer changing stuff
```

jedwards4b:
In this case we are changing the JRA dataset to be a different set of data - it doesn't make sense to have JRA mean one thing for one model and something else for another model, so we are redoing the new dataset with a different name and maintaining the old JRA.

PR #3246, opened 2019-09-23
address objection to using case in workflow

### LINKED ISSUE #3245: "workflow-case" language is confusing since the word "case" has a special meaning in CIME
This is to strongly advocate for not using the word "case" to describe a set of workflow tasks.  "Case" already has a special meaning in CIME, i.e., an instance of a model configuration.  Overloading this word to also refer to a workflow will cause confusion for the user.  We are still early enough where this change can be made with relatively little fallout, so I'm advocating that we do this ASAP.

Specifically, in the XML, the attribute "case" should not be used - instead something like "id" or "name" could be used to describe a workflow.  In addition, on the create_newcase command, "--workflow-case" should be changed to simply "--workflow"  or maybe "--workflow-id"

jedwards4b:
Address objection to use of the word case in the workflow xml, id is now used instead.

Test baseline: 
Test namelist changes: 

Fixes #3245 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/machines/config_workflow.xml
+++ b/config/cesm/machines/config_workflow.xml
@@ -1,5 +1,5 @@
 <?xml version="1.0"?>
-<config_workflow version="2.0">
+<config_workflow version="2.1">
   <!--
   File: config_workflow.xml
   Purpose: Define the jobs to be run, the order to run them, the job size and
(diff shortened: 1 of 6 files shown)


rsdunlapiv:
Thank you @jedwards4b.  A hardy approval for this!

PR #3251, opened 2019-09-27
Add namelist for minimum wind speed for atmOcn fluxes

cacraigucar:
Minimum wind speed for atmOcn fluxes had been hardwired up until this point.  Introduced a namelist variable and changed the default setting only for AQP11.  

Test suite:  all aux_cam tests, cime_developer tests
Test baseline: 
Test namelist changes: Ran the AQP11 test, and it's namelist has the new value of 1
Test status: [bit for bit, roundoff, climate changing] BFB

Fixes [CIME Github issue #]

User interface changes?: Added "seq_flux_atmocn_minwind" namelist.  The default for all jobs is 0.5, which is the previous jobs.  The only exception is AQP11, which now has the value of 1.0

Update gh-pages html (Y/N)?:n

Code review:

### DIFF
--- a/src/drivers/mct/cime_config/namelist_definition_drv.xml
+++ b/src/drivers/mct/cime_config/namelist_definition_drv.xml
@@ -4266,4 +4266,17 @@
     </values>
   </entry>
 
+  <entry id="seq_flux_atmocn_minwind">
+    <type>real</type>
+    <category>seq_flux_mct</category>
+    <group>seq_flux_mct_inparm</group>
+    <desc>
+       minimum wind speed for atmOcn flux calculations
+    </desc>
+    <values>
+      <value>0.5</value>
+      <value aqua_planet_sst_type="sst_aquap11">1.0</value>
+    </values>
+  </entry>
+
 </entry_id>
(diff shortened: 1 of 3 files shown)


cacraigucar:
@brianpm  @kevinareed @whannah1 

Here is minimum windspeed pull request that we've discussed.

src/drivers/mct/cime_config/namelist_definition_drv.xml
```diff
@@ -4266,4 +4266,17 @@
     </values>
   </entry>
 
+  <entry id="seq_flux_atmocn_minwind">
+    <type>real</type>
+    <category>seq_flux_mct</category>
+    <group>seq_flux_mct_inparm</group>
+    <desc>
+       minimum wind speed for atmOcn flux calculations
```
whannah1: Can you change this to say "...idealized flux calculations" here so it's clear what this is for?

whannah1: oh wait, is this always operating even in normal simulations?

PR #3253, opened 2019-09-30
Update gnu modules for cheyenne

fischer-ncar:
Update to gnu 8.3.0 and associated modules

Test suite:  scripts_regression_tests, ERP_Ln9.f19_f19_mg17.F2000climo.cheyenne_gnu
Test baseline: cesm2_2_alpha03c
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -441,11 +441,11 @@ This allows using a different mpirun command to launch unit tests
 	<command name="load">pgi/19.3</command>
       </modules>
       <modules compiler="gnu">
-        <command name="load">gnu/7.3.0</command>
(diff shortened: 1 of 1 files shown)


billsacks:
@fischer-ncar - in the past, compiler updates have caused problems, leading to either test failures or answer changes. If I remember correctly, I think we've said in the past that compiler updates like this should be accompanied with a run of the prealpha & prebeta tests on the relevant machine-compiler. Would you be able to do this in order to verify that this update doesn't cause any problems, and/or to document any expected answer changes?

billsacks:
Awesome, thanks a lot for taking the time to confirm that.

PR #3273, opened 2019-10-28
Refactor how run_cmd handles strings

jgfouca:
run_cmd should always return a string type for stdout and stderr
that matches the system's base str type.

Also, minor refactor to e3sm_cime_mgmt.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


jgfouca:
@jedwards4b , I did a fairly deep dive trying to understand the interactions between subprocess and string/bytes/uncode on py2 3. I think this PR finally gets things right on both 2 and 3, which removed the need for all those encode calls we had to put everywhere.

PR #3277, opened 2019-10-29
fix issue with e3sm field

### LINKED ISSUE #3276: PR 3272 causing test failures in cesm
PR 3272 is causing failures in cesm test SMS_P1.f09_g16.X

jedwards4b:
Recent addition of some exchange fields in e3sm via PR #3272 broke the X compset in cesm.

Test baseline: 
Test namelist changes: 
Test status: bit for bit, 

Fixes #3276 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/drivers/mct/main/cime_comp_mod.F90
+++ b/src/drivers/mct/main/cime_comp_mod.F90
@@ -532,6 +532,7 @@ module cime_comp_mod
        &Sa_co2diag:Sa_co2prog'
 
   ! --- other ---
+  character(len=cs)        :: cime_model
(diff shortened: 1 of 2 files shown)


jedwards4b:
@jgfouca I think that this may be broken in e3sm X compsets as well.  If it is my change won't fix it.

jgfouca:
@jedwards4b , I think it must be OK for us since it passed scripts_regression_tests (has some X tests) and our e3sm_developer test has an X test as well.

PR #3280, opened 2019-10-30
fix glob for bless_test_results

### LINKED ISSUE #3278: bless_test_results failing with ref1 and ref2 directories
bless_test_results is processing the .ref1 and .ref2 directories that are produced by the ERI tests.

The TestStatus file in the .ref directories isn't completed by the ERI tests, so it looks like those tests
failed to run and aren't blessed.  

Another issue is the .ref directories might be the most recent time stamp, which is causing all of the other test to be skipped.

jedwards4b:
Improved patern matching for bless_test_results.

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #3278

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/bless_test_results.py
+++ b/scripts/lib/CIME/bless_test_results.py
@@ -59,7 +59,7 @@ def bless_history(test_name, case, baseline_name, baseline_root, report_only, fo
 def bless_test_results(baseline_name, baseline_root, test_root, compiler, test_id=None, namelists_only=False, hist_only=False,
                        report_only=False, force=False, bless_tests=None, no_skip_pass=False):
 ###############################################################################
-    test_id_glob = "*{}*".format(compiler) if test_id is None else "*{}*".format(test_id)
+    test_id_glob = "*{}[0-9][0-9]".format(compiler) if test_id is None else "*{}".format(test_id)
     test_status_files = glob.glob("{}/{}/{}".format(test_root, test_id_glob, TEST_STATUS_FILENAME))
     expect(test_status_files, "No matching test cases found in for {}/{}/{}".format(test_root, test_id_glob, TEST_STATUS_FILENAME))
 


scripts/lib/CIME/bless_test_results.py
```diff
@@ -59,7 +59,7 @@ def bless_history(test_name, case, baseline_name, baseline_root, report_only, fo
 def bless_test_results(baseline_name, baseline_root, test_root, compiler, test_id=None, namelists_only=False, hist_only=False,
                        report_only=False, force=False, bless_tests=None, no_skip_pass=False):
 ###############################################################################
-    test_id_glob = "*{}*".format(compiler) if test_id is None else "*{}*".format(test_id)
+    test_id_glob = "*{}[0-9][0-9]".format(compiler) if test_id is None else "*{}".format(test_id)
```
billsacks: I'm confused how this can work in the case where `test_id` is None. My test names often look like `SMS_D_Ld1_P4x1.f10_f10_musgs.I2000Clm50BgcCropQianRsGs.bishorn_gnu.clm-default.C.20191029_170705_zyp3pv`; it doesn't seem like a glob on '*gnu[0-9][0-9]' would pick that up. What am I missing?

jedwards4b: @billsacks is this not a test-id?  20191029_170705_zyp3pv

jedwards4b: oh - I see, it is a test id but it's not passed in.

jedwards4b: Why is this same assignment in two different places instead of in a shared location?   
@jgfouca I think that maybe you can find a better solution, these are tools that I don't use in my workflow but e3sm and cam developers do.

jedwards4b:
closed for a better fix

PR #3283, opened 2019-10-31
changes to be compatible with the latest cmeps PR to master

mvertens:
Update for compatibility with latest CMEPS - almost ALL of the modifications do not affect the mct  code base

Test baseline: (tested with CMEPS testlist - and everything was bfb except for X compsets)
Test namelist changes: no
Test status: bit for bit (except for X compsets with nuopc)
Fixes:
User interface changes?: No
Code review: jedwards4b


scripts/lib/CIME/case/case_run.py
```diff
@@ -171,7 +171,11 @@ def _post_run_check(case, lid):
     driver = case.get_value("COMP_INTERFACE")
 
     if driver == 'nuopc':
-        file_prefix = 'med'
+        compset = case.get_value("COMPSET")
```
jedwards4b: We should be able to come up with something general here instead of line 175

src/share/util/shr_const_mod.F90
```diff
@@ -7,81 +7,94 @@ MODULE shr_const_mod
 
    use shr_kind_mod
 
-   integer(SHR_KIND_IN),parameter,private :: R8 = SHR_KIND_R8 ! rename for local readability only
+   !----------------------------------------------------------------------------
```
jedwards4b: These changes are not okay, new variables need to follow naming convention and most lines in this file should not change from original - I will submit an update.

jedwards4b:
As of 
5239520 passes all mct cime tests except the expected failure on cheyenne.

PR #3292, opened 2019-11-08
cleanup env batch

jgfouca:
Clean up env_batch
    
Try, wherever possible, to have our queue handle be a qnode object, not
just a queue name. This will result in much better handling of the case
where multiple entries have the same name.

Also, add a test specifically for testing the case where there are multiple queue entries with the same name.

Test suite: scripts_regression_tests.K_TestCimeCase
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


jgfouca:
@jedwards4b , I didn't worry about trying to compute overlap. I think continuing to match the first queue meeting spec is fine.

PR #3295, opened 2019-11-08
initialize albdif and albdir

### LINKED ISSUE #3294: maint-5.6 backward compatibility is broken
PR #3266 introduced a change that is only valid with cam version 
https://github.com/ESCOMP/CAM/tree/cam_cesm2_1_rel_36 and breaks backward compatibility of cime maint-5.6 with older cesm and cam tags.   We need to introduce a fix to make cime backward compatible with older cesm and cam tags so that the maint-5.6 compatibility is maintained.

jedwards4b:
These variables should get reasonable initial values if the namelist isn't read. 

Test suite: scripts_regression_tests.py had testing of SMS_Ld5.f09_g17.B1850G.cheyenne_intel.allactive-cism-test_coupling
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #3294 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/drivers/mct/main/seq_flux_mct.F90
+++ b/src/drivers/mct/main/seq_flux_mct.F90
@@ -109,8 +109,8 @@ module seq_flux_mct
   real(r8),parameter :: const_deg2rad = const_pi/180.0_r8  ! deg to rads
 
   ! albedo reference variables - set via namelist
-  real(r8)  :: seq_flux_mct_albdif  ! albedo, diffuse
-  real(r8)  :: seq_flux_mct_albdir  ! albedo, direct
+  real(r8)  :: seq_flux_mct_albdif = 0.06_r8  ! albedo, diffuse
+  real(r8)  :: seq_flux_mct_albdir = 0.07_r8  ! albedo, direct
   real(r8)  :: seq_flux_atmocn_minwind ! minimum wind temperature for atmocn flux routines
 
   ! Coupler field indices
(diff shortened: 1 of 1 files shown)


cacraigucar:
I would propose that the values be initialized to something like -99 and a check be instituted that looks for this value and halts the program if they are not set.  

I believe that not reading these values should be an error condition as opposed to silently setting the values to something reasonable.

billsacks:
I was about to say something similar to @cacraigucar . In CTSM, we prefer to initialize namelist variables to NaN (using shr_infnan_mod), so that it will be very obvious if a namelist variable isn't read. However, given the way shr_infnan_mod is implemented, I'm not sure if you can initialize to NaN using this module-level initialization: I think you need to do it at runtime. We do this initialization in the routine that reads the namelist, so it will trigger an error if the namelist variable is absent from the namelist. If you're trying to catch the particular error that just happened (where the routine isn't called at all) then I guess that wouldn't help, but it would help in the possibly-more-common case where the routine is called but the namelist wasn't set up properly.

whannah1:
I think any negative value would be fine for initialization. I would use -1.

jedwards4b:
In reviewing the code I don't see how initializing these variables with an obviously incorrect value would have addressed the issue at hand and not just lead to bigger problems.  I invite you to update this pr with your own proposed solution.

jedwards4b:
okay how does this look?

PR #3302, opened 2019-11-13
Two issues on maint-5.6 for izumi

### LINKED ISSUE #3262: cime5.6.24 PATH setting is missing running with pgi on izumi
Simple change to allow izumi/pgi to run with cime5.6.24. I added the following line in env_mach_specific.xml and it would run.

` xml
<env name="PATH">$ENV{PATH}:/cluster/torque/bin</env>
`
This allows the compute nodes to issue a queue submit.

By, the way I changed the original documentation of this issue to split this into two issue. #3302 is now separated from this one.

### LINKED ISSUE #3301: Add PFUNIT for izumi
I'd like to add PFUNIT for izumi. I'll create a PR for this as well.

Here's the diff I have for it right now.

``` xml
diff --git a/config/cesm/machines/config_compilers.xml b/config/cesm/machines/config_compilers.xml
index cc4820086..f6a7b73fb 100644
--- a/config/cesm/machines/config_compilers.xml
+++ b/config/cesm/machines/config_compilers.xml
@@ -887,6 +887,10 @@ using a fortran linker.
   </SLIBS>
 </compiler>
 
+<compiler MACH="izumi" COMPILER="intel">
+  <PFUNIT_PATH MPILIB="mpi-serial" compile_threaded="false">/fs/cgd/csm/tools/pFUnit/pFUnit3.3.3_izumi_Intel19.0.1_noMPI_noOpenMP</PFUNIT_PATH>
+</compiler>
+
 <compiler MACH="hobart" COMPILER="intel">
   <CFLAGS>
     <append> -lifcore</append>
```

ekluzek:
Add path for PFUNIT on izumi and add path for batch commands to $PATH

Testing:
   Ran CLM PFunit testing on izumi and it works in what will be release-clm5.0.28
   Tried to run PGI testing that was failing on izumi, it still fails, but we know the PATH
      change is at least one required thing to get it working. izumi/pgi tests are working for
      me in latest ctsm master ctsm1.0.dev078, and the only important difference I see is
      this PATH change. If someone can show me what else is needed to get this working
      I'd gladly add it to the branch, but we might as well get these two things going even if
      we can't figure out how to get izumi_pgi to run.

Fixes #3301
Address an issue in #3262 

User interface changes?: No


ekluzek:
Good point. I changed the comment to say that it addresses an issue in, rather than it fixes it.

jedwards4b:
@ekluzek Once again I must ask you to provide a test case.  Nether in this pr nor in the issue #3262 do you clearly explain how to reproduce the problem.

jedwards4b:
I tried SMS_P96.f19_g16.X.izumi_pgi and it failed in both maint-5.6 (with izulmipgi) 
and in master.

PR #3307, opened 2019-11-15
Jgfouca/branch for acme split 2019 11 15

jgfouca:
Features:
* Add (or re-add) support for NLDAS configurations

Bug fixes:
* Set SCM to run with netcdf
* Fix py3 issue with floating point data in env_batch
* create_test: fix baseline_root for non-default compiler

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


scripts/lib/CIME/XML/env_batch.py
```diff
@@ -402,8 +402,15 @@ def get_submit_args(self, case, job):
                             rval = val
                     else:
                         rval = val
+
+                    # We don't want floating-point data
+                    try:
```
jgfouca: py3 bug fix

scripts/lib/CIME/test_scheduler.py
```diff
@@ -201,15 +201,13 @@ def __init__(self, test_names, test_data=None,
         self._baseline_cmp_name = baseline_cmp_name # Implies comparison should be done if not None
         self._baseline_gen_name = baseline_gen_name # Implies generation should be done if not None
 
-        # Compute baseline_root
-        self._baseline_root = baseline_root if baseline_root is not None \
+        # Compute baseline_root. Need to set some properties on machobj in order for
+        # the baseline_root to resolve correctly.
+        self._machobj.set_value("COMPILER", self._compiler)
```
jgfouca: BASELINE_ROOT bug fix.

scripts/lib/CIME/test_scheduler.py
```diff
@@ -638,6 +636,7 @@ def _xml_phase(self, test):
                         envtest.set_test_parameter("NTASKS_"+comp, "1")
                         envtest.set_test_parameter("NTHRDS_"+comp, "1")
                         envtest.set_test_parameter("ROOTPE_"+comp, "0")
+                        envtest.set_test_parameter("PIO_TYPENAME", "netcdf")
```
jedwards4b: Why would you hard code this?  You can set it in config_pio.xml if you need to.

src/components/data_comps/datm/cime_config/config_component.xml
```diff
@@ -52,6 +53,7 @@
       <value compset="%CRU">CLMCRUNCEP</value>
       <value compset="%CRUv7">CLMCRUNCEPv7</value>
       <value compset="%GSWP3v1">CLMGSWP3v1</value>
+      <value compset="%NLDAS">CLMNLDAS</value>
```
jedwards4b: Have you checked that this still works with NLDAS2?  I'm a little concerned about how compset matching works.

scripts/lib/CIME/XML/env_batch.py
jgfouca: 99% sure we'll never want to pass a floating point to the batch submission command. Can you think of a context where we would?

scripts/lib/CIME/test_scheduler.py
jgfouca: @amametjanov , can you comment here?

src/components/data_comps/datm/cime_config/config_component.xml
jgfouca: @jonbob can you comment here?

config/e3sm/config_grids.xml
```diff
@@ -168,7 +168,7 @@
       <mask>360x720cru</mask>
     </model_grid>
 
-    <model_grid alias="NLDAS_NLDAS" compset="(DATM_SATM).+CLM">
```
ekluzek: OK, I don't get this change. And I don't think it's correct either. We have compsets that run NLDAS with either MOSART or SROF, and need it to function either way. Do you need to enable compsets that aren't running CLM, but are running MOSART?

scripts/lib/CIME/test_scheduler.py
amametjanov: Tasks, threads and rootpe are already hard-wired. Setting PIO_TYPENAME=netcdf here keeps all SCM-specific settings nearby. Error message is at https://github.com/E3SM-Project/E3SM/pull/3300#issuecomment-551270871

src/components/data_comps/datm/cime_config/config_component.xml
```diff
@@ -10,12 +10,13 @@
...
   -->
   <description modifier_mode="1">
-    <desc atm="DATM[%QIA][%WISOQIA][%CRU][%CRUv7][%GSWP3v1][%NLDAS2][%CPLHIST][%1PT][%NYF][%IAF][%JRA]"> Data driven ATM </desc>
+    <desc atm="DATM[%QIA][%WISOQIA][%CRU][%CRUv7][%GSWP3v1][%NLDAS][%NLDAS2][%CPLHIST][%1PT][%NYF][%IAF][%JRA]"> Data driven ATM </desc>
     <desc option="QIA"> QIAN data set </desc>
     <desc option="WISOQIA">QIAN with water isotopes</desc>
     <desc option="CRU"> CRUNCEP data set </desc>
     <desc option="CRUv7"> CLM CRU NCEP v7 data set </desc>
     <desc option="GSWP3v1"> GSWP3v1 data set </desc>
+    <desc option="NLDAS"> NLDAS data set </desc>
```
ekluzek: Can you give more info. on NLDAS here? And how it's different than NLDAS2? Maybe the name should be NLDAS1? Like NLDAS2 is it only over US, if so the warning below should also apply to it.

src/components/data_comps/datm/cime_config/namelist_definition_datm.xml
```diff
@@ -285,6 +290,7 @@
       <value stream="CLMCRUNCEP">domain.lnd.360x720.130305.nc</value>
       <value stream="CLMGSWP3v1">domain.lnd.360x720_gswp3.0v1.c170606.nc</value>
       <value stream="CLMGSWP3">domain.lnd.360x720_gswp3.0v1.c170606.nc</value>
+      <value stream="CLMNLDAS">domain.lnd.nldas2_0224x0464_c110415.nc</value>
```
ekluzek: IS the NLDAS domain really different than the NLDAS2 domain? It is the same size. Also we decided that 0.125nldas2_0.125nldas2 was a more descriptive name rather than the grid size as it gives the grid cell resolution. You can query the file easy enough to get the number of grid cells.

src/components/data_comps/datm/cime_config/namelist_definition_datm.xml
```diff
@@ -453,6 +459,7 @@
       <value stream="CLMGSWP3.Solar">$DIN_LOC_ROOT_CLMFORC/atm_forcing.datm7.GSWP3.0.5d.v1.c170516/Solar3Hrly</value>
       <value stream="CLMGSWP3.Precip">$DIN_LOC_ROOT_CLMFORC/atm_forcing.datm7.GSWP3.0.5d.v1.c170516/Precip3Hrly</value>
       <value stream="CLMGSWP3.TPQW">$DIN_LOC_ROOT_CLMFORC/atm_forcing.datm7.GSWP3.0.5d.v1.c170516/TPHWL3Hrly</value>
+      <value stream="CLMNLDAS">$DIN_LOC_ROOT/atm/datm7/NLDAS</value>
```
ekluzek: I took a quick look at your data. It's hourly so not dividing into streams as you have done is the right thing to do. I was going to recommend dividing it like the NLDAS2 data which is less frequent if not.

My one concern to point out is that you only have a single year of it (at least in inputdata). This will be problematic if you cycle over that single year of forcing to run a land model for multi-year climate simulations. 

It also looks like it's only for 1980, so some of the year start, end, align assignments aren't going to work right for it.

rljacob:
Original PR for the NLDAS forcing for E3SM:  https://github.com/E3SM-Project/E3SM/pull/3264

config/e3sm/config_grids.xml
jgfouca: @jonbob ^

src/components/data_comps/datm/cime_config/config_component.xml
jgfouca: @jonbob ^

src/components/data_comps/datm/cime_config/namelist_definition_datm.xml
jgfouca: @jonbob ^

src/components/data_comps/datm/cime_config/namelist_definition_datm.xml
jgfouca: @jonbob ^

jonbob: This configuration was first introduced some months ago by @apcraig and got lost, maybe in a CIME merge? I don't know much about it, just trying to shepherd it back in as requested.

src/components/data_comps/datm/cime_config/namelist_definition_datm.xml
jonbob: Again, our PR was just to get a lost capability back into the system. I don't know much about the specifics....

config/e3sm/config_grids.xml
jonbob: I was asked to support a configuration in which MOSART is the only active model, so yes.

src/components/data_comps/datm/cime_config/config_component.xml
jonbob: I did not test that, but can

jonbob: except we don't have any support for NLDAS2 in E3SM. You may have to test it in CESM with a merge of this branch

billsacks: You should be able to test this by using a compset long name, or by introducing a temporary compset for the sake of testing.

I'm hopeful that this particular instance will be okay, because of the `match="last"` above, but I'm concerned that there may be other, similar instances. This seems to be begging for problems, if not now, then in the future.

billsacks:
@jonbob if you're not the right one to weigh in on the NLDAS changes, can you please ensure that the appropriate person or people comment here? If we're going to have this confusing near-duplication – which is potentially error prone both from a user perspective and from a maintenance perspective (e.g., https://github.com/ESMCI/cime/pull/3307#discussion_r347046056) – I'd at least like to hear some arguments that this truly needs to be distinct from the existing NLDAS2 mode, as well as comments addressing points raised by @jedwards4b @ekluzek and myself.

src/components/data_comps/datm/cime_config/config_component.xml
jonbob: Thanks @billsacks - I'll look at that today

jonbob:
@billsacks - my understanding is that @apcraig first introduced this NLDAS configuration some time ago, but it was apparently lost. I can't figure out when and where, but we've been squash-merging these CIME updates and I think lose the history that way. So I don't know if it was replaced by the NLDAS2 configuration and never checked with the right E3SM developers? In any case, I've just been asked to reimplement what had previously been there, but I have little to no knowledge of the history. I do understand your concerns but am unsure how to address them.

src/components/data_comps/datm/cime_config/config_component.xml
jonbob: @billsacks - I had tried this last week but there are files for the NLDAS2 configuration that we do not have

billsacks:
> @billsacks - my understanding is that @apcraig first introduced this NLDAS configuration some time ago, but it was apparently lost. I can't figure out when and where, but we've been squash-merging these CIME updates and I think lose the history that way. So I don't know if it was replaced by the NLDAS2 configuration and never checked with the right E3SM developers? In any case, I've just been asked to reimplement what had previously been there, but I have little to no knowledge of the history. I do understand your concerns but am unsure how to address them.

I spent some time looking back through the CIME history – in the actual CIME repo, not in E3SM. At the time when I introduced the NLDAS2 datm forcing, there was no such forcing available: this was an addition, not a replacement of existing functionality.

If you aren't able to address these comments, then can you please get the appropriate person or people to respond to the review comments here?

src/components/data_comps/datm/cime_config/config_component.xml
billsacks: They should be available from our public inputdata repository via the check_input_data script. Did you have trouble even when using that?

rljacob:
@hydrotian or tanzeli1982 can you answer the questions about NLDAS forcing above?

ekluzek:
Also note that this older NLDAS data only includes 1980. While the newer NLDAS2 data goes from 1980 to 2018. Another solution for this is to just abandon the older data and have both E3SM and CESM use the same NLDAS2 data. The only reason I can think of to keep the older NLDAS data is if there is a reason to preserve answers exactly. Since, this is forcing data I wouldn't think that would matter much.

billsacks:
@tanzeli1982 fixing your call-out (see https://github.com/ESMCI/cime/pull/3307#issuecomment-560531477)

config/e3sm/config_grids.xml
hydrotian: Yes this "MOSART only" compset was originally added by @apcraig in a old E3SM development branch and was recently merged to E3SM master by @jonbob . River model developers like myself need this configuration to run MOSART without running the land model (to save time).

hydrotian:
I'm not sure about the NLDAS to NLDAS2 changes. I think @tanzeli1982 is the right person to answer this question as he created a NLDAS compset to test his sediment scheme in ELM.

config/e3sm/config_grids.xml
billsacks: I think @ekluzek 's point is that you shouldn't need to explicitly specify MOSART here. You can just delete the "CLM" from the compset match without replacing it with anything else. Actually, I'm guessing that the important thing here is that you don't have an active atmosphere and don't have an active ocean, or something like that.

But I won't hold this PR up for this particular point, since at least this change doesn't impact models other than E3SM.

tanzeli1982:
@billsacks @ekluzek @rljacob 
I am not sure whether NLDAS has been introduced to the master branch before. But it works in the Tony's apcraig/mosart/add-inundation branch and is very useful for the simulations of water resources. As you did, I also noticed that E3SM introduced NLDAS2 several months ago. I did try to make use of NLDAS2 to run ELM + MOSART and MOSART only simulations over the continental US but made no success. This is the reason that we requested @jonbob to bring NLDAS to the master branch. However, we will be more than happy to use NLDAS2 only if my described issues are fixed. I agree that it will cause far less confusion.

billsacks:
> I did try to make use of NLDAS2 to run ELM + MOSART and MOSART only simulations over the continental US but made no success. This is the reason that we requested @jonbob to bring NLDAS to the master branch. However, we will be more than happy to use NLDAS2 only if my described issues are fixed. I agree that it will cause far less confusion.

Okay, great, it sounds like we're in agreement about this. Can you please describe the issues that you ran into, or point us to an existing discussion of these specific issues?

tanzeli1982:
Sure. This is the discussion I made with Gautam long before.

Hi Gautam,
 
We are trying to set up a DATM+ELM+MOSART case over the NLDAS grid based on the master branch code but met some problems. Could you help us with it or point to the right person who can help us?
 
Before, I did set up a DATM+ELM+MOSART NLDAS case successfully on the Tony’s add-inundation branch. What I did includes: 1) adding NLDAS as a new DATM mode in data_comps/datm/cime_config/config_component.xml; 2) defining the NDLAS datm in data_comps/datm/cime_config/namelist_definition_datm.xml; 3) creating a new compset in clm/cime_config/config_compsets.xml and giving it an alias (for example, 20TR_DATM%NLDAS_CLM45%CNPECACNTBC_SICE_SOCN_MOSART_SGLC_SWAV).
 
When I tried to mimic the above steps on the master branch using NLDAS2 (because someone has already completed the 1st and 2nd steps on the master, I only did the 3rd step), it sends me an error like “ERROR: grid alias NLDAS_NLDAS not valid for compset 20TR_DATM%NLDAS2_CLM45%CNPECACNTBC_SICE_SOCN_MOSART_SGLC_SWAV_SIAC_SESP”.  I checked the definition of the NLDAS_NLDAS resolution using “./query_config –grids”. It shows that “alias: NLDAS_NLDAS (only for compsets that are (DATM_SATM).+CLM ) non-default grids are: atm:NLDAS  lnd:NLDAS  ocnice:NLDAS  rof:NLDAS  glc:null  wav:null, mask is: NLDAS”. I guess that the key point here is to modify (DATM_SATM).+CLM to (DATM|SATM).+CLM but did not figure out how to do it.

billsacks:
@tanzeli1982 okay, I see how that issue required a small change in E3SM's config_grids.xml. However, I'm more concerned about the changes in datm, which introduce a NLDAS datm forcing mode in addition to the already-existing NLDAS2 datm forcing mode. The issue that you mention above doesn't seem related to the datm changes. Did you run into some other issues that required you to introduce this separate datm forcing mode?

tanzeli1982:
> @tanzeli1982 okay, I see how that issue required a small change in E3SM's config_grids.xml. However, I'm more concerned about the changes in datm, which introduce a NLDAS datm forcing mode in addition to the already-existing NLDAS2 datm forcing mode. The issue that you mention above doesn't seem related to the datm changes. Did you run into some other issues that required you to introduce this separate datm forcing mode?

Likely not. But as I was blocked by the error, I cannot predict what would happen next.

bishtgautam:
> Could you help us with it or point to the right person who can help us?

@tanzeli1982, I don't know who the right person on E3SM-side is to sort this out.

All, Here is a naive question. Is there truly a difference between `NLDAS` and `NLDAS2`? Or, both groups added the same capability but with different names?

jonbob:
@tanzeli1982 - why don't we see if we can get past the problem you were having and use the NLDAS2 datm setup? I believe that's the biggest conflict, though @billsacks please let me know if I'm missing something. I'm testing E3SM cases with both NLDAS and NLDAS2 for datm

rljacob:
@billsacks said `the time when I introduced the NLDAS2 datm forcing, there was no such forcing available: this was an addition, not a replacement of existing functionality.`

Then why is there a "2" in "NLDAS2" ?

Also, can someone tell me what "NLDAS" stands for?

hydrotian:
> Then why is there a "2" in "NLDAS2" ?
> 
> Also, can someone tell me what "NLDAS" stands for?

@rljacob NLDAS stands for "North American Land Data Assimilation System". NLDAS2 is the 2nd Phase of the NLDAS project. The difference between NLDAS1 and 2 can be found [here](https://ldas.gsfc.nasa.gov/faq/nldas).

rljacob:
It sounds like we could solve this by just renaming "NLDAS" to something more descriptive like "RIVERTEST".   The comment can indicate that its 1 year of hourly data from NLDAS1.  We probably shouldn't call it "NLDAS" unless it is the full NLDAS dataset (phase 1 or 2)

billsacks:
> All, Here is a naive question. Is there truly a difference between `NLDAS` and `NLDAS2`? Or, both groups added the same capability but with different names?

@bishtgautam From above comments from @ekluzek it sounds like the newly-introduced NLDAS forcing may only have a single year. I'm not sure if that was intentional or not. I haven't been able to determine from these comments whether the newly-introduced one uses NLDAS1 or NLDAS2 forcing.

> @tanzeli1982 - why don't we see if we can get past the problem you were having and use the NLDAS2 datm setup? I believe that's the biggest conflict, though @billsacks please let me know if I'm missing something. I'm testing E3SM cases with both NLDAS and NLDAS2 for datm

Thanks @jonbob - that sounds like a good plan. Indeed, if the existing NLDAS2 datm forcing works fine for E3SM's needs, then that seems like the simplest and easiest-to-maintain solution moving forwards.

> It sounds like we could solve this by just renaming "NLDAS" to something more descriptive like "RIVERTEST". The comment can indicate that its 1 year of hourly data from NLDAS1. We probably shouldn't call it "NLDAS" unless it is the full NLDAS dataset (phase 1 or 2)

@rljacob This seems like a good plan if @jonbob 's tests don't work out.

jonbob:
@billsacks - is the biggest issue adding the NLDAS mode to datm? If so, I do have our compsets working with datm set to NLDAS2, and we can move to using that and get rid of the NLDAS mode. Also, are you OK with us specifying a compset in the datm config_component that uses a single year of the NLDAS2 data? That would give us a test that requires downloading much less data, sort of like the NLDAS dataset we were using. @tanzeli1982 seems happy with this resolution. I know we've added a mode to dlnd to support a compset with MOSART as the only active component -- is that OK with everyone there?

billsacks:
> @billsacks - is the biggest issue adding the NLDAS mode to datm? If so, I do have our compsets working with datm set to NLDAS2, and we can move to using that and get rid of the NLDAS mode. Also, are you OK with us specifying a compset in the datm config_component that uses a single year of the NLDAS2 data? That would give us a test that requires downloading much less data, sort of like the NLDAS dataset we were using. @tanzeli1982 seems happy with this resolution. I know we've added a mode to dlnd to support a compset with MOSART as the only active component -- is that OK with everyone there?

@jonbob yes, this all sounds like a great plan. I think you're suggesting adding something like this

https://github.com/ESMCI/cime/blob/2a7d36454a01d1b5308b38b253cf2f88669956e0/src/components/data_comps/datm/cime_config/config_component.xml#L250

but for the NLDAS2 forcing data. That sounds like a great plan.

Thank you for all of your work towards this solution.

jonbob:
@billsacks - that's exactly right. It would allow developers to test the new compsets without downloading all the NLDAS2 files for all months. Thanks for all your help -- I'll work on getting it all implemented.

jgfouca:
@jonbob , are you going to do your work on this branch or over in E3SM?

jonbob:
@jgfouca - I started doing it in E3SM, just so I can test it a little easier. Does that work for you?

jgfouca:
@jonbob , yes. Just make sure I know when it's done. I'll redo this split when it's done.

jonbob:
@jgfouca - OK, I'm trying to get it done this week while things are relatively quiet

jgfouca:
@jonbob , any update on this? Should I try to do another split?

jonbob:
@jgfouca - I worked up a new branch over the past few weeks but need to run the changes past the developers. Maybe give me a week before another split, if you want to include this? Thanks

jgfouca:
@jonbob , that sounds fine. Thanks for the update.

jgfouca:
@jonbob , @rljacob , the cime split/merge cycle has been blocked for an extended period and I'm starting to get concerned.

jonbob:
Thanks for the reminder, @jgfouca . I'll push on the nldas PR and see if we can remove that issue from blocking this cycle. We're also working here to be able to update to the newer version of jra forcing that the NCAR cime supports

klindsay28:
FYI @jonbob, in https://github.com/ESMCI/cime/pull/3361 I re-implemented the JRA updates, to the maint-5.6 branch of CIME, to not change answers compared to previous versions, which I think is what E3SM developers requested.

Specifying DATM%JRA and DROF%JRA in the longname provides the JRA, I think v1.2, to 2016.

Updated JRA forcing is now available with DATM%JRA-1p4-2018, DROF%JRA-1p4-2018 in the longname.

With this longname convention, the same longname produces the same forcing, and it provides a mechanism to extend the forcing in time, and update the version of the forcing.

jonbob:
Thanks @klindsay28 -- that's perfect. I don't believe we have access to those changes yet, since the merging cycle has been blocked for a bit. But we're looking forward to that capability

jgfouca:
Any action item on my part?

jonbob:
I don't believe so -- I just need to re-orient myself with the changes I made and check with the developers who wanted the old NLDAS support

jgfouca:
@jonbob , update?

PR #3340, opened 2020-01-08
Implement automatic test traiging in e3sm test infrastructure

jgfouca:
Should have no impact on cesm, but I asked for a review anyway in case there was interest.

Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


rljacob:
What does this do?

jgfouca:
@rljacob , this is the automated triaging we talked about on the IG telecon a couple weeks before break. If we keep track of when tests transition into passing/failing states, when a test reports a fail, we can offer a list of merges that could be the culprit.

Issue #3151, opened 2019-06-20
walltime in workflows doesn't work with lsf


jedwards4b:
The walltime argument in config_workflow.xml needs to conform to the walltime_format in config_batch.xml

rljacob:
Is this because lsf requires "HH:MM:  and doesn't like "HH:MM:SS"?

jedwards4b:
yes

rljacob:
We had the same issue in the python-based testing.  In tests.py, there's a way to specify the max time for a suite:
```
    "e3sm_land_developer" : { 
        "time"  : "0:45:00",
        "tests" : ( 
```

Try to run that suite with an lsf queue and you get errors.   Attn:  @jgfouca

rljacob:
If you give slurm " 01:00"  will it call that 1 hour or 1 minute?

jedwards4b:
It seems to be system dependent and we have a field in config_batch.xml to deal with it but that field is not being used in this case. We should make sure that the walltime_format field is used consistantly https://github.com/ESMCI/cime/blob/master/config/cesm/machines/config_batch.xml#L64

rljacob:
So the convention in CIME config files should be to use HH:MM:SS and the library can use walltime_format to fit that string to what the queue wants?

jedwards4b:
yes

jgfouca:
Fixed in E3SM. Will arrive in the next split.

jgfouca:
https://github.com/E3SM-Project/E3SM/pull/3446

jgfouca:
This was fixed by a CIME split a while ago.

Issue #3235, opened 2019-09-10
different behavior for JOB_WALLCLOCK_TIME on cheyenne?


mnlevy1981:
I think this is a variant of #3163 that #3164 didn't address. On cheyenne, you used to be able to run

```
$ ./xmlchange JOB_WALLCLOCK_TIME=HH:MM
```

but now (as of early July, I suppose), `cheyenne` interprets that as `MM:SS`. From a user perspective, it would be nice to not have to retrain myself to add the extra `:00` and instead have CIME define a standard for this variable (e.g. `DD:HH:MM` with `DD` assumed to be 0 if omitted) and then translate it to the appropriate format via something in `config_machines.xml`. This will also make some scripts more portable from machine to machine and make it easier for users to move from one machine to another.

jedwards4b:
See variable walltime_format
https://github.com/ESMCI/cime/blob/master/config/cesm/machines/config_batch.xml#L64

PR #3342, opened 2020-01-09
Convert user requested walltime to match walltime_format.

fischer-ncar:
If the user requested walltime format is different than the config_batch.xml walltime_format, convert
to the correct format.

If user requested walltime is %H:%M, and walltime_format is %H:%M:%S, then add :00 to user walltime.
If user requested walltime is %H:%M:%S, and walltime format is %H:%M, then remove %S from user walltime.
If user requested walltime is the same format at walltime_format, or walltime_format is None, then do
nothing.

Test suite:  scripts_regression_tests, create_test SMS.f09_g17.A.cheyenne_intel --walltime 00:35
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #] #3235

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -221,7 +221,8 @@ class EnvBatch(EnvBase):
 
             walltime    = case.get_value("USER_REQUESTED_WALLTIME", subgroup=job) if case.get_value("USER_REQUESTED_WALLTIME", subgroup=job) else None
             force_queue = case.get_value("USER_REQUESTED_QUEUE", subgroup=job) if case.get_value("USER_REQUESTED_QUEUE", subgroup=job) else None
-            logger.info("job is {} USER_REQUESTED_WALLTIME {} USER_REQUESTED_QUEUE {}".format(job, walltime, force_queue))
+            walltime_format = case.get_value("walltime_format", subgroup=job) if case.get_value("walltime_format", subgroup=job) else None
+            logger.info("job is {} USER_REQUESTED_WALLTIME {} USER_REQUESTED_QUEUE {} WALLTIME_FORMAT {}".format(job, walltime, force_queue, walltime_format))
             task_count = int(jsect["task_count"]) if "task_count" in jsect else case.total_tasks
             walltime = jsect["walltime"] if ("walltime" in jsect and walltime is None) else walltime
             if "task_count" in jsect:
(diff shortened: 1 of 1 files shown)


scripts/lib/CIME/XML/env_batch.py
```diff
@@ -273,6 +274,18 @@ def set_job_defaults(self, batch_jobs, case):
                     walltime = specs[3]
 
                 walltime = self._default_walltime if walltime is None else walltime # last-chance fallback
+            else:
+                # Set the walltime to the correct walltime_format, if set.
+                # Assuming correct format is #H:#M or %H:%M:%S.
+                if walltime_format is not None:
+                    components=walltime.split(":")
+                    if len(components) > len(walltime_format.split(":")):
+                        walltime = components[0] + ":" + components[1]
```
mnlevy1981: You are implying that if a user provides more `:` than the format expects, then the format must expect exactly 2. I think you want

```
walltime = ':'.join(components[:len(walltime_format.split(":"))])
```

scripts/lib/CIME/XML/env_batch.py
```diff
@@ -273,6 +274,18 @@ def set_job_defaults(self, batch_jobs, case):
...
+                        logger.info(" Changing USER_REQUESTED_WALLTIME to {} to match walltime_format {}".format(walltime,walltime_format))
+                    if len(components) < len(walltime_format.split(":")):
+                        walltime=walltime + ":00"
```
mnlevy1981: Similarly, you are assuming `len(walltime_format.split(":"))` is just one greater than `len(components)`. Something like

```
walltime = walltime + ':' + ':'.join(["00"]*(len(walltime_format.split(":")) - len(components)))
```

seems more general.

fischer-ncar:
@mnlevy1981 How often is a user specifying days for a walltime?  The only current walltime_formats for
cesm and e3sm are %H:%M;:%S or %H:%M.  The help for create_newcase has the usual format as
HH:MM:SS.  I agree with the changes that you recommended to make things more general.

mnlevy1981:
@fischer-ncar:

> How often is a user specifying days for a walltime? 

I would guess somewhere between "not very often" and "never" :) I could be mis-remembering, but I thought in the past I've had access to machines that accepted `DD:HH:MM` as a valid walltime, though treated `DD=0` as default and assumed users entered `HH:MM` if it wasn't specified so it has stuck with me as a source of confusion when I see `HH:MM:SS`.

Note that I'm not claiming this is previous behavior of old CESM or CCSM versions, just what I'm recalling from submitting parallel jobs to the queue on my own.

>  I agree with the changes that you recommended to make things more general

Glad to hear that, the general comment was less important but I do think the specific suggests are better for the code.

rljacob:
Will this also fix #3151 ?

fischer-ncar:
This doesn't fix #3151.  The walltime for the short term archiver is still 0:20:00.

fischer-ncar:
I just pushed Mike's recommended changes.  Still waiting for the tests to finish.  Since I don't know of any
systems that use DD for walltime, I'm not going to address that issue.

Issue #3347, opened 2020-01-10
Clean up extension match for the sake of pop dd files


billsacks:
As noted here https://github.com/ESMCI/cime/pull/3346#pullrequestreview-341461981 I think we can clean up the logic. I'll do this.

PR #3346, opened 2020-01-10
Pop hist dd fix

jedwards4b:
pop dd history files were not identified by get_extension
Fixed problem, added a test

Test suite: scripts_regression_tests.py
Test baseline: 
Test namelist changes: 
Test status: bit for bit,

Fixes #3344 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

PR #3348, opened 2020-01-10
Remove redundant logic in _get_extension

billsacks:
Consider these cases:
- Search succeeds with .nc extension: then the search will succeed
  identically without the .nc extension
- Search fails with .nc extension: then my suggestion will be no
  different from the current behavior (we'll just skip the first, failed
  search)

So we don't need to do the search twice

(See also https://github.com/ESMCI/cime/pull/3346#pullrequestreview-341461981)

Test suite: scripts_regression_tests: underway
Test baseline: n/a
Test namelist changes: n/a
Test status: bit for bit

Fixes ESMCI/cime#3347

User interface changes?: N

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/archive_base.py
+++ b/scripts/lib/CIME/XML/archive_base.py
@@ -165,14 +165,9 @@ def _get_extension(model, filepath):
     ext_regexes.append(r'\w+')
 
     for ext_regex in ext_regexes:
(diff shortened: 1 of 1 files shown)


billsacks:
scripts_regression_tests is underway. I'll update the PR when I get a chance to check it.

billsacks:
For the record, `scripts_regression_tests` passed on cheyenne.

PR #3356, opened 2020-01-16
Allow writing T-forcing cplhist files even when running with SGLC

billsacks:
If the user has set the namelist option that writes the cpl aux hist
files needed for running a later T compset (histaux_l2x1yrg = .true.),
then do the prep_glc_accum and prep_glc_accum_avg calls that are needed,
even if running with a stub GLC model.

Test suite: scripts_regression_tests on cheyenne, and extensive manual
testing as noted below
Test baseline: 22afa61fe
Test namelist changes: none
Test status: bit for bit

Manual testing: In the context of CTSM at release-clm5.0.29, I ran the
following tests with this new branch and the maint-5.6 branch (at
22afa61fe); for each, I compared the last cism (if available), clm,
cpl.ha, cpl.hi and cpl.hl2x1yr_glc files to ensure they are bit-for-bit
identical to the baseline:

(1) Ensuring these changes do not change answers in a case with
    CISM2%NOEVOLVE:

 create_newcase with `--res f10_f10_musgs --compset I1850Clm50Sp`,
 then:

    ./xmlchange JOB_WALLCLOCK_TIME=2:00:00,STOP_OPTION=nyears,STOP_N=3,DOUT_S=FALSE,HIST_OPTION=nyears,HIST_N=1,AVGHIST_OPTION=nyears,AVGHIST_N=1
    ./case.setup
    echo "histaux_l2x1yrg = .true." >> user_nl_cpl

(2) Ensuring these changes do not change answers in a case with
    CISM2%EVOLVE:

 create_newcase with `--res f10_f10_musgs --compset I1850Clm50SpG`, then:

    ./xmlchange JOB_WALLCLOCK_TIME=2:00:00,STOP_OPTION=nyears,STOP_N=3,DOUT_S=FALSE,HIST_OPTION=nyears,HIST_N=1,AVGHIST_OPTION=nyears,AVGHIST_N=1
    ./case.setup
    echo "histaux_l2x1yrg = .true." >> user_nl_cpl

(3) Ensuring that the new functionality works as expected: a case with
    SGLC from this branch compared with a baseline case with
    CISM2%NOEVOLVE with GLC_TWO_WAY_COUPLING=FALSE

This was similar to (1), but the baseline for comparison had
GLC_TWO_WAY_COUPLING=FALSE, whereas the new case used a user compset
that had SGLC in place of CISM2%NOEVOLVE. (Note that, in the
baseline, it doesn't work to use SGLC with histaux_l2x1yrg = .true.;
that is the new functionality allowed by this PR.)

Fixes ESCMI/cime#3218

User interface changes?: N

Code review:

### DIFF
--- a/src/drivers/mct/main/prep_glc_mod.F90
+++ b/src/drivers/mct/main/prep_glc_mod.F90
@@ -116,6 +116,7 @@ contains
     logical                          :: esmf_map_flag ! .true. => use esmf for mapping
     logical                          :: iamroot_CPLID ! .true. => CPLID masterproc
     logical                          :: glc_present   ! .true. => glc is present
+    logical                          :: do_hist_l2x1yrg ! .true. => create aux files: l2x 1yr glc forcings
     character(CL)                    :: lnd_gnam      ! lnd grid
     character(CL)                    :: glc_gnam      ! glc grid
     type(mct_avect), pointer         :: l2x_lx
@@ -127,6 +128,7 @@ contains
     call seq_infodata_getData(infodata , &
          esmf_map_flag=esmf_map_flag   , &
          glc_present=glc_present       , &
+         histaux_l2x1yrg=do_hist_l2x1yrg, &
          lnd_gnam=lnd_gnam             , &
          glc_gnam=glc_gnam)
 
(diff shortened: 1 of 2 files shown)


src/drivers/mct/main/prep_glc_mod.F90
```diff
@@ -136,27 +138,32 @@ subroutine prep_glc_init(infodata, lnd_c2_glc)
...
+
+       l2x_lx => component_get_c2x_cx(lnd(1))
+       lsize_l = mct_aVect_lsize(l2x_lx)
+
+       allocate(l2gacc_lx(num_inst_lnd))
+       do eli = 1,num_inst_lnd
+          call mct_aVect_init(l2gacc_lx(eli), rList=seq_flds_l2x_fields_to_glc, lsize=lsize_l)
+          call mct_aVect_zero(l2gacc_lx(eli))
+       end do
+       l2gacc_lx_cnt = 0
```
billsacks: The code inside this conditional block is not new: I simply moved a subset of the code from the below conditional up here, into this new conditional which is executed under the additional condition of `do_hist_l2x1yrg` (in addition to the old condition of `glc_present .and. lnd_c2_glc`).

billsacks:
FYI @whlipscomb @Katetc @gunterl

billsacks:
@jedwards4b I just want to make sure this gets merged by the end of the day today, in time for the code freeze. Let me know if you'd like me to merge it myself.

PR #3358, opened 2020-01-17
Allow writing T-forcing cplhist files even when running with SGLC (for master)

billsacks:
If the user has set the namelist option that writes the cpl aux hist
files needed for running a later T compset (histaux_l2x1yrg = .true.),
then do the prep_glc_accum and prep_glc_accum_avg calls that are needed,
even if running with a stub GLC model.

(This PR is the analog of #3356, but this one is destined for master
rather than maint-5.6. Most of it needed to be reimplemented due to a
major refactor on master relative to maint-5.6, as well as fitting this
into new capabilities for ocean-ice coupling on master. I reran all
testing noted below on this version.)

Test suite: scripts_regression_tests on cheyenne, and extensive manual
testing as noted below
Test baseline: 44673e0b8
Test namelist changes: none
Test status: bit for bit

Manual testing: In the context of CTSM at ctsm1.0.dev081, I ran the
following tests with this new branch and master (at 44673e0b8); for
each, I compared the last cism (if available), clm, cpl.ha, cpl.hi and
cpl.hl2x1yr_glc files to ensure they are bit-for-bit identical to the
baseline:

(1) Ensuring these changes do not change answers in a case with
    CISM2%NOEVOLVE:

 create_newcase with `--res f10_f10_musgs --compset I1850Clm50Sp`,
 then:

    ./xmlchange JOB_WALLCLOCK_TIME=2:00:00,STOP_OPTION=nyears,STOP_N=3,DOUT_S=FALSE,HIST_OPTION=nyears,HIST_N=1,AVGHIST_OPTION=nyears,AVGHIST_N=1
    ./case.setup
    echo "histaux_l2x1yrg = .true." >> user_nl_cpl

(2) Ensuring these changes do not change answers in a case with
    CISM2%EVOLVE:

 create_newcase with `--res f10_f10_musgs --compset I1850Clm50SpG`, then:

    ./xmlchange JOB_WALLCLOCK_TIME=2:00:00,STOP_OPTION=nyears,STOP_N=3,DOUT_S=FALSE,HIST_OPTION=nyears,HIST_N=1,AVGHIST_OPTION=nyears,AVGHIST_N=1
    ./case.setup
    echo "histaux_l2x1yrg = .true." >> user_nl_cpl

(3) Ensuring that the new functionality works as expected: a case with
    SGLC from this branch compared with a baseline case with
    CISM2%NOEVOLVE with GLC_TWO_WAY_COUPLING=FALSE

This was similar to (1), but the baseline for comparison had
GLC_TWO_WAY_COUPLING=FALSE, whereas the new case used a user compset
that had SGLC in place of CISM2%NOEVOLVE. (Note that, in the
baseline, it doesn't work to use SGLC with histaux_l2x1yrg = .true.;
that is the new functionality allowed by this PR.)

Fixes ESCMI/cime#3218

User interface changes?: N

Code review:

### DIFF
--- a/src/drivers/mct/main/cime_comp_mod.F90
+++ b/src/drivers/mct/main/cime_comp_mod.F90
@@ -213,6 +213,7 @@ module cime_comp_mod
   private :: cime_run_lnd_setup_send
   private :: cime_run_lnd_recv_post
   private :: cime_run_glc_setup_send
+  private :: cime_run_glc_accum_avg
   private :: cime_run_glc_recv_post
   private :: cime_run_rof_setup_send
   private :: cime_run_rof_recv_post
@@ -2331,6 +2332,7 @@ contains
     type(ESMF_Time)       :: etime_curr           ! Current model time
     real(r8)              :: tbnds1_offset        ! Time offset for call to seq_hist_writeaux
     logical               :: lnd2glc_averaged_now ! Whether lnd2glc averages were taken this timestep
+    logical               :: prep_glc_accum_avg_called ! Whether prep_glc_accum_avg has been called this timestep
 
 101 format( A, i10.8, i8, 12A, A, F8.2, A, F8.2 )
 102 format( A, i10.8, i8, A, 8L3 )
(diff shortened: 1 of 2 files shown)


billsacks:
@jedwards4b this is conceptually the same as #3356, which you approved, but the details of the implementation differ due to massive changes on master in the relevant pieces of code.

@jonbob I want to give you or someone else on E3SM the opportunity to review this in case you think it might have any issues on your end - though I don't foresee any.

jedwards4b:
@jonbob this is waiting for your review...

PR #3368, opened 2020-01-24
add timeouts for ftp and wget

### LINKED ISSUE #3365: Attempt to download inputdata files with wget is hanging
The check_input_data script is hanging when attempting to check the server with protocol wget.

This problem was originally reported in the CESM forums: https://xenforo.cgd.ucar.edu/cesm/threads/downloading-input-data-b1850-compset-no-server-response.4999/ . I have reproduced it on cheyenne as follows:

```
git clone https://github.com/ESCOMP/CESM.git
cd CESM
git checkout cesm2.1-alphabranch
./manage_externals/checkout_externals
cd cime/scripts/
./create_newcase --case /glade/scratch/sacks/test_inputdata_0124a --compset B1850 --res f09_g17
cd /glade/scratch/sacks/test_inputdata_0124a
mkdir /glade/scratch/sacks/cesm_test_inputdata
./xmlchange DIN_LOC_ROOT=/glade/scratch/sacks/cesm_test_inputdata
./case.setup
./preview_namelists
```

The last command produces:

```
Setting resource.RLIMIT_STACK to -1 from (307200000, -1)
Refcase not found in /glade/scratch/sacks/cesm_test_inputdata/cesm2_init/b.e20.B1850.f09_g17.pi_control.all.299_merge_v3/0134-01-01, will attempt to download from inputdata
Checking server ftp://gridanon.cgd.ucar.edu:2811/cesm/inputdata/ with protocol gftp
Setting resource.RLIMIT_STACK to -1 from (-1, -1)
Client protocol gftp not enabled
Checking server ftp://ftp.cgd.ucar.edu/cesm/inputdata with protocol wget
Setting resource.RLIMIT_STACK to -1 from (-1, -1)
Using protocol wget with user anonymous and passwd user@example.edu
```

then hangs.

It's possible that the solution to this problem is simply to reboot the system used with wget; that would address the "CRITICAL" nature of this issue.

However, I wonder if it would be possible for the check_input_data script to handle timeouts more gracefully. Now that we support more inputdata servers, it seems like it will be more common for one of them to be down. If the data can be retrieved from the servers that are up, then it seems like the scripts should gracefully handle an unnecessary server being down, and just move on without complaint. (But maybe I'm misunderstanding the problem here, or what I'm suggesting is harder than I think.)

jedwards4b:
Test baseline: master
Test namelist changes: 
Test status: bit for bit

Fixes #3365 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/Servers/ftp.py
+++ b/scripts/lib/CIME/Servers/ftp.py
@@ -6,33 +6,44 @@ from CIME.XML.standard_module_setup import *
 from CIME.Servers.generic_server import GenericServer
 from ftplib import FTP as FTPpy
 from ftplib import all_errors as all_ftp_errors
-
+import socket
 
 logger = logging.getLogger(__name__)
 # I think that multiple inheritence would be useful here, but I couldnt make it work
 # in a py2/3 compatible way.
 class FTP(GenericServer):
-    def __init__(self, address, user='', passwd=''):
-        ftp_server, root_address = address.split('/', 1)
-        logger.info("server address {} root path {}".format(ftp_server, root_address))
-        self.ftp = FTPpy(ftp_server)
+    def __init__(self, address, user='', passwd='', server=None):
         if not user:
             user = ''
         if not passwd:
             passwd = ''
-
+        expect(server," Must call via ftp_login function")
+        root_address = address.split('/', 1)[1]
+        self.ftp = server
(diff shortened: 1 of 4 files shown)


jgfouca:
See #3370

PR #3373, opened 2020-01-27
move orbital updates to nuopc caps and update nuopc dmodels

mvertens:
move orbital updates to nuopc caps and update nuopc dmodels

This PR accomplishes two things.

1. The orbital updates have now been moved to every model cap that requires an orbital calculation. This is required since the run sequence in nuopc is not done in hard-wired code (like cpl7) but rather in an run sequence ascii file. So calls to update the orbital calculation cannot be done in one place (like in the cpl7 driver. 
2. All mct references have been removed from the nuopc data models. New data structures have been created in order to map stream data to model data (rather than using the translate functions of the mct.  This is the first step in replacing the mct dependence in the nuopc data models with ESMF.
3. This PR must be accompanied by a CMEPS PR (for the cmeps hash 8716a2c).

Test suite: scripts_regressions_tests were run for mct - all passed.
Test baseline: 
Test namelist changes: none
Test status: bit for bit(for mct)
Fixes: none
User interface changes?: no
Code review:


src/components/data_comps/dshr_nuopc/dshr_dfield_mod.F90
```diff
@@ -0,0 +1,292 @@
...
+    ! note that if the attribute vector is not found in the streams
+
+    ! loop over all input streams
+    do ns = 1, sdat%nstreams
+       ! determine if the strm_fld is in the attribute vector of stream ns
+       kf  = mct_aVect_indexRA(sdat%avs(ns), trim(strm_fld), perrWith='quiet')
+       if (kf > 0) then
+          dfield_new%sdat_stream_index = ns
+          dfield_new%sdat_avect_index = kf
+          allocate(dfield_new%stream_data1d(lsize))
```
jedwards4b: Allocate statements should use the optional stat variable and explicitly check for errors.

src/components/data_comps/dshr_nuopc/dshr_dfield_mod.F90
```diff
@@ -0,0 +1,292 @@
...
+
+    ! local variables
+    type(dfield_type), pointer :: dfield_new
+    integer :: ns, kf, lsize
+    character(len=*), parameter :: subname='(dfield_add_1d)'
+    ! ----------------------------------------------
+
+    rc = ESMF_SUCCESS
+
+    allocate(dfield_new)
```
jedwards4b: allocate should use stat and trap errors

src/components/data_comps/dshr_nuopc/dshr_dfield_mod.F90
```diff
@@ -0,0 +1,292 @@
...
+    ! local variables
+    type(dfield_type), pointer :: dfield_new
+    integer :: n, i, kf, ns, nf
+    integer :: nflds, lsize, num
+    character(len=*), parameter :: subname='(dfield_add_2d)'
+    ! ----------------------------------------------
+
+    rc = ESMF_SUCCESS
+
+    allocate(dfield_new)
```
jedwards4b: allocate should use stat

mvertens: done

mvertens:
All allocations were done as part of a new linked list element - so since
that element is new all allocations are new.

PR #3375, opened 2020-01-27
fix issue with resubmit on systems that use ssh

### LINKED ISSUE #3374: resubmit not working on systems where submit not allowed on compute nodes
The ssh command used on TACC systems for resubmit is not working, it currently results in a failure after ssh timeout.

jedwards4b:
Fix issue with resubmit when ssh to login is used.

Test suite: scripts_regression_tests.py on stampede and cheyenne
Test baseline: 
Test namelist changes: 

Fixes #3374 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -690,9 +690,12 @@ class EnvBatch(EnvBase):
             sequence = (batchsubmit, submitargs, batchredirect, get_batch_script_for_job(job), run_args)
 
         submitcmd = " ".join(s.strip() for s in sequence if s is not None)
+        if submitcmd.startswith("ssh"):
+            submitcmd = submitcmd.replace("cd $CASEROOT","\'cd $CASEROOT") + "\'"
         if dry_run:
             return submitcmd
         else:
+            submitcmd = case.get_resolved_value(submitcmd)
             logger.info("Submitting job script {}".format(submitcmd))
             output = run_cmd_no_fail(submitcmd, combine_output=True)
             jobid = self.get_job_id(output)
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/XML/env_batch.py
```diff
@@ -690,9 +690,12 @@ def _submit_single_job(self, case, job, dep_jobs=None, allow_fail=False,
             sequence = (batchsubmit, submitargs, batchredirect, get_batch_script_for_job(job), run_args)
 
         submitcmd = " ".join(s.strip() for s in sequence if s is not None)
+        if submitcmd.startswith("ssh"):
+            submitcmd = submitcmd.replace("cd $CASEROOT","\'cd $CASEROOT") + "\'"
```
jgfouca: Are the escape characters necessary here? Python has good support for nesting quotes:
```
s = " 'hi' "
```

jgfouca: Also, I'm a bit confused that the second single quote happens outside of the replace. I think this would be more clear:
```
if submitcmd.startswith("ssh"):
    stmt_to_quote = "cd $CASEROOT"
    submitcmd = submitcmd.replace(stmt_to_quote, "'{}'".format(stmt_to_quote))
```

jedwards4b: I'm looking into removing the escape chars.

The second quote is outside of the replace because there is more than just cd $CASEROOT that needs to be in quotes the rest of the submitcmd needs to be in it.

jedwards4b: @jgfouca The escapes are required.

PR #3377, opened 2020-01-27
Update nuopc esmf

jedwards4b:
Update cheyenne nuopc esmf to 8.1.0b05, fix bug in env_batch/env_workflow.py which caused an error in xmlquery --listall

Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


jgfouca:
So we're removing support for PROJECT_REQUIRED?

rljacob:
Yeah what was the bug?  This seems like the wrong fix.

jgfouca:
It looks like maybe he thought about moving PROJECT_REQUIRED to env_workflow.xml. But the driver config_component.xml files still have it going to env_batch.xml.

jedwards4b:
I see what happened, in mct it still goes in env_batch.xml but in nuopc it goes in env_workflow.xml
I think I had a reason for moving it but can't remember now.  I'll move it back until I can recall again.

jedwards4b:
We found some significant differences with this change, we are backing it out until we can determine the difference.

PR #3382, opened 2020-01-30
Update intel and cprnc on izumi

fischer-ncar:
Old intel compilers on Izumi were removed, so needed to update to newer 19.0.2 compiler.
cprnc was also rebuilt and had a new filename.

Test suite: prealpha tests on izumi.
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -1375,7 +1375,7 @@ This allows using a different mpirun command to launch unit tests
     <DIN_LOC_ROOT_CLMFORC>/project/tss</DIN_LOC_ROOT_CLMFORC>
     <DOUT_S_ROOT>/scratch/cluster/$USER/archive/$CASE</DOUT_S_ROOT>
     <BASELINE_ROOT>/fs/cgd/csm/ccsm_baselines</BASELINE_ROOT>
-    <CCSM_CPRNC>/fs/cgd/csm/tools/cime/tools/cprnc/cprnc.izumi</CCSM_CPRNC>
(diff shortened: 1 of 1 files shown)


fischer-ncar:
PGI looks to be messed up on Izumi.  I haven't been able to get it to work.  I'm going to address the pgi
compiler after release testing, since I can still test pgi on hobart.

gold2718:
I think using both "PGI" and "messed up" in a sentence is redundant.

config/cesm/machines/config_machines.xml
```diff
@@ -1375,7 +1375,7 @@ This allows using a different mpirun command to launch unit tests
     <DIN_LOC_ROOT_CLMFORC>/project/tss</DIN_LOC_ROOT_CLMFORC>
     <DOUT_S_ROOT>/scratch/cluster/$USER/archive/$CASE</DOUT_S_ROOT>
     <BASELINE_ROOT>/fs/cgd/csm/ccsm_baselines</BASELINE_ROOT>
-    <CCSM_CPRNC>/fs/cgd/csm/tools/cime/tools/cprnc/cprnc.izumi</CCSM_CPRNC>
```
billsacks: @fischer-ncar  I think this will break cprnc on hobart, since now it looks like they are both using the same executable. I don't use hobart anymore myself, but this should be fixed if anyone does use hobart for testing.

jedwards4b: I just tried it on hobart and it seems to work fine.

PR #3387, opened 2020-02-04
DO_NOT_MERGE: new separation of nuopc and mct component directories

mvertens:
new separation of nuopc and mct component directories for data, xcpl and stub

This PR creates a new cime/src/components/ directory structure permitting the complete separation of  nuopc/ and mct/ code for data, dead and stub components. This is needed now since the nuopc and mct cime components will effectively share not common code once the data model share code is reimplemented with ESMF (a proccess that is occurring now). 

The e3sm and ufs config_files.xml were modified to account for this change - but testing was only done for cesm.

Test suite: scripts_regression_tests and cesm prealpha tests on cheyenne
Test baseline: cesm2_2_alpha04c (for cesm)
Test namelist changes: no

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


src/components/nuopc/data_comps/dice/cime_config/config_archive.xml
```diff
@@ -0,0 +1,15 @@
+<components version="2.0">
+  <comp_archive_spec compname="dice" compclass="ice">
+    <rest_file_extension>r</rest_file_extension>
```
jedwards4b: My testing indicated that you also needed the rs1 listed as a rest_file_extension.

rljacob:
I'd like to keep the components directory flat.  So instead of adding "mct" and "nuopc" subdirs, I'd like to see those added as additional names on the existing directories.  So you would have
```
components/data_comps_mct
components/data_comps_nuopc
components/stub_comps_mct
components/stub_comps_nuopc
etc.
```

mvertens:
@rjacob - I'm okay with that.

PR #3397, opened 2020-02-12
hist_utils tests using cprnc doesnt work on binary files

### LINKED ISSUE #3396: non-netcdf history files cause a fail in testing baselines
For example pop saves a binary file pop.dd.  - currently testing will call cprnc with this file which causes a fail since it is not netcdf.

jedwards4b:
Non-netcdf files should not be compared with cprnc - look for the .nc extension on history files and only compare files that have that extension.

Test suite: scripts_regression_tests.py + hand testing of SMS.T62_g37.C.cheyenne_intel.pop-default
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3396 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -62,6 +62,9 @@ def copy_histfiles(case, suffix):
         num_copied += len(test_hists)
         for test_hist in test_hists:
             test_hist = os.path.join(rundir,test_hist)
+            if not test_hist.endswith('.nc'):
+                logger.info("Will not compare non-netcdf file {}".format(test_hist))
+                continue
             new_file = "{}.{}".format(test_hist, suffix)
             if os.path.exists(new_file):
                 os.remove(new_file)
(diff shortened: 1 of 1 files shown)


scripts/lib/CIME/hist_utils.py
```diff
@@ -246,6 +249,9 @@ def _compare_hists(case, from_dir1, from_dir2, suffix1="", suffix2="", outfile_s
         num_compared += len(match_ups)
 
         for hist1, hist2 in match_ups:
+            if not '.nc.' in hist1:
```
billsacks: Wouldn't this fail if hist1 didn't have a suffix, and so ended with just `.nc` rather than `.nc.suffix1`?

I'm wondering if this should be changed to simply `if not '.nc' in hist1:`?

rljacob:
Can you change the title?  I thought this was a mod to cprnc but its a mod to hist_utils.py.

scripts/lib/CIME/hist_utils.py
jedwards4b: fixed, thanks

PR #3400, opened 2020-02-14
Wpcoomb/improve cprnc build process

### LINKED ISSUE #2457: Build cprnc as part of create_test
It seems like the python implementation of cprnc (#1031 ) has stalled out (due to no fault of @mfdeakin-sandia 's -- we just hit some roadblocks with this, particularly related to netcdf dependencies). It occurred to me: The primary motivation for that python implementation was to avoid needing to maintain a build of cprnc on all of the systems we test on (users need to pre-build this on their system, we need to keep this up-to-date... which we're often not very good about... etc.).

Now that the cprnc build is more robust, can we simply build cprnc as part of each `create_test` instantiation? I feel that would make the Fortran version of cprnc a sustainable solution.

### LINKED ISSUE #3389: Improve cprnc build process
The "cprnc" tool is required for CIME regression testing.  I recently tried to install on a new platform and found that the README was not accurate and I had to examine the internals of the Makefile in order to figure out how to build it.  Since CIME is being ported to a lot of different platforms (e.g., for the UFS release), the build of cprnc is currently a major liability--i.e., it may be hard to a regular user to install it and therefore the entire regression testing process would not be viable.  

Recommendations:

1.  Since CIME is capable of building entire models, the build of cprnc should be simplified and standardized, e.g., ./configure && make.  Generation of the env_machine_specific.sh should be hidden entirely from the user and sensible defaults (e.g., mpi-serial) should be preset.

2.  Consider automated build of cprnc when the regression tests are run.  In CIME we already have the full machine/compiler information needed.  Maintaining cprnc in a fixed place has advantages, but it should be straightforward to check if cprnc is already built and if not, to build it in a temporary location using the case compiler info, and use that for the regression tests.  This would completely remove the need for users to know about/install cprnc.

WesCoomber:
Improved cprnc build process for 
better installs on new platforms.
This PR adds functionality to build cprnc on the fly as part of create_test if it doesn't already exist. One can skip defining the xml entry, CCSM_CPRNC, and the sharedlib build system will automatically build cprnc once for each compiler, with DEBUG set to FALSE.

Test baseline: 
Test namelist changes: 
Test status: 

Fixes #3389
Fixes #2457

User interface changes?: N

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -247,6 +247,10 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
     if uses_kokkos(case):
         libs.append("kokkos")
 
+    cprnc_loc = case.get_value("CCSM_CPRNC")
+    if not cprnc_loc or not os.path.exists(cprnc_loc):
+        libs.append("cprnc")
+
     logs = []
     sharedlibroot = os.path.abspath(case.get_value("SHAREDLIBROOT"))
     for lib in libs:
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -247,12 +247,10 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
     if uses_kokkos(case):
         libs.append("kokkos")
 
+    # Check if we need to build our own cprnc
     cprnc_loc = case.get_value("CCSM_CPRNC")
     if not cprnc_loc or not os.path.exists(cprnc_loc):
         libs.append("cprnc")
-    #libs.append("cprnc")
-    logger.info("wpc:: appending cprnc lib to the libs list")
-
 
     logs = []
     sharedlibroot = os.path.abspath(case.get_value("SHAREDLIBROOT"))
@@ -267,7 +265,7 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
             full_lib_path = os.path.join(sharedlibroot, sharedpath, "mct", lib)
         elif lib == "cprnc":
             full_lib_path = os.path.join(sharedlibroot, sharedpath, lib)
-            case.set_value("CCSM_CPRNC", os.path.join(full_lib_path, "/bin/cprnc"))
+            case.set_value("CCSM_CPRNC", os.path.join(full_lib_path, "cprnc"))
         else:
             full_lib_path = os.path.join(sharedlibroot, sharedpath, lib)
         # pio build creates its own directory
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -264,7 +264,7 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
         elif lib == "mpi-serial":
             full_lib_path = os.path.join(sharedlibroot, sharedpath, "mct", lib)
         elif lib == "cprnc":
-            full_lib_path = os.path.join(sharedlibroot, sharedpath, lib)
+            full_lib_path = sharedlibroot
             case.set_value("CCSM_CPRNC", os.path.join(full_lib_path, "cprnc"))
         else:
             full_lib_path = os.path.join(sharedlibroot, sharedpath, lib)
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -250,6 +250,9 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
     cprnc_loc = case.get_value("CCSM_CPRNC")
     if not cprnc_loc or not os.path.exists(cprnc_loc):
         libs.append("cprnc")
+    #libs.append("cprnc")
+    logger.info("wpc:: appending cprnc lib to the libs list")
+
 
     logs = []
     sharedlibroot = os.path.abspath(case.get_value("SHAREDLIBROOT"))
@@ -262,6 +265,9 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
             full_lib_path = os.path.join(sharedlibroot, sharedpath)
         elif lib == "mpi-serial":
             full_lib_path = os.path.join(sharedlibroot, sharedpath, "mct", lib)
+        elif lib == "cprnc":
+            full_lib_path = os.path.join(sharedlibroot, sharedpath, lib)
+            case.set_value("CCSM_CPRNC", os.path.join(full_lib_path, "/bin/cprnc"))
         else:
             full_lib_path = os.path.join(sharedlibroot, sharedpath, lib)
         # pio build creates its own directory
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -248,9 +248,10 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
         libs.append("kokkos")
 
     # Check if we need to build our own cprnc
-    cprnc_loc = case.get_value("CCSM_CPRNC")
-    if not cprnc_loc or not os.path.exists(cprnc_loc):
-        libs.append("cprnc")
+    if case.get_value("TEST"):
+        cprnc_loc = case.get_value("CCSM_CPRNC")
+        if not cprnc_loc or not os.path.exists(cprnc_loc):
+            libs.append("cprnc")
 
     logs = []
     sharedlibroot = os.path.abspath(case.get_value("SHAREDLIBROOT"))
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -266,7 +266,7 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
             full_lib_path = os.path.join(sharedlibroot, sharedpath, "mct", lib)
         elif lib == "cprnc":
             full_lib_path = os.path.join(sharedlibroot, compiler, "cprnc")
-            case.set_value("CCSM_CPRNC", os.path.join(full_lib_path))
+            case.set_value("CCSM_CPRNC", os.path.join(full_lib_path, "cprnc"))
         else:
             full_lib_path = os.path.join(sharedlibroot, sharedpath, lib)
         # pio build creates its own directory
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -265,8 +265,8 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
         elif lib == "mpi-serial":
             full_lib_path = os.path.join(sharedlibroot, sharedpath, "mct", lib)
         elif lib == "cprnc":
-            full_lib_path = sharedlibroot
-            case.set_value("CCSM_CPRNC", os.path.join(full_lib_path, "cprnc"))
+            full_lib_path = os.path.join(sharedlibroot, compiler, "cprnc")
+            case.set_value("CCSM_CPRNC", os.path.join(full_lib_path))
         else:
             full_lib_path = os.path.join(sharedlibroot, sharedpath, lib)
         # pio build creates its own directory
--- a/src/build_scripts/buildlib.cprnc
+++ b/src/build_scripts/buildlib.cprnc
@@ -3,7 +3,7 @@
 from standard_script_setup import *
 from CIME.utils import expect, run_bld_cmd_ensure_logging, run_cmd_no_fail, run_cmd
 from CIME.case import Case
-from CIME.build import get_standard_makefile_args
+from CIME.build import get_standard_cmake_args
 
 logger = logging.getLogger(__name__)
 
@@ -44,15 +44,23 @@ formatter_class=argparse.ArgumentDefaultsHelpFormatter
 ###############################################################################
 def buildlib(bldroot, installpath, case):
 ###############################################################################
-    import pdb
-    pdb.set_trace()
-    make_args = get_standard_makefile_args(case, shared_lib=True)
-    sfc = run_cmd_no_fail("make -f Macros.make {} -p | grep SFC".format(make_args)).split(":=")[-1].strip()
-    np = run_cmd_no_fail("make -f Macros.make {} -p | grep NETCDF_PATH".format(make_args)).split(":=")[-1].strip()
+    caseroot = case.get_value("CASEROOT")
+    cimeroot = case.get_value("CIMEROOT")
+
+    # Generate macros and environment
+    run_bld_cmd_ensure_logging("{}/tools/configure --mpilib=mpi-serial --macros-format=CMake".format(cimeroot), logger, from_dir=bldroot)
+
+    cmake_args = get_standard_cmake_args(case, shared_lib=True)
     cr = "$CIMEROOT/tools/cprnc"
 
-    cmake_cmd = "cmake -DNETCDF_PATH={netcdf_path} -DCMAKE_Fortran_COMPILER_ID={sfc} {cr} -DCMAKE_PREFIX_PATH{dest_path}".format(netcdf_path=np, sfc=sfc, cr=cr, destpath=installpath)
+    cmake_cmd = "CIMEROOT={cimeroot} . .env_mach_specific.sh && cmake {cmake_args} -DMPILIB=mpi-serial -C Macros.cmake {cimeroot}/tools/cprnc -DCMAKE_PREFIX_PATH={dest_path}".\
+        format(cimeroot=cimeroot, dest_path=installpath, caseroot=caseroot, cmake_args=cmake_args)
     run_bld_cmd_ensure_logging(cmake_cmd, logger, from_dir=bldroot)
+
+    gmake_cmd = case.get_value("GMAKE")
+    gmake_j = case.get_value("GMAKE_J")
+
+    run_bld_cmd_ensure_logging("{} -j {}".format(gmake_cmd, gmake_j), logger, from_dir=bldroot)
 
 def _main(argv, documentation):
     bldroot, installpath, caseroot = parse_command_line(argv, documentation)
--- a/src/build_scripts/buildlib.cprnc
+++ b/src/build_scripts/buildlib.cprnc
@@ -51,7 +51,7 @@ def buildlib(bldroot, installpath, case):
 
     cmake_args = get_standard_cmake_args(case, shared_lib=True)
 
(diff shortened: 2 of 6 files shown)


rsdunlapiv:
This is fantastic!  Can we clean up the variable name CCSM_CPRNC since cprnc is not tied to CESM?  Something like CPRNC_EXE?  Old name would be deprecated but supported for backward compatibility.

jedwards4b:
@WesCoomber you have a few pylint issues to deal with:
```

FAIL: test_pylint_src_build_scripts_buildlib_cprnc (__main__.B_CheckCode)

  File "./scripts_regression_tests.py", line 3327, in test

    self.assertTrue(result == "", msg=result)

AssertionError: False is not true : ************* Module buildlib

W: 47, 4: Unused variable 'caseroot' (unused-variable)

W:  4, 0: Unused run_cmd_no_fail imported from CIME.utils (unused-import)

W:  4, 0: Unused expect imported from CIME.utils (unused-import)

W:  4, 0: Unused run_cmd imported from CIME.utils (unused-import)
```

WesCoomber:
@jedwards4b Got it. I fixed the pylint issues and I am re-running the scripts_regression_tests.py locally on Melvin.

jedwards4b:
@billsacks Can you please review this morning?

tools/cprnc/CMakeLists.txt
```diff
@@ -1,24 +1,19 @@
...
 SET(CMAKE_MODULE_PATH ${CMAKE_MODULE_PATH} $ENV{CIMEROOT}/src/externals/pio2/cmake)
 
-# Defining NETCDF_LIBRARIES and NETCDF_INCLUDE_DIR enables bypassing
-#   the search for Netcdf
-IF (NOT DEFINED NETCDF_LIBRARIES OR NOT DEFINED NETCDF_INCLUDE_DIR)
-  SET(NETCDF_INCLUDE_DIR ${NETCDF_PATH}/include)
-  SET(NETCDF_LIBRARIES ${NETCDF_PATH}/lib/libnetcdff.a)
-ENDIF ()
-
 ADD_CUSTOM_COMMAND(
```
billsacks: This is a pre-existing issue, so I'm okay deferring it to a separate issue, though it feels more important now that cprnc may be built all the time as part of testing: I believe that the `ADD_CUSTOM_COMMAND` command should have a `DEPENDS` clause:

```
DEPENDS ${PROJECT_SOURCE_DIR}/compare_vars_mod.F90.in ${PROJECT_SOURCE_DIR}/../../src/externals/genf90/genf90.pl
```

Without that, `compare_vars_mod.F90` isn't regenerated even if its dependencies change.

Maybe connected with this: I don't understand the purpose of the `ADD_CUSTOM_TARGET` command on line 23. Does anyone understand why that's there and whether it's actually needed?

Again, I'm happy to open a separate issue for this if you don't want to muddy up this PR with this change.

tools/cprnc/CMakeLists.txt
```diff
@@ -53,9 +48,6 @@ SET (CPRNC_SRCS
...
 
-TARGET_LINK_LIBRARIES(cprnc ${NETCDF_LIBRARIES})
-
-INSTALL(TARGETS cprnc
-        RUNTIME DESTINATION bin)
-
-execute_process(COMMAND ../configure --clean WORKING_DIRECTORY ${CMAKE_CURRENT_LIST_DIR})
+# SLIBS comes from Macros.cmake
+separate_arguments(SLIBS_LIST UNIX_COMMAND "${SLIBS}")
+TARGET_LINK_LIBRARIES(cprnc ${SLIBS_LIST})
```
billsacks: It looks like the assumption here is that the netcdf libraries are listed in SLIBS. This doesn't seem to be the case in general, at least for CESM machines. I think we need a more robust / general way of getting the netcdf libraries.

scripts/lib/CIME/build.py
```diff
@@ -258,6 +263,9 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
             full_lib_path = os.path.join(sharedlibroot, sharedpath)
         elif lib == "mpi-serial":
             full_lib_path = os.path.join(sharedlibroot, sharedpath, "mct", lib)
+        elif lib == "cprnc":
+            full_lib_path = os.path.join(sharedlibroot, sharedpath, lib)
```
jedwards4b: I suggest that here full_lib_path = sharedlibroot.  This will avoid building more copies of cprnc than you need.

WesCoomber: Good idea. I implemented your suggestion to build cprnc higher up.

billsacks:
@WesCoomber thanks for the change to the directory in which cprnc is built. Maybe you're still working on this, but as I note in the first bullet point of my PR, I think this change needs to be combined with overriding some build options:

(1) Debug should be set to false

(2) Either (a) the compiler should be set to the default compiler for the machine or (b) we should continue to do a different build for each compiler in a test suite

The way it currently stands, it is unpredictable which compiler and debug option would be used to build cprnc in a suite that uses multiple compilers and debug on/off.

However, I just realized that there may be an issue with switching the compiler used for a single library, unless the module environment gets reloaded for each library. So this could argue for using (2b), which feels like maybe the safest thing to do anyway. This could theoretically be an issue for the debug option, too: I see that, for CESM, we do slightly different module loads (for ESMF) depending on whether we're using debug or not. But in practice, I don't think this will cause a problem for the debug option, so it's probably safe to use the test's environment even if we change the debug flag for the sake of building esmf.

jedwards4b:
@billsacks Regarding your point 2 - as long as the rpath option is used when linking shared libraries you shouldn't need any special module environment for cprnc to work.  You can build it once with the first compiler in the list and it will work for all cases of model environments.

billsacks:
> @billsacks Regarding your point 2 - as long as the rpath option is used when linking shared libraries you shouldn't need any special module environment for cprnc to work. You can build it once with the first compiler in the list and it will work for all cases of model environments.

But I'm partly concerned with predictability / reproducibility: if the compiler used to build cprnc depends on the order of tests in the test suite, that could cause some hard-to-track-down problems. For example, I'm imagining someone who first runs a test suite that has tests across multiple compilers, then reruns just a subset of the tests that only includes a single compiler. It would be confusing if the cprnc build was done differently for a given test depending on what other tests are in that test suite. (At best it would lead to roundoff-level differences in the cprnc results; at worst it could lead to a failed build in one run of the test suite but a successful build in another.) For this reason, I suggest making the compiler used to build cprnc predictable: either the compiler of the given test, or the default compiler for the machine regardless of what compiler is used for the given test.

jedwards4b:
I would agree to the default compiler for the system.

rljacob:
On many machines with multiple compiles, we make "cprnc" a script that first loads any needed modules and then executes the corresponding binary called something like "cprnc.exe".  Maybe the build should always have a script like that as its target?

billsacks:
> On many machines with multiple compiles, we make "cprnc" a script that first loads any needed modules and then executes the corresponding binary called something like "cprnc.exe". Maybe the build should always have a script like that as its target?

@rljacob - For the sake of building cprnc as part of the test suite, I'm having trouble seeing the advantage of that approach over building a separate cprnc for each compiler (in the sharedlib area) and pointing each test to the appropriate compiler-specific cprnc. But maybe I'm missing something.

> I would agree to the default compiler for the system.

@jedwards4b and I just discussed this. I think this will only work if you make some changes so that you load the correct module environment for the default compiler, rather than the module environment given in `env_mach_specific.sh` (which is the module environment for the compiler being used in this test). If you can see a way to do that without too much trouble, that could be a good approach. But doing a separate build for each compiler also seems acceptable, and may be significantly easier.

jgfouca:
@jedwards4b , @billsacks ,

I don't think CIME has good support for changing key settings like compiler, debug, etc midway through a `case.build --sharedlib-only`. Cprnc is not expensive to build, so I think the best thing is probably for Wesley to restore the placement of cprnc to a fully-qualified sharelib path.

jedwards4b:
I agree that changing compiler is complicated but Debug should be easy.   how about building cprnc once for each compiler - not the fully qualified path but sharedlibroot/compiler?

billsacks:
> I agree that changing compiler is complicated but Debug should be easy. how about building cprnc once for each compiler - not the fully qualified path but sharedlibroot/compiler?

I agree with @jedwards4b . It looks like you're already changing the mpi library to use mpi-serial, so can you similarly force use of debug=False? (I'm not sure if you can do it with the same mechanism or need a different mechanism for that.)

jgfouca:
I see. Yes, we can easily force debug=false in buildlib.cprnc.

tools/cprnc/CMakeLists.txt
jgfouca: How does CESM get netcdf libs?

billsacks: I think based on logic in the Makefile, at least for some machines. I can't comment on what the vision is for how it *should* be done, though.

WesCoomber:
@billsacks Hi Bill, 

I documented the mechanism for a machine not having a pre-installed cprnc. If the machine does not define CCSM_CPRNC (or points to a non-existent path) in the .xml we use the shared lib mechanism to build cprnc once for each compiler. 

I also have forced the use of DEBUG=FALSE in the CMake command called in buildlib.cprnc. Jim helped fix the macro issue to turn on the compiler optimizations.

tools/cprnc/CMakeLists.txt
WesCoomber: Ok I added the suggested `DEPENDS` clause and removed the mystery `ADD_CUSTOM_TARGET` command. Jim Foucar and I did not understand that `ADD_CUSTOM_TARGET` command too. I am running the regression tests with these changes.

tools/cprnc/CMakeLists.txt
jedwards4b: The problem is every system does it differently, and for a while even every version of netcdf did it differently - though that's long enough in the past that I'm not seeing the problem anymore.   On cheyenne for example when you load the netcdf module it adds the include path and the link steps to the mpi compiler wrappers so that you really don't need to explicitly do anything, on tacc systems they define TACC_NETCDF_PATH and on cray they define CRAY_NETCDF_PATH so then you need to translate that to what the model wants - usually that translation happens in config_compilers.xml - then there is cmake that wants a FindNetcdf.cmake file of which there are a lot out there and non of them are standard.   Oh if only it could be standard...

jedwards4b:
@WesCoomber great job - thanks!

jedwards4b:
@billsacks do you want to review this again?

billsacks:
Sorry for the delay... I've been trying to get some other things done before cheyenne goes down tomorrow morning. I'll plan to look at this tomorrow.

scripts/lib/CIME/build.py
```diff
@@ -258,6 +264,9 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
             full_lib_path = os.path.join(sharedlibroot, sharedpath)
         elif lib == "mpi-serial":
             full_lib_path = os.path.join(sharedlibroot, sharedpath, "mct", lib)
+        elif lib == "cprnc":
+            full_lib_path = os.path.join(sharedlibroot, compiler, "cprnc")
+            case.set_value("CCSM_CPRNC", os.path.join(full_lib_path))
```
billsacks: You need an extra 'cprnc' at the end of this path: right now it points to the directory, not the executable.

tools/cprnc/CMakeLists.txt
billsacks: I don't see this change in the latest commit. Can you double-check that the changes you mention were actually committed?

tools/cprnc/CMakeLists.txt
billsacks: What should we do with this one? I'm okay bringing this PR in as is, but opening an issue that this will need to be generalized. @jedwards4b @jgfouca @WesCoomber do you feel that's the best plan, or should this be fixed for this PR so that it's more likely to work across systems?

scripts/lib/CIME/build.py
jgfouca: I'm surprised wesley's testing didn't fail.

tools/cprnc/CMakeLists.txt
jgfouca: We should make some effort to generalize this.

tools/cprnc/CMakeLists.txt
jgfouca: @WesCoomber

WesCoomber: @billsacks @jgfouca 
The change was on my local branch. I have pushed the commit with the appropriate changes to remote. 
c2f7b31af5418f6fad01a2aeb206b70028f5badb

billsacks: Thanks!

scripts/lib/CIME/build.py
jedwards4b: @billsacks I think you are misinterpreting this  the directory is sharedlibroot/compiler/ and cprnc is the executable.

tools/cprnc/CMakeLists.txt
jedwards4b: I would like to get this into the ufs_release branch before a code freeze this Friday so I vote for merging and opening a new issue.

scripts/lib/CIME/build.py
billsacks: No, I tested it:

```
$ ./xmlquery -p CPRNC
	CCSM_CPRNC: /Users/sacks/projects/scratch/SMS_D_Ld1_P4x1.f10_f10_musgs.I2000Clm50BgcCropQianRsGs.bishorn_gnu.clm-default.20200225_113043_ioba26/bld/gnu/cprnc

$ ls -ld /Users/sacks/projects/scratch/SMS_D_Ld1_P4x1.f10_f10_musgs.I2000Clm50BgcCropQianRsGs.bishorn_gnu.clm-default.20200225_113043_ioba26/bld/gnu/cprnc
drwxr-xr-x  18 sacks  1202617992  576 Feb 25 11:32 /Users/sacks/projects/scratch/SMS_D_Ld1_P4x1.f10_f10_musgs.I2000Clm50BgcCropQianRsGs.bishorn_gnu.clm-default.20200225_113043_ioba26/bld/gnu/cprnc/

$ ls -ld /Users/sacks/projects/scratch/SMS_D_Ld1_P4x1.f10_f10_musgs.I2000Clm50BgcCropQianRsGs.bishorn_gnu.clm-default.20200225_113043_ioba26/bld/gnu/cprnc/cprnc
-rwxr-xr-x  1 sacks  1202617992  114032 Feb 25 11:32 /Users/sacks/projects/scratch/SMS_D_Ld1_P4x1.f10_f10_musgs.I2000Clm50BgcCropQianRsGs.bishorn_gnu.clm-default.20200225_113043_ioba26/bld/gnu/cprnc/cprnc*
```

tools/cprnc/CMakeLists.txt
billsacks: I'm fine with that, though you may want to confirm that this will actually work on the machines you need it on for the ufs_release (it didn't work out of the box on my mac with the way I had things set up).

scripts/lib/CIME/build.py
jedwards4b: I see, sorry.

tools/cprnc/CMakeLists.txt
jedwards4b: So this is a cmake build - what about using this?
https://github.com/ESMCI/cime/blob/master/src/externals/pio2/cmake/FindNetCDF.cmake

jgfouca: It seems like there must be something we can rely on to find netcdf. Any env variable or case setting?

scripts/lib/CIME/build.py
jgfouca: Is this machine specific also? I don't see how our internal testing could have passed if CCSM_CPRNC were not set correctly. @WesCoomber , can you confirm?

jgfouca:
@WesCoomber , we should be a good citizen and add a regression test for this. Please add a test to scripts_regression_tests.

1. Setup a simple ERS case
2. Sabotage CCSM_CPRNC via xmlchange
3. Complete the case (build + submit), it should succeed

scripts/lib/CIME/build.py
WesCoomber: @billsacks You are correct that build.py pointed to the directory instead of the executable. I have implemented the fix and I added a test to scripts_regression_tests.py to catch cases like this in the future. In this commit:
2b882e4d6df38903d835263ab7514f1fd2ed3d75

billsacks: Thanks!

jgfouca:
@billsacks , @jedwards4b , I think we have a nice solution now. buildlib.cprnc passes a NETCDF env hint to the cprnc cmake build and it uses the findnetcdf from pio2.

We also added a regression test for this capability.

This should be all ready to go.

src/build_scripts/buildlib.cprnc
```diff
@@ -0,0 +1,69 @@
...
+def buildlib(bldroot, installpath, case):
+###############################################################################
+    cimeroot = case.get_value("CIMEROOT")
+
+    # Generate macros and environment
+    run_bld_cmd_ensure_logging("{}/tools/configure --mpilib=mpi-serial --macros-format=CMake".format(cimeroot), logger, from_dir=bldroot)
+
+    cmake_args = get_standard_cmake_args(case, shared_lib=True)
+
+    cmake_cmd = "CIMEROOT={cimeroot} . .env_mach_specific.sh && NETCDF=$(dirname $(dirname $(which nf-config))) CIMEROOT={cimeroot} cmake {cmake_args} -DMPILIB=mpi-serial -DDEBUG=FALSE -C Macros.cmake {cimeroot}/tools/cprnc -DCMAKE_PREFIX_PATH={dest_path}".\
```
billsacks: Is nf-config guaranteed to be present? If not, what will happen here?

jedwards4b:
I think all modern netcdf builds install this file and cprnc needs the
fortran build of netcdf, so i don't see a problem.

jedwards4b:
@WesCoomber I'm ready to merge but will defer to you since you are assignee.

jedwards4b:
@WesCoomber I'm ready to merge but will defer to you since you are assignee.

src/build_scripts/buildlib.cprnc
jgfouca: @billsacks , at this point in the code, the user will have their cime-env loaded, so nf-config should be available. If not, NETCDF will be set to an empty string and therefore findNetcdf in cprnc/CMakeLists.txt will probably fail; I think this behavior is fine.

billsacks: Sounds good, thanks.

PR #3414, opened 2020-02-24
UFS release: add generic Linux configuration

climbfuji:
I am targeting the ufs_release_v1.0 for this PR.

This PR adds a generic Linux configuration, identical to the macOS configuration, to CIME. provided that the setup instructions for the machine are followed (NCEPLIBS-external, NCEPLIBS; these are also very similar for macOS and Linux), the end-to-end workflow runs out of the box on a Ubuntu Linux platform and with exactly the same commands as on macOS (including Grib2 output from the post-processor).

This PR requires the issue https://github.com/ufs-community/ufs-mrweather-app/issues/98 to be solved. In order to test the PR, I added an entry for `linux` in `src/model/FV3/cime/cime_config/config_pes.xml` as what is in there for `macos`.

### DIFF
--- a/config/ufs/machines/config_machines.xml
+++ b/config/ufs/machines/config_machines.xml
@@ -345,6 +345,43 @@ This allows using a different mpirun command to launch unit tests
 -->
   </machine>
 
+  <machine MACH="linux">
+    <DESC>
+      Customize these fields as appropriate for your system,
+      particularly changing MAX_TASKS_PER_NODE and MAX_MPITASKS_PER_NODE to the
+      number of cores on your machine.
+    </DESC>
+    <NODENAME_REGEX> something.matching.your.machine.hostname </NODENAME_REGEX>
+    <OS>LINUX</OS>
+    <COMPILERS>gnu</COMPILERS>
+    <MPILIBS>mpich</MPILIBS>
+    <CIME_OUTPUT_ROOT>$ENV{UFS_SCRATCH}</CIME_OUTPUT_ROOT>
+    <DIN_LOC_ROOT>$ENV{UFS_INPUT}/ufs-inputdata</DIN_LOC_ROOT>
+    <DOUT_S_ROOT>$ENV{UFS_SCRATCH}/archive/$CASE</DOUT_S_ROOT>
+    <BASELINE_ROOT>$ENV{UFS_INPUT}/baselines</BASELINE_ROOT>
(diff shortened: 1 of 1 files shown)


config/ufs/machines/config_machines.xml
```diff
@@ -345,6 +345,43 @@ This allows using a different mpirun command to launch unit tests
...
+    <DOUT_S_ROOT>$ENV{UFS_SCRATCH}/archive/$CASE</DOUT_S_ROOT>
+    <BASELINE_ROOT>$ENV{UFS_INPUT}/baselines</BASELINE_ROOT>
+    <CCSM_CPRNC>$CIMEROOT/tools/cprnc/build/cprnc</CCSM_CPRNC>
+    <GMAKE>make</GMAKE>
+    <GMAKE_J>4</GMAKE_J>
+    <BATCH_SYSTEM>none</BATCH_SYSTEM>
+    <SUPPORTED_BY>__YOUR_NAME_HERE__</SUPPORTED_BY>
+    <MAX_TASKS_PER_NODE>8</MAX_TASKS_PER_NODE>
+    <MAX_MPITASKS_PER_NODE>8</MAX_MPITASKS_PER_NODE>
+    <TOTAL_TASKS>8</TOTAL_TASKS>
```
jedwards4b: This variable does not exist in the config_machines.xsd definition as found in
config/xml_schemas/config_machines.xsd 
There is no way to specify a generic task count for FV3 - currently at C96 we support 8 and 108 tasks.   That's it.  Unless and until that situation changes the fix is to set the task count for this system in config_pes.xml

climbfuji: Thanks for the info. Too bad. I'll take the line out and add something to `config_pes.xml`. Do you know if this is documented somewhere in the mr-weather-app users' guide?

jedwards4b: @climbfuji Is there documentation in the weather_model on how to set the layout and io tasks?  If so we can document how use cime with any task count that the model can use.

climbfuji: I don't know, not the documentation guy. Will take a look.

climbfuji: https://ufs-mrapp.readthedocs.io/en/latest/faq.html#how-do-i-change-the-processor-layout

jedwards4b: That's already cime documentation - thanks @uturuncoglu!

PR #3417, opened 2020-02-26
CIME UFS: port to gaea (DRAFT)

climbfuji:
This works apart from the issue noted here: https://github.com/NOAA-EMC/EMC_post/issues/105


config/ufs/machines/config_compilers.xml
```diff
@@ -289,6 +289,74 @@ using a fortran linker.
   <SUPPORTS_CXX>TRUE</SUPPORTS_CXX>
 </compiler>
 
+<compiler COMPILER="cray-intel">
```
jedwards4b: This should not be needed at all

config/ufs/machines/config_machines.xml
```diff
@@ -451,7 +451,77 @@ This allows using a different mpirun command to launch unit tests
...
   </machine>
-
+ <machine MACH="gaea">
+    <DESC>
+      port to gaea
+    </DESC>
+    <NODENAME_REGEX>gaea*</NODENAME_REGEX>
+    <OS>CNL</OS>
+    <COMPILERS>cray-intel</COMPILERS>
+    <MPILIBS>mpich</MPILIBS>
```
jedwards4b: I think that this should be mpt

config/ufs/machines/config_machines.xml
```diff
@@ -451,7 +451,77 @@ This allows using a different mpirun command to launch unit tests
...
+    <PROJECT>esrl</PROJECT>
+    <SAVE_TIMING_DIR> </SAVE_TIMING_DIR>
+    <CIME_OUTPUT_ROOT>$ENV{SCRATCH}/$ENV{USER}/ufs-runs/scratch</CIME_OUTPUT_ROOT>
```
jedwards4b: Set CIME_OUTPUT_ROOT to $ENV{SCRATCH}/$ENV{USER} 
but then set the other fields to $CIME_OUTPUT_ROOT/whatever
so that the user only needs to change one field.

config/ufs/machines/config_machines.xml
```diff
@@ -451,7 +451,77 @@ This allows using a different mpirun command to launch unit tests
...
+    <MAX_MPITASKS_PER_NODE>16</MAX_MPITASKS_PER_NODE>
+    <PROJECT_REQUIRED>FALSE</PROJECT_REQUIRED>
+    <mpirun mpilib="default">
+     <executable>srun</executable>
+      <arguments>
+        <arg name="label"> --label</arg>
+        <arg name="num_tasks"> -n {{ total_tasks }}</arg>
+    </arguments>
+    </mpirun>
+    <module_system type="module" allow_error="true">
```
jedwards4b: Do you know that we need the allow_error="true" option - lets not use it if we don't need to.  (Some module setups write info messages to stderr - that's when we need this flag)

config/ufs/machines/config_machines.xml
```diff
@@ -451,7 +451,77 @@ This allows using a different mpirun command to launch unit tests
...
+      <init_path lang="perl">/opt/cray/pe/modules/3.2.10.5/init/perl.pm</init_path>
+      <init_path lang="python">/opt/cray/pe/modules/3.2.10.5/init/python.py</init_path>
+      <init_path lang="csh">/opt/cray/pe/modules/3.2.10.5/init/csh</init_path>
+      <init_path lang="sh">/opt/cray/pe/modules/3.2.10.5/init/sh</init_path>
+      <cmd_path lang="perl">/opt/cray/pe/modules/4.1.3.1/bin/modulecmd perl</cmd_path>
+      <cmd_path lang="python">/opt/cray/pe/modules/4.1.3.1/bin/modulecmd python</cmd_path>
+      <cmd_path lang="sh">module</cmd_path>
+      <cmd_path lang="csh">module</cmd_path>
+      <modules>
+        <command name="purge"/>
```
jedwards4b: purge does not work on cray.  Instead I usually have a list of specific modules to remove such as this https://github.com/ESMCI/cime/blob/master/config/cesm/machines/config_machines.xml#L712

config/ufs/machines/config_machines.xml
```diff
@@ -451,7 +451,77 @@ This allows using a different mpirun command to launch unit tests
...
     </environment_variables>
   </machine>
-
+ <machine MACH="gaea">
+    <DESC>
+      port to gaea
+    </DESC>
+    <NODENAME_REGEX>gaea*</NODENAME_REGEX>
+    <OS>CNL</OS>
+    <COMPILERS>cray-intel</COMPILERS>
```
jedwards4b: should just be intel

config/ufs/machines/config_machines.xml
```diff
@@ -451,7 +451,77 @@ This allows using a different mpirun command to launch unit tests
...
+      <init_path lang="csh">/opt/cray/pe/modules/3.2.10.5/init/csh</init_path>
+      <init_path lang="sh">/opt/cray/pe/modules/3.2.10.5/init/sh</init_path>
+      <cmd_path lang="perl">/opt/cray/pe/modules/4.1.3.1/bin/modulecmd perl</cmd_path>
+      <cmd_path lang="python">/opt/cray/pe/modules/4.1.3.1/bin/modulecmd python</cmd_path>
+      <cmd_path lang="sh">module</cmd_path>
+      <cmd_path lang="csh">module</cmd_path>
+      <modules>
+        <command name="purge"/>
+      </modules>
+      <modules compiler="cray-intel">
```
jedwards4b: should just be intel

config/ufs/machines/config_machines.xml
```diff
@@ -451,7 +451,77 @@ This allows using a different mpirun command to launch unit tests
...
+        <command name="use">/lustre/f2/pdata/esrl/gsd/ufs/modules/modulefiles/intel-18.0.3.222</command>
+        <command name="load">NCEPlibs/1.0.0.beta03</command>
+      </modules>
+    </module_system>
+    <environment_variables>
+      <env name="OMP_STACKSIZE">256M</env>
+      <env name="NCEP_LIBS">$ENV{NCEPLIBS_DIR}</env>
+      <env name="SCRATCH">/lustre/f2/scratch</env>
+      <env name="NETCDF">$ENV{NETCDF}</env>
+      <env name="CMAKE_Platform">gaea.intel</env>
```
jedwards4b: don't need to define CMAKE_Platform

config/ufs/machines/config_machines.xml
climbfuji: the module is called cray-mpich on gaea

jedwards4b: I'm not entirely sure on this one, we can try it.

climbfuji:
Iteration 1 fails with same error:

CMake Error at CMakeLists.txt:16 (project):
  The CMAKE_Fortran_COMPILER:

    mpif90

  is not a full path and was not found in the PATH.

  Tell CMake where to find the compiler by setting either the environment
  variable "FC" or the CMake cache entry CMAKE_Fortran_COMPILER to the full
  path to the compiler, or to the compiler name if it is in the PATH.

need to go out for an hour now ...

jedwards4b:
Where is that message coming from? The atm.bldlog?

jedwards4b:
Can you please send the Macros.cmake file from the case directory and the atm.bldlog to me?

Thanks

climbfuji:
[logs.tar.gz](https://github.com/ESMCI/cime/files/4257880/logs.tar.gz)

Here they are. Will also take a look inside.

jedwards4b:
You still have compiler set to cray-intel - was this case created before your latest changes?

climbfuji:
I believe I sent you an old build log. Anyway, tried it again with everything cleaned out. Same results.

[logs.tar.gz](https://github.com/ESMCI/cime/files/4258006/logs.tar.gz)

jedwards4b:
I'm still seeing Calling cmake  -DCOMPILER=cray-intel  
Did you start from a fresh case or are you trying to build the same one again?
./xmlquery COMPILER
should result in intel

climbfuji:
```
Dom.Heinzeller@gaea15:/lustre/f2/dev/esrl/Dom.Heinzeller/ufs-mrweather-app/ufs-mrweather-app-20200225/c96-gfsv15p2> ./xmlquery COMPILER
	COMPILER: intel
```

climbfuji:
I'll manually remove the bld and run directories under the scratch space for the next try.

jedwards4b:
Please remove the case directory as well.

climbfuji:
That one was gone, I use rsync --delete to copy the data across.

climbfuji:
[logs_again.tar.gz](https://github.com/ESMCI/cime/files/4258037/logs_again.tar.gz)

climbfuji:
This is the culprit! `src/model/FV3/cime/cime_config/buildlib`
```
            os.environ["CMAKE_C_COMPILER"]="mpicc"
            os.environ["CMAKE_CXX_COMPILER"]="mpicxx"
            os.environ["CMAKE_Fortran_COMPILER"]="mpif90"
            os.environ["CMAKE_Platform"]="cime"
```
Hardcoded values!!!

climbfuji:
Note that even on hera we prefer mpiicc, mpiifort, mpiicpc.

config/ufs/machines/config_machines.xml
```diff
@@ -451,7 +451,66 @@ This allows using a different mpirun command to launch unit tests
...
+        <command name="load">intel/18.0.3.222</command>
+        <command name="use">/lustre/f2/pdata/esrl/gsd/ufs/modules/modulefiles/generic</command>
+        <command name="load">cmake/3.16.4</command>
+        <command name="use">/lustre/f2/pdata/esrl/gsd/ufs/modules/modulefiles/intel-18.0.3.222</command>
+        <command name="load">NCEPlibs/1.0.0.beta03</command>
+      </modules>
+    </module_system>
+    <environment_variables>
+      <env name="OMP_STACKSIZE">256M</env>
+      <env name="NCEP_LIBS">$ENV{NCEPLIBS_DIR}</env>
```
jedwards4b: Do we need this variable? I don't think cime uses it.

config/ufs/machines/config_machines.xml
```diff
@@ -451,7 +451,66 @@ This allows using a different mpirun command to launch unit tests
...
+      <env name="SCRATCH">/lustre/f2/scratch</env>
```
jedwards4b: Since you are setting SCRATCH here I suggest that we change it here and above to UFS_SCRATCH to be consistant with other systems.  Also for consistancy perhaps we should require users to define UFS_SCRATCH before using cime?

config/ufs/machines/config_machines.xml
```diff
@@ -451,7 +451,66 @@ This allows using a different mpirun command to launch unit tests
...
+      <env name="NETCDF">$ENV{NETCDF}</env>
```
jedwards4b: All this does is throw an error if $NETCDF isn't defined in the environment - is that your intent?

climbfuji:
I'll try to make the changes you suggested. 

Note: the grib table is still wrong. If I do a diff between params_grib2_tbl_new in the run directory and
```
 /lustre/f2/pdata/esrl/gsd/ufs/modules/NCEPlibs-ufs-v1.0.0.beta03/intel-18.0.3.222/cray-mpich-7.4.0/share/params_grib2_tbl_new params_grib2_tbl_new
```
or
```
 /lustre/f2/pdata/esrl/gsd/ufs/modules/NCEPlibs-ufs-v1.0.0.beta04/intel-18.0.3.222/cray-mpich-7.4.0/share/params_grib2_tbl_new params_grib2_tbl_new
```
I see a lot of differences.

jedwards4b:
You should be using hash 28a5ca9 of FV3/cime - can you make sure that is correct.

climbfuji:
I just checked out the code again. It's a little obscure to me with manage externals, with git submodules I have more experience. But I see the latest hash in FV3/cime to be 28a5ca98ea56355fc7f3f8f4eb444f6be6ac3946 ... hopefully this will fix it.

Dom

climbfuji:
Ok, now the grib param files are identical. Let's hope it runs through!

config/ufs/machines/config_batch.xml
```diff
@@ -262,4 +245,19 @@
...
+    <submit_args>
+      <arg flag="--time" name="$JOB_WALLCLOCK_TIME"/>
+      <arg flag="--qos" name="$JOB_QUEUE"/>
+      <arg flag="--account" name="$CHARGE_ACCOUNT"/>
+    </submit_args>
+    <directives>
+      <directive>--clusters=c4</directive>
+    </directives>
+    <queues>
+      <queue walltimemax="00:30:00" nodemin="1" nodemax="171">debug</queue>
```
jedwards4b: This is good for testing but is there another queue you should add for production use?  there is a default attribute that you can use to set the preferred queue if they overlap

climbfuji: Yes, I think when our testers work with this the next days on the larger test cases, they will need to add other queues (maybe - debug is fairly generous on gaea). I wanted to get the initial config going with the C96 case.

climbfuji:
Ready to merge.

PR #3430, opened 2020-03-02
only e3sm has the unique sharedlib setup

### LINKED ISSUE #3427: cprnc build fails when running suite of tests
I keep getting failures because more than one test is trying to create the directory at the same time - around line 278 of build.py

jgfouca:
This should fix #3427 

Test suite: None
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/test_scheduler.py
+++ b/scripts/lib/CIME/test_scheduler.py
@@ -803,7 +803,7 @@ class TestScheduler(object):
             return total_pes
 
         elif (phase == SHAREDLIB_BUILD_PHASE):
-            if self._cime_model == "cesm":
+            if self._cime_model != "e3sm":
                 # Will force serialization of sharedlib builds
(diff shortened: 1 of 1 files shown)


jgfouca:
@jedwards4b , this raises a broader issue. Now that we have more models, the following:

```
if get_model() == "cesm"
```

and

```
if get_model() != "e3sm"
```

Are no longer semantically equivalent. We probably need to do a quick review of the repo to make sure we are being consistent.

rljacob:
Why does e3sm do shared libs that way?  Can we switch back?

We should have a separate issue for use of` get_model =` throughout the code.

jgfouca:
@rljacob , I can't even remember; I made the change years because something was broken. It's on my todo-list to re-evaluate.

PR #3434, opened 2020-03-04
Wpcoomb/improve PETtest baseline generation and comparison

### LINKED ISSUE #2916: PET test does baseline generation and comparison twice
I noticed that the PET test is doing baseline generation and comparison twice: once before the base vs. single-therad comparison and once after it.

While this doesn't cause any major problems that I can see, it is a bit confusing, especially when looking at the copied cprnc.out files in the test directory from a failed baseline comparison (which contain two copies of the cprnc output concatenated together).

From a quick glance, my best guess is that this is due to the PET test being a multi-submit test, but I'm not sure about that.

WesCoomber:
Improvement to baseline generation and comparison for the PET multi-submit test.

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes #2916 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


scripts/lib/CIME/SystemTests/system_tests_common.py
```diff
@@ -141,12 +141,12 @@ def run(self, skip_pnl=False):
             with self._test_status:
                 self._test_status.set_status(RUN_PHASE, TEST_PEND_STATUS)
 
+            resub_val = self._case1.get_value("RESUBMIT")
```
jgfouca: I think this is close, but you cannot rely on case1/case2 existing for all test types.

jedwards4b:
A minor nit - can you avoid adding new whitespace to files please.

scripts/lib/CIME/SystemTests/system_tests_common.py
jgfouca: I think this is what you want

```
if (isinstance(self, SystemTestsCompareTwo)):
    resub_val = self._case1.get_value("RESUBMIT")
else:
    resub_val = 0
```

jgfouca:
As part of this PR, please add some Fake tests that are resubmit-enabled.

WesCoomber:
_This is still WiP, there are some deeper issues with the scripts_regression_tests.py script on the Chama Machine regarding file permissions and test_bless_test_results timing on the run phase for batch systems._ I am investigating to figure out the proper solution and will either add it to this PR or open up a new one depending on the scope of the implemented fix. So far it has been a good learning experience to make sure I understand how the 'Fake Tests' and regression tests work on batch systems and how to properly diagnose them.

jgfouca:
Asking for review from @billsacks

PR #3435, opened 2020-03-05
Bugfix for Gaea

climbfuji:
When the NCEPLIBS module was updated to beta04, the required changes to the modules environment were not made in the CIME config. This PR fixes the problem.


climbfuji:
@jedwards4b @uturuncoglu this must be merged in order to run CIME on gaea (when the user does not load the required modules manually beforehand).

PR #3436, opened 2020-03-06
CIME config for MSU Orion

climbfuji:
This is the CIME config for MSU Orion. It was tested for the c96-gfsv15p2 test case and produced wgrib2 output from the model run. Most of this work was done by @GeorgeVandenberghe-NOAA, I only moved things around into the right place and adjusted a few knobs.

Also contains a tiny bugfix for jet, which has been tested successfully.

@GeorgeVandenberghe-NOAA @arunchawla-NOAA @jedwards4b

### DIFF
--- a/config/ufs/machines/config_batch.xml
+++ b/config/ufs/machines/config_batch.xml
@@ -217,6 +217,22 @@
     </queues>
   </batch_system>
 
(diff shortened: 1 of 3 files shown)


climbfuji:
Thanks, @jedwards4b!

PR #3439, opened 2020-03-08
Documentation: provide a version dropdown menu and change theme

billsacks:
The documentation of cime is starting to diverge on the various
important branches (master, maint-5.6, and ufs_release_v1.0). This PR,
along with similar PRs for maint-5.6 and ufs_release_v1.0, and some
changes to the gh-pages branch, support having versioned documentation,
with a dropdown menu. This also changes the documentation theme to a
theme that supports this version dropdown. Note that this requires a
branch of the sphinx_rtd_theme; I will update the cime wiki on building
the documentation to note the new pip install step needed for that. (See
below for more details.)

At a high level, the way the versioned documentation works is to have
separate subdirectories in the gh-pages branch for each version of the
documentation we want to support. There is then a bit of JavaScript code
which uses a json file in the gh-pages branch to determine which
versions exist and how these should be named in the dropdown menu. This
changes the documentation build process slightly; I will update the wiki
to reflect these changes.

@mvertens and I made the various changes that you can see in
`doc/source` in this PR, which mainly borrowed from
https://github.com/ESCOMP/CISM-wrapper/pull/23. That, in turn, is a
slight modification of an implementation provided by @mnlevy for the
MARBL documentation, which in turn borrowed from an implementation put
together by Unidata (credit where credit is due).

As mentioned above, I am not aware of out-of-the-box support for a
version pull-down in out-of-the-box sphinx themes (though the last time
I looked was in Fall, 2018, so there may be something available
now). However, support for a version dropdown exists in an open PR in
the sphinx readthedocs theme repository:
https://github.com/readthedocs/sphinx_rtd_theme/pull/438. I have pushed
this branch to a new repository in ESMCI
(https://github.com/ESMCI/sphinx_rtd_theme) to facilitate long-term
maintenance of this branch in case it disappears from the official
sphinx_rtd_theme repository. I have also cherry-picked a commit onto
that branch, which is needed to fix search functionality in sphinx1.8
(from https://github.com/readthedocs/sphinx_rtd_theme/pull/672) (which
is another reason for maintaining our own copy of this branch). The
branch in this repository is now named version-dropdown-with-fixes
(branching off of the version-dropdown branch in the sphinx_rtd_theme
repository). I will update the wiki to give the additional pip install
command needed to install this repository before doing the build. In the
long-term, I am a little concerned about using this theme that isn't
showing any signs of being merged to the main branch of the readthedocs
theme, but this has been working for us in other projects for the last 2
years, so I feel this is a reasonable approach in the short-medium term.

A few changes were also needed in the gh-pages branch:

- A `versions` directory was introduced at the top level, with
  subdirectories for each documentation version we want to support. We
  are using the convention that these subdirectories should have names
  that match the corresponding cime branches, to avoid uncertainty.

- A top-level index.html page was introduced that redirects to the
  master version of the documentation:

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta http-equiv="refresh" content="0; url=versions/master/html/index.html" />
  </head>
</html>
```

- Other than those files and an empty `.nojekyll` file, everything else
  was removed from the top level directory of the documentation.

- In the versions directory, we also introduced a `versions.json`
  file. This lists the versions and provides a mapping between directory
  names and the names used in the pull-down menu:

```json
{
    "master": "master",
    "maint-5.6": "cime5.6",
    "ufs_release_v1.0": "ufs_release_v1.0"
}
```

- The complete contents of each version subdirectory are generated by
  the documentation build script. Thus, it is safe to remove the
  contents of (e.g.) `versions/master/` before rebuilding the
  documentation for that version. The documentation build is done
  similarly to before, but now providing a BUILDDIR argument to the make
  command in order to build in the correct subdirectory. For example,
  when building the documentation for master, you would provide
  `BUILDDIR=/path/to/cime_gh_pages_repo/versions/master/`.

Test suite: none
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes none

User interface changes?: none

Code review:


billsacks:
@rljacob - as I mentioned in the earlier PR, note that we have already pushed the resulting documentation build to http://esmci.github.io/cime - so you can take a look at that to see the end result.

Once this is approved, I will open similar PRs to the maint-5.6 and ufs_release_v1.0 branches.

rljacob:
I suppose this was inevitable but was hoping we could put it off a bit longer.  There are still some holes in the documentation which now would need to be replicated in 3 places when filled?

billsacks:
> I suppose this was inevitable but was hoping we could put it off a bit longer. There are still some holes in the documentation which now would need to be replicated in 3 places when filled?

Yes, that's certainly a downside. My understanding from @mvertens, though, is that this is truly needed at this point because documentation is starting to diverge between the different branches.

@mvertens I'll merge once you give your formal approval.

rsdunlapiv:
I did not think that this was about diverging documentation, but instead the need for a tag to lock down this point in time.  In my view the CIME docs should not diverge, but contain information about all common parts of CIME, regardless of the modeling system.  If there are diverging aspects, then in draws into question whether a separate modeling-system specific documentation layer is needed.  However, in this specific case, I don't think it is about diverging docs.  In fact, didn't we tag the docs "as is" ("as are?") without adding anything specific about UFS?

mvertens:
@rsdunlapiv - you really do need to have 3 branches for the cime documentation right now. There are changes in the data models that only relate to the latest developments and are not in the cesm release. The same is true for other updates to CIME. As for the UFS release, there could be other additions that do not relate to cesm or e3sm. I'm not sure that the documentation is divergent just that there is a separation of concerns that having the 3 branches permits us to address.

rsdunlapiv:
Will all the branches be merged back together, e.g., when the updated data models are available in CESM?  Just trying to understand if this branching is needed due to fundamental differences between the modeling systems, or if it used to manage the fact that different modeling systems are using different versions of CIME.

rljacob:
I'm now wondering this too.   If the UFS data models are going to live on master, they should be documented on master.   Is the UFS branch going to live forever and diverge?  Why would UFS not want to track master?

mvertens:
The new ESMF-based data models will need to have their own separate
documentation - but it could be on master as well. Its not clear that UFS
wants to track master - and they need their own frozen release
documentation. We should discuss this some more and come up with a longer
term plan - but the hooks are now there to version control the different
branches that are required for the releases.

jedwards4b:
I am working on a PR now to merge the ufs branch back to master and the next ufs release should be a new branchpoint from master, but the ufs developers wanted a documentation branch that was fixed with respect to the release version.

rljacob:
Any specific version UFS wants to use should be tagged (on master) and we should find a way to build docs for specific tags.   If the docs or other things need to be fixed, a maintenance branch would be made off that tag and the docs for that version built from the head of that branch.

Everything in master should be documented on master.

jedwards4b:
@rljacob I think that's what this does, the ufs_v1.0.0 documentation is on a branch and future ufs development will occur on master, with only bug fixes going on the ufs_release_v1.0 branch.

jedwards4b:
@rljacob I should say the documentation is tied to the tag ufs_v1.0.0 which is on a branch.

rljacob:
Ok.  That branch point should have a cimeX.Y.X tag as well.  Anytime UFS wants a release, there should be a corresponding cime tag.

PR #3441, opened 2020-03-10
Thunder port

jedwards4b:
Port to system Thunder - 4 node ARM system at NCAR. 

Test suite: scripts_regression_tests.py on thunder with armgcc and gnu compilers. 
Test baseline: 
Test namelist changes: 
Test status:
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -260,7 +260,7 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
             case.set_value("CCSM_CPRNC", os.path.join(full_lib_path, "cprnc"))
             if not os.path.isdir(full_lib_path):
                 os.makedirs(full_lib_path)
-                libs.append("cprnc")
+                libs.insert(0,"cprnc")
 
 
     logs = []
--- a/src/build_scripts/buildlib.cprnc
+++ b/src/build_scripts/buildlib.cprnc
@@ -51,15 +51,15 @@ def buildlib(bldroot, installpath, case):
     run_bld_cmd_ensure_logging("{}/tools/configure --mpilib=mpi-serial --macros-format=CMake --machine={} --compiler={}".format(cimeroot, case.get_value("MACH"), compiler), logger, from_dir=bldroot)
 
     cmake_args = get_standard_cmake_args(case, shared_lib=True)
-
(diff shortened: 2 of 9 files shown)


scripts/lib/CIME/build.py
```diff
@@ -260,7 +260,7 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
             case.set_value("CCSM_CPRNC", os.path.join(full_lib_path, "cprnc"))
             if not os.path.isdir(full_lib_path):
                 os.makedirs(full_lib_path)
-                libs.append("cprnc")
+                libs.insert(0,"cprnc")
```
jedwards4b: @WesCoomber these are changes to get the cprnc build and test working so I thought you should review.

jedwards4b:
@jtruesdal Can you make sure this change in the Makefile is okay - it just prevents a warning about the CAM_DYCORE variable when cam is not involved in the build.

src/build_scripts/buildlib.cprnc
```diff
@@ -51,15 +51,15 @@ def buildlib(bldroot, installpath, case):
     run_bld_cmd_ensure_logging("{}/tools/configure --mpilib=mpi-serial --macros-format=CMake --machine={} --compiler={}".format(cimeroot, case.get_value("MACH"), compiler), logger, from_dir=bldroot)
 
     cmake_args = get_standard_cmake_args(case, shared_lib=True)
-
-    cmake_cmd = "CIMEROOT={cimeroot} . ./.env_mach_specific.sh && NETCDF=$(dirname $(dirname $(which nf-config))) CIMEROOT={cimeroot} cmake {cmake_args} -DMPILIB=mpi-serial -DDEBUG=FALSE -C Macros.cmake {cimeroot}/tools/cprnc -DCMAKE_PREFIX_PATH={dest_path} -DBLDROOT={bldroot}".\
+    os.environ["CIMEROOOT"] = cimeroot
```
WesCoomber:
```suggestion
    os.environ["CIMEROOT"] = cimeroot
```
I think this is a typo and should be CIMEROOT with two O's.

jedwards4b: Good catch

scripts/lib/CIME/build.py
WesCoomber: I agree inserting 'cprnc' to the front of the libs list, similar to how we already insert 'mpilib' into libs, makes sense to me.

PR #3449, opened 2020-03-12
Improve case.setup clean (happens for --clean or --reset)

jgfouca:
Need to remove Macro files, otherwise the setup does not re-run
the BuildTools.configure.

Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


jedwards4b:
I actually prefer this feature as is and use it to test new compiler options by editing the Macros file directly.   Can we just document the need to remove this file if you want to regenerate it?  Or maybe have a ./case.setup --reset-full option?

jgfouca:
You could still edit Macros, right? You'd just have to not do a `case.setup --clean` in order to not lose them.

PR #3470, opened 2020-03-31
use shrmap by reference instead of copy

jedwards4b:
This PR replaces a copy of three arrays in shr_dmodel_mod with a reference, thus reducing the memory footprint considerably and improving performance too. 

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/share/streams/shr_dmodel_mod.F90
+++ b/src/share/streams/shr_dmodel_mod.F90
@@ -38,7 +38,7 @@ module shr_dmodel_mod
   end interface shr_dmodel_gsmapCreate
 
   interface shr_dmodel_mapSet; module procedure &
-       !shr_dmodel_mapSet_dest!, &
+       !shr_dmodel_mapSet_dest, &
        shr_dmodel_mapSet_global
   end interface shr_dmodel_mapSet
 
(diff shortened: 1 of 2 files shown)


jedwards4b:
@rljacob @mvertens waiting for reviews

jedwards4b:
@jgfouca can you review?

src/share/streams/shr_dmodel_mod.F90
```diff
@@ -38,7 +38,7 @@ module shr_dmodel_mod
   end interface shr_dmodel_gsmapCreate
 
   interface shr_dmodel_mapSet; module procedure &
-      !shr_dmodel_mapSet_dest, &
+       !shr_dmodel_mapSet_dest!, &
```
rljacob, line 41: Does this comment out the line?  Why not just remove it?

jedwards4b: The original already had this line commented out  - not sure why the change was made, I can remove it or restore to the original.

PR #3477, opened 2020-04-05
Add standard UCAR Code of Conduct for the maint-5.6 branch since that…

### LINKED ISSUE #3476: Add a Code of Conduct to cime
It would be good for CIME to have a Code of Conduct (COC) for expected behaviors

Because the maint5.6 branch is solely CESM, we should probably adopt the standard UCAR COC 
which has a DOI here: https://doi.org/10.5065/6w2c-a132. I'll probably give a PR for that directly.

The question is what should a COC looks like for CIME as a whole on master? It possibly should be more expansive than the UCAR one. The UCAR talks about how to report on issues to UCAR resources (such as the Office of Diversity and Inclusion, Human Resources, and Ethics and Legal). Those resources can be involved even for issues involving people outside of UCAR. But, possibly reporting should be more expansive for CIME with a significant set of people outside UCAR.

Having a  COC in each repository is a github best practice. It's something that NSF wants all of its projects to have. And it's a good idea in order to have  and maintain diversity and inclusion in a project. It helps everyone on a project to be more productive, by having expectations on behavior that encourage everyone to be respectful and understanding with each other.

ekluzek:
Add Code of Conduct (COC) to maint-5.6

This adds the standard UCAR COC to the maint-5.6 branch. Since this branch is maintained just by CESM, this seems the right thing to do. A COC on master for CIME might need to be broader.

Test suite: None (Documentation only)

Fixes #3476 

User interface changes?: N

Code review:

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,84 @@
+# Contributor Code of Conduct
+_The Contributor Code of Conduct is for participants in our software projects and community._
+
+## Our Pledge
+We, as contributors, creators, stewards, and maintainers (participants), of the Common Infrastructure for Modeling the Earth (CIME) pledge to make participation in our software, system or hardware project and community a safe, productive, welcoming and inclusive experience for everyone.
+All participants are required to abide by this Code of Conduct.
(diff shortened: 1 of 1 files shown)


ekluzek:
There was some discussion about this, and the thinking was that the code of conduct should be the same everywhere. So This should just be closed.

PR #3486, opened 2020-04-08
Bring back the setting of dlnd strm_domdir and strm_domfil to null

### LINKED ISSUE #3485: dlnd strm_domdir and strm_domfil left unset when they should be null
@fischer-ncar is seeing failures in T compset tests (CISM with dlnd) after the merge of #3457 . It looks like the problem is: When dlnd was changed to support a new mode for MOSART (I believe by @jonbob ), the default settings of `null` for `strm_domdir` and `strm_domfil` were accidentally removed. It turns out that these explicit settings to `null` are important for the logic in nmlgen.py: `create_stream_file_and_update_shr_strdata_nml`.

I will open a PR shortly that brings back these default settings.

billsacks:
When dlnd was changed to support a new mode for MOSART in
ESMCI/cime#3457, the default settings of null for strm_domdir and
strm_domfil were accidentally removed. It turns out that these explicit
settings to null are important for the logic in nmlgen.py:
create_stream_file_and_update_shr_strdata_nml.

This commit brings back these settings to null

Test suite: NONE!
Test baseline: n/a
Test namelist changes: Will change namelists for T compsets
(reverting to behavior prior to #3457 for these compsets)
Test status: bit for bit

Fixes #3485 

User interface changes?: N

Code review: @fischer-ncar

### DIFF
--- a/src/components/data_comps_mct/dlnd/cime_config/namelist_definition_dlnd.xml
+++ b/src/components/data_comps_mct/dlnd/cime_config/namelist_definition_dlnd.xml
@@ -80,6 +80,7 @@
     <group>streams_file</group>
     <desc>Stream domain file directory.</desc>
     <values>
+      <value>null</value>
       <value stream="clm.nldas">$DLND_CPLHIST_DIR</value>
     </values>
   </entry>
@@ -90,6 +91,7 @@
     <group>streams_file</group>
     <desc>Stream domain file path(s).</desc>
     <values>
+      <value>null</value>
       <value stream="clm.nldas">mosart_nldas_hist_1975.clm2.h1.1975-01-01-00000.nc</value>
     </values>
   </entry>


billsacks:
@jonbob FYI. Also, I wonder if you want to keep these two variables set to `null` for the new dlnd mode you added for mosart. My gut thinking is that yes, you should use `null` for that, too. You may want to test that and see if that's true. Note this logic in `nmlgen.py` that is designed to support leaving the domain settings as `null` so that the domain information will be read from the data files, which I think is what you want in your case, too:

```python
            # Overwrite domain_file if should be set from stream data
            if domain_filenames == 'null':
                domain_filepath = data_filepath
                domain_filenames = data_filenames.splitlines()[0]
```

jonbob:
Thanks @billsacks - I need to look at that. I couldn't get dlnd to work and will try with your fix.

Issue #3487, opened 2020-04-09
ERIO_Ln11.f09_g16.X and ERI_Ln9.f09_g16.X are broken on mappy


jgfouca:
These CIME developer tests appear to have been broken on melvin for some time. @jedwards4b have these tests been working for you guys?

jedwards4b:
Yes they are.

jgfouca:
@WesCoomber , I tried to triage this problem and was unsuccessful. I went back 3 years in the git history and was unable to find a commit where these tests worked. This is strange because I don't remember seeing these failures in scripts_regression_tests until recently. Anyway, this will require a deep dive and will be part of your effort to get us back to 100% pass on melvin and chama.

rljacob:
The ERIO test last passed on March 27 using d308e3ea80bdd1
https://my.cdash.org/index.php?project=E3SM&date=2020-03-27
You can't reproduce that?

jgfouca:
@rljacob , that's a different test than what's in the cime_developer suite, but I can try it.

rljacob:
Oh ok.

jgfouca:
ERIO.ne30_g16_rx1.A.melvin_gnu seems to work just fine.

rljacob:
How does the pelayout between the 2 differ?

rljacob:
Still broken.

jgfouca:
Old issue. We are going to address this with a new one.

PR #3490, opened 2020-04-13
 nuopc_updates for scripts_regression_tests, esmf on cheyenne

jedwards4b:
Updates the scripts_regression_tests.py to work with nuopc interface, updates the esmf build on cheyenne.

Test suite: CIME_MODEL=cesm; CIME_DRIVER=nuopc; ./scripts_regression_tests.py 
Test baseline: 
Test namelist changes: 
Test status: bit for bit,
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/tests/scripts_regression_tests.py
+++ b/scripts/tests/scripts_regression_tests.py
@@ -2456,7 +2456,7 @@ class K_TestCimeCase(TestCreateTestCommon):
     ###########################################################################
     def test_self_build_cprnc(self):
     ###########################################################################
-        testname = "ERS_Ln7.f19_g16.X"
+        testname = "ERS_Ln7.f19_g16_rx1.A"
         self._create_test([testname, "--no-build"], test_id=self._baseline_name)
 
         casedir = os.path.join(self._testroot,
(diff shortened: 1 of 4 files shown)


jedwards4b:
I implemented the increase walltime change incorrectly - undoing and pushing the correct change.

jedwards4b:
@mvertens @jgfouca this is ready for review.

scripts/tests/scripts_regression_tests.py
```diff
@@ -2453,7 +2456,7 @@ def test_configure(self):
     ###########################################################################
     def test_self_build_cprnc(self):
     ###########################################################################
-        testname = "ERS.f19_g16_rx1.A"
+        testname = "ERS_Ln7.f19_g16.X"
```
jgfouca, line 2459: Why this change?

jedwards4b: Because the test is built and run and it's only intended to test the cprnc build so building and running the cheapest possible test makes sense.

jgfouca: This test doesn't pass on melvin unfortunately.

jedwards4b: Shouldn't it?  I can change this back to ERS_Ln7.f19_g16_rx1.A - will that work?

jedwards4b: @jgfouca please try again now.

jgfouca: @jedwards4b , yes it shouldn't, but currently does not. See https://github.com/ESMCI/cime/issues/3487 . I think any "A" case will work; it's the X cases that stopped working.

PR #3508, opened 2020-04-28
Allow user to specify an extra machines directory in create_newcase

### LINKED ISSUE #3493: Add command-line flag to specify alternate location custom machine files, rather than ~/.cime
Connected with #3489 , I'd like to add a command-line flag to create_newcase and create_test that lets you specify the location of the .cime directory, rather than assuming it's in your home directory.

### LINKED ISSUE #3506: scripts_regression_tests: J_CreateNewCase doesn't respect --no-teardown
I noticed that, in scripts_regression_tests, class J_CreateNewCase doesn't respect the `--no-teardown` argument. The fix is simple. I'll add it to my branch where I'm addressing #3493 .

billsacks:
Allow user to specify an extra machines directory in create_newcase

This is to support making an on-the-fly cime port (see #3493 for more
rationale).

Note that this extra machines directory is used in addition to the
~/.cime directory.

I have also made a few changes in `config_machines_template.xml` to make
it closer to working out of the box, so that I could more easily use it
in a test. (And it seems to be better for usability if the default
values in there are ones that will work.)  With the changes here, I am
able to get through the case.setup step (which is how far I want to
get), using a command like:

`./create_newcase --case test_mymachine_0427b --compset X --res f19_g16 --mach mymachine -n --output-root .`

DEPENDENCIES: For NUOPC cases, depends on https://github.com/ESCOMP/CMEPS/pull/62

Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Additional testing: I tested this on my mac (which uses porting
information in ~/.cime), by first creating a case and running case.setup
as normal from cime master; then moving ~/.cime to a temporary location,
doing the same create_newcase from this branch but with
--extra-machines-dir pointing to the temporary location of my .cime
directory, running .case.setup in that new case, and comparing the
resulting Macros.make and Macros.cmake with the baseline. They were
identical, as hoped for.

Fixes #3493 
Fixes #3506 

User interface changes?: Adds `--extra-machines-dir` argument to `create_newcase`

Code review:


utils/perl5lib/Config/SetupTools.pm
```diff
@@ -218,180 +218,6 @@ sub getxmlvars
     }
 }
 
-#-------------------------------------------------------------------------------
```
billsacks, line 221: This set_compiler perl function doesn't appear to be used anymore. I didn't want to try to rewrite it to accept a different location for the .cime directory, so I'm just removing this unused code.

config/xml_schemas/config_machines_template.xml
```diff
@@ -37,7 +37,7 @@
          the batch script (ex. #PBS -A charge_account). Will default
          to PROJECT if not set.
 	 can be overridden in environment or $HOME/.cime/config -->
-    <CHARGE_ACCOUNT></CHARGE_ACCOUNT>
+    <CHARGE_ACCOUNT>couldbethis</CHARGE_ACCOUNT>
```
billsacks, line 40: An empty string didn't work with the schema

config/xml_schemas/config_machines_template.xml
```diff
@@ -72,14 +72,14 @@
     <CCSM_CPRNC>$ENV{CESMDATAROOT}/tools/cime/tools/cprnc/cprnc.cheyenne</CCSM_CPRNC>
 
     <!-- GMAKE: gnu compatible make tool, default is 'gmake' -->
-    <GMAKE></GMAKE>
+    <GMAKE>gmake</GMAKE>
```
billsacks, line 75: An empty string didn't work with the schema

config/xml_schemas/config_machines_template.xml
```diff
@@ -72,14 +72,14 @@
...
 
     <!-- GMAKE_J: optional number of threads to pass to the gmake flag -->
     <GMAKE_J>8</GMAKE_J>
 
     <!-- BATCH_SYSTEM: batch system used on this machine,
       supported values are: none, cobalt, lsf, pbs, slurm -->
-    <BATCH_SYSTEM>pbs</BATCH_SYSTEM>
+    <BATCH_SYSTEM>none</BATCH_SYSTEM>
```
billsacks, line 82: With pbs, it seems like I would have needed a `config_batch.xml` entry for this dummy machine in order to use it in tests. To keep things simple, I changed the default to `none`.

scripts/tests/scripts_regression_tests.py
```diff
@@ -842,7 +898,7 @@ def tearDownClass(cls):
                     shutil.rmtree(tfile)
                 except BaseException:
                     print("Could not remove directory {}".format(tfile))
-        if rmtestroot:
+        if rmtestroot and do_teardown:
```
billsacks, line 901: Unrelated fix (fix for #3506)

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -2026,6 +2026,14 @@
     <desc>Machines directory location</desc>
   </entry>
 
+  <entry id="EXTRA_MACHDIR">
+    <type>char</type>
+    <default_value></default_value>
+    <group>case_def</group>
+    <file>env_case.xml</file>
+    <desc>Path to an extra directory containing supplementary machines files</desc>
+  </entry>
```
billsacks, line 2035: It looks like I'll need to make a similar change for cmeps/nuopc. I'll open a PR with that change soon.

billsacks:
Note: I am deliberately NOT using this mechanism for the workflow xml files: it feels most intuitive and useful to limit this to the location of machine porting files.

src/drivers/mct/cime_config/config_component.xml
billsacks: Done - see https://github.com/ESCOMP/CMEPS/pull/62

billsacks:
@fischer-ncar I wanted to call to your attention the dependency of this PR on https://github.com/ESCOMP/CMEPS/pull/62 for cases with the NUOPC driver. We'll want to bring these into an alpha tag together.

fischer-ncar:
@billsacks thanks for the heads up.

PR #3515, opened 2020-05-01
Highres jra bgc

mnlevy1981:
Updates to CIME used for 0.1 degree JRA-forced G run with BGC enabled. So far that is just

1. Updating filename in `ROF2OCN_LIQ_RMAPNAME` and `ROF2OCN_ICE_RMAPNAME` because I got it wrong in f078f12
1. Updating `ATM_NCPL` and `OCN_NCPL` for 0.1 degree ocean grids (`ATM_NCPL` update only applies to `DATM` / `SLND` / `CICE` / `POP` runs)

Test suite: just testing by hand so far, will update in comments when more testing is available
Test baseline: N/A
Test namelist changes: N/A
Test status: all test suites should be bit-for-bit, I don't think there are high-res runs being tested

User interface changes?: No

Code review: setting up this draft PR to make code review a little simpler


mnlevy1981:
This got the okay from Frank (one of the OMWG co-chairs) as well as @klindsay28 so I think it's ready to be merged in as-is. Thanks!

PR #3519, opened 2020-05-05
fix atm level spec so that it does not match 1x1_brazil

### LINKED ISSUE #3510: scripts_regression_tests needs to not be so sensitive to current dir
The main function should chdir to `os.path.dirname(__file__)` so that it works the same regardless of where it is run from.

### LINKED ISSUE #3518: Support for `--res 1x1_brazil` broken
@jedwards4b #3492 broke the support for `--res 1x1_brazil`. With the new code, the `gridvalue['atm'] = 1x1_bra` while the `gridvalue['lnd'] = 1x1_brazil`. In E3SM, I created the case via:

```
./create_newcase -case 1x1_brazil.I1850CLM45 -res 1x1_brazil -compset I1850CLM45
```

jedwards4b:
PR #3492 was too general in matching grids, refine to only match z\d to a vertical grid. 

Test suite: scripts_regression_tests.phy 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3518 
Fixes #3510 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/tests/scripts_regression_tests.py
+++ b/scripts/tests/scripts_regression_tests.py
@@ -3567,6 +3567,8 @@ OR
     GLOBAL_TIMEOUT = ns.timeout
     NO_TEARDOWN    = ns.no_teardown
 
+    os.chdir(os.path.dirname(__file__))
+
     if ns.machine is not None:
         MACHINE = Machines(machine=ns.machine)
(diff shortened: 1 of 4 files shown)


jgfouca:
Waiting for review from @bishtgautam

PR #3530, opened 2020-05-11
Fix div0s in cprnc comparisons

### LINKED ISSUE #3525: avoiding div0 in cprnc
I had been running into an issue with cprnc while running regression tests with CTSM-FATES on cheyenne.

https://github.com/ESCOMP/CTSM/pull/1007#issuecomment-626871444

When comparing two arrays (base and test), if the base had only zeros in an array, but the test had non-zero values, it was enabling a comparison between the two. However, the normalization factor was using the zero value, and hence trying to calculate a div0.  

*Note, in this case, the arrays that were generating the div0 were integer arrays defining the dimensions.

I hacked up some modification, and was able to get the test to stop trying to divide by zero. It also identified that the fields were different in the final summary.

Here are my modifications:
https://github.com/ESMCI/cime/compare/master...rgknox:fix-div0-compare_vars

I'm not too familiar with the code, so I didn't want to issue a PR with any changes until some others looked at this and made suggestions.

rgknox:
When comparing two arrays (base and test), if the base had only zeros in an array, but the test had non-zero values, it was enabling a comparison between the two. However, the normalization factor was using the zero value, and hence trying to calculate a div0.

Tests: I performed a build test on the final changes.  The changes, as per commit 
ceae267, were necessary in completing tests for this PR: https://github.com/NGEET/fates/pull/647

Test baseline: NA 
Test namelist changes: None
Test status: NA

Fixes #3525

User interface changes?: no

Update gh-pages html (Y/N)?: no

Code review:

### DIFF
--- a/tools/cprnc/compare_vars_mod.F90.in
+++ b/tools/cprnc/compare_vars_mod.F90.in
@@ -364,10 +364,10 @@ contains
           ! can be sensitive to outliers).
           if (n1 > 0 .and. n2 > 0 .and. rms > 0) then
              rms_normalized_denom = (avgval(1) + avgval(2)) / 2.0
-             if(abs(rms_normalized_denom)>1.e-20_r8)then
+             if(abs(rms_normalized_denom)>0)then
                 rms_normalized = rms / rms_normalized_denom
              else
-                rms_normalized = 0
+                rms_normalized = huge(rms)
              end if
           else
              ! don't try to compute rms_normalized in any of the following conditions:
--- a/tools/cprnc/compare_vars_mod.F90.in
+++ b/tools/cprnc/compare_vars_mod.F90.in
@@ -364,7 +364,11 @@ contains
           ! can be sensitive to outliers).
           if (n1 > 0 .and. n2 > 0 .and. rms > 0) then
              rms_normalized_denom = (avgval(1) + avgval(2)) / 2.0
-             rms_normalized = rms / rms_normalized_denom
+             if(abs(rms_normalized_denom)>1.e-20_r8)then
+                rms_normalized = rms / rms_normalized_denom
+             else
+                rms_normalized = 0
+             end if
           else
              ! don't try to compute rms_normalized in any of the following conditions:
              ! n1 = 0 -- then we won't have avgval(1)


billsacks:
@rgknox thank you for opening this PR. These changes are fine, although I want to make some notes about what I think is the underlying cause of this problem, now that I have looked at it more closely:

In particular, I think you may have misdiagnosed the details of what's going on in your summary (which is completely excusable: this cprnc code is really tangled and in desperate need of a rewrite):

> When comparing two arrays (base and test), if the base had only zeros in an array, but the test had non-zero values, it was enabling a comparison between the two. However, the normalization factor was using the zero value, and hence trying to calculate a div0.

In fact, it looks like this line:

```Fortran
rms_normalized_denom = (avgval(1) + avgval(2)) / 2.0
```

is correctly using both the base and test arrays in determining the denominator. I think the problem is that `rms_normalized_denom` is declared as an integer in this case. In your case (if I understand it correctly), avgval(1) = 0, avgval(2) = 1. So this leads to (0 + 1) / 2.0 = 0.5. But when storing 0.5 in an integer, it gets truncated to 0.

The problem goes beyond this `rms_normalized_denom` variable, though: For example, the avgval values have the same issue, since they are also integers in this case. So, for example, if array1 was `[0, 1, 1, 0, 0]` and array2 was `[1, 0, 0, 1, 0]` then both `avgval` values would end up as 0 – so just changing `rms_normalized_denom` to a real isn't enough to solve this problem.

I think the ideal solution would be to change many of the terms in these calculations to be declared as `real(r8)` rather than as `{VTYPE}`. However, this would take more analysis than I have time for at the moment. For now, I'm happy to bring in your workaround for the issue.

billsacks:
@jedwards4b and/or @jgfouca - are you okay with bringing this in as is? As I note above, I believe this is a workaround for a more significant underlying issue when comparing integer fields. I personally feel that it's fine to bring in this workaround for now, but chime in if you'd prefer _not_ to have this workaround brought in, but instead wait for a fix for the underlying issue (i.e., having a bunch of intermediate variables declared as integer rather than real).

rgknox:
Your explanation does make sense @billsacks .

billsacks:
With #3531, I'm not sure if the changes in this PR are still necessary. However, I'm going to go ahead and merge it because it seems like it doesn't hurt to have some extra code to make sure we avoid dividing by 0.

PR #3538, opened 2020-05-15
Jgfouca/branch for acme split 2020 05 15

jgfouca:
Changes list:

Features:
* create_test: print important resource usage info at start

Bug fixes:
* archive_base: cprnc.out files should never be treated as hist files
* mct: fluxes to the river model are not always from the land
* Fix PIO xml case setup: Component-specific settings need to be set after generic settings so that they take precedence.
* cime_bisect: Submodule operations must now be recursive
* buildTools macro writer: Fix Makefile macro generator to support negation

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


src/drivers/mct/main/cime_comp_mod.F90
```diff
@@ -1323,6 +1323,7 @@ subroutine cime_init()
        write(logunit,F00) 'Initialize each component: atm, lnd, rof, ocn, ice, glc, wav, esp, iac'
        call shr_sys_flush(logunit)
     endif
+    call seq_infodata_GetData(infodata, cime_model=cime_model)
```
jedwards4b, line 1326: pr #3536 already addresses this - I believe this should be removed.

billsacks: I thought that too, at first, but then I wondered about the timing of when `cime_model` is set vs. when it's needed. It's probably cleanest if this call is made in a single place fairly early in initialization.

config/e3sm/machines/Depends.summit.pgigpu
```diff
@@ -0,0 +1,222 @@
+
+ACCFLAGS :=  -Minline -ta=nvidia,cc70,fastmath,loadcache:L1,unroll,fma,managed,ptxinfo -Mcuda -Minfo=accel
+
+advect_scalar.o : advect_scalar.F90
```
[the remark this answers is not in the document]
jgfouca: This is deprecated (CMake is the preferred build method) so I'm not going to sweat this.

jgfouca:
I've asked @rljacob to look at the mct changes.

rljacob:
@billsacks I made the changes you requested.

src/drivers/mct/main/cime_comp_mod.F90
jedwards4b: This should be removed, it is already called at line 3431 of this file.  @billsacks I don't understand your 'but'.

billsacks: Looking more closely at the calling sequence of various routines, I agree that I think things would work if this were removed. However, I'm uncomfortable only having this at line 3431, which is inside the routine `cime_printlogheader`: It happens right now that that routine is called before `cime_model` is needed, but it's very non-obvious that the setting of `cime_model` in that routine – which is about printing information – is critical for the correct operation of the code. What if someone reworked that print routine to no longer need `cime_model`, for example, or changed when it is called in initialization? In these cases, a seemingly innocent change would lead to a hard-to-understand bug.

My thinking was to avoid this problem by either:

1. Getting rid of both of the current calls to `seq_infodata_GetData(infodata, cime_model=cime_model)`, instead making this call as early as possible in either `cime_pre_init1` or `cime_pre_init2`, where it will be more obvious that it's needed.

2. Make `cime_model` a local variable in the subroutines that need it rather than a module-level variable, and calling `seq_infodata_GetData(infodata, cime_model=cime_model)` from each subroutine that needs this variable – because (as we saw from the bug that led to #3536 as well as the need for the addition here) the use of module-level data can be more error-prone.

I'm fine with either of the above solutions.

jgfouca: @billsacks , done. I went with option 1 since it seemed easier.

billsacks: @jgfouca thanks, but as I look at the code, I'm wondering if the new location is now too early: It looks like the infodata object may not be initialized until the start of `cime_pre_init2`:

```Fortran
    !| Initialize infodata

    if (len_trim(cpl_inst_tag) > 0) then
       call seq_infodata_init(infodata,nlfilename, GLOID, pioid, &
            cpl_tag=cpl_inst_tag)
    else
       call seq_infodata_init(infodata,nlfilename, GLOID, pioid)
    end if
```

So I'm thinking that the new call may need to be after that.

jgfouca: @billsacks , thanks. Restarting my testing.

billsacks: Looks good to me now, thanks.

jgfouca:
OK, testing checks-out and I think all comments have been addressed. Merging.

PR #3539, opened 2020-05-16
Wpcoomb/improve cime buildlib function debug logging

### LINKED ISSUE #3532: python logger debug statements aren't printed when used in "buildlib".
While updating CAM's buildlib script, it was found that `logger.debug` statements are not printed to any file in the case, even if the log level is set to debug in buildlib itself.  However, `logger.info` statements do make it to the `*.bldlog.*` file. 

This seems to hold even if the `DEBUG` variable is set to True in `env_build.xml`, and if the `--debug` flag is provided to `case.build`.  It also doesn't seem to be CAM-specific, as the same behavior happens in CLM's buildlib script as well (although I didn't test other components' buildlib scripts).

I specifically used the CESM `release-cesm2.1.3` tag on Izumi to run these tests (although I ran a similar test on Cheyenne with the same results).

Any ideas as to why this is the happening would be greatly appreciated.  Thanks!

WesCoomber:
The buildlib scripts for external components are called from subthreads.
We now forward the --debug option if used to the python buildlib scripts.
If a component's buildlib is called as a subprocess, then 
the logging options (eg. --debug or --silent) do get passed down now.

Test baseline: 
Test namelist changes: 
Test status: [bit for bit]

Fixes #3532 

User interface changes?: N

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -3,7 +3,7 @@ functions for building CIME models
 """
 import glob, shutil, time, threading, subprocess, imp
 from CIME.XML.standard_module_setup  import *
-from CIME.utils                 import get_model, analyze_build_log, stringify_bool, run_and_log_case_status, get_timestamp, run_sub_or_cmd, run_cmd, get_batch_script_for_job, gzip_existing_file, safe_copy
+from CIME.utils                 import get_model, analyze_build_log, stringify_bool, run_and_log_case_status, get_timestamp, run_sub_or_cmd, run_cmd, get_batch_script_for_job, gzip_existing_file, safe_copy, check_for_python
 from CIME.provenance            import save_build_provenance as save_build_provenance_sub
 from CIME.locked_files          import lock_file, unlock_file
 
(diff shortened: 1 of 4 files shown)


WesCoomber:
@jedwards4b Yes I have tested that this works.  `case.build --debug` successfully changes the logger level of subprocesses for building components and debug messages like "logger.debug("wpc7.buildlib HERE I AM")" are recorded in a log file.

WesCoomber:
@jedwards4b Yes I have tested that this works.  `case.build --debug` successfully changes the logger level of subprocesses for building components and debug messages like "logger.debug("wpc7.buildlib HERE I AM")" are recorded in a log file.

scripts/lib/CIME/build.py
```diff
@@ -423,6 +423,11 @@ def _build_model_thread(config_dir, compclass, compname, caseroot, libroot, bldr
     if get_model() != "ufs":
         compile_cmd = "SMP={} {}".format(stringify_bool(smp), compile_cmd)
 
+    if (check_for_python(cmd)):
```
jgfouca, line 426: You don't need to wrap if statements in parens.

scripts/lib/CIME/build.py
```diff
@@ -423,6 +423,11 @@ def _build_model_thread(config_dir, compclass, compname, caseroot, libroot, bldr
...
+        logging_options = get_logging_options()
+        if (logging_options != ""):
```
jgfouca, line 428: ditto

scripts/lib/CIME/build.py
WesCoomber: Good catch. Resolved in commit b87f893.

scripts/lib/CIME/build.py
WesCoomber: Resolved in commit b87f893.

PR #3543, opened 2020-05-19
add mvsource script

### LINKED ISSUE #3542: Keeping an old case running when a username changes
One of our clusters recently updated its accounting and a bunch of usernames changed.

As you might expect, this plays havoc with cases already created in a user's $HOME directory with the old name.

There should be a way to fix this without re-creating the case form scratch.

Here's what we tried so far:
use xmlchange to update: CASEROOT, CIMEROOT, SRCROOT, MACHDIR
do ./case.setup --reset

When trying to do /.case.build, there's an error because the softlinks in $CASEDIR/Tools are still pointing to the old pathname (with the old username) and the Makefile can't be found.   Shouldn't case.setup --rest remake those links?   Is that the last step?

jedwards4b:
Add script cime/scripts/Tools/movesrc which accepts two arguments caseroot and cimeroot
and updates a case directory with these paths.  If either the case directory or the code source is moved use this script to correct the case and keep it functional. 

Test suite: 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3542 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/Tools/mvsource
+++ b/scripts/Tools/mvsource
@@ -17,11 +17,10 @@ except:
     print("ERROR: {} does not appear to be a valid CIMEROOT directory\n".format(cimeroot))
     exit
 
-logger = logging.getLogger(__name__)
 
 # simple way to verify we are in a caseroot
 if not os.path.exists(os.path.join(caseroot,"env_case.xml")):
-    logger.critical("ERROR: {} does not appear to be a valid CASEROOT directory\n".format(caseroot))
+    print("ERROR: {} does not appear to be a valid CASEROOT directory\n".format(caseroot))
     exit
 
 for dirpath, dirnames, filenames in os.walk(caseroot):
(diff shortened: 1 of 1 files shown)


jedwards4b:
I'll work on a test, but wanted to put this up for review and comments.

billsacks:
At a glance, this looks like useful functionality - thanks for working on it. Are you planning to change the script to look more like other cime tools, both from a coding perspective and from a user interface perspective - e.g., use of argparse, logging, etc.?

jedwards4b:
@billsacks I started with the assumption that this would need to work without access to the cime python library, but then I did back off that assumption a little but I don't think it can use the argparse stuff because you can't access it until you have the cimeroot.    @jgfouca what do you think?

billsacks:
You could use python's argparse library for argument parsing without having access to cimeroot. But I see your point about not being able to access some of the other cime stuff until you know where you're getting it from.

jedwards4b:
@rljacob you were the one to request this feature - do you want to review?

PR #3545, opened 2020-05-21
allow compiler to modify MAX_TASKS_PER_NODE and MAX_MPITASKS_PER_NODE

### LINKED ISSUE #3544: MAX_TASKS_PER_NODE is static for each machine entry
For the E3SM-MMF configuration on Summit we have configurations to run the CRM on either CPUs or GPUs depending on whether the compiler is pgi or pgigpu, which requires changing MAX_TASKS_PER_NODE and MAX_MPITASKS_PER_NODE. However, in config_machines.xml these entries cannot be set based on the compiler, so we have created a separate machine entry "summit-cpu" which is identical to "summit" except for the task per node setting. 

We would like some way to fix this in order to get rid of the redundant machine configuration because it's causes headaches every time the summit config needs to be updated.

jedwards4b:
Allows a compiler attribute to be added to the MAX_TASKS_PER_NODE and MAX_MPITASKS_PER_NODE elements in config_machines.xml

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3544 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -254,7 +254,7 @@ class GenericXML(object):
         """
         This is the critical function, its interface and performance are crucial.
 
-        You can specify attributes={key:None} if you want to select chilren
+        You can specify attributes={key:None} if you want to select children
         with the key attribute but you don't care what its value is.
         """
         root = root if root is not None else self.root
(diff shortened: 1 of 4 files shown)


scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -292,6 +292,14 @@ def get_child(self, name=None, attributes=None, root=None, err_msg=None):
 
     def get_optional_child(self, name=None, attributes=None, root=None, err_msg=None):
         children = self.get_children(root=root, name=name, attributes=attributes)
+        if len(children) > 1:
```
jgfouca, line 295: I'm wondering why this code is needed. get_children should already be able to down-select based on attributes.

jedwards4b: I also realize I should add this to get_child and will update with that changed.
If attributes is None then get_children doesn't check to see if the node has any attributes - it just returns all matching nodes.

PR #3547, opened 2020-05-21
Buildsub

### LINKED ISSUE #3532: python logger debug statements aren't printed when used in "buildlib".

jedwards4b:
Allows threads to call run_sub_or_cmd for calling buildlib 

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3532 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/standard_module_setup.py
+++ b/scripts/lib/CIME/XML/standard_module_setup.py
@@ -1,4 +1,3 @@
-
 # pragma pylint: disable=unused-import
 
 import logging, os, sys, re
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -3,7 +3,9 @@ functions for building CIME models
 """
 import glob, shutil, time, threading, subprocess, imp
 from CIME.XML.standard_module_setup  import *
-from CIME.utils                 import get_model, analyze_build_log, stringify_bool, run_and_log_case_status, get_timestamp, run_sub_or_cmd, run_cmd, get_batch_script_for_job, gzip_existing_file, safe_copy, check_for_python, get_logging_options
+from CIME.utils                 import get_model, analyze_build_log, stringify_bool, run_and_log_case_status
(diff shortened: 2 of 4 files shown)


jgfouca:
Didn't @WesCoomber 's solution to #3532 work?

Also, I'm a little worried about how we can trust that components' buildlib scripts won't make thread unsafe changes. We would need to somehow verify that none of them did any chdir or modified os.environ.

jgfouca:
OK, I just read some of the comments on #3532 , so I guess Wesley's solution didn't work for you guys. He did some testing and it looks to me like it should work, so I'd like to understand why his solution didn't work before we pursue using run_or_sub_cmd, which raises a lot of complex issues.

jedwards4b:
I think that you can test this all inside cime - just add a logger.debug or logger.info call inside one of the data model buildlib scripts.

jgfouca:
I'm pretty sure that's exactly what @WesCoomber tested. Wes, can you chime in here?

WesCoomber:
@jgfouca I spent some time checking my tests for the subprocess feature, #3539, and they still work on Melvin. So I am following up in the original issue thread with the steps I used to verify the feature is working correctly and we will figure out the issue on the other machines. Passing the same logging options as was used for current executable to subprocesses should work.

PR #3555, opened 2020-05-28
When writing .env_mach_specific files, only write settings for main job

### LINKED ISSUE #3553: env_mach_specific files should just contain settings for main job
The hidden `.env_mach_specific.sh` and `.env_mach_specific.csh` files contain settings that apply for all jobs, one after the other. This is a problem on cheyenne, at least, where we set `MPI_USE_ARRAY=false` on the share queue (which is used by the short-term archiver), but we don't want this setting for the main job. Currently, if you source the env_mach_specific file, you'll end up with this undesired setting.

I propose changing these files so that they only contain settings that are applicable to the main job in a case (case.run or case.test).

I will open a PR with this change soon. @jedwards4b @jgfouca you can let me know if you think this should not be changed for some reason that I'm not aware of.

### LINKED ISSUE #3554: configure tool doesn't honor --output-dir for the .env_mach_specific files
In testing my fix for #3553, I noticed that the standalone configure tool doesn't honor `--output-dir` for the `.env_mach_specific` files: It puts most of the content of these files in the current working directory rather than in the directory specified by --output-dir. The fix is relatively simple. I'll include this in the PR where I fix #3553 .

billsacks:
The main change here is:

When writing .env_mach_specific files, only write settings for main job

Previously, the .env_mach_specific.sh and .env_mach_specific.csh files
had settings for all jobs. At least on cheyenne, this led to the
settings for the st_archive job being written after the settings for the
main job - in particular, setting MPI_USE_ARRAY=false. I typically want
the settings for the main job. This commit changes the behavior so that
the env_mach_specific files just contain the settings for the main job.

Note that this doesn't change the behavior when running the standalone
configure tool, because I'm not sure how to fix that case.

An additional fix is:

configure tool: honor --output-dir for .env_mach_specific files

Previously, these files were being written in the current working
directory, even if --output-dir was given.

Test suite: Manual testing: (1) ran configure tool before and after on cheyenne; (2) examined .env_mach_specific.sh files before and after on cheyenne (both for a non-test case and for a test case); (3) scripts_regression_tests on cheyenne
Test baseline: n/a
Test namelist changes: 
Test status: bit for bit

Fixes #3553 
Fixes #3554 

User interface changes?: N

Code review:

### DIFF
--- a/scripts/lib/CIME/BuildTools/configure.py
+++ b/scripts/lib/CIME/BuildTools/configure.py
@@ -97,7 +97,7 @@ def _generate_env_mach_specific(output_dir, machobj, compiler, mpilib, debug,
         logger.warning("{} already exists, delete to replace".format(ems_path))
         return
 
-    ems_file = EnvMachSpecific(output_dir, unit_testing=unit_testing)
+    ems_file = EnvMachSpecific(output_dir, unit_testing=unit_testing, standalone_configure=True)
     ems_file.populate(machobj)
     ems_file.write()
 
--- a/scripts/lib/CIME/BuildTools/configure.py
+++ b/scripts/lib/CIME/BuildTools/configure.py
@@ -107,7 +107,7 @@ def _generate_env_mach_specific(output_dir, machobj, compiler, mpilib, debug,
     fake_case = FakeCase(compiler, mpilib, debug, comp_interface)
     ems_file.load_env(fake_case)
     for shell in ('sh', 'csh'):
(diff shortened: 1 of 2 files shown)


billsacks:
@jedwards4b @jgfouca - the main issue I'm fixing here (#3553 ) has caused problems for me on cheyenne. However, I don't know if the current behavior is perhaps intentional; if so, feel free to let me know.

billsacks:
I have run scripts_regression_tests on cheyenne on this branch; all tests pass.

PR #3559, opened 2020-06-01
Esmf datamodels vector stream update

jedwards4b:
Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/share/streams_nuopc/dshr_strdata_mod.F90
+++ b/src/share/streams_nuopc/dshr_strdata_mod.F90
@@ -23,7 +23,7 @@ module dshr_strdata_mod
   use dshr_stream_mod  , only : shr_stream_getnextfilename, shr_stream_getprevfilename, shr_stream_getData
   use dshr_tinterp_mod , only : shr_tInterp_getCosz, shr_tInterp_getAvgCosz, shr_tInterp_getFactors
   use dshr_methods_mod , only : dshr_fldbun_getfldptr, dshr_fldbun_getfieldN, dshr_fldbun_fldchk, chkerr
-  use dshr_methods_mod , only : dshr_fldbun_diagnose, dshr_fldbun_regrid
+  use dshr_methods_mod , only : dshr_fldbun_diagnose, dshr_fldbun_regrid, dshr_field_getfldptr
 
   use pio              , only : file_desc_t, iosystem_desc_t, io_desc_t, var_desc_t
   use pio              , only : pio_openfile, pio_closefile, pio_nowrite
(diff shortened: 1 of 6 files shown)


src/share/streams_nuopc/dshr_strdata_mod.F90
```diff
@@ -71,13 +70,14 @@ module dshr_strdata_mod
...
      character(CS), allocatable          :: fldlist_stream(:)               ! names of stream file fields
      character(CS), allocatable          :: fldlist_model(:)                ! names of stream model fields
-     type(ESMF_FieldBundle)              :: fldbun_stream_lb                ! stream n field bundle for lb of time period (stream grid)
-     type(ESMF_FieldBundle)              :: fldbun_stream_ub                ! stream n field bundle for ub of time period (stream grid)
-     type(ESMF_FieldBundle)              :: fldbun_model_lb                 ! stream n field bundle for lb of time period (model grid)
-     type(ESMF_FieldBundle)              :: fldbun_model_ub                 ! stream n field bundle for ub of time period (model grid)
-     type(ESMF_FieldBundle)              :: fldbun_model                    ! stream n field bundle for model time (model grid)
-     integer                             :: ustrm                           ! index of vector u in stream
-     integer                             :: vstrm                           ! index of vector v in stream
+     integer                             :: stream_lb
```
mvertens, line 73: Please add a comment to what this is. It would be good to have every attribute in these derived types commented.

src/share/streams_nuopc/dshr_strdata_mod.F90
```diff
@@ -71,13 +70,14 @@ module dshr_strdata_mod
...
+     integer                             :: stream_ub
+     type(ESMF_Field)                    :: field_stream                    ! a field on the stream data domain
```
mvertens, line 75: Can you please clarify if the field has the stream file name or the model filename. This was a huge source of confusion in the past.

src/share/streams_nuopc/dshr_strdata_mod.F90
```diff
@@ -870,48 +818,21 @@ subroutine shr_strdata_advance(sdat, ymd, tod, logunit, istr, timers, rc)
...
-                if (ChkErr(rc,__LINE__,u_FILE_u)) return
-             endif
-             call t_stopf(trim(lstr)//trim(timname)//'_map')
           endif
 
           call t_stopf(trim(lstr)//trim(timname)//'_readLBUB')
 
        enddo ! end of loop over streams
-
+#ifdef DOTHIS
```
mvertens, line 826: So the vector field implementation has not been tested yet? If you have tested it - then we don't need this if-def. In fact it looks like you are handling the vector streams in another place - so why is this still here?

jedwards4b:
@mvertens I think that I've addressed your issues - I also created a difference file comparing test
SMS_Vnuopc_Ln11_D.f19_g17_rx1.A.cheyenne_intel to the apr26 baseline for this case.  
/glade/scratch/jedwards/SMS_Vnuopc_Ln11_D.f19_g17_rx1.A.cheyenne_intel.C.20200601_084934_gul611/diffat19800.nc
there may be some reason for concern about differences in U and V around the poles.

PR #3573, opened 2020-06-11
add short data type and decompression for era5

jedwards4b:
Add support for compressed ERA5 data streams

Test suite: hand tested with ERS_Vnuopc_D.f09_g17.A.cheyenne_intel modified for era5
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/share/streams_nuopc/dshr_strdata_mod.F90
+++ b/src/share/streams_nuopc/dshr_strdata_mod.F90
@@ -5,7 +5,7 @@ module dshr_strdata_mod
 
   use ESMF
 
-  use shr_kind_mod     , only : r8=>shr_kind_r8, cs=>shr_kind_cs, cl=>shr_kind_cl, cxx=>shr_kind_cxx, r4=>shr_kind_r4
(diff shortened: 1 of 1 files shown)


src/share/streams_nuopc/dshr_strdata_mod.F90
```diff
@@ -1266,6 +1271,16 @@ subroutine shr_strdata_readstrm(sdat, ns, stream, stream_mesh, &
           rcode = pio_get_att(pioid, varid, "_FillValue", fillvalue_r4)
        else if (pio_iovartype == PIO_DOUBLE) then
           rcode = pio_get_att(pioid, varid, "_FillValue", fillvalue_r8)
+       else if (pio_iovartype == PIO_SHORT) then
```
uturuncoglu, line 1274: I think that there could be short data type without add_offset and scale_factor such as land-sea mask etc. So, i think the interface just need to check those attributes and use them if they are available otherwise the data can be passed directly from PIO.

jedwards4b: I think that I would like to wait until we encounter such a case before we deal with it - is that okay with you?

uturuncoglu: Okay, that is fine. Thanks for the short data type support!

PR #3589, opened 2020-06-24
Clean up screen output & logging in check_input_data

### LINKED ISSUE #3588: Change check_input_data output to avoid using logging.warning for normal output?
I ran into this issue when calling `case.check_all_input_data` from my own script (using CIME as a python library). So I didn't call `CIME.utils.parse_args_and_handle_standard_logging_options` (because I use a somewhat different logging configuration than CIME does). When I did this, I got a ton of WARNING messages; here is a small set of them:

```
WARNING:   Model clm missing file paramfile = '/Users/sacks/projects/scratch/lilac_usermach_0624a/inputdata/lnd/clm2/paramdata/clm5_params.c200519.nc'
WARNING:   Model clm missing file fsnowoptics = '/Users/sacks/projects/scratch/lilac_usermach_0624a/inputdata/lnd/clm2/snicardata/snicar_optics_5bnd_c090915.nc'
WARNING:   Model clm missing file fsnowaging = '/Users/sacks/projects/scratch/lilac_usermach_0624a/inputdata/lnd/clm2/snicardata/snicar_drdt_bst_fit_60_c070416.nc'
WARNING:   Model clm missing file finidat = '/Users/sacks/projects/scratch/lilac_usermach_0624a/inputdata/lnd/clm2/initdata_map/clmi.I2000Clm50BgcCrop.2011-01-01.1.9x2.5_gx1v7_gl4_simyr2000_c190312.nc'
WARNING:   Model clm missing file stream_fldfilename_urbantv = '/Users/sacks/projects/scratch/lilac_usermach_0624a/inputdata/lnd/clm2/urbandata/CLM50_tbuildmax_Oleson_2016_0.9x1.25_simyr1849-2106_c160923.nc'
WARNING: Client protocol gftp not enabled
WARNING: Client protocol wget not enabled
WARNING: Client protocol None not enabled
WARNING: Client protocol gftp not enabled
WARNING: Client protocol wget not enabled
```

It seems like all of the output from this function is using warnings, when in fact many of these messages seem to be just normal, expected operation – in particular, the messages that files are missing should not (in my view) be warnings, but rather are exactly what you expect. It is therefore misleading to have `WARNING` appear on these messages, and it means that true issues get lost, since both true issues and expected output have the same `WARNING` prefix.

When running the `check_input_data` script from a case, this issue doesn't arise, presumably because of how logging is set up in CIME. However, I feel that we should follow best practices in this respect, and reserve `logging.warning` for true warnings. Based on the guidance in https://docs.python.org/3/howto/logging.html#when-to-use-logging, I feel that the messages about missing files should be made using `print` statements, since my impression is that these fall under the category of "Display console output for ordinary usage of a command line script or program".

billsacks:
Changes to check_input_data screen output / logging to make it slightly less verbose and reduce the presence of warnings that are really expected behavior. I'll explain the details in line comments.

Test suite: Manual tests plus scripts_regression_tests - just A_RunUnitTests and B_CheckCode
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes #3588 

User interface changes?: A bit of reduced output from check_input_data

Code review:

### DIFF
--- a/scripts/lib/CIME/case/check_input_data.py
+++ b/scripts/lib/CIME/case/check_input_data.py
@@ -24,7 +24,7 @@ def _download_checksum_file(rundir):
     while protocol is not None:
         protocol, address, user, passwd, chksum_file,_ = inputdata.get_next_server()
         if protocol not in vars(CIME.Servers):
-            logger.warning("Client protocol {} not enabled".format(protocol))
+            logger.info("Client protocol {} not enabled".format(protocol))
             continue
         logger.info("Using protocol {} with user {} and passwd {}".format(protocol, user, passwd))
         if protocol == "svn":
@@ -50,7 +50,7 @@ def _download_checksum_file(rundir):
         full_path = os.path.join(rundir, local_chksum_file)
         new_file = full_path + '.raw'
         protocol = type(server).__name__
-        logging.info("Trying to download file: '{}' to path '{}' using {} protocol.".format(rel_path, new_file, protocol))
+        logger.info("Trying to download file: '{}' to path '{}' using {} protocol.".format(rel_path, new_file, protocol))
         tmpfile = None
         if os.path.isfile(full_path):
             tmpfile = full_path+".tmp"
@@ -122,7 +122,7 @@ def _download_if_in_repo(server, input_data_root, rel_path, isdirectory=False, i
     full_path = os.path.join(input_data_root, rel_path)
     if ic_filepath:
         full_path = full_path.replace(ic_filepath, "/")
-    logging.info("Trying to download file: '{}' to path '{}' using {} protocol.".format(rel_path, full_path, type(server).__name__))
+    logger.info("Trying to download file: '{}' to path '{}' using {} protocol.".format(rel_path, full_path, type(server).__name__))
     # Make sure local path exists, create if it does not
     if isdirectory or full_path.endswith(os.sep):
         if not os.path.exists(full_path):
@@ -288,7 +288,7 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
     no_files_missing = True
     if download:
         if protocol not in vars(CIME.Servers):
-            logger.warning("Client protocol {} not enabled".format(protocol))
+            logger.info("Client protocol {} not enabled".format(protocol))
             return False
         logger.info("Using protocol {} with user {} and passwd {}".format(protocol, user, passwd))
         if protocol == "svn":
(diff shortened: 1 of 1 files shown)


scripts/lib/CIME/case/check_input_data.py
```diff
@@ -24,7 +24,7 @@ def _download_checksum_file(rundir):
     while protocol is not None:
         protocol, address, user, passwd, chksum_file,_ = inputdata.get_next_server()
         if protocol not in vars(CIME.Servers):
-            logger.warning("Client protocol {} not enabled".format(protocol))
+            logger.info("Client protocol {} not enabled".format(protocol))
```
billsacks, line 27: This one felt to me like expected behavior, so I downgraded it from a warning to info. (It is still printed when running the `check_input_data` script because of how cime sets up logging, but if you set up logging to suppress info statements, this no longer appears.)

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -288,7 +288,7 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
     no_files_missing = True
     if download:
         if protocol not in vars(CIME.Servers):
-            logger.warning("Client protocol {} not enabled".format(protocol))
+            logger.info("Client protocol {} not enabled".format(protocol))
```
billsacks, line 291: As above: downgraded from warning to info.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -356,7 +356,7 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
                         isdirectory=rel_path.endswith(os.sep)
 
                         if ("/" in rel_path and not os.path.exists(full_path) and not full_path.startswith('unknown')):
-                            logger.warning("  Model {} missing file {} = '{}'".format(model, description, full_path))
+                            print("Model {} missing file {} = '{}'".format(model, description, full_path))
```
billsacks, line 359: Same rationale as above.

billsacks:
@jedwards4b @jgfouca I'm done making line comments. I'm not tied to any of these changes if you disagree with my rationale on them.

scripts/lib/CIME/case/check_input_data.py
```diff
@@ -50,7 +50,7 @@ def _download_checksum_file(rundir):
         full_path = os.path.join(rundir, local_chksum_file)
         new_file = full_path + '.raw'
         protocol = type(server).__name__
-        logging.info("Trying to download file: '{}' to path '{}' using {} protocol.".format(rel_path, new_file, protocol))
+        logger.info("Trying to download file: '{}' to path '{}' using {} protocol.".format(rel_path, new_file, protocol))
```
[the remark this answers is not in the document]
jedwards4b: Good catch, yes it should be logger.

PR #3499, opened 2020-04-22
fix issue with mpi-serial io

jedwards4b:
The mpi-serial library was not interpreting a derived mpi datatype correctly, I found that forcing the alltoall interface when using a single processor fixes the problem.  It also allows 0 as a valid value for pio_rearr_comm_max_pend_req_comp2io and pio_rearr_comm_max_pend_req_io2comp.

Test suite: scripts_regression_tests.py, SMS_D_Ln9_Mmpi-serial.ne5_ne5_mg37.FADIAB.cheyenne_intel 
Test baseline: 
Test namelist changes: 
Test status: bit for bit,

Fixes

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

PR #3594, opened 2020-06-26
Fix max pending requests default

jayeshkrishna:
Using a new default value, -2, for PIO_REARR_COMM_MAX_PEND_REQ_COMP2IO to indicate that CIME needs to set the value of maximum I/O pending requests, since 0 is now a valid value for the variable.

Also see related PR #3499 

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes

User interface changes?: 

Code review:

### DIFF
--- a/src/drivers/mct/cime_config/config_component.xml
+++ b/src/drivers/mct/cime_config/config_component.xml
@@ -2400,10 +2400,10 @@
   <entry id="PIO_REARR_COMM_MAX_PEND_REQ_COMP2IO">
     <type>integer</type>
     <valid_values></valid_values>
-    <default_value>0</default_value>
+    <default_value>-2</default_value>
     <group>run_pio</group>
     <file>env_run.xml</file>
-    <desc>pio rearranger communication max pending requests (comp2io) : 0 implies that CIME internally calculates the value ( = max(64, 2 * PIO_NUMTASKS) ), -1 implies no bound on max pending requests </desc>
+    <desc>pio rearranger communication max pending requests (comp2io) : -2 implies that CIME internally calculates the value ( = max(64, 2 * PIO_NUMTASKS) ), -1 implies no bound on max pending requests </desc>
   </entry>
 
   <entry id="PIO_REARR_COMM_ENABLE_HS_COMP2IO">
@@ -2430,7 +2430,7 @@
     <default_value>64</default_value>
     <group>run_pio</group>
     <file>env_run.xml</file>
-    <desc>pio rearranger communication max pending requests (io2comp) : -1 implies no bound on max pending requests </desc>
+    <desc>pio rearranger communication max pending requests (io2comp) : -2 implies that CIME internally calculates the value ( = 64), -1 implies no bound on max pending requests </desc>
   </entry>
 
   <!-- Handshaking is a very bad idea for large process counts and for -->
--- a/src/share/util/shr_pio_mod.F90
+++ b/src/share/util/shr_pio_mod.F90
@@ -872,17 +872,6 @@ contains
       else
         buf(8) = 0
       end if
-
-      ! Log the rearranger options
-      write(shr_log_unit, *) "PIO rearranger options:"
-      write(shr_log_unit, *) "  comm type     =", pio_rearr_comm_type
-      write(shr_log_unit, *) "  comm fcd      =", pio_rearr_comm_fcd
-      write(shr_log_unit, *) "  max pend req (comp2io)  =", pio_rearr_comm_max_pend_req_comp2io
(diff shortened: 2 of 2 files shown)


billsacks:
@jayeshkrishna or @jedwards4b -

I have a namelist file with pio settings that is somewhat outside of the cime infrastructure. It currently sets:

```
  pio_rearr_comm_max_pend_req_comp2io = 0
  pio_rearr_comm_max_pend_req_io2comp = 64
```

Does the change in this PR imply that I should change the first setting to -2, or can/should this be left at 0?

jayeshkrishna:
@jedwards4b would be the right person to comment on that, since the value would depend on the behavior that you need with PIO.
Before PR #3499 a value of 0 would result in CIME calculating it for you (to say 64 and would result in PIO using point to point messages for data rearrangement) and after the PR  a value of 0 would be directly passed on to PIO (and it would result in PIO using MPI_Alltoallw() for data rearrangement).
So after merging this PR setting the value to -2 would force CIME to calculate it for you (the first case above - the behavior before PR #3499) and a value of 0 would force PIO to use collectives for data rerrangement (the second case above - the behavior after PR #3499)

billsacks:
Thanks @jayeshkrishna .

@jedwards4b can you confirm that a setting of `pio_rearr_comm_max_pend_req_comp2io = 0` still makes sense in the Fortran namelist for LILAC?

jedwards4b:
No, it should probably be changed to -2.

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -2400,10 +2400,10 @@
   <entry id="PIO_REARR_COMM_MAX_PEND_REQ_COMP2IO">
     <type>integer</type>
     <valid_values></valid_values>
-    <default_value>0</default_value>
+    <default_value>-2</default_value>
     <group>run_pio</group>
     <file>env_run.xml</file>
-    <desc>pio rearranger communication max pending requests (comp2io) : 0 implies that CIME internally calculates the value ( = max(64, 2 * PIO_NUMTASKS) ), -1 implies no bound on max pending requests </desc>
+    <desc>pio rearranger communication max pending requests (comp2io) : -2 implies that CIME internally calculates the value ( = max(64, 2 * PIO_NUMTASKS) ), -1 implies no bound on max pending requests </desc>
```
jedwards4b, line 2406: Should add here that 0 implies use MPI_ALLTOALL

billsacks:
> No, it should probably be changed to -2.

Okay, will do - thanks.

src/drivers/mct/cime_config/config_component.xml
jayeshkrishna: There is another variable, PIO_REARR_COMM_TYPE, that was intended to control whether the user wants a collective or point to point messages for data rearrangement.

jedwards4b: So when I submitted PR #3499 which you reviewed and approved - why was this not mentioned?  The implication is that #3499 should have set PIO_REARR_COMM_TYPE instead.

jayeshkrishna: I haven't looked at the PIO code in a while, I assumed that you wanted to use the maximum pending requests setting to force the collective usage in PIO. In any case, a value of 0 can be considered a valid setting for the maximum pending requests and we need to move to a different value for the default setting (to indicate CIME calculate the value for the user)

jedwards4b: Perhaps making that change to #3499 would remove the need for this PR?

jayeshkrishna: AFAICT, this behavior (using the comm type to choose the behavior) would also require change in PIO. In any case, this PR allows users to set maximum pending requests = 0, if needed.

rljacob:
#3499 was already merged.  What exactly do you want to do with it?

rljacob:
@jayeshkrishna Open an issue describing the problem created by PR #3499 and discuss there how to fix it.

jayeshkrishna:
ok, I will create a github issue and we can discuss on how to fix the issue

jayeshkrishna:
PR #3597 fixes the issue fixed here (hence this PR is obsolete)

jayeshkrishna:
@billsacks : Just FYI, PR #3597 fixes the issue in a way that does not require you to change the value of pio_rearr_comm_max_pend_req_comp2io in the older namelists (no need to change your old namelists).

billsacks:
Thanks for the note, @jayeshkrishna . If I understand the code correctly, it looks like it will work to set `pio_rearr_comm_max_pend_req_comp2io` to either 0 or -2. So I'll keep the -2 setting for now because that will work with multiple cime versions. Please correct me if I'm wrong.

PR #3599, opened 2020-06-27
Wpcoomb/improve error message for create newcase test option

### LINKED ISSUE #3595: create_newcase --test mode does not create a viable case without create_test
The XML_PHASE of tests has bugged me for a while. Every other phase of a the case-control system can be run standalone in a case via a command-line tool, but the XML_PHASE can only be run through create_test. If a user copy/pastes the create_newcase command, and then tries to run `./case.setup` and `./case.build`, the build will fail with `No variable CCSM_CPRNC found in case`. Even if we were to fix this particular problem, the behavior of the case will still not match create_test since so many other important env XML settings would be missing.

At the very least, I'd like `create_newcase ... --test ...` to throw a very stern warning to the user if it detects that it was invoked outside of create_test.

I think a better, but more difficult, solution would be to merge the functionality of the xml_phase into either create_newcase or case_setup.

WesCoomber:
Calling create_newcase ... --test ... from the command line
now throws a CIMEError along with a very stern warning message 
to the user when it detects that it was invoked outside of create_test.

Test baseline:
Test namelist changes:
Test status: [bit for bit]
Fixes #3595
User interface changes?: N
Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -224,7 +224,6 @@ def _main_func(description):
         expect(("FROM_CREATE_TEST" in os.environ and os.environ["FROM_CREATE_TEST"] == "True"),
            "The --test argument is intended to only be called from inside create_test. Invoking this option from the command line is not appropriate usage.")
 
-
     with Case(caseroot, read_only=False) as case:
         # Configure the Case
         case.create(casename, srcroot, compset, grid, user_mods_dir=user_mods_dir,
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -217,6 +217,14 @@ def _main_func(description):
     expect(not (os.path.exists(caseroot) and not test),
            "Case directory {} already exists".format(caseroot))
 
+    # create_newcase ... --test ... throws a CIMEError along with
+    # a very stern warning message to the user
+    # if it detects that it was invoked outside of create_test
+    if test:
+        expect(("FROM_CREATE_TEST" in os.environ and os.environ["FROM_CREATE_TEST"] == "True"),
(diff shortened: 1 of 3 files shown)


scripts/create_newcase
```diff
@@ -217,6 +217,14 @@ def _main_func(description):
     expect(not (os.path.exists(caseroot) and not test),
            "Case directory {} already exists".format(caseroot))
 
+    # create_newcase ... --test ... throws a CIMEError along with
+    # a very stern warning message to the user
+    # if it detects that it was invoked outside of create_test
+    if test:
+        expect(("FROM_CREATE_TEST" in os.environ and os.environ["FROM_CREATE_TEST"] == "True"),
```
jgfouca, line 224: We only want a warning, not an error.

scripts/create_test
```diff
@@ -633,6 +633,7 @@ def _main_func(description):
 
     success = False
     run_count = 0
+    os.environ["FROM_CREATE_TEST"] = "True"
```
jgfouca, line 636: you can remove this

jgfouca:
I see @jedwards4b wanted it to be an error. I guess I'm fine with that.

scripts/create_test
WesCoomber: Good catch-- this is not needed because the test_scheduler.py change will catch the error.

WesCoomber: I removed the extra line in create_test.

scripts/create_newcase
WesCoomber: JimE mentioned in the original issue that the message should be an error, so I am resolving this for now.

PR #3600, opened 2020-06-28
Gptl threaded

jedwards4b:
This THREADED_OMP flag and the compiler SMP flag is needed when compiling gptl with SMP. 

Test suite: by hand, SMS_D_P36x3_Ln9.f09_f09_mg17.QSC6.cheyenne_intel.cam-outfrq9s
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/share/timing/Makefile
+++ b/src/share/timing/Makefile
@@ -7,7 +7,7 @@ VPATH := $(GPTL_DIR)
 ifndef MOD_SUFFIX
    MOD_SUFFIX := mod
 endif
-ifeq ($(SMP),TRUE)
+ifeq ($(strip $(SMP)),TRUE)
   CPPDEFS +=  -DTHREADED_OMP
   compile_threaded=TRUE
 endif
--- a/src/share/timing/Makefile
+++ b/src/share/timing/Makefile
@@ -7,7 +7,10 @@ VPATH := $(GPTL_DIR)
 ifndef MOD_SUFFIX
    MOD_SUFFIX := mod
 endif
-
+ifeq ($(SMP),TRUE)
+  CPPDEFS +=  -DTHREADED_OMP
+  compile_threaded=TRUE
+endif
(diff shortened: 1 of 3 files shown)


src/share/timing/Makefile
```diff
@@ -7,7 +7,10 @@ VPATH := $(GPTL_DIR)
 ifndef MOD_SUFFIX
    MOD_SUFFIX := mod
 endif
-
+ifeq ($(SMP),TRUE)
```
billsacks, line 10: In cime's main makefile, we have:

```Makefile
ifeq ($(strip $(SMP)),TRUE)
```

Note the use of `strip`. Should this also use `strip` for consistency?

jedwards4b:
@billsacks gptl was built threaded but did not have additional code added for individual threads because the THREADED_OMP flag was not on.  So prior to 3591 GPTL was built threaded sorta, after that it was not built threaded, this PR builds with full thread support.

jedwards4b:
@billsacks I would like this change in for the release - would you have another look please.

billsacks:
@jedwards4b thanks for the explanation; this makes more sense now. Can you just comment on whether this should use `strip`, as is done for the main cime Makefile? (See my inline comment above.)

src/share/timing/Makefile
jedwards4b: sure, I've updated it to use strip

jedwards4b:
reverts really screw things up - I tried reapplying #3591 here but it won't take.  I'll open another pr

billsacks:
> reverts really screw things up - I tried reapplying #3591 here but it won't take. I'll open another pr

For future reference, I'm curious what you did that didn't work? Did you try cherry-picking the given commit?

PR #3601, opened 2020-06-29
Revert "fix force_build_smp, did not work after case.setup"

billsacks:
Reverts ESMCI/cime#3591

Apparently #3591 caused problems with the gptl build. #3600 is supposed to fix those problems, but I have some outstanding questions there, and @jedwards4b won't be able to help resolve those until next week. So, in order for us to move forward with CESM testing, I'd like to revert #3591 for now. In the next couple of weeks, we will revisit #3591 together with #3600 to determine the appropriate fix.

### DIFF
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -19,7 +19,7 @@ _CMD_ARGS_FOR_BUILD = \
 
 def get_standard_makefile_args(case, shared_lib=False):
     make_args = "CIME_MODEL={} ".format(case.get_value("MODEL"))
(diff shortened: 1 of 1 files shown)


billsacks:
@amametjanov I'm adding you as a reviewer since you were a reviewer on both #3591 and #3600 . If you're able to give this your approval (or let us know if you have issues with this temporary revert) in the next day, that would be ideal, so we can move ahead with testing. Please let us know if that will be a problem.

billsacks:
@amametjanov thanks for your quick turnaround of this. If you're able to explain why @jedwards4b needed to make the changes he did in #3600 then we could move forward with that; if not, I'll wait for him to be able to comment on that.

PR #3609, opened 2020-07-07
Refactor ensemble.py and single_run.py for python3

johnsonb-ucar:
Edit print statements in each script to use the print_function.
Include from __future__ import print_function to ensure uniform
behavior in both python2 and python3.

Assignments of j and ll within ensemble.get_pert_lim no longer
assume the python2 behavior in which integer division always
returns an integer. Integer division returns a float in python3
when the dividend is not evenly divisible by the divisor. Refactor
uses integer type conversion to ensure the assignments of j and ll
work identically in python2 and python3.

The assignment of ippt within ensemble.get_pert_lim no longer
returns a ValueError in python3. Refactor uses string type
conversion, the zfill method and concatenation to ensure the
assignment of ptlim works identically in python2 and python3.

User interface changes?: N

### DIFF
--- a/tools/statistical_ensemble_test/ensemble.py
+++ b/tools/statistical_ensemble_test/ensemble.py
@@ -1,4 +1,5 @@
 #!/usr/bin/python
+from __future__ import print_function
 import os, sys, getopt
 import random
 from single_run import process_args_dict, single_case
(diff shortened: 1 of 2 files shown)


johnsonb-ucar:
My mistake -- I used integer type conversion too late in the operation when assigning ll in ensemble.get_pertlim_uf. The second push uses the type conversion immediately after the division so the value assigned to ll is the same in both python2 and python3.

jedwards4b:
In the future please create a feature branch in your fork and apply the PR from your branch instead of from master.

johnsonb-ucar:
Yup, will do.

PR #3610, opened 2020-07-07
Jgfouca/branch for acme split 2020 07 07

jgfouca:
Change list

Features:
* Add paths to I/O libraries that is required for static builds with Scorpio.
* mct-configure: Initial support for the Intel GPU compiler on the JLSE iris nodes
* Add Antarctica RRM support for F2010SC5-CMIP6 compset
* create_test/test_scheduler: model builds will never take more than 16 procs in the proc pool

Fixes:
* The maximum I/O pending requests setting in env_run.xml, PIO_REARR_COMM_MAX_PEND_REQ_COMP2IO, is by default set to 0. CIME needs to set the variable to an appropriate value in this case (when its set to 0). Re-introducing this behavior that was disabled by ESMCI/cime#3499.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


src/components/data_comps_nuopc/dshr_nuopc/dshr_mod.F90
```diff
@@ -0,0 +1,1497 @@
+module dshr_mod
+
+  use NUOPC            , only : NUOPC_CompAttributeGet, NUOPC_CompFilterPhaseMap
```
jedwards4b, line 3: This file should be removed.

src/drivers/nuopc/cime_config/config_component.xml
```diff
@@ -0,0 +1,2527 @@
+<?xml version="1.0"?>
```
jedwards4b, line 1: This file should be removed.

src/externals/mct/configure
```diff
@@ -4311,6 +4311,10 @@ case $ac_fc_v_output in
   *mGLOB_options_string*)
     ac_fc_v_output=`echo $ac_fc_v_output | sed 's/"-mGLOB[^"]*"/ /g'` ;;
 
+  # With Intel ifx, remove -loopopt=0
```
jedwards4b, line 4314: Could you please provide some detail re this change.

src/share/streams_nuopc/dshr_methods_mod.F90
```diff
@@ -0,0 +1,595 @@
+module dshr_methods_mod
```
jedwards4b, line 1: This directory should be removed.

rljacob:
@jgfouca why are you doing this before merging https://github.com/E3SM-Project/E3SM/pull/3657 ?

jgfouca:
@rljacob , i can always do another merge/split. I think we are still a ways from having CIME as a submodule in e3sm.

rljacob:
Ok but doing that PR results in less code to merge/split.

jgfouca:
Yeah, I should have done it first, but at this point it's easier to just continue with this split.

src/externals/mct/configure
jgfouca: Here is the PR: https://github.com/E3SM-Project/E3SM/pull/3654

jgfouca:
Ty, @jedwards4b. Seeing a few anomalies in my testing that I will fix and then merge.

PR #3614, opened 2020-07-09
Addition of new aux_cdeps tests to CDEPS

mvertens:
Addition of new aux_cdeps tests to CDEPS

Test suite: aux_cdeps
Test baseline: generated baselines for jul09 (since these are new tests there is nothing to compare agains)
Test namelist changes: NA
Test status: NA
Fixes: None
User interface changes?: No
Code review:

### DIFF
--- a/config/xml_schemas/testlist.xsd
+++ b/config/xml_schemas/testlist.xsd
@@ -16,7 +16,7 @@
         <xs:element minOccurs="1" maxOccurs="1" ref="machines"/>
         <xs:element minOccurs="0" ref="options"/>
       </xs:sequence>
-      <xs:attribute name="compset" use="required" type="xs:NCName"/>
+      <xs:attribute name="compset" use="required" type="xs:string"/>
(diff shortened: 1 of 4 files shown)


config/xml_schemas/testlist.xsd
```diff
@@ -16,7 +16,7 @@
         <xs:element minOccurs="1" maxOccurs="1" ref="machines"/>
         <xs:element minOccurs="0" ref="options"/>
       </xs:sequence>
-      <xs:attribute name="compset" use="required" type="xs:NCName"/>
+      <xs:attribute name="compset" use="required" type="xs:string"/>
```
jedwards4b, line 19: That this change was required leads me to think that perhaps some new compset name violates the naming convention - can you explain?

mvertens: This was needed in order to have the long case name in the test - which is what was needed for the aux_cdeps tests.
If I recall - you pointed out that we needed to change this.

PR #3622, opened 2020-07-17
Fix build clean for e3sm

jgfouca:
Was broken for things still built by classic Makefile (sharedlibs). Add a regression test to ensure it stays working.

Also, make _create_test return casedirs which is very convenient for a large number of tests.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes https://github.com/E3SM-Project/E3SM/issues/3653

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


jgfouca:
@billsacks , no, just needed a review from you since JimE is out.

PR #3625, opened 2020-07-20
Make hist_utils a bit more robust to existing cprnc.out files

### LINKED ISSUE #3606: Remove any existing cprnc output files before running cprnc
I should preface this by saying that this showed up for me in a rather screwey test case. I'm not 100% sure this will show up in all testing, but looking at what I think is the offending code, I think it would.

I ran a system test that did baseline comparisons; this passed (including the baseline comparisons - it was bfb with baselines). Then I made a change that caused the results to differ from the baselines and reran the test. To my surprise, this reported that the BASELINE phase PASSed.

Digging into this, it looks like the output from cprnc is appended to the end of any existing cprnc.out file, rather than overwriting the old file. Then this whole file (now containing multiple sets of output) is read back in, and the first thing the code looks for is an occurrence of "files seem to be IDENTICAL". Here is the relevant code:

https://github.com/ESMCI/cime/blob/13a8d5c5e89cc10719a0b23acf86331ab83c9448/scripts/lib/CIME/hist_utils.py#L346-L374

In this case, both "files seem to be IDENTICAL" and "the two files seem to be DIFFERENT" appear in the output file; the code treats this as being identical, even though the most recent run led to a difference.

I'm labeling this as CRITICAL because I view any bug that causes tests to be reported as a PASS when they should be a FAIL to be a critical issue.

I don't have time to fix this right now, given other things that I'm scrambling to finish before taking a few days off. But I'd suggest:

- Definitely: removing any existing file (i.e., the `output_filename`) before running cprnc, so that only the most recent output appears in there. (If the appending was by design, then the old file could be saved away somewhere rather than just removed.)

- Possibly: reordering the logic in the bottom half of the above code snippet: The safest thing to do is to first check for any failure conditions and then finally check for "files seem to be IDENTICAL" as the last check. This way, if we do run into a situation where both "DIFFERENT" and "IDENTICAL" appear in the output, we err on the side of treating this as DIFFERENT, which seems safest.

@jgfouca @jedwards4b @WesCoomber

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3606 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @billsacks

### DIFF
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -343,6 +343,10 @@ def cprnc(model, file1, file2, case, rundir, multiinst_driver_compare=False, out
     if outfile_suffix:
         output_filename += ".{}".format(outfile_suffix)
 
+    # Remove existing output file if it exists
+    if os.path.exists(output_filename):
+        os.remove(output_filename)
+
     if outfile_suffix is None:
         cpr_stat, out, _ = run_cmd("{} -m {} {}".format(cprnc_exe, file1, file2), combine_output=True)
     else:
@@ -360,9 +364,7 @@ def cprnc(model, file1, file2, case, rundir, multiinst_driver_compare=False, out
             # have no differences.
             files_match = " 0 had non-zero differences" in out
         else:
-            if "files seem to be IDENTICAL" in out:
-                files_match = True
-            elif "the two files seem to be DIFFERENT" in out:
+            if "the two files seem to be DIFFERENT" in out:
                 files_match = False
             elif "the two files DIFFER only in their field lists" in out:
                 if ignore_fieldlist_diffs:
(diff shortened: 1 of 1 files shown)


scripts/lib/CIME/hist_utils.py
```diff
@@ -343,6 +343,10 @@ def cprnc(model, file1, file2, case, rundir, multiinst_driver_compare=False, out
     if outfile_suffix:
         output_filename += ".{}".format(outfile_suffix)
 
+    # Remove existing output file if it exists
+    if os.path.exists(output_filename):
+        os.remove(output_filename)
+
```
billsacks, line 349: My read of the logic in this function is that this removal should only be done in the `else` block a few lines down – otherwise, nothing is written to `output_filename`, and in fact `output_filename` may not be writeable (and thus may not be removable) due to permission issues. Do you agree?

PR #3635, opened 2020-07-23
PET Test Baseline Generation and Comparison Improvement

### LINKED ISSUE #2916: PET test does baseline generation and comparison twice

WesCoomber:
Fixed the PET test issue with baseline generation and baseline comparison happening twice in resubmit system tests.

Test suite: scripts_regression_tests (both batch and non-batch)
Test baseline: 
Test namelist changes: 

Fixes #2916 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


scripts/lib/CIME/SystemTests/system_tests_common.py
```diff
@@ -145,20 +145,37 @@ def run(self, skip_pnl=False):
...
             self._resetup_case(RUN_PHASE)
+            do_baseline_ops = True
             with self._test_status:
                 self._test_status.set_status(RUN_PHASE, TEST_PEND_STATUS)
 
+            # We do not want to do multiple repetitions of baseline operations for
+            # multi-submit tests. We just want to do them upon the final submission.
+            # Other submissions will need to mark those phases as PEND to ensure wait_for_tests
+            # waits for them.
+            if self._case.get_value("BATCH_SYSTEM") != "none":
```
billsacks, line 156: Can you explain why this is only done when there is a batch system? If you're running on a machine without a batch system, my intuition is that you'd still only want to do the baseline generation and comparison after the final piece of the test, but I don't have a good enough understanding of the relevant pieces of cime to know for sure.

jgfouca: @billsacks , I just confirmed that things work as intended on non-batch machines. I'm sure this has something to do with the fact that non-batch machines don't do multiple submissions and therefore things stay in one process.

PR #3658, opened 2020-08-11
Improve check_input_data.py to use os.path.isfile() rather than os.pa…

### LINKED ISSUE #3654: check_input_data check passed for directory, does not confirm if listing is a file
check_input_data appears only to check for the existence of elements listed in the input_data_list, and not whether or not they are files. For example, check_input_data will pass the following:

Buildconf/cice.input_data_list
grid_file = /work/02441/bcash/stampede2/UFS/ufs_inputdata/
kmt_file = /work/02441/bcash/stampede2/UFS/ufs_inputdata/

The subsequent run will fail, however, because grid_file and kmt_file are required.

If I understand the code correctly (which I probably do not) the issue is that check_input_data.py is using os.path.exists in lines 342 and 358 to perform the existence check, but os.path.exists does not distinguish between files and directories.

WesCoomber:
Improve check_input_data.py to use os.path.isfile() rather than os.path.exists() to better distinguish between files and directories.

Test baseline: 
Test namelist changes: 
Test status: [bit for bit]

Fixes #3654 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jgfouca

### DIFF
--- a/scripts/lib/CIME/case/check_input_data.py
+++ b/scripts/lib/CIME/case/check_input_data.py
@@ -339,7 +339,7 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
                         # User pointing to a file outside of input_data_root, we cannot determine
                         # rel_path, and so cannot download the file. If it already exists, we can
                         # proceed
-                        if not os.path.exists(full_path):
+                        if not os.path.isfile(full_path):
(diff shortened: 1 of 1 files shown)


jgfouca:
Thx @WesCoomber !

PR #3660, opened 2020-08-11
CIME changes to support upcoming addition of MALI gis1to10km configuration

jonbob:
This PR makes two small changes in CIME that are necessary prior to an E3SM PR to bring in the MALI gis1to10km configuration. The two changes are:
* add mpas.gis1to10km as option for GLC_GRID; and
* increase i90_MXDEP from 4 to 5 in mpeu.

Neither change should have any impact on current configurations.

[BFB]

### DIFF
--- a/src/drivers/mct/cime_config/config_component.xml
+++ b/src/drivers/mct/cime_config/config_component.xml
@@ -1062,7 +1062,7 @@
 
   <entry id="GLC_GRID">
     <type>char</type>
-    <valid_values>gland20,gland10,gland5,gland5UM,gland4,mpas.aisgis20km,mpas.gis20km,mpas.ais20km,null</valid_values>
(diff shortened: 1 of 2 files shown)


rljacob:
Whats the mpeu change for?

jonbob:
@rljacob - it was necessary to get the full B-case configuration with MALI to run. Otherwise it complained about an overflow. I don't understand that part of the coupler well, so I'd appreciate any insight you have

jonbob:
Thanks @jgfouca

jonbob:
@jgfouca, @rljacob - what is the process then? Do I need to make another PR to bring this version of CIME into E3SM?

rljacob:
yes.

jonbob:
@rljacob - a PR for CIME master will bring in a lot more changes than just mine. We are very careful in MPAS to avoid this situation

rljacob:
We always want to track CIME master.  How would you propose updating the submodule in E3SM with just your changes?

PR #3666, opened 2020-08-17
Cmeps driver env var

jedwards4b:
We needed to add a way to differentiate between different drivers using the same coupler.

Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/ufs/config_inputdata.xml
+++ b/config/ufs/config_inputdata.xml
@@ -8,6 +8,7 @@
   <!-- it will be searched for filename and chksum of each downloaded file.  -->
   <!-- see the file ftp://ftp.cgd.ucar.edu/cesm/inputdata_chksum.dat for proper format. -->
 
+  <!-- server for sample data -->
   <server>
     <comment>ftp site for ufs release</comment>
     <protocol>ftp</protocol>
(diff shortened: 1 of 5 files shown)


rsdunlapiv:
@jedwards4b would you consider another name than `CMEPS_DRIVER`?  Drivers and different than mediators, so you might have a configuration not using CMEPS that still needs to set its driver.

jedwards4b:
@rsdunlapiv This code is specific to the cmeps mediator - we are telling cmeps what driver it should use so the name seems to me to be entirely appropriate.

rsdunlapiv:
@jedwards4b I disagree.  CMEPS actually does not know what driver it uses, which is why it works both under NEMS and with the CESM/CIME driver.  That's by design.  This variable is setting what driver to use for the coupled system, regardless of whether CMEPS is a part of it.

uturuncoglu:
Yes. The variable name can be changed to more convenient one such as DRIVER_NAME?

mvertens:
I agree with @rsdunlapiv regarding the naming convention. CMEPS needs to be treated like any other component and I plan to actually move it out of CIME and with the other CESM components in an upcoming PR.

jedwards4b:
How about UFS_DRIVER?  I am concerned that this solves a very particular problem and the suggested names are too general implying this may do more that it is intended to.

uturuncoglu:
I updated CMEPS_DRIVER as UFS_DRIVER.

PR #3670, opened 2020-08-18
add a warning for tests

### LINKED ISSUE #3669: xmlchange should warn if a var is in env_test.xml
When using xmlchange on a test with a variable defined in env_test.xml the change is overwritten by the value in env_test.xml.  We should add a warning to the user with the following text:
Warning:  This variable is defined in env_test.xml and will be overwritten if you wish to change the value for the test add the flags --file env_test.xml to your xmlchange command line.

jedwards4b:
Add a warning if the user attempts to change a value set by env_test.xml - the xmlchange request is still honored but a warning is printed to stdout.   This is because values in env_test.xml overwrite values in other files by design in tests and this is not always apparent to the user. 

Test suite: scripts_regression_tests.py, hand testing of xmlchange
Test baseline: 
Test namelist changes: 

Fixes #3669 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/Tools/xmlchange
+++ b/scripts/Tools/xmlchange
@@ -129,9 +129,12 @@ def parse_command_line(args, description):
 
     return args.caseroot, listofsettings, args.file, args.id, args.val, args.subgroup, args.append, args.noecho, args.force , args.dryrun
 
-def xmlchange_single_value(case, xmlid, xmlval, subgroup, append, force, dryrun):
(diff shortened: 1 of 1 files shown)


billsacks:
Thanks @jedwards4b - I have run into this gotcha a number of times; I like how you have fixed it!

PR #3678, opened 2020-08-19
Add flexibility in dlnd domain_fracname

### LINKED ISSUE #3667: E3SM PR #3619 resulted in undesirable change in dlnd landfrac when used in sno mode
@fischer-ncar noticed answer changes in T compsets (active glc driven by dlnd) in CESM due to #3619 . We still need to do one more test to 100% confirm the cause of this, but I'm pretty sure it's due to the changes in `src/components/data_comps_mct/dlnd/src/dlnd_comp_mod.F90` in that PR. Looking at the history, it looks like @apcraig made those changes. I see that the old lines were commented out, which makes me wonder if perhaps the change was meant to be temporary?

We'd like this to be changed back to how it was before, but we don't want to just make this change and risk breaking E3SM. Can @apcraig or someone else involved in these changes please weigh in ASAP on this, because is holding up our release testing (and I'm labeling it as critical for that reason).

@jgfouca @rljacob pinging you two so that you can get someone else appropriate on the line if @apcraig isn't able to respond promptly.

For context on the code in question, see a20cf6de2185380522188ec241e60fe02472412b.

billsacks:
Resolves #3667 in a more general way

Test suite: ERS_Ly7.f09_g17_gl4.T1850G.cheyenne_intel
Test baseline: https://github.com/escomp/cism-wrapper at cism2_1_69
Test namelist changes: none
Test status: bit for bit

Fixes #3667 (again)

User interface changes?: N

Code review: @jgfouca

### DIFF
--- a/src/components/data_comps_mct/dlnd/src/dlnd_comp_mod.F90
+++ b/src/components/data_comps_mct/dlnd/src/dlnd_comp_mod.F90
@@ -133,6 +133,7 @@ CONTAINS
     integer(IN)        :: nflds_glc_nec  ! number of snow fields separated by elev class
     integer(IN)        :: field_num ! field number
     character(nec_len) :: nec_str   ! elevation class, as character string
+    character(*), parameter :: domain_fracname_unset = 'null'
 
     !--- formats ---
     character(*), parameter :: F00   = "('(dlnd_comp_init) ',8a)"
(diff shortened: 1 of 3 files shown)


src/components/data_comps_mct/dlnd/src/dlnd_comp_mod.F90
```diff
@@ -169,12 +170,21 @@ subroutine dlnd_comp_init(Eclock, x2l, l2x, &
...
+       end if
     else
-       call shr_strdata_init(SDLND,mpicom,compid,name='lnd', calendar=calendar, &
-            dmodel_domain_fracname_from_stream=domain_fracname)
+       if (domain_fracname == domain_fracname_unset) then
+          call shr_strdata_init(SDLND,mpicom,compid,name='lnd', calendar=calendar)
+       else
+          call shr_strdata_init(SDLND,mpicom,compid,name='lnd', calendar=calendar, &
+               dmodel_domain_fracname_from_stream=domain_fracname)
+       end if
```
billsacks, line 187: I started down the path of trying to avoid the need for having near duplication of this code, but it was getting messy, so I decided it was simplest just to have these four different subroutine calls based on which arguments are relevant.

billsacks:
@jgfouca @fischer-ncar - I'm going to run a test of a T compset in CESM now to make sure this new change is bfb. @jgfouca if you are happy with this change, can you run whatever testing you'd like to do on your end to verify that this works for e3sm?

billsacks:
Hmmm, while it *looks* good, it apparently doesn't work. I forgot that domain_fracname is not a stream-specific variable... let me see if I can find a fix.

jgfouca:
I'm testing ERS.f19_g16_rx1.A.mappy_gnu now.

billsacks:
@jgfouca please try with the latest commit... I think this will work... will comment definitively soon.

billsacks:
Okay, I have verified that this is still bit-for-bit for a T compset. Specifically, I ran ERS_Ly7.f09_g17_gl4.T1850G.cheyenne_intel from https://github.com/escomp/cism-wrapper master (cism2_1_69) to generate baselines, then cherry-picked the changes in this PR into the cime version used there (cime5.8.28) and verified that results were bit-for-bit.

jgfouca:
@billsacks , ok. your previous commit worked, for what it's worth.

jgfouca:
Works for our A case.

jgfouca:
Shoot,  I just realized our A cases use slnd, not dlnd.

PR #3680, opened 2020-08-20
new cdeps stream schema and stream definition file

mvertens:
New cdeps stream schema and stream definition file

This PR accompanies a CDEPS PR that changed the path for CDEPS and that split the original `namelist_definition_comp.xml` file into 2 files with a new `stream_definition_comp.xml`

Test suite: cdeps_aux tests and testlist_drv.xml test with cheyenne/intel
Test baseline: compared to aug11  and generated aug19 in both cmeps_baselines and cdeps_baselines
Test namelist changes: Yes
Test status: bit for bit

Fixes: None

User interface changes?: No for CIME

Code review:


scripts/lib/CIME/build.py
```diff
@@ -353,7 +354,7 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
     cdeps_build_script = None
     if comp_interface == "nuopc":
         libs.append("CDEPS")
-        cdeps_build_script = os.path.join(cimeroot, "src", "components", "cdeps", "cime_config", "buildlib")
+        cdeps_build_script = os.path.join(srcroot, "components", "cdeps", "cime_config", "buildlib")
```
jedwards4b, line 357: This should use config_files and not assume a path.

mvertens: I am fine doing that - but this was not done in the original implementation so can we please discuss how we waht to have this implemented. I simply changed a path that had already been hard-wired.

PR #3686, opened 2020-08-24
Avoid artificial limit on string lengths in shr_string_listMerge

### LINKED ISSUE #3683: Limit on string length in shr_string_listMerge can sometimes be exceeded
A user who was trying to run CESM with 36 glacier elevation classes ran into a string length limit in `shr_string_listMerge` that caused the model to abort. The fix is relatively straightforward. I'll run some tests and submit a patch soon.

billsacks:
Test suite: create_test cime_developer on cheyenne
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes #3683 

User interface changes?: none

Code review: @jedwards4b

### DIFF
--- a/src/share/util/shr_string_mod.F90
+++ b/src/share/util/shr_string_mod.F90
@@ -1237,7 +1237,8 @@ contains
     !EOP
 
     !----- local -----
-    character(SHR_KIND_CX) :: l1,l2   ! local char strings
+    character(len=len(list1)) :: l1
+    character(len=len(list2)) :: l2
     integer(SHR_KIND_IN)   :: rCode   ! return code
     integer(SHR_KIND_IN)   :: t01 = 0 ! timer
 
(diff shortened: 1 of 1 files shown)


billsacks:
I would say this is *not* important for CESM2.2, but is probably relatively safe to include in CESM2.2 if that's convenient.

src/share/util/shr_string_mod.F90
```diff
@@ -1237,7 +1237,8 @@ subroutine shr_string_listMerge(list1,list2,listout,rc)
     !EOP
 
     !----- local -----
-    character(SHR_KIND_CX) :: l1,l2   ! local char strings
+    character(len=len(list1)) :: l1
```
jedwards4b, line 1240: Are you sure that this initialization is supported by all our current compilers?

PR #3693, opened 2020-08-27
Support concurrent env XML changes during case.run

### LINKED ISSUE #3338: Cannot change RESUBMIT while model is running
In one of my recent simulation, I tried to increase the number of automatic resubmissions using:

`./xmlchange -id RESUBMIT -val 6`

Unfortunately, when the next job started, it stopped with the following error:

`ERROR: File /compyfs/gola749/E3SM_simulations/20191216.alpha20.piControl.ne30_r05_oECv3_ICG.compy/case_scripts/env_run.xml appears to have changed without a corresponding invalidation, modtimes 1577796971.00 != 1577892633.00 `

Resubmitting the job manually afterwards with

`./case.submit`

worked and the job ran.

jgfouca:
This became mostly an exercise in coming up with a good test.

Adds a new type of FakeTest that uses both a custom shell script and a built model exe.

Adds a new system test to see that, if the user does this during case.run:
```
./xmlchange DOUT_S=TRUE
./xmlchange DOUT_S=TRUE --file env_test.xml
./xmlchange RESUBMIT=1
./xmlchange STOP_N={2*original_stopn}
./xmlchange CONTINUE_RUN=FALSE
./xmlchange RESUBMIT_SETS_CONTINUE_RUN=FALSE
```
... that CIME will successfully resubmit and run with more timesteps.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3338

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -343,8 +343,7 @@ class GenericXML(object):
     def validate_timestamp(self):
         timestamp_ok = self.check_timestamp()
         expect(timestamp_ok,
-               """File {} appears to have changed without a corresponding invalidation.
-It's possible an entry needs to be moved to env_volatile""".format(self.filename))
+               "File {} appears to have changed without a corresponding invalidation.".format(self.filename))
 
     def write(self, outfile=None, force_write=False):
         """
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -109,7 +109,8 @@ class GenericXML(object):
             self._FILEMAP[infile] = self.CacheEntry(self.tree, self.root, os.path.getmtime(infile))
 
     def read_fd(self, fd):
-        expect(self.read_only or not self.filename or not self.needsrewrite, "Reading into object marked for rewrite, file {}"               .format(self.filename))
+        expect(self.read_only or not self.filename or not self.needsrewrite,
+               "Reading into object marked for rewrite, file {}".format(self.filename))
         read_only = self.read_only
         if self.tree:
             addroot = _Element(ET.parse(fd).getroot())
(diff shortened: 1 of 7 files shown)


scripts/lib/CIME/XML/generic_xml.py
```diff
@@ -329,21 +330,30 @@ def get_version(self):
...
     def write(self, outfile=None, force_write=False):
         """
         Write an xml file from data in self
         """
-        #self.check_timestamp()
-
         if not (self.needsrewrite or force_write):
             return
 
+        self.validate_timestamp()
```
jgfouca, line 355: Restores timestamp checking when a file gets written. I think we need this since any user changes would be blown-away by this write.

scripts/lib/CIME/case/case_run.py
```diff
@@ -111,12 +111,17 @@ def _run_model_impl(case, lid, skip_pnl=False, da_cycle=0):
 
         model_log("e3sm", logger, "{} MODEL EXECUTION BEGINS HERE".format(time.strftime("%Y-%m-%d %H:%M:%S")))
         run_func = lambda: run_cmd_no_fail(cmd, from_dir=rundir)
+        case.flush()
```
jgfouca, line 114: The critical changes are this line and a few lines below.

jgfouca:
@golaz , once this is in, I'd like for you to try to use it.

jgfouca:
As a side note, I still feel CONTINUE_RUN is awkward. My initial test set it to FALSE and I was confused until I realized I also had to set RESUBMIT_SETS_CONTINUE_RUN to FALSE.

jedwards4b:
I don't really understand the issue you are having with CONTINUE_RUN.  And the name RESUBMIT_SETS_CONTINUE_RUN tells you exactly what that variable does.

jgfouca:
@jedwards4b the confusion is that setting it had no effect on my subsequent run until I also set RESUBMIT_SETS_CONTINUE_RUN. I understand why it's there, just wondering if the situation is too confusing for users.

golaz:
Don't get me started on CONTINUE_RUN...

jgfouca:
@billsacks , done.

billsacks:
Thanks @jgfouca . What would you think about putting a similar comment in `case_run`? I thought that's where it may be most needed, because a read-through of that function gives no indication that the call to `_run_model` in that function will result in the whole case object being reloaded.

jgfouca:
I forgot that I had a nice test for checking the timestamp validation code. I just re-enabled it.

jgfouca:
Ah, sorry, should have read your comment more closely.

jgfouca:
@billsacks done

PR #3701, opened 2020-09-10
Change NCPL's for G-case default and grids EC30to60E2r2 and WC14to60E2r3

jonbob:
Set the coupling frequency for new MPAS-Ocean/Seaice grids EC30to60E2r2 and WC14to60E2r3 in E3SM. Also update the default G-case coupling frequency in E3SM.

[BFB]

Test suite: N/A
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: none

Code review:

### DIFF
--- a/src/drivers/mct/cime_config/config_component_e3sm.xml
+++ b/src/drivers/mct/cime_config/config_component_e3sm.xml
@@ -429,6 +429,8 @@
       <value compset="_MPASO" grid="oi%oEC60to30v3">48</value>
       <value compset="_MPASO" grid="oi%oEC60to30wLI">48</value>
       <value compset="_MPASO" grid="oi%ECwISC30to60E1r2">48</value>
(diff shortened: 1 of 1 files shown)


darincomeau:
@jonbob I've been setting `OCN_NCPL` in this file as well for new grids. It should take `ATM_NCPL` as the default, but maybe safer to add it explicitly (and to be consistent with existing grids)?

jonbob:
@darincomeau - that's probably a good idea. Let me fix that right now

darincomeau (approved):
Approved by visual inspection.

jonbob:
@jedwards4b - if you're around today, could you please take a quick peek at this one? We'd like to get it in quickly for some runs we're trying to start. Thanks

PR #3703, opened 2020-09-11
S2sport 

jedwards4b:
Changes in cime for s2s model build.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit,
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/ufs/config_files.xml
+++ b/config/ufs/config_files.xml
@@ -136,7 +136,7 @@
     <default_value>unset</default_value>
     <values>
       <value component="pop"       >$SRCROOT/components/pop/</value>
-      <value component="mom"       >$SRCROOT/src/model/mom/</value>
+      <value component="mom"       >$SRCROOT/src/model/MOM6/</value>
       <value component="nemo"      >$SRCROOT/components/nemo/</value>
       <value component="hycom"     >$SRCROOT/src/model/HYCOM/</value>
       <value component="docn"      >$SRCROOT/components/cdeps/docn</value>
@@ -183,7 +183,7 @@
     <type>char</type>
     <default_value>unset</default_value>
     <values>
-      <value component="cice"      >$SRCROOT/src/model/cice/</value>
+      <value component="cice"      >$SRCROOT/src/model/CICE/</value>
       <value component="dice"      >$CIMEROOT/src/components/cdeps/dice</value>
       <value component="sice"      >$CIMEROOT/src/components/stub_comps_$COMP_INTERFACE/sice</value>
       <value component="xice"      >$CIMEROOT/src/components/xcpl_comps_$COMP_INTERFACE/xice</value>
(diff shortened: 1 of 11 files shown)


config/ufs/config_files.xml
```diff
@@ -183,8 +183,8 @@
     <type>char</type>
     <default_value>unset</default_value>
     <values>
-      <value component="cice"      >$SRCROOT/components/cice/</value>
-      <value component="dice"      >$SRCROOT/components/cdeps/dice</value>
+      <value component="cice"      >$SRCROOT/src/model/CICE/</value>
+      <value component="dice"      >$CIMEROOT/src/components/cdeps/dice</value>
```
jedwards4b, line 187: dice path is wrong

scripts/lib/CIME/build.py
```diff
@@ -139,7 +139,8 @@ def _build_model(build_threaded, exeroot, incroot, complist,
             if "cpl" in l:
                 cpl_in_complist = True
 
-        if cime_model == 'ufs' and not cpl_in_complist:
+#        if cime_model == 'ufs' and not cpl_in_complist:
```
jedwards4b, line 142: Need to fix this.

jedwards4b:
@jgfouca Changes in build.py affect e3sm - the external library paths are now explicit in config_files.xml for each model.

config/ufs/config_files.xml
```diff
@@ -107,7 +107,7 @@
   <entry id="COMP_ROOT_DIR_ATM">
     <type>char</type>
     <values>
-      <value component="datm"      >$SRCROOT/components/cdeps/datm</value>
+      <value component="datm"      >$SRCROOT/src/model/CDEPS/datm</value>
```
uturuncoglu, line 110: we are checking out CDEPS under components/cdeps for HAFs application. I think this needs to be changed and we need to checkout under src/model/CDEPS/ after this PR. Am I right?

config/ufs/config_files.xml
```diff
@@ -122,8 +122,8 @@
   <entry id="COMP_ROOT_DIR_CPL">
     <type>char</type>
     <values>
+      <value component="cpl" >$SRCROOT/src/model/CMEPS</value>
```
uturuncoglu, line 125: This is okay but i think still need to have clear distinction between mediator and driver under CIME. So, maybe we could introduce  COMP_ROOT_DIR_DRV etc.

config/ufs/config_files.xml
jedwards4b: Yes, for s2s I added it as a submodule to ufs-s2s-model, you can do that or keep it in the Externals.cfg file
but you will need the change to manage_externals too.

uturuncoglu (approved):

config/ufs/config_files.xml
jedwards4b: I have opened https://github.com/ESMCI/cime/issues/3712 to address this issue.

uturuncoglu: Thanks. That is great

jgfouca (approved):

PR #3711, opened 2020-09-15
Wpcoomb/improve check input data tool

### LINKED ISSUE #3654: check_input_data check passed for directory, does not confirm if listing is a file

WesCoomber:
I improved check_input_data.py to better 
distinguish between correct files and directories. 

If an unsupported directory is detected, 
the error message is more clear and informative.
I also added a small optimization to the
check_input_data tool for entries that point to /dev/null.

Test baseline:
Test namelist changes:
Test status: [bit for bit]

Fixes #3654

User interface changes?:

Update gh-pages html (Y/N)?:

Code review: @jedwards4b


scripts/lib/CIME/case/check_input_data.py
```diff
@@ -373,6 +373,7 @@ def check_input_data(case, protocol="svn", address=None, input_data_root=None, d
                             if chksum:
                                 verify_chksum(input_data_root, rundir, rel_path.strip(os.sep), isdirectory)
                                 logger.info("Chksum passed for file {}".format(os.path.join(input_data_root,rel_path)))
+                            expect((not isdirectory), "Unsupported directory in input_data_list '{}'".format(full_path))
```
billsacks, line 376: Can you explain the rationale for doing this check at this point? This will only do the check when the given file/directory already exists, as opposed to doing the isdirectory check regardless of existence. Perhaps more importantly, it looks like the other code around here is built to work with isdirectory true or false, so what is the rationale for adding this expectation that it is not a directory?

I don't understand the requirements here, so I'm asking these questions out of naivety.

WesCoomber: Hi Bill, the original issue was opened up by Benjamin Cash, and he was running into trouble with the check_input_data tool passing incorrect entries that pointed to existing directories, when they should have been pointing to existing files. Such as:

Buildconf/cice.input_data_list
grid_file = /work/02441/bcash/stampede2/UFS/ufs_inputdata/
kmt_file = /work/02441/bcash/stampede2/UFS/ufs_inputdata/

The subsequent run will fail, however, because grid_file and kmt_file are required.

I put the check there to fulfill his improvement request of having the check_input_data tool catch the type of errors described in his issue.

billsacks: I guess I'm questioning whether this response to the request is really the right one. It seems like this code has been designed to support directories as well as individual files, and so adding this assertion feels like it's changing this in a not-totally-consistent way. I feel like this needs more discussion. I'm also wondering how this change aligns with the discussion in #3654 , which showed that there are some cases where there are, indeed, directory paths in input data lists.

To summarize, I'm asking two things:

(1) I feel there needs to be broader discussion about whether check_input_data is intended to support directory paths or only individual files

(2) If it is only intended to support individual files, then – at a glance – it looks to me like your addition may be inconsistent with some existing code, and it may be that a bigger rework / cleanup is warranted. Along with this, the check should probably be moved to somewhere higher up in the conditional structure, so that it is checked whether or not a file already exists.

jgfouca:
@WesCoomber , we need to resolve @billsacks 's concerns and get this merged.

WesCoomber:
@billsacks 

Hi Bill,

I revisited this PR and did some more work to meet the original issue author, Benjamin Cash’s, requirements along with addressing your concerns. The input_data_list files with entries that specify directory paths have key values with the name ending with the string “datapath” or “data_path”, for example `“linoz_data_path = /sems-data-store/ACME/inputdata/atm/cam/chem/trop_mozart/ub”` and `“mam_mom_datapath = /sems-data-store/ACME/inputdata/atm/cam/chem/trop_mam/marine_BGC/”`. Additionally, many of the required files in input_data_list files, such as `“grid_file = /sems-data-store/ACME/inputdata/share/domains/domain.ocn.ne4np4_oQU240.160614.nc”`, `“kmt_file = /sems-data-store/ACME/inputdata/share/domains/domain.ocn.ne4np4_oQU240.160618.nc”`, or “mam_mom_filename = /data-store/ACME/inputdata/atm/cam/chem/trop_mam/marine_BGC//monthly_macromolecules_0.1deg_bilinear_latlon_year01_merge_date.nc”, have key value entries ending with the string “file” or “filename”. So, I moved the check to higher up in the conditional structure, so that it is checked whether or not a file already exists. This means the check is consistent with the existing code, and still supports downloading missing files or directories over the network. I also improved the quality and communication of the expect() error message.

Please take a look at this updated PR when you get a chance, so we can get this merged in.

jedwards4b (approved):

billsacks (approved):
Thanks for your rework of this.

PR #3719, opened 2020-09-23
Skip preview namelist

### LINKED ISSUE #2981: --skip-preview-namelists option doesn't work for tests.
I found in an SMS test that this option doesn't work because the tests invoke case_setup.py which calls preview_namelists without consideration for this flag.   An SMS test can run without invoking preview_namelists on the compute nodes but there are other tests which probably cannot.  
We should either fix this or error out if a user trys to use the option in a test.

### LINKED ISSUE #3720: scripts_regression_tests.py test TESTRUNUSERXMLCHANGE problematic
The test writes a script which overwrites itself 
[here](https://github.com/ESMCI/cime/blob/master/scripts/lib/CIME/SystemTests/system_tests_common.py#L757)
On cheyenne the line mv {modelexe} {modelexe}.old overwrites the script currently being executed so that the next line
mv {modelexe}.real {modelexe} 
is never executed and the test fails.

jedwards4b:
the --skip-preview-namelist flag was not being honored by case.submit

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #2981 
Fixes #3720 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -1841,7 +1841,7 @@ This allows using a different mpirun command to launch unit tests
   </machine>
 
 
-  <machine MACH="mappy">
+  <machine MACH="melvin">
     <DESC>Linux workstation for Jenkins testing</DESC>
     <NODENAME_REGEX>(melvin|watson)</NODENAME_REGEX>
     <OS>LINUX</OS>
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -1841,7 +1841,7 @@ This allows using a different mpirun command to launch unit tests
   </machine>
 
 
-  <machine MACH="melvin">
+  <machine MACH="mappy">
     <DESC>Linux workstation for Jenkins testing</DESC>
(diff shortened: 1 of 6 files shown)


config/cesm/machines/config_machines.xml
```diff
@@ -1841,7 +1841,7 @@ This allows using a different mpirun command to launch unit tests
   </machine>
 
 
-  <machine MACH="melvin">
+  <machine MACH="mappy">
```
WesCoomber, line 1844: Hi Jim, it looks like the Travis tests do not like the machine melvin to mappy change.

jedwards4b: yes, I have a fix for this I haven't pushed yet - it was actually mappy-> melvin that messed it up.

jedwards4b:
@WesCoomber all good now.

WesCoomber (approved):
_Thumbs Up._ I see that case.submit with the skip-preview-namelist flag still calls create_namelists and create_dirs during case.run, however it now skips the buildnml cmd within those functions and I trust that this is the intended behavior of the skip-preview-namelist flag.

jgfouca (approved):

PR #3723, opened 2020-09-29
Change name of cam to eam in E3SM

rljacob:
Make necessary changes in E3SM cime config files and other files to allow switching cam to eam for E3SM.

Test suite: e3sm_integration
Test namelist changes: none
Test status: bit for bit

User interface changes?: user_nl_cam changed to user_nl_eam

Code review:

### DIFF
--- a/src/drivers/mct/cime_config/buildnml
+++ b/src/drivers/mct/cime_config/buildnml
@@ -387,6 +387,12 @@ def buildnml(case, caseroot, component):
         if "aquaplanet" in cam_config_opts:
             infile_text = "aqua_planet = .true. \n aqua_planet_sst = 1"
 
+    if case.get_value('COMP_ATM') == 'eam':
+        # cam is actually changing the driver namelist settings
+        cam_config_opts = case.get_value("CAM_CONFIG_OPTS")
+        if "aquaplanet" in cam_config_opts:
+            infile_text = "aqua_planet = .true. \n aqua_planet_sst = 1"
+
     user_nl_file = os.path.join(caseroot, "user_nl_cpl")
     namelist_infile = os.path.join(confdir, "namelist_infile")
     create_namelist_infile(case, user_nl_file, namelist_infile, infile_text)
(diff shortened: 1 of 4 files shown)


rljacob:
Note that I had to make some small changes in driver/mct/cime_config/buildnml and namelise_definition_drv.xml where the atmosphere was assumed to be called "cam".

src/drivers/mct/cime_config/buildnml
```diff
@@ -387,6 +387,12 @@ def buildnml(case, caseroot, component):
         if "aquaplanet" in cam_config_opts:
             infile_text = "aqua_planet = .true. \n aqua_planet_sst = 1"
 
+    if case.get_value('COMP_ATM') == 'eam':
+        # cam is actually changing the driver namelist settings
+        cam_config_opts = case.get_value("CAM_CONFIG_OPTS")
+        if "aquaplanet" in cam_config_opts:
+            infile_text = "aqua_planet = .true. \n aqua_planet_sst = 1"
+
```
billsacks, line 395: Unless I'm missing something, it looks like the bodies of these conditionals are identical. Is there a reason you copied this block of code rather than changing the conditional above to something like `if case.get_value('COMP_ATM') in ['cam', 'eam']`?

jedwards4b (approved):
Approved with the change suggested by @billsacks

src/drivers/mct/cime_config/buildnml
rljacob: They are currently identical but we might change CAM_CONFIG_OPTS to EAM_CONFIG_OPTS later so made the full copy.

billsacks: It seems like that could still be generalized with variable substitution rather than hard-coding "CAM_CONFIG_OPTS", but since this isn't too much duplicated code and I could see how these might need to diverge even further in the future, I don't feel strongly about this myself.

rljacob: I'll make the change you requested since there probably is a better solution to that.

rljacob: What the right syntax for that if clause?  I'm getting a syntax error with what you wrote.

rljacob: I forgot the colon at the end.

billsacks (approved):
thanks!

PR #3726, opened 2020-10-01
change permissions on ref case rpointer files

### LINKED ISSUE #3707: Permissions should be changed after rpointer files are copied so that write for author is allowed.
We'd like to have files in inputdata to be read-only, so that someone can't accidentally end up changing or removing them. However, as it is now, cime relies on having the rpointer files to have write for author allowed. When cime copies the files for a REFCASE it doesn't change the permission, but cime requires that the author be able to overwrite them when running a case.

Thus the best way around this is when cime copies the rpointer files for a case that it also change the file permissions to open up author write.

@gold2718 @fischer-ncar @cacraigucar @fvitt

jedwards4b:
chmod of rpointer files when copied to run directory so that user has write permissions

Test suite: hand tested with SMS.f19_g17.B1850.cheyenne_intel by adding a refcase
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3707

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


billsacks (approved):
Thanks, @jedwards4b !

ekluzek (approved):

PR #3727, opened 2020-10-01
Add support for MPAS grids

fischer-ncar:
Add support for MPAS grids from @brian-eaton.

                Y_TestUserConcurrentMods failed, but it's also failing for cime master.
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/Tools/Makefile
+++ b/scripts/Tools/Makefile
@@ -568,14 +568,17 @@ ifeq ($(findstring -cosp,$(CAM_CONFIG_OPTS)),-cosp)
 endif
 
 COMP_ATM ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) COMP_ATM --value)
+CAM_DYCORE ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) CAM_DYCORE --value)
 ifeq ($(strip $(COMP_ATM)),cam)
-  CAM_DYCORE ?= $(shell $(CASEROOT)/xmlquery --caseroot $(CASEROOT) CAM_DYCORE --value)
   ifeq ($(CAM_DYCORE),fv3)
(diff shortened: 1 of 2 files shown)


jedwards4b:
Just want to acknowledge that Y_TestUserConcurrentMods  is failing on master.

scripts/Tools/Makefile
```diff
@@ -604,6 +607,29 @@ $(FV3CORE_LIBDIR)/libfv3core.a: $(EXEROOT)/FMS/libfms.a
 
 dyn_grid.o: $(FV3CORE_LIBDIR)/libfv3core.a
 endif
+
+ifdef MPAS_LIBDIR
```
jedwards4b, line 611: Could this be done in the cam buildlib and libmpas.a combined into libatm.a so that no changes are needed in the cime Makefile?
The following will combine liba.a and libb.a into libab.a
ar crsT libab.a liba.a libb.a

fischer-ncar: I'll let @brian-eaton answer this.

brian-eaton: We were following the example of the COSP and FV3 library builds.  Not sure how this is different.

I've looked in buildlib and don't see how it would be done from there.  There are interdependencies between the mpas lib and the cam build which are expressed in the Makefile.  libmpas depends on cam_abortutils.o, and dyn_comp.o depends on libmpas.  So building libmpas is synchronized between the builds of cam_abortutils.o and dyn_comp.o.  Note that this is also true for the COSP library build: cospsimulatro_intr.o depends on libcosp which depends on cam_abortutils.o.

jedwards4b (approved):
Thank you for considering.

PR #3730, opened 2020-10-06
Fix check_input_data to fail over to svn when a wget download fails

lukaszlacinski:
That commit, https://github.com/ESMCI/cime/commit/897f79b94b2706d3e2c41719ae5b3aee2f8f3264, causes that check_input_data(download=True) returns False when the last wget downloads fails, and returns True when the last wget download succeeds even if all other wget downloads have failed. This PR fixes this bug.


jedwards4b (changes requested):
This change leaves no case in which variable no_files_missing is ever True.

lukaszlacinski:
> This change leaves no case in which variable no_files_missing is ever True.

no_files_missing is initialized as True, https://github.com/ESMCI/cime/blob/60d324eb2a2abdb4faf4ef0dfe8f3d3a0c70c9c7/scripts/lib/CIME/case/check_input_data.py#L288.

jedwards4b (approved):
Thanks

PR #3734, opened 2020-10-09
Some updates to E3SM test infrastructure

jgfouca:
1) When recording walltimes, include commit for provenance. This data can now be more useful than simply creating more-accurate walltime estimates; for example, we could potentially create time plots of performance.
2) Be more aggressive about uploading logs: include logs from all potential sources, even when the only problem is NML diff.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @rljacob @amametjanov

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -3,7 +3,7 @@ Base class for CIME system tests
 """
 from CIME.XML.standard_module_setup import *
 from CIME.XML.env_run import EnvRun
-from CIME.utils import append_testlog, get_model, safe_copy, get_timestamp, CIMEError, expect
+from CIME.utils import append_testlog, get_model, safe_copy, get_timestamp, CIMEError, expect, get_current_commit
 from CIME.test_status import *
 from CIME.hist_utils import copy_histfiles, compare_test, generate_teststatus, \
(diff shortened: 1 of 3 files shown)


rljacob:
What does uploading more logs for an NML diff do?

jedwards4b (approved):

jgfouca:
@rljacob , the logic prior to this PR was to avoid uploading any logs if the only problem with a test was that it had a NML diff. That logic is now changed, logs will be uploaded for NML diffs. Also, the prior logic limited log scans based on test outcome (build fail -> only look in EXEROOT for logs). Now, we just gather all the logs from all potential dirs containing logs regardless.

rljacob:
Ok but uploading more logs won't necessarily help with identifying NML diffs via cdash.   The problem there is the column that lists the NML diffs doesn't provide a way to click to the TestStatus log.   I'd rather do that then download a tar file  to find the NML diffs.

rljacob (approved):

jgfouca:
@rljacob , yeah, there's possibly more we can do to help with NML diffs. The uploading of logs for NMLs was a feature request on slack a couple weeks ago, I think by @amametjanov

amametjanov:
@bishtgautam ^^

amametjanov (approved):

PR #3747, opened 2020-10-16
Esmf and pgi cheyenne update

jedwards4b:
Updating the pgi compiler version and esmf for pgi and gnu on cheyenne, add a workaround for pgi include path length

Test suite: scripts_regression_tests.py on cheyenne with pgi and gnu and openmpi and mct
Test baseline: 
Test namelist changes: 
Test status: bit for bit,

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -501,7 +501,7 @@ This allows using a different mpirun command to launch unit tests
 	<command name="load">openblas/0.3.6</command>
       </modules>
       <modules compiler="pgi">
-	<command name="load">pgi/19.3</command>
+	<command name="load">pgi/19.9</command>
       </modules>
 
       <modules compiler="intel" mpilib="!mpi-serial" DEBUG="TRUE">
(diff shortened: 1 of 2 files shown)


jedwards4b:
@billsacks I added you because you wanted to be aware of compiler version changes.

fischer-ncar (approved):

billsacks (approved):
Thanks for bringing this to my attention, @jedwards4b . I'm less concerned than I often am about compiler updates, since it looks like this just updates the pgi version on cheyenne (and adds some esmf modules).

billsacks:
> I'm less concerned than I often am about compiler updates, since it looks like this just updates the pgi version on cheyenne (and adds some esmf modules).

However, that said: I'm thinking we should wait to bring this in until after #3732 , so that cesm2_3_beta01 won't have this compiler change on top of the pio2 change. @jedwards4b @fischer-ncar @mvertens do you agree?

fischer-ncar:
Since it's just pgi, I think it would be fine to change it at any time.

billsacks:
> Since it's just pgi, I think it would be fine to change it at any time.

But it looks like we do have a number of pgi tests in the prealpha / prebeta test suites, and I thought we wanted to ensure that cesm2_3_beta01 is bit-for-bit... or are we pretty sure that this pgi update _is_ bit-for-bit?

billsacks:
(I'll add, though, that I don't feel _that_ strongly about ensuring that `cheyenne_pgi` tests are bit-for-bit in cesm2_3_beta01, so I'm okay with folding this in if you are.)

PR #3748, opened 2020-10-19
Create srt.yml

jedwards4b:
Change CI testing from travis to github, include a greater percentage of scripts_regression_tests.py
in testing.  I have added a new option to scripts_regression_tests.py --no-fortran-run so that all tests except those that
run fortran programs (or are considered slow) are included.  Remove travis-ci tests this replaces it. 

Test suite:  github actions 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?:  adds --no-run option to scripts_regression_tests.py 

Update gh-pages html (Y/N)?:

Code review:


scripts/lib/CIME/case/case.py
```diff
@@ -980,10 +980,11 @@ def configure(self, compset_name, grid_name, machine_name=None,
 
         for nodename in nodenames:
             value = machobj.get_value(nodename, resolved=False)
-            type_str = self.get_type_info(nodename)
-            if type_str is not None:
-                logger.debug("machine nodname {} value {}".format(nodename, value))
-                self.set_value(nodename, convert_to_type(value, type_str, nodename))
+            if value:
```
jedwards4b, line 983: Bug fix here - when --machine is specified on command line not having a nodename_regex field should not be an error.

scripts/tests/scripts_regression_tests.py
```diff
@@ -3537,6 +3553,10 @@ def _main_func(description):
                         help="Do not submit jobs to batch system, run locally."
                         " If false, will default to machine setting.")
 
+    parser.add_argument("--no-run", action="store_true",
+                        help="Do not run any fortran jobs."
+                        " used for github actions")
```
billsacks, line 3558: The name of this argument (and the related variables in the code) makes it sound more general than it is (based on the help description). Would something like `--no-compiled-tests` or `--no-fortran-tests` or `--python-only` (or something similar) be more appropriate? Also, I'm unclear how this relates to the `--fast` argument; can you clarify that? Also, I'm confused: if you aren't running any fortran jobs, then why is there information in the yaml file that seems to be needed for compiling fortran?

billsacks (changes requested):
Can you briefly describe what this does in the overall PR comment? My impression is that this is adding a github action that will trigger on pushes to the master and maint-5.6 branch that runs some subset of scripts_regression_tests (though I'm not sure exactly what subset)?

See also my one line comment below.

scripts/tests/scripts_regression_tests.py
jedwards4b: We are compiling fortran jobs, but I encountered problems when I attempted to run them.  Compiling fortran allows us to include several tests that compile but do not run fortran codes.

billsacks: Okay, thanks. I'm still not totally clear on this, though. Does this option imply that the system tests (class Z, if I remember correctly) are all built but not run, or does this also skip those system tests?

billsacks: (If possible, can you please answer that last question in the help description for this argument?)

jedwards4b: I have updated the --no-fortran-run description.   @billsacks Perhaps for your own edification you should review the 
tests that are run by selecting the "Checks" option at the top of this PR.

billsacks: Thanks, @jedwards4b . I don't see an updated description; maybe you haven't pushed it back yet? I do see, based on the checks and a second look at the code, that `--no-fortran-run` implies `--fast`, so simply noting that in the description would be sufficient (if you haven't already done so).

jedwards4b: oops - pushed now.

billsacks (approved):
Thanks!

jgfouca (approved):

PR #3755, opened 2020-10-26
Fix E3SM nightly TSC test failure

mkstratos:
In c66e1c9, casename was added as an argument to
`ArchiveBase.get_all_histfiles()`. This function is called from
`TSC._run_with_specified_dtime()`, and causes the error here:
https://my.cdash.org/test/10724501

User interface changes?: N

Code review:


jgfouca:
Nice job, @mkstratos !

jgfouca (approved):

PR #3756, opened 2020-10-27
Some build system updates for e3sm

jgfouca:
1) Add new --separate-builds flag for case.build. Component builds
   will be done one at a time and produce separate log files for each
   component.
2) Remove --use-old flag. The old system is not being maintained or tested
   and it's no longer safe to use.
3) Produce more logging output for case.build.
4) Re-establish support for building only specific components with case.build.
   This capability had been lost in the initial cmake conversion. Building components
   this way will cause them to get their own log file.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes E3SM 3769 and 3758

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/Tools/case.build
+++ b/scripts/Tools/case.build
@@ -63,13 +63,13 @@ def parse_command_line(args, description):
                         "Default is current directory.")
 
     if get_model() == "e3sm":
-        parser.add_argument("--use-old", action="store_true",
-                            help="Use old Makefile build system (not cmake)")
-
         parser.add_argument("--ninja", action="store_true",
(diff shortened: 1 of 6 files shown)


rljacob:
What is the difference between 1. and 4. ?

jgfouca:
@rljacob ,

To build all components separately, with separate logs:
```
./case.build --separate-builds
```

To build a specific set of components (each getting its own log) (atm and lnd in this example):
```
./case.build -b atm -b lnd
```

rljacob (approved):

PR #3760, opened 2020-10-29
Add an error check for separate builds

jgfouca:
It won't work for test cases.

Test suite: code-checker
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/Tools/case.build
+++ b/scripts/Tools/case.build
@@ -151,7 +151,10 @@ def _main_func(description):
                     ts.set_status(phase_to_fail, TEST_FAIL_STATUS, comments="failed to initialize")
                 raise
 
-            expect(buildlist is None, "Build lists don't work with tests")
+            expect(buildlist is None,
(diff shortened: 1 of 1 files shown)


rljacob:
What is this?  You can't do separate builds if you started with create_test ?

jgfouca:
@rljacob , that's right. There's actually a fair number of features of case.build that are not available when using create_test because we'd have to pass all those parameters into the system_test infrastructure. Building individual models is the other feature that's explicitly checked, but I strongly suspect other features are also not supported. Basically, any flag you see being passed to build.case_build but not test.build is not going to work for create_test.

That said, if you think that --separate-builds is important enough to be supported for create_test, I can make the necessary changes for that.

jedwards4b:
It makes sense that you can't call create_test with the --separate-builds flag - but once a test is created can you not use this option when running case.build on an existing test?

PR #3765, opened 2020-11-03
Fix python3 incompatibilities in case.py

johnsonb-ucar:
[ Explicit type conversion must be used when comparing ninst because in python3
'>' is not supported between instances of 'str' and 'int'.]

Test baseline: NA
Test namelist changes: NA

Fixes [CIME Github issue #] #3766 

User interface changes?: N

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -865,7 +865,7 @@ class Case(object):
 
         if other is not None:
             logger.info("setting additional fields from config_pes: {}".format(other))
-            for key, value in other.items():
+            for key, value in list(other.items()):
                 self.set_value(key, value)
 
         totaltasks = []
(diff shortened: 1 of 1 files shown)


jedwards4b:
We regularly test against python 3 - in what version of python did this fail?  Please open an issue and refer to it in the template above, also run scripts_regression_tests.py in cime/scripts/tests

johnsonb-ucar:
This fails in Python 3.7.3. Here is the traceback:
```
  File "/lustre/project/k1421/cesm2_1_3/cime/scripts/create_newcase", line 218, in <module>
    _main_func(__doc__)
  File "/lustre/project/k1421/cesm2_1_3/cime/scripts/create_newcase", line 213, in _main_func
    input_dir=input_dir, driver=driver, workflowid=workflow)
  File "/lustre/project/k1421/cesm2_1_3/cime/scripts/Tools/../../scripts/lib/CIME/case/case.py", line 1448, in create
    input_dir=input_dir, driver=driver, workflowid=workflowid)
  File "/lustre/project/k1421/cesm2_1_3/cime/scripts/Tools/../../scripts/lib/CIME/case/case.py", line 858, in configure
    if multi_driver and ninst>1:
TypeError: '>' not supported between instances of 'str' and 'int'
```
The issue might be that the testing suite does not attempt to create a multi-instance case.

jedwards4b (approved):

PR #3770, opened 2020-11-04
Avoid building CDEPS with CTSM's LILAC

### LINKED ISSUE #3769: Avoid building CDEPS with CTSM's LILAC
For now, we want to avoid building CDEPS with CTSM's LILAC because it's not needed in this configuration and CDEPS relies on unreleased ESMF code. I will submit a PR that avoids building CDEPS when using LILAC, similarly to what's done for ufs.

billsacks:
For now, avoid building CDEPS with CTSM's LILAC because it's not needed
in this configuration and CDEPS relies on unreleased ESMF
code. Eventually we will require CDEPS (for its streams functionality),
at which point CDEPS should only require released ESMF code; then we
should revert the changes in this commit.

Test suite: **Running testing now; will comment when complete**
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3769 

User interface changes?: N

Code review: @jedwards4b

### DIFF
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -370,6 +370,7 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
             os.makedirs(shared_item)
 
     mpilib = case.get_value("MPILIB")
+    lilac_mode = case.get_value("LILAC_MODE")
     ufs_driver = os.environ.get("UFS_DRIVER")
     cpl_in_complist = False
     for l in complist:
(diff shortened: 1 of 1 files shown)


billsacks:
@jedwards4b I'm about to run scripts_regression_tests and at least some testing that exercises cmeps / cdeps. I'll add a comment when that testing is done. In the meantime, I wanted to put this out for your review.

scripts/lib/CIME/build.py
```diff
@@ -390,7 +391,13 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
 
     # Build shared code of CDEPS nuopc data models
     build_script = {}
-    if comp_interface == "nuopc" and (not ufs_driver or ufs_driver != 'nems'):
+    if (comp_interface == "nuopc" and (not ufs_driver or ufs_driver != 'nems')
+        and lilac_mode != 'on'):
```
billsacks, line 395: Note that LILAC_MODE is undefined when not using CTSM, but that's okay: lilac_mode will end up as None, which gives equivalent results to 'off' (i.e., not 'on').

jedwards4b (approved):

billsacks:
Thanks a lot for the quick turnaround of this, @jedwards4b .

For the record, I have finished scripts_regression_tests on cheyenne (pass) and, in a CESM checkout, `SMS_Vnuopc_Ld5.T62_g17.2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_SROF_SGLC_SWAV_SESP.cheyenne_intel` (passes).

PR #3771, opened 2020-11-05
WIP Add timeout values to avoid forever-hangs

### LINKED ISSUE #3370: We should try to avoid forever-hangs wherever possible
1. Add an optional timeout argument to run_cmd and run_cmd_no_fail. Implementation should use the Timeout context manager in utils.py.
2. Survey CIME's use of run_cmd/run_cmd_no_fail, looking for operations (network, large-file) that could hang and add a generous timeout.

WesCoomber:
I added a generous timeout (LONG_OP_TIMEOUT
is 4500 seconds or 1 hour and 15 minutes.)
for operations that could hang, such as
network operations, and compress/uncompress
operations. I also added a shorter timeout value,
SHORT_OP_TIMEOUT, of 60 seconds for
quick 'does file exist' network operations.
These timeout values are global constants 
located in utils.py.

Test baseline: 
Test namelist changes: 
Test status: [bit for bit]

Fixes #3370 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jgfouca


jedwards4b (approved):

jgfouca (changes requested):
Instead of repeating the same hardcoded values in multiple places, let's instead use some global constants. These can be kept in utils.py. We'd want two values, SHORT_OP_TIMEOUT, LONG_OP_TIMEOUT. Also, I'm a bit concerned that LONG_OP_TIMEOUT is too short. I think we want at least an hour, maybe more.

scripts/tests/scripts_regression_tests.py
```diff
@@ -249,7 +249,7 @@ def assert_dashboard_has_build(tester, build_name, expected_count=1):
 
         wget_file = tempfile.mktemp()
 
-        run_cmd_no_fail("wget https://my.cdash.org/api/v1/index.php?project=ACME_test --no-check-certificate -O %s" % wget_file)
+        run_cmd_no_fail("wget https://my.cdash.org/api/v1/index.php?project=ACME_test --no-check-certificate -O %s" % wget_file, timeout=900)
```
jgfouca, line 252: This one can use SHORT_OP_TIMEOUT.

WesCoomber: @jgfouca Okay I improved the wget timeout in scripts_regression_tests to use the SHORT_OP_TIMEOUT global constant.

WesCoomber:
@jgfouca Yes that makes sense. I added LONG_OP_TIMEOUT (which is 
is 4500 seconds or 1 hour and 15 minutes) for operations that could hang, such as
network operations, and compress/uncompress operations. I also added a shorter timeout value, SHORT_OP_TIMEOUT, of 60 seconds for quick 'does file exist' network operations.These timeout values are global constants located in utils.py.

jgfouca:
This is looking better but I really don't want to have to spend time chasing down problems in CIME caused by timeouts. Some of the input files are quite large and sometimes clusters have slow filesystems and less-than-ideal download speeds from the external web. @jedwards4b , in your judgement, what would be a safe timeout for downloading input files?

jedwards4b:
Most users would ctrl-c before waiting an hour.  I think that maybe indicating in stdout that the
process is in a call with a timeout setting and maybe having some kind of progress indicator print out may be a good complement to these.

WesCoomber:
There were issues with the git timeline and git history, so I did some work on a new branch to get back to a working state that can be more easily merged with master. 

Github does not have the feature to change target branches in an open pull request, so I created a new PR #3887 which references this first pull request and preserves the conversation history.

PR #3777, opened 2020-11-13
build nuopc_cap with csm_share

jedwards4b:
Build the nuopc_cap code with csm_share instead of in components (it's currently built anew for each component)
This means that the code source can be added to SourceMods/src.csm_share - currently it would need to be added
to each components SourceMods which creates a possibility of conflict

Test suite: scripts_regression_tests.py with CIME_DRIVER=nuopc and SMS_Vnuopc.f19_g17.B1850.cheyenne_intel
(with components build of that directory removed)

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


billsacks (approved):
Thanks a lot for this @jedwards4b !

mvertens (approved):

PR #3781, opened 2020-11-17
Update OS process id error-checking in GPTL's get_memusage

amametjanov:
On newer machines with AMD Epyc + CentOS 8, processes on
compute nodes have pids larger than 6-digit numbers. This logs
many messages of the form:
```
  59: get_memusage: pid 1031078 is too large
```
This PR updates OS process id error-checking in GPTL's get_memusage

Test suite: by-hand
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

### DIFF
--- a/src/share/timing/GPTLget_memusage.c
+++ b/src/share/timing/GPTLget_memusage.c
@@ -118,8 +118,8 @@ int GPTLget_memusage (int *size, int *rss, int *share, int *text, int *datastack
   */
 
   pid = (int) getpid ();
-  if (pid > 999999) {
-    fprintf (stderr, "get_memusage: pid %d is too large\n", pid);
+  if (pid <= 0) {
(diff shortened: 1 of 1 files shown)


src/share/timing/GPTLget_memusage.c
```diff
@@ -118,7 +118,7 @@ int GPTLget_memusage (int *size, int *rss, int *share, int *text, int *datastack
   */
 
   pid = (int) getpid ();
-  if (pid > 999999) {
+  if (pid > 9999999) {
```
sarats, line 121: Curious why this check is needed?

Looking at the man page (https://man7.org/linux/man-pages/man2/getpid.2.html), this call should always be successful at least on Linux. Perhaps check for a positive value or what's the corner case we are trying to handle?

amametjanov: Updated

PR #3782, opened 2020-11-20
More e3sm build upgrades

jgfouca:
Change list:
1) Make time gathering for cmake and make commands more robust.
2) When saving build provenance, collect and save the times needed to build targets to a "build_times.txt" file in similar approach to how build_environment.txt is saved.

Test suite: scripts_regression_tests, by-hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/provenance.py
+++ b/scripts/lib/CIME/provenance.py
@@ -34,13 +34,14 @@ def _extract_times(zipfiles, target_file):
 
     contents ="Target Build_time\n"
     for zipfile in zipfiles:
-        output = run_cmd_no_fail("zgrep 'built in' {}".format(zipfile))
-        for line in output.splitlines():
-            line = line.strip()
-            if line:
-                items = line.split()
-                target, the_time = items[1], items[-2]
-                contents += "{} {}\n".format(target, the_time)
+        stat, output, _ = run_cmd("zgrep 'built in' {}".format(zipfile))
+        if stat == 0:
+            for line in output.splitlines():
+                line = line.strip()
+                if line:
+                    items = line.split()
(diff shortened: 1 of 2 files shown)


jgfouca:
@sarats ^

scripts/lib/CIME/provenance.py
```diff
@@ -71,9 +86,18 @@ def _save_build_provenance_e3sm(case, lid):
...
     env_module.save_all_env_info(env_prov)
 
+    # Save build times
+    build_times = os.path.join(exeroot, "build_times.{}.txt".format(lid))
+    if os.path.exists(build_times):
+        os.remove(build_times)
+    globstr = "{}/*bldlog*{}.gz".format(exeroot, lid)
+    matches = glob.glob(globstr)
+    if matches:
+        _extract_times(matches, build_times)
```
sarats, line 96: I presume this means the overall component build times are written at the end of this file, right?

jgfouca: @sarats , remember that, with CMake, the entire build is done at once, not component-by-component (unless --separate-builds is provided to case.build). So we can't code our provenance to expect per-component build times.

sarats: Oh right, I was looking at Az's output that probably used separate builds.

sarats (approved):

worleyph:
@jgfouca , I will be unable to review this for 2-3 weeks. Best to remove me as a reviewer.

PR #3794, opened 2020-12-10
fix issue with time variable in ha files

jedwards4b:
Fixes an issue in cpl hist avg files that caused the time variable not to be written correctly in pio2.

Test suite: scripts_regression_tests.py, ERS_Lm3.T62_g16.AIAF.cheyenne_intel
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/drivers/mct/main/seq_hist_mod.F90
+++ b/src/drivers/mct/main/seq_hist_mod.F90
@@ -131,7 +131,6 @@ contains
        atm, lnd, ice, ocn, rof, glc, wav, iac, &
        fractions_ax, fractions_lx, fractions_ix, fractions_ox, fractions_rx,  &
        fractions_gx, fractions_wx, fractions_zx, cpl_inst_tag)
-
     implicit none
     !
     ! Arguments
@@ -172,6 +171,7 @@ contains
     type(mct_gsMap), pointer :: gsmap
     type(mct_gGrid), pointer :: dom    ! comp domain on cpl pes
     character(CL) :: model_doi_url
+
     !-------------------------------------------------------------------------------
     !
     !-------------------------------------------------------------------------------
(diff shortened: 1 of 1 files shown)


src/drivers/mct/main/seq_hist_mod.F90
```diff
@@ -248,17 +248,9 @@ subroutine seq_hist_write(infodata, EClock_d, &
...
-             call seq_io_write(hist_file,&
-                  time_units=time_units, time_cal=calendar, time_val=curr_time, &
-                  nt=1,whead=whead, wdata=wdata)
-          else
-             call seq_io_write(hist_file, &
-                  time_units=time_units, time_cal=calendar, time_val=curr_time, &
-                  nt=1,whead=whead, wdata=wdata, tbnds=tbnds)
-          endif
-
+          call seq_io_write(hist_file,&
```
jedwards4b, line 251: Remove this do nothing if block introduced by Tony "cut-n-paste" Craig.

src/drivers/mct/main/seq_hist_mod.F90
```diff
@@ -862,16 +854,18 @@ subroutine seq_hist_writeavg(infodata, EClock_d, &
              end if
 
              avg_time = 0.5_r8 * (tbnds(1) + tbnds(2))
+
              !---------- tcx nov 2011 tbnds of same values causes problems in ferret
-             if (tbnds(1) >= tbnds(2)) then
+             if (tbnds(1) == tbnds(2)) then
```
jedwards4b, line 859: This if block was backward.

billsacks: I'm not sure I'd say it was backward, but your change seems okay. My understanding is that, in general, we have either:
- tbnds(1) < tbnds(2) – the normal case
- tbnds(1) == tbnds(2) – the case this conditional is trying to avoid

I can see arguments either way for how to handle the case where tbnds(1) > tbnds(2), but I doubt it ever happens in practice.

billsacks (approved):
Thanks for your fix, @jedwards4b !

billsacks:
@jedwards4b @fischer-ncar - I have gone ahead and reviewed this in place of @mvertens so that we can get it merged so that @fischer-ncar can include it in his next round of testing, if desired.

Issue #3790, opened 2020-12-09
Attempt to fetch from allocatable variable DRIVER_TASK_NODE_MAP when it is not allocated


jedwards4b:
The code introduced in #3788 appears to be generating an error when compiled in debug mode.
I tried the following test SMS_D.T62_g16.AIAF and got error:

0:forrtl: severe (408): fort: (8): Attempt to fetch from allocatable variable DRIVER_TASK_NODE_MAP when it is not allocated
0:
0:Image              PC                Routine            Line        Source             
0:cesm.exe           000000000158DCB6  Unknown               Unknown  Unknown
0:cesm.exe           000000000042E6C7  cime_comp_mod_mp_        2482  cime_comp_mod.F90
0:cesm.exe           0000000000464BCC  MAIN__                    153  cime_driver.F90
0:cesm.exe           0000000000407922  Unknown               Unknown  Unknown
0:libc.so.6          00002B97AEF726E5  __libc_start_main     Unknown  Unknown
0:cesm.exe           0000000000407829  Unknown               Unknown  Unknown

amametjanov:
I've been trying to reproduce the error. SMS_D.T62_g16.AIAF works with `CIME_MODEL=e3sm`. Tried with `CIME_MODEL=cesm` on cori-haswell, but ran into other unrelated errors.

The array is allocated in src/drivers/mct/shr/seq_comm_mct.F90, TIMING macro is [defined here](https://github.com/ESMCI/cime/blob/master/src/build_scripts/buildlib.csm_share#L125) and is not e3sm-specific. If the build above did not involve `seq_comm_mct.F90`, can you please post steps to reproduce on Cori ? Thanks.

klindsay28:
The allocation of `driver_task_node_map` in `src/drivers/mct/shr/seq_comm_mct.F90` is inside a `(info_taskmap_model > 0)` conditional, while the usage of `driver_task_node_map` in `src/drivers/mct/main/cime_comp_mod.F90` is not inside such a conditional.

The default value of `info_taskmap_model` in `src/drivers/mct/cime_config/config_component_e3sm.xml` is `2`, while the default in `src/drivers/mct/cime_config/config_component_cesm.xml` is `0`. Perhaps you can reproduce the error if you set `info_taskmap_model=0`.

amametjanov:
[...]

PR #3801, opened 2020-12-15
Add memory usage logging

amametjanov:
This logs GPTL memory usage and highwater numbers in a gnuplot format
for debugging and profiling of memory usage. Logging can be configured
for component ROOTPE tasks, all tasks, component ROOTPE nodes (default)
and all nodes. The frequency of logging can be configured in seconds of
simulation time: by default once-a-day.

The output is captured in 
`$RUNDIR/memory.$info_mprof.$info_mprof_dt.log file` in comma-separated
format with one line per snapshot
```
yymmdd.fraction, msize_0, mrss_0, msize_1, mrss_1, ...
```

Test suite: SMS[_D][_P18x2].T62_g16.AIAF
Test baseline: none
Test namelist changes: cesm drv_in's info_taskmap_model changes from 0 to 1, new info_mprof and info_mprof_dt
Test status: [bit for bit]

Fixes [ESMCI/cime#3790]

User interface changes?: none

Code review: Jim Edwards.

### DIFF
--- a/src/drivers/mct/cime_config/buildnml
+++ b/src/drivers/mct/cime_config/buildnml
@@ -171,6 +171,22 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
         nmlgen.set_value('esp_cpl_dt', value=esp_time)
     # End if pause is active
 
+    # Sanity checks on mem-prof options
+    info_mprof = case.get_value('INFO_MPROF')
+    expect(info_mprof in [0,1,2,3,4],
+           "Expected INFO_MPROF value is 0,1,2,3 or 4; given {:d}".format(info_mprof))
+
+    info_taskmap_model = case.get_value('INFO_TASKMAP_MODEL')
+    expect(not (info_mprof > 2 and info_taskmap_model < 1),
+           "Node-level memory profiling (INFO_MPROF={:d}) expects positive INFO_TASKMAP_MODEL, given {:d}".\
+           format(info_mprof, info_taskmap_model))
+
+    info_mprof_dt = case.get_value('INFO_MPROF_DT')
+    expect(info_mprof_dt >= 0 and 
+           info_mprof_dt <= 86400,
+           "Expected INFO_MPROF_DT between 0 and 86400 secs; given {:d}".format(info_mprof_dt))
+    # end mprof checks
+
     #--------------------------------
     # (1) Write output namelist file drv_in and  input dataset list.
     #--------------------------------
--- a/src/drivers/mct/cime_config/buildnml
+++ b/src/drivers/mct/cime_config/buildnml
@@ -173,9 +173,6 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
 
     # Sanity checks on mem-prof options
     info_mprof = case.get_value('INFO_MPROF')
-    expect(info_mprof in [0,1,2,3,4],
-           "Expected INFO_MPROF value is 0,1,2,3 or 4; given {:d}".format(info_mprof))
-
     info_taskmap_model = case.get_value('INFO_TASKMAP_MODEL')
     expect(not (info_mprof > 2 and info_taskmap_model < 1),
            "Node-level memory profiling (INFO_MPROF={:d}) expects positive INFO_TASKMAP_MODEL, given {:d}".\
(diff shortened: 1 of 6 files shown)


src/drivers/mct/main/cime_comp_mod.F90
```diff
@@ -2432,21 +2458,92 @@ subroutine cime_run()
     call seq_timemgr_EClockGetData( EClock_d, curr_ymd=ymd, curr_tod=tod)
 #ifndef CPL_BYPASS
     ! Report on memory usage
+    call shr_mem_getusage(msize,mrss)
+
+    allocate( msizeOnTask(0:npes_GLOID-1), mrssOnTask(0:npes_GLOID-1), stat=ierr)
```
jedwards4b, line 2463: I'm not sure this is going to be fully scalable and suggest that there be a way to turn it off for large cases.

src/drivers/mct/shr/seq_comm_mct.F90
```diff
@@ -408,15 +415,22 @@ subroutine seq_comm_init(global_comm_in, driver_comm_in, nmlfile, drv_comm_id)
           call shr_sys_flush(logunit)
        endif
 
+       allocate( driver_task_node_map(0:global_numpes-1), stat=ierr)
```
jedwards4b, line 418: This could be a problem when global_numpes is large.

jedwards4b: The problem of info_taskmap_model == 0 has not been solved.

jedwards4b (changes requested):

src/drivers/mct/main/cime_comp_mod.F90
```diff
@@ -2432,21 +2458,111 @@ subroutine cime_run()
...
+       msizeOnTask(:) = 0
+       mrssOnTask(:)  = 0
+       call mpi_gather (msize, 1, mpi_real8, &
+                        msizeOnTask, 1, mpi_real8, &
+                        0, mpicom_GLOID, ierr)
+       call mpi_gather (mrss, 1, mpi_real8, &
+                        mrssOnTask, 1, mpi_real8, &
+                        0, mpicom_GLOID, ierr)
+
+       if (info_mprof > 2) then ! aggregate task-level to node-level mem-usage
```
jedwards4b, line 2493: This check should be done in python not here.

src/drivers/mct/cime_config/config_component_cesm.xml
```diff
@@ -52,7 +52,7 @@
   <entry id="INFO_TASKMAP_MODEL">
     <type>integer</type>
     <valid_values>0,1,2</valid_values>
-    <default_value>0</default_value>
+    <default_value>1</default_value>
```
jedwards4b, line 55: Is there really a need for another variable here?  Why not just base this on info_mprof > 2?  That seems to be the requirement below.

jedwards4b (changes requested):

src/drivers/mct/cime_config/buildnml
```diff
@@ -171,6 +171,22 @@ def _create_drv_namelists(case, infile, confdir, nmlgen, files):
         nmlgen.set_value('esp_cpl_dt', value=esp_time)
     # End if pause is active
 
+    # Sanity checks on mem-prof options
+    info_mprof = case.get_value('INFO_MPROF')
+    expect(info_mprof in [0,1,2,3,4],
```
jedwards4b, line 176: There is no need for this check, the <valid_values> field from xml is already checked when you try to change a value using xmlchange.

src/drivers/mct/cime_config/config_component_cesm.xml
amametjanov: `*TASKMAP*` and `*MPROF*` serve different uses: first one is for checking task-to-node allocations, second one is for profiling memory growth and out-of-memory errors in high-resolution cases. Per-task memory profiling (info_mprof={1,2}) provides relative and finer granularity for tracking component ROOTPE and PIO_ROOT vs. other tasks. Task-to-node mapping can be reimplemented inside mem-prof to make the two independent.

jedwards4b (changes requested):
Almost there

jedwards4b (approved):
thanks

PR #3809, opened 2020-12-17
Remove unused logic

jgfouca:
E3SM no longer uses this Makefile to build components, so logic
that is only for options used by e3sm component builds can be
removed.

There might be more we can remove but I'm quite sure we can at least remove this much.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:  @jedwards4b @billsacks

### DIFF
--- a/scripts/Tools/Makefile
+++ b/scripts/Tools/Makefile
@@ -59,17 +59,6 @@ else
    ESMFDIR = noesmf
 endif
 
-# Determine whether any C++ code will be included in the build;
-# currently, C++ code is included if and only if we're linking to the
-# trilinos library or the Albany library.
-USE_CXX = FALSE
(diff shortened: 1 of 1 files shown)


billsacks (approved):
Thanks for doing this, @jgfouca .

@jedwards4b - much of the reason I didn't take this on myself a few weeks ago was that it now leaves a bunch of non-functional C++-related variables in our config_compilers.xml, and I wasn't sure what to do about that (SUPPORTS_CXX and a bunch of other CXX variables). Should we set SUPPORTS_CXX to FALSE across the board and then remove the other CXX-related variables from our config_compilers to avoid giving the impression that we support CXX code?

jedwards4b (approved):

PR #3820, opened 2021-01-04
Pass integer and not floating point

adityakavalur:
Even though the calculation for cpus-per-task (in slurm) results in an integer, it passes it as a floating point i.e. 2.0 instead of 2. I was encountering issue on Cori (NERSC)

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -183,7 +183,8 @@ class Case(object):
 
         os.environ["OMP_NUM_THREADS"] = str(self.thread_count)
 
-        self.srun_binding = smt_factor*max_mpitasks_per_node / self.tasks_per_node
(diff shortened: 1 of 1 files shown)


jedwards4b (changes requested):
Thank you for the contribution.  I believe that the prefered way to do this is
        self.srun_binding = (smt_factor*max_mpitasks_per_node) // self.tasks_per_node

adityakavalur:
Thanks for the quick response. 
While '//' does do the floor calculation it still returns the value as floating point e.g. 10.0 // 5.0 = 2.0 and 10.0 // 3.0 = 3.0 in python3. This still causes issues with Slurm, let me know if anything is unclear.

jedwards4b:
In that case, I will accept the change.  Thanks again.

jedwards4b (approved):

PR #3823, opened 2021-01-13
More grids for mizuRoute

### LINKED ISSUE #3608: Add r05mz grid for half degree grid with land only for mizuRoute
For mizuRoute I need a way to distinguish between the full half degree MOSART grid that includes ocean and the mizuRoute grid that is land only. mizuRoute currently can't handle one with ocean, but that's the default for MOSART. Right now we have to get around this by hand, but doing that became a problem. So it'll obviously be a problem for users who don't understand this subtlety. By adding another grid, MOSART won't be able to use it, and mizuRoute can be configured to only use the land-only version -- hence handling this problem without the user needing to know about it.

@mvertens @jedwards4b what is your opinion on a name for this?

### LINKED ISSUE #3758: New grids needed for mizuRoute
We need to add some new grids for use by mizuRoute: HDMAmz over CONUS, USGSmz over CONUS for nldas2 grid for land.

### LINKED ISSUE #3778: mesh file needed for nuopc driver for nldas2 grid
nuopc mesh files are only provided for a subset of the grids. Specifically the CTSM regional grids 5x5pt_amazon are missing as well as its hcru_hcru grid and the nldas2 grid.

We really want to at least add the nldas2 grid for work with mizuRoute.

The single point mesh files are missing also. Apparently the single point grids, 5x5_amazon, and hcru are no longer needed through some magic in CMEPS where it uses the domain file to create the mesh signaled with "create_mesh" to define the mesh file in config_grids.xml. For some reason the nldas2 grid doesn't work unless the mesh file is explicitly pointed to.

### LINKED ISSUE #3793: Add MERIT-Hmz grid for mizuRoute
We'd like to add another grid for mizuRoute for the MERIT-Hydro grid. This global grid is even more complex than HDMA and we aren't successful at using ESMF regridding to to. do the mapping. However, we have our own offline way to create a mapping file in a different format than SCRIP map format. We plan to figure out how to convert the file format and then use it for the mapping for this grid. Once we can do that we can then add the appropriate mesh and mapping files in place.

### LINKED ISSUE #3957: Add a grid alias for hcru_hcru_mt13
We found issues with the domain file for the hcru_hcru resolution. It turns out it calls large lakes as ocean, so the land model doesn't run over large lakes. See this issue in CTSM

https://github.com/ESCOMP/CTSM/issues/1357

So I think we should remove the old hcru_hcru grid alias and add in hcru_hcru_mt13 with a tenth degree ocean mask that includes the Caspian Sea.

I've created a domain file for this use here:

$DIN_LOC_ROOT/share/domains/domain.clm/domain.lnd.360x720_tx0.1v3.210506.nc

### LINKED ISSUE #4087: 360x720cru mesh file is missing
A mesh file for nuopc is missing for the 360x720cru grid. We've decided this is important so it needs to be added in. It looks like there is one created here:

/glade/p/cesmdata/cseg/inputdata/share/meshes/360x720_120830_ESMFmesh_c20210507b.nc

The entry in config_grids.xml for 360x720cru is like this

``` xml
    <domain name="360x720cru">
      <nx>720</nx> <ny>360</ny>
      <file grid="atm|lnd">$DIN_LOC_ROOT/share/domains/domain.clm/domain.lnd.360x720_cruncep.100429.nc</file>
      <desc>Exact half-degree CRUNCEP datm forcing grid with CRUNCEP land-mask -- only valid for DATM/CLM compset</desc>
    </domain>
```

An entry for "mesh driver=nuopc" needs to be added to the listing.

### LINKED ISSUE #4089: tx0.1v3 mesh file is missing needed for running grids with t13 for NUOPC
The mesh file is missing from config_grids.xml for tx0.1v3. This means that any grids that use t13 fail when you try to run the case.

The nuopc.runconfig file is missing a setting for mesh_mask which is set to UNSET.

cesm.log returns with a bunch of errors like this...

```
29:NetCDF Error: UNSET : No such file or directory
30:NetCDF Error: UNSET : No such file or directory
3:MPT ERROR: Rank 3(g:3) is aborting with error code 1.
3:	Process ID: 72703, Host: r11i1n29, Program: /glade/scratch/erik/SMS_Vnuopc_Ld1.hcru_hcru_mt13.I2000Clm50SpMizGs.cheyenne_intel.mizuroute-default.20210909_143420_zcv0b5/bld/cesm.exe
3:	MPT Version: HPE MPT 2.22  03/31/20 15:59:10
3:
3:MPT: --------stack traceback-------
-1:MPT ERROR: MPI_COMM_WORLD rank 17 has terminated without calling MPI_Finalize()
-1:	aborting job
```

ekluzek:
More grids for mizuRoute.

Test suite: manual tests of specific cases
Test baseline: None
Test namelist changes: none
Test status: bit-for-bit

Fixes #3793
Fixes #3778 
Fixes #3758
Fixes #3608
Fixes #3957
Fixes #4087
Fixes #4089

User interface changes?: New grids available
  5x5_amazon_rHDMA
  nldas2_rUSGS_mnldas2
  nldas2_rHDMA_mnldas2
  nldas2_rMERIT_mnldas2
  f19_g17_rHDMA
  f19_g17_rMERIT
 hcru_hcru_mt13
 hcru_hcru_mt13_rMERIT
 hcru_hcru_mt13_rHDMA


ekluzek:
I added this in for visibility. I need to update this to a later version. And there are some other changes that I will get in before asking it to be merged. So I'll convert it to a draft.

ekluzek:
Should also add: f19_g17_rMERIT and f09_g17_rMERIT and f09_g17_rHDMA

mvertens:
@erik - sorry for the delay in my review. What is the current state of this PR?

ekluzek:
We've also decided we need to add in the half degree grid hcru_hcru. So that will be added in as well. To be fully usable that also relates to this CTSM issue: https://github.com/ESCOMP/CTSM/issues/1483

ekluzek:
@alperaltuntas and @mnlevy1981 there are some changes to create_ESMF_map.sh in this PR, could one or both of you take a look and give me some feedback on the changes?

ekluzek:
Note #3957 relates to the half degree issue here.

ekluzek:
Also note the new issue #4087 relates to this as well.

tools/mapping/gen_mapping_files/gen_ESMF_mapping_file/create_ESMF_map.sh
```diff
@@ -217,10 +217,7 @@ if [ $MACH == "UNSET" ]; then
     r+([0-9])i+([0-9])n+([0-9]) )
       MACH="cheyenne"
     ;;
-    geyser* )
-      MACH="dav"
-    ;;
-    caldera* )
+    casper* )
```
mnlevy1981, line 220: Good catch! (And just in time for derecho :) )

tools/mapping/gen_mapping_files/gen_ESMF_mapping_file/create_ESMF_map.sh
```diff
@@ -314,22 +317,24 @@ case $MACH in
...
     module purge
-    module load intel/17.0.1 esmflibs/7.1.0r
+    module load intel/19.1.1 esmflibs/${esmfvers}
     if [ "$serial" == "TRUE" ]; then
       # No MPIEXEC
       if [ -z "$MPIEXEC" ]; then
         MPIEXEC=""
       fi
-      module load esmf-7.1.0r-ncdfio-uni-O
+      module load esmf-${esmfvers}-ncdfio-uni-O
```
mnlevy1981, line 337: so you're using 8.0.0 on casper but 8.1.0b23 on cheyenne?

tools/mapping/gen_mapping_files/gen_ESMF_mapping_file/create_ESMF_map.sh
```diff
@@ -352,6 +357,7 @@ if [ ! -z $ESMFBIN_PATH ]; then
 else
   ESMF_REGRID="ESMF_RegridWeightGen"
 fi
+ESMF_REGRID="$ESMF_REGRID --ignore_degenerate "
```
mnlevy1981, line 360: I think this argument should be hard-coded in line 403

```diff
-cmd="$MPIEXEC $ESMF_REGRID --ignore_unmapped -m $mapping -w $fmap -s $fsrc -d $fdst $pass_thru"
+cmd="$MPIEXEC $ESMF_REGRID --ignore_degenerate --ignore_unmapped -m $mapping -w $fmap -s $fsrc -d $fdst $pass_thru"
```

We definitely don't need it before the `command -v` call in the next block (which just makes sure the executable exists)

mnlevy1981 (changes requested):
This looks good. There are a few small changes and a couple questions I have about module versions, but it doesn't look too intrusive (and I do appreciate you stripping out support for machines that no longer exist)

tools/mapping/gen_mapping_files/gen_ESMF_mapping_file/create_ESMF_map.sh
ekluzek: Yes. I looked for the system installed version on casper, but we use a custom installed version on cheyenne. I don't know if we can use the same version on both?

tools/mapping/gen_mapping_files/gen_ESMF_mapping_file/create_ESMF_map.sh
ekluzek: Yeah, put it with the other hardcoded settings. That makes sense.

I think the flag is something that we want to always use. The ESMF people recommended it for particular work I was doing, but I don't see a reason not to have it on all the time.

ekluzek: This is the comment about --ignore_degenerate in ESMF:

>  --ignore_degenerate - ignore degenerate cells in the input grids. If not specif
>  ied,
>                            the default is to stop with an error.

ekluzek: @mnlevy1981 I pushed some changes, let me know if that's better now. I didn't test, but it's straightforward enough that it shouldn't be a problem. And I will at least run a test for it.

ekluzek: I did run a simple test. I give details on it in regular discussion.

mnlevy1981 (approved):
`create_ESMF_map.sh` looks good to me, thanks for cleaning this up @ekluzek!

ekluzek:
@fischer-ncar could you bring this to cime main-dev? Thanks..

PR #3824, opened 2021-01-13
Better create_test/wait_for_tests integration

jgfouca:
This started out of frustration that I could not bisect a DIFF for E3SM because my bisection kept tripping over NML diffs that I wasn't interested in. create_test did support a `--wait-ignore-namelists` option, but it only did anything on batch systems and I was on a desktop. The basic problem is that create_test (test_scheduler.py) was doing it's own reporting in one case and relying on wait_for_test in another. This PR changes that so it always goes through wait_for_tests. In order to have wait_for_tests has as-good-of reports as create_tests, test_status.get_overall_test_status had to be augmented to not only return the overall status for a case but to also return the phase responsible for that status.

This PR also brings cime_bisect into better working order and it can now bisect on both CIME and the host model.

This is the biggest change to the core of CIME I've done in a while, so I've asked for a couple reviewers.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @jedwards4b , @billsacks

### DIFF
--- a/scripts/Tools/cime_bisect
+++ b/scripts/Tools/cime_bisect
@@ -4,6 +4,9 @@
 A script to help track down the commit that caused tests to fail. This script
 can do bisections for both cime and the model that houses it, just be sure
 you run this script from the root of the repo you want to bisect.
+
+NOTE: this tool will only work for models that use git and, for bisecting CIME,
+bring in CIME via submodule or clone.
 """
 
 from standard_script_setup import *
--- a/scripts/Tools/cime_bisect
+++ b/scripts/Tools/cime_bisect
@@ -1,7 +1,9 @@
 #!/usr/bin/env python
 
 """
-A script to help track down the commit that caused tests to fail.
+A script to help track down the commit that caused tests to fail. This script
+can do bisections for both cime and the model that houses it, just be sure
+you run this script from the root of the repo you want to bisect.
 """
 
 from standard_script_setup import *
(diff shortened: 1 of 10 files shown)


jedwards4b (commented):
I don't have any concerns but this assumes that the parent repository uses only git and git subtrees and will not work with cesm.   How can we make that clear in the documentation?

billsacks (approved):

billsacks:
@fischer-ncar FYI. I'm not sure if you leverage the behavior of wait_for_tests at all.

Here's a message from @jgfouca from slack:

> i think this is a good change for CIME, but it's behavior changing
> the return code for the wait_for_tests tool will now be 0 even if some tests are still PEND as long as it is expected that it's possible for some tests to be legitimately pending (you did not tell wait_for_tests to wait until all tests completed and you are on a batch system)
> also the text reporting from create_test/wait_for_tests has been enhanced and will be slightly different (edited) 
> this had some ripple effects in scripts_regression_tests because it was parsing wait_for_tests output. i don't think anything else is doing that, but i'm not 100% sure

rljacob:
Is anyone outside of E3SM using wait_for_tests?   @jgfouca would that script still work if you started a test suite with the XML file instead of test.py ?

jgfouca:
@jedwards4b , I assume you're talking about cime_bisect? You're correct. Since cime_bisect leverages git bisect, it will only work for git repos.

@rljacob , anyone using create_test/test_scheduler is using wait_for_tests, so I assume it is widely used.

jgfouca:
@jedwards4b , I added a comment in the cime_bisect description indication that it only works for git models.

jedwards4b (approved):

fischer-ncar (approved):

PR #3836, opened 2021-01-25
add machine frontera and fix run_exe setting issue

### LINKED ISSUE #3833: default_run_exe
I have a need to overwrite default_run_exe on a system (frontera).  The reason is that performance on this system is much improved if the exe is read from a local disk instead of the shared one.  So 
 <executable>sbcast ${EXEROOT}/cesm.exe /tmp/cesm.exe; ibrun</executable>

and 
 <default_run_exe>/tmp/cesm.exe </default_run_exe>

But I don't see a way to change the run_exe for one system or a subset of systems.

jedwards4b:
Add the TACC Frontera HPC system to machine definitions.  In the process I found that overrides to the run_exe setting
were not working and fixed that issue.   Also added a test for this by modifying the FakeTest to change run_exe instead of
moving the executable. 

Test suite: scripts_regression_tests.py on frontera and cheyenne.
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3833 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


billsacks (commented):
This isn't a part of cime that I'm very familiar with. Please let me know if there are particular things you'd like me to review. Otherwise I'm happy to trust you on this.

jgfouca (approved):

jgfouca (approved):

billsacks (approved):

PR #3840, opened 2021-01-28
Fix some documentation of EXEROOT

billsacks:
I happened to notice some things in the documentation that looked wrong to me. I'd be happy to get a second set of eyes on this to make sure I didn't cause more errors than I fixed.

Test suite: none
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes none

User interface changes?: N

Update gh-pages html (Y/N)?: not yet; needs to be done

Code review: @mvertens


mvertens (approved):

billsacks:
I rebuilt the documentation. I got a ton of warnings from the documentation build, some related to the api generation (which points to some non-existent files and possibly other issues) and some related to other parts of the documentation build. This will need some attention at some point.

I realized that I only made this change on master, not on all of the branches. My inclination is to let these minor errors remain on the other documentation branches, but let me know if you feel differently.

PR #3843, opened 2021-02-03
Turn PIO_USE_MALLOC back on

jgfouca:
If this setting doesn't hurt anything, I'd like to leave it in since we need it for scorpio. If that's a problem, I can look into an indirection based on model.

Test suite: by-hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


jedwards4b (approved):
We have removed this Flag from PIO, it won't have any effect for us.

PR #3844, opened 2021-02-03
Add a new env_run XML value for E3SM: CASE_GROUP

### LINKED ISSUE #3825: Add a "case-family" argument to create_newcase
In order to complete all the simulations required for DECK, many cases will have to be run.   Each run (pi-control, 20th centruy, SSP85) will start with separate invocations of "create_newcase" with a different argument to "--case" for each simulation and that argument is recorded in the CASE variable in env_run.xml.

For constructing databases and other provenance tools, it would be useful to indicate that each case is part of something larger (in this instance, a DECK run).   The proposal is to add an optional argument like --case-family to create_newcase.  That value would be recorded in env_run.xml under CASE_FAMILY.

So you would have "create_newcase --case picontolCESM3 --case-family DECK_CESM3  --compset ....."

I'm not married to "--case-family" as the name of this argument.  "--simulation-campaign" was also suggested.

jgfouca:
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3825

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: @rljacob

### DIFF
--- a/scripts/create_newcase
+++ b/scripts/create_newcase
@@ -154,8 +154,7 @@ def parse_command_line(args, cimeroot, description):
                         "\nIf provided, the contents of these files will be appended to"
                         "\nthe standard machine files (and any files in ~/.cime).")
 
-    if model == "e3sm":
-        parser.add_argument("--case-group", help="Add this case to a case group")
+    parser.add_argument("--case-group", help="Add this case to a case group")
 
     args = CIME.utils.parse_args_and_handle_standard_logging_options(args, parser)
 
@@ -176,10 +175,6 @@ def parse_command_line(args, cimeroot, description):
     if model == "cesm":
         run_unsupported = args.run_unsupported
 
-    case_group = None
-    if model == "e3sm":
-        case_group = args.case_group
-
     expect(CIME.utils.check_name(args.case, fullpath=True),
            "Illegal case name argument provided")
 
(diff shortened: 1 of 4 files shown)


jgfouca:
@rljacob , I assume we do not want to tie this value into create_test, just create_newcase?

rljacob:
Yes just create_newcase.

scripts/create_newcase
```diff
@@ -154,6 +154,9 @@ def parse_command_line(args, cimeroot, description):
                         "\nIf provided, the contents of these files will be appended to"
                         "\nthe standard machine files (and any files in ~/.cime).")
 
+    if model == "e3sm":
```
rljacob, line 157: Don't restrict it to e3sm.  cesm users might find this useful .  As long as its not a required argument.

src/drivers/mct/cime_config/config_component_e3sm.xml
```diff
@@ -795,6 +795,15 @@
     <desc>2**n relative cost of machine (DO NOT EDIT)</desc>
   </entry>
 
+  <entry id="CASE_GROUP">
+    <type>char</type>
+    <group>run_desc</group>
+    <file>env_run.xml</file>
```
rljacob, line 801: I think it should be in env_case.xml.   You can't change it without re-running create_newcase.

jgfouca:
@rljacob , I made the changes you wanted. @jedwards4b , I added you as a review since Rob wanted this feature to be opened up to all models.

rljacob (approved):

jedwards4b (approved):

PR #3853, opened 2021-02-09
Update driver-mct support for trigrid configuration where lnd is not on the atm grid

jonbob:
Brings in changes to the mct coupler to fix both the diagnostics
and lnd2rof coupling for trigrid configurations, where atm is on
one grid, lnd and rof on another, and ocn/ice on a third. Issues
with this configuration were discovered during E3SM testing,
and these changes fix all uncovered issues.

These updates fix the diagnostics and lnd2rof coupling for the
case when lnd and rof are on the same grid. The diagnostics issue 
is easily seen in a small test configuration of the trigrid in E3SM,
in this case a fully-active run of ne4pg2-r05-oQU480, where the
atm is on ne4pg2, lnd/rof are on r05 (half-degree grid), and
ocn/ice are on an oQU480 grid (quasi-uniform 480-km spacing 
MPAS grid). Coupler output shows an issue even in the area budget:
```
(seq_diag_print_mct) NET AREA BUDGET (m2/m2): period =  monthly: date =     10201     0
                           atm            lnd            ocn         ice nh         ice sh        *SUM*
            area    -1.00000000     0.24638554     0.63952167     0.02429621     0.03700782    -0.05278876
```
The modification to the diagnostics fixes this problem but exposes a
similar issue in the actual lnd2rof coupling. The area budget is drastically
improved:
```
(seq_diag_print_mct) ATM_to_CPL AREA BUDGET (m2/m2): period =  monthly: date =     10201     0
                       a2c_atm        a2c_lnd        a2c_inh        a2c_ish        a2c_ocn        *SUM*
            area    -1.00000000     0.29916044     0.02427324     0.03706142     0.63949937    -0.00000553
```
but now shows a water budget problem:
```
(seq_diag_print_mct) NET WATER BUDGET (kg/m2s*1e6): period = all_time: date =    110101     0
                           atm            lnd            rof            ocn         ice nh         ice sh            glc        *SUM*
         wfreeze     0.00000000     0.00000000     0.00000000    -0.13322464     0.06696238     0.06626226     0.00000000    -0.00000000
           wmelt     0.00000000     0.00000000     0.00000000     0.76839575    -0.28894893    -0.47944686     0.00000000    -0.00000003
           wrain   -38.03418055     9.00948767     0.00000000    28.98419372     0.01967490     0.02089916     0.00000000     0.00007490
           wsnow    -0.88243926     0.46788059     0.00000000     0.29608527     0.04496143     0.07348853     0.00000000    -0.00002345
           wevap    39.01586187    -7.04244184     0.00000000   -31.97266799    -0.00010192    -0.00144629     0.00000000    -0.00079617
         wrunoff     0.00000000    -1.67305761     0.01005664     1.21279836     0.00000000     0.00000000     0.00000000    -0.45020262
         wfrzrof     0.00000000    -0.14139458     0.00000000     0.12515946     0.00000000     0.00000000     0.00000000    -0.01623512
           *SUM*     0.09924206     0.62047423     0.01005664    -0.71926006    -0.15745215    -0.32024321     0.00000000    -0.46718249
```
The other changes to the coupling resolve this.

Test suite: scripts_regression_test, hand testing with E3SM tri-grid cases.
Test baseline: 
Test namelist changes: 
Test status: bit-for-bit for bi-grid, climate-changing for trigrid (due to current errors)

Fixes [CIME Github issue #]

User interface changes?: N

Update gh-pages html: N

Code review:  billsacks, jedwards4b, jgfouca, apcraig

### DIFF
--- a/src/drivers/mct/main/prep_rof_mod.F90
+++ b/src/drivers/mct/main/prep_rof_mod.F90
@@ -82,6 +82,8 @@ module prep_rof_mod
   character(CXX) :: lnd2rof_normal_fluxes
   ! whether the model is being run with a separate irrigation field
   logical :: have_irrig_field
+  ! samegrid atm and lnd
+  logical :: samegrid_al   ! samegrid atm and lnd
   !================================================================================================
 
 contains
@@ -175,6 +177,8 @@ contains
 
        samegrid_lr = .true.
        if (trim(lnd_gnam) /= trim(rof_gnam)) samegrid_lr = .false.
+       samegrid_al = .true.
+       if (trim(atm_gnam) /= trim(lnd_gnam)) samegrid_al = .false.
 
        if (lnd_c2_rof) then
           if (iamroot_CPLID) then
@@ -451,10 +455,11 @@ contains
     integer, save :: index_l2x_coszen_str
     integer, save :: index_x2r_coszen_str
 
-    integer, save :: index_lfrac
+    integer, save :: index_frac
+    real(r8)      :: frac
+    character(CL) :: fracstr
     logical, save :: first_time = .true.
     logical, save :: flds_wiso_rof = .false.
-    real(r8)      :: lfrac
     integer       :: nflds,lsize
     logical       :: iamroot
     character(CL) :: field        ! field string
@@ -519,27 +524,32 @@ contains
           index_x2r_Flrl_rofl_HDO = mct_aVect_indexRA(x2r_r,'Flrl_rofl_HDO' )
           index_x2r_Flrl_rofi_HDO = mct_aVect_indexRA(x2r_r,'Flrl_rofi_HDO' )
        end if
-       index_lfrac = mct_aVect_indexRA(fractions_r,"lfrac")
 
-       index_lfrac = mct_aVect_indexRA(fractions_r,"lfrac")
+       if (samegrid_al) then
+          index_frac = mct_aVect_indexRA(fractions_r,"lfrac")
+          fracstr = 'lfrac'
+       else
+          index_frac = mct_aVect_indexRA(fractions_r,"lfrin")
+          fracstr = 'lfrin'
+       endif
 
        mrgstr(index_x2r_Flrl_rofsur) = trim(mrgstr(index_x2r_Flrl_rofsur))//' = '// &
-            'lfrac*l2x%Flrl_rofsur'
+            trim(fracstr)//'*l2x%Flrl_rofsur'
        mrgstr(index_x2r_Flrl_rofgwl) = trim(mrgstr(index_x2r_Flrl_rofgwl))//' = '// &
-            'lfrac*l2x%Flrl_rofgwl'
+            trim(fracstr)//'*l2x%Flrl_rofgwl'
        mrgstr(index_x2r_Flrl_rofsub) = trim(mrgstr(index_x2r_Flrl_rofsub))//' = '// &
-            'lfrac*l2x%Flrl_rofsub'
+            trim(fracstr)//'*l2x%Flrl_rofsub'
        mrgstr(index_x2r_Flrl_rofdto) = trim(mrgstr(index_x2r_Flrl_rofdto))//' = '// &
-            'lfrac*l2x%Flrl_rofdto'
+            trim(fracstr)//'*l2x%Flrl_rofdto'
        mrgstr(index_x2r_Flrl_rofi) = trim(mrgstr(index_x2r_Flrl_rofi))//' = '// &
-            'lfrac*l2x%Flrl_rofi'
+            trim(fracstr)//'*l2x%Flrl_rofi'
        if (trim(cime_model).eq.'e3sm') then
           mrgstr(index_x2r_Flrl_demand) = trim(mrgstr(index_x2r_Flrl_demand))//' = '// &
-               'lfrac*l2x%Flrl_demand'
+               trim(fracstr)//'*l2x%Flrl_demand'
        endif
        if (have_irrig_field) then
           mrgstr(index_x2r_Flrl_irrig) = trim(mrgstr(index_x2r_Flrl_irrig))//' = '// &
-               'lfrac*l2x%Flrl_irrig'
+               trim(fracstr)//'*l2x%Flrl_irrig'
        end if
        if(trim(cime_model) .eq. 'e3sm') then
           mrgstr(index_x2r_Flrl_Tqsur) = trim(mrgstr(index_x2r_Flrl_Tqsur))//' = '//'l2x%Flrl_Tqsur'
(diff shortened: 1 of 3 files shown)


jedwards4b (commented):
Please document the testing you have done in the opening comment of the PR. 
Did you add any test for trigrid?

jonbob:
I didn't add any trigrid test, but tested in E3SM with several trigrid configurations as well as making sure it didn't change answers for non-trigrid cases. I'm not sure how much documentation you want from the E3SM side? Our initial tests showed the water budget was not closed for trigrid, so we came up with a patch to seq_diags. Runs with that patch showed an actual problem with lnd2rof coupling when they were on the same grid, so the rest of this PR fixes that. Does CESM run any of these configurations? We can wrap more of the changes with cime_model logic, if that helps

billsacks:
I had some discussion with @mvertens and @apcraig in 2015 on the issue of lfrac vs. lfrin which neither I nor anybody else ever followed up on. The result of that discussion was that we all felt that the actual error is that lfrin is used at all: we all felt at the time that we should drop most uses of lfrin (such as the normalization factor in lnd -> atm mapping) and use lfrac in its place. I'm sorry that I never opened an issue on this. This discussion was based on drawing and examining a lot of pictures of example grids, and not based on any empirical evidence from coupler diagnostics.

Here was my original email, which admittedly does not express a lot of confidence:

> Hi Tony and Mariana,
> 
> (Tony: I had a long conversation with Mariana yesterday about lfrin vs. lfrac, since I was still confused about the use of these two subtly different variables.)
> 
> I spent a couple of hours this morning working through some examples, where I was considering whether we should use lfrac or lfrin in the normalization when mapping from LND -> GLC. See section 3.2.3 in https://docs.google.com/document/d/144PZ63gwpz875U4EYEI7NuxnF__V955QJOxV9LcEAL0/edit# (starting on p. 14).
> 
> Based on these examples, I have convinced myself that:
> 
> (1) In order to conserve, we need to normalize using lfrac rather than lfrin in the LND -> GLC mapping.
> 
> (2) Moreover, normalization by lfrin in the LND -> ATM mapping breaks conservation, regardless of the normalization factor we choose for the LND -> GLC mapping.
> 
> Thus, I concluded based on this that lfrin should never be used for normalization. Its only purpose in the system should be to set lfrac when you're running without ice/ocean. So it should be a local variable in seq_frac_mct, never referenced outside of there.
> 
> 
> Do you agree with my assessment? I have based these examples on my latest understanding of lfrac and lfrin, but I am not at all confident of this understanding. So it would be really good to get an independent check by one of you.
> 
> If you agree with my assessment, I would propose the following:
> 
> (a) I use normalization by lfrac in my new code, even though it is inconsistent with other mappings, because it appears to be needed for conservation.
> 
> (b) In a separate set of changes, which can come before or after my LND <-> GLC changes, we replace lfrin normalization with lfrac normalization, in prep_atm_mod and prep_rof_mod.
> 
> 
> Bill

And here was Tony's reply, which gave me more confidence:


> This is what I was expecting.  When all we were doing is mapping land to atm,
> we could use lfrin.  It gave us better science and conserved.  That was easy.
> When we separated rtm and then added flooding, this became more complicated
> but still worked because neither of those coupling involved conservation with a
> third component.  But now that we are coupling to glc that does have conservation
> issues with other models (runoff, lnd, etc), I think we have to abandon lfrin completely
> as it not only breaks conservation for lnd-glc coupling but that interaction also
> makes the lfrin invalid for the other land mappings as well.
> 
> I agree with your plan (a) and (b) below.  I think we have no choice.
> 
> tony.........


Based on this, I would suggest that you see if changing lfrin to lfrac in prep_atm_mod and prep_rof_mod gives conservation according to the coupler diagnostics you've been looking at.

billsacks:
By the way: In CESM, trigrid is technically supported, but as far as I know, it is rarely if ever used for science. This is why we let this issue drop when it came up 6 years ago.

jonbob:
@billsacks - I should have said that Tony helped me with all this. For E3SM, it does fix conservation issues in the diagnostics, as well as ensure the cpl budget matches that inside the lnd model. Would it help for me to post some of those results?

apcraig:
This update fixes the conservation and budget diagnostics in trigrid configurations.  It also preserves the non-trigrid configurations bit-for-bit and imposes no new requirements on them.  A more general solution might be better, but it would require more effort and could have a significant impact on backwards compatibility.  One issue I'll call out is that definition of lfrin is different in trigrid and non-trigrid.  That would probably be the first thing that needs to get resolved and why I advocated getting rid of lfrin all-together many years ago.  But that might require that we fundamentally redefine lfrac in the land model and the atm model and impose new constraints on non-trigrid land fractions in the land model input.  Finally, the solution proposed here does not address any potential existing glc issues, those were not considered.  But non-trigrid configurations are unchanged, and I'm not sure how much testing has been done with trigrid with glc on.

jonbob:
Following up on @apcraig's comment, I haven't done any testing with GLC on. E3SM still doesn't really run much with active GLC.

billsacks (changes requested):
Thank you for the explanation, @jonbob and @apcraig . Given your explanation, I am okay with this change, but @jonbob please fill out the PR template more completely, documenting all testing done (note that we expect at least the testing documented here: https://github.com/ESMCI/cime/wiki/Testing#changes-to-fortran-or-to-the-test-scripts-in-systemtests) and, importantly, documenting the configurations in which this changes answers.

jgfouca:
I'm running scripts_regression_tests now.

jonbob:
Thanks @jgfouca

jgfouca:
Tests passed.

jgfouca (approved):

jgfouca:
@billsacks , did @jonbob address your concern about documented testing?

billsacks:
In the answer changes section, I think it would be better to start with what DOES change answers and by how much (roundoff vs. climate changing). This line is very important to our process and listing it as bit-for-bit (even with a qualification) may lead to issues down the line for us.

jonbob:
@billsacks - I added more information about what changes to the description above, but if it has more detail than is appropriate I'm happy to move it to this discussion part. The idea is that the PR should be BFB for configurations with atm/lnd on the same grid but it is climate-changing for trigrid cases, but only because they are incorrect currently. As-is, the lnd model does not pass fields correctly to the rof, so over time it has a continuous uptake of water. In E3SM, this results in a steady decrease of sea surface height in the ocn.

billsacks (approved):
Thank you for updating the top-level description. This looks good now.

jonbob:
Thanks @jgfouca for shepherding this through

jgfouca:
@rljacob , is this ready to merge?

rljacob:
Its been merged to CIME.  You can update the cime module in E3SM anytime.

jgfouca:
@rljacob , apologies, my tab was in a stale state. I merged this 6 days ago :)

PR #3860, opened 2021-02-20
Fix compare test results undefined

### LINKED ISSUE #3814: compare_test_results: logfile_name can be undefined
In some circumstances, logfile_name can be undefined when running compare_test_results.

This happened to me when running like this:

```
./compare_test_results --hist-only -b ctsm5.1.dev019 --baseline-root $scratch/cesm_baselines -r /scratch/cluster/sacks/tests_1222-131830iz -t 1222-131830iz

SKIP ERS_Ld5.f10_f10_musgs.I2000Clm50BgcCropRtm.izumi_nag.rtm-default BASELINE Run phase did not pass

  File "./compare_test_results", line 105, in <module>
    _main_func(__doc__)
  File "./compare_test_results", line 99, in _main_func
    success = compare_test_results(baseline_name, baseline_root, test_root, compiler, test_id, compare_tests, namelists_only, hist_only)
  File "/home/sacks/ctsm_code/ctsm/cime/scripts/Tools/../../scripts/lib/CIME/compare_test_results.py", line 158, in compare_test_results
    append_status_cprnc_log(brief_result, logfile_name, test_dir)
UnboundLocalError: local variable 'logfile_name' referenced before assignment
```

It looks like logfile_name is only defined when it finds a test that actually needs comparison (so not a skipped test like this one), yet it is referenced for all tests in the append_status_cprnc_log line.

I'm not positive,  but I think this error was introduced in #2079

@jgfouca

jasonb5:
Fixes case where a skipped test would result in logfile_name being
undefined in compare_test_results.py.

Test baseline: 
Test namelist changes: 
Test status:

Fixes #3814 

User interface changes?:  

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/tests/test_compare_test_results.py
+++ b/scripts/lib/CIME/tests/test_compare_test_results.py
@@ -1,0 +1,99 @@
+#/usr/bin/env python
+
+"""
+This module contains unit tests for compare_test_results
+"""
+
+import time
+import unittest
+import tempfile
+import os
+import shutil
+
+from CIME import utils
+from CIME import compare_test_results
+from CIME.test_status import *
+from CIME.tests.case_fake import CaseFake
+
+
+class TestCaseFake(unittest.TestCase):
+
+    def setUp(self):
+        self.tempdir = tempfile.mkdtemp()
+        self.test_root = os.path.join(self.tempdir, "tests")
+        self.baseline_root = os.path.join(self.test_root, "baselines")
+
+        # TODO switch to unittest.mock
+        self._old_strftime = utils.time.strftime
+        utils.time.strftime = lambda *args: "2021-02-20"
+
+        self._old_init = CaseFake.__init__
+        CaseFake.__init__ = lambda x, y, *args: self._old_init(x, y, create_case_root=False)
+
+        self._old_case = compare_test_results.Case
+        compare_test_results.Case = CaseFake
+
+    def tearDown(self):
+        utils.time.strftime = self._old_strftime
+        CaseFake.__init__ = self._old_init
+        compare_test_results.Case = self._old_case
+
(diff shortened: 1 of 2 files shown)


billsacks:
Thank you for working on this!

scripts/lib/CIME/tests/test_compare_test_results.py
```diff
@@ -0,0 +1,98 @@
+#/usr/bin/env python
+
+"""
+This module contains unit tests for compare_test_results
+"""
+
+import unittest
```
jgfouca, line 7: @jasonb5 , thanks for adding a test! We have been putting most of our testing in scripts/tests/scripts_regression_tests.py. You can put this code there or keep it here and just make sure scripts_regression_tests.py calls this test.

billsacks: @jgfouca that's not true: we have a number of unit tests in the directory where @jasonb5 added this, and if things are set up right, they should all be invoked from A_RunUnitTests. We had a discussion a few years ago in which a few of us agreed that it would probably be better to put new unit tests in their own module there rather than adding to the already-huge scripts_regression_tests. Which is all to say: I like how @jasonb5 set up this test.

jasonb5: @jgfouca @billsacks I'll keep the test were it is and double check that it's being excercised with A_RunUnitTests.

jgfouca: Thanks guys, I remember that discussion.

jasonb5: Ok double checked the tests are picked up by A_RunUnitTests. This can be resolved and i'll move the PR out of draft.

jgfouca (approved):

PR #3861, opened 2021-02-20
Logs tail of cprnc outputs to TestStatus.log

### LINKED ISSUE #3831: Add tail from *cprnc.out files in TestStatus.log
If possible, can we have a tail -n 20 for *nc.cprnc.out files in TestStatus.log ? it is a pain to open each file just for the last few lines. One example is when diffs are only fill diffs, yet, each *cprnc.out file in a suite needs to be opened to confirm that the diffs are harmless.

jasonb5:
Adds tail of cprnc logs to TestStatus.log.

Test baseline: 
Test namelist changes: 
Test status:

Fixes #3831 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: Bill Sacks


billsacks (commented):
Thank you for working on this. One question I have is: It looks like `_log_cprnc_output_tail` is potentially called multiple times (once for `_compare_baseline` and one or more (e.g., twice for the ERI test) times for `_component_compare_test`). Will this lead to the tail from each cprnc file appearing in the TestStatus.log file multiple times? It seems like this should be changed to operate on only the cprnc out files that were generated from the current comparison, by making the globbing more specific.

jasonb5:
@billsacks I see what you mean with ERI. It looks like I missed some classes that overrride `_compare_baseline` and don't call the parent's method as well.

billsacks:
Does this address my earlier comment? I don't see how, but if you can explain or show an example of output (for example, from an ERI test that also does baseline comparisons) then that would be great.

jgfouca:
@billsacks , it looks like @jasonb5 has addressed your comment in a recent commit. Can you re-review?

billsacks (approved):
That looks better now - thank you!

PR #3864, opened 2021-02-22
SCORPIO needs _PNETCDF

jgfouca:
Test suite: ./scripts_regression_tests --fast
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: jedwards

### DIFF
--- a/scripts/Tools/Makefile
+++ b/scripts/Tools/Makefile
@@ -249,6 +249,7 @@ ifeq ($(findstring -DLINUX,$(CPPDEFS)),-DLINUX)
 endif
 
 ifdef LIB_PNETCDF
+   CPPDEFS += -D_PNETCDF
    SLIBS += -L$(LIB_PNETCDF) -lpnetcdf
 endif
 


jedwards4b (approved):

jedwards4b:
Is the only reason this needs to be put back because of the macro in shr_pio_mod.F90?  I believe that netcdf versions have
supported that flag for a couple of years now and the macro can be removed.

rljacob:
We still support building SCORPIO without netcdf (pnetcdf-only) so maybe that's why?

dqwu:
> We still support building SCORPIO without netcdf (pnetcdf-only) so maybe that's why?

Without this macro, PIO_NETCDF_FORMAT will not use 64bit_data
```
#ifdef _PNETCDF
    elseif ( pio_netcdf_format .eq. '64BIT_DATA' ) then
       pio_netcdf_ioformat = PIO_64BIT_DATA
#endif
```

jedwards4b:
I believe you read that incorrectly.  Removing the macro means the code would be 
```
    elseif ( pio_netcdf_format .eq. '64BIT_DATA' ) then
       pio_netcdf_ioformat = PIO_64BIT_DATA
```
and so would support this option even if pnetcdf is not used.

dqwu:
> I believe you read that incorrectly. Removing the macro means the code would be
> 
> ```
>     elseif ( pio_netcdf_format .eq. '64BIT_DATA' ) then
>        pio_netcdf_ioformat = PIO_64BIT_DATA
> ```
> 
> and so would support this option even if pnetcdf is not used.

I mean if "#ifdef _PNETCDF" is still being used in shr_pio_mod.F90 (as
of current version), we definitely need to define _PNETCDF.

Of course, if shr_pio_mod.F90 will be updated in the future to remove
the use of that macro (since NetCDF should also support PIO_64BIT_DATA
now), "CPPDEFS += -D_PNETCDF" would be optional for component csm_share
as you have mentioned. We still need to make sure that this macro is not
used in other source files of component csm_share, though.

PR #3870, opened 2021-03-01
make 64bit data default format

jedwards4b:
Makes the 64bit_data format the default.  Removes a requirement that pnetcdf must be built to use this format.

Test baseline: 
Test namelist changes: 

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/share/unit_test_stubs/pio/pio.F90.in
+++ b/src/share/unit_test_stubs/pio/pio.F90.in
@@ -63,6 +63,7 @@ module pio
        r8        = selected_real_kind(13)
 
   integer, parameter, public :: PIO_64BIT_OFFSET = 0
+  integer, parameter, public :: PIO_64BIT_DATA = 0
   integer(i4), parameter, public :: PIO_BCAST_ERROR = 0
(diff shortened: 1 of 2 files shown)


fischer-ncar (approved):

billsacks:
@jedwards4b I'm curious: You mentioned last week that this might have some performance implications. Do you have any sense of how much performance penalty we might expect to pay with this change? Or do you think it will be negligible?

I'm just thinking that it's probably quite rare that someone will be doing a run that actually requires this change (since you need a variable with over 4 billion elements to require this), so if the performance hit is noticeable, we may want to make this conditional on some heuristic - e.g., only making this change for high res runs, for some definition of high res. (Even normal high res runs – like 1/4 or 1/10 degree – seem very unlikely to require this.)

jedwards4b:
I mentioned that there is a file size implication.  I have not seen any performance implications.

jedwards4b:
1/4 and 1/10 degree runs do require this for some components.

billsacks:
Okay, sounds good - thanks for the clarification.

PR #3873, opened 2021-03-01
Add run scripts to provenance capture

sarats:
When a run script is used to setup simulations, it is copied in
the case directory under 'run_script_provenance'. Adding that to
the list of files captured in the provenance/performance archive.

Test suite: 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes : https://github.com/E3SM-Project/E3SM/issues/4118

User interface changes?: N

Code review: @jgfouca

### DIFF
--- a/scripts/lib/CIME/provenance.py
+++ b/scripts/lib/CIME/provenance.py
@@ -228,6 +228,7 @@ def _save_prerun_timing_e3sm(case, lid):
     os.mkdir(case_docs)
     globs_to_copy = [
         "CaseDocs/*",
+        "run_script_provenance/*",
         "*.run",
         ".*.run",
         "*.xml",


jgfouca (approved):

rljacob:
What happens if "run_scripts_provenance" is not present?  It won't be for cases started directly with create_newcase but PACE will still need to capture its metadata.

sarats:
As I understand, if 'run_scripts_provenance' is not present, the the glob below would return [] and the copy would not be attempted.
> for item in glob.glob(os.path.join(caseroot, glob_to_copy)):

Not sure what you mean when they use 'create_newcase'. We can still capture a run script if it's present, otherwise just capture all XMLs and inputs like we always do.

rljacob:
ok sounds good.

PR #3881, opened 2021-03-12
Disable kokkos tests when building for E3SM

jgfouca:
Generate_makefile (which is how E3SM builds Kokkos) has Kokkos_ENABLE_TESTS ON by default and this is causing slower builds and increasing the chances of build problems. I cant imagine why E3SM would ever want kokkos tests, so I'm thinking we should turn this off.

Test suite: By hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:


jgfouca:
@oksanaguba , @bartgol

PR #3883, opened 2021-03-13
Support for SCM functionality for multiple columns

bogensch:
PR changes the name of the ambiguously named flag “iop_mode” with “scm_multcols”, which turns on single column model (SCM) functionality but for multiple columns.  Specifically, its purpose is to take the column on the globe intended to be run in SCM mode, but propagates that point (i.e. meaning surface type and location etc.) to multiple columns, as defined by the user.  All columns are then forced identically according to the Intensive Observation Period (IOP) file, which is typically required to run in SCM mode.  This PR also adds the flexibility for the user to determine the number of columns in the x & y direction (PTS_NX, PTX_NY and scm_nx, scm_ny).

Background/Motivation for changes: This PR is intended to support running a doubly periodic cloud resolving model with the planar version of HOMME (specifically, the doubly periodic version of SCREAM) but could be potentially useful for other applications.

Test suite: scripts_regression_tests on cori
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/components/data_comps_mct/docn/src/docn_comp_mod.F90
+++ b/src/components/data_comps_mct/docn/src/docn_comp_mod.F90
@@ -95,7 +95,7 @@ CONTAINS
        seq_flds_x2o_fields, seq_flds_o2x_fields, &
        SDOCN, gsmap, ggrid, mpicom, compid, my_task, master_task, &
        inst_suffix, inst_name, logunit, read_restart, &
-       scmMode, iop_mode, scmlat, scmlon, iop_nx, iop_ny)
+       scmMode, scm_domain, scmlat, scmlon, scm_nx, scm_ny)
 
     ! !DESCRIPTION: initialize docn model
     use pio        , only : iosystem_desc_t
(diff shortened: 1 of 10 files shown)


billsacks:
At a glance, I'm thinking this should wait until we have completed the cime split (separating out the Fortran code so CESM is using a different version from E3SM); otherwise it probably needs a review from the CESM side as well as the E3SM side.

mvertens:
I agree with the point that @billsacks raises. I think this should wait until the CIME source code split is finalized. We are aiming for the end of March for this on the CESM end.

bogensch:
Thanks @billsacks and @mvertens.  Understood and sounds good.

jgfouca:
@bogensch , RobJ has completed the CIME split. I recommend you move this work over the E3SM master. I can help with the details if you need.

rljacob:
This PR makes changes to src/share which I haven't moved yet.  But it should be redone in the E3SM repo anyway.

bogensch:
Thanks @rljacob !

PR #3899, opened 2021-03-26
support for threading in two different ways for esmf

jedwards4b:
Adds threading support for nuopc in two ways with the introduction of a new xml variable in cmeps 
ESMF_AWARE_THREADING.   When this variable is False or None the threading implementation is the same as that in mct, this is the default.  When ESMF_AWARE_THREADING is True the driver uses all available mpi tasks on a given node and component threads are launched using pthreads.  This allows for each component to specify its thread count independent of the other components.  It also changes the definition of ROOTPE.  The pelayout was changed to add PSTRID and to document the value of ESMF_AWARE_THREADING and how that value affects the ROOTPE variable.  

Test suite: scripts_regression_tests.py + manual tests of pelayout script 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]
Depends on https://github.com/ESCOMP/CMEPS/pull/144
 
User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -635,7 +635,6 @@ This allows using a different mpirun command to launch unit tests
       <env name="UGCSFIXEDFILEPATH">/glade/work/turuncu/FV3GFS/fix_am</env>
       <env name="UGCSADDONPATH">/glade/work/turuncu/FV3GFS/addon</env>
       <env name="OMP_WAIT_POLICY">PASSIVE</env>
-<!--      <env name="OMP_PROC_BIND">true</env> -->
       <env name="MPI_DSM_VERBOSE">true</env>
     </environment_variables>
     <environment_variables unit_testing="true">
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -644,8 +644,6 @@ This allows using a different mpirun command to launch unit tests
       <env name="TMPDIR">/glade/scratch/$USER</env>
     </environment_variables>
     <resource_limits>
-      <!--      <resource name="RLIMIT_STACK"> 80000000</resource> -->
-      <!-- set to unlimited -->
       <resource name="RLIMIT_STACK">-1</resource>
     </resource_limits>
   </machine>
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -644,7 +644,9 @@ This allows using a different mpirun command to launch unit tests
       <env name="TMPDIR">/glade/scratch/$USER</env>
     </environment_variables>
     <resource_limits>
-      <resource name="RLIMIT_STACK"> 80000000</resource>
+      <!--      <resource name="RLIMIT_STACK"> 80000000</resource> -->
+      <!-- set to unlimited -->
+      <resource name="RLIMIT_STACK">-1</resource>
     </resource_limits>
   </machine>
 
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -472,6 +472,7 @@ This allows using a different mpirun command to launch unit tests
       <arguments>
         <arg name="labelstdout">-p "%g:"</arg>
         <arg name="num_tasks"> -np {{ total_tasks }}</arg>
+        <arg name="zthreadplacement"> omplace -tm open64 -vv</arg>
       </arguments>
     </mpirun>
     <mpirun mpilib="mpt" queue="share" comp_interface="nuopc">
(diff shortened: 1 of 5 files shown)


config/cesm/machines/config_machines.xml
```diff
@@ -472,6 +472,7 @@ This allows using a different mpirun command to launch unit tests
       <arguments>
         <arg name="labelstdout">-p "%g:"</arg>
         <arg name="num_tasks"> -np {{ total_tasks }}</arg>
+        <arg name="zthreadplacement"> omplace -tm open64 -vv</arg>
```
jedwards4b, line 475: Needed when ESMF_AWARE_THREADING is off, not desirable when it is on. leaving it on for now.

config/cesm/machines/config_machines.xml
```diff
@@ -633,16 +634,18 @@ This allows using a different mpirun command to launch unit tests
...
     <environment_variables unit_testing="true">
       <env name="MPI_USE_ARRAY">false</env>
     </environment_variables>
     <environment_variables queue="share">
       <env name="TMPDIR">/glade/scratch/$USER</env>
-<!--      <env name="MPI_USE_ARRAY">false</env> -->
     </environment_variables>
     <resource_limits>
-      <resource name="RLIMIT_STACK"> -1</resource>
+      <resource name="RLIMIT_STACK"> 80000000</resource>
```
jedwards4b, line 648: -1 value does not work with pthreads, the value here was chosen via trial and error.  We need to find a better way to set this.

scripts/lib/CIME/get_timing.py
```diff
@@ -90,7 +90,8 @@ def _gettime2_mct(self, heading_padded):
     def _gettime2_nuopc(self):
         self.nprocs = 0
         self.ncount = 0
-        expression = re.compile(r'\s*\MED:\s*\(med_phases_profile\)\s+(\d+)\s+(\d+)')
+#        expression = re.compile(r'\s*\MED:\s*\(med_phases_profile\)\s+(\d+)\s+(\d+)')
```
jedwards4b, line 93: unrelated bug fix for timing output

jgfouca (approved):

jgfouca (approved):

scripts/lib/CIME/get_timing.py
```diff
@@ -459,11 +459,10 @@ def _getTiming(self, inst=0):
...
         tmaxr = 0
         if tmax > 0:
-            print ("HERE adays {} tmax {}".format(adays, tmax))
             tmaxr = adays*86400.0/(tmax*365.0)
 
         self.write("\n")
-        self.write("  total pes active           : {} \n".format(totalpes*maxthrds*smt_factor ))
+        self.write("  total pes active           : {} \n".format(totalpes*smt_factor ))
         self.write("  mpi tasks per node               : {} \n".format(max_mpitasks_per_node))
         self.write("  pe count for cost estimate : {} \n".format(pecost))
```
amametjanov, line 467: Since you're here anyways, could you consider updating this?
1. Indentation of `:` for "mpi tasks per node" line
2. It's hard to estimate the cost of a job: e.g. "charge factor" at NERSC Cori. I think it'd be useful to simplify and write
```
  total nodes         : N (num_nodes)
  mpi tasks per node  : M (MAX_MPITASKS_PER_NODE)
  threads per mpi task: T (max_threads)
  pes per node        : MxT
  total pes active    : MxTxN
```
i.e. remove `smt_factor` from all PE calculations. It's hard to account for SMT/hyper-threading correctly without introducing additional XML variable for actual number of cores per node: e.g. OLCF Summit node with 42 cores per node capable of SMT[1,2,4].

fischer-ncar (approved):

scripts/lib/CIME/get_timing.py
jedwards4b: I have addressed the indentation.  The smt_factor change is more extensive and not something I feel can be handled as a minor addition.  I would be happy to review a new PR if you would like to submit one.

amametjanov: One-line change:
```
diff --git a/scripts/lib/CIME/get_timing.py b/scripts/lib/CIME/get_timing.py
index 68c2e7769..6e197ead4 100644
--- a/scripts/lib/CIME/get_timing.py
+++ b/scripts/lib/CIME/get_timing.py
@@ -405,7 +405,7 @@ class _TimingParser:
                 comp_label = m.comp + inst_label
             else:
                 comp_label = m.comp
-            self.write("  {} = {:<8s}   {:<6d}      {:<6d}   {:<6d} x {:<6d}  {:<6d} ({:<6d}) \n".format(m.name.lower(), comp_label, (m.ntasks*m.nthrds *smt_factor), m.rootpe, m.ntasks, m.nthrds, m.ninst, m.pstrid))
+            self.write("  {} = {:<8s}   {:<6d}      {:<6d}   {:<6d} x {:<6d}  {:<6d} ({:<6d}) \n".format(m.name.lower(), comp_label, (m.ntasks*m.nthrds), m.rootpe, m.ntasks, m.nthrds, m.ninst, m.pstrid))
             if m.nthrds > maxthrds:
                 maxthrds = m.nthrds
         if self._driver == 'nuopc':
```

PR #3901, opened 2021-03-30
Point to new dlnd scpl forcing data

### LINKED ISSUE #3835: Plan to update dlnd forcings for forcing glc (any objections from E3SM?)
We'd like to update the dlnd forcings that are used to force glc (in T compsets). These currently point to `$DIN_LOC_ROOT/lnd/dlnd7/CPLHIST_SNO/i.e20.I1850Clm50Sp.f09_g17.001_c180502`, which (if I remember correctly) aren't actually scientifically meaningful for anything, but were just created for software testing purposes. We'd like to update these to point to forcings from CESM's CMIP6 PI-control run. Moving forward, we're mainly going to be using CDEPS (the nuopc-based data models) for land ice science, so it isn't absolutely critical that we update this in the mct data models. But it would help us to keep mct consistent with nuopc/CDEPS for now, if possible. (And this needs to happen soon, before we'll be done with the separation of the data models into their own repository, as we've discussed.)

The simplest way of making this change would potentially impact E3SM configurations that also use these data for forcing GLC from DLND, that is currently being done. I would put the new forcing data in our inputdata repository so any such configurations should continue to work, but they would change answers. @matthewhoffman @stephenprice @jonbob do you have any objections to my making this change?

billsacks:
Point to new dlnd scpl forcing data.

These data are used to force cases with GLC - i.e., T compsets.

Test suite: Just ran `SMS_Ly35.f09_g17_gl4.T1850G.cheyenne_intel` in the context of CESM's nuopc_dev branch
Test baseline: 
Test namelist changes: 
Test status: climate changing - just for T compsets

Fixes #3835 

User interface changes?: N

Code review:

### DIFF
--- a/src/components/data_comps_mct/dlnd/cime_config/config_component.xml
+++ b/src/components/data_comps_mct/dlnd/cime_config/config_component.xml
@@ -71,7 +71,7 @@
     <default_value>UNSET</default_value>
     <values match="last">
       <value compset="DLND%LCLMR" grid="l%NLDAS">$DIN_LOC_ROOT/lnd/dlnd7/NLDAS</value>
-      <value compset="DLND%SCPL" grid="l%0.9x1.25">$DIN_LOC_ROOT/lnd/dlnd7/CPLHIST_SNO/i.e20.I1850Clm50Sp.f09_g17.001_c180502</value>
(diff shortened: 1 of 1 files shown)


billsacks:
@matthewhoffman gave approval for this for E3SM in discussion in #3835

mvertens (approved):

PR #3906, opened 2021-04-02
fix order of timers

jedwards4b:
Correct timer order in piodarray.F90.in, out of order timers causes a gptl error at end of run.

Test suite:hand testing with SMS.f19_g17.B1850cmip6.cheyenne_intel 
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/.github/workflows/srt.yml
+++ b/.github/workflows/srt.yml
@@ -39,7 +39,7 @@ jobs:
 
       - id: load-env
         run: |
-          #sudo apt-get update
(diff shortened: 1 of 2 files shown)


fischer-ncar (approved):
Do you plan to do a similar PR to master?  Or since it's pio1, just skio it.

PR #3917, opened 2021-04-07
Generalize build directory paths, and prep for CESM share repository

fischer-ncar:
- Generalize paths to source code in build.py and buildlib.py, to use 
   config_files.xml.
- Update Makefile to be able to handle different locations of of mct 
   and pio.  Also add an if statement place holder for furture CESM 
   share repository.
- Update case_clone.py to be able to handle the future CESM share 
   repository.
- Add get_src_root to utils.py, similar to get_cime_root.  Needed this 
   to correct an assumption in utilities.py that srcroot=cimeroot/...
- Added place holder if statement in scripts_regression_tests for the 
  new CESM share repository.
- Did some minor cleanup of build_scripts.

There's a couple of place holder if statements that I added that will be 
updated when we bring in the new CESM share repository.
 
Build_scripts and CMake files will still need to be updated when the 
CESM share repository is ready.  I didn't include any changes since
the structure of the CESM share repository will be changing.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/Tools/Makefile
+++ b/scripts/Tools/Makefile
@@ -612,17 +612,25 @@ endif
 # Drive configure scripts for support libraries (mct)
 #------------------------------------------------------------------------------
 
+ifeq ($(CIME_MODEL),cesm)
+    #This will be changed to EXTERN_PATH = $(SRCROOT)/libraries when the new
+    #share repos are used for cesm2.
+    EXTERN_PATH = $(CIMEROOT)/src/externals
+else
+    EXTERN_PATH = $(CIMEROOT)/src/externals
+endif
+
 $(SHAREDLIBROOT)/$(SHAREDPATH)/mct/Makefile.conf:
 	@echo "SHAREDLIBROOT |$(SHAREDLIBROOT)| SHAREDPATH |$(SHAREDPATH)|"; \
-	$(CONFIG_SHELL) $(CIMEROOT)/src/externals/mct/configure $(CONFIG_ARGS) --srcdir $(CIMEROOT)/src/externals/mct
+	$(CONFIG_SHELL) $(EXTERN_PATH)/mct/configure $(CONFIG_ARGS) --srcdir $(EXTERN_PATH)/mct
 
 $(SHAREDLIBROOT)/$(SHAREDPATH)/mct/mpi-serial/Makefile.conf:
 	@echo "SHAREDLIBROOT |$(SHAREDLIBROOT)| SHAREDPATH |$(SHAREDPATH)|"; \
-	$(CONFIG_SHELL) $(CIMEROOT)/src/externals/mct/mpi-serial/configure $(CONFIG_ARGS) --srcdir $(CIMEROOT)/src/externals/mct/mpi-serial
+	$(CONFIG_SHELL) $(EXTERN_PATH)/mct/mpi-serial/configure $(CONFIG_ARGS) --srcdir $(EXTERN_PATH)/mct/mpi-serial
 
 ifndef IO_LIB_SRCROOT
   ifndef PIO_SRCROOT
-    PIO_SRCROOT = $(CIMEROOT)/src/externals
+    PIO_SRCROOT = $(EXTERN_PATH)
   endif
 
   ifeq ($(PIO_VERSION),2)
@@ -673,7 +681,7 @@ CMAKE_OPTS += -Wno-dev -D CMAKE_Fortran_FLAGS:STRING="$(FFLAGS) $(EXTRA_PIO_FPPD
 	      -D GPTL_PATH:STRING=$(INSTALL_SHAREDPATH) \
 	      -D PIO_ENABLE_TESTS:BOOL=OFF \
 	      -D PIO_USE_MALLOC:BOOL=ON \
-	      -D USER_CMAKE_MODULE_PATH:LIST="$(CIMEROOT)/src/CMake;$(CIMEROOT)/src/externals/pio2/cmake"
+	      -D USER_CMAKE_MODULE_PATH:LIST="$(CIMEROOT)/src/CMake;$(EXTERN_PATH)/pio2/cmake"
 
 ifeq ($(DEBUG), TRUE)
   CMAKE_OPTS += -D PIO_ENABLE_LOGGING=ON
--- a/scripts/lib/CIME/case/case_clone.py
+++ b/scripts/lib/CIME/case/case_clone.py
@@ -3,7 +3,7 @@ create_clone is a member of the Case class from file case.py
 """
 import os, glob, shutil
 from CIME.XML.standard_module_setup import *
-from CIME.utils import expect, check_name, safe_copy
+from CIME.utils import expect, check_name, safe_copy, get_model
 from CIME.simple_compare            import compare_files
 from CIME.locked_files         import lock_file
 from CIME.user_mod_support import apply_user_mods
@@ -39,7 +39,10 @@ def create_clone(self, newcaseroot, keepexe=False, mach_dir=None, project=None,
         logger.warning(" It is NOT recommended to clone cases from different versions of CIME.")
 
     # *** create case object as deepcopy of clone object ***
-    srcroot = os.path.join(newcase_cimeroot,"..")
+    if os.path.isdir(os.path.join(newcase_cimeroot,'share')) and get_model() == "cesm":
+        srcroot = newcase_cimeroot         
+    else:
+        srcroot = os.path.join(newcase_cimeroot,"..")
     newcase = self.copy(newcasename, newcaseroot, newsrcroot=srcroot)
     with newcase:
         newcase.set_value("CIMEROOT", newcase_cimeroot)
--- a/scripts/lib/CIME/utils.py
+++ b/scripts/lib/CIME/utils.py
@@ -254,7 +254,7 @@ def get_cime_root(case=None):
     logger.debug( "CIMEROOT is " + cimeroot)
     return cimeroot
 
-def get_src_root(case=None):
+def get_src_root():
     """
     Return the absolute path to the root of SRCROOT.
 
--- a/scripts/lib/CIME/utils.py
+++ b/scripts/lib/CIME/utils.py
@@ -259,18 +259,11 @@ def get_src_root(case=None):
     Return the absolute path to the root of SRCROOT.
 
     """
-    script_absdir = os.path.abspath(os.path.join(os.path.dirname(__file__),".."))
-    assert script_absdir.endswith(get_python_libs_location_within_cime()), script_absdir
     # This if statement will need to be updated when cesm brings in the new share repos.
     if get_model() == "cesm":
         srcroot = os.path.abspath(os.path.join(get_cime_root(),".."))
     else:
         srcroot = os.path.abspath(os.path.join(get_cime_root(),".."))
-
-    if case is not None:
-        case_srcroot = os.path.abspath(case.get_value("SRCROOT"))
-        srcroot = os.path.abspath(srcroot)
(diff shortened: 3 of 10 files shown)


scripts/tests/scripts_regression_tests.py
```diff
@@ -585,7 +585,12 @@ def test_f_createnewcase_with_user_compset(self):
 
         cls._testdirs.append(testdir)
 
-        pesfile = os.path.join("..","src","drivers","mct","cime_config","config_pes.xml")
+        if CIME.utils.get_model() == "cesm":
```
jedwards4b, line 588: This test should be changed to use the nuopc driver. I have opened issue https://github.com/ESMCI/cime/issues/3920

src/build_scripts/buildlib.internal_components
```diff
@@ -7,7 +7,7 @@ components.
 
 import sys, os
 
-_CIMEROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..","..","..","..", "..")
+_CIMEROOT = os.environ.get("CIMEROOT")
```
jedwards4b, line 10: I think we should add an error message here for the case when CIMEROOT is not defined.

src/build_scripts/buildlib_cmake.internal_components
```diff
@@ -7,7 +7,7 @@ components.
 
 import sys, os
 
-_CIMEROOT = os.path.join(os.path.dirname(os.path.abspath(__file__)), "..","..","..","..", "..")
+_CIMEROOT = os.environ.get("CIMEROOT")
```
jedwards4b, line 10: I think we should add an error message here for the case when CIMEROOT is not defined.

jedwards4b (changes requested):
Just a minor change.

jedwards4b (approved):

jgfouca (approved):

jedwards4b:
@fischer-ncar have you looked at the github workflow at all yet?  I think that we need to address it for this PR.

fischer-ncar:
@jedwards4b I've looked at the github workflow for the new repositories a bit.  But didn't think about it for cime until you mentioned it.  We would need to run the tests with CIME_MODEL=cesm and e3sm.  Would doubling the number of tests be an issue?  We would also need to wget the new repos.

jedwards4b:
@fischer-ncar you've fixed the issues in this pr, thanks.  Let's discuss the increased testing in our next meeting.

fischer-ncar:
@rljacob and @billsacks do you want to review at this PR?

rljacob:
Yes I'll review it today.

scripts/lib/CIME/case/case_clone.py
```diff
@@ -39,7 +39,10 @@ def create_clone(self, newcaseroot, keepexe=False, mach_dir=None, project=None,
         logger.warning(" It is NOT recommended to clone cases from different versions of CIME.")
 
     # *** create case object as deepcopy of clone object ***
-    srcroot = os.path.join(newcase_cimeroot,"..")
+    if os.path.isdir(os.path.join(newcase_cimeroot,'share')) and get_model() == "cesm":
+        srcroot = newcase_cimeroot
```
billsacks, line 43: I don't feel a need to hold up this PR for this comment, but I'll make it here and we can open a separate issue if we don't want to hold up the PR:

I'm concerned about this logic. It seems very possible that the share directory would exist within cime, because someone ran manage_externals from within cime, either by accident or on purpose, yet the user's intent is to use a srcroot one level up from cime. It seems like most/all "normal" users will be using cime in the situation where srcroot is one level up from cimeroot, and so that case should always be the default. The situation where you're running create_clone from a standalone cime checkout can be handled the same way as what I think is done for create_newcase: requiring a `--srcroot` argument. I realize that something still would need to be done for the sake of system tests.

Here's an email I sent about 6 weeks ago; I'm wondering if this would work or if there are problems with this suggestion:

> Maybe you've already seen this, but in case not: I poked around for a few minutes and found this:
> 
> https://github.com/esmci/cime/blob/master/scripts/create_newcase#L101-L108
> 
> However, it looks like there isn't a corresponding --srcroot argument to create_test. I'm thinking that, if you added a --srcroot argument to create_test then that would get you part of the way.
> 
> But this would also be a problem, since some tests use create_clone:
> 
> https://github.com/esmci/cime/blob/master/scripts/lib/CIME/case/case_clone.py#L42
> 
> I'm thinking there are actually a few problems there:
> 
> (1) For user use, create_clone should have the same srcroot logic as create_newcase, with the ability to override it just like in create_newcase. (This isn't important for getting your testing to work, but would be a good thing in general.)
> 
> (2) Maybe for the purposes of system tests, the create_clone function needs an option like keep_original_srcroot; when called from the system tests, that should be set to true, in which case the clone's SRCROOT is set equal to the original case's SRCROOT.
> 
> With those things in place, I'm thinking you'd be able to get this to work if you specify '--srcroot ..' (assuming you're running this from within cime/scripts) to the create_test command.
>

scripts/lib/CIME/utils.py
```diff
@@ -254,6 +254,27 @@ def get_cime_root(case=None):
     logger.debug( "CIMEROOT is " + cimeroot)
     return cimeroot
 
+def get_src_root(case=None):
+    """
+    Return the absolute path to the root of SRCROOT.
+
+    """
+    script_absdir = os.path.abspath(os.path.join(os.path.dirname(__file__),".."))
+    assert script_absdir.endswith(get_python_libs_location_within_cime()), script_absdir
```
billsacks, line 263: What's the point of these two lines?

scripts/lib/CIME/utils.py
```diff
@@ -254,6 +254,27 @@ def get_cime_root(case=None):
...
+    # This if statement will need to be updated when cesm brings in the new share repos.
+    if get_model() == "cesm":
+        srcroot = os.path.abspath(os.path.join(get_cime_root(),".."))
+    else:
+        srcroot = os.path.abspath(os.path.join(get_cime_root(),".."))
+
+    if case is not None:
+        case_srcroot = os.path.abspath(case.get_value("SRCROOT"))
+        srcroot = os.path.abspath(srcroot)
+        expect(srcroot == case_srcroot, "Inconsistent SRCROOT variable: case -> '{}', file location -> '{}'".format(case_srcroot, srcroot))
```
billsacks, line 273: It doesn't seem like this function is currently called with the case argument. Do you envision that it will be in some situations? If so, I'm trying to understand (out of curiosity) if this error-check is catching something funny that a user can do, or if it's mainly checking internal consistency in the cime code.

scripts/lib/CIME/utils.py
```diff
@@ -254,6 +254,27 @@ def get_cime_root(case=None):
     logger.debug( "CIMEROOT is " + cimeroot)
     return cimeroot
 
+def get_src_root(case=None):
+    """
+    Return the absolute path to the root of SRCROOT.
```
billsacks, line 259: Currently this just seems to assume that srcroot is one level up from cimeroot. Can you explain your vision for what this function will become?

billsacks (changes requested):
Thanks a lot, @fischer-ncar ! I have a few questions to get some clarification on the `get_src_root` function, and a concern I'd like to discuss about `create_clone`. None of this needs to hold up this PR, though.

billsacks:
One other question: If I understand some of your logic / placeholders right, it seems like you may be assuming that a standalone cime checkout is always run with model=='cesm'. Am I understanding that right, and if so, is that what we want?

rljacob (approved):

rljacob:
I confirmed I can still build and submit a fully coupled E3SM case on this branch.

rljacob:
@billsacks I think its been true for a while that standalone CIME can only run with cime_model=cesm because of the CMake build e3sm now uses.  @jgfouca can confirm.

scripts/lib/CIME/case/case_clone.py
billsacks: I am deferring this to its own issue: #3923

billsacks:
Thanks @rljacob . Then I'm okay with there being CESM-specific logic for now if that's really what's wanted - though I would still suggest that, if the logic can be written in a way so that it is *not* model-dependent (which I think could be done with some of my thoughts in #3923) then that would be better: I find the proliferation of model-dependent logic to be confusing, especially as we now have 3 possible models, and may have more in the future.

From discussion with @fischer, it sounds like he is planning to remove some of the unnecessary logic from `get_src_root` that is leftover from copying the logic in `get_cime_root`. Once he does that, I'll approve this PR - with further thoughts moved to #3923 that we can hopefully address soon.

scripts/lib/CIME/utils.py
fischer-ncar: No point, I just copied get_cime_root.  I'll remove it.

scripts/lib/CIME/utils.py
fischer-ncar: Just was just coping get_cime_root, I'll remove it.

scripts/lib/CIME/utils.py
fischer-ncar: My vision is if cime_model is cesm and the share directory is in cimeroot, then srcroot=cimeroot.  This would be for a cime stand alone clone.  Otherwise, srcroot=../cimeroot.

fischer-ncar:
@billsacks I'm assuming that if the cime_model is cesm, then the share, cpl7, and libraries directory will be in different places.  Those directories will be in different places based on a cesm2 of cime clone.  If cime_model isn't cesm or the share directory isn't in cimeroot, then do ever thing the way it was done before.

fischer-ncar:
@billsacks I cleaned up get_src_root.

scripts/Tools/Makefile
```diff
@@ -612,17 +612,25 @@ endif
 # Drive configure scripts for support libraries (mct)
 #------------------------------------------------------------------------------
 
+ifeq ($(CIME_MODEL),cesm)
+    #This will be changed to EXTERN_PATH = $(SRCROOT)/libraries when the new
+    #share repos are used for cesm2.
+    EXTERN_PATH = $(CIMEROOT)/src/externals
```
rljacob, line 618: Is this ultimately going to be moved to config_files.xml and set there?

jgfouca:
@rljacob , that's right. E3SM-mode doesn't work without the cmake files in the host repository.

scripts/lib/CIME/utils.py
```diff
@@ -254,6 +254,20 @@ def get_cime_root(case=None):
     logger.debug( "CIMEROOT is " + cimeroot)
     return cimeroot
 
+def get_src_root(case=None):
```
billsacks, line 257: Thanks for the cleanup, @fischer-ncar . One more change: the optional case argument should be removed now, right?

billsacks (changes requested):

billsacks (approved):
Thanks!

scripts/lib/CIME/utils.py
fischer-ncar: Yes.  I've removed it.

scripts/Tools/Makefile
fischer-ncar: My understanding is once we switch to nuopc the mct build will go away.  We will also switch to using only pio2.  At that time I can switch the pio2 path to $SRCROOT/libraries.  This is assuming the e3sm isn't going to be using pio2.  So EXTERN_PATH isn't going to move to config_files.xml, it's just going to go away.

PR #3918, opened 2021-04-07
no need to test ESMF_AWARE_THREADING here

### LINKED ISSUE #3916: Logic for maxinst in env_mach_pes fails with nuopc cases with python3
The maxinst expression in this block of code:

https://github.com/ESMCI/cime/blob/81d4f2033b5cbe7fb4c0b78fe71221b5e2182573/scripts/lib/CIME/XML/env_mach_pes.py#L105-L107

fails with python3:

```
  File "./create_newcase", line 244, in <module>
    _main_func(__doc__)
  File "./create_newcase", line 239, in _main_func
    extra_machines_dir=extra_machines_dir, case_group=case_group)
  File "/glade/work/sacks/ctsm_code/current_branch2/cime/scripts/Tools/../../scripts/lib/CIME/case/case.py", line 1672, in create
    extra_machines_dir=extra_machines_dir, case_group=case_group)
  File "/glade/work/sacks/ctsm_code/current_branch2/cime/scripts/Tools/../../scripts/lib/CIME/case/case.py", line 1120, in configure
    self.initialize_derived_attributes()
  File "/glade/work/sacks/ctsm_code/current_branch2/cime/scripts/Tools/../../scripts/lib/CIME/case/case.py", line 174, in initialize_derived_attributes
    self.total_tasks = env_mach_pes.get_total_tasks(comp_classes) + self.iotasks
  File "/glade/work/sacks/ctsm_code/current_branch2/cime/scripts/Tools/../../scripts/lib/CIME/XML/env_mach_pes.py", line 107, in get_total_tasks
    maxinst = max(maxinst, ninst)
TypeError: '>' not supported between instances of 'NoneType' and 'int'
```

Reproducer on cheyenne: From https://github.com/billsacks/ctsm/commit/bc4758979e36cbb6292962c5f65f9397797a4390

```
module load python
ncar_pylib
./create_newcase --case test_0407b --compset I2000Clm50BgcCropRtm --res f10_f10_mg37 --driver nuopc
```

It looks like python2 allows comparison between None and int, but python3 does not. The issue seems to be that ninst isn't defined per-component when running with nuopc, so ninst gets set to None here.

jedwards4b:
Test suite: hand tested as in #3916 and scripts_regression_tests.py 
Test baseline: 
Test namelist changes: 

Fixes #3916

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/XML/env_mach_pes.py
+++ b/scripts/lib/CIME/XML/env_mach_pes.py
@@ -102,7 +102,7 @@ class EnvMachPes(EnvBase):
             else:
                 nthrds = 1
 
-            if comp != "CPL" and (comp_interface!="nuopc" or not esmf_aware_threading):
+            if comp != "CPL" and comp_interface!="nuopc":
                 ninst = self.get_value("NINST", attribute={"compclass":comp})
                 maxinst = max(maxinst, ninst)
             tt = rootpe + nthrds*((ntasks - 1) * pstrid + 1)


billsacks (approved):
Thanks for the fix. I have confirmed that it fixes the issues I ran into, both on my mac and on cheyenne.

PR #3924, opened 2021-04-10
Fixes phase 1 of multisubmit not executing during rerun

### LINKED ISSUE #3805: PET test does not always rerun correctly
When rerunning a PET test after a failed run the state is not restored correctly.   If the POP ocn model is involved the
new run will stop in the case.build with an error:
```
ERROR: Missing required pointer_file 
```
this is because the CONTINUE_RUN variable in the case is set to TRUE, you can work around this by manually setting CONTINUE_RUN=FALSE.   But then when the PET test is submitted it does not run the first case and goes directly to the case2run.  Once that run completes (and fails because run1 was not done) the test appears to be in a correct initial state again and another run will proceed correctly.

jasonb5:
On a batch system with multisubmit, the first phase would be skipped
if rerunning after a failed phase. This was due to case 1's "RESUBMIT"
variable being decremented before phase 2. During a rerun if the specific 
condition is met, the test will call `_setup_case` which triggers the default
test settings to be loaded.

Test baseline: 
Test namelist changes: 
Test status: 

Fixes #3805 

User interface changes?: n/a
Code review:

### DIFF
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -2,6 +2,16 @@
 Base class for CIME system tests that involve doing two runs and comparing their
 output.
 
+NOTE: Below is the flow of a multisubmit test.
+Non-batch:
+case_submit -> case_run     # PHASE 1
+            -> case_run     # PHASE 2
+
+batch:
+case_submit -> case_run     # PHASE 1
+case_run    -> case_submit
+case_submit -> case_run     # PHASE 2
+
 In the __init__ method for your test, you MUST call
     SystemTestsCompareTwo.__init__
 See the documentation of that method for details.
@@ -224,15 +234,17 @@ class SystemTestsCompareTwo(SystemTestsCommon):
             self._run_one_suffix,
             self._run_two_suffix)
 
-        # On a batch system with a multisubmit test "RESUBMIT" is always 
-        # decremented before running the second phase. If the test is rerun it
-        # will always skip the first phase. To fix this "RESUBMIT" will be set
-        # back to 1 when the test is marked with "IS_FIRST_RUN", is multisubmit
-        # and "RESUBMIT" does not have the correct value.
+        # On a batch system with a multisubmit test "RESUBMIT" is used to track
+        # which phase is being ran. By the end of the test it equals 0. If the
+        # the test fails in a way where the RUN_PHASE is PEND then "RESUBMIT" 
+        # does not get reset to 1 on a rerun and the first phase is skiped 
+        # causing the COMPARE_PHASE to fail. This ensures that "RESUBMIT" will 
+        # get reset if the test state is not correct for a rerun.
+        # NOTE: "IS_FIRST_RUN" is reset in "case_submit.py"
         if (is_first_run and
                 self._multisubmit and
                 self._case1.get_value("RESUBMIT") == 0):
-            self._resetup_case(compare_phase_name)
+            self._resetup_case(RUN_PHASE, reset=True)
 
         first_phase = self._case1.get_value("RESUBMIT") == 1 # Only relevant for multi-submit tests
         run_type = self._case1.get_value("RUN_TYPE")
--- a/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_compare_two.py
@@ -217,8 +217,27 @@ class SystemTestsCompareTwo(SystemTestsCommon):
         Runs both phases of the two-phase test and compares their results
         If success_change is True, success requires some files to be different
         """
+        is_first_run = self._case1.get_value("IS_FIRST_RUN")
+
+        compare_phase_name = "{}_{}_{}".format(
+            COMPARE_PHASE,
+            self._run_one_suffix,
+            self._run_two_suffix)
+
+        # On a batch system with a multisubmit test "RESUBMIT" is always 
+        # decremented before running the second phase. If the test is rerun it
+        # will always skip the first phase. To fix this "RESUBMIT" will be set
+        # back to 1 when the test is marked with "IS_FIRST_RUN", is multisubmit
+        # and "RESUBMIT" does not have the correct value.
+        if (is_first_run and
+                self._multisubmit and
+                self._case1.get_value("RESUBMIT") == 0):
+            self._resetup_case(compare_phase_name)
+
         first_phase = self._case1.get_value("RESUBMIT") == 1 # Only relevant for multi-submit tests
         run_type = self._case1.get_value("RUN_TYPE")
+
+        logger.info("_multisubmit {} first phase {}".format(self._multisubmit, first_phase))
 
         # First run
         if not self._multisubmit or first_phase:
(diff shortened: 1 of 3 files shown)


billsacks (commented):
Thanks for this fix. I especially appreciate the explanatory comment you have added in
`SystemTestsCompareTwo:run_phase` and the test you have added of the new behavior.

I don't think I have ever fully understood the logic for multisubmit tests, so take this comment with a grain of salt: I am confused about the call you added to `self._resetup_case(compare_phase_name)`: do you understand why this is needed, given that `SystemTestsCommon:run` already contains a call to `self._resetup_case(RUN_PHASE)`? Is the problem that the `RUN_PHASE` may be in a `PEND` state at this point, and so the call to `self._resetup_case(RUN_PHASE)` doesn't do anything due to the conditional logic, `if reset or (self._case.get_value("IS_FIRST_RUN") and phase_status != TEST_PEND_STATUS):`?

If that's the problem, then I'm wondering if your new code might have a similar issue, at least in some situations. If I understand correctly, this will *not* do the resetup if this phase has a PEND status. So what would happen in this scenario: The first run succeeds (leading RESUBMIT to be set to 0, right?), then the second run fails. I *think* the compare stage would be left in a PEND state in this scenario. When you go to rerun the test, I'm thinking that the resetup is then *not* triggered.

So I'm wondering if the right solution here might be to change your new call to `self._resetup_case(RUN_PHASE, reset=True)` – i.e., *forcing* a reset in this scenario? With `reset=True`, it looks like the phase doesn't matter for the logic in `_resetup_case`, but it somehow feels more correct to me to set the phase here to `RUN_PHASE` rather than `compare_phase_name`, since what we're actually rerunning is the `RUN_PHASE`.

Again, I don't fully understand this logic, so I'm not positive my suggestion is right – and it may be that my suggestion leads to a too aggressive application of `_resetup_case`. But I wanted to at least put this out for consideration, because I'm having trouble understanding why the fix currently in this PR would solve the problem in all cases given the existing call to `self._resetup_case(RUN_PHASE)` in `SystemTestsCommon:run`.

jgfouca (approved):

jedwards4b (approved):

billsacks:
Also, a question mainly for @jgfouca and @jedwards4b - I'm trying to remember why the PET test is set up as a multi-submit test. I found https://github.com/ESMCI/cime/pull/538 and later https://github.com/ESMCI/cime/pull/2255. Those, together with my vague recollections, make me think that it was changed to a multi-submit test due to issues on one system (or maybe a couple of systems), but that outside of those systems, there is no fundamental reason why it needs to be a multi-submit test. For example, we don't seem to have any trouble with multi-threaded ERP tests on our test systems; those also change nthrds for the second case but aren't set up as multi-submit tests.

I'll be honest: My motivation in asking this question is: Given the trouble we seem to have with the PET test and multi-submit tests in general, my confidence in these multi-submit tests is lower than for other system tests. (I don't mean to criticize whoever put in place the multi-submit logic: I think it's just an inherently very hard problem. A lot of the motivation for setting up SystemTestsCompareTwo using two separate cases was to move away from the need to track and restore state between the two different runs of system tests, because we were constantly having trouble with that logic. But the multi-submit tests inherently bring back that need.) So I'm thinking a bit about adding a new PETS (PET-single-submit) test that works like PET but without multi-submit, and switching to using that in some of our testing. I'm wondering if this will cause its own trouble.

jasonb5:
What I reproduced was a situation where `RUN_PHASE` had the status `PASS` and the `COMPARE_PHASE` would have the `FAIL` status. This was a result of only one of the two outputs being available when the `COMPARE_PHASE` was executed. I missed the `_resetup_case` call in `SystemTestsCommon:run`, now that I see that the second call on the `COMPARE_PHASE` would be redundant.

I'm going to take a second look to figure out why adding that second call fixed it.

jgfouca:
@billsacks , I can't remember the PET details and I also struggle to understand our multisubmit tests. Is it possible the configuration of the second case required a different batch submit command?

billsacks:
@jasonb5 thanks for the clarification. The description in #3805 sounded different to me (a failure during the run, which I assume would have left the compare phase in a PEND state), which is part of what led me to wonder what, if anything, the changes in this PR would change for this situation.

@jgfouca - It sounded to me like the change in thread count led to issues launching the mpi job on skybridge for unclear (to me, anyway) reasons, when there was a single submission. As far as I know, we don't have that issue on most systems.

billsacks (changes requested):
It may be that the current code is fine/right and no other changes are needed, but based on the conversation, I'm going to mark this as "request changes" just to track that I want to re-review before this is merged.

jasonb5:
OK I'm going to convert this to a draft, I need to look at this again. At some point while I was working on understanding how multisubmit work I mutated what I was reproducing.

jgfouca:
@billsacks , this is another related issue: https://github.com/ESMCI/cime/issues/3104

jasonb5:
Moving this out of draft for review, a description of the scenario this fixes can be found [here](https://github.com/ESMCI/cime/issues/3805#issuecomment-818331261).

billsacks:
I'm still confused about how the changes here fix either of the situations that have been discussed:

- If the RUN phase was left in a PEND state (https://github.com/ESMCI/cime/issues/3805#issuecomment-818331261), then I think that, in practice, the compare phase will also be left in a PEND state, due to this code:

https://github.com/ESMCI/cime/blob/8b48a2e2d4b6e1581229e0e6e12c916a44b64585/scripts/lib/CIME/SystemTests/system_tests_compare_two.py#L227-L230

I think the new `_resetup_case` call then won't do anything. I'd be interested for you to try manually setting both RUN and the COMPARE state to PEND in your reproducer to confirm that, if you haven't already. This scenario would be fixed by my above suggestion of forcing a reset via something like `self._resetup_case(RUN_PHASE, reset=True)`

- If the RUN phase was left in a PASS state (https://github.com/ESMCI/cime/pull/3924#issuecomment-818080115) then it seems like the existing call to `_resetup_case` here:

https://github.com/ESMCI/cime/blob/8b48a2e2d4b6e1581229e0e6e12c916a44b64585/scripts/lib/CIME/SystemTests/system_tests_common.py#L147

should be sufficient. If you managed to create a situation where that wasn't enough, then I think we have a bigger problem on our hands, which calls for a more general solution: If a PASSed test doesn't get reset when you rerun it, then I'm concerned that that could impact non-multisubmit tests as well as multisubmit tests. Would you be able to describe the exact setup you used to reproduce this issue with the RUN phase being left in a PASS state?

See also #3940 where I wonder if we need a more general solution.

jasonb5:
Below is the state I used to reproduce the issue. I was not able to determine what led to this state but had @jedwards4b (original reporter) check that this seemed correct. 

- IS_FIRST_RUN=FALSE & RESUBMIT=0
  - This is the state before the second phase of the test runs
- RUN phase is set to PEND
  - This was testing with COMPARE missing, PEND and FAIL
- Incomplete ouput files in `run/`

I just went back to test and you are correct with **my current fix**, RUN set to PEND and COMPARE set to PEND still produce the original error with the above state.

I could change the new call to `_resetup_case(RUN_PHASE, reset=True)` as suggested by @billsacks which would always force the reset with the above conditions. Alternatively I could close this PR and continue the disscussion under #3940 and maybe determine a better fix for the long term.

@jgfouca @jedwards4b @billsacks Which path should we move forward with?

billsacks:
How about we discuss this, along with #3939 and #3940, on Wednesday's cime call?

billsacks:
Oops, I just realized that we meant to discuss @jasonb5 's latest comment specifically but didn't at today's cime call. But I'd say: based on the fact that we decided to move ahead with this PR, I think the new call should be changed to `_resetup_case(RUN_PHASE, reset=True)`. Of course, here I'm just supporting my own initial suggestion, so happy to hear from others on this :-)

jasonb5:
@billsacks I agree the new call should be `_resetup_case(RUN_PHASE, reset=True)`, I'll make the change.

billsacks (approved):
I'm happy with this now - thank you @jasonb5 !

PR #3925, opened 2021-04-13
Change behavior of check on DIN_LOC_ROOT in case_setup.py

### LINKED ISSUE #3892: Support for CESM containers - DIN_LOC_ROOT directory check
The CESM/CESM-Lab containers require a few minor modifications,including a minor change for convenience to how CIME handles the inputdata directory - currently, it fails if it doesn't exist, it seems we can simply change this to creating it when it doesn't exist without impacting existing systems, since their config entries already specify a shared input directory.  

(Changed this from two issues to one, since each will be handled by a different PR.)

briandobbins:
This PR seeks to change the behavior of the check in case_setup  
on the DIN_LOC_ROOT directory so that when it doesn't exist an 
attempt is made to create it.  This is primarily useful for containers 
on laptops/desktops, since the non-shared environment means 
there isn't usually a pre-existing directory.  Any supported machines 
will already have a valid entry for inputdata, so it shouldn't effect 
them.

Test suite:  None yet; would like input on which to run, if needed.
Test baseline: None; not answer changing
Test namelist changes: No; doesn't affect namelists
Test status: Doesn't impact model

Fixes #3892 

User interface changes?:  In a very minor way; previous fails where 
a DIN_LOC_ROOT directory didn't exist will now succeed if it 
can be created.

Code review:

### DIFF
--- a/scripts/lib/CIME/case/case_setup.py
+++ b/scripts/lib/CIME/case/case_setup.py
@@ -94,7 +94,7 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False, 
                 os.makedirs(din_loc_root, exist_ok=True)
             except OSError as e:
                 if e.errno == errno.EACCES:
-                    logger.info("Invalid permissions to create {}".format(din_loc_root)
+                    logger.info("Invalid permissions to create {}".format(din_loc_root))
             
         expect(not (not os.path.isdir(din_loc_root) and testcase != "SBN"),
                "inputdata root is not a directory or is not readable: {}".format(din_loc_root))
--- a/scripts/lib/CIME/case/case_setup.py
+++ b/scripts/lib/CIME/case/case_setup.py
@@ -83,10 +83,15 @@ def _case_setup_impl(case, caseroot, clean=False, test_mode=False, reset=False, 
     extra_machines_dir = case.get_value("EXTRA_MACHDIR")
     expect(mach is not None, "xml variable MACH is not set")
 
-    # Check that $DIN_LOC_ROOT exists - and abort if not a namelist compare tests
+    # Check that $DIN_LOC_ROOT exists or can be created:
     if not non_local:
         din_loc_root = case.get_value("DIN_LOC_ROOT")
         testcase     = case.get_value("TESTCASE")
+
+        if not os.path.isdir(din_loc_root):
+          logger.info("Making inputdata directory: {}".format(din_loc_root))
+          os.mkdir(din_loc_root)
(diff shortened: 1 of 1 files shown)


jedwards4b (approved):

lukaszlacinski:
`os.mkdir()` does not create a directory recursively, so it will fail in many cases. I would replace [the three lines](https://github.com/ESMCI/cime/blob/b2899395da8c1584884939c479a197226b25e802/scripts/lib/CIME/case/case_setup.py#L91) with one:
should rather be `os.makedirs(din_loc_root, exist_ok=True)`

briandobbins:
@lukaszlacinski Thanks very much for the catch!  I'll prep a new PR with that fix this morning and tag you and Jim.

briandobbins:
@jedwards4b, @lukaszlacinski 

My python and CIME knowledge is still pretty limited -- it looks like 'exist_ok' is only for Python3.  We're still supporting Python2, I think, right?  Is there a preferred / recommended way of doing this?  I see potential workarounds using exceptions, but wasn't sure what the right approach was here.

Do we support Python2 still?  I seem to recall recent discussions about this.  If so, I see there are several possible workarounds, but is one recommended / used elsewhere in CIME?

jedwards4b:
This seems like a trivial place to drop support for py2, just change the one line:

os.mkdirs(din_loc_root)

lukaszlacinski:
>  I see potential workarounds using exceptions, but wasn't sure what the right approach was here.
I think an exception has to be caught anyway in case a directory cannot be created because a regular file with the same name already exists.

rljacob:
@briandobbins you don't need to make a new PR.  Just make another commit to the branch for this PR.

briandobbins:
Apologies for the naive question, but per Jim's comment, if we're dropping support for python2 here, how do I just run the python3 regression tests?  The python2 one fails (as expected, since 'exist_ok' isn't valid), but that seems to trigger the cancelation of the python3 tests too.

The latest code does add a check on whether a directory can be created if it doesn't exist, and uses os.makedirs w/ exist_ok to create the full path.

jedwards4b:
Perhaps I wasn't clear in my comment - what I meant to say is that it is easy to maintain support for py2 here, I don't think that this should be the point where we break it.  When and if we drop py2 we will need to remove it from the github actions so that only py3 is tested.

lukaszlacinski:
Please change `os.makedirs(din_loc_root, exist_ok=True)` to `os.makedirs(din_loc_root)` to still support Python 2.

PR #3929, opened 2021-04-16
Add optional fields from atmosphere for calculation of surface stress

quantheory:
This PR adds two new options to env_run.xml:

 1. `ATM_FLUX_INTEGRATION_METHOD`
 2. `ATM_SUPPLIES_GUSTINESS`

If the first option is set to `"explicit"` (the default), there is
no change in the behavior of the model. If this option is set
to `"implicit_stress"`, the atmosphere model is required to
export two new fields, `wsresp` and `tau_est`. These fields
represent an approximate linearization of the response of
the atmosphere's boundary layer scheme to stress, and can
be used to perform an explicit solution to the stress in
surface flux calculations. The implementation of this method
in the E3SM atmosphere/land/ice models is currently on a
branch here:

https://github.com/E3SM-Project/E3SM/tree/quantheory/equilibrium-surface-fluxes

This change is motivated by oscillations in the surface winds
that can occur in both E3SM and CESM, which are discussed
in these two issues:

 - E3SM-Project/E3SM#4073
 - ESCOMP/CAM#342

These oscillations are largely due to instabilities arising from
the surface stresses over land at a coarse time step, and are
therefore reduced by the use of an implicit method.

The second option is a logical flag, which keeps the current
behavior if set to `FALSE`, and if `TRUE` adds an
additional "ugust" field to be exported by the atmosphere. This
change is currently only useful for E3SM, which uses the
variance in horizontal winds produced by CLUBB as part of
its gustiness parameterization.

Test suite: Have not run a real test suite, but E3SM F2010 cases
        with default settings run BFB with the version of the code I
        branched from.
Test baseline: 
Test namelist changes: 
Test status: bit for bit

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review: None yet!

### DIFF
--- a/src/drivers/mct/cime_config/config_component_e3sm.xml
+++ b/src/drivers/mct/cime_config/config_component_e3sm.xml
@@ -199,6 +199,17 @@
     </desc>
   </entry>
 
+  <entry id="ATM_SUPPLIES_GUSTINESS">
+    <type>logical</type>
+    <valid_values>TRUE,FALSE</valid_values>
+    <default_value>FALSE</default_value>
+    <group>run_flags</group>
+    <file>env_run.xml</file>
+    <desc>
+      Whether or not the atmosphere should supply an extra gustiness field.
+    </desc>
+  </entry>
+
   <!-- ===================================================================== -->
   <!-- component coupling options and frequencies -->
   <!-- ===================================================================== -->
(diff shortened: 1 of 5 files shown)


quantheory:
There's no need to merge this yet, since it goes with some features that haven't been added to E3SM (and haven't even been ported to CAM). But I did want some feedback on whether this is the right approach. Even if these exact mods aren't added, both models will probably need to export *some* extra information from the atmosphere to solve their respective wind oscillation issues.

mvertens:
@quantheory - since we are splitting the e3sm and cesm mct coupling infrastructure into separate repositories (including the shr_flux_mod.F90)  I am not sure it makes sense for me to review this since this will not be going into the cesm shr_flux_mod at this point. What are your thoughts?

PR #3930, opened 2021-04-19
Allow E3SM to use its versions of coupler and data,stub,x models.

rljacob:
Modify config_files.xml in config/e3sm to point to new locations
of the driver (MCT and MOAB) and MCT data,stub,x models within E3SM

Slightly modify buildlib.csm_share to use COMP_ROOT_DIR_CPL to find shr dir.

Still using cime/src/externals and cime/src/share

Test with e3sm_integration (build only) and an E3SM F-case (to test BFB)

[BFB]

### DIFF
--- a/config/e3sm/config_files.xml
+++ b/config/e3sm/config_files.xml
@@ -104,7 +104,7 @@
   <entry id="COMP_ROOT_DIR_CPL">
     <type>char</type>
     <values>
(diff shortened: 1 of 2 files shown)


rljacob:
Related PR in E3SM  https://github.com/E3SM-Project/E3SM/pull/4225

jedwards4b (approved):

fischer-ncar (approved):

PR #3932, opened 2021-04-21
Updates the coupler heat budget

bishtgautam:
Currently, the land ground heat flux is excluded from the coupler heat
budget. The code modifications add the land ground heat
flux in the coupler heat budget diagnostic.

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/src/drivers/mct/main/seq_diag_mct.F90
+++ b/src/drivers/mct/main/seq_diag_mct.F90
@@ -137,41 +137,42 @@ module seq_diag_mct
   integer(in),parameter :: f_hlatf     = 8     ! heat : latent, fusion, snow
   integer(in),parameter :: f_hioff     = 9     ! heat : latent, fusion, frozen runoff
   integer(in),parameter :: f_hsen      =10     ! heat : sensible
-  integer(in),parameter :: f_wfrz      =11     ! water: freezing
-  integer(in),parameter :: f_wmelt     =12     ! water: melting
-  integer(in),parameter :: f_wrain     =13     ! water: precip, liquid
-  integer(in),parameter :: f_wsnow     =14     ! water: precip, frozen
-  integer(in),parameter :: f_wevap     =15     ! water: evaporation
-  integer(in),parameter :: f_wsalt     =16     ! water: water equivalent of salt flux
-  integer(in),parameter :: f_wroff     =17     ! water: runoff/flood
-  integer(in),parameter :: f_wioff     =18     ! water: frozen runoff
-  integer(in),parameter :: f_wfrz_16O  =19     ! water: freezing
-  integer(in),parameter :: f_wmelt_16O =20     ! water: melting
-  integer(in),parameter :: f_wrain_16O =21     ! water: precip, liquid
-  integer(in),parameter :: f_wsnow_16O =22     ! water: precip, frozen
-  integer(in),parameter :: f_wevap_16O =23     ! water: evaporation
-  integer(in),parameter :: f_wroff_16O =24     ! water: runoff/flood
-  integer(in),parameter :: f_wioff_16O =25     ! water: frozen runoff
-  integer(in),parameter :: f_wfrz_18O  =26     ! water: freezing
-  integer(in),parameter :: f_wmelt_18O =27     ! water: melting
-  integer(in),parameter :: f_wrain_18O =28     ! water: precip, liquid
-  integer(in),parameter :: f_wsnow_18O =29     ! water: precip, frozen
-  integer(in),parameter :: f_wevap_18O =30     ! water: evaporation
-  integer(in),parameter :: f_wroff_18O =31     ! water: runoff/flood
-  integer(in),parameter :: f_wioff_18O =32     ! water: frozen runoff
-  integer(in),parameter :: f_wfrz_HDO  =33     ! water: freezing
-  integer(in),parameter :: f_wmelt_HDO =34     ! water: melting
-  integer(in),parameter :: f_wrain_HDO =35     ! water: precip, liquid
-  integer(in),parameter :: f_wsnow_HDO =36     ! water: precip, frozen
-  integer(in),parameter :: f_wevap_HDO =37     ! water: evaporation
-  integer(in),parameter :: f_wroff_HDO =38     ! water: runoff/flood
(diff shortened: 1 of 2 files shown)


bishtgautam:
An example of the new coupler budget output is [here](https://gist.githubusercontent.com/bishtgautam/b37080d8db909cd2b12a78134480065d/raw/4bff738778e361c8e0e8de6dfa4804163159196d/gistfile1.txt).

billsacks:
I don't know enough about the budget calculations and how they're used to be able to review this. Is the feeling in E3SM that this is a bug and that, to have correct budgets you need to include this new term? Or is this more a choice where you can go either way depending on how you want to think about the budgets? If it's a bug, then I appreciate your bringing it to our attention and we can pass it along to some of the scientists who may want to update CESM similarly.

In either case, though, it's probably best if this waits for the soon-upcoming split of the Fortran source code between CESM & E3SM: it looks like this will require a new field to be passed from the land, so wouldn't currently work in CESM.

billsacks (changes requested):
I'm marking this as requesting changes to track that this should wait until the cime split.

bishtgautam:
I believe this fixes a bug in the coupler heat budget. A new field, ground heat flux, needs to be sent to the coupler, and here is my PR in E3SM that passes this additional field from the land model. https://github.com/E3SM-Project/E3SM/pull/4237. Let me know if you want to discuss this potential bug on a Github issue (probably on CTSM repo) or on an email thread.

bishtgautam:
FYI, the annual error in the land heat budget with the fix reduced from 1.088 W/m^2 to -0.122 W/m^2 for a 1-year e3sm land-only run.

dlawrenncar:
Yes.  Need to know how the heat budget is used to understand this.  If trying to understand the coupled system heat budget, then it's not clear to me that removing the mean ground heat flux makes sense.  Within coupled system, don't we want to know how much net energy is entering the soil?  In perfect model, that would be zero or it would be matched by energy exiting the soil in via runoff.  So, if the mean ground heat flux is non-zero in long-term mean, then that indicates an energy leak, which we know exists.  Subtracting off the ground heat flux, then, is more of an indication that the surface energy budget closes, which we try to ensure within the land model error checks, but then you no longer are learning about energy leaks.

bishtgautam:
@dlawrenncar, I tried to close the following land surface energy budget equation in the coupler at each coupler time step:
```
Rnet = S_down - S_up + L_down - L_up = LE + SH + G
```
The land model heat budget balance includes G, but the coupler budget does not include G. The code changes here add G in the coupler heat budget.

dlawrenncar:
Yes.  That is obviously the surface energy balance budget.  The question is
whether or not that is what we are trying to track with the heat budget
balance in the coupler.  If the rationale for the heat budget in the
coupler is to assess whether or not the model is correctly solving the
surface energy budget, then you clearly want to include the ground heat
flux term.  If the goal is to track whether or not the land is gaining or
losing heat either due to a lack of equilibrium or an energy sink or source
due to model formulation (i.e., not tracking heat exchange with streams) or
due to a bug (e.g., lack of conservation of energy during phase change),
then I think you (maybe) do not want to include the ground heat flux.

rljacob:
Let me try:  the goal of the budgets in the coupler is to track energy **that flows through the coupler**.   What physicaly is this "ground heat flux" ?

bishtgautam:
> If the goal is to track whether or not the land is gaining or
losing heat either due to a lack of equilibrium or an energy sink or source
due to model formulation

Okay, I see your point. I believe the coupler heat "budget" is only the conservation of heat fluxes. For the coupler to check whether the whole coupled model (or any component) is gaining/losing heat, the coupler would additionally need information regarding the total amount of heat stored in each component.

rljacob:
The conservation of fluxes between components.  Where does G go?

rljacob:
And yes the coupler can't check conservation of the whole system without a lot more info.

You could add a budget calculation to ELM that includes G.  Output to the lnd.log file.

bishtgautam:
G is the heat that goes into the soil and is computed by the land model. Since no other component needs G, it does not ***flow*** through the coupler. But G needs to be accounted for in the coupler heat diagnostics.

rljacob:
Into the soil and....out the bottom of the soil column?

bishtgautam:
>You could add a budget calculation to ELM that includes G. Output to the lnd.log file.

The land model heat budget already includes G.

bishtgautam:
>Into the soil and....out the bottom of the soil column?

The heat enters the top of the soil and there is a no-flux boundary condition at the bottom of the soil column.

golaz:
Interesting conversation. I had assumed initially that there was a bug in the coupler land heat budget, but that may not be the case then. The coupler heat budget for the land should average out to zero (over the long term) if:

1. The model is in equilibrium.
2. The land is perfectly conserving energy.

Assuming (1) and (2) are satisfied, would the current coupler land heat budget show a bottom line value of 0?

rljacob:
I think so but it really depends on what exactly G is.  Gautam and I are conversing offline (in Slack) and I am no clearer.

oksanaguba:
@bishtgautam could I please ask, what exactly is G in
Rnet = S_down - S_up + L_down - L_up = LE + SH + G ?
I assume LE is the latent term  L_down - L_up? I tried to follow this discussion + your PR but still have a question about G. Thanks!

golaz:
From https://www.researchgate.net/figure/Linking-the-land-surface-energy-and-water-balances-through-E-evaporation-and-energy_fig1_226620656

See middle panel: the amount of energy going into the ground. G only exists in the land model. The coupler only knows about radiation, latent and sensible.

image

rljacob:
Closing this.  If these changes are to be done, will be made directly in E3SM repo and its copy of the coupler.

oksanaguba:
@bishtgautam following plots above, do you know how the latent heat flux to atm. is computed in the land component? or maybe it is computed in the coupler from water mass? Also, would you know how big this flux is in the model in W/m^2? What is discharge, Q, above? The linked paper does not give details and 'discharged + land' is usually used for mass fluxes (from quick googling), not for energy. thanks!

PR #3935, opened 2021-04-21
Add support for WAV as primary component

### LINKED ISSUE #3934: Active WAV component not recongized as primary component
In a new compset I created with an active WAV component and data ATM, the pe layout defaults to the datm pe specification file: (`src/drivers/mct/cime_config/config_pes.xml`)

I *think* this could be fixed by adding:
```
        elif progcomps["WAV"]:
            primary_component = spec["WAV"]
```
before the `else`  in  the following `if` statement:
https://github.com/ESMCI/cime/blob/44b241d263e3d2ace62f5340a1110f7174ca79e5/scripts/lib/CIME/case/case.py#L520-L573

sbrus89:
This PR allows WAV to be identified as the primary component for compsets with WAVEWATCH III, such as:
`2000_DATM%IAF_SLND_SICE_SOCN_SROF_SGLC_WW3_SIAC_SESP`

Test suite: 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3934 

User interface changes?: 

Update gh-pages html (Y/N)?:

Code review:

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -566,6 +566,9 @@ class Case(object):
         elif progcomps["ROF"]:
             # This is a "R" compset
             primary_component = spec["ROF"]
+        elif progcomps["WAV"]:
+            # This is a "V" compset
+            primary_component = spec["WAV"]
         else:
             # This is "A", "X" or "S"
             primary_component = "drv"


sbrus89:
I'm not sure if this is how letter cases are officially claimed. I'm happy change the comment that says this is a V" case.

PR #3939, opened 2021-04-22
Reset a test that is rerun after a run failure

### LINKED ISSUE #3933: ERI doesn't rerun correctly when using the --retry flag to create_test
When running create_test with the `--retry` option set to 1 or more: If an ERI test fails in the last run, STOP_N is set incorrectly in the retried run. If you instead rerun it manually, by running `./case.submit`, it seems to rerun correctly.

I haven't investigated carefully yet, but I think the problem is that `--retry` sets `use_existing`, which in turn changes a FAILed RUN phase into a PEND. I think the issue with that is that it changes the behavior of the call to `_resetup_case(RUN_PHASE)` in `system_tests_common:run`, since the logic in `_resetup_case` is that PENDing phases are *not* reset.

I will do a bit more investigation and testing, then submit a fix. This may impact other test types besides just ERI.

billsacks:
When running create_test with the --retry option set to 1 or more: If a
test (ERS, ERI and possibly others) failed in the last run, some xml
values would be set incorrectly in the retried run. This is because
--retry sets use_existing, which in turn changes a FAILed RUN phase into
a PEND. Then, in the call to _resetup_case(RUN_PHASE) from
system_tests_common:run, the test reset wasn't done, since PENDing
phases are not reset.

This commit solves this problem by adding a comment that the now-PENDing
phase is actually being RERUN, and checking for that in _resetup_case.

There may be other situations where we should add this RERUN comment,
but this change fixes my original issue, and it wasn't obvious to me
what, if anything, else should be changed from looking through the
code (especially test_scheduler.py).

Test suite:
- Tested with ERS_Ln3 with the diff in https://github.com/ESMCI/cime/issues/3933#issuecomment-824387449, on both my Mac (no batch system) and cheyenne (batch system): in both cases, the retried run failed with `stop_n value 1 too short` before this fix, but with this fix got all the way to the end 

Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes #3933 

User interface changes?: N

### DIFF
--- a/config/config_tests.xml
+++ b/config/config_tests.xml
@@ -427,6 +427,15 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
     <DOUT_S>FALSE</DOUT_S>
   </test>
 
+  <test NAME="TESTRUNFAILRESET" INFRASTRUCTURE_TEST="TRUE">
+    <DESC>For testing infra only. Insta-fail run step, also testing that settings get reset upon rerun.</DESC>
+    <INFO_DBUG>1</INFO_DBUG>
+    <STOP_OPTION>ndays</STOP_OPTION>
+    <STOP_N>11</STOP_N>
+    <CHECK_TIMING>FALSE</CHECK_TIMING>
+    <DOUT_S>FALSE</DOUT_S>
+  </test>
+
   <test NAME="TESTRUNSTARCFAIL" INFRASTRUCTURE_TEST="TRUE">
     <DESC>For testing infra only. Insta-fail st archive test with exception.</DESC>
     <INFO_DBUG>1</INFO_DBUG>
--- a/scripts/lib/CIME/SystemTests/system_tests_common.py
+++ b/scripts/lib/CIME/SystemTests/system_tests_common.py
@@ -772,7 +772,9 @@ class TESTRUNFAILRESET(TESTRUNFAIL):
         # upon a rerun
         self._case.set_value("STOP_N", stop_n + 1)
 
-        super().run_indv(suffix=suffix, st_archive=st_archive, submit_resubmits=submit_resubmits)
+        super(TESTRUNFAILRESET, self).run_indv(suffix=suffix,
+                                               st_archive=st_archive,
+                                               submit_resubmits=submit_resubmits)
 
 class TESTRUNFAILEXC(TESTRUNPASS):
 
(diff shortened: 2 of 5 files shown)


billsacks:
@jasonb5 I'm assigning you as a reviewer since you've been thinking about _resetup_case recently.

billsacks:
scripts_regression_tests pass on cheyenne

billsacks:
I'll also add: I considered a few different ways of solving this problem, but anything else I could think of was much more complex than this solution. I'm not tied to this solution, though, if others see a cleaner way to fix the problem.

jedwards4b (approved):

jgfouca:
@billsacks , is this functionality already tested in scripts_regression?

billsacks:
@jgfouca Thanks for giving me a gentle prod to add a test for this :-)   I have now. I have confirmed that `O_TestTestScheduler:test_d_retry` fails with my testing-related commits (1faa36f and c15311a) cherry-picked onto master, but passes with the full changes in this branch.

At first I tried changing the existing TESTRUNFAIL class (rather than adding a new class), but this caused a failure in test_c_use_existing, I think because that test sets the SUBMIT phase to PEND for the second run:

https://github.com/billsacks/cime/blob/c15311a7d99dcd9de1cb5551d6e546c6cefd0b8c/scripts/tests/scripts_regression_tests.py#L1484-L1487

which then must trigger a different flow that doesn't end up resetting the case. At first I got worried that a flow like that might happen in reality. My current not-very-solid sense is that it won't – i.e., it wouldn't typically happen that the RUN fails and then the SUBMIT phase is set to PEND – but if others think it will, then some more changes are needed in this PR to handle that possibility.

jgfouca (approved):

jasonb5 (approved):
LGTM

billsacks:
If people have time to give some thought to #3940 early next week – or maybe during next week's CIME call – then this PR can wait to be merged until we decide what, if anything, to do about those thoughts. However, I don't want to let this linger for too long (more than a week or two anyway), since it is impacting CTSM testing.

billsacks:
Based on cime telecon discussion, we are going to bring this in (*not* the more general suggestions in #3940 ). @jgfouca please feel free to merge at your convenience or give me the go-ahead to merge myself.

PR #3948, opened 2021-04-28
change shebang to explict python3

### LINKED ISSUE #3919: Charting a path towards dropping python2 support
This question came up in yesterday's CESM Software Engineering Group meeting (raised by @gold2718 , also relevant to @alperaltuntas ): We have some python code coming into CESM that will assume python3. Some of this code will be invoked by cime scripts (buildnml / buildlib), so will run under whatever python version is used to invoke the cime scripts, as controlled by your environment together with the shebang lines in the various cime scripts. Our feeling in CESM is that it is reasonable at this point to require python3, since python2 support ended over a year ago.

**I guess the first question here is: Do others (e.g., in E3SM) agree that it's reasonable to require python3 moving forward?**

Currently, our shebang lines are `#!/usr/bin/env python`. My understanding (based on https://stackoverflow.com/questions/6908143/should-i-put-shebang-in-python-scripts-and-what-form-should-it-take, https://www.python.org/dev/peps/pep-0394/, etc.) is that that is correct usage when your python code is compatible with python2 and python3. However, some systems (like cheyenne) have `python` pointing to `python2`, even if you have loaded a python3 module, and it may be a while before we can assume `python` always refers to `python3`.

So I'd like to start a discussion about how we want to proceed here. I'm going under the assumption that all users will have python3 installed on their system, so the question is: **How do we ensure that the cime scripts are run under python3 rather than python2?**

Two possible solutions came up in our CSEG discussion:

1. Change our shebang lines to `#!/usr/bin/env python3`. My understanding (e.g., from the links a couple of paragraphs above) is that this is a generally recommended solution when you require python3. I'm not 100% sure that this is 100% reliable in all situations, though: it sounds to me like systems *should* define a python3 executable, but there may be some that don't.

2. Instruct users to use a python environment where `python` points to a python3 executable. From PEP 0394, it sounds like this will be the case if people use a python virtual environment. I have been wary of requiring use of a python virtual environment, because it feels like something that might make porting to a new machine more difficult for non-tech-savvy users, and might be hard for users to remember even on systems with tech-savvy sys admins: Users would need to remember to launch the right python environment before invoking any cime scripts (create_newcase, case.build, etc., etc., etc.).

My gut feeling leans towards (1), because part of the value of cime has been that it manages the environment for you, so (2) feels like it might add too much friction for the use of cime. But I'd like to hear thoughts from others, especially others who have more experience than I do with python system configuration & virtual environments (which is a low bar to be honest). One possible advantage of (2) is that it could help give a path towards requiring a limited number of python packages outside the standard library (for CESM, we've talked about yaml, numpy, and either scipy.io.netcdf or netcdf4).

Thoughts?

jedwards4b:
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes #3919 

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/.github/workflows/srt.yml
+++ b/.github/workflows/srt.yml
@@ -18,7 +18,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [3.8, 3.9, 3.10]
+        python-version: [3.8, 3.9]
     env:
       CC: mpicc
       FC: mpifort
--- a/.github/workflows/srt.yml
+++ b/.github/workflows/srt.yml
@@ -18,14 +18,14 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        python-version: [2.7, 3.8]
+        python-version: [3.8, 3.9, 3.10]
     env:
       CC: mpicc
       FC: mpifort
       CXX: mpicxx
       CPPFLAGS: "-I/usr/include -I/usr/local/include"
       # Versions of all dependencies can be updated here
-      PNETCDF_VERSION: pnetcdf-1.12.1
+      PNETCDF_VERSION: pnetcdf-1.12.2
       NETCDF_FORTRAN_VERSION: v4.5.2
       NETCDF_C_PATH: /usr
       NETCDF_FORTRAN_PATH: ${HOME}/netcdf-fortran
--- a/scripts/Tools/archive_metadata
+++ b/scripts/Tools/archive_metadata
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 """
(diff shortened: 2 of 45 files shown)


rljacob:
Just a thought:  should we put off dropping python2 until cime6 ?

jedwards4b:
@rljacob I can't think of a good reason to delay this.  Everyone should have a compatible python3 already available.

billsacks (approved):
Awesome, thanks a lot for just going ahead and taking care of this @jedwards4b .

fischer-ncar (approved):

rljacob:
I have no objections.

jgfouca (approved):

PR #3943, opened 2021-04-27
Rename GLC Greenland grid from "gland" to "gris"

billsacks:
We are renaming both the grid long name and alias to "gris". For now we
are maintaining backwards compatibility with old
aliases ("_gl4"/"_gl20").

Test suite: Ran a subset of tests from the aux_glc test list:
```
ERS_Ly11.f09_g17_gris20.T1850Gg.cheyenne_gnu
ERS_Vnuopc_Ly11.f09_g17_gris20.T1850Gg.cheyenne_gnu
ERS_Ly7.f09_g17_gris4.T1850Gg.cheyenne_intel
ERS_Vnuopc_Ly7.f09_g17_gris4.T1850Gg.cheyenne_intel
ERI.f10_f10_mg37.I1850Clm50SpG.cheyenne_gnu.cism-test_coupling
ERI_Vnuopc.f10_f10_mg37.I1850Clm50SpG.cheyenne_gnu.cism-test_coupling
SMS_D.T31_g37_gris20.I1850Clm50SpG.cheyenne_gnu.cism-test_coupling
SMS_Vnuopc_D.T31_g37_gris20.I1850Clm50SpG.cheyenne_gnu.cism-test_coupling
```

Test baseline: cism2_1_78
Test namelist changes: Yes, due to new grid name
Test status: bit for bit

Fixes none

User interface changes?: N

Update gh-pages html (Y/N)?: N

**Depends on cism changes in https://github.com/ESCOMP/CISM-wrapper/pull/58**

PR #3954, opened 2021-05-05
add this back for mct CLM_USRDAT

### LINKED ISSUE #3905: USRDAT resolution issues for both MCT and NUOPC drivers
The following test is failing for me in ctsm5.1.dev030

SMS_D_Lm1_Mmpi-serial.CLM_USRDAT.I1PtClm50SpRs.cheyenne_intel.clm-USUMB

with cime5.8.39

It fails to submit with the following error...

> env_batch.xml appears to have changed, regenerating batch scripts
> manual edits to these file will be lost!
> 
>     Cannot download file since it lives outside of the input_data_root '/glade/p/cesm/cseg/inputdata'
>     Cannot download file since it lives outside of the input_data_root '/glade/p/cesm/cseg/inputdata'
>     Cannot download file since it lives outside of the input_data_root '/glade/p/cesm/cseg/inputdata'
>     Cannot download file since it lives outside of the input_data_root '/glade/p/cesm/cseg/inputdata'
>     Cannot download file since it lives outside of the input_data_root '/glade/p/cesm/cseg/inputdata'
>     Cannot download file since it lives outside of the input_data_root '/glade/p/cesm/cseg/inputdata'
> ERROR: Could not find all inputdata on any server

check_input_data shows the following...

> ./check_input_data 
> Setting Environment OMP_STACKSIZE=1024M
> Setting Environment TMPDIR=/glade/scratch/erik
> Setting Environment MPI_TYPE_DEPTH=16
> Setting Environment MPI_IB_CONGESTED=1
> Setting resource.RLIMIT_STACK to  -1 from (307200000, -1)
> Loading input file list: 'Buildconf/ctsm.input_data_list'
> Loading input file list: 'Buildconf/cpl.input_data_list'
> Loading input file list: 'Buildconf/datm.input_data_list'
> Model datm missing file domain1 = 'null/null'
> Model datm missing file file1 = 'null/null'

DOMAIN files seem to be set correctly. But, the datm streams files are getting set like this in datm_in...

> &shr_strdata_nml
>   datamode = "CLMNCEP"
>   domainfile = "/glade/p/cesm/cseg/inputdata/lnd/clm2/PTCLMmydatafiles.c171024/1x1pt_US-UMB/domain.lnd.1x1pt_US-UMB_navy.171024.nc"
>   dtlimit = 1.5, 1.5, 1.5
>   fillalgo = "nn", "nn", "nn"
>   fillmask = "nomask", "nomask", "nomask"
>   fillread = "NOT_SET", "NOT_SET", "NOT_SET"
>   fillwrite = "NOT_SET", "NOT_SET", "NOT_SET"
>   mapalgo = "nn", "nn", "nn"
>   mapmask = "nomask", "nomask", "nomask"
>   mapread = "NOT_SET", "NOT_SET", "NOT_SET"
>   mapwrite = "NOT_SET", "NOT_SET", "NOT_SET"
>   readmode = "single", "single", "single"
>   streams = "datm.streams.txt.CLM1PT.UNSET 1999 -999 -999",
>       "datm.streams.txt.presaero.clim_2000 1 2000 2000",
>       "datm.streams.txt.topo.observed 1 1 1"
>   taxmode = "extend", "extend", "extend"
>   tintalgo = "linear", "linear", "lower"
>   vectors = "null"
> /

All of these values set to NOT_SET or UNSET are causing problems.

jedwards4b:
Add back domain for CLM_USRDAT for mct driver. 

Test suite: SMS_D_Lm1_Mmpi-serial.CLM_USRDAT.I1PtClm50SpRs.cheyenne_intel.clm-USUMB (All pass)
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3905 

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/config/cesm/config_grids.xml
+++ b/config/cesm/config_grids.xml
@@ -37,8 +37,8 @@
       <grid name="rof"       compset="DROF"        >rx1</grid>
       <grid name="rof"       compset="DROF%CPLHIST">r05</grid>
       <grid name="rof"       compset="XROF"        >r05</grid>
-      <grid name="glc"	     compset="SGLC"        >null</grid>
-      <grid name="glc"	     compset="CISM2"       >gland4</grid>
+      <grid name="glc"       compset="SGLC"        >null</grid>
+      <grid name="glc"       compset="CISM2"       >gland4</grid>
       <grid name="glc"       compset="XGLC"        >gland4</grid>
       <grid name="wav"       compset="SWAV"        >null</grid>
       <grid name="wav"       compset="DWAV"        >ww3a</grid>
@@ -1872,6 +1872,13 @@
       <file>domain.ocn.01col.ArcticOcean.20150824.nc</file>
       <desc>01col is a single-column grid for datm and POP:</desc>
     </domain>
+
+    <domain name="CLM_USRDAT">
+      <nx>1</nx> <ny>1</ny>
+      <file>$DIN_LOC_ROOT/share/domains/domain.clm/domain.lnd.${CLM_USRDAT_NAME}_navy.nc</file>
+      <desc>user specified domain - only valid for DATM/CLM compset</desc>
+    </domain>
+
     <domain name="1x1_numaIA">
       <nx>1</nx> <ny>1</ny>
       <file grid="atm|lnd">$DIN_LOC_ROOT/share/domains/domain.clm/domain.lnd.1x1pt-numaIA_navy.110106.nc</file>
--- a/config/cesm/config_grids.xml
+++ b/config/cesm/config_grids.xml
@@ -37,8 +37,8 @@
       <grid name="rof"       compset="DROF"        >rx1</grid>
       <grid name="rof"       compset="DROF%CPLHIST">r05</grid>
       <grid name="rof"       compset="XROF"        >r05</grid>
-      <grid name="glc"       compset="SGLC"        >null</grid>
-      <grid name="glc"       compset="CISM2"       >gland4</grid>
+      <grid name="glc"	     compset="SGLC"        >null</grid>
+      <grid name="glc"	     compset="CISM2"       >gland4</grid>
       <grid name="glc"       compset="XGLC"        >gland4</grid>
       <grid name="wav"       compset="SWAV"        >null</grid>
       <grid name="wav"       compset="DWAV"        >ww3a</grid>


ekluzek (changes requested):
We may need to chat about this. But, I actually do NOT want this to come in. We never used these generically named files and when I asked you to remove them I was just asking you to remove unused code. This isn't the problem that I'm seeing. I can't find the issue I made regarding this right now, but when I do I'll point it out.

jedwards4b:
@ekluzek in your issue #3905 the problem is that ATM_GRID and LND_GRID are unset - this is the code that causes them to be set.  It's done in grids.py as part of create_newcase.

ekluzek:
Ahhh, OK. You are right.  Of course (as you often are).  I was convinced you were wrong, but I had also tracked it down to _get_domains in grid.py which corresponds to the XML you are pointing out. So this does now totally make sense.

I really don't want the domain file set though, but it looks like the block doesn't need it as far as I can see at this point. But, that just means removing that one line. And I suggest adding a comment to say you have to set it by hand in the case with xmlchange.

config/cesm/config_grids.xml
```diff
@@ -37,8 +37,8 @@
       <grid name="rof"       compset="DROF"        >rx1</grid>
       <grid name="rof"       compset="DROF%CPLHIST">r05</grid>
       <grid name="rof"       compset="XROF"        >r05</grid>
-      <grid name="glc"	     compset="SGLC"        >null</grid>
-      <grid name="glc"	     compset="CISM2"       >gland4</grid>
+      <grid name="glc"       compset="SGLC"        >null</grid>
+      <grid name="glc"       compset="CISM2"       >gland4</grid>
```
billsacks, line 41: This change (while a good fix of whitespace) is going to cause a conflict with the changes in #3943 . Can you please back this out here and I'll fix the whitespace in that PR?

jedwards4b: This PR has correct whitespace in it.  If there is a conflict the issue should be addressed in #3943.

billsacks: I have pushed a whitespace fix to #3943 . The problem is that, even with the same whitespace fix in that PR, merging these two branches will produce a merge conflict. Since resolving merge conflicts can be error-prone, I was asking if we can preemptively solve the problem by backing out the whitespace change here and just letting it be fixed when #3943 is merged. If you give me your okay, I'm happy to make that change myself. The end result will be the same, but the merge conflict will be avoided.

jedwards4b: sure, go ahead

billsacks: Done; thanks

ekluzek (approved):
Ok. Jim is right this is what's needed. And I don't see anything else needed in cime for the generic NUOPC case.

PR #3955, opened 2021-05-05
[WIP] Allow building csm_share and gptl in different locations

rljacob:
Add some "if model=" logic to allow building of csm_share and gptl when they are in a different location
in E3SM source.

Test suite:  Built an E3SM F-case with cime/src/share moved.
Test baseline: 
Test namelist changes:  NONE
Test status: BFB

### DIFF
--- a/src/build_scripts/buildlib.csm_share
+++ b/src/build_scripts/buildlib.csm_share
@@ -47,11 +47,22 @@ def buildlib(bldroot, installpath, case):
 ###############################################################################
     gmake_args = get_standard_makefile_args(case, shared_lib=True)
     comp_interface = case.get_value("COMP_INTERFACE")
+    cime_model = case.get_value("MODEL")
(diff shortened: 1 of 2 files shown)


rljacob:
@fischer-ncar and @jedwards4b I'm not sure if this is what I should do or if E3SM should just have its own version of these build scripts somewhere else.  Looks like config_files.xml can point to them anywhere.

Note there is already some "if model" logic in buildlib.pio

rljacob:
I guess I could make a "build_scripts" directory in our version of share and put the gptl and csm_share ones there.

rljacob:
Changing to a WIP.  May not even want this.

jedwards4b:
@rljacob I think that it would be better to move them in config_files.xml 
If and when you do move them I can clean up the pio buildlib.

rljacob:
Yes.  If we're going to have our own versions of the code, we should have our own build scripts too.

PR #3956, opened 2021-05-06
update this so that paths with % in them do not break cism build

jedwards4b:
The % symbol has a special meaning in Makefile and when we create tests with long compset names this
was causing a problem with the cism build not being triggered.  But Phony targets are phony so you can call them whatever you want.  This fixes the issue when called from cism buildlib with cmake GLCMakefile 

Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:


billsacks (approved):
Thank you very much for this, @jedwards4b !

PR #3959, opened 2021-05-10
Switch location of E3SM share buildlib scripts

rljacob:
Switch locaion of E3SM build scripts in config_files
for csm_share and gptl.

Test suite:  e3sm_integration
Test namelist changes: none
Test status: BFB


rljacob:
E3SM master is already using this branch.  This must be merged to cime master before updating E3SM to use cime master.

PR #3967, opened 2021-05-13
Update climate reproducibility tests

mkstratos:
This updates the three E3SM climate reproducibility tests (MVK, PGN, 
and TSC) to work with updated component names (eam and elm) and
time step namelist variables. Additionally, the MVK test uses
`seed_clock=True` so unique ensembles are generated.

Test suite: e3sm_atm_nbfb
Test baseline: 
Test namelist changes: 

User interface changes?: N

### DIFF
--- a/scripts/lib/CIME/SystemTests/pgn.py
+++ b/scripts/lib/CIME/SystemTests/pgn.py
@@ -44,7 +44,7 @@ PERTURBATIONS = OrderedDict([('woprt', 0.0),
 FCLD_NC = 'cam.h0.cloud.nc'
 INIT_COND_FILE_TEMPLATE = \
     "SMS_Ly5.ne4_ne4.FC5AV1C-04P2.eos_intel.ne45y.{}.{}.0002-{:02d}-01-00000.nc"
-INSTANCE_FILE_TEMPLATE = '{}eam_{:04d}.h0.0001-01-01-00000{}.nc'
+INSTANCE_FILE_TEMPLATE = '{}{}_{:04d}.h0.0001-01-01-00000{}.nc'
 
 
 class PGN(SystemTestsCommon):
@@ -54,6 +54,12 @@ class PGN(SystemTestsCommon):
         initialize an object interface to the PGN test
         """
         super(PGN, self).__init__(case)
+        if self._case.get_value("MODEL") == "e3sm":
+            self.atmmod = "eam"
+            self.lndmod = "elm"
+        else:
+            self.atmmod = "cam"
+            self.lndmod = "clm"
 
     def build_phase(self, sharedlib_only=False, model_only=False):
         ninst = NUMBER_INITIAL_CONDITIONS * len(PERTURBATIONS)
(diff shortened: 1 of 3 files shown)


rljacob (changes requested):
Wherever the code refers to "eam" or "elm", add some logic to possibly set that to "clm" or "cam" so that CESM could also use these tests.

You could use case.get_value("MODEL") which will return "e3sm" or "cesm".

scripts/lib/CIME/SystemTests/pgn.py
```diff
@@ -44,8 +44,7 @@
 FCLD_NC = 'cam.h0.cloud.nc'
 INIT_COND_FILE_TEMPLATE = \
     "SMS_Ly5.ne4_ne4.FC5AV1C-04P2.eos_intel.ne45y.{}.{}.0002-{:02d}-01-00000.nc"
-# FIXME: should 'cam' be 'atm' now?
-INSTANCE_FILE_TEMPLATE = '{}cam_{:04d}.h0.0001-01-01-00000{}.nc'
+INSTANCE_FILE_TEMPLATE = '{}eam_{:04d}.h0.0001-01-01-00000{}.nc'
```
singhbalwinder, line 47: This is a file name template which still uses `cam` as it was created in the older version of E3SM. This should not be changed otherwise the test won't be able to find the files.

singhbalwinder (changes requested):
Thanks a lot for fixing these! I am planning to revive these tests soon.

scripts/lib/CIME/SystemTests/pgn.py
mkstratos: > 
> 
> Wherever the code refers to "eam" or "elm", add some logic to possibly set that to "clm" or "cam" so that CESM could also use these tests.
> 
> You could use case.get_value("MODEL") which will return "e3sm" or "cesm".

As @singhbalwinder mentioned above, the current baseline is "cam" for these tests in E3SM. Should I add a switch for that as well so the tests can be compared across model versions? Something as simple as:

```python
if any(["cam" in _ for _ in os.listdir(base_dir)]):
    baseline_component = "cam"
else:
    baseline_component = "eam"
```

rljacob: @singhbalwinder a new baseline has to be created and it should use eam histories.  One reason we stopped running this test was because there WAS a climate changing PR and no one knew how to update the baseline.

singhbalwinder: Ok, I will create the baselines with eam history files. I am currently looking into this test and will find out how to update baselines. After that, I will try reviving it in our test suites again.

singhbalwinder: Sorry, I misread the code. This change looks good to me. That is, it is okay to change this string to `eam` vs. `cam` based on the host model (E3SM or CESM).

The code which actually forms the initial condition file name is on lines [93](https://github.com/ESMCI/cime/pull/3967/files#diff-885e3f4da13b63c9bb4a447c2b313c11a37fc0ec731992b555cde2c7d8eed615R93) and [94](https://github.com/ESMCI/cime/pull/3967/files#diff-885e3f4da13b63c9bb4a447c2b313c11a37fc0ec731992b555cde2c7d8eed615R94). These lines are not modified in this PR, so we should be good with this change.

singhbalwinder (approved):
Looks great!

jgfouca:
@rljacob , do you approve of this PR?

scripts/lib/CIME/SystemTests/pgn.py
```diff
@@ -44,8 +44,7 @@
 FCLD_NC = 'cam.h0.cloud.nc'
 INIT_COND_FILE_TEMPLATE = \
     "SMS_Ly5.ne4_ne4.FC5AV1C-04P2.eos_intel.ne45y.{}.{}.0002-{:02d}-01-00000.nc"
```
rljacob, line 46: @singhbalwinder why run FC5AVIC-04P2?   Does that compset still exist?

rljacob (approved):

scripts/lib/CIME/SystemTests/pgn.py
rljacob: Actually I'm not sure what this TEMPLATE variable is doing.  "eos" is also a machine that doesn't exist anymore.

rljacob:
@jgfouca you can merge this.

scripts/lib/CIME/SystemTests/pgn.py
singhbalwinder: @rljacob : The TEMPLATE variable is used for generating the name of the initial condition files. The initial conditions are generated using FC5AVIC-04P2 on eos machine. That is why you see both the old compset name and the machine name in the TEMPLATE variable. Since this is just an initial condition file, the compset used doesn't matter. More specifically, it should not make a difference for these particular tests. We can still run these tests using new v2 compsets.

PR #3968, opened 2021-05-14
Adds CASE_HASH to env_case

### LINKED ISSUE #3944: Create hash at create_newcase time.
To help uniquely identify cases in experiment data bases, create a hash using:

- the create_newcase command arguments,
- the date and time,
- the local host name
- the user id.

Store the value of this hash in env_case.xml.  Only way to change it is re-create the case.

This should be enough to create unique hashes for every case created with cime.

jasonb5:
Adds CASE_HASH to env_case.xml. The hash is made up of 
create_newcase arguments, date/time, user, and hostname. This
is meant to be finer grain than case-group for case databases.

Test baseline: n/a
Test namelist changes: n/a
Test status: bit for bit

Fixes #3944 

User interface changes?: n/a

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -5,7 +5,7 @@ All interaction with and between the module files in XML/ takes place
 through the Case module.
 """
 from copy import deepcopy
-import glob, os, shutil, math, six, time
+import glob, os, shutil, math, six, time, hashlib
 from CIME.XML.standard_module_setup import *
 #pylint: disable=import-error,redefined-builtin
 from six.moves import input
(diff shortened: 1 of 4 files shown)


rljacob:
Now if there are several submissions from the same CASEDIR, they will all have the same hash and the provenance system can identify them as being from the same case.  

But I'm now wondering what we should do if a user needs to re-run create_newcase to continue the same simulation.  Maybe they moved it to a different machine.   @sarats what do you think?

jedwards4b (changes requested):
What is the exact intent here?  I generated a case and then ran create_clone to create a new case.  Both cases have the same CASE_HASH - I don't think that is the intent.

rljacob:
That's true, the hashes should be different then.

jasonb5:
Fixed it so anytime `Case.copy` is called a new hash is generated. This covers `create_clone` and if called explicitly via a script (more of a future case).

jedwards4b (approved):

PR #3974, opened 2021-05-19
Implementing GEOS-Chem chemistry in CESM2.1 (CESM2-GC)

### LINKED ISSUE #3366: Addition of a `srcExclude` file for selective compiling
As part of the coupling between CESM and GEOS-Chem, we propose adding a `srcExclude` file that would list any files that do not need to be compiled. For instance, when coupling the GEOS-Chem chemistry modules with CESM, the advection files from GEOS-Chem do not require to be compiled, as CESM performs the advection.

We intend to add two types of `srcExclude` files:
- Any directory in `Filepath` could have a **local `srcExclude`** file. This file would be folder-specific and contain only the **local name of files to exclude**. Any missing files from the list would be discarded.
- A **global `srcExclude`** with **full pathnames relative to the component's root folder**. This file could be used, for instance, to exclude unnecessary GEOS-Chem files. The GEOS-Chem folder would be unmodified from its Github repository, thus we do not want to add any local `srcExclude`.
 
These files could be read from `mkSrcfiles`. Any new `srcExclude` file would push the new content to an array of excluded files.

I want to open this up for discussion and I welcome any potential suggestions.

fritzt:
Hello! 

I have been working on implementing the GEOS-Chem (version 
13.0.0) chemistry module in CESM2.1.1. This new option is 
designated as CESM2-GC. The implementation of GEOS-Chem 
chemistry inside CESM2 aims to provide an additional chemistry 
option alongside CAM-Chem. 

This is the third PR out of three (CAM, CLM, cime). The main PR 
can be found at https://github.com/ESCOMP/CAM/pull/378.

This implementation comes with a few modifications in cime and 
CLM. First, we wanted to have the option to let GEOS-Chem 
compute its own dry deposition velocities. This requires the coupler 
to pass land types from CLM to CAM. We have modified CLM, 
cime, and CAM accordingly. Additionally, not all files in the 
GEOS-Chem subdirectories should be compiled. To deal with 
this, we introduced a feature that can exclude files from the 
compilation process (https://github.com/ESMCI/cime/issues/3366).
Finally, this PR adds new Henry coefficients and oxidation
factors in `seq_drydep_mod.F90` for GEOS-Chem species. 
Some species overlap between CAM-Chem and GEOS-Chem, 
but still have different Henry coefficients. 

Test baseline: 
Test namelist changes: 

Fixes #3366

User interface changes?: No


jedwards4b:
@fritzt what is the status of this PR? Ready for review?

fritzt:
Hi @jedwards4b! Yes, this is ready for review.

scripts/Tools/mkSrcfiles
```diff
@@ -20,37 +20,103 @@ if ( open(FILEPATH,"< Filepath") ) {
...
+    # Filepath generated for CAM by CIME.
+    if ( (-e "$dir/.exclude") && (open(EXCLUDE,"$dir/.exclude") ) ) {
+        #@excludes = <EXCLUDE>;
+        push @excludes, <EXCLUDE>;
+        close( EXCLUDE );
+        $excludeDir = $dir;
+        $foundExclude = 1;
+        print "Found .exclude in: $excludeDir\n";
+    }
+#    if ( (-e "$dir/../.exclude") && (open(EXCLUDE,"$dir/../.exclude") ) ) {
```
jedwards4b, line 42: Please remove commented code.

src/drivers/mct/shr/seq_drydep_mod.F90
```diff
@@ -753,6 +1344,93 @@ subroutine seq_drydep_init( )
              test_name = 'CO'  ! this is just a place holder. values are set in drydep_fromlnd
           case( 'NH4NO3' )
              test_name = 'HNO3'
+          ! Below add GEOS-Chem species or mappings
```
jedwards4b, line 1347: Can we do this with a namelist option instead of code that must be edited by the user?

jedwards4b (changes requested):
We are moving to a new coupler: [CMEPS](https://github.com/ESCOMP/CMEPS)
these changes will need to be repeated in that repository in order to be available to the development version of cesm.

fvitt (commented):
It looks like github is comparing against an old version of seq_drydep_mod.F90.  Is this because you mods are in cesm2.1?  The drydep module would updated for CESM2.2.

lizziel:
Hi @fvitt and @jedwards4b, I'm Lizzie from the [GEOS-Chem Support Team](http://wiki.seas.harvard.edu/geos-chem/index.php/GEOS-Chem_Support_Team). I'll be working on updates in this PR for Thibaud in the near future. From the conversation it looks like the primary action items for merge are:
1. Re-implement the addition of GEOS-Chem species and mappings via namelist
2. Make sure the merging branch is off of CESM2.2 rather than 2.1
3. Apply the same changes to CMEPS

It see some higher level questions as well:
- Is computation of drydep velocity in GEOS-Chem rather than CTSM an approach approved by both CAM-Chem and GEOS-Chem communities?
- Who will do the work of moving hard-coded data out of `seq_drydep_mod.F90` and into an external file for read? 

I can reach out to CAM-Chem folks to find out their thoughts on where drydep velocities are calculated. If they would prefer computation in CTSM I wonder if that change should happen before or after merging the initial work for GEOS-Chem in CESM.

Regarding the data move, from the conversation it sounds like this may have been picked up again by @fvitt. What is the status? Should it hold up the merge of submitted changes?




Let me know if I am missing anything and I'll follow-up with any questions I have on the action items.

jedwards4b:
Hi @lizziel I believe that it should be from cesm2.3 not cesm2.2.  I'll let @fvitt handle the hard questions.

ekluzek:
Yes, as @jedwards4b points out it should be in the latest development versions of CESM which means cesm2_3.

I can check on this, but I think overall the CTSM community is happy to let the CAM-Chem people working on this lead this effort. So if they think it should be in CAM so be it. The complexity here is that some of what you need to calculate drydep are details that are hidden in CTSM. CTSM is the expert in regard to vegetation for example. We want to construct this so that the component that has the most expert knowledge on the process in question communicates the least amount of information to the other component that doesn't have as much expert knowledge about it. Both models are going to change, doing it that way should be the best way to future proof it against changes that happen in either model.

fvitt:
I will let @lkemmons comment on which component should compute dry deposition velocities for GEOS-Chem.   

I started the work of moving hard-wired data from seq_drydep_mod to an input file some time ago.   I got distracted with many other projects.  Perhaps this will motivate me to pick it up again.

CESM is transitioning from MCT to CMEPS coupling (NUOPC) and the component shared code has been pull out of CIME.  Thus, changes to seq_drydep_mod should be done in share/cmeps (https://github.com/ESCOMP/CESM_share).

lkemmons:
I think it would be best to keep the dry deposition velocity calculation in CLM/CTSM, and for GEOS-Chem to use the existing deposition velocity code. Future CESM/GEOS-Chem users would be welcome to contribute to improving our existing scheme. 
Perhaps the first step is to implement reading the input file Francis @fvitt created (containing the Henry's Law constants) in place of the data statements in seq_drydep_mod.  
Then we should allow the mapping of species for dry dep in a namelist (as we do for MEGAN emissions). 
An alternative might be to add all the species needing dry dep to the Henry's Law input file, essentially providing any mapping in that file.  Though, I guess this may not be as efficient as it might be calculating the same velocity for many species and passing them from land to atmosphere, when they could just be mapped in CAM.
@lizziel if you would like to discuss this in a meeting, I (Louisa Emmons) am happy to do that.

ekluzek:
@lkemmons thanks for that comment. We just had a meeting this morning to discuss some of this. Overall I feel you are right, as this process sends the internals of CTSM to CAM, so I'm not sure it's a good way to go.  My main concern is future-proofing this capability based on changes that will happen inside of CTSM.

That said an issue we brought up this morning is that GEOS-Chem is being brought in a way that allows it's code to remain unchanged. For that to happen GEOS-Chem does need to do the calculation inside of CAM. So for the short term I think it's likely that this is the way we need to go for linking in GEOS-Chem.

But, I do think we need everyone on the same page about this so we likely do need another meeting that involves at least you and Francis and myself as well as @lizziel. @lizziel is going to be doing the long term support of this in CESM for the GEOS-Chem community, so our desire to make this work in a way that's sustainable for future changes to CTSM is directly relevant for her.

src/drivers/mct/shr/seq_drydep_mod.F90
```diff
@@ -35,10 +35,11 @@ module seq_drydep_mod
...
   ! !PRIVATE ARRAY SIZES
 
-  integer, private, parameter :: maxspc = 100              ! Maximum number of species
-  integer, public,  parameter :: n_species_table = 77      ! Number of species to work with
+  integer, private, parameter :: maxspc = 160              ! Maximum number of species
+  integer, public,  parameter :: n_species_table = 252     ! Number of species to work with
   integer, private, parameter :: NSeas = 5                 ! Number of seasons
-  integer, private, parameter :: NLUse = 11                ! Number of land-use types
+  integer, public,  parameter :: NLUse = 11                ! Number of land-use types
+  integer, public,  parameter :: NPatch = 80               ! Number of patch types
```
ekluzek, line 42: We are concerned about NLUse and NPatch being constants. In this context does NLUse need to be coordinated with CTSM land-use types? Or are they separate from CTSM? It looks like having NPatch as a constant is for the maximum number of different vegetation patches that might be in CTSM. If it's a max number for efficiency it would be good to have this have the ability to change as a non-crop simulation will only have up to 17, and actually crop simulations only actively use about 24.

lkemmons:
It looks everything is being worked out to satisfy the GEOS-Chem team (in a different issue).
@fvitt, perhaps we should open a new issue to implement the input file for Henry's Law constants, so that does not get lost. 
That will impact wet dep as well as dry dep.

lizziel:
I'm now looking a CMEPS tag `cmeps0.13.25` to figure out what updates in this PR need to be adapted. @mvertens, I think a meeting to go over this would be helpful.

The questions I have right now have looking through Thibaud's changes in this PR, the conversation thus far, and CMEPS code versus CIME:
1. It looks like CMEPS can be built with either GNU make or CMake. Are you transitioning to CMake only, and therefore should I not make any build updates for GNU?
2. As discussed in https://github.com/ESMCI/cime/issues/3366, CMake does not use mkSrcfiles. For other CMake projects such as GEOS-Chem in GEOSgcm we list files to exclude in the CMake build using the `REMOVE_ITEM` subcommand in the CMakeLists.txt in the directory containing GEOS-Chem. Do you want to do something similar for CESM, at least for the case of building with CMake? This would mean not having a dedicated exclude file since the list would be in CMakeLIsts.txt.
3. I cannot find the CMEPS equivalent file for seq_drydep_mod.F90. Is there one?
4. Given the work by @fvitt to create an input file for Henry's Law constants, should hard-coding these for use by GEOS-Chem be abandoned in favor of waiting for that update?
5. From discussion in https://github.com/ESCOMP/CTSM/pull/1475 it looks like we will not use NPatch and NLUse but make updates to pass all the lai for every vegetation type as a single field, which I think bypasses the need for them. Am I understanding this correctly?

ekluzek:
@lizziel let's put together another meeting to discuss this more. We need at least you @lizziel and also @fvitt, @mvertens, myself, and @billsacks would also like to be involved. Anyone else that should be added? @lkemmons do you want to be in this group as well? This will mainly be focused on what needs to be passed between components and how it needs to be done.

On the build issues CMEPS is automatically built as part of CESM from cime, so the build details aren't really an issue, unless you want to build GEOS-Chem standalone outside of CESM. We do build with many compilers: gnu, intel, pgi, nag, as well as many different machines. But, all of that is handled by cime.

The interface on the land side should be constructed to send the minimum amount of information in a form that will be stable for use even as the land model model changes. So I think we need to do some work to future-proof how this is done.

lkemmons:
I will be happy to join this meeting if I am available, but do not schedule
it around me.
Louisa

lizziel:
This PR is superseded by https://github.com/ESMCI/cime/pull/4180 and can be closed. Changes in this PR that are not in the new one are no longer applicable to this repository.

PR #3975, opened 2021-05-19
Fixes new_hash method

### LINKED ISSUE #3971: Calling new_hash fails on some systems where HOSTNAME is not defined
Method `new_hash` fails if HOSTNAME does not exists.

https://github.com/ESMCI/cime/blob/d224588ef7af63138d0f479398ef9ee8f0bed20b/scripts/lib/CIME/case/case.py#L1733-L1743

Should switch to `socket.getfqdn` which is already used in CIME.

jasonb5:
A call to new_hash would fail if the HOSTNAME environment variable
was not available. To fix this we moved to a more portable function 
(socket.getfqdn) to get the HOSTNAME, also did the same for USER
(getpass.getuser). Also updated tests to utilize unittest.mock.

Test baseline: n/a
Test namelist changes: n/a
Test status: bit for bit

Fixes #3971 

User interface changes?: n/a

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -1738,8 +1738,6 @@ directory, NOT in this subdirectory."""
         hostname = socket.getfqdn()
         user = getpass.getuser()
 
-        print(args, ctime, hostname, user)
-
         data = "{}{}{}{}".format(args, ctime, hostname, user)
 
         return hashlib.sha256(data.encode()).hexdigest()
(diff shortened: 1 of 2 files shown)


scripts/lib/CIME/case/case.py
```diff
@@ -1733,10 +1733,13 @@ def create(self, casename, srcroot, compset_name, grid_name,
     def new_hash(self):
         """ Creates a hash
         """
+        print(sys.argv)
```
jgfouca, line 1736: I think you can get rid of these prints.

jgfouca (changes requested):

scripts/lib/CIME/case/case.py
jasonb5: @jgfouca Removed the print statements.

jgfouca (approved):

PR #3977, opened 2021-05-21
Add chksum argument

### LINKED ISSUE #3467: check_input_data --chksum option to create_test
check_input_data has an option --chksum that is currently only invoked if a user explicitly runs it, 
this request is to add --chksum as an argument to create_test so that it is automatically invoked for each test in the suite.

jasonb5:
Adds --chksum argument to case.submit and create_test allowing 
users to verify checksums without explicitly calling 
check_input_data with --chksum.

Test baseline: n/a
Test namelist changes: n/a 
Test status: bit for bit

Fixes #3467 

User interface changes?: n/a

### DIFF
--- a/scripts/lib/CIME/tests/test_case.py
+++ b/scripts/lib/CIME/tests/test_case.py
@@ -1,11 +1,65 @@
 import os
 import sys
 import unittest
+from unittest import mock
+import tempfile
 
+from CIME.case import case_submit
 from CIME.case import Case
 from CIME import utils as cime_utils
 
 from . import utils
+
+class TestCaseSubmit(unittest.TestCase):
+
+    def test_check_case(self):
+        case = mock.MagicMock()
+
+        case_submit.check_case(case, chksum=True)
+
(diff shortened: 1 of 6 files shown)


jgfouca (approved):

scripts/lib/CIME/tests/test_case.py
```diff
@@ -203,7 +255,7 @@ def test_init(self, strftime, get_value, flush, init): # pylint: disable=unused-
     @mock.patch("sys.argv", ["/src/scripts/create_newcase"])
     def test_sub_relative(self, strftime, get_value, flush, init): # pylint: disable=unused-argument
         Case._force_read_only = False # pylint: disable=protected-access
-
+
```
jedwards4b, line 258: remove whitespace here.

jedwards4b (changes requested):
minor whitespace change request, otherwise looks good.

jedwards4b (approved):

PR #3980, opened 2021-05-24
[Obsolete] Add a multi-GPU-per-node configuration to Casper

sjsprecious:
Based on @jedwards4b 's suggestions, we added a new command line argument `ngpus_per_node` to the `create_newcase` script so that a user could specify the number of GPUs per node on Casper to run a GPU-lized subroutine. We also added the `NGPUS_PER_NODE` and `MAX_GPUS_PER_NODE` xml variables in order to make this configuration work properly.

This PR is an enhancement based on the previous PR https://github.com/ESMCI/cime/pull/3922.

An example of usage looks lie:
`./create_newcase --case $CASEROOT --mach casper --res f19_f19_mg17 --compset F2000climo --compiler pgi-gpu --ngpus_per_node 8`

Test suite: [aux_cam]
Test baseline: Cheyenne
Test status: [bit for bit]

Code review: @jedwards4b @johnmauff 

See an updated PR at: https://github.com/ESMCI/cime/pull/3989


jedwards4b:
@sjsprecious Please move this code to a branch on your fork and reissue the PR from the branch (not master).   Thanks

config/xml_schemas/config_machines.xsd
```diff
@@ -159,6 +160,8 @@
         <!-- MAX_TASKS_PER_NODE: maximum number of threads*tasks per
              shared memory node on this machine-->
         <xs:element ref="MAX_TASKS_PER_NODE" minOccurs="1" maxOccurs="4"/>
+        <!-- MAX_GPUS_PER_NODE: maximum number of GPUs per node on this machine-->
+        <xs:element ref="MAX_GPUS_PER_NODE" minOccurs="0" maxOccurs="4"/>
```
jedwards4b, line 164: Should maxOccurs here really be 1?  I see that MAX_TASKS_PER_NODE also has 4 but I think that this must be a mistake.

scripts/lib/CIME/case/case.py
```diff
@@ -1487,6 +1509,20 @@ def get_mpirun_cmd(self, job=None, allow_unresolved_envvars=True, overrides=None
         if self.get_value("BATCH_SYSTEM") == "cobalt":
             mpi_arg_string += " : "
 
+        ngpus_per_node = self.get_value("NGPUS_PER_NODE")
+        if ngpus_per_node > 0:
+            # generate a wrapper script under the case dir to set the device id for each MPI rank
+            casedir = self.get_value("CASEROOT")
+            with open (casedir+'/set_device_rank.sh', 'w+') as f:
```
jedwards4b, line 1516: This code seems like it may be pretty specific to casper.  Can we think of a way to do it that will be more adaptable to other systems?

src/drivers/mct/cime_config/config_component.xml
```diff
@@ -3003,6 +3003,23 @@
    <desc>TRUE=>turn on CPP variable COMPARE_TO_NUOPC</desc>
  </entry>
 
+ <entry id="NGPUS_PER_NODE">
```
jedwards4b, line 3006: Should this be implemented on a per component basis?  See the NTASKS and NTHRDS variables for example.

jedwards4b (commented):
I have a few questions and need you to reopen in a branch on your fork.

config/xml_schemas/config_machines.xsd
sjsprecious: Thanks @jedwards4b . I think it makes more sense to set `maxOccurs` to 1 but I just followed the setting for `MAX_TASKS_PER_NODE`. I could update it to 1 if it looks fine to you.

scripts/lib/CIME/case/case.py
sjsprecious: Thanks @jedwards4b . Could you provide more details about which part of this code do you think is too specific to Casper?

config/xml_schemas/config_machines.xsd
jedwards4b: Yes, please change to 1.

scripts/lib/CIME/case/case.py
jedwards4b: I am referring to the set_device_rank.sh script and the way that it is included in the mpi_arg_string.  I think that it's okay as is for now, but we may need to revisit this as we extend gpu support to more systems.

rljacob:
Tagging @amametjanov since this is adding general capability to CIME to describe node architecture with GPUs.

I know we've used current CIME to run cases on Summit that use the CPU and GPU but I'm not sure what exactly CIME is doing.

src/drivers/mct/cime_config/config_component.xml
sjsprecious: Thanks @jedwards4b . I have changed this in my work branch and it will be reflected in another PR.

scripts/lib/CIME/case/case.py
sjsprecious: Thanks for the details, @jedwards4b. Now I see your point and I agree that this setting may be too specific. Since Casper is the only machine I have the access with GPU, I could not test this configuration in another machine. But I could include a comment section here in case we need to revisit it in the future.

jedwards4b:
Replaced by #3989

PR #3982, opened 2021-05-25
Allow multiple grids for a component

### LINKED ISSUE #3962: Reworking grids.py to support multiple grids for a component
I'm in the process of reworking grids.py to support multiple grids for a component. The purpose is to allow multiple grids for the glacier component, in conjunction with some changes @mvertens has been making to CMEPS to support multiple glc components at the mediator level. This way, we can have separate grids for Greenland and Antarctica without having a combinatoric issue.

In doing this, I'm finding myself reworking grids.py more than I originally envisioned: this has become a bit of a peeling onion. To support this work, I am introducing a number of unit tests around grids.py, where I am setting up fake little config_grids.xml files to test aspects of the grids.py logic, to try to make sure I don't break anything now and that future changes won't break the changes I'm making.

I'll open a PR for this when it's ready for review, but for now wanted to let others know that I'm working on this.

billsacks:
This is intended to support the GLC component having multiple grids (e.g., Greenland & Antarctica), but the implementation handles any component having multiple grids.

At a high level: If a component's grid is a colon-delimited string, then it is assumed to specify multiple grids, separated by the colons. Then, anywhere where we use the component's grid to create derived variables (domain name, mapping file names, etc.), I have introduced a loop over the list of grids for each component. For a component with multiple grids, the derived variables then also end up as colon-delimited strings, with one colon-delimited element per grid. I introduced similar loops in two places in nmlgen where we process file names.

To avoid adding more complexity and deep nesting to the Grids class, I have refactored it significantly in a few places. Main goals of the refactoring were (1) extracting some logic into a helper class, _ComponentGrids; among other things, this replaces various lookups that were being done using tuples and dicts; and (2) extracting the bodies of nested loops into new methods to avoid super-deep nesting. I also applied some refactorings like (2) to the nmlgen class for a similar reason.

I have also added unit tests of grids.py. The high-level tests of the Grids class work by constructing little fake xml files on the fly. There are also some tests of the lower-level helper functions. These unit tests don't cover all of the logic in grids.py, but they do cover a lot of the logic, with a particular focus on any parts of the code that I had to change significantly for this work.

A caveat about this implementation – particularly for the changes in nmlgen – is that it assumes that colons will only appear in file paths for this purpose of using multiple files. If people feel that is a dangerous assumption, we will either need to change the delimiter or more significantly rework the implementation. (I considered using an array of strings instead of a colon-delimited string. However, I prefer the colon-delimited approach because (1) @mvertens suggested it as an approach that could work well with the Fortran code in CMEPS, (2) I believe that changing this to use lists of strings would require *many* changes to namelist_definition xml files as well as a need to be vigilant moving forward that all relevant variables are declared as lists in the xml files, and (3) I'm not sure if the list-of-strings approach would work at all for the case xml files.) I did check all of the CESM inputdata that I could find on both cheyenne and izumi. The only files in CESM's inputdata that I can find that have a colon are these two very old files, which I don't believe would ever appear in inputdata lists in modern code bases:
- `./atm/ccm3/SEP1.T05.0198.nc:t.adump`
- `./lnd/clm2/rawdata/urban_data_Feddema:080410_0.5x0.5_avg_c080908.nc`

The driver/mediator code will also need to be extended to permit colon-delimited strings for various variables. However, this only needs to be done on an as-needed basis: As long as no component has multiple grids (the status quo), there will be no changes to the namelist variables. Soon, we will allow multiple grids for the GLC component with nuopc/cmeps in CESM, so we will extend CMEPS accordingly. We have no plans to extend the MCT driver/coupler to allow this, but the changes in this PR will work fine with the existing code.

@mvertens has provided guidance for the high-level design of this feature.

Test suite:
- scripts_regression_tests on cheyenne: all pass
- cime_developer test suite on cheyenne: all pass and bit-for-bit
- CESM's aux_cime_baselines on cheyenne: all pass and bit-for-bit
- CESM's prealpha tests on cheyenne and izumi: all pass and bit-for-bit except for tests that seem to have failed or weren't run at all in the baselines (`SMS_D_Ln9.f09_f09_mg17.FCHIST.cheyenne_intel.cam-outfrq9s_ocnemis` and the equivalent nuopc test had an exception in NLCOMP, but hand comparisons showed that namelists were effectively identical)

Test baseline: cime5.8.47, cesm2_3_alpha03a
Test namelist changes: none
Test status: bit for bit

Fixes #3962

User interface changes?: N

### DIFF
--- a/scripts/lib/CIME/nmlgen.py
+++ b/scripts/lib/CIME/nmlgen.py
@@ -754,12 +754,12 @@ class NamelistGenerator(object):
         for one_file_path in file_path.split(GRID_SEP):
             # NOTE - these are hard-coded here and a better way is to make these extensible
             if one_file_path == 'UNSET' or one_file_path == 'idmap' or one_file_path == 'idmap_ignore':
-                return
+                continue
             if input_pathname == 'abs':
                 # No further mangling needed for absolute paths.
                 # At this point, there are overwrites that should be ignored
                 if not os.path.isabs(one_file_path):
-                    return
+                    continue
                 else:
                     pass
             elif input_pathname.startswith('rel:'):
--- a/scripts/lib/CIME/nmlgen.py
+++ b/scripts/lib/CIME/nmlgen.py
@@ -740,8 +740,8 @@ class NamelistGenerator(object):
 
         It's possible that file_path actually contains multiple files delimited by
         GRID_SEP. (This is the case when a component has multiple grids, and so has a file
-        for each grid.) In this case, we split it on GRID_SEP and handle each separated portion as a
-        separate file.
+        for each grid.) In this case, we split it on GRID_SEP and handle each separated
+        portion as a separate file.
 
         Args:
         - input_data_list: file handle
@@ -771,7 +771,12 @@ class NamelistGenerator(object):
             else:
                 expect(False,
                        "Bad input_pathname value: {}.".format(input_pathname))
+
             # Write to the input data list.
+            #
+            # Note that the same variable name is repeated for each file. This currently
+            # seems okay for check_input_data, but if it becomes a problem, we could
+            # change this, e.g., appending an index to the end of variable_name.
             string = "{} = {}".format(variable_name, one_file_path)
             hashValue = hashlib.md5(string.rstrip().encode('utf-8')).hexdigest()
             if hashValue not in lines_hash:
--- a/scripts/lib/CIME/nmlgen.py
+++ b/scripts/lib/CIME/nmlgen.py
@@ -17,6 +17,7 @@ from CIME.namelist import Namelist, parse, \
 from CIME.XML.namelist_definition import NamelistDefinition
 from CIME.utils import expect, safe_copy
 from CIME.XML.stream import Stream
+from CIME.XML.grids import GRID_SEP
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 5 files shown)


billsacks:
Some notes for reviewing this:
- As noted above, I'm going to do extensive testing over the next day or two in the CESM context to make sure my grids.py changes haven't broken anything. I'll report back here when that is done.
- Many of the diffs here look bigger than they are, because I have extracted some chunks of code into new methods. Ignoring whitespace in the diffs could help make it easier to see the real diffs. I'll also add some line comments.
- @jgfouca you might want to run at least a little testing with E3SM before approving, given the magnitude of changes. I'll leave it up to you.

scripts/lib/CIME/nmlgen.py
```diff
@@ -707,33 +729,61 @@ def _write_input_files(self, data_list_path):
...
+            # Note that the same variable name is repeated for each file. This currently
+            # seems okay for check_input_data, but if it becomes a problem, we could
+            # change this, e.g., appending an index to the end of variable_name.
+            string = "{} = {}".format(variable_name, one_file_path)
+            hashValue = hashlib.md5(string.rstrip().encode('utf-8')).hexdigest()
+            if hashValue not in lines_hash:
+                logger.debug("Adding line {} with hash {}".format(string,hashValue))
+                input_data_list.write(string+"\n")
+            else:
+                logger.debug("Line already in file {}".format(string))
```
billsacks, line 786: The body of this for loop is essentially the same as the pre-existing code, except `file_path` is replaced with `one_file_path`

billsacks:
I am finished running extensive testing on this PR. I have updated the top-level comment.

jedwards4b (approved):
Looks good, thanks for all of the notes.

jgfouca (approved):

PR #3985, opened 2021-06-03
logical args to Make should not be quoted, and update esmf library and cheyenne_gnu compiler versions

jedwards4b:
Update the esmf library module on cheyenne for all compilers, update gnu compiler to 10.1.0 on cheyenne.  Fix an issue in build.py that was causing the ESMF_VERSION flags not to be available to 
the compile line (USE_ESMF_LIB=TRUE not USE_ESMF_LIB="TRUE")

Test suite: scripts regression tests with intel, gnu and pgi compilers on cheyenne with cime_driver=nuopc 
Test baseline: 
Test namelist changes: 
Test status: bit for bit
Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/config/cesm/machines/config_compilers.xml
+++ b/config/cesm/machines/config_compilers.xml
@@ -896,6 +896,9 @@ using a fortran linker.
     <append MODEL="pio1"> -DNO_MPIMOD </append>
   </CPPDEFS>
   <FFLAGS>
+    <!-- -fallow-argument-mismatch is needed for building with the mpich library with gfortran10;
+         -fallow-invalid-boz is needed for some components that have non-standard uses of BOZ literal constants with gfortran10.
+         See https://github.com/ESMCI/cime/issues/3846 for details. -->
     <append> -fallow-argument-mismatch -fallow-invalid-boz </append>
   </FFLAGS>
   <SLIBS>
--- a/config/cesm/machines/config_compilers.xml
+++ b/config/cesm/machines/config_compilers.xml
@@ -200,7 +200,7 @@ using a fortran linker.
   <FFLAGS>
     <!-- -ffree-line-length-none and -ffixed-line-length-none need to be in FFLAGS rather than in FIXEDFLAGS/FREEFLAGS
        so that these are passed to cmake builds (cmake builds don't use FIXEDFLAGS and FREEFLAGS). -->
-    <base>  -fconvert=big-endian -ffree-line-length-none -ffixed-line-length-none </base>
+    <base>  -fconvert=big-endian -ffree-line-length-none -ffixed-line-length-none -fallow-argument-mismatch -fallow-invalid-boz </base>
     <append compile_threaded="TRUE"> -fopenmp </append>
     <!-- Ideally, we would also have 'invalid' in the ffpe-trap list. But at
          least with some versions of gfortran (confirmed with 5.4.0, 6.3.0 and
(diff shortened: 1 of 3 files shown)


fischer-ncar (approved):

config/cesm/machines/config_compilers.xml
```diff
@@ -200,7 +200,7 @@ using a fortran linker.
   <FFLAGS>
     <!-- -ffree-line-length-none and -ffixed-line-length-none need to be in FFLAGS rather than in FIXEDFLAGS/FREEFLAGS
        so that these are passed to cmake builds (cmake builds don't use FIXEDFLAGS and FREEFLAGS). -->
-    <base>  -fconvert=big-endian -ffree-line-length-none -ffixed-line-length-none </base>
+    <base>  -fconvert=big-endian -ffree-line-length-none -ffixed-line-length-none -fallow-argument-mismatch -fallow-invalid-boz </base>
```
billsacks, line 203: Thanks for adding these flags. But, since they are workarounds for issues that will hopefully be resolved, I think it would be helpful to add a comment stating that – i.e., something like:

```
-fallow-argument-mismatch is needed for building with the mpich library with gfortran10; -fallow-invalid-boz is needed for some components that have non-standard uses of BOZ literal constants with gfortran10. See https://github.com/ESMCI/cime/issues/3846 for details.
```

However, I'm also wondering if these flags work with older versions of gfortran, or if they require gfortran10+.

config/cesm/machines/config_machines.xml
```diff
@@ -559,76 +559,76 @@ This allows using a different mpirun command to launch unit tests
         <command name="load">mkl</command>
       </modules>
       <modules compiler="gnu">
-        <command name="load">gnu/9.1.0</command>
-        <command name="load">openblas/0.3.6</command>
+        <command name="load">gnu/10.1.0</command>
```
billsacks, line 562: Have you confirmed that this doesn't change answers for any of our CESM testing? It would be good to know that before merging.

jedwards4b:
gnu and intel compilers are working but I'm having issues with pgi.

billsacks (changes requested):

config/cesm/machines/config_machines.xml
billsacks: In the past, I think we've said we should do at least prealpha testing when doing a major compiler version update like this. @fischer-ncar , if you or @jedwards4b hasn't already done so, would you be able to run the cheyenne_gnu prealpha tests with baseline comparisons? (And prebeta cheyenne_gnu if you feel it's warranted; I notice that prealpha cheyenne_gnu has very little coverage.)

fischer-ncar: @billsacks I've start prealpha and prebeta tests for cheyenne_gnu.

jgfouca (approved):

config/cesm/machines/config_machines.xml
fischer-ncar: Testing is finished.  gnu/10.1.0 is bfb with gnu/9.1.0.

billsacks: Awesome, thanks!

fischer-ncar:
@jedwards4b are you ready to merge this PR?

config/cesm/machines/config_machines.xml
ekluzek: Just FYI, we did find a case where answers changed because of this update. See https://github.com/ESCOMP/CTSM/issues/1464

I want to highlight it just to reinforce that our paranoia over compiler updates is often justified. :-) It did take us a bit to track it down to this change. It's also helpful that the compiler change was in one PR and not mixed with anything else, so thanks for that.

jedwards4b:
@ekluzek It also turns out that the smartsim tool doesn't work with gnu 10.   PR #4045 backs out this change.

PR #3988, opened 2021-06-05
Raise an error if no value can be found for a namelist variable

### LINKED ISSUE #3984: Missing default namelist values should raise an error but do not
If a namelist variable has defaults defined for certain attribute values but not others, and no match is found, the current code simply puts an empty string in as the value and moves on (even if the variable is not a character variable). Instead, an error should be raised in this situation.

I feel like this error checking is important because it is easy to accidentally leave out certain sets of matches from a `namelist_definition` file, and it would be much better for this issue to be caught at `preview_namelists` time rather than generating a presumably cryptic error at model runtime. It looks like CIME has been missing this error checking ever since the introduction of the python-based nmlgen, but I just noticed it now because I have only started working with python-based buildnmls in the last few months.

To reproduce this issue, introduce diffs like these:

``` diff
diff --git a/src/drivers/mct/cime_config/namelist_definition_drv.xml b/src/drivers/mct/cime_config/namelist_definition_drv.xml
index 05800db03..1927e53fc 100644
--- a/src/drivers/mct/cime_config/namelist_definition_drv.xml
+++ b/src/drivers/mct/cime_config/namelist_definition_drv.xml
@@ -750,10 +750,6 @@
       Setting this to zero will always do flux_max_iteration
     </desc>
     <values>
-      <!-- the value for f09 grids is maintained as 0.0 for compatibility with cmip6 experiments
-           the 1% convergence criteria improves performance of the atm ocn flux calculation in shr_flux_mod.F90 -->
-      <value cime_model='cesm'>0.01</value>
-      <value cime_model='cesm' atm_grid="0.9x1.25">0.0</value>
       <value cime_model='e3sm'>0.0</value>
     </values>
   </entry>
```

then run a test like the following, using `cime_model` = 'cesm' (note that I have removed the default values for 'cesm' in the above diffs):

```
./create_test --no-build SMS.f09_g16.X.bishorn_gnu
# cd to the test directory
./preview_namelists
grep flux_convergence CaseDocs/*
```

(I have reproduced the issue for integer & real values. For strings, the issue ends up triggering the valid value check if valid values are defined for that variable. That still doesn't seem like the right behavior, but at least it's better than not triggering an error at all.)

It looks like the issue is that https://github.com/ESMCI/cime/blob/8ec59c02d0042b9e1e134ffbeba9fccb1f31a061/scripts/lib/CIME/XML/namelist_definition.py#L195-L196 prevents the error checking that is supposed to happen in https://github.com/ESMCI/cime/blob/8ec59c02d0042b9e1e134ffbeba9fccb1f31a061/scripts/lib/CIME/nmlgen.py#L308-L310 . It looks like the behavior change came in 9991c862d32254b9d78131560309804b7e03d34a; I can't tell from that commit what the purpose of the change was and in what situations that change was needed. Without understanding the details, I would naively think that, if certain variables need the ability to have a fall-back value of an empty string, then they should do it explicitly in their `namelist_definition` file, by setting `<value> </value>`.

billsacks:
Previously, if no value was found for a namelist variable, the code simply put an empty string in as the value, without raising any errors. This PR adds error checking.

Fixing this issue revealed some other latent issues, which were being masked by this one. So this PR also fixes those other latent issues. See https://github.com/ESMCI/cime/issues/3984 for extensive details.

Test suite:
- Manual testing of the situation described in https://github.com/ESMCI/cime/issues/3984
- cime_developer tests on cheyenne, with baseline comparisons
- CESM's aux_cime_baseline tests on cheyenne (mct & nuopc), with baseline comparisons
- CESM's prealpha tests on cheyenne and izumi (mct & nuopc), with baseline comparisons

Test baseline: cesm2_3_alpha03a
Test namelist changes: Adds pio_netcdf_format for ESP, which was previously undefined; other than that, namelists are identical to baseline
Test status: bit for bit

Fixes #3984

User interface changes?: N

### DIFF
--- a/scripts/lib/CIME/XML/namelist_definition.py
+++ b/scripts/lib/CIME/XML/namelist_definition.py
@@ -174,6 +174,9 @@ class NamelistDefinition(EntryID):
         """This function is not implemented."""
         raise TypeError("NamelistDefinition does not support `set_value`.")
 
+    # In contrast to the entry_id version of this method, this version doesn't support the
+    # replacement_for_none argument, because it is hard-coded to ''.
+    # pylint: disable=arguments-differ
     def get_value_match(self, vid, attributes=None, exact_match=True, entry_node=None):
         """Return the default value for the variable named `vid`.
 
@@ -190,6 +193,8 @@ class NamelistDefinition(EntryID):
 
         if entry_node is None:
             entry_node = self._nodes[vid]
+        # NOTE(wjs, 2021-06-04) In the following call, replacement_for_none='' may not
+        # actually be needed, but I'm setting it to maintain some old logic, to be safe.
         value = super(NamelistDefinition, self).get_value_match(vid.lower(),attributes=all_attributes, exact_match=exact_match,
                                                                 entry_node=entry_node,
                                                                 replacement_for_none='')
--- a/scripts/lib/CIME/XML/namelist_definition.py
+++ b/scripts/lib/CIME/XML/namelist_definition.py
@@ -156,6 +156,10 @@ class NamelistDefinition(EntryID):
     def add_attributes(self, attributes):
         self._attributes = attributes
 
+    def get_attributes(self):
+        """Return this object's attributes dictionary"""
+        return self._attributes
+
     def get_entry_nodes(self):
         return self._entry_nodes
 
--- a/scripts/lib/CIME/XML/namelist_definition.py
+++ b/scripts/lib/CIME/XML/namelist_definition.py
@@ -191,10 +191,9 @@ class NamelistDefinition(EntryID):
         if entry_node is None:
             entry_node = self._nodes[vid]
         value = super(NamelistDefinition, self).get_value_match(vid.lower(),attributes=all_attributes, exact_match=exact_match,
-                                                                entry_node=entry_node)
-        if value is None:
-            value = ''
-        else:
+                                                                entry_node=entry_node,
+                                                                replacement_for_none='')
+        if value is not None:
             value =  self._split_defaults_text(value)
 
         return value
@@ -445,4 +444,4 @@ class NamelistDefinition(EntryID):
             all_attributes.update(attribute)
 
         value = self.get_value_match(item.lower(), all_attributes, True)
-        return self._split_defaults_text(value)
+        return value
--- a/src/drivers/mct/cime_config/namelist_definition_modelio.xml
+++ b/src/drivers/mct/cime_config/namelist_definition_modelio.xml
@@ -172,6 +172,7 @@
     <value component="glc">$GLC_PIO_NETCDF_FORMAT</value>
     <value component="wav">$WAV_PIO_NETCDF_FORMAT</value>
     <value component="iac">$IAC_PIO_NETCDF_FORMAT</value>
+    <value component="esp">$ESP_PIO_NETCDF_FORMAT</value>
     </values>
   </entry>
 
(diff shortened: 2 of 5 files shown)


billsacks:
I have done extensive testing of this, but I would be happy to have a critical eye on this, given the subtleties in the namelist generation code.

@jgfouca I found a few places where defaults weren't being set correctly in CESM, which was ignored before this but picked up after this PR. So it could be worth running some E3SM testing before merging this, or at least keeping an eye out for issues after this is merged – both to catch pre-existing problems with default values not being available, and also in case these changes inadvertently change any namelist defaults (they shouldn't, and don't appear to for CESM, but there's a chance that I missed some subtlety in the logic). If you run into problems, I'm happy to help resolve them.

scripts/lib/CIME/XML/namelist_definition.py
```diff
@@ -190,11 +197,12 @@ def get_value_match(self, vid, attributes=None, exact_match=True, entry_node=Non
...
             entry_node = self._nodes[vid]
+        # NOTE(wjs, 2021-06-04) In the following call, replacement_for_none='' may not
+        # actually be needed, but I'm setting it to maintain some old logic, to be safe.
         value = super(NamelistDefinition, self).get_value_match(vid.lower(),attributes=all_attributes, exact_match=exact_match,
-                                                                entry_node=entry_node)
-        if value is None:
-            value = ''
-        else:
+                                                                entry_node=entry_node,
+                                                                replacement_for_none='')
```
billsacks, line 204: As I say in the comment, I don't know if this replacement_for_none is actually needed; I added it to be safe to maintain existing logic. @jedwards4b @jgfouca do you know off-hand if a None value can be generated from the text value retrieved by entry_id's get_value_match? If you'd like, I can do some testing of whether this is ever triggered in CESM testing and remove it if it isn't.

Also: I don't love that the interface of get_value_match now differs from the entry_id version (requiring `#pylint: disable=arguments-differ`), but this seemed better to me than other implementations I could think of. I'm open to suggestions if you see a better way to do this.

scripts/lib/CIME/XML/namelist_definition.py
```diff
@@ -445,4 +453,4 @@ def get_default_value(self, item, attribute=None):
             all_attributes.update(attribute)
 
         value = self.get_value_match(item.lower(), all_attributes, True)
-        return self._split_defaults_text(value)
+        return value
```
billsacks, line 456: The original motivation for this change was seeing that `self.get_value_match` can now return `None`; but as I looked closely, I saw that there was a double-call to `self._split_defaults_text(value)` (at the end of `self.get_value_match` and here) that looks like a bug, so I have removed that double-call. I don't see any calls to this `get_default_value` routine from within cime, but it's possible that components use it in their buildnml.

src/drivers/mct/cime_config/namelist_definition_modelio.xml
```diff
@@ -172,6 +172,7 @@
     <value component="glc">$GLC_PIO_NETCDF_FORMAT</value>
     <value component="wav">$WAV_PIO_NETCDF_FORMAT</value>
     <value component="iac">$IAC_PIO_NETCDF_FORMAT</value>
+    <value component="esp">$ESP_PIO_NETCDF_FORMAT</value>
```
billsacks, line 175: I'm not sure if this is the right setting: other esp pio values seem to be set to some "undefined" value. But this seemed more right than the earlier behavior, where this was left as blank in the namelist file.

jgfouca (approved):

jgfouca:
Thanks, @billsacks . I will keep this in mind the next time I update CIME for E3SM.

PR #3990, opened 2021-06-08
Seperation of fortran code to new repositories.

fischer-ncar:
Split out shared fortran code into seperate git repositories.  

- Only use the relocated code when CIME_MODEL=cesm.  
- Update build scripts to build from new locations.  
- Need to run manage_externals/checkout_externals in order to clone the 
  git repos in that are found in Externals.cfg.

These are the git repos that are used, and the files that were moved into them.
CESM_CPL7andDataComps and CESM_share are new git repositories.

 The following are in  https://github.com/ESCOMP/CESM_CPL7andDataComps
```
        Relocate src/components/*_mct to components/cpl7/components.
        Relocate src/drivers/mct      to components/cpl7/driver.
```

The following are in https://github.com/ESCOMP/CESM_share
```
	Relocate files form src/share/util to share/src and components/cpl7/mct_shr
	Relocate src/share/nuopc           to share/cmeps
	Relocate src/share/include         to share/include.
	Relocate src/share/RandNum         to share/RandNum.
	Relocate src/share/test            to share/test.
	Relocate src/share/unit_test_stubs to share/unit_test_stubs
```

Now using mct from https://github.com/MCSclimate/MCT
```
	Relocate src/externals/mct to libraries/mct
```

Now using pio2 from https://github.com/NCAR/ParallelIO
```	
        Relocate src/exteranls/pio2 to libraries/parallelio
```

Test suite: scripts_regression_tests for mct and nuopc, prealpha on cheyenne
Test baseline: cesm2_3_alpha03a
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

Code review:


.github/workflows/srt.yml
```diff
@@ -52,7 +82,7 @@ jobs:
         uses: actions/cache@v2
         with:
           path: ~/pnetcdf
-          key: ${{ runner.os }}-${{ env.PNETCDF_VERSION}}-pnetcdf-redo
+          key: ${{ runner.os }}-foo-v3-${{ env.PNETCDF_VERSION}}-pnetcdf-redo
```
jedwards4b, line 85: do we want to remove the foo-v part and the redo part?

.github/workflows/srt.yml
```diff
@@ -108,13 +138,21 @@ jobs:
...
+          ls $HOME/work/cime/cime/components
+          echo "-----"
           mkdir -p $HOME/cesm/scratch
           mkdir -p $HOME/cesm/inputdata
           cd $HOME/work/cime/cime/scripts/tests
           export NETCDF=$HOME/netcdf-fortran
           export PATH=$NETCDF/bin:$PATH
           export LD_LIBRARY_PATH=$NETCDF/lib:$HOME/pnetcdf/lib:$LD_LIBRARY_PATH
-          ./scripts_regression_tests.py --no-fortran-run --compiler gnu --mpilib openmpi --machine ubuntu-latest
+#CAF          ./scripts_regression_tests.py --no-fortran-run --compiler gnu --mpilib openmpi --machine ubuntu-latest
```
jedwards4b, line 155: so we aren't running the test at this time?

.github/workflows/srt_nuopc.yml
```diff
@@ -0,0 +1,195 @@
+# CIME scripts regression tests
+
+name: nuopc scripts regression tests
+
+# Controls when the action will run. Triggers the workflow on push or pull request
+# events but only for the master branch
+on:
+  push:
+    branches: [ master, maint-5.6 ]
```
jedwards4b, line 9: I think that we don't need (or want) cmeps tests on maint-5.6

.github/workflows/srt_nuopc.yml
```diff
@@ -0,0 +1,195 @@
...
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+
+      - name: cache pnetcdf
+        id: cache-pnetcdf
+        uses: actions/cache@v2
+        with:
+          path: ~/pnetcdf
+          key: ${{ runner.os }}-nuopc-${{ env.PNETCDF_VERSION}}-pnetcdf-redo
```
jedwards4b, line 104: ultimately the key here should be the same as the key for the mct build

.gitignore
```diff
@@ -19,4 +19,10 @@ scripts/Tools/JENKINS*
 /scripts/cases/
 
 #Ignore nuopc driver available seperatly
-src/drivers/nuopc
\ No newline at end of file
+src/drivers/nuopc
```
jedwards4b, line 22: didn't we move this to components?

config/cesm/config_files.xml
```diff
@@ -108,9 +108,11 @@
     <type>char</type>
     <values>
       <value component="datm" comp_interface="nuopc" >$SRCROOT/components/cdeps/datm</value>
-      <value component="datm" comp_interface="mct"   >$CIMEROOT/src/components/data_comps_$COMP_INTERFACE/datm</value>
-      <value component="satm"                        >$CIMEROOT/src/components/stub_comps_$COMP_INTERFACE/satm</value>
-      <value component="xatm"                        >$CIMEROOT/src/components/xcpl_comps_$COMP_INTERFACE/xatm</value>
+      <value component="datm" comp_interface="mct"   >$SRCROOT/components/cpl7/components/data_comps_$COMP_INTERFACE/datm</value>
+      <value component="satm" comp_interface="nuopc" >$CIMEROOT/src/components/stub_comps_$COMP_INTERFACE/satm</value>
+      <value component="satm" comp_interface="mct"   >$SRCROOT/components/cpl7/components/stub_comps_$COMP_INTERFACE/satm</value>
+      <value component="xatm" comp_interface="nuopc" >$CIMEROOT/src/components/xcpl_comps_$COMP_INTERFACE/xatm</value>
```
jedwards4b, line 114: we no longer have stub comps for cmeps - I think this line and the other stubs with comp_interface="nuopc" can be removed.

scripts/Tools/Makefile
```diff
@@ -898,15 +899,15 @@ $(PIOLIB) : $(MPISERIAL) $(GPTLLIB)
...
+ifneq ($(findstring csm_share,$(MODEL)),csm_share)
   $(OBJS):  $(CSMSHARELIB)
 else
   complib: install_lib
 endif
 
 install_lib: $(COMPLIB)
 	$(CP) -p $(COMPLIB) $(CSMSHARELIB)
-	$(CP) -p *.$(MOD_SUFFIX) *.h $(INCLUDE_DIR)
+	$(CP) -p *.$(MOD_SUFFIX)  $(INCLUDE_DIR)
```
jedwards4b, line 910: Why remove *.h?   I think that it may be important for some compsets or compilers.

jedwards4b (approved):

.github/workflows/srt_nuopc.yml
fischer-ncar: Does removing references to maint-5.6 include srt.yml?

.github/workflows/srt.yml
fischer-ncar: Not a problem, I was setting foo-v and redo while I was figuring out github workflows.

.github/workflows/srt.yml
fischer-ncar: I commented this out for debugging.  We're suppose to be running the test.

scripts/Tools/Makefile
fischer-ncar: When building the library for the CESM_share repository, there are no *.h.  Which causes an error.  Let me see what I can 
do about this.

config/cesm/config_files.xml
fischer-ncar: done

.gitignore
fischer-ncar: Yes, I removed it.

.github/workflows/srt_nuopc.yml
fischer-ncar: pnetcdf keys are now the same.

.github/workflows/srt.yml
fischer-ncar: Fixed

.github/workflows/srt.yml
fischer-ncar: Done

config/cesm/config_files.xml
fischer-ncar: I had to put back in the stub comps for cmeps.  scripts_regression_tests were failing without them.  I believe it's just an issue with matching.

rljacob:
So this doesn't remove the code from their current locations?

CMakeLists.txt
```diff
@@ -9,7 +9,11 @@ project(cime_tests Fortran C)
 
 # We rely on pio for cmake utilities like findnetcdf.cmake, so that we don't
 # need to duplicate this cmake code
-list(APPEND CMAKE_MODULE_PATH "${CIME_ROOT}/src/externals/pio2/cmake")
+if (EXISTS ${SRC_ROOT}/libraries/parallelio/cmake)
+  list(APPEND CMAKE_MODULE_PATH "${SRC_ROOT}/libraries/parallelio/cmake")
+else()
+  list(APPEND CMAKE_MODULE_PATH "${CIME_ROOT}/src/externals/pio2/cmake")
+endif()
```
rljacob, line 16: Should we put the path in config_files.xml?  Same for MCT below.

rljacob:
@jgfouca does cime/CMakelists.txt have a role in the e3sm build?

CMakeLists.txt
fischer-ncar: No code was removed.  I have no issues with adding the paths to config_files.xml.  I would want to update several other files to remain consistent.  So I would prefer to do that in a separate PR.  But if you want, I'll do it for this PR.

jgfouca:
@rljacob , no.

CMakeLists.txt
rljacob: After telecon:  Nevermind.  Each model can effectively set the location in config_files.xml already, this file is only for the cime unit tests and the logic will be removed after the old code is removed.

rljacob (approved):

jgfouca (approved):

billsacks (changes requested):
@fischer-ncar - thanks a lot for your work on this. I have a few inline comments below.

The change to `CSMSHARELIB` (which looks wrong to me but you could convince me that it's actually right) also made me think more generally: When I change Makefile logic, I'm often concerned that I'm failing to capture a dependency relationship that's needed for a rebuild to work correctly when some code has changed. For example, if I understand the original logic correctly (which I may not be), it was saying: if the csm_share library is updated, then rebuild all component object files. Breakages in this logic might not show up in a clean build from scratch, but can show up in an incremental rebuild after you have changed some source file(s) in the depended-on library (e.g., csm_share). Have you done some testing where you change some code in a lower-level library like csm_share and ensure that the correct other libraries are and are not rebuilt?

billsacks:
> That raises a good question, what is it used for? I'm nearly certain it's not used in the cime or E3SM build systems.

It's used for the unit test build.

scripts/lib/CIME/build.py
```diff
@@ -389,6 +390,8 @@ def _build_libraries(case, exeroot, sharedpath, caseroot, cimeroot, libroot, lid
         logger.info("UFS_DRIVER is set to {}".format(ufs_driver))
     if ufs_driver and ufs_driver == 'nems' and not cpl_in_complist:
         libs = []
+    elif case.get_value("MODEL") == "cesm":
+        libs = ["gptl", "mct", "pio", "csm_share", "csm_share_cpl7"]
```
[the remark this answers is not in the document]
fischer-ncar: Yes, csm_share_cpl7 was being built for nuopc.  I've updated build.py so csm_share_cpl7 isn't being build for nuopc cases.

scripts/Tools/Makefile
```diff
@@ -612,10 +612,8 @@ endif
 # Drive configure scripts for support libraries (mct)
 #------------------------------------------------------------------------------
 
-ifeq ($(CIME_MODEL),cesm)
-    #This will be changed to EXTERN_PATH = $(SRCROOT)/libraries when the new
-    #share repos are used for cesm2.
-    EXTERN_PATH = $(CIMEROOT)/src/externals
+ifneq ("$(SRCROOT)/libraries",)
```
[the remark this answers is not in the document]
fischer-ncar: This logic is wrong, I've fixed it.

scripts/Tools/Makefile
```diff
@@ -873,7 +874,7 @@ endif
 
 # libcsm_share.a is in ULIBDEP, but -lcsm_share is in ULIBS rather than CLIBS,
 # so this needs to be added after creating CLIBS above
-CSMSHARELIB = $(INSTALL_SHAREDPATH)/$(COMP_INTERFACE)/$(ESMFDIR)/$(NINST_VALUE)/lib/libcsm_share.a
+CSMSHARELIB = $(INSTALL_SHAREDPATH)/$(COMP_INTERFACE)/$(ESMFDIR)/$(NINST_VALUE)/lib/lib$(MODEL).a
```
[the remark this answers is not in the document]
fischer-ncar: This is wrong, I'm surprised it worked.  CSMSHARELIB is now set to either lib_cesm_share or lib_cesm_share_cpl7.  I've also added CSMSHARELIB_CPL7 to address an issue you brought up below.

scripts/Tools/Makefile
```diff
@@ -898,8 +899,10 @@ $(PIOLIB) : $(MPISERIAL) $(GPTLLIB)
 
 $(CSMSHARELIB):  $(MCTLIBS) $(PIOLIB) $(GPTLLIB)
 
-ifneq ($(MODEL),csm_share)
+ifneq ($(findstring csm_share,$(MODEL)),csm_share)
   $(OBJS):  $(CSMSHARELIB)
```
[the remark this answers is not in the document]
fischer-ncar: Your right, I've added CSMSHARELIB_CPL7 to the dependency for the rebuild.  The original reason I changed the logic here was csm_share_cpl7 was throwing a copy error because there was there was no *.h files.

scripts/Tools/Makefile
```diff
@@ -873,7 +878,16 @@ endif
...
 # libcsm_share.a is in ULIBDEP, but -lcsm_share is in ULIBS rather than CLIBS,
 # so this needs to be added after creating CLIBS above
-CSMSHARELIB = $(INSTALL_SHAREDPATH)/$(COMP_INTERFACE)/$(ESMFDIR)/$(NINST_VALUE)/lib/libcsm_share.a
+ifeq ($(CIME_MODEL),cesm)
+  ifneq ($(COMP_INTERFACE), nuopc)
+    CSMSHARELIB_CPL7 = $(INSTALL_SHAREDPATH)/$(COMP_INTERFACE)/$(ESMFDIR)/$(NINST_VALUE)/lib/libcsm_share_cpl7.a
+  endif
+endif
+ifeq ($(MODEL),csm_share_cpl7)
+  CSMSHARELIB = $(CSMSHARELIB_CPL7)
```
billsacks, line 887: Can you please add a comment explaining the purpose of this unintuitive line?

scripts/Tools/Makefile
```diff
@@ -908,6 +924,10 @@ install_lib: $(COMPLIB)
 	$(CP) -p $(COMPLIB) $(CSMSHARELIB)
 	$(CP) -p *.$(MOD_SUFFIX) *.h $(INCLUDE_DIR)
 
+install_lib_cpl7: $(COMPLIB)
+	$(CP) -p $(COMPLIB) $(CSMSHARELIB)
```
billsacks, line 928: Should CSMSHARELIB here be CSMSHARELIB_CPL7? Is this maybe the reason for the unintuitive line I called out above?

billsacks (changes requested):
Thanks @fischer-ncar . This looks better now. I just have a couple of remaining comments below.

scripts/Tools/Makefile
fischer-ncar: Yes, using CSMSHARELIB_CPL7 here removes the need for the unintuitive line above.

scripts/Tools/Makefile
fischer-ncar: This isn't needed anymore when using CSMSHARELIB_CPL7 in install_lib_cpl7 mentioned below.

billsacks (approved):
Looks good now - thank you!

PR #4004, opened 2021-06-17
add a check for case_dir (should contain env_case.xml if it exists

### LINKED ISSUE #4003: xmlchange should detect if it is in a case directory or not
I was helping a user update the account key charged by a case, and instructed her to run

```
$ ./xmlchange PROJECT=#####
```

I wasn't clear about where to run this, though, and she ran the command from `cime/scripts/Tools` rather than her case root. The result of such a mistake is

```
$ git status
HEAD detached at cime5.8.42
Untracked files:
  (use "git add <file>..." to include in what will be committed)

	scripts/Tools/env_archive.xml
	scripts/Tools/env_batch.xml
	scripts/Tools/env_build.xml
	scripts/Tools/env_case.xml
	scripts/Tools/env_mach_pes.xml
	scripts/Tools/env_mach_specific.xml
	scripts/Tools/env_run.xml
	scripts/Tools/env_workflow.xml
```

It would be great if, rather than blindly creating a set of xml files, `xmlchange` reported an error along the lines of "This command must be run from a case directory." I talked with @billsacks about this, and we came up with two possibilities:

1. add a check in the [`xmlchange()`](https://github.com/ESMCI/cime/blob/1e2bbff/scripts/Tools/xmlchange#L183) function before creating the `Case` object. This would be easy to implement but would not catch issues with running other scripts in the wrong location
2. add a check in the `Case` class itself. This would be a little bit more involved, and obviously there are some instances (such as `create_newcase`) where we want to allow a new `Case` object to generate files if `caseroot` is not currently a directory, so this would require a flag of some sort in the constructor... but it would probably be worth the extra effort to follow this path. I would recommend have the default for this flag be to report an error if `caseroot` is not already a case dir, and then explicitly allow scripts like `create_newcase` to proceed anyway.

jedwards4b:
For the case object, either the env_case.xml file must be present in case_root or case_root should not exist.

Test suite: scripts_regression_tests.py (maint-5.6) 
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #4003

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -84,6 +84,9 @@ class Case(object):
 
         if case_root is None:
             case_root = os.getcwd()
+
+        expect(not os.path.isdir(case_root) or os.path.isfile(os.path.join(case_root,"env_case.xml")), "Directory {} does not appear to be a valid case directory".format(case_root))
+
         self._caseroot = case_root
         logger.debug("Initializing Case.")
         self._read_only_mode = True
--- a/scripts/Tools/xmlchange
+++ b/scripts/Tools/xmlchange
@@ -161,6 +161,7 @@ def xmlchange(caseroot, listofsettings, xmlfile, xmlid, xmlval, subgroup,
               append, noecho, force, dryrun):
 
     with Case(caseroot, read_only=False) as case:
+        expect(os.path.isdir(caseroot), "Directory {} does not appear to be a valid case directory".format(caseroot))
         comp_classes = case.get_values("COMP_CLASSES")
         if xmlfile:
             case.set_file(xmlfile)


scripts/lib/CIME/case/case.py
```diff
@@ -84,6 +84,9 @@ def __init__(self, case_root=None, read_only=True):
 
         if case_root is None:
             case_root = os.getcwd()
+
+        expect(not os.path.isdir(case_root) or os.path.isfile(os.path.join(case_root,"env_case.xml")), "Directory {} does not appear to be a valid case directory".format(case_root))
```
mnlevy1981, line 88: At first blush I was a little confused by the logical statement... but this error will only be triggered if `case_root` exists but `{case_root}/env_case.xml` does not? And when `case_root` does not exist, there's already logic to handle whether that's allowed or not?

mnlevy1981 (approved):
assuming my interpretation of the `expect` statement is correct, this looks good to me. Thanks for the quick fix!

scripts/lib/CIME/case/case.py
jedwards4b: If the case_root can be created with mkdir -p it will be by create_newcase and create_test.
In master I was able to use the record_cmd function to confirm that the directory exists if expected.   In maint-5.6 I was only able to put this check in xmlchange itself.

PR #4008, opened 2021-06-19
Allow multiple methods for running unit tests

billsacks:
Besides "scripts_regression_tests.py A_RunUnitTests", there are two other methods for running unit tests that I like to use:

1. From scripts/lib, run "python3 -m unittest discover -s CIME/tests -t .": this is similar to what's run by A_RunUnitTests, but doesn't run the doc tests. This runs significantly faster than running "scripts_regression_tests.py A_RunUnitTests".
2. From scripts/lib, run an individual test file, like: "./CIME/tests/test_user_mod_support.py". This runs a single module's tests very fast. I like this method for doing TDD, because I can run unit tests super frequently, in the blink of an eye.

A recent change broke a few unit tests under method (1); this PR fixes that issue. In addition, method (2) wasn't working consistently for the different unit test files; this PR fixes that usage for all of the unit tests under CIME/tests and its subdirectories.

Test suite: scripts_regression_tests.py on cheyenne, and hand-testing of the above alternative options for running unit tests
Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes none

User interface changes?: N


scripts/lib/CIME/case/case.py
```diff
@@ -84,8 +84,7 @@ def __init__(self, case_root=None, read_only=True, record=False):
 
         if case_root is None:
             case_root = os.getcwd()
-        if not (in_doctest() or in_unit_test()):
```
billsacks, line 87: A bit of explanation for this change and the related changes in this file – doing away with the recently-added in_doctest() and in_unit_test() functions: I have replaced these functions with explicit handling in the unit tests and doctests that needed it, either by ensuring that caseroot is a non-existent directory or by putting an empty `env_case.xml` file in the caseroot directory. Reasons I prefer this method: (1) (main reason) it works for `python -m unittest discover`; (2) I found the other method hard to understand; (3) it makes me nervous to have unit test-specific code sprinkled in the production code, especially when that test-specific code is hard to understand: it's hard to be confident that things will work right in production.

scripts/lib/CIME/case/case.py
```diff
@@ -589,24 +588,28 @@ def _valid_compset_impl(self, compset_name, compset_alias, comp_classes, comp_ha
         <comp_hash> is a dictionary where each key is a supported component
         (e.g., datm) and the associated value is the index in <comp_classes> of
         that component's class (e.g., 1 for atm).
-        >>> Case(read_only=False)._valid_compset_impl('2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV', None, ['CPL', 'ATM', 'LND', 'ICE', 'OCN', 'ROF', 'GLC', 'WAV', 'ESP'], {'datm':1,'satm':1,'dlnd':2,'slnd':2,'dice':3,'sice':3,'docn':4,'socn':4,'drof':5,'srof':5,'sglc':6,'swav':7,'ww3':7,'sesp':8})
+        >>> import os, shutil, tempfile
+        >>> workdir = tempfile.mkdtemp()
+        >>> caseroot = os.path.join(workdir, 'caseroot')  # use non-existent caseroot to avoid error about not being a valid case directory in Case __init__ method
+        >>> Case(caseroot, read_only=False)._valid_compset_impl('2000_DATM%NYF_SLND_DICE%SSMI_DOCN%DOM_DROF%NYF_SGLC_SWAV', None, ['CPL', 'ATM', 'LND', 'ICE', 'OCN', 'ROF', 'GLC', 'WAV', 'ESP'], {'datm':1,'satm':1,'dlnd':2,'slnd':2,'dice':3,'sice':3,'docn':4,'socn':4,'drof':5,'srof':5,'sglc':6,'swav':7,'ww3':7,'sesp':8})
```
billsacks, line 594: In addition to the general comment above: For these doctests, it looks like they were creating a case in the current directory (wherever that happens to be), which feels like it could cause trouble in some circumstances, so it felt safer to fix these tests.

scripts/lib/CIME/tests/test_utils.py
```diff
@@ -7,7 +7,7 @@
 from unittest import mock
 from CIME.utils import indent_string, run_and_log_case_status
 
-from . import utils
+from CIME.tests import utils
```
billsacks, line 10: Relative import doesn't work if you try to run this as a standalone script

scripts/lib/CIME/tests/test_provenance.py
```diff
@@ -1,10 +1,12 @@
+#!/usr/bin/env python3
+
 import os
 import sys
 import unittest
 
 from CIME import provenance
 
-from . import utils
+from CIME.tests import utils
```
billsacks, line 9: Relative import doesn't work if you try to run this as a standalone script

billsacks:
As noted here https://github.com/ESMCI/cime/pull/4009#pullrequestreview-688928542 there will be conflicts with #4009 in at least a couple of files; we should keep the version in this PR for the conflicted shebang lines.

jedwards4b (approved):

jedwards4b:
@billsacks Thanks!

jedwards4b:
@jasonb5 Is this still waiting for your review?

billsacks:
I just merged latest master into this and resolved some minor conflicts.

billsacks:
I just merged up to the latest master, resolving conflicts. @jasonb5 I assume that your LGTM comment means you approve even though you didn't formally approve this PR? If I haven't heard from you by, say, mid next week, I'd like to go ahead and merge this.

jasonb5 (approved):
Forgot to formally approve this, it's good to go though.

PR #4009, opened 2021-06-21
More python churn

### LINKED ISSUE #3987: More shebang lines that need to be changed to python3
scripts_regression_tests failed for me due to a python3 feature I (accidentally) introduced. This made me realize that there are still a lot more shebang lines that should be changed to python3 beyond those changed in #3948 .

e.g., do this from the top level of cime:
```
git grep 'usr/bin/env python' | grep -v python3
```

Many files are ones that I think are going to be split out of cime with @fischer-ncar 's cime split. So it might make sense to wait to fix this until that's done.

@fischer-ncar would you be willing to take this on after you finish the cime split?

jgfouca:
The python situation in CIME has become untenable on some of our Sandia machines. We have some E3SM ecosystem projects that are using python3 for their scripting and also trying to run CIME test suites and the py2/py3 situation is causing problems. This PR aims to remove all potential python2 invocations from CIME with the exception of some stuff from `tools` which is not part of the CIME CCS.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3987 

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/config/cesm/machines/template.case.run
+++ b/config/cesm/machines/template.case.run
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # Batch system directives
 {{ batchdirectives }}
 
--- a/config/cesm/machines/template.case.test
+++ b/config/cesm/machines/template.case.test
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 {{ batchdirectives }}
 """
 This is the system test submit script for CIME. This should only ever be called
--- a/config/cesm/machines/template.st_archive
+++ b/config/cesm/machines/template.st_archive
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python3
 # Batch system directives
 {{ batchdirectives }}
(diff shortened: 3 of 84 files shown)


billsacks (approved):
Thanks for this @jgfouca . Note that this resolves #3987 , which was assigned to @fischer-ncar , so looping him in here. As noted in that issue, I think some of the impacted files are about to be removed from cime, though I'm not positive. I'll let @fischer-ncar add any comments if he'd like.

Also note that there will be some conflicts with #4008 . I changed the shebang lines of unit test files there. Importantly, there were a couple of unit test files with malformed shebang lines (missing `!`), which I fixed in that PR, so when we resolve conflicts, we should keep the version from that PR.

gold2718:
FWIW, tools/utils/find_circular_dependency.py works just fine if you change the shebang (just tested it with and without a circular dependency).

jedwards4b (approved):

fischer-ncar:
@billsacks it looks like the only files that are impacted by my move are the buildnml and buildexe scripts.

PR #4013, opened 2021-06-25
code to check and download neon data to rundir

jedwards4b:
Changes the inputdata structure to add a field "inventory", currently this field is only used for NEON tower data and is a pointer to a CSV format file with filenames and modification dates for all available data on the server.  This file was deemed necessary because query access to the data server is not allowed.   This also extends the data download options so that if the data path begins with the case RUNDIR download is allowed.  Previously only DIN_LOC_ROOT was allowed. 

Test suite: Hand tested with SMS_D_Vnuopc_Mmpi-serial.CLM_USRDAT.I1PtClm51Bgc.cheyenne_intel.clm-NEON-NIWO , scripts_regression_tests.py

Test baseline: 
Test namelist changes: 
Test status: bit for bit,

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/config/cesm/config_inputdata.xml
+++ b/config/cesm/config_inputdata.xml
@@ -40,6 +40,9 @@
     <comment> NEON Tower data for datm </comment>
     <protocol>wget</protocol>
     <address>https://s3.data.neonscience.org/neon-ncar/NEON/</address>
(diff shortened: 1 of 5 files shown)


billsacks (approved):
Looks good, thanks for these enhancements!

PR #4015, opened 2021-06-25
Updates from imp to importlib

### LINKED ISSUE #4007: imp module is deprecated; replace with importlib
I notice that we have a few uses of the 'imp' module in CIME. According to python documentation, this was deprecated as of python3.4, and we should switch to using importlib.

That said, I'm adding another use of the 'imp' module that follows our current patterns, because it seems like it will be easiest to stay consistent and then switch everything at once. The right way to switch our usage of imp.load_source over to importlib seems to depend on the python versions we want to support. According to https://stackoverflow.com/questions/67631/how-to-import-a-module-given-the-full-path, for python3.5+, we'll need this lovely bit of code (apparently the python maintainers really don't want you to do this):

```python
import importlib.util
spec = importlib.util.spec_from_file_location("module.name", "/path/to/file.py")
foo = importlib.util.module_from_spec(spec)
spec.loader.exec_module(foo)
```

See also https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly, which adds a line, `sys.modules[module_name] = module`; I'm not sure if that's needed.

So I guess, if we are comfortable only supporting python3.5+, we should add a utility function that does the above and then switch our current usage to use that.

Thoughts?

jasonb5:
Updates the use to imp to importlib.

Test baseline: n/a
Test namelist changes: n/a
Test status: bfb

Fixes #4007

User interface changes?: n/a

### DIFF
--- a/scripts/lib/CIME/build.py
+++ b/scripts/lib/CIME/build.py
@@ -1,9 +1,12 @@
 """
 functions for building CIME models
 """
-import glob, shutil, time, threading, subprocess, imp
+import glob, shutil, time, threading, subprocess
 from CIME.XML.standard_module_setup  import *
-from CIME.utils                 import get_model, analyze_build_log, stringify_bool, run_and_log_case_status, get_timestamp, run_sub_or_cmd, run_cmd, get_batch_script_for_job, gzip_existing_file, safe_copy, check_for_python, get_logging_options
+from CIME.utils                 import get_model, analyze_build_log, \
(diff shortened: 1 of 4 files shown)


jedwards4b (approved):

billsacks:
Thanks @jasonb5 ! I also see uses of the 'imp' module in build.py and utils.py.

jasonb5:
@billsacks Thanks! I'll take care of those.

billsacks (approved):
Looks great, thanks!

PR #4016, opened 2021-06-28
Set default python to 'python3' for scripts tests CMake file

jgfouca:
Test suite: by-hand
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:


jgfouca:
Since we've already switched nearly all of CIME to use python3, I'm thinking no one will object to this. Merging.

PR #4020, opened 2021-06-29
use testmod default when usermod in test

jedwards4b:
This PR automatically includes the primary component default testmod if it exists when a
test is invoked with a user mod. 

Test suite: scripts_regression_test and hand testing with SMS_Vnuopc_Mmpi-serial.CLM_USRDAT.I1PtClm51Bgc.cheyenne_intel.clm-NEON-JERC (and other NEON sites)

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:


ekluzek (approved):
We talked about this at the CSEG meeting. There was agreement on adding this for usermods that don't already have testmods for them. testmods that already duplicate a usermod can control the include hierarchy on their own. And in CTSM for example we explicitly use the "basic" testmod rather than "default" for some cases.

billsacks (changes requested):
Now that I look at the implementation, I think I understand the concern that @ekluzek raised in the CSEG meeting. In that meeting, my understanding was that this change – i.e., adding the 'default' directory – would only apply to testmods that ended up resolving to a usermods directory instead of the testmods directory. But if I'm understanding the implementation correctly, it looks like this change applies to *all* testmods. We do not want that in CTSM: testmods that don't include the 'default' directory are often set up that way deliberately.

What I thought you were doing based on the summary you gave in the CSEG meeting was: If a test has a testmod, and that testmod resolves to be a directory in the usermods space instead of the testmods space, then also apply the 'default' testmod from the testmod space. If you do it that way (or explain to me how the current implementation is already doing it that way, if I'm misunderstanding), then I'm happy with this feature. (I see pros and cons: I don't love that it leads to implicit behavior that may be confusing to someone who isn't expecting it, but I see how it can make testing a usermod significantly easier.)

One comment on the implementation details: I feel it would be clearer if, instead of adding the 'default' testmod to the list of include_dirs, you instead called `apply_user_mods` one extra time pointing to the 'default' testmod directory. As noted in the docstring for `apply_user_mods`, it can be called multiple times, with settings from later calls taking precedence over earlier calls. So the call to `apply_user_mods` for the 'default' directory should happen first (so that its settings can be overridden). In contrast, with your current implementation, the 'default' directory gets implicitly added to the include list of *every* user mod in the `all_user_mods` list, which (at least for me) makes it confusing to reason about its precedence relative to other mods.

billsacks:
I think my previous comment could be addressed by allowing multiple `--user-mods-dir` arguments to `--create_newcase`, then adding the necessary logic in `test_scheduler.py`, in the conditional block added in https://github.com/ESMCI/cime/pull/4013, to look for the 'default' testmods directory and adding it when using a usermods directory.

However, I also wonder about the feasibility of an alternative implementation: Would it work to allow multiple testmods in a test name? It seems like the main additional thing this would require would be: In the parsing of a test name, allow any number of final testmods rather than just one. (It may also require extending the schema for testlist.xml.) A downside would be that this limits our ability to add other components of a test name in the future. However, a significant upside would be that it would help considerably with our management of testmods, at least in CTSM.

rljacob:
The title and description are confusing me.  How does one use a usermod in an invocation of create_test ?

jedwards4b:
@billsacks The intent was to not include this in tests but I just tried and it isn't working yet.  Hang on I'll fix it.   I think that your alternatives are difficult at best.  
@rjacob If a testmod is specified and the test-mod is not found, a user-mod will be searched for and used if it exists - this is already on master - this PR extends that so that if a user-mod is specified in a test then the testmods/default is also used if it exists.

jedwards4b:
@billsacks It now only does this if the mod is not already a test-mod.

jedwards4b:
superceeded by #4031

PR #4027, opened 2021-07-06
Additional support for ww3 as a component in the e3sm config file

jonbob:
Add more support for ww3 as a component by including it in the 
COMPSETS_SPEC_FILE and PES_SPEC_FILE entries in the e3sm version of
config_files.xml, needed by the Ocean NGD project. Limited testing showed
no impact in typical E3SM use.

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?:  N

Update gh-pages html (Y/N)?:

### DIFF
--- a/config/e3sm/config_files.xml
+++ b/config/e3sm/config_files.xml
@@ -260,6 +260,7 @@
       <value component="mali"     >$COMP_ROOT_DIR_GLC/cime_config/config_compsets.xml</value>
       <value component="mpassi"   >$COMP_ROOT_DIR_ICE/cime_config/config_compsets.xml</value>
       <value component="mosart"   >$COMP_ROOT_DIR_ROF/cime_config/config_compsets.xml</value>
+      <value component="ww3"      >$COMP_ROOT_DIR_WAV/cime_config/config_compsets.xml</value>
     </values>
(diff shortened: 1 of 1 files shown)


config/e3sm/config_files.xml
```diff
@@ -279,6 +280,7 @@
       <value component="mpaso"    >$SRCROOT/cime_config/allactive/config_pesall.xml</value>
       <value component="mali"     >$SRCROOT/cime_config/allactive/config_pesall.xml</value>
       <value component="mpassi"   >$SRCROOT/cime_config/allactive/config_pesall.xml</value>
+      <value component="ww3"      >$SRCROOT/cime_config/allactive/config_pesall.xml</value>
```
rljacob, line 283: Maybe this should be the time we start to move component specific layouts to their own file? since its a new component  If you have ones for ww3 standalone, they would go in $SRCROOT/components/ww3/cime_config/config_pes.xml

rljacob (approved):

config/e3sm/config_files.xml
jonbob: Agreed, that probably makes sense.

PR #4042, opened 2021-07-15
fix testmods path and fail if xmllint not found

### LINKED ISSUE #4039: Have cime abort if xmllint is not found as it fails anyway
Jackie ran into the following problem on izumi, where because she didn't have xmllint in her path her case failed to be created.

This is with branch_tags/cime5.8.32_a02 but it looks like the same problem would be on master as well.

> job is case.run USER_REQUESTED_WALLTIME None USER_REQUESTED_QUEUE None WALLTIME_FORMAT %H:%M:%S
> 
> job is case.st_archive USER_REQUESTED_WALLTIME None USER_REQUESTED_QUEUE None WALLTIME_FORMAT %H:%M:%S
> 
>  Creating Case directory /home/jkshuman/FATES_cases/C4Gr_TropTrees/BCI_ens_test/fire_test_main_4x5_2a90c206_64c557e0
> 
> 
> Setting Environment OMP_STACKSIZE=64M
> 
> Setting Environment PATH=/usr/lib64/qt-3.3/bin:/usr/local/bin:/usr/local/sbin:/usr/sbin:/cluster/torque/bin
> 
> Setting Environment ESMFMKFILE=/home/dunlap/ESMF-INSTALL/8.0.0bs16/lib/libg/Linux.intel.64.mvapich2.default/esmf.mk
> 
> Setting resource.RLIMIT_STACK to -1 from (8388608, -1)
> 
> xmllint not found, could not validate file /home/jkshuman/git/fates_main_api/cime/config/cesm/config_files.xml
> 
> xmllint not found, could not validate file /home/jkshuman/git/fates_main_api/cime/config/cesm/machines/config_machines.xml
> 
> xmllint not found, could not validate file /home/jkshuman/git/fates_main_api/cime/config/cesm/machines/config_compilers.xml
> 
> /home/jkshuman/FATES_cases/C4Gr_TropTrees/BCI_ens_test/fire_test_main_4x5_2a90c206_64c557e0/env_mach_specific.xml already exists, delete to replace
> 
> /home/jkshuman/FATES_cases/C4Gr_TropTrees/BCI_ens_test/fire_test_main_4x5_2a90c206_64c557e0/env_mach_specific.xml already exists, delete to replace
> 
> ERROR: File /home/jkshuman/FATES_cases/C4Gr_TropTrees/BCI_ens_test/fire_test_main_4x5_2a90c206_64c557e0/env_mach_pes.xml appears to have changed without a corresponding invalidation.
> 
> Traceback (most recent call last):
> 
>   File "./xmlchange", line 225, in <module>

### LINKED ISSUE #4041: testmods path wrong in nuopc test
./create_test SMS_Vnuopc.f19_g17.X.casper_gnu.drv-smartsim 
should look in components/cmeps/cime_config/testdefs/testmods_dirs/drv/smartsim

but I get an error:
 Missing testmod file 'smartsim', checked /glade/work/jedwards/sandboxes/cesm2_x_alpha.smartsim/components/cpl7/driver/cime_config/testdefs/testmods_dirs and /glade/work/jedwards/sandboxes/cesm2_x_alpha.smartsim/components/cpl7/driver/cime_config/usermods_dirs

jedwards4b:
Fails if xmllint not found in path.  Fix testmods dir for cmeps cases by setting _cime_driver prior to looking for testmods

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #4041 
Fixes #4039 

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/XML/generic_xml.py
+++ b/scripts/lib/CIME/XML/generic_xml.py
@@ -363,6 +363,7 @@ class GenericXML(object):
 
         # xmllint provides a better format option for the output file
         xmllint = find_executable("xmllint")
+
         if xmllint is not None:
             if isinstance(outfile, six.string_types):
                 run_cmd_no_fail("{} --format --output {} -".format(xmllint, outfile), input_str=xmlstr)
@@ -549,11 +550,10 @@ class GenericXML(object):
         expect(os.path.isfile(filename),"xml file not found {}".format(filename))
         expect(os.path.isfile(schema),"schema file not found {}".format(schema))
         xmllint = find_executable("xmllint")
-        if xmllint is not None:
-            logger.debug("Checking file {} against schema {}".format(filename, schema))
-            run_cmd_no_fail("{} --xinclude --noout --schema {} {}".format(xmllint, schema, filename))
-        else:
-            logger.warning("xmllint not found, could not validate file {}".format(filename))
+        expect(os.path.isfile(xmllint), " xmllint not found, xmllint is required for cime")
+
+        logger.debug("Checking file {} against schema {}".format(filename, schema))
+        run_cmd_no_fail("{} --xinclude --noout --schema {} {}".format(xmllint, schema, filename))
 
     def get_raw_record(self, root=None):
         logger.debug("writing file {}".format(self.filename))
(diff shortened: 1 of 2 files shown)


billsacks (approved):
Thanks for these fixes!

ekluzek (changes requested):
Thanks for doing this @jedwards4b!

The one suggestion I have for a change is to mention that if xmllint isn't found that $PATH is likely bad. This should point the user into looking at their $PATH and getting it set correctly. Hopefully, they'll then get it set correctly on their own. If you don't say that, they might not know what to do.

jedwards4b:
@ekluzek I have added to the message to indicate that the problem may be in the PATH variable and print out the value of that variable.  I don't think that we should make a value judgement here, just saying what we are looking for and where we are looking should be enough.

ekluzek (approved):

Issue #4050, opened 2021-07-21
replay.sh should not cause fail if it cannot write 


jedwards4b:
I was trying to examine another users case:
```
./pelayout
Traceback (most recent call last):
  File "./pelayout", line 244, in <module>
    _main_func(__doc__)
  File "./pelayout", line 230, in _main_func
    with Case(caseroot, record=True) as case:
  File "/glade/work/turuncu/TESTING/CESM_newalarm/cime/scripts/Tools/../../scripts/lib/CIME/case/case.py", line 104, in __init__
    self.record_cmd()
  File "/glade/work/turuncu/TESTING/CESM_newalarm/cime/scripts/Tools/../../scripts/lib/CIME/case/case.py", line 1683, in record_cmd
    with open(os.path.join(caseroot, "replay.sh"), "a") as fd:
PermissionError: [Errno 13] Permission denied: '/glade/scratch/turuncu/IRT_N3_PM3_Ld7_Vnuopc.f19_g17.BHIST.cheyenne_intel.allactive-defaultio.20210719_160023_50zo6w/replay.sh'
```
I think that instead of failing it should just print a message that the script is not writable

jasonb5:
This was fixed in #4053.

PR #4053, opened 2021-07-22
Fix replay archive

### LINKED ISSUE #4049: replay.sh: Capture script in provenance.py for archiving (E3SM)
We need to archive replay.sh in performance archive so that it can be picked up by PACE for E3SM.
Logic needs to be added to https://github.com/ESMCI/cime/blob/master/scripts/lib/CIME/provenance.py#L400

cc @jasonb5 @rljacob

jasonb5:
- Fixes capturing `replay.sh` in performance archive.
- Fixes handling error when `replay.sh` is not writable.

Test baseline: n/a
Test namelist changes: n/a 
Test status: bfb

Fixes #4049 #4050 

User interface changes?: n/a

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -1729,8 +1729,11 @@ directory, NOT in this subdirectory."""
         lines_len = len(lines)
         lines.insert(lines_len-1 if init else lines_len, "{}\n\n".format(cmd))
 
-        with open(os.path.join(caseroot, "replay.sh"), "a") as fd:
-            fd.writelines(lines)
+        try:
+            with open(os.path.join(caseroot, "replay.sh"), "a") as fd:
+                fd.writelines(lines)
+        except PermissionError:
+            logger.warning("Could not write to 'replay.sh' script")
 
     def create(self, casename, srcroot, compset_name, grid_name,
                user_mods_dirs=None, machine_name=None,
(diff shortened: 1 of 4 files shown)


sarats:
On the E3SM side, we definitely want replay.sh to be created for every case. As this script would be written in a user's case directory, I'm not sure how they might run into permission issues. Perhaps I'm missing something here?

jedwards4b:
@sarats often in the course of debugging issues for users I want to run a command in their case directory.  I attempted to run the pelayout script in a users case and got a permission error.

jedwards4b (approved):

rljacob:
The commit message about a "capturing error" is confusing.  Provenance just wants to copy replay.sh to another directory so only needs read permission on the file.   But if it needs to be writeable for other people to run commands like "preview_run", that makes sense.

jasonb5:
The "capturing error" should have been "handling error", this is a separate fix for #4050. My understanding was a user may want to execute a CIME command in another users directory where they do not have write permission.

The two fixes in this PR are loosely related by the "replay.sh" script.

rljacob:
@jgfouca make sure this is also in your next CIME update to E3SM.

jgfouca:
@rljacob I was thinking the same thing, I just pinged Jason to fix the conflict. I'm doing a CIME update the minute this gets merged.

sarats:
It would be good to get the new Git files as well for provenance. We can try asking WC to update their v2 tag.

jgfouca:
@sarats , what needs to happen to get those files? Do we have an issue or PR for that?

rljacob:
PR #4055 did that.  Just need to get the newest CIME master in to E3SM and then have the WC team update their clone.  I don't think they'll need to rebuild their executable for everything to work.

jasonb5:
@jgfouca This is ready to merge.

PR #4057, opened 2021-07-28
Prepend the casename to mpas component restart file search in st_archive

jonbob:
Changes the pattern used to search for mpas/mali restart files to 
include the casename prepended to the current specification. All the
MPAS components modified their output filenames to include the
casename some time ago as other components do. While the st_archive
script continues to function correctly for the most part, the has been an
issue reported for hybrid configurations and this fixes it.

Test suite: 
Test baseline: 
Test namelist changes: 

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/case/case_st_archive.py
+++ b/scripts/lib/CIME/case/case_st_archive.py
@@ -413,7 +413,7 @@ def _archive_restarts_date_comp(case, casename, rundir, archive, archive_entry,
 #        logger.debug("suffix is {} ninst {}".format(suffix, ninst))
         restfiles = ""
         if compname.find('mpas') == 0 or compname == 'mali':
-            pattern = compname + r'\.' + suffix + r'\.' + '_'.join(datename_str.rsplit('-', 1))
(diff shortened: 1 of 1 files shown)


jonbob:
@jgfouca - can I ask you to run any required tests?

scripts/lib/CIME/case/case_st_archive.py
```diff
@@ -413,7 +413,7 @@ def _archive_restarts_date_comp(case, casename, rundir, archive, archive_entry,
 #        logger.debug("suffix is {} ninst {}".format(suffix, ninst))
         restfiles = ""
         if compname.find('mpas') == 0 or compname == 'mali':
```
jedwards4b, line 415: @jonbob  - if I understand this correctly mpas now conforms to the standard naming convention and doesn't need to have a special block here?

jonbob: @jedwards4b - close but not completely. The MPAS components still have a different date string.

jedwards4b: that's unfortunate - thanks for clarifying.

rljacob (approved):

PR #4058, opened 2021-07-30
Support the creation of multiple user_nl files for a component

### LINKED ISSUE #3999: Support the creation of multiple user_nl files for a component
CISM in CESM will soon support running multiple ice sheets in the same case. Each of these ice sheets will have its own config file (similar to a namelist file), with different settings. Therefore, in a run containing Greenland and Antarctica (for example) we need a way for users to be able to separately specify settings in the `user_nl_cism` file that apply (1) to both ice sheets; (2) just Greenland; (3) just Antarctica.

Based on discussion here https://github.com/ESCOMP/CISM-wrapper/issues/50 the user_nl_cism file will look like this:

```
! Settings that apply to all ice sheets go here
myvar = myvalue

[gris]
! Settings that apply to Greenland (if present) go here
foo = value1
bar = value2

[ais]
! Settings that apply to Antarctica (if present) go here
foo = value3
baz = value4
```

I propose adding a function to CIME (in `user_mod_support.py`) to parse a file like this, outputting a new file that just contains the requested subset of settings. In CISM's buildnml, we would then call this function three times:

First, we'd request a file that just contains settings that apply to all ice sheets (this is needed to create the top-level `cism_in` file): `separate_user_nl(..., section=None)`, which would produce:

```
! Settings that apply to all ice sheets go here
myvar = myvalue
```

Then we would call `separate_user_nl(..., section='gris')`, which would produce:

```
! Settings that apply to all ice sheets go here
myvar = myvalue

[gris]
! Settings that apply to Greenland (if present) go here
foo = value1
bar = value2
```

Finally, we would call `separate_user_nl(..., section='ais')`, which would produce:

```
! Settings that apply to all ice sheets go here
myvar = myvalue

[ais]
! Settings that apply to Antarctica (if present) go here
foo = value3
baz = value4
```

(Note: It's intentional that the settings that apply to all ice sheets appear in all cases. This section can contain a mix of settings that (a) go in `cism_in`, which contains non-icesheet-specific settings, and (b) go in *all* `cism.config` files – i.e., ice sheet-specific settings, but for which the user wants the same setting to apply to all ice sheets.)

I hope to implement this in the next week or so. I am opening this to gather any feedback on:

1. **Are there any objections to putting this functionality in CIME?** I could put it in CISM, but I prefer putting it in CIME for two reasons: (i) I would like to cover this functionality with unit tests; putting it in CISM would require me to build up a new python unit testing infrastructure in CISM. (ii) Maybe this could be useful for other components for some related or different purpose.

2. **Are there any objections to the format of the `user_nl` file proposed above?** We are not completely tied to this format, though we (@whlipscomb, @mvertens and I) prefer it over other options we considered (see https://github.com/ESCOMP/CISM-wrapper/issues/50).

billsacks:
In CESM, we are extending CISM to allow multiple user_nl_cism* files: an overall user_nl_cism, plus a separate file for each ice sheet (user_nl_cism_ais, user_nl_cism_gris, etc.). This PR adds support for staging multiple user_nl files for a component at case.setup time, via an optional function that can be defined in the component's buildnml. See https://github.com/ESCOMP/CISM-wrapper/blob/23b9a83c81db4a952f889496f5ca1dbe62f91178/cime_config/buildnml#L66-L78 for the definition of this function for CISM.

To support this change, I have rewritten the CIME functionality that determines if a given function is present in a python module so that it is more robust and covered by unit tests.

Test suite:
- scripts_regression_tests on izumi: most tests pass; for failures, there were similar failures in a run from an unmodified copy of cime6.0.3 (some of the same tests failed, though there were a few differences that appeared to be random system issues)
- manual testing of the changes to `run_sub_or_cmd`, ensuring that the `file_contains_python_function` logic is working correctly
- extensive manual testing of the case_setup changes in this PR together with https://github.com/ESCOMP/CISM-wrapper/tree/multiple_user_nl_cisms/ – both with and without the new `get_user_nl_list` function

Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes #3999

User interface changes?: Allows components to define a `get_user_nl_list` function in their buildnml file, which supports the creation of multiple user_nl files for the component


scripts/lib/CIME/case/case_setup.py
```diff
@@ -50,24 +52,60 @@ def _build_usernl_files(case, model, comp):
...
+        default_nlfile = "user_nl_{}".format(comp)
+        model_nl = os.path.join(model_dir, default_nlfile)
+        user_nl_list = _get_user_nl_list(case, default_nlfile, model_dir)
+        # Note that, even if there are multiple elements of user_nl_list (i.e., we are
+        # creating multiple user_nl files for this component with different names), all of
+        # them will start out as copies of the single user_nl_comp file in the model's
+        # source tree.
+        for nlfile in user_nl_list:
+            if ninst > 1:
+                for inst_counter in range(1, ninst+1):
```
jedwards4b, line 64: Have you run any multiinstance tests to confirm that that functionality is still okay?  Multiinstance with multiple ice sheets?

scripts/lib/CIME/tests/test_utils.py
```diff
@@ -1,15 +1,17 @@
...
+import shutil
 import sys
 import tempfile
 
 import unittest
 from unittest import mock
 from CIME.utils import indent_string, run_and_log_case_status, \
-    import_from_file
+    import_from_file, \
+    _line_defines_python_function, file_contains_python_function
```
jedwards4b, line 12: should we remove the underscore in _line_defines_python_function since we are calling it from outside CIME.utils?

jedwards4b (approved):

scripts/lib/CIME/case/case_setup.py
billsacks: I didn't run tests all the way through the build & run, but I did run through case.setup with multi-instance and verified that the appropriate user_nl files were staged in all cases. I did the following tests both with and without the new `get_user_nl_list` function defined:
- Initial case.setup (so no user_nl files present yet in the case)
- Destination file (e.g., user_nl_cism_gris_0001) already present with some contents: verified that this is not overwritten
- Destination files (e.g., user_nl_cism_gris_0001) absent, but single-instance version present with some contents (e.g., user_nl_cism_gris): verified that the contents from the single-instance version were copied into all multi-instance versions

I checked single instance with multiple ice sheets. I didn't check multi-instance with multiple ice sheets, but note that even with a single ice sheet we now have multiple user_nl_cism files, so my tests of multi-instance did cover the situation where a component has multiple user_nl files.

scripts/lib/CIME/tests/test_utils.py
billsacks: My thinking was that, since this should only be accessed by unit tests and not any external production code, I would keep it with the underscore. But if you feel I should remove the leading underscore due to its access by the unit tests, I'm happy to do so.

scripts/lib/CIME/case/case_setup.py
jedwards4b: sounds good thanks

scripts/lib/CIME/tests/test_utils.py
jedwards4b: No, fine as written.

jgfouca (approved):

billsacks:
@jedwards4b can I merge this branch?

jedwards4b:
@billsacks sure

PR #4066, opened 2021-08-10
Fix namelist duplications

### LINKED ISSUE #3795: Duplicate call to set_nodes in nmlgen:set_defaults looks suspicious
I'm not 100% sure this is a bug, but it looks like one to me... if so, my guess is that it's just slowing things down a bit, probably not causing incorrect results, but it could still be worth fixing:

It looks to me like set_nodes is unnecessarily called twice here, and that only the second call (which saves the return results) should be done:

https://github.com/ESMCI/cime/blob/c258cdf083e55c950b47562408e96a27a8411b74/scripts/lib/CIME/nmlgen.py#L106-L109

But again, I'm not 100% sure that this is a bug: it may be that it needs to be called twice for some reason.

It looks like this came in with @mvertens 's overhaul of this code in #965. @mvertens can you comment on this?

jasonb5:
Fixes issue that was effectively creating 4x as many entries.

Test baseline: n/a
Test namelist changes: n/a 
Test status: bfb

Fixes #3795 

User interface changes?: N


billsacks (approved):
Thanks a lot - looks good to me!

jasonb5:
@billsacks This is ready to merge should contain the fixes for failing tests in #4071.

fischer-ncar:
I see the real problem now.  I've removed the testing for f19_g16.  So it would probably be better to remove the if statement for nuopc and just use --res f19_g17.

jasonb5:
@fischer-ncar @billsacks Would it be sufficient to do as @fischer-ncar suggested and remove the `--run-unsupported` flag?

fischer-ncar:
I'm testing scripts_regression_tests.py J_TestCreateNewcase right now to see if my suggestion does work.

fischer-ncar:
@jasonb5 @billsacks 
I just looked through scripts_regression_tests.  There's several other tests that are using f19_g16  But those tests also add.

        if CIME.utils.get_model() == "cesm":
            args += " --run-unsupported"

So I would say for now leave the --run-unsupported you added.  Then the bigger question becomes, do we keep testing f19_g16 resolution, or switch to f19_g17.  Which I think we should.  But that's not a question we should answer on a Friday afternoon.

fischer-ncar:
My J_TestCreateNewcase did pass.

billsacks:
I like having `--run-unsupported` in any case, because otherwise these tests will be prone to breakage in the future if there are other changes to what is and isn't tested in our test suites.

Changing the tests is also a good idea, but can be done in a follow-up set of changes.

PR #4070, opened 2021-08-11
Issue #2644 solved.

### LINKED ISSUE #2644: In documentation, "Addendum" should be changed to "Appendices"
The cime documentation lists "addendum" at the end. But I think the use of this term is incorrect. I think this really means "Appendices". See http://www.learnersdictionary.com/qa/what-s-the-difference-between-an-appendix-and-an-addendum .

Sukhamjot-Singh:
Changed Addendum to Appendices

Test suite: none
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes #2644 

User interface changes?: N

Update gh-pages html (Y/N)?: Y (needed)

### DIFF
--- a/doc/source/index.rst
+++ b/doc/source/index.rst
@@ -23,7 +23,7 @@ Table of contents
    build_cpl/index.rst
    misc_tools/index.rst
 
-Addendum
+Appendices
 ---------
 .. toctree::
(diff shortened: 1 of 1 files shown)


billsacks:
Thank you for your contribution!

PR #4079, opened 2021-09-02
if nuopc and any datamodel add a cdeps directory

jedwards4b:
Add a SourceMods/src.cdeps directory for shared cdeps sourcemods

Test suite: scripts_regression_tests.py and hand testing of SMS.f19_g17.A.cheyenne_intel
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes  https://github.com/ESCOMP/CDEPS/issues/98

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/case/case.py
+++ b/scripts/lib/CIME/case/case.py
@@ -1303,7 +1303,11 @@ class Case(object):
 
     def _create_caseroot_sourcemods(self):
         components = self.get_compset_components()
+        datamodels = [x for x in components if x.startswith('d')]
         components.extend(['share', 'drv'])
+        if self._comp_interface == 'nuopc' and datamodels:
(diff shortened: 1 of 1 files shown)


jedwards4b:
@mvertens can you confirm this solves the issue and review.   Note that component specific cdeps files go in the component subdirectory and non-component specific files go in the added src.cdeps directory.

billsacks:
CDEPS is built even if you don't have any data models in your case, right? (Because it's needed for streams functionality, right?). So I'm thinking that there should always be a cdeps SourceMods directory when building with driver nuopc?

jedwards4b:
Right - that's a good point, I'll remove that conditional.  Thanks

billsacks (approved):
Thanks!

PR #4081, opened 2021-09-03
Initial changes to config_compilers/config_machines for container w/ GNU

jedwards4b:
Adds a container machine definition.  
Corrects an issue with determining the path to cime which caused an erroneous warning to be output if a linked directory was referred to in two different ways.  

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:


jedwards4b:
@briandobbins I think that you may want to add your changes to config_inputdata.xml here.

config/cesm/machines/config_machines.xml
```diff
@@ -862,7 +862,7 @@ This allows using a different mpirun command to launch unit tests
     <OS>LINUX</OS>
     <COMPILERS>gnu</COMPILERS>
     <MPILIBS>mpich</MPILIBS>
-    <CIME_OUTPUT_ROOT>$CASEROOT/</CIME_OUTPUT_ROOT>
+    <CIME_OUTPUT_ROOT>$ENV{HOME}/$CASE</CIME_OUTPUT_ROOT>
```
briandobbins: How about we go with /scratch/${USER} instead, which is close to Cheyenne?  I'm not fond of forcing it into the home directory, since if someone creates a case in /home/user/foobar/test1, this creates a confusing directory in /home/user/test1 with the bld/run directories, right?  Secondly, this should allow us to also map in /scratch/user directories for people who want to mount some other storage for run stuff (faster storage, perhaps, like an SSD).

This may require simple changes to check / make that directory if it doesn't exist, but I think that's entirely in the container process.

jedwards4b: So then you would create a /scratch file system in the container?  I'm fine with that.

briandobbins: Either /scratch or /home/scratch seems fine.  Maybe one is better than the other, but I need to think about that more.  For now, this all looks good, so I'll approve and we'll return to this if we need to change it.  (You had the ${HOME}/scratch in a later commit I hadn't seen when going through them chronologically.)

PR #4085, opened 2021-09-08
Fix scripts_regression_tests on mappy for e3sm

### LINKED ISSUE #4083: CIME mappy scripts_regression_tests in bad shape
I'm seeing tons of fails:

```
FAILED (failures=48, errors=7, skipped=21)
```

jgfouca:
1) We have a newer version of pylint that's picking up a lot of
   new warnings that aren't useful, so disable arguments-renamed.
2) The test_case.py test was changing CIME_MODEL, which caused all
   subsequent tests to fail for e3sm since the model was set to cesm.
   Use mock instead to set this env var.
3) pesfile was wrong for some tests for e3sm
4) test_clean_build was using a grid alias that is unknown to e3sm. Use
   a more common one instead.
5) Many additional changes are needed on the E3SM side which will be pushed
   to that repo soon.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #4083 

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/code_checker.py
+++ b/scripts/lib/CIME/code_checker.py
@@ -19,7 +19,7 @@ def _run_pylint(on_file, interactive):
 
     cmd_options = " --disable=I,C,R,logging-not-lazy,wildcard-import,unused-wildcard-import"
     cmd_options += ",fixme,broad-except,bare-except,eval-used,exec-used,global-statement"
-    cmd_options += ",logging-format-interpolation,no-name-in-module"
+    cmd_options += ",logging-format-interpolation,no-name-in-module,arguments-renamed"
     cimeroot = get_cime_root()
     srcroot = get_src_root()
 
--- a/scripts/tests/scripts_regression_tests.py
+++ b/scripts/tests/scripts_regression_tests.py
@@ -166,8 +166,6 @@ class A_RunUnitTests(unittest.TestCase):
                     else:
                         print("{} has no doctests".format(filepath))
 
-        self.assertTrue(get_model() == "e3sm")
-
 ###############################################################################
 class C_TestGridGeneration(unittest.TestCase):
 ###############################################################################
--- a/scripts/tests/scripts_regression_tests.py
+++ b/scripts/tests/scripts_regression_tests.py
@@ -24,7 +24,7 @@ import collections
 
 from CIME.utils import run_cmd, run_cmd_no_fail, get_lids, get_current_commit, \
     safe_copy, CIMEError, get_cime_root, get_src_root, Timeout, \
-    import_from_file
+    import_from_file, get_model
 import get_tests
 import CIME.test_scheduler, CIME.wait_for_tests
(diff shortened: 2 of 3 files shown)


jasonb5 (approved):

jgfouca:
With this PR and my E3SM PR, we should be back to just the ERIO and ERI tests failing for mappy Z_FullSystemTest; everything else should pass.

scripts/lib/CIME/code_checker.py
```diff
@@ -19,7 +19,7 @@ def _run_pylint(on_file, interactive):
 
     cmd_options = " --disable=I,C,R,logging-not-lazy,wildcard-import,unused-wildcard-import"
     cmd_options += ",fixme,broad-except,bare-except,eval-used,exec-used,global-statement"
-    cmd_options += ",logging-format-interpolation,no-name-in-module"
+    cmd_options += ",logging-format-interpolation,no-name-in-module,arguments-renamed"
```
jedwards4b, line 22: will this present a problem for older pylint versions?

scripts/tests/scripts_regression_tests.py
```diff
@@ -164,6 +166,8 @@ def test_lib_doctests(self):
                     else:
                         print("{} has no doctests".format(filepath))
 
+        self.assertTrue(get_model() == "e3sm")
```
jedwards4b, line 169: How will this work for cesm?

scripts/tests/scripts_regression_tests.py
```diff
@@ -2580,7 +2584,7 @@ class G_TestBuildSystem(TestCreateTestCommon):
     ###########################################################################
     def test_clean_rebuild(self):
     ###########################################################################
-        casedir = self._create_test(["--no-run", "SMS.f19_g17_rx1.A"], test_id=self._baseline_name)
+        casedir = self._create_test(["--no-run", "SMS.f19_g16_rx1.A"], test_id=self._baseline_name)
```
jedwards4b, line 2587: I think that this needs to be g17 for cesm, we have dropped the g16 grid.

jedwards4b (changes requested):

scripts/tests/scripts_regression_tests.py
jgfouca: @jedwards4b , sorry, this line was mistakenly left in the code. I just removed it.

scripts/tests/scripts_regression_tests.py
jgfouca: This grid is used in cime_developer test suite, so it should work for all models.

jedwards4b (approved):

scripts/lib/CIME/code_checker.py
jgfouca: I don't think so, I just checked that adding a disabling of an unknown warning is not an error.

PR #4088, opened 2021-09-09
Change E3SM to use cmake macro file system

### LINKED ISSUE #2965: Add compiler modifier for MAX_TASKS_PER_NODE, MAX_MPITASKS_PER_NODE, and mpirun arguments
On Summit, we have an issue where the PE layout will be different for GPU and CPU runs. CPU runs should use 84 tasks per node while GPU runs should use 36 or fewer. I'd like to use the compiler as the modifier for this, e.g.,

```
<MAX_TASKS_PER_NODE compiler="pgigpu">36</MAX_TASKS_PER_NODE>
<MAX_MPITASKS_PER_NODE compiler="pgigpu">36</MAX_MPITASKS_PER_NODE>
<MAX_TASKS_PER_NODE compiler="!pgigpu">84</MAX_TASKS_PER_NODE>
<MAX_MPITASKS_PER_NODE compiler="!pgigpu">84</MAX_MPITASKS_PER_NODE>

<mpirun mpilib="spectrum-mpi">
<!-- Use a helper script to tweak jsrun options -->
<executable>/gpfs/alpinetds/world-shared/csc190/e3sm/mpirun.summit</executable>
<!-- <executable>jsrun</executable> -->
<arguments compiler="pgigpu">
  <arg name="num_tasks"> -n {{ total_tasks }} -N $MAX_MPITASKS_PER_NODE -gpu </arg>
</arguments>
<arguments compiler="!pgigpu">
  <arg name="num_tasks"> -n {{ total_tasks }} -N $MAX_MPITASKS_PER_NODE </arg>
</arguments>
</mpirun>
```

It seems that CIME is not currently able to use a compiler modifier for these fields when parsing the XML file, but I think this will be necessary for CIME to support Summit's configurations.

### LINKED ISSUE #3287: Better encapsulation of compiler flag customization
Hi all,

I'm back to spending a lot of cycles look at our build system. One that I noticed is that there's some confusion for concept of customizing compiler flags.

If you want to customize compiler flags, there's many places to look depending on the context:
* If you want a global customization for a file, it's done directly in  tools/Makefile (or CMakeLists.txt) (e.g. rrtmg_lw_k_g.f90)
* If you want compiler/platform specific changes, you use config_compilers.xml
* If you want compiler/platform specific changes for a particular file, you use a Depends file

I'm thinking that it sure would be nice if all compiler flag customizations could be encapsulated by config_compilers.xml. We could enhance it to have filename selectors.

Thoughts?

### LINKED ISSUE #3341: Unexpected behavior with parsing of config_compilers.xml into Macros.cmake (not overriding defaults)
In `config_compilers.xml` We have compiler-specific cases such as:
```
<compiler COMPILER="gnu">
```
where we have all GNU compiler basic flags/settings. If we need to make a change to those settings for a specific machine, my understanding is that we could so something like the following:

```
<compiler MACH="cori-knl" COMPILER="gnu">
  <ALBANY_PATH>/global/project/projectdirs/acme/software/AlbanyTrilinos20190823/albany-build/install</ALBANY_PATH>
  <CONFIG_ARGS>
    <base> --host=Linux </base>
  </CONFIG_ARGS>
  <CPPDEFS>
    <append> -DARCH_MIC_KNL </append>
  </CPPDEFS>
  <CFLAGS>
    <base DEBUG="TRUE"> -g </base>
    <base DEBUG="FALSE"> -O2</base>
    <append compile_threaded="TRUE"> -fopenmp </append>
    <append COMP_NAME="csm_share"> -std=c99 </append>
  </CFLAGS>
  <FFLAGS>
    <base DEBUG="TRUE"> -g</base>
    <base DEBUG="FALSE">  -O2</base>
    <append compile_threaded="TRUE"> -fopenmp </append>
  </FFLAGS>
...
```

Here in the example, I have made the changes more visible (ie wanting DEBUG flags to simply be `-g` for cori-knl with gnu) where in practice my desired change is more subtle. 

It also looks any commands of a "section" in config_compilers.xml is simply added to other commands -- such that even now, I see multiple commands in Macros.cmake that were likely not what was intended (and only work as it's a simple flag append), such as:
 ```
    set(FFLAGS "${FFLAGS}  -O2 -debug minimal")
    set(FFLAGS "${FFLAGS}  -O2 -debug minimal -qno-opt-dynamic-align")
```
where the first command is coming from the `<compiler COMPILER="intel">` section, and the second is coming from `<compiler MACH="cori-knl" COMPILER="intel">` section. Again, it still works, but I would think the expectation is that the settings in `<compiler MACH="cori-knl" COMPILER="intel">` would override the other default settings.

Note I see this in both Macros.cmake and Macros.make. But I sure thought this was NOT happening a while back. Would need to go back to older repos to test.

I hope I've made this clear, but can explain a practical issue to be more concrete.

### LINKED ISSUE #3446: Critical flaw in Macro generation: does not preserve order
Some background: I was debugging a problem on summit and I discovered that I needed to have the pnetcdf link come earlier in the link line, so I changed the config_compilers.xml like this:

FROM:
```
   <SLIBS>
    <append>-lxlopt -lxl -lxlsmp -L$ENV{NETCDF_C_PATH}/lib -lnetcdf -L$ENV{NETCDF_FORTRAN_PATH}/lib -lnetcdff -L$ENV{ESSL_PATH}/lib64 -lessl -L$ENV{NETLIB_LAPACK_PATH}/lib -llapack</append>
     <append MPILIB="!mpi-serial"> -L$ENV{PNETCDF_PATH}/lib -lpnetcdf -L$ENV{HDF5_PATH}/lib -lhdf5_hl -lhdf5 </append>
   </SLIBS>
```

TO:
```
   <SLIBS>
     <append MPILIB="!mpi-serial"> -L$ENV{PNETCDF_PATH}/lib -lpnetcdf -L$ENV{HDF5_PATH}/lib -lhdf5_hl -lhdf5 </append>
    <append>-lxlopt -lxl -lxlsmp -L$ENV{NETCDF_C_PATH}/lib -lnetcdf -L$ENV{NETCDF_FORTRAN_PATH}/lib -lnetcdff -L$ENV{ESSL_PATH}/lib64 -lessl -L$ENV{NETLIB_LAPACK_PATH}/lib -llapack</append>
   </SLIBS>
```

In other words, just moved the pnetcdf SLIBS append earlier. To my surprise, this had no impact on the link flag order generated in Macros.cmake. I checked out BuildTools/macroconditiontree.py and discovered that it does not appear to be order-preserving. Things are being sorted, stored in unordered dicts, and written without regard to the original order.

Since compilation and link flags can be order-sensitive, this strikes me as a fatal flaw... I think the entire BuildTools library may need to be rewritten.

### PR DESCRIPTION
**Pull request by jgfouca**
This is a big change for E3SM but should have no impact on other models and did not require big changes to CIME code.

This PR lays the groundwork for E3SM leaving the config_compilers.xml/BuildTools.configure behind for good, replacing that system with a Cmake-based cache file system that can, if needed, be used to generate a Makefile macro when needed (for some sharedlibs).

I will lay out detail documentation for this change when I create the corresponding PR in E3SM that activates these new macros.

Test suite: scripts_regression_tests (both with and without new macros in the host repo)
Test baseline: 
Test namelist changes: 

Fixes #3287 
Fixes #3446 
Fixes #3341 
Fixes #2965 

User interface changes?: 

Update gh-pages html (Y/N)?:

### CODE REVIEW

File: scripts/lib/CIME/build.py [hunk 1]
```diff
@@ -21,6 +21,38 @@
...
+
+    cmake_args = get_standard_cmake_args(case, "DO_NOT_USE", shared_lib=True)
+    output = run_cmd_no_fail("cmake -DCONVERT_TO_MAKE=ON -DCOMP_NAME={} {} -C Macros.cmake |& grep E3SM_SET_MAKEFILE_VAR | grep -v BUILD_INTERNAL_IGNORE".format(comp_name, cmake_args), from_dir=caseroot)
+
+    # The Tools/Makefile may have already adding things to CPPDEFS and SLIBS
+    real_output = output.replace("E3SM_SET_MAKEFILE_VAR ", "").\
+                  replace("CPPDEFS := ", "CPPDEFS := $(CPPDEFS) ").\
+                  replace("SLIBS := ", "SLIBS := $(SLIBS) ")
+
+    with open(os.path.join(caseroot, "Macros.make"), "w") as fd:
```
jedwards4b, line 46: pylint is complaining unless encoding is added to open statements.

jedwards4b (approved):

scripts/lib/CIME/build.py
jgfouca: @jedwards4b , both my pylint and the CI test machine OK'd this file. Can you paste the pylint warning you are getting?

jedwards4b:
It's coming from pylint 2.10 used in travis-ci and is described here https://pylint.pycqa.org/en/latest/whatsnew/2.10.html
> Added unspecified-encoding: Emitted when open() is called without specifying an encoding

jgfouca:
@jedwards4b , thanks, I will fix that. How come I'm seeing "All checks have passed" here on GitHub?

jedwards4b:
I think that the travis.ci were turned off in cime and turned on in cmeps and travis is the only one using this new version of pylint.  So you probably didn't need to change anything yet.

jgfouca:
@jedwards4b , I'm going to leave it as is then. I grepped around and we have tons of `open` 's without encoding. We will likely need to just disable this warning.

PR #4093, opened 2021-09-17
add cmake macros

jedwards4b:
This extends the work that Jim Foucar did with cmake to allow the inclusion of CMAKE_LANG_COMPILER_VERSION in cmake_macros files and use cmake to generate the case Macros.make file. 
It requires a change in the cdeps CMakeLists.txt for cdeps to build.  I have demonstrated the ability to handle compiler version in the file cmake_macros/gnu.cmake where I add flags only if the gfortran is version 10 or newer.  This is a big change and should get thoroughly tested before adoption. 

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 
User interface changes?: 

Update gh-pages html (Y/N)?:


jgfouca (approved):
Looks great. I'm excited for support for version checking. I imagine that would have been pretty hard to achieve in the prior implementation.

billsacks:
Thanks for your work on this, @jedwards4b !

I would like to leave the detailed review to @fischer-ncar and @jgfouca , but just have a couple of big picture questions:

1. Is the intent that this replaces config_compilers.xml? If so, should that file be removed in this PR, or is there a reason to keep it around?
2. Will this change the method for porting to a user-defined machine? If so, the documentation for that probably needs to be updated – though I'm fine with that being deferred until a separate PR if you don't have time to do that now (with an issue opened as a reminder).

jedwards4b:
@billsacks It does replace config_compilers.xml, that file and I think some python files like compilers.py can be removed once we are comfortable with this change, but I haven't removed them yet.

We will need to update documentation for porting - thanks for the reminder.

jgfouca:
@billsacks , the old system is still offered via env `CIME_NO_CMAKE_MACRO` , so we will need those files until support for that is removed.

billsacks:
Thanks for the clarifications, @jedwards4b and @jgfouca . Can one of you clarify what this means for someone who has an existing cime port via config_compilers and config_machines in a `.cime` directory? Does it mean that they can continue to use their current `.cime` directory as long as they set `CIME_NO_CMAKE_MACRO`? Is there any way to do something like issue a warning if we detect that a `.cime` directory is using the old config_compilers method, instructing the user to (1) set `CIME_NO_CMAKE_MACRO` and (2) consider updating to the new method as soon as possible?

Since I maintain my own cime port on my Mac, I'm thinking about how I'll do this so that – for some time – my single `.cime` directory can support both new and old versions of cime. I'm thinking that I'll have both a `config_compilers.xml` and a cmake macros file for my machine with gnu. So older cime versions will pick up my `config_compilers.xml` and newer ones will pick up my macros file instead. Will that work? If so, then the above suggestion of detecting the old method would probably mean seeing if `.cime` contains `config_compilers.xml` and does not contain any macros files.

jedwards4b:
@billsacks sure I can work on backward compatibility

billsacks:
Thanks @jedwards4b . Also, will this new method support user-defined cmake macros files inside your `.cime` directory? For the LILAC build in CTSM for a user-defined machine, we'll also need it to support finding these files in the directory specified by the `--extra-machines-dir` argument to `create_newcase` (just like that directory is currently searched for config_compilers and config_machines). (At a glance, I don't see support for checking those directories yet, but I may be missing something.)

jgfouca:
@billsacks , great catch. We will need something to support macro changes in .cime.

fischer-ncar (approved):

jedwards4b:
I intend to incorporate suggested changes but I am on PTO next week.

jedwards4b:
I have confirmed that using CIME_NO_CMAKE_MACRO with files in ~/.cime works.  I have also added support for cmake macro files in ~/.cime.

billsacks:
Thanks a lot @jedwards4b . Your latest commit partly addresses what I was asking for, but there are two other things that I don't see:

1. Can you please add support for the `--extra-machines-dir` to `create_newcase`? We need this for CTSM's LILAC build. For example, see

https://github.com/ESMCI/cime/blob/80220601778a29d2254cc2759cfc034350ba0b57/scripts/lib/CIME/XML/compilers.py#L66-L69

for how this is used to add information from a config_compilers.xml file in this extra machines dir. Basically, any time we look for machine-related files in `~/.cime`, we should also check the path given by `--extra-machines-dir` (if any).

2. Thank you for adding the warning if no macros files are found in the `~/.cime` directory. But I'm thinking that the warning should only be issued if a `config_compilers.xml` file is found there (and no macros files are found there): people use `.cime` for purposes other than machine ports, and I'm thinking we don't want to issue a warning in that case.

billsacks (approved):
Thanks a lot for this @jedwards4b . Your most recent commit addresses my final requests - thank you!

jedwards4b:
@jgfouca I've run into a problem.  In cmake_macros/universal.cmake we have 
```
if (MODEL STREQUAL mom)
  string(APPEND FFLAGS " $(FC_AUTO_R8) -Duse_LARGEFILE")
endif()
```
This is not currently being set in the generated Macros.make

jgfouca:
@jedwards4b , I hit that exact same problem in E3SM. You need to use `COMP_NAME` instead of `MODEL`.

jedwards4b:
@jgfouca this would imply changes in Tools/Makefile, since I am trying to maintain backward compatibility I think I need to use MODEL not COMP_NAME.

jedwards4b:
With the latest push the Macros.make for case SMS.T62_t061.GMOM.cheyenne_intel now looks like
```
                                                                                                                                                                                                                                 
# This file is auto-generated, do not edit. If you want to change                                                                                                                                                                
# sharedlib flags, you can edit the cmake_macros in this case. You                                                                                                                                                               
# can change flags for specific sharedlibs only by checking COMP_NAME.                                                                                                                                                           
                                                                                                                                                                                                                                 
CFLAGS :=  -qno-opt-dynamic-align -fp-model precise -std=gnu99 -O2 -debug minimal -qopt-report -xCORE_AVX2 -no-fma                                                                                                               
CPPDEFS := $(CPPDEFS)  -DCESMCOUPLED -DFORTRANUNDERSCORE -DCPRINTEL                                                                                                                                                              
CXX_LDFLAGS :=  -cxxlib                                                                                                                                                                                                          
CXX_LINKER := FORTRAN                                                                                                                                                                                                            
FC_AUTO_R8 := -r8                                                                                                                                                                                                                
FFLAGS :=  -qno-opt-dynamic-align  -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model source -O2 -debug minimal -qopt-report -xCORE_AVX2 -no-fma                                                 
FFLAGS_NOOPT := -O0                                                                                                                                                                                                              
FIXEDFLAGS := -fixed                                                                                                                                                                                                             
FREEFLAGS := -free                                                                                                                                                                                                               
HAS_F2008_CONTIGUOUS := TRUE                                                                                                                                                                                                     
MACRO_FILE :=                                                                                                                                                                                                                    
MPICC := mpicc                                                                                                                                                                                                                   
MPICXX := mpicxx                                                                                                                                                                                                                 
MPIFC := mpif90                                                                                                                                                                                                                  
NETCDF_PATH := /glade/u/apps/ch/opt/netcdf-mpi/4.8.0/mpt/2.22/intel/19.1.1/                                                                                                                                                      
PIO_FILESYSTEM_HINTS := gpfs                                                                                                                                                                                                     
PNETCDF_PATH := /glade/u/apps/ch/opt/pnetcdf/1.12.2/mpt/2.22/intel/19.1.1/                                                                                                                                                       
SCC := icc                                                                                                                                                                                                                       
SCXX := icpc                                                                                                                                                                                                                     
SFC := ifort                                                                                                                                                                                                                     
SLIBS := $(SLIBS)  -mkl=cluster                                                                                                                                                                                                  
SUPPORTS_CXX := TRUE                                                                                                                                                                                                             
ifeq "$(MODEL)" "mom"                                                                                                                                                                                                            
  CPPDEFS := $(CPPDEFS)  -DCESMCOUPLED -Duse_LARGEFILE -DFORTRANUNDERSCORE -DCPRINTEL                                                                                                                                            
                                                                                                                                                                                                                                 
endif                                                                                                                                                                                                                            
ifeq "$(MODEL)" "mom"                                                                                                                                                                                                            
  FFLAGS :=  $(FC_AUTO_R8)  -qno-opt-dynamic-align  -convert big_endian -assume byterecl -ftz -traceback -assume realloc_lhs -fp-model source -O2 -debug minimal -qopt-report -xCORE_AVX2 -no-fma                                
                                                                                                                                                                                                                                 
endif                                                                                                                                                                                                                            
ifeq "$(MODEL)" "gptl"                                                                                                                                                                                                           
  CPPDEFS := $(CPPDEFS)  -DCESMCOUPLED -DFORTRANUNDERSCORE -DCPRINTEL -DHAVE_NANOTIME -DBIT64 -DHAVE_VPRINTF -DHAVE_BACKTRACE -DHAVE_SLASHPROC -DHAVE_COMM_F2C -DHAVE_TIMES -DHAVE_GETTIMEOFDAY                                  
                                                                                                                                                                                                                                 
endif                                                                                                                                                                                                                            
                                                                                                                                                                                                                                 
```

jgfouca (approved):

PR #4104, opened 2021-10-11
Remove MODEL as a setting for the build system

jgfouca:
Use COMP_NAME instead.

If CESM has any build-scripts that are maintained outside of CIME, they may need to be updated. E3SM still had some that were using both MODEL= and COMP_NAME= and even a couple that were only using MODEL=.

This should hopefully remove some ambiguity from the system as it was never clear whether MODEL was supposed to be the component name or component class. COMP_NAME and COMP_CLASS make it very clear.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/Tools/Makefile
+++ b/scripts/Tools/Makefile
@@ -4,6 +4,17 @@
 #
 #===============================================================================
 
+ifdef MODEL
+  ifndef COMP_NAME
+    $(warning "Variable MODEL is deprecated, please use COMP_NAME instead")
+    COMP_NAME:=MODEL
+  else
+    ifneq ($(MODEL), $(COMP_NAME))
+      $(error "MODEL is inconsistent with COMP_NAME")
+    endif
+  endif
+endif
+
 # Set up special characters
 null  :=
 comma := ,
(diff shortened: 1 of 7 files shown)


jedwards4b:
MODEL in this context is a legacy variable that was in the Makefile pre cime.

jedwards4b (approved):

rljacob:
Right I was thinking of CIME_MODEL.  Nevermind.

jedwards4b:
@jgfouca if we add the following to the top of the Makefile I believe that this change will be backward compatible:
 ifdef MODEL
   ifndef COMP_NAME
     $(warning Variable MODEL is deprecated, please use COMP_NAME instead)
     COMP_NAME=MODEL
   endif
 endif

billsacks:
@jedwards4b @jgfouca thanks for thinking about backwards compatibility and consistency checking. If that can be done, that seems to be the best solution to avoid requiring everyone to update at once. If for some reason that doesn't work, I'm okay with this being merged given that it could be a pain for E3SM for it not to be.

jedwards4b:
@jgfouca looks good, thanks.

PR #4106, opened 2021-10-13
port to nersc perlmutter and CPE on cheyenne

jedwards4b:
Port to the NERSC system perlmutter  - this port is still a work in progress.  python socket.getfqdn() was not returning the fully quailified name and so I have added an alternate for identifying the machine.   <NODENAME_REGEX>$ENV{NERSC_HOST}:perlmutter</NODENAME_REGEX>  Where the part before the : is the environment variable you would look for and after the : is the expected match.

Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes 

User interface changes?: 

Update gh-pages html (Y/N)?:


jedwards4b:
@fischer-ncar NERSC_HOST env variable is cori and I wanted that to match the machine we called cori-haswell, the easiest way to do that was to rename cori-haswell to cori.   I realize that I don't need to make any cori change and can revert that.

jgfouca (approved):

PR #4110, opened 2021-10-20
add IRT test to exceptions for generate history

### LINKED ISSUE #3978: IRT test in cime_developer test suite fails in generate phase
I just ran the cime_developer test suite with the --generate flag. This produced:

```
    FAIL IRT_N2_Ln9.f19_g16_rx1.A.cheyenne_intel GENERATE exception
```

The problem seems to be that there are no history files in the run directory that can be used for generation. It looks like the test *does* produce a history file (IRT_N2_Ln9.f19_g16_rx1.A.cheyenne_intel.G.20210521_174039_x66o19.cpl.hi.0001-01-01-16200.nc in my case) but it is sent to the archive directory. I thought there is general logic to copy the last set of history files back into the run directory, but maybe it doesn't apply to coupler history files?

jedwards4b:
ADD IRT Test to list of exceptions for history file generate.

Test suite: scripts_regression_tests.py, IRT_C2_Ln9.f19_g16_rx1.A.cheyenne_intel.G
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #3978 

User interface changes?: 

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/SystemTests/irt.py
+++ b/scripts/lib/CIME/SystemTests/irt.py
@@ -7,6 +7,7 @@ This test the model's restart capability as well as the short term archiver's in
 (3) Recover first interim restart to the case2 run directory
 (4) Start case2 from restart and run to the end of case1
 (5) compare results.
+(6) this test does not save or compare history files in baselines.
 
 """
 
--- a/scripts/lib/CIME/hist_utils.py
+++ b/scripts/lib/CIME/hist_utils.py
@@ -494,7 +494,7 @@ def _generate_baseline_impl(case, baseline_dir=None, allow_baseline_overwrite=Fa
     testname = case.get_value("TESTCASE")
     testopts = parse_test_name(case.get_value("CASEBASEID"))[1]
     testopts = [] if testopts is None else testopts
-    expect(num_gen > 0 or (testname in ["PFS", "TSC"] or "B" in testopts),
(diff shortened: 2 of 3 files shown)


billsacks (changes requested):
Interesting fix. Two questions:

(1) Do you see this as a short-term workaround or a long-term solution? I am personally okay with either, but:
- If you see this as a short-term workaround, then we should keep the issue open for a more robust fix later.
- If you see this as a long-term solution, then we should clearly document that this test doesn't generate baselines to make it less likely that people will get tripped up by this difference in behavior. I'd suggest adding a comment both here https://github.com/ESMCI/cime/blob/adc6adc3f8fed9d6f8d3dfa1d6b4d93d081d0050/config/config_tests.xml#L294-L305 and here https://github.com/ESMCI/cime/blob/adc6adc3f8fed9d6f8d3dfa1d6b4d93d081d0050/scripts/lib/CIME/SystemTests/irt.py#L2-L9

(2) Have you tested whether the compare phase will also fail similarly if you run a test with `--compare` against previous baselines? I'm guessing it will and that you need to also add this test to https://github.com/ESMCI/cime/blob/adc6adc3f8fed9d6f8d3dfa1d6b4d93d081d0050/scripts/lib/CIME/hist_utils.py#L277 . To avoid duplication, it might be best to have a variable like `_TESTS_WITHOUT_BASELINES = ["PFS", "TSC", "IRT"]` at the top of this file and replace these two uses with a reference to that variable.

PR #4117, opened 2021-10-28
add python 3.7.0 module load for izumi (CESM only)

### LINKED ISSUE #4116: Add python module load command for izumi
To ensure CAM runs with python 3.7.0, add a module load command to the izumi section in config_machines.xml.

>  < command name="load" >lang/python/3.7.0</command>

peverwhee:
Add module load for python 3.7 on izumi for CESM in config_machines.xml. 
Without this update, the python version is set in cime regardless of 
whether a user does a module load before building. This addition ensures 
that CAM specifically is being built & run using python 3.7.

Test suite: scripts_regression_tests.py, clean builds of CAM, CESM
Test baseline: 
Test namelist changes: 
Test status: bit for bit

Fixes #4116 

User interface changes?: N

### DIFF
--- a/config/cesm/machines/config_machines.xml
+++ b/config/cesm/machines/config_machines.xml
@@ -2064,6 +2064,7 @@ This allows using a different mpirun command to launch unit tests
       <cmd_path lang="sh">module</cmd_path>
       <modules>
         <command name="purge"></command>
+        <command name="load">lang/python/3.7.0</command>
       </modules>
       <modules compiler="gnu">
         <command name="load">compiler/gnu/9.3.0</command>


gold2718:
> But you also need to update the externals for the new broken up cime
directory structure.

Thanks!

billsacks:
@peverwhee @gold2718 can you summarize where this stands? Do you still want this change or did you decide that this is unnecessary for now? I'm still okay with it if it's needed, and my impression is that @jedwards4b and @fischer-ncar are, too, though my feeling is that if it isn't actually needed, it seems best to skip it for now.

gold2718:
> can you summarize where this stands?

We would like to proceed for the reasons I recently added to #4116 because that is (what we believe to be) code improvement we cannot make in 3.6. An important consideration here is whether it is okay to require python 3.7 in CIME build scripts.

gold2718:
@billsacks, Thanks!

PR #4130, opened 2021-11-24
Support importing buildnml from cime_config/{compname}_cime_py

### LINKED ISSUE #4076: Naming conventions for files in cime_config/
## Background

In recent CSEG meetings, we talked about file naming conventions to maintain consistency among all the components -- namely, should python files have the `.py` extension or not. The prevailing view was that python modules that are meant to be imported should have the extension, scripts that should be run from the command line should not have it, and files that currently do double duty via `if __name__ == "__main__"` blocks should be split into two files (a `.py` module and a extension-free wrapper).

One big hitch to this plan is how CIME treats `cime_config/buildnml`. Currently there is logic that tries to import specific functions from `buildnml`, and falls back to executing the script via `subprocess`. @billsacks and I chatted about how to modify CIME to follow these conventions, and came up with some interesting ideas we thought were worth discussing.

## Potential paths forward

Assuming other groups agree that this would be a nice convention to follow, there a few possible ways to proceed.

1. The simplest fix would be to update `run_sub_or_cmd()` to try to import `f"{filename}.py"` before falling back on the current behavior of trying to import `filename` and running `filename` if the import fails. Perhaps in a later version this could collapse to "try to import `filename.py` or run `filename`", but leaving the middle check would maintain backwards compatibility for the current version.
2. Another possibility would be to allow components to make `cime_config` a package (i.e. include `__init__.py`). Then we could try to import `cime_config`, and fall back to the current behavior in CIME6 (again, perhaps cutting out the `import filename` option in future versions). This could potentially clean up some python code in components -- one example is that POP has a python layer that sits between `buildnml`; if these functions are included in the `cime_config` package it is easier to import them into `buildnml`.
3. A third possibility would be to enforce a rule that each component has `cime_config/${COMPNAME}/__init__.py`; this would let us more cleanly separate the python code in `cime_config` from the XML and the test suite files. It might also let us clean up `import` statements since there is no longer a namespace conflict every time you import a new `buildnml` (or `cime_config` in the above option)

## Questions

1. Does it make sense to make `buildnml` a wrapper script that imports `buildnml.py`?
2. If so, are any of the above paths the right answer for how CIME should make use of this renamed module?

In case it changes how you answer the above questions, I'll note that I am happy to take lead on implementing whatever solution results from this issue ticket.

mnlevy1981:
Proof of concept commit: see https://github.com/ESCOMP/POP2-CESM/pull/62 for companion component PR.

Basically, we want CIME to import `buildnml()` for each component from
 a package rather than using the logic in `run_sub_or_cmd()`. For the
 first commit, I just modified `preview_namelists.py` directly, but I think
 these changes probably belong in `run_sub_or_cmd()` itself?

Test suite: no testing yet, still working out the details of the change
Test baseline: N/A
Test namelist changes: N/A
Test status: N/A

Fixes #4076

User interface changes?: N

Update gh-pages html (Y/N)?: N (I don't think so, at least)

### DIFF
--- a/scripts/lib/CIME/case/preview_namelists.py
+++ b/scripts/lib/CIME/case/preview_namelists.py
@@ -6,6 +6,7 @@ create_dirs and create_namelists are members of Class case from file case.py
 from CIME.XML.standard_module_setup import *
 from CIME.utils import run_sub_or_cmd, safe_copy
 import time, glob
+import importlib
 logger = logging.getLogger(__name__)
 
 def create_dirs(self):
@@ -65,6 +66,7 @@ def create_namelists(self, component=None):
     # Create namelists - must have cpl last in the list below
     # Note - cpl must be last in the loop below so that in generating its namelist,
     # it can use xml vars potentially set by other component's buildnml scripts
+    comp_cime_py = dict()
     models = self.get_values("COMP_CLASSES")
     models += [models.pop(0)]
     for model in models:
@@ -84,10 +86,22 @@ def create_namelists(self, component=None):
             else:
                 # otherwise look in the component config_dir
                 cmd = os.path.join(config_dir, "buildnml")
-            expect(os.path.isfile(cmd), "Could not find buildnml file for component {}".format(compname))
-            logger.info("Create namelist for component {}".format(compname))
-            run_sub_or_cmd(cmd, (caseroot), "buildnml",
-                           (self, caseroot, compname), case=self)
+                # For now, only try to import if not using SourceMods
+                try:
+                    # can we import buildnml?
+                    sys.path.append(config_dir)
+                    comp_cime_py[compname] = importlib.import_module(f"{compname}_cime_py")
+                except:
+                    pass
+            # If successfully imported {compname}_cime_py, call buildnml(),
+            # otherwise just run buildnml via run_sub_or_cmd()
+            if compname in comp_cime_py:
+                comp_cime_py[compname].buildnml(self, caseroot, compname)
+            else:
+                expect(os.path.isfile(cmd), "Could not find buildnml file for component {}".format(compname))
+                logger.info("Create namelist for component {}".format(compname))
+                run_sub_or_cmd(cmd, (caseroot), "buildnml",
+                               (self, caseroot, compname), case=self)
 
         logger.debug("Finished creating component namelists, component {} models = {}".format(component, models))
 


mnlevy1981:
@billsacks and I talked yesterday, and ended up with some pretty nice changes. `preview_namelists` calls `import_and_run_sub_or_cmd()` instead of `run_sub_or_cmd()`. This new subroutine tries to import `{compname}_cime_py` as a module and run `subname` from that module -- if that fails, it falls back to just calling `run_sub_or_cmd()`.

One big remaining concern that needs to be addressed before taking this PR out of draft status is how to deal with users who want to put `buildnml` (now `buildnml.py`) in source mods. As the code currently stands, even if a user puts a file named `buildnml` in source mods, the subroutine will give preference to code in `cime_config/{compname}_cime_py` and only look at what is in source mods if that import or command fails. Two possible solutions:

1. Use an optional argument to `import_and_run_sub_or_cmd()` and move the check for source mods into that routine
2. have `preview_namelists` check for source mods and call `run_sub_or_cmd()` directly (without the `import_and_run_sub_or_cmd()`) if a script is found there

But I think supporting code in source mods may make it very difficult to refactor python code in `{compname}_cime_py`, since we'd have to somehow support letting users change one or more of those files in source mods and then somehow import the correct version of everything.

billsacks:
Thanks for your work on this @mnlevy1981 ! I agree that it's going to be awkward to support testmods with this, but if we want to keep this backward compatible, I would vote for option (2) in your last comment.

billsacks:
Maybe we should also issue a warning that this use of SourceMods for buildnml is deprecated if we detect that someone is using it. One idea (maybe overkill?) is that we could open an issue here in cime to remove support for this deprecated feature, then in the warning invite people to make a comment in the issue if this feature is critical for their workflow. If we haven't heard from many people in some amount of time, then we could remove that feature?

jedwards4b:
I agree that we should not support putting cime python code into SourceMods.

mnlevy1981:
> I agree that we should not support putting cime python code into SourceMods.

I'm okay with this rule of thumb, but should it apply to `cime_config` code in each component as well? I.e. is `${POP}/cime_config/buildnml` cime python code or POP code?

jedwards4b:
I would call it cime code.   Are there frequent cases where we make this substitution?  If the user needs to make a change in buildnml then it seems to me that they should create a branch of the component and make the change there.

mnlevy1981:
I've removed support for putting `buildnml` in Source Mods, so this PR is ready for review. It does not require any component changes - `cime_config/buildnml` will still run (or be imported) if `cime_config/{compname}_cime_py` does not exist or `{compname}_cime_py.buildnml()` is not available. Once this is available in CESM, I can update POP to use this new setup.

One question: should the component put any other modules in this directory? I think we've discussed how `buildlib` is not written in a way to easily refactor in this way, but I wasn't sure if this new import scheme could be used for any other scripts.

billsacks (approved):
Thanks a lot for your work on this @mnlevy1981 ! This looks good to me.

Regarding your question:

> One question: should the component put any other modules in this directory? I think we've discussed how buildlib is not written in a way to easily refactor in this way, but I wasn't sure if this new import scheme could be used for any other scripts.

The one other thing I can think of is `cime_config/SystemTests` (used by ctsm and other components: https://github.com/ESCOMP/CTSM/tree/master/cime_config/SystemTests). I'm fine with that being added later, though.

I feel like the next step here will be figuring out a robust & consistent pattern for allowing components to also include some python code from a different location. For ctsm, for example, I think we'll eventually want buildnml (and the SystemTests) to be able to access python code in the main ctsm python module (https://github.com/ESCOMP/CTSM/tree/master/python/ctsm ). Maybe this should be handled in an "official" way via config_files (adding a specification for an optional directory to add to sys.path for this component), or maybe we should just do this in a more ad-hoc way (e.g., having `buildnml.py` adjust the sys.path itself). (I prefer the more official way of using config_files, because I think it will be less error-prone long-term.) But that can also be added later.

jedwards4b (approved):

PR #4134, opened 2021-12-04
Fix probing machine

### LINKED ISSUE #4133: PR #4106 causes failures in auto-detecting some machine names
The changes in #4106 cause failures in auto-detecting some machine names, resulting in the following (at least for CESM):

```
ERROR: No machine $ENV{NERSC_HOST}:perlmutter found
```

The problem arises when the first call to `_probe_machine_name_one_guess` from `probe_machine_name` fails (because `socket.getfqdn()` doesn't provide the match, but the later call to `socket.gethostname()` will provide the match). The specific issue is that the `machine` variable was used for multiple purposes in the changes to `machines.py` in #4106, so when `_probe_machine_name_one_guess` was supposed to return `None`, it was instead returning the text value of `NODENAME_REGEX` from the last machine with an `$ENV{XXX}:YYY` pattern.

I have a fix on a branch and will open a PR shortly.

billsacks:
See #4133 for details of the issue.

Test suite: scripts_regression_tests - only A_RunUnitTests and B_CheckCode
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes #4133 

User interface changes?: N

### DIFF
--- a/scripts/lib/CIME/XML/machines.py
+++ b/scripts/lib/CIME/XML/machines.py
@@ -175,13 +175,13 @@ class Machines(GenericXML):
                 logger.debug("machine regex string is " + regex_str)
                 # an environment variable can be used
                 if regex_str.startswith("$ENV"):
-                    machine = self.get_resolved_value(regex_str, allow_unresolved_envvars=True)
-                    if not machine.startswith("$ENV"):
+                    machine_value = self.get_resolved_value(regex_str, allow_unresolved_envvars=True)
+                    if not machine_value.startswith("$ENV"):
                         try:
-                            match,machine = machine.split(":")
+                            match, this_machine = machine_value.split(":")
                         except ValueError:
                             expect(False,"Bad formation of NODENAME_REGEX.  Expected envvar:value, found {}".format(regex_str))
-                        if match == machine:
+                        if match == this_machine:
                             machine = machtocheck
                             break
                 else:    


scripts/lib/CIME/XML/machines.py
```diff
@@ -175,13 +175,13 @@ def _probe_machine_name_one_guess(self, nametomatch):
                 logger.debug("machine regex string is " + regex_str)
                 # an environment variable can be used
                 if regex_str.startswith("$ENV"):
-                    machine = self.get_resolved_value(regex_str, allow_unresolved_envvars=True)
-                    if not machine.startswith("$ENV"):
+                    machine_value = self.get_resolved_value(regex_str, allow_unresolved_envvars=True)
```
billsacks, line 178: This is the critical change

scripts/lib/CIME/XML/machines.py
```diff
@@ -175,13 +175,13 @@ def _probe_machine_name_one_guess(self, nametomatch):
...
+                    if not machine_value.startswith("$ENV"):
                         try:
-                            match,machine = machine.split(":")
+                            match, this_machine = machine_value.split(":")
```
billsacks, line 181: We don't absolutely need a new variable here (`this_machine`): we could reuse `machine_value` like the old code reused `machine`. However, I have introduced a new variable for clarity, because I find it confusing when code uses the same variable for multiple purposes.

jedwards4b (approved):

PR #4141, opened 2021-12-16
Adds enforcing code formatting to pre-commit

jasonb5:
The goal of this PR is to enforce a coding style using pre-commit
plugins (check-xml, end-of-file-fixed, trailing-whitespace and black).
Fixing linting/formatting issues is now a single command 
`pre-commit run -a`. This auto formatting is performed by Black
which enforces a strict subset of PEP8. Pre-commit runs all of
these plugins for every PR. Pylint still runs as a redundancy. By
enforcing a uniformed coding style PR's will now be style agnostic 
and less development time used on formatting code.

### Running pre-commit
```
pip install pre-commit
pre-commit run -a
```
### References
- https://black.readthedocs.io/en/stable/index.html
- https://black.readthedocs.io/en/stable/the_black_code_style/current_style.html

Test suite: n/a
Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes n/a
User interface changes?: n/a


ekluzek:
I have become a fan of using black. Is automatically reformatting with black on commit something that's now approved by the wider cime community?  Or is this PR being done as a way to introduce the idea to the community?

I'm also wondering about your views on doing this as a pre-commit hook rather than a check on push? I know there are pros and cons to each. Partially just curious as we are looking how to bring in using black for CTSM, and I'm wondering about the best ways to bring it into the process.

jedwards4b (approved):
My easier to view if you exclude whitespace changes.

billsacks:
I haven't tried to review this, but have a few questions about how this will be integrated in the workflow moving forward.

First, though, in terms of reviewing this:

(1) Are there substantive changes that warrant review here alongside the essentially non-reviewable formatting changes? If so, would it be easy for you to point to specific files or commits that warrant review?

Now my questions about workflow:

(2) Can you give more description of how you envision the running of black and the other formatters fitting into the workflow? The fact that you are using a tool called "pre-commit" made me initially think that you are suggesting that people should set up pre-commit hooks, but from your description that doesn't seem to necessarily be the case. So are you instead imagining that people will run "pre-commit" manually? I think in my glance through the diffs I also saw that you are integrating it into the GitHub actions. If so, will that actually commit the changes back to a PR, or does the GitHub action just run black in "check" mode?

I'm not sure what the best workflow is here, but there are a few things I would like to avoid:
- (A) The top thing I'd like to avoid is having someone run black in the midst of a PR review. The specific situation I want to avoid is: I review commit abc123 which did not have black run on it, then my re-review (after the author has addressed my comments) is on commit def456 which did have black run on it. The problem in this case is that I want to review the changes made between abc123 and def456, but diffing those two points will give me a mix of real changes and changes created by black (the latter fixing formatting issues with abc123). I can see two solutions to this:
  - (i) (Ideal) A reviewer only reviews code that has had black run on it. It's okay if you make intermediate commits that have not been run through black (although maybe the absolute best thing would be to have all commits be cleaned by black), but a reviewer will only look at the code after it has been run through black. In this case, if I'm thinking about it right, then the review of diffs between any two points won't be contaminated by extra black-induced formatting cleanup.
  - (ii) If (i) doesn't happen for some reason, and a reviewer reviews a point that is not clean according to black, then either: (a) black should be run immediately after the review (essentially reverting to option (i)) or, (b) if that doesn't happen, and other commits have been made in the meantime, then the author should wait to run black at all until later, to make the reviewer's job easier.
- (B) I also foresee some additional issues (whose magnitude is hard for me to judge up-front) if reformatting commits are frequently separated from "real" commits: I can at least see issues for git blame becoming more difficult to use in this situation. So at least for myself, I'd like to try to adopt a workflow where black is applied either by my editor (on save, for example) or in a pre-commit hook, ensuring that all of my git commits are black-clean.

(3) Since the giant reformatting commit(s) in this PR will interfere with git blame, what would you think about adding a file like .git-blame-ignore-revs in the root of the ctsm repository listing the commit(s) that ran reformatting operations? See https://akrabat.com/ignoring-revisions-with-git-blame/ for details.

(4) Once this PR is merged, what would you recommend for how the updated, reformatted master is merged with existing branches that predated the running of black? I know one option would be for all of those branches to run black themselves before merging with master, but I'm a little worried about how multiple giant-reformatting commits might interfere with git blame, so my hope is that we can avoid that if possible – i.e., having there only be a single giant-reformatting commit anywhere. But if it's necessary for all of these branches to run black themselves, then that could argue more for using the mechanism in (3) so we can maintain a better-working git blame (which I often find helpful to understand the history of some code).

(5) Finally, I just wanted to point out that this is a bit of a mis-statement:

> Pylint still runs as a redundancy.

My understanding is that pylint does a lot of checks beyond what black handles. We could / should remove the style checking of pylint after adopting black, but I think we still want pylint for its ability to catch programming errors (or else adopt a different tool that serves that purpose).

jasonb5:
@ekluzek 
This PR was to introduce the idea, there has been some talk about using black. Right now
this just checks on push, but it's still up for discussion as we could force a pre-commit hook or
utilize the free pre-commit.ci service to run and commit changes.

@billsacks
1. These are the only files that warrant a review, everything else is just formatting.

   - .github/workflows/srt.yml
   - .pre-commit-config.yaml
   - CONTRIBUTING.md

2. There was never an intention of making developers use pre-commit hooks. The vision was that a developer could choose how they want to run pre-commit (black and other plugins) but that they would have to do it before their PR is merged. The "enforcement" portion is GitHub action running the pre-commit tool which would cause a merge to block if issues arise.
Maybe I'm missing some detail here but I'm not sure I see any new issues using black with respect to git blame. We already
deal with PRs containing a mix of "real" changes and syntactic changes. I think the only way to mitigate the effects on git blame
would be to ensure syntactic changes are included in commits containing "real" changes.

3. I don't see any issue adding .git-blame-ignore-revs for the big commit.

4. The existing branches would need to merge master then run black. The PR merge into master would only result 
in the relevant changes from black on that branch which is what all future PRs will look like when they contribute these type of changes.

5. Yes you're correct this is a miss-statement, I was referring to the stylistic portion of pylint. Pylint is still needed
to detect other possible errors.

CONTRIBUTING.md
```diff
@@ -67,12 +67,22 @@ to tell CIME about your development machine. See the [CIME users guide](https://
...
+
+    pytest scripts/lib/CIME/tests
 
 Make your change. Add tests for your change. Make the tests pass to the same level as before your changes.
 
-    scripts_regression_tests.py
+    cd scripts/lib/CIME/tests
+    python scripts_regression_tests.py
+
+Run [pre-commit](https://pre-commit.com/#usage) before commit changes and submitting a PR.
```
billsacks, line 82:
```suggestion
Run [pre-commit](https://pre-commit.com/#usage) before committing changes and submitting a PR.
```

CONTRIBUTING.md
```diff
@@ -90,8 +100,9 @@ Some things that will increase the chance that your pull request is accepted:
 * Follow [PEP8][pep8] for style. (The `flake8` utility can help with this.)
 * Write a [good commit message][commit].
 
-Pull requests will automatically have tests run by Travis. This includes
-running both the unit tests as well as the `flake8` code linter.
+Pull requests will automatically have tests run by Github Action. This
```
billsacks, line 103:
```suggestion
Pull requests will automatically have tests run by a Github Action. This
```

CONTRIBUTING.md
```diff
@@ -90,8 +100,9 @@ Some things that will increase the chance that your pull request is accepted:
...
+includes running both the unit tests as well as `pre-commit`, which checks
+liniting.
```
billsacks, line 105:
```suggestion
linting.
```

billsacks (changes requested):
I made a few inline suggestions just fixing some very small typos.

Thanks a lot for the extra description in response to my last comment – it's helpful to have this extra information.

I realized that I made a copy & paste error in point (3): I initially asked about adding .git-blame-ignore-revs in the root of the ctsm repository, but I really meant the cime repository – and so was wondering what you would think about including that in this PR.

Regarding git blame: Maybe this is not a huge deal (after the initial big reformat), but what I'm thinking about is that you will often need to go back multiple commits to see the commit that actually gives a useful log message description of what changed, because the most recent commit is likely to be just a formatting change, given how many formatting changes black seems to make. This is why I see advantages to ensuring that every commit is black-clean (via editor integration or a pre-commit hook) rather than doing a separate reformatting commit. Again, maybe not a big deal, but just something we may want to consider in terms of recommended workflows.

Finally, I wanted to more succinctly state my view related to your statement, "they would have to [run pre-commit] before their PR is merged": I don't feel strongly for or against black, but I do feel that *if* we adopt black, then we should have some pretty firm guidelines to avoid making reviewers' jobs more difficult. In some initial reviews in CTSM when people were first playing around with black, a re-review was challenging because it included a mix of real changes plus black-ification of some things that hadn't been black-ified for the initial review, and it was hard to separate the two in my re-review. To avoid that happening again, **I would be tempted to refuse to review (or re-review) a PR unless it was passing black's checks (this is in line with the guidance you added to the CONTRIBUTING file – thanks).**

jgfouca (approved):
Nice!

CONTRIBUTING.md
jasonb5: Fixed.

CONTRIBUTING.md
jasonb5: Fixed.

CONTRIBUTING.md
jasonb5: Fixed.

billsacks (approved):

jasonb5:
@billsacks 

I don't see any issue adding .git-blame-ignore-revs, this should take care of the bulk changes. 

I agree the recommended workflow should be running black with each commit. 

My goal is to keep the review process simple, and would recommend that reviewers wait until pre-commit and all tests have passed.

jasonb5:
Here's some notes about `.git-blame-ignore-revs`.
- Two ways to utilize with CLI (https://black.readthedocs.io/en/stable/guides/introducing_black_to_your_project.html#avoiding-ruining-git-blame)
  - Add `--ignore-revs-file .git-blame-ignore-revs` to `git blame`
  - Set in git config `git config blame.ignoreRevsFile .git-blame-ignore-revs`
- Github blame view doesn't support `.git-blame-ignore-revs` but it's in the works and high priority (https://github.com/github/feedback/discussions/5033)

jgfouca:
@jasonb5 , if this is ready, please merge it. Looks like everyone approves and checks have passed.

billsacks:
@jasonb5 thanks for adding the `.git-blame-ignore-revs` file and for your comments on it!

PR #4150, opened 2022-01-10
handle ESMF_Profile format change (backward compatible)

jedwards4b:
Format of the ESMF_Profile.summary file which is used to get cesm timing change in ESMF8_3_0_beta_snapshot_04
this handles the new format as well as the old format.

Test suite: by hand
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #

User interface changes?:

Update gh-pages html (Y/N)?:


mvertens (approved):
Could you please add some comments regarding what ESMF version these changes are compatible with? Also - are these changes backwards compatible with older ESMF versions? That would be good to comment on as well.

PR #4154, opened 2022-01-11
-o should remove not just overwrite baseline contents

### LINKED ISSUE #4152: Overwriting baselines should not leave stale files
The directory should probably be blown away and re-created to avoid leaving stale files that can cause confusion. See:
https://github.com/E3SM-Project/E3SM/issues/3374

jedwards4b:
Remove rather than overwrite baseline contents with the create_test -o flag.

Test suite: hand testing
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #4152

User interface changes?:

Update gh-pages html (Y/N)?:


jgfouca (approved):

jgfouca:
Thanks, @jedwards4b !

PR #4156, opened 2022-01-12
This fixes a number of fails in scripts_regression for e3sm on mappy

jgfouca:
It looks like these problems were missed because we aren't running in e3sm mode in GitHub actions.

Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/tests/test_sys_test_scheduler.py
+++ b/scripts/lib/CIME/tests/test_sys_test_scheduler.py
@@ -14,9 +14,11 @@ from CIME.tests import base
 
 
 class TestTestScheduler(base.BaseTestCase):
-    @mock.patch.dict(os.environ, {"CIME_MODEL": "cesm"})
     @mock.patch("time.strftime", return_value="00:00:00")
     def test_chksum(self, strftime):  # pylint: disable=unused-argument
+        if utils.get_model() != "cesm":
+            self.skipTest("Skipping chksum test. Depends on CESM settings")
+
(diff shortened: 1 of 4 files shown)


scripts/lib/CIME/tests/test_sys_test_scheduler.py
```diff
@@ -14,9 +14,11 @@
 
 
 class TestTestScheduler(base.BaseTestCase):
-    @mock.patch.dict(os.environ, {"CIME_MODEL": "cesm"})
     @mock.patch("time.strftime", return_value="00:00:00")
     def test_chksum(self, strftime):  # pylint: disable=unused-argument
+        if utils.get_model() != "cesm":
```
jgfouca, line 19: I think the intent of this test was to use mock to have it work everywhere, but that doesn't work because:
```
CIME.utils.CIMEError: ERROR: Makes no sense to have empty read-only file: /home/e3sm-jenkins/jenkins-ws/workspace/mappy_cime/E3SM/ccs_config/machines/config_machines.xml
```
The E3SM config files don't live under ccs_config.

jasonb5: Yea that was the intent, this should be fine now.

jasonb5:
LGTM

jasonb5 (approved):

PR #4161, opened 2022-01-18
Remove the remaining users of config_compilers.xml

### LINKED ISSUE #4136: PFUNIT-based unit tests should use new CMake Macros files, not old config_compilers.xml
In updating my Mac's CIME port to the new CMake Macro-based system, I noticed that `scripts/fortran_unit_testing/run_tests.py` still needs things defined in config_compilers.xml: defining PFUNIT_PATH (and maybe other things?) in the new macros format does not seem sufficient.

For example, if I remove the path to PFUNIT from the cheyenne section in CESM's config_compilers.xml, I get

```
ERROR: PFUNIT_PATH not found for this machine and compiler, with MPILIB=mpi-serial and compile_threaded=FALSE.
You must specify PFUNIT_PATH in config_compilers.xml, with attributes MPILIB and compile_threaded.
```

(I tried this out of the latest CTSM master, with its externals, running `../cime/scripts/fortran_unit_testing/run_tests.py --build-dir unit_tests.temp` from CTSM's `src` directory.)

Since I think we want to eventually remove the need for config_compilers.xml (I assume?), I'm thinking this run_tests.py script should be updated to use the new macros-based method.

By the way, this exercise made me explore the new CMake Macro-based system in more detail. I feel like this is a definite improvement over the previous config_compilers.xml. Thank you @jgfouca and @jedwards4b for all of your work on this!

jgfouca:
The Compilers XML object should not even be constructed unless
the env var enabling the deprecated macro system is on.

Test baseline:
Test namelist changes:
Test status: bfb

Fixes #4136 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/fortran_unit_testing/run_tests.py
+++ b/scripts/fortran_unit_testing/run_tests.py
@@ -9,12 +9,14 @@ sys.path.append(os.path.join(_CIMEROOT, "scripts", "fortran_unit_testing", "pyth
 
 from standard_script_setup import *
 from CIME.BuildTools.configure import configure, FakeCase
-from CIME.utils import run_cmd_no_fail, stringify_bool, expect, get_src_root
+from CIME.utils import run_cmd_no_fail, stringify_bool, expect, get_src_root, safe_copy
 from CIME.XML.machines import Machines
 from CIME.XML.compilers import Compilers
 from CIME.XML.env_mach_specific import EnvMachSpecific
+from CIME.build import get_makefile_vars
 from xml_test_list import TestSuiteSpec, suites_from_xml
-import socket
+import socket, shutil
+from pathlib import Path
 
 # =================================================
 # Standard library modules.
@@ -287,31 +289,34 @@ def make_stage(name, output, make_j, clean=False, verbose=True):
     run_cmd_no_fail(" ".join(make_command), combine_output=True)
 
 
-def find_pfunit(compilerobj, mpilib, use_openmp):
+def find_pfunit(caseroot, cmake_args):
     """Find the pfunit installation we'll be using, and print its path
 
     Aborts if necessary information cannot be found.
 
     Args:
-    - compilerobj: Object of type Compilers
-    - mpilib: String giving the mpi library we're using
-    - use_openmp: Boolean
+    - case: A fake case
+    - caseroot: The dir with the macros
     """
-    attrs = {"MPILIB": mpilib, "compile_threaded": "TRUE" if use_openmp else "FALSE"}
+    new_cmake_dir = os.path.join(caseroot, "cmake_macros")
+    cmake_lists = os.path.join(new_cmake_dir, "CMakeLists.txt")
+    Path(os.path.join(caseroot, "cmaketmp")).mkdir(parents=False, exist_ok=True)
+    safe_copy(cmake_lists, "cmaketmp")
 
-    pfunit_path = compilerobj.get_optional_compiler_node(
-        "PFUNIT_PATH", attributes=attrs
-    )
-    expect(
-        pfunit_path is not None,
-        """PFUNIT_PATH not found for this machine and compiler, with MPILIB={} and compile_threaded={}.
-You must specify PFUNIT_PATH in config_compilers.xml, with attributes MPILIB and compile_threaded.""".format(
-            mpilib, attrs["compile_threaded"]
-        ),
-    )
-    logger.info("Using PFUNIT_PATH: {}".format(compilerobj.text(pfunit_path)))
-    return compilerobj.text(pfunit_path)
+    all_vars = get_makefile_vars(None, caseroot, cmake_args=cmake_args)
+    shutil.rmtree(os.path.join(caseroot, "cmaketmp"))
 
+    all_vars_list = all_vars.splitlines()
+    for all_var in all_vars_list:
+        if ":=" in all_var:
+            expect(all_var.count(":=") == 1, "Bad makefile line: {}".format(all_var))
+            varname, value = [item.strip() for item in all_var.split(":=")]
+            if varname == "PFUNIT_PATH":
+                logger.info("Using PFUNIT_PATH: {}".format(value))
+                return value
+
+    expect(False, "PFUNIT_PATH not found for this machine and compiler")
+    return None
 
 # =================================================
 # Iterate over input suite specs, building the tests.
@@ -402,10 +407,6 @@ def _main():
         compiler = machobj.get_default_compiler()
         logger.info("Compiler is {}".format(compiler))
 
-    compilerobj = Compilers(machobj, compiler=compiler, mpilib=mpilib)
-
-    pfunit_path = find_pfunit(compilerobj, mpilib=mpilib, use_openmp=use_openmp)
-
     debug = not build_optimized
     os_ = machobj.get_value("OS")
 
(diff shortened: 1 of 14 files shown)


jgfouca:
This is a bit of a work in progress. Without a proper case to work off of, interrogating CIME for build info is a bit awkward.

scripts/lib/CIME/BuildTools/configure.py
```diff
@@ -79,6 +79,31 @@ def configure(
                     new_cmake_macros_dir, os.path.join(output_dir, "cmake_macros")
                 )
 
+            if form == "Makefile":
+                # Use the cmake macros to generate the make macros
```
jgfouca, line 83: This feel tacky. Fortunately, it appears as though using `configure` to create Makefile macros is very uncommon. With this PR, the only remaining use is in `tools/mapping/gen_domain_files/test_gen_domain.sh`

scripts/lib/CIME/tests/utils.py
```diff
@@ -353,34 +353,6 @@ def assert_variable_matches(self, var_name, regex, env=None, var=None):
         self.parent.assertRegexpMatches(self.query_var(var_name, env, var), regex)
 
 
-def get_macros(macro_maker, build_xml, build_system):
```
jgfouca, line 356: This was a duplicate definition.

tools/cprnc/README
```diff
@@ -12,30 +12,19 @@ your case directory.
 Quick Start Guide:
 ------------------
 
-On cime supported systems you can generate a Macros file using the following
+On cime supported systems you can generate a cmake Macros file using the following
 (assuming you are running the command from the directory cime/tools/cprnc):
 
-CIMEROOT=../.. ../configure --macros-format=Makefile --mpilib=mpi-serial
```
jgfouca, line 18: We want users to use CMake. Unfortunately, there is still a Makefile build of cprnc in `tools/mapping/gen_domain_files/test_gen_domain.sh`, so I couldn't get rid of the Makefile.

scripts/lib/CIME/tests/test_sys_unittest.py
```diff
@@ -7,30 +7,32 @@
...
-            "PFUNIT_PATH", attributes=attrs
-        )
-        if pfunit_path is None:
-            return False
+            compiler = self.TEST_COMPILER
+
+        mach = self.MACHINE.get_machine_name()
+        cmake_macros_dir = Files().get_value("CMAKE_MACROS_DIR")
+
+        machine_macro = os.path.join(cmake_macros_dir, "{}_{}.cmake".format(compiler, mach))
```
jgfouca, line 29: Kinda hacky.

scripts/lib/CIME/tests/test_sys_cmake_macros.py
```diff
@@ -28,6 +29,9 @@ def xml_to_tester(self, xml_string):
     def setUp(self):
         super().setUp()
 
+        if "CIME_NO_CMAKE_MACRO" not in os.environ:
```
jgfouca, line 32: Don't test the deprecated system unless the user indicates they want the deprecated system in their env.

scripts/lib/CIME/tests/test_sys_full_system.py
```diff
@@ -12,8 +12,7 @@
 class TestFullSystem(base.BaseTestCase):
     def test_full_system(self):
         # Put this inside any test that's slow
-        if self.FAST_ONLY:
-            self.skipTest("Skipping slow test")
+        self.skipTest("Skipping slow test")
```
jgfouca, line 15: Need to revert this change.

scripts/lib/CIME/BuildTools/configure.py
jedwards4b: I don't see any reason that we cannot replace this (in  test_gen_domain.sh) with a cmake build - do you?  That would allow us to remove this ```if form```?

jedwards4b (approved):

scripts/lib/CIME/BuildTools/configure.py
jgfouca: @jedwards4b , it looks like the source code in `tools/mapping/gen_domain_files/src/*` requires a Makefile build system, so we'd need to change that in order to be able to remove Makefile mode from `configure`.

jgfouca:
As a side note, it looks like there was a bit of confusion in the code as to where local cmake macros should be located.

The test `CIME.tests.test_sys_create_newcase.TestCreateNewcase.test_ka_createnewcase_extra_machines_dir` was assuming they would live directly under extra_mach_dir, e.g. `$extra_mach_dir/*.cmake`. `case_setup` was looking for local macros in `$extra_mach_dir/cmake_macros/*.cmake`. This was an easy fix, but I'll note that `case_setup` also looks for local macros in `$HOME/.cime/*cmake`. So, we expect extra-mach-dir to provide cmake macros under the directory `cmake_macros`, but we do not expect this in `$HOME/.cime` cmake macros which are expected to live directly in the directory, so things are not consistent between these two approaches. Was this a deliberate choice or accident? @jedwards4b @billsacks

scripts/fortran_unit_testing/run_tests.py
```diff
@@ -425,19 +421,24 @@ def _main():
     )
     machspecific = EnvMachSpecific(build_dir, unit_testing=True)
 
-    fake_case = FakeCase(compiler, mpilib, debug, comp_interface)
+    fake_case = FakeCase(compiler, mpilib, debug, comp_interface, threading=use_openmp)
     machspecific.load_env(fake_case)
+    # JGF bug? original contents of cmake_args is lost
```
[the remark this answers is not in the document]
jedwards4b: In the previous version cmake_args was not an input argument so I don't think this condition was ever satisfied, but now that it is, this does appear to be a bug.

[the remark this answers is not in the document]
jgfouca: @jedwards4b , I pushed what I think is the appropriate fix for this.

billsacks:
> As a side note, it looks like there was a bit of confusion in the code as to where local cmake macros should be located.
> 
> The test `CIME.tests.test_sys_create_newcase.TestCreateNewcase.test_ka_createnewcase_extra_machines_dir` was assuming they would live directly under extra_mach_dir, e.g. `$extra_mach_dir/*.cmake`. `case_setup` was looking for local macros in `$extra_mach_dir/cmake_macros/*.cmake`. This was an easy fix, but I'll note that `case_setup` also looks for local macros in `$HOME/.cime/*.cmake`. So, we expect extra-mach-dir to provide cmake macros under the directory `cmake_macros`, but we do not expect this in `$HOME/.cime` cmake macros which are expected to live directly in the directory, so things are not consistent between these two approaches. Was this a deliberate choice or accident? @jedwards4b @billsacks 

I ran into this confusion, too. Once I figured out what was going on, I came to feel that the way it's done for the extra_machines_dir (having a cmake_macros subdirectory) makes sense in that it parallels what's done for the actual machines directory, though I didn't (and still don't) feel strongly one way or the other. (This led me to add clarification in the documentation in #4137 ). I can see arguments either way for `.cime`. I do see that it could be more clear if these two were consistent. In that case, I don't have much preference as to whether they are both done the way `.cime` is or the way `extra_machines_dir` is done, though if I had to choose, I'd vote for using a `cmake_macros` subdirectory in both cases.

jedwards4b:
I think that it makes most sense to have .cime/cmake_macros/*.cmake

jgfouca:
@jedwards4b , @billsacks , I agree that it makes sense to have `~/.cime/cmake_macros`. Unfortunately, this change would break backwards compatibility and I know some E3SM devs that are always using custom macros, so I am a bit hesitant to just make the change here. Perhaps I will take this on in a separate PR.

billsacks:
> I agree that it makes sense to have `~/.cime/cmake_macros`. Unfortunately, this change would break backwards compatibility and I know some E3SM devs that are always using custom macros, so I am a bit hesitant to just make the change here. Perhaps I will take this on in a separate PR.

That makes sense - thanks for all of this.

PR #4168, opened 2022-01-28
need to move version check out of utils.py

jedwards4b:
The check_minimum_python_version script needed to be moved from utils.py.  Otherwise you get an import error on utils.py before the version message is printed.  

Test suite: hand testing with python 2.7, 3.4, 3.6 and 3.9
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/Tools/standard_script_setup.py
+++ b/scripts/Tools/standard_script_setup.py
@@ -7,6 +7,29 @@ that every script should do.
 import sys, os
 import __main__ as main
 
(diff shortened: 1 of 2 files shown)


jgfouca (approved):

jasonb5 (approved):
LGTM

Issue #3222, opened 2019-08-31
Generalize system_tests_compare_two to system_tests_compare_n


billsacks:
@fischer-ncar has requested generalizing system_tests_compare_two to system_tests_compare_n (well actually he asked for just three, but if we're going to support three we might as well support an arbitrary n). In the past, I have also thought this might be valuable.

Off-hand, I think this could be done with a moderate amount of effort. I believe that we wouldn't have to change anything fundamental about the logic and flow of system_tests_compare_two, but would need to generalize it by:

- Adding an argument to the `__init__` method specifying how many cases to run for this test (defaults to 2, which would be the minimum allowed value)

- Where we have hard-coded variables referring to case1 or case2, we'd need to turn this into a list.

- In many places where we do something for case2, we'd need to have a loop over 1..(n-1), where n is the number of cases (here I'm using 0-based indexing, where case 0 is what used to be case 1)

- Specific tests would implement methods with numerals in their names rather than spelled-out numbers. For example, `_case_two_setup` would become `_case_setup_2`. In the loops over case, we'd use introspection (`getattr` or similar) to call the appropriate method based on a generated string with the current case number. (And for things where we currently have an empty implementation in the base class, like `_case_two_custom_prerun_action`, we could first check hasattr to see if the given method is implemented for this test.)

- For multi-submit tests with more than two phases: We could check the value of RESUBMIT compared with the number of phases expected, and determine the current phase based on that.

Overall, these changes wouldn't be trivial, but I think they'd be relatively straightforward. (Famous last words.) They can be tested with a combination of:

- Existing unit tests of system_tests_compare_two (which may need to be tweaked a bit)

- One or a few new unit tests that test the 3-case situation

- Some manual testing of 2-case and 3-case tests, including verifying that failures are still reported properly for the run phase as well as comparison between case 1 and any of the following cases

I believe we may be able to even do this in a backwards-compatible way, by maintaining a system_tests_compare_two that implements methods like `_case_one_setup` and `_case_two_setup` (and all other methods that can be implemented by individual tests) like this:

```python
def _case_setup_2(self):
    self._case_two_setup()
```

(though I haven't thought this through carefully). The point of this would be to at least temporarily support tests in various components. (If this backwards compatibility piece is trivial enough and seems like it won't be a maintenance burden, we could even support this long-term, in order to keep the interface somewhat simpler for the common two-case version. But I'm not sure if that's worth doing.)

PR #4169, opened 2022-02-01
Comparison test class with arbitrary number of instances

alperaltuntas:
This PR introduces a new test class, `SystemTestsCompareN`, which is
a generalization of already existing `SystemTestsCompareTwo` class.
The new test class supports arbitrary number of subsequent case
instances (N-1) as opposed to only one to be compared against the
base case.

We needed this new class to implement a novel MOM6 test within CESM.
The MOM6 test that initiated this PR is called "Dimensional Consistency
Test", or DIMCS. In summary the DIMCS test involves a base case that is
absent any modifications and 12 subsequent cases with modified 
dimensional scaling factors. Each subsequent case corresponds to one 
of 6 different scaling factors (T, L, H, Z, R, and Q) to be set to either `2**11`
or `2**-11`. Hence 12 subsequent cases to be compared against the base 
case. We expect all cases to be bitwise identical to the base case which 
is absent any dimensional scaling.

More information regarding the DIMCS test and another similar test that
use the SystemTestsCompareN class can be found in the following
MOM_interface draft PR: https://github.com/ESCOMP/MOM_interface/pull/97

The above PR also showcases how this new class may be utilized.

The API mostly mimics what was done in SystemTestsCompareTwo.
Here are a few specifics:

 - The argument `N` to the `__init__` specifies the number of case instances
 including the base case.
 - The optional arguments `run_suffixes` and `run_descriptions` may be provided.
 These must be list of strings with lengths of N.
- The `_case_setup` method must be implemented in the derived classes. This
method expects an integer `i` as a required argument. `i==0` corresponds to 
the base case while `i>0` correspend to subsequent cases throughout the 
SystemTestsCompareN class.

More API details are provided in the docstrings. 

Note: I believe we need to add a unit test and documentation before this PR can
be merged, but I thought I should get an early feedback before adding
those.

Test suite: aux_mom suite, DIMCS_Ld1.T62_t061.GMOM.cheyenne_intel, MEMCS_Ld1_Vnuopc_D.T62_t061.GMOM.cheyenne_gnu
Test baseline:
Test namelist changes:
Test status: b4b for already existing cases

Fixes https://github.com/ESMCI/cime/issues/3222

User interface changes?: a new test class added

Update gh-pages html (Y/N)?:

### DIFF
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,6 +1,169 @@
 ======================================================================
 
 Originator: Chris Fischer 
+Date: 12-16-21
+Tag: cime6.0.12
+Answer Changes: None
+Tests: scripts_regression_tests
+Dependencies:
+
+Brief Summary:
+    - Rename izumi.unified to izumi.cgd
+    - Fix test ptr to run_cmd_no_fail.
+    - chdir is redundent in CMakeLists.txt
+    - Fix paths in cdash testing
+    - Hot fix cdash testing/
+    - Update unit tests and linting.
+    - Fix SSL certification for testing database.
+    - Fix probing machine.
+    - cprnc compares NaN.
+    - Update izumi ESMF lib to 8.2.0 release.
+    - Update CCE for crayenv2 on cheyenne.
+    - Cleanup Scorpio I/O performance data in the run directory.
+    - Fix the N_TestUnitTest by setting CIME_NO_CMAKE_MACRO
+    - Add python 3.7.0 module load for izumi (CESM only).
+    - Fix broken links in documentation.
+    - Update cime externals.
+    - tools/configure was using deprecated config_compilers.xml system.
+    - Remove documentation sections for pieces no longer included in CIME.
+    - More grids for mizuRoute. 
+
+User interface changes: 
+
+PR summary: git log --oneline --first-parent [previous_tag]..master
+5e99566ae Merge pull request #4142 from ESMCI/fischer/izumi_cgd
+6ffadc5fd fix test ptr to run_cmd_no_fail
+d271ebbd8 chdir here is redundent, done in CMakeLists.txt
+1faf45586 fix paths in cdash testing
+3955917de hot fix cdash testing
+baf320df2 Merge pull request #4129 from jasonb5/updates_testing
+e725ab7cb Merge pull request #4128 from ESMCI/fischer/testreporter_ssl
+f0dca0d96 Merge pull request #4134 from billsacks/fix_machine_probe
+4961f0357 Merge pull request #4132 from jedwards4b/cprnc_nan_detect
+188e9fe2b Merge pull request #4124 from ESMCI/fischer/izumi_esmf
+b0f7080a9 update CCE for crayenv2 on cheyenne
+8c515e31c Merge pull request #4127 from ESMCI/jayeshkrishna/consolidate_spio_timing
+5a98a511b fix the N_TestUnitTest by setting CIME_NO_CMAKE_MACRO
+cde1dafda Merge pull request #4117 from peverwhee/python_load_module
+f9fe41210 Merge pull request #4123 from billsacks/fix_doc_links
+1686f03b8 update externals
+9f263ed3d Merge pull request #4122 from ESMCI/jgfouca/fix_tools_configure
+7f709b709 Merge pull request #4108 from billsacks/remove_some_docs
+c7813bce7 Merge pull request #3823 from ekluzek/mizuRoute
+
+
+Modified files: git diff --name-status [previous_tag]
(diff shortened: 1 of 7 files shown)


billsacks:
Thank you very much for your work on this @alperaltuntas ! I think this is going to be very useful moving forward. 

So far I have only looked at the example tests in the MOM6 PR, but based on what I see there, I think this looks great. I'm happy with your choice of having a single `_case_setup` routine that accepts an argument for which case it's dealing with: I see how that made one of the MOM6 tests much cleaner than my original thought of having separate, named routines for each case. (And in the situation where there is a non-trivial amount of work for each case, the test implementation could always use separate routines like `_case_one_setup` that are called by the `_case_setup` routine to do the actual work.)

I am totally swamped with work right now and need to stay focused on some other projects for the next few weeks. So I won't be able to give this a careful review for a while. If you prefer to bring this in on a sooner time-frame, we should get someone else to review it.

As you say in your initial comment, it would be good to have some unit tests for this, similar to what's done for SystemTestsCompareTwo. (I'm not sure if you necessarily need equivalents for all of the tests of SystemTestsCompareTwo, but some testing would be good to make sure the critical functionality continues to work over time.)

Also, can you briefly mention whether you have done some manual testing to ensure that failures are reported when they should be at various stages? – For example, if the build fails for one of the cases (first case or a later case), if the run fails for one of the cases (first case or a later case), and if the comparison fails for one of the cases (second case or a later case, assuming that they are all compared against the first case).

I'm trying to think about what the long-term vision is here: It seems like, long-term, we don't want to maintain separate SystemTestsCompareN and SystemTestsCompareTwo. I originally proposed making SystemTestsCompareTwo be a lightweight wrapper on top of SystemTestsCompareN, maintaining backwards compatibility for existing component tests as well as providing a slightly simpler interface for the common two-case tests. However, I'm not sure whether that makes the most sense. An alternative would be to leave them both in place for some time but treat SystemTestsCompareTwo as deprecated, and ask components to modify their tests to use the SystemTestsCompareN relatively soon; then we could eventually remove SystemTestsCompareTwo. Do you have a thought on the best approach? I think it depends on how much work it would be to turn SystemTestsCompareTwo into a lightweight wrapper on top of SystemTestsCompareTwo. One related question is: Are all features from SystemTestsCompareTwo now implemented in SystemTestsCompareN, or are there some things that you haven't implemented yet?

Again, thanks a lot for this work!

alperaltuntas:
Thanks, @billsacks. This is needed on a sooner time-frame, but I am not sure when I'll be able to add a unit test. If I can do that soon, we can get someone else to review this.

> Also, can you briefly mention whether you have done some manual testing to ensure that failures are reported when they should be at various stages? ...

I have only confirmed successful reporting of build/runtime failures for the base case and comparison failures for the subsequent cases. I'll check others too.

As for the long-term vision, I don't have a strong preference but leaving them both in place for some time and deprecating SystemTestsCompareTwo makes the most sense to me.

> Are all features from SystemTestsCompareTwo now implemented in SystemTestsCompareN, or are there some things that you haven't implemented yet?

Unless I missed something, all features should be there in SystemTestsCompareN.

jedwards4b:
@alperaltuntas   Can you run black on the file scripts/lib/CIME/SystemTests/system_tests_compare_n.py ?
This should fix the github test failure.  Is there more that needs to be done here before merging?

billsacks:
> Is there more that needs to be done here before merging?

Realistically, I don't know if / when I'll get a chance to do a careful review of this. Given that it is currently only used for some MOM tests, I'm happy to trust that Alper has done sufficient review and testing of it, so I'm happy to let this come in whenever Alper feels it's ready, if others are okay with it coming in without careful review. Alper, do let us know if there are pieces of this that you feel could warrant a review – such as pieces you are less certain about.

That said, I do feel it's important to get some unit tests in place for this before long – whether it's done before the PR is merged or relatively soon afterwards – if this will start to be used more generally, which I think it eventually should be. A big part of my motivation for putting together SystemTestsCompareTwo initially was that we kept finding issues in test scripts, some of which led to the test reporting a PASS when in fact there was a failure. Given the criticality of this testing infrastructure, this experience made me feel that we should have extensive unit testing in place for the test system, because otherwise it can be hard to detect the worst kinds of errors that can creep into the test system under maintenance – i.e., errors where tests are reported as PASSing when in fact they should be failing.

alperaltuntas:
I ran black on the new test module and carried out more extensive tests. I believe this is now ready to be reviewed and merged. We need this module for MOM6 development sooner rather than later, so if that's okay, I suggest we omit unit tests for now and I can subsequently try submitting a PR with a unit test.

jedwards4b (approved):

PR #4170, opened 2022-02-02
changes for implementing nemo in cesm2.3

daniele-peano:

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/Tools/mkDepends
+++ b/scripts/Tools/mkDepends
@@ -346,7 +346,7 @@ foreach $f (keys %file_includes) {
     }
 
     # Expand
-    for ($i = 0; $i <= $#expand_incs; ++$i) {
+    foreach $i (0..$#expand_incs) {
 	push @expand_incs, @{ $include_depends{$expand_incs[$i]} };
     }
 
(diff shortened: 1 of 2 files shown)


scripts/Tools/mkDepends
```diff
@@ -346,7 +346,7 @@ foreach $f (keys %file_includes) {
     }
 
     # Expand
-    for ($i = 0; $i <= $#expand_incs; ++$i) {
```
jedwards4b, line 349: why is this change required?

jedwards4b (approved):

scripts/Tools/mkDepends
daniele-peano: When running the model on Zeus with NEMO, the build of NEMO becomes slow and uses a lot of memory with the original code. 
With the suggested code this problem is solved.

jedwards4b (approved):

jedwards4b:
Minor changes needed for the NEMO ocean model.

PR #4172, opened 2022-02-04
Changes to CIME to make CICE6 the default component.

dabail10:
This changes the default sea ice component from CICE5 to CICE6. The only CIME changes are to replace "cice6" with "cice5" in three files.

Test suite: aux_cice (formerly aux_cice6)
Test baseline:
Test namelist changes:
Test status: Changing from CICE5 to CICE6 is answer changing for all active sea ice cases. They have been deemed to have the same climate.

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/XML/archive_base.py
+++ b/scripts/lib/CIME/XML/archive_base.py
@@ -99,7 +99,7 @@ class ArchiveBase(GenericXML):
         # remove when component name is changed
         if model == "fv3gfs":
             model = "fv3"
-        if model == "cice6":
+        if model == "cice5":
             model = "cice"
         if model == "ww3dev":
             model = "ww3"
@@ -205,7 +205,7 @@ def _get_extension(model, filepath, ext_regexes):
     # Remove with component namechange
     if model == "fv3gfs":
         model = "fv3"
-    if model == "cice6":
+    if model == "cice5":
         model = "cice"
     if model == "ww3dev":
         model = "ww3"
--- a/ChangeLog
+++ b/ChangeLog
@@ -1,5 +1,29 @@
 ======================================================================
 
+Originator: David Bailey
+Date: 1-27-22
+Tag: cime6.0.1?
+Answer Changes: Yes. CICE6 is not bfb with CICE5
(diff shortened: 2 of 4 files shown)


jedwards4b (approved):

scripts/lib/CIME/XML/archive_base.py
```diff
@@ -99,7 +99,7 @@ def get_all_hist_files(self, casename, model, from_dir, suffix="", ref_case=None
         # remove when component name is changed
         if model == "fv3gfs":
             model = "fv3"
-        if model == "cice6":
+        if model == "cice5":
```
jedwards4b, line 102: the changes in config_files.xml are specific to model cesm.  So the changes in archive_base.py and case_st_archive.py will need to be applied only when cime_model == cesm.

jedwards4b (changes requested):
We will need to add something so that the changes in python scripts only apply for cime_model == cesm.

scripts/lib/CIME/XML/archive_base.py
dabail10: Does E3SM have a cice6 or a cice5? Won't it ignore this anyway?

rljacob: Could you just add an .or. statement?  If model = (cice5 or cice6)

rljacob: Actually we don't have cice5 or cice6.  We just have "cice".  So this code probably doesn't affect E3SM.

jedwards4b: oh good point - it only has cice so I think that it's okay.

jedwards4b (approved):

jedwards4b:
False alarm - e3sm will never trigger these statements.

mvertens:
@jedwards4b @rljacob - can this PR be merged?

jedwards4b:
Yes

rljacob:
yes

rljacob (approved):

PR #4179, opened 2022-02-10
check current machine and exit if not case machine

### LINKED ISSUE #4178: Throw a sensible error message when submitting from the wrong machine.
I think this is the right place for this! (@mvertens is this the right place?)
This may just be me, but I am often moving back and forth between cheyenne and casper for example, and occasionally end up submitting a job I had built on cheyenne on casper which throws a traceback of course, and puzzles me for a while. Could we have a sensible exit that says straight up that I am not submitting on the machine/architecture I had originally built it on. Trivial I know, but hey.

jedwards4b:
When initializing the case object for an existing case make sure that it matches the machine you are on,
throw an error otherwise

Test suite:scripts_regression_tests.py  hand testing on casper and cheyenne
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #4178

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/Tools/case.setup
+++ b/scripts/Tools/case.setup
@@ -72,16 +72,25 @@ def parse_command_line(args, description):
         "Use should use this if you have local modifications to these files that you want to keep.",
     )
 
+    parser.add_argument(
+        "-N",
+        "--non-local",
+        action="store_true",
+        help="Use when you've requested a machine that you aren't on. "
+        "Will reduce errors for missing directories etc.",
+    )
+
+
     args = CIME.utils.parse_args_and_handle_standard_logging_options(args, parser)
 
-    return args.caseroot, args.clean, args.test_mode, args.reset, args.keep
+    return args.caseroot, args.clean, args.test_mode, args.reset, args.keep, args.non_local
 
 
 ###############################################################################
 def _main_func(description):
     ###############################################################################
-    caseroot, clean, test_mode, reset, keep = parse_command_line(sys.argv, description)
-    with Case(caseroot, read_only=False, record=True) as case:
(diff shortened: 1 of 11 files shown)


jedwards4b:
Are there situations where this check will be problematic?  Should we have an override option?

jedwards4b:
@jgfouca is there a way to know (when initializing the case) if we are using that feature?

jgfouca:
@jedwards4b  `--non-local` flag for create_newcase.

jgfouca (approved):

fischer-ncar (approved):

jedwards4b:
Having trouble with:
```
FAIL: test_rebless_namelist (CIME.tests.test_sys_bless_tests_results.TestBlessTestResults)
  File "/glade/work/jedwards/sandboxes/cesm2_3_alpha08a.cice5/cime/scripts/lib/CIME/tests/test_sys_bless_tests_results.py", line 212, in test_rebless_namelist
    self._create_test(compargs)
  File "/glade/work/jedwards/sandboxes/cesm2_3_alpha08a.cice5/cime/scripts/lib/CIME/tests/base.py", line 236, in _create_test
    output = self.run_cmd_assert_result(
  File "/glade/work/jedwards/sandboxes/cesm2_3_alpha08a.cice5/cime/scripts/lib/CIME/tests/base.py", line 137, in run_cmd_assert_result
    self.assertEqual(stat, expected_stat, msg=msg)
AssertionError: 100 != 0 : 
```

jgfouca:
@jedwards4b , you may need to add some verbose printing (command + output/stderr) to see what the problem is.

PR #4184, opened 2022-02-22
Allow unbounded compilers to modify max tasks per node

amametjanov:
This is needed on OLCF's Summit and Ascent machines with 6 configured compilers.

Test suite: by-hand
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

User interface changes?: none


rljacob:
What does "unbounded" mean in this?  Was "MaxOccurs" a limit on the number of compilers on a machine?

jedwards4b (approved):

jedwards4b:
It was a limit on the number of times the MAX_TASKS_PER_NODE setting could appear in a machine definition.  I assume that it appears for each compiler used and he needs more than 4 compilers on a certain machine.

amametjanov:
Yes, when I tried adding one more entry to max tasks per node to the existing 4, I ran into an `xmllint` schema validation error.

PR #4190, opened 2022-03-08
neon inputdata update

jedwards4b:
Updates pointer to neon data server, the new address is a virtual one so we hopefully won't need to change this again.

Test suite: hand testing neon data sites
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


ekluzek (approved):
Simple change to add another location to download NEON data from. There's not much to review. There are some whitespace formatting changes as well which is fine.

The one thing I wondered about was upgrading the SUCCESS from info logging level to warning. I suppose that's done to make sure that it always happens. But, it's not really a warning.

PR #4195, opened 2022-03-11
Fix generated baseline from disappearing when using --retry

### LINKED ISSUE #4193: Baselines being removed from generated baseline directory
PR #4154 is overly aggressive in remove baselines when using --retry in create_test.

For example
./create_test --xml-category aux_cime_baselines --generate TEST_BASELINE2 --compare cesm2_3_alpha08b --no-build --retry 2
./create_test --xml-category aux_cime_baselines --generate TEST_BASELINE1 --compare cesm2_3_alpha08b --no-build

TEST_BASELINE2 has been cleaned out, while TEST_BASELINE1 still has the baseline namelist files.

fischer-ncar:
Using --retry in create_newcase was causing generated baseline to be removed from the
generated baselines directory.  Add a check to see if retry is set before the -o option removes
the old baselines.

Test suite: ./create_test by hand, scripts_regression_tests(still running)
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #4193

User interface changes?:

Update gh-pages html (Y/N)?:


scripts/lib/CIME/test_scheduler.py
```diff
@@ -340,7 +342,7 @@ def __init__(
                     test_baseline = os.path.join(full_baseline_dir, test_name)
                     if os.path.isdir(test_baseline):
                         existing_baselines.append(test_baseline)
-                        if allow_baseline_overwrite:
+                        if allow_baseline_overwrite and retry == "None":
```
billsacks, line 345: Should this really be `retry == "None"` or should it be `retry is None`?

fischer-ncar: Yes is should be.  I just pushed the change.

jedwards4b (approved):

scripts/lib/CIME/test_scheduler.py
fischer-ncar: That is it should be retry is None.

billsacks:
With the change here, will there be a regression in #4152 / #4154 when running with both the `--retry` and `-o` flags? My understanding is that:
1. Prior to #4154 there were problems with `-o` sometimes leaving stale files around
2. #4154 fixed that problem, but introduced new problems when combining `-o` with `--retry`
3. This PR fixes the problems in (2) but I think brings back the original problems (1) when combining `-o` with `--retry`

It seems like the ideal solution would be: when running with `-o` and `--retry`, do an initial clearing of any baselines in the first go-around, but then do not reclear them for each retry. I'm not sure how hard that would be.

PR #4197, opened 2022-03-15
Only clear baselines when run_count is 0 and -o is used.

fischer-ncar:
PR #4195 introduced an issue that caused a regression to #4152/#4154.  When 
create_test uses -o and -retry the baselines were no longer removed, which could 
leave behind stale files.
Now the baselines can only removed when run_count == 0 instead of when retry is None.

Test suite: by hand, scripts_regression_tests
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes

User interface changes?:

Update gh-pages html (Y/N)?:


billsacks (approved):
Great, thanks a lot for the fix!

jedwards4b (approved):

PR #4199, opened 2022-03-17
make pio_async_interface a compvar

jedwards4b:
First step in the direction of implementing async IO in CESM.   We need to make the async option a component option rather that a global.   We also need some change so that pio options can go into 
nuopc.runconfig instead of namelists. 

Test suite: scripts_regression_tests.py cesm prealpha testing
Test baseline:
Test namelist changes:

Requires CMEPS PR 275

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/.github/workflows/srt.yml
+++ b/.github/workflows/srt.yml
@@ -55,7 +55,7 @@ jobs:
       PNETCDF_PATH: ${HOME}/pnetcdf
       CIME_MODEL: cesm
       CIME_DRIVER: mct
-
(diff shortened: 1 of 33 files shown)


mvertens (approved):

jedwards4b:
Note to self - test  qcmd -- ./create_test ERS.ne5_ne5_mg37.FADIAB.cheyenne_intel.cam-outfrq3s_bwic
is not working here.   Fails while compiling shr_pio_mod.F90

jedwards4b:
With the merge of #4215 this test now passes.

Issue #4202, opened 2022-03-24
run_tests.py doesn't work if you try to let it auto-detect the machine


billsacks:
Currently run_tests.py requires you to specify the `--machine` argument; otherwise, it dies with `ERROR: PFUNIT_PATH not found for this machine and compiler`.

I think I see how to fix it; PR incoming shortly.

PR #4203, opened 2022-03-24
Fix run_tests.py when no machine is given

billsacks:
Test suite: Currently running scripts_regression_tests on cheyenne; will amend if any tests fail
Test baseline: none
Test namelist changes: no
Test status: bit for bit

Fixes ESMCI/cime#4202

User interface changes?: N

### DIFF
--- a/scripts/fortran_unit_testing/run_tests.py
+++ b/scripts/fortran_unit_testing/run_tests.py
@@ -427,7 +427,7 @@ def _main():
         "{}-DOS={} -DMACH={} -DCOMPILER={} -DDEBUG={} -DMPILIB={} -Dcompile_threaded={} -DCASEROOT={}".format(
             "" if not cmake_args else "{} ".format(cmake_args),
             os_,
-            machine,
+            machobj.get_machine_name(),
             compiler,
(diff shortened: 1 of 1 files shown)


scripts/fortran_unit_testing/run_tests.py
```diff
@@ -427,7 +427,7 @@ def _main():
         "{}-DOS={} -DMACH={} -DCOMPILER={} -DDEBUG={} -DMPILIB={} -Dcompile_threaded={} -DCASEROOT={}".format(
             "" if not cmake_args else "{} ".format(cmake_args),
             os_,
-            machine,
+            machobj.get_machine_name(),
```
billsacks, line 430: If the machine argument is given, then I think this will give the same result as before; but this version works if the machine argument isn't given and needs to be inferred.

billsacks:
Hold off on merging for a bit: I ran into another issue that I want to try to fix here.

billsacks:
Actually, the changes to fix the other issue are getting extensive. Let's keep this PR as is and I'll open a new one for the other issue.

jgfouca (approved):

Issue #4204, opened 2022-03-24
run_tests.py doesn't seem to look for cmake macros files in ~/.cime


billsacks:
run_tests.py doesn't pick up my gnu_bishorn.cmake file in my ~/.cime directory. It looks like, in a full case, this is picked up by case_setup; the configure function that run_tests.py calls looks at extra-machines-dir, but not .cime.

I propose changing the configure function so that it checks for macros files in ~/.cime as well as extra-machines-dir. I'm working on a fix and will open a PR once it's ready.

PR #4205, opened 2022-03-25
Extend configure to get cmake macros files from the .cime directory

billsacks:
Extend configure to get cmake macros files from the .cime directory. This also fixes the use of cmake macros from the extra machines dir in configure, which seemed to be broken before.

This is needed for run_tests.py.

- `ERIO_Ln11.f09_g16.X.cheyenne_intel` failed initially. Reran (directly via `create_test`) and it passed

Test baseline: N/A
Test namelist changes: none
Test status: bit for bit

Fixes ESMCI/cime#4204

User interface changes?: N

### DIFF
--- a/scripts/lib/CIME/BuildTools/configure.py
+++ b/scripts/lib/CIME/BuildTools/configure.py
@@ -29,7 +29,7 @@ from CIME.XML.env_mach_specific import EnvMachSpecific
 from CIME.XML.files import Files
 from CIME.build import CmakeTmpBuildDir
 
-import shutil, glob
+import shutil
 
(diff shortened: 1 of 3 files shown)


jgfouca (approved):
Thanks!

jgfouca:
@billsacks , looks like you need to run the pre-commit script (I assume for black formatting).

billsacks:
Okay, should be fixed now. The pre-commit thing is nice! (Thanks @jasonb5 .) I had forgotten about that. I had actually been careful to run everything through black (which I now do in my editor), but forgot to check pylint. pre-commit makes that easy now.

PR #4207, opened 2022-03-25
wait_for_tests: Try multiple ctest drop methods

jgfouca:
... for uploading test results. cori appears to work with http but not https as a drop method.

Test suite: scripts_regression_tests (wait_for_tests, Jenkins only)
Test baseline:
Test namelist changes:
Test status: bit for bit

User interface changes?:

Update gh-pages html (Y/N)?:


jasonb5 (approved):
LGTM

PR #4208, opened 2022-03-28
Error on PROJECT number starting with zero

daniele-peano:

Some project names on Zeus are numbers starting with zero, such as 0490.
We have to skip this check on the retrieving of the PROJECT name otherwise the name is converted into 490, which is not recognized as the correct project number.

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/scripts/lib/CIME/XML/env_batch.py
+++ b/scripts/lib/CIME/XML/env_batch.py
@@ -597,11 +597,12 @@ class EnvBatch(EnvBase):
                     else:
                         rval = val
 
+                    if flag != "-P":
                     # We don't want floating-point data
-                    try:
-                        rval = int(round(float(rval)))
-                    except ValueError:
-                        pass
+                        try:
+                            rval = int(round(float(rval)))
+                        except ValueError:
+                            pass
 
                     # need a correction for tasks per node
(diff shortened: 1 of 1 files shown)


jedwards4b:
Hi @daniele-peano how about this?  It doesn't require the introduction of any new flags.   

```
                    # We don't want floating-point data                                                                                
                    if '.' in str(rval):                                                                                               
                        try:                                                                                                           
                            rval = int(round(float(rval)))                                                                             
                        except ValueError:                                                                                             
                            pass                                                                                                       
  ```

daniele-peano:
I tested your suggested change and it is working in our case.
Then, I agree with using your solution.

sarich:
> rval = int(round(float(rval)))
I don't recall the exact circumstances either, but @jfouca is right that int() can't handle a string with a decimal. It also truncates the number which might cause trouble in floating point arithmetic, hence the round() function. In python2 round() returned another floating point, so the int() was still needed. But I believe python3's round() will return an integer if no second argument is used.

jedwards4b:
Superseded

PR #4212, opened 2022-03-30
add a valid PTS_DOMAIN_FILE in aquaplanet mode

mvertens:
Add a valid PTS_DOMAIN_FILE in aquaplanet mode. 
This was a problem for the CAM standard testing for SCAM.

Test suite: Verified that ./create_newcase --case test3 --compset QPC6 --res T42_T42 --run-unsupported now
has a valid PTS_DOMAIN_FILE.
Test baseline: NA
Test namelist changes: NA
Test status: bit for bit

Fixes:None

User interface changes?: No


jedwards4b (approved):
Why 'unset' and not None?

mvertens:
@jedwards - because unset its the value of the file node - which by default is unset if there is no domain file setting.
I have tried with None and it does not work.

PR #4215, opened 2022-04-01
Correct ATM_GRID definition for CAM SE grids.

### LINKED ISSUE #4214: CAM SE grids are not fully defined in case
In particular test ERS.ne5_ne5_mg37.FADIAB.cheyenne_intel.cam-outfrq3s_bwic does not
correctly define ATM_GRID.

jedwards4b:
Unindent the line that defined {COMP}_GRID so that atm grid is defined in test ERS.ne5_ne5_mg37.FADIAB.cheyenne_intel.cam-outfrq3s_bwic

Test suite: scripts_regression_tests.py, ERS.ne5_ne5_mg37.FADIAB.cheyenne_intel.cam-outfrq3s_bwic

Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #4214 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/grids.py
+++ b/CIME/XML/grids.py
@@ -220,7 +220,7 @@ class Grids(GenericXML):
 
         for comp_name in component_grids.get_compnames(include_mask=True):
             for grid_name in component_grids.get_comp_gridlist(comp_name):
-                # Determine grid name with no nlev suffix if there is one
+                # Determine grid name with no nlev suffix if there is one 
                 grid_name_nonlev = grid_name
                 levmatch = re.match(atmlevregex, grid_name)
                 if levmatch:
@@ -312,9 +312,6 @@ class Grids(GenericXML):
                     domains[comp_name + "_NX"] = 1
                     domains[comp_name + "_NY"] = 1
 
-            # set up dictionary of domain files for every component
-            _add_grid_info(domains, comp_name + "_GRID", grid_name)
-
             if driver == "mct":
                 # mct
                 file_nodes = self.get_children("file", root=domain_node)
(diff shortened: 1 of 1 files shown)


mvertens (approved):

gold2718:
This change has been in for nearly a year. It was made by @billsacks in #3982.
Do you know if this change preserves that functionality?

mvertens:
@gold2718 - good point about #3982. @jedwards4b - can we please ask @billsacks to review this PR as well.

jedwards4b:
I'm not sure how it was missed.  It came up in this simple grid test - have you run this test in nuopc before?
I think that the reason it comes up now may be because of the refactor of grids done in https://github.com/ESMCI/cime/pull/4145  
I haven't seen any evidence in testing that it is causing any side-effects.

gold2718:
Our pretag test (which is passing) is:
`ERC_D_Ln9_Vnuopc.ne16_ne16_mg17.FADIAB.cheyenne_intel.cam-terminator`
We do test `ne5_ne5_mg37` on Izumi / nag:
`ERI_D_Ln18_Vnuopc.ne5_ne5_mg37.FADIAB.izumi_nag.cam-outfrq3s_bwic` is also passing.

The problem you ran into this morning could be a chicken and egg thing since we have been working on getting CAM to pass the NUOPC tests (as announced in the CSEG meeting).

jedwards4b:
What I know is that this test was failing on my PR #4199 and applying the fix from this PR solves the issue.
Are you saying that if I update the cam tag this test will pass without the change?   What cam tag should I use?

jedwards4b:
Well what cam tag do I need to overcome this problem?

gold2718:
To repeat what I just wrote and what was said in the CSEG meeting, we are _working on_ the PR where CAM passes all the NUOPC tests. Normally, this means that the tag has not yet been made and that is true here. The process has been made more difficult by the churn in CIME and we now have a catch-22 where a fix we need (#4212) is on top of a change we cannot accept (#4172) because it interferes with the CESM tag sequence (with the CICE changes) we all agreed to at the CSEG meeting.
Chris is working on a solution to that but we need a stable CIME tag in order to be able to make the CAM tag.

billsacks:
Thanks for pointing me to this. It looks like, even though I reworked this code pretty significantly in #3982, I did not change the placement of the relevant line of code relative to the surrounding blocks / indentations. I don't see anything about this current PR that would break the changes in #3982. That is to say: I'm fine with this change if others are.

PR #4217, opened 2022-04-04
Add Total_Build time to end of build_times.txt

jgfouca:
... in e3sm's provenance code.

Also, minor fix for list_e3sm_tests.

Test suite: by hand
Test baseline:
Test namelist changes:

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/provenance.py
+++ b/CIME/provenance.py
@@ -66,6 +66,7 @@ def _extract_times(zipfiles, target_file):
         fd.write(contents)
         fd.write("Total_Build {}".format(str(total_build_time)))
 
+
 def _run_git_cmd_recursively(cmd, srcroot, output):
     """Runs a git command recursively
 
--- a/CIME/provenance.py
+++ b/CIME/provenance.py
@@ -44,6 +44,7 @@ def _get_batch_job_id_for_syslog(case):
 def _extract_times(zipfiles, target_file):
 
     contents = "Target Build_time\n"
+    total_build_time = 0.0
     for zipfile in zipfiles:
         stat, output, _ = run_cmd("zgrep 'built in' {}".format(zipfile))
         if stat == 0:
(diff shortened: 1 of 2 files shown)


jedwards4b (approved):

jedwards4b:
@jgfouca these should be fixed by a merge of master.

jgfouca:
@jedwards4b , git is telling me my branch is up to date.

jgfouca:
Ah, ok. Thanks

sarats:
@jgfouca It would be great if you can you apply this change to maint-2.0 as well.

CIME/provenance.py
```diff
@@ -54,8 +55,16 @@ def _extract_times(zipfiles, target_file):
                     target, the_time = items[1], items[-2]
                     contents += "{} {}\n".format(target, the_time)
 
+        stat, output, _ = run_cmd("zgrep -E '^user [0-9.]+$' {}".format(zipfile))
```
sarats, line 58: I think we should use the "real" time reported by the time command to account for all the I/O and system/kernel overheads involved.

CIME/provenance.py
```diff
@@ -44,6 +44,7 @@ def _get_batch_job_id_for_syslog(case):
 def _extract_times(zipfiles, target_file):
 
     contents = "Target Build_time\n"
+    total_build_time = 0.0
     for zipfile in zipfiles:
         stat, output, _ = run_cmd("zgrep 'built in' {}".format(zipfile))
```
sarats, line 49: In a similar vein, is the target build time reporting elapsed (real) or user time?

jgfouca: The individual targets use the python `time` library, so should not have the issue of distinguishing between user and real times.

CIME/provenance.py
jgfouca: I think you're right.

CIME/provenance.py
sarats: Just for my reference, where is the CMAKE wrapper you use for recording this time info?

jgfouca: @sarats, `cime/CIME/Tools/e3sm_compile_wrap.py` . This is the customizable target wrapper for E3SM. The default version (version the lives in the repo) just does timings.

PR #4221, opened 2022-04-07
Changes needed for E3SM

jgfouca:
Our PIO (scorpio) does not work with the latest genf90 changes,
specifically the addition of 'short' support. This commit changes
the gen90 submodule to point to a genf90 commit with short support
reverted.

Also, our Jenkins jobs rely on the 'reuse host credential' feature
and so we must have all submodules be using the ssh protocol.

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jgfouca:
@jedwards4b , @jasonb5 , I'm not really sure what to do here since we (cesm and e3sm) are sharing this repo.

jedwards4b (changes requested):
We cannot accept this, we need the short data type.

jgfouca:
@jedwards4b , yeah, I was not expecting to merge this in to master. I was more looking to open a discussion on how to manage the genf90 submodule. I suspect you also do not want to manage it via ssh keys since that requires all users to have a github acct, but we need it to be this way for Jenkins to work.

jgfouca:
I believe I can coax the scorpio maintainers to get it working with the latest genf90. The ssh/https issue on the submodule is a longterm problem though.

jedwards4b:
I thought I could modify genf90 to make the short type optional, but we can't 
require everyone to have a github account so the only other solution would be to
move genf90 out of cime.

rljacob:
Why can't E3SM use its own genf90 we could keep in E3SM/externals ?

rljacob:
Oh is it because this is buried deep in the CIME python package?

rljacob:
Remind me why CIME, the python package, even needs genf90 ?

jgfouca:
@rljacob , genf90 is coupled with our build system. Any fortran file with the `.F90.in` gets processed through gen90 to create a compilable .F90 file.

rljacob:
another option:  undo the submodule and just put genf90 in CIME as part of the repo.  That's not a code getting lots of updates.

But that doesn't solve the "short" problem.   Instead of hardcoding the use of genf90 in that location, make it a variable that can be set in config_files.xml.

jgfouca:
@rljacob , I agree. I think any benefits of having it a submodule are outweighed by the ssh/https protocol problem.

The maintainers of scorpio will need to make sure the code works with latest genf90.

rljacob:
So if any CIME-build user wants to autogenerate Fortran, they must use genf90 and the genf90 included in CIME?   I could see how different models/libs may want their own tool for that.  SCORPIO should be able to call its own internal version of genf90 when built in E3SM (which it has in scorpio/src/genf90).

jedwards4b:
We can in fact use different genf90 for scorpio build - the path to genf90
is set [here](https://github.com/ESMCI/cime/blob/master/CIME/build_scripts/buildlib.pio#L98) and we can add a different path for scorpio.

rljacob:
Great. Lets do that.

jedwards4b:
Note that that won't resolve the issue of using ssh to git.

rljacob:
Lets remove the submodule and put it back as regular code for that.

jgfouca:
Had to also remove the probed-machine check which broke a few machines on our side. I don't know why this check was added.

CIME/case/case.py
```diff
@@ -184,12 +184,7 @@ def __init__(self, case_root=None, read_only=True, record=False, non_local=False
                     machobj = Machines(machine=mach, extra_machines_dir=extra_machdir)
                 else:
                     machobj = Machines(machine=mach)
-                probed_machine = machobj.probe_machine_name()
```
jedwards4b, line 187: This check was added because there are instances of multiple supported machines which share a common file system.  Users were creating a case on one and then trying to use the case on another which generated a nonsensical error.   This probe was added to create a reasonable error message.   Perhaps rather than remove this we can discuss why it is causing issues on your systems.

jedwards4b: The issue associated with this change: https://github.com/ESMCI/cime/issues/4178

rljacob: One case was cori which is one front end for two machines (cori-knl and cori-haswell).   You can't tell which one you're intending to run on by probing the login machine name.   For connivence, the probed name is set to "cori-knl" and if you want to run on haswell you have to use the --machine argument.  This check overrides what the user specified with --machine.

jedwards4b: Good point - we can fix that by introducing the alternative NODENAME_REGEX as we use on 
perlmutter, that is:  <NODENAME_REGEX>$ENV{NERSC_HOST}:perlmutter</NODENAME_REGEX>
We can add the same for cori-knl.

rljacob: How does that work and where does it get added?

I still think that probe statement should be surrounded by an if (user didn't specify machine).

jedwards4b: It's in the machine definition in config_machines.xml.   NERSC sets an env variable, NERSC_HOST, for each system.
I tested this at NCAR where we have a common file system but different front ends, nersc has both the filesystem and the frontend in common.    I'll try it at nersc and see if I can refine the solution.

ndkeen: I don't think NERSC_HOST env var will help here as it is "cori". We login to cori and then user can build&submit for 2 diff types of compute nodes that are chosen via slurm batch.

jedwards4b:
I've opened issue https://github.com/ESMCI/cime/issues/4226 and will try to address this tomorrow.

jgfouca:
@jedwards4b , it looks like https://github.com/ESMCI/cime/pull/4228 is not merged yet and @rljacob appeared to have an objection. Once that goes in, that still leaves the issue of how to manage the genf90 submodule. Rob suggested just going back to storing a copy like before, which I'm fine with but I don't know if everyone agrees.

jedwards4b (approved):

PR #4222, opened 2022-04-08
Add help to the --driver option in create_newcase, for CESM, change options for it for CESM and E3SM. leave off for UFS

### LINKED ISSUE #4220: Remove argparse.SUPPRESS for --driver option in create_newcase
The --driver option is suppressed in the help for create_newcase. It is available for create_test. For CESM it would be nice to have help for it so that users can see that it's an option, especially for our transition from MCT to NUOPC. I assume the only reason it was suppressed was it was considered an advanced feature that most people wouldn't use.

There's two other options that are suppressed --script-root and --test and for those I see why they are suppressed.

The change is just to remove the help=argparse.SUPPRESS and add a help string such as

"Override the top level driver type and use this one (changes xml variable COMP_INTERFACE) [this is an advanced option]"

ekluzek:
Add help to --driver option, but only for the CESM model

Adds help to the --driver option, for CESM and only allow nuopc and mct. For E3SM leave help suppressed, and only allow mct and moad. For UFS leave the option off.

Test namelist changes: none
Test status: bit-for-bit

Ran ./create_newcase --help

Fixes #4220 

User interface changes?: Y

### DIFF
--- a/CIME/scripts/create_newcase.py
+++ b/CIME/scripts/create_newcase.py
@@ -233,8 +233,13 @@ def parse_command_line(args, cimeroot, description):
     )
     if model == "cesm":
         drv_choices = ("mct", "nuopc")
+        drv_help = (
+            "Override the top level driver type and use this one "
+            + "(changes xml variable COMP_INTERFACE) [this is an advanced option]"
+        )
     elif model == "e3sm":
         drv_choices = ("mct", "moab")
+        drv_help = argparse.SUPPRESS
     else:
         drv_choices = None
 
@@ -243,8 +248,7 @@ def parse_command_line(args, cimeroot, description):
             "--driver",
             default=get_cime_default_driver(),
             choices=drv_choices,
-            help="Override the top level driver type and use this one "
-            "(changes xml variable COMP_INTERFACE) [this is an advanced option]",
+            help=drv_help,
         )
 
     parser.add_argument(
--- a/CIME/scripts/create_newcase.py
+++ b/CIME/scripts/create_newcase.py
@@ -232,10 +232,17 @@ def parse_command_line(args, cimeroot, description):
         help="Use a non-default location for input files. This will change the xml value of DIN_LOC_ROOT.",
     )
     if model == "cesm":
+        drv_choices = ("mct", "nuopc")
+    elif model == "e3sm":
+        drv_choices = ("mct", "moab")
+    else:
+        drv_choices = None
+
+    if drv_choices is not None:
         parser.add_argument(
             "--driver",
             default=get_cime_default_driver(),
-            choices=("mct", "nuopc", "moab"),
+            choices=drv_choices,
             help="Override the top level driver type and use this one "
             "(changes xml variable COMP_INTERFACE) [this is an advanced option]",
         )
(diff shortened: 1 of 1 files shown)


jgfouca (approved):

CIME/scripts/create_newcase.py
```diff
@@ -231,12 +231,14 @@ def parse_command_line(args, cimeroot, description):
...
-        "--driver",
-        default=get_cime_default_driver(),
-        choices=("mct", "nuopc", "moab"),
-        help=argparse.SUPPRESS,
-    )
+    if model == "cesm":
+        parser.add_argument(
+            "--driver",
+            default=get_cime_default_driver(),
+            choices=("mct", "nuopc", "moab"),
```
jedwards4b, line 238: moab is not supported by cesm, you should remove that choice.

jedwards4b (changes requested):

CIME/scripts/create_newcase.py
ekluzek: OK, I'll do that.

ekluzek:
OK, I added back in --driver for both cesm and e3sm, the only difference between the two are the list of valid options for each.

ekluzek:
@uturuncoglu what I have now --driver will be an option for both e3sm and cesm (with different valid options), but it won't be an option for UFS. Is that the right choice here?

CIME/scripts/create_newcase.py
```diff
@@ -231,12 +231,21 @@ def parse_command_line(args, cimeroot, description):
...
-    parser.add_argument(
-        "--driver",
-        default=get_cime_default_driver(),
-        choices=("mct", "nuopc", "moab"),
-        help=argparse.SUPPRESS,
-    )
+    if model == "cesm":
+        drv_choices = ("mct", "nuopc")
+    elif model == "e3sm":
+        drv_choices = ("mct", "moab")
```
jedwards4b, line 237: I think that you should put the argparse.SUPPRESS option back for e3sm.
UFS only uses nuopc so they don't need this option.

ekluzek: OK, I'll do that.

fischer-ncar (approved):

CIME/scripts/create_newcase.py
```diff
@@ -231,12 +231,21 @@ def parse_command_line(args, cimeroot, description):
...
+    else:
+        drv_choices = None
+
+    if drv_choices is not None:
```
jedwards4b, line 241: make the help option a variable and set it to help=argparse.SUPPRESS for e3sm and to
"Override the top level driver type and use this one "
            "(changes xml variable COMP_INTERFACE) [this is an advanced option]"
for cesm.

CIME/scripts/create_newcase.py
ekluzek: OK, did that.

CIME/scripts/create_newcase.py
ekluzek: Got it.

jedwards4b (approved):

rljacob (approved):

rljacob:
Title and description for this PR should be changed.

ekluzek:
Can this come in now that all the changes requested are done, and everyone has approved it?

PR #4228, opened 2022-04-14
only do this check on ncar systems

### LINKED ISSUE #4226: machine id fix #4179 breaks nersc systems
The PR https://github.com/ESMCI/cime/pull/4179 solved the machine id problem
at NCAR but creates a new problem at NERSC.  Since both cori-knl and cori-haswell
share the same front end, this change makes it impossible to submit to cori-knl.

jedwards4b:
This check should only be done on systems with a common filesystem but separate login nodes (ncar)
Test suite: Ran SMS.f19_g17.X.cheyenne_intel and tried to run on casper, ran  ./create_test SMS.f19_g17.X.cori-knl_intel SMS.f19_g17.X on cori.   
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #4226 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/case/case.py
+++ b/CIME/case/case.py
@@ -184,12 +184,13 @@ class Case(object):
                     machobj = Machines(machine=mach, extra_machines_dir=extra_machdir)
                 else:
                     machobj = Machines(machine=mach)
-                probed_machine = machobj.probe_machine_name()
-                if probed_machine:
-                    expect(
-                        mach == probed_machine,
-                        f"Current machine {probed_machine} does not match case machine {mach}.",
-                    )
+                if "NCAR_HOST" in os.environ:
+                    probed_machine = machobj.probe_machine_name()
+                    if probed_machine:
+                        expect(
+                            mach == probed_machine,
+                            f"Current machine {probed_machine} does not match case machine {mach}.",
+                        )
             self.initialize_derived_attributes()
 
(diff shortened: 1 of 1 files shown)


CIME/case/case.py
```diff
@@ -184,12 +184,13 @@ def __init__(self, case_root=None, read_only=True, record=False, non_local=False
                     machobj = Machines(machine=mach, extra_machines_dir=extra_machdir)
                 else:
                     machobj = Machines(machine=mach)
-                probed_machine = machobj.probe_machine_name()
-                if probed_machine:
-                    expect(
-                        mach == probed_machine,
-                        f"Current machine {probed_machine} does not match case machine {mach}.",
-                    )
+                if "NCAR_HOST" in os.environ:
```
rljacob, line 187: We should not be putting site-specific stuff like this in the CCS.

jedwards4b: I welcome suggestions for alternative solutions.

rljacob: I think the more portable way is a machine file setting that marks such a config.  But I'm fine with this going in now as is.

jedwards4b: @rljacob I've given some thought to how a field in the machine definition could resolve this issue, but haven't come up with anything, if you have any specific suggestions, I'm interested.

PR #4230, opened 2022-04-25
Add numeric time-stamp to jenkins archiving

amametjanov:
To prevent substring matching: e.g. `prod` archiving `prod_bench`

Test suite: e3sm_prod, e3sm_prod_bench
Test baseline: none
Test namelist changes: none
Test status: bit for bit

User interface changes?: N


jgfouca:
@amametjanov , you don't think we should remove old `cs.submit` files?

amametjanov:
This now matches both in my tests with
```
./cime/CIME/Tools/jenkins_generic_job -t e3sm_prod -O master --baseline-compare -b master --no-submit
```

jgfouca:
@amametjanov , looks good. I took the liberty of running the black code formatter on your PR so that it would pass all the checks. I will merge once everything passes.

PR #4243, opened 2022-05-04
Fix E3SM provenance

### LINKED ISSUE #4229: E3SM Provenance: Need to check archival of preview_run.log and GIT_* files
Related issue: https://github.com/ESMCI/cime/issues/3966

There were previously two attempts at recording this information in provenance. However, these files still don't appear in our performance archive. I can see preview_run.log files in a user's run directory but those didn't get archived.
https://github.com/ESMCI/cime/pull/3998
https://github.com/ESMCI/cime/pull/4055

Something to check, the following code snippet from last commit in above PR (https://github.com/ESMCI/cime/pull/4055/commits/b52ce0a07e34e895cd88a8d2545fb5eca7d6c0e0) is _apparently deleting_ the baseline preview_run.log file before making a copy with lid suffix. It should probably make a copy first and then delete. 

https://github.com/ESMCI/cime/blob/master/CIME/provenance.py#L516:L520
```
   if os.path.exists(base_preview_run):
        os.remove(base_preview_run)

    with open(base_preview_run, "w") as fd:
        case.preview_run(lambda x: fd.write("{}\n".format(x)), None)
```

We also need to check why GIT_* files never show up.
https://github.com/ESMCI/cime/blob/master/CIME/provenance.py#L669:675
`        "GIT_STATUS",        "GIT_DIFF",        "GIT_LOG",        "GIT_REMOTE",        "GIT_CONFIG",        "GIT_SUBMODULE_STATUS",
`

jasonb5:
Fixes copying GIT_* and preview_run.log into E3SM performance
archives.

Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes #4229 
User interface changes?: n

### DIFF
--- a/CIME/provenance.py
+++ b/CIME/provenance.py
@@ -411,6 +411,7 @@ def _save_prerun_timing_e3sm(case, lid):
         "Depends.{}.{}".format(mach, compiler),
         "software_environment.txt",
     ]
+
     for glob_to_copy in globs_to_copy:
         for item in glob.glob(os.path.join(caseroot, glob_to_copy)):
             safe_copy(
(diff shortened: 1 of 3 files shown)


jasonb5:
WIP, might have one more change to add.

jasonb5:
@sarats Has verified the output of my test to contain the correct provenance in the performance archive.

jgfouca (approved):

PR #4245, opened 2022-05-04
Allow any case env to be used as a directives selector

jgfouca:
Some users want to be able to select directives based on compset. It is a lot easier to skip XSD attribute verification for directives element rather than list every possible case env setting.

Test suite: ./scripts-regression-tests
Test baseline:
Test namelist changes:
Test status:  bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/data/config/xml_schemas/config_batch.xsd
+++ b/CIME/data/config/xml_schemas/config_batch.xsd
@@ -146,10 +146,7 @@
       <xs:sequence>
         <xs:element maxOccurs="unbounded" ref="directive"/>
       </xs:sequence>
-      <xs:attribute ref="queue"/>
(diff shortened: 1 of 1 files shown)


jgfouca:
@jedwards4b , I'm going to go ahead and merge this since the need is time sensitive for E3SM and I suspect you don't have an issue with this tiny PR. We can discuss if this is incorrect.

jedwards4b (commented):
Looks good

rljacob:
@jgfouca do you remember the motivation for this?

jgfouca:
@rljacob , I think on some machines, the batch submission changes depending on if you selected a GPU compiler vs. a standard compiler.

PR #4246, opened 2022-05-05
Fix sys path for configure

jgfouca:
It needed an additional `..`.

Test suite: by-hand
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jasonb5 (approved):
LGTM

PR #4255, opened 2022-05-11
Fix buildlib.pio genf90 path for scorpio

jgfouca:
The path was incorrect. The error was hidden because scorpio uses its internal
genf90 if GENF90_PATH is not found but this does not work for scorpio_classic.

Test suite: by-hand
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


dqwu (approved):
Manually tested on ANL GCE nodes, looks good to me.

PR #4261, opened 2022-06-02
Allow models to own their own config_files file

### LINKED ISSUE #4259: Refactor config_files.xml in CIME to contain bare minimum
Really, all it should point to is the "real" config_files.xml that lives in the host repo. I think something like this:

```
<entry_id version="2.0">

  <entry id="MODEL_CONFIG_FILES">
    <type>char</type>
    <default_value>$SRCROOT/cime_config/config_files.xml</default_value>
    <group>case_last</group>
    <file>env_case.xml</file>
    <desc>file containing paths</desc>
  <entry>
</entry_id>
```

The motivation for this change is that I feel E3SM developers should be able to tweak things like PES location for components without having to issue PRs to CIME.

jgfouca:
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #4259 

User interface changes?:

Update gh-pages html (Y/N)?:


jedwards4b (approved):

jasonb5 (approved):
LGTM

PR #4263, opened 2022-06-03
Fixes source command

### LINKED ISSUE #4260: "source" not available in all systems
env_mach_specific.py has this line:
`source_cmd = "source {} && ".format(init_path)`

Turns out "source" is not available in Ubuntu.  Most systems link "/bin/sh" to bash but Ubuntu links it to "dash" which is faster and does not define the "source" command.

Solution is to use "." which is POSIX-standard command for executing commands from a file.
`source_cmd = ". {} && ".format(init_path)`

Thanks to @dqwu who found this and the solution.

jasonb5:
Changes source command from `source` to `.` for `sh` shell, leaves
`source` for `csh`.

Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes #4260 
User interface changes?: N

### DIFF
--- a/CIME/XML/env_mach_specific.py
+++ b/CIME/XML/env_mach_specific.py
@@ -184,7 +184,7 @@ class EnvMachSpecific(EnvBase):
         # setup script if it exists.
         init_path = self.get_module_system_init_path("sh")
         if init_path:
-            source_cmd = "source {} && ".format(init_path)
+            source_cmd = ". {} && ".format(init_path)
         else:
             source_cmd = ""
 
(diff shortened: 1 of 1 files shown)


CIME/XML/env_mach_specific.py
```diff
@@ -231,6 +231,7 @@ def make_env_mach_specific_file(self, shell, case, output_dir=""):
         case: case object
         output_dir: string - path to output directory (if empty string, uses current directory)
         """
+        source_cmd = "." if shell == "sh" else "source"
```
rljacob, line 234: Why this if test?  "." is the standard so just use that.

jasonb5: I tried using `.` in `csh` and got the error `/usr/bin/.: Permission denied.`, so I checked the `csh` man page and only saw references to using `source` in `csh`.

jasonb5: @rljacob I tried this on my local ubuntu machine and anvil.

rljacob (approved):

PR #4279, opened 2022-07-08
Add a new system test type, REUSEINITFILES

billsacks:
This test does two runs:

(1) A standard initial run

(2) A run that reuses the init-generated files from run (1).

This verifies that it works to reuse these init-generated files, and that you can get
bit-for-bit results by doing so. This is important because these files are typically
reused whenever a user reruns an initial case.

Test suite:
- Manual testing, ensuring the test passes but fails when it should (test `REUSEINITFILES_D_Ld1_P8x1.f10_f10_mg37.I2000Clm50BgcCropQianRs.green_gnu.clm-default`)
Test baseline: n/a
Test namelist changes: none
Test status: bit for bit

Fixes none

User interface changes?: N

### DIFF
--- a/CIME/SystemTests/reuseinitfiles.py
+++ b/CIME/SystemTests/reuseinitfiles.py
@@ -1,0 +1,60 @@
+"""
+Implementation of the CIME REUSEINITFILES test
+
+This test does two runs:
+
+(1) A standard initial run
+
+(2) A run that reuses the init-generated files from run (1).
+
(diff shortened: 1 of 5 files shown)


billsacks:
My scripts_regression_tests job finished on cheyenne. All tests passed except for the unit tests that I have now pushed a fix for.

mvertens (approved):

PR #4180, opened 2022-02-11
Use of optional .exclude files for selective exclusion during build

lizziel:
As part of the coupling between CESM and GEOS-Chem, Thibaud Fritz
(@fritzt) added functionality of optional `.exclude` files to list files in the
sub-tree that do not need to be compiled. This is useful for coupling the
GEOS-Chem chemistry within CESM since some files contained in the
GEOS-Chem repository, such as the advection modules, are not used in
CESM. Using `.exclude` enables the GEOS-Chem source folder to be
unmodified from its Github repository.

The `.exclude` files are read from `mkSrcfiles` during creation of
`Srcfiles`. Any directory in `Filepath` could have a local `.exclude`
file containing filenames to exclude across its subdirectories.

An example of using `.exclude` for coupling GEOS-Chem within CESM
is here at [CAM PR #484](https://github.com/ESCOMP/CAM/pull/484/files#diff-4d5f6caee0a1d0ecb5cb75fa4346fc1182a5696f74d0cad280d2239798bbd456).

This PR supersedes PR https://github.com/ESMCI/cime/pull/3974. It is currently draft until I I get
to running the tests and updating docs as needed. Feedback welcome!

I looked at CIME documentation in doc/source/ but did not see any docs that need updating for this feature.

Test suite: **in progress**
Test baseline: 
Test namelist changes: 
Test status: [bit for bit, roundoff, climate changing] 

Fixes https://github.com/ESMCI/cime/issues/3366

User interface changes?: No

Update gh-pages html (Y/N)?: No

PR #4302, opened 2022-09-12
Expand capability to exclude files from build using .exclude

lizziel:
Updates in this PR expand the capability of using an optional `.exclude`
file to list files in the sub-tree to skip during compilation. That
feature was originally submitted and merged as https://github.com/ESMCI/cime/pull/4180 for use with
GEOS-Chem in CESM. A limitation of the feature was recently
discovered when a second `.exclude` file was added for use with HEMCO.
The content of only one of the files was excluded during build.

This PR features the following improvements in using .`exclude`:
1. Using more than one `.exclude` file in CESM is now possible.
2. Skipping one file when more than one file of the same name exists
is now possible.

It is important to note that `.exclude` must now contain the relative
path of the file to exclude. Previously only the filename
was present regardless of where it existed in the sub-tree.

Test suite: **none yet**
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #] No associated issue 

User interface changes?: N

### DIFF
--- a/CIME/Tools/mkSrcfiles
+++ b/CIME/Tools/mkSrcfiles
@@ -22,17 +22,17 @@ if ( open(FILEPATH,"< Filepath") ) {
 chomp @paths;
 unshift(@paths, '.');
 
-my $excludeDir = '';
 my $foundExclude = 0;
 foreach $dir (@paths) {  # (could check that directories exist here)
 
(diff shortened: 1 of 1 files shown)


jedwards4b (approved):
Looks good, thanks.

jgfouca (approved):

PR #4305, opened 2022-09-14
clean up paths passed to make

### LINKED ISSUE #3109: Trailing slash in casename creates unexpected behavior
If the `casename` argument to `CIME.case.create()` has a trailing slash, the method used to extract the `casename` from a path results in an empty string for `lookups['CASE']`, which I suspect is not an intended result.  In CESM2, this causes the `CASE` variable to not be set in `env_case.xml`, which prevents building the model.

This results from the use of `os.path.basename()` in [scripts/lib/CIME/case/case.py#L1561](https://github.com/ESMCI/cime/blob/b7892a96367a2f144c6a5bea2e71ca7fdd9a41c6/scripts/lib/CIME/case/case.py#L1561), which [returns an empty string if the input path has a trailing slash](https://docs.python.org/3.6/library/os.path.html#os.path.basename).

I suspect that it isn't intended that users give casenames with a trailing slash, but this might inadvertently occur when users (like me) are constructing shell scripts to configure/build/run simulations.  In my case, this resulted from me referring to my case as both an argument to `create_newcase` and as a path that I would refer to in the shell script; I habitually add trailing slashes to directory paths when writing shell scripts.

If it is the case that users should not give a trailing slash, then perhaps `case.py` could be modified to flag an error if the `casename` argument has a trailing slash?

jedwards4b:
Remove all // or trailing / before passing paths to Make arguments. 

Test suite: scripts_regression_tests.py testing by hand with script from issue
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #3109 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/build.py
+++ b/CIME/build.py
@@ -272,7 +272,11 @@ def xml_to_make_variable(case, varname, cmake=False):
         return ""
     if isinstance(varvalue, bool):
         varvalue = stringify_bool(varvalue)
-
+    elif isinstance(varvalue, str):
+        # assure that paths passed to make do not end in / or contain //
+        varvalue = varvalue.replace("//", "/")
+        if varvalue.endswith("/"):
+            varvalue = varvalue[:-1]
     if cmake or isinstance(varvalue, str):
         return '{}{}="{}" '.format("-D" if cmake else "", varname, varvalue)
     else:


jgfouca (approved):

rgknox:
I ran a quick test on this branch @jedwards4b and I no longer get the build errors I was experiencing when having double slashes before.

PR #4307, opened 2022-09-16
Updates config_batch.xml to facilitate queue specific flags

### LINKED ISSUE #4120: Easier split of case jobs into different queues
The system on which I (the user) run CESM2.1.3 (archer2) uses Slurm and has two queues: "standard" and "short", the latter being suited for small jobs shorter than 20 minutes.

1) Is there a way to configure CESM2 (presumably using config_batch.xml) so that any jobs with walltime ≤ 20 minutes are automatically submitted to the short queue, rather than the standard one (without having to manually edit the case.st_archive file)?

2) Additionally, for jobs submitted to the short queue, this system requires an extra sbatch option, "--reservation=shortqos". Again, I wondered if there's way to set up config_batch.xml so that any jobs submitted to the short queue will automatically include that option.

jasonb5:
Updates `config_batch.xml` to facilitate queue specific flags e.g. 
`<argument job_queue="long">-w docker</argument>` would only
apply when the queue name is `long`. The behavior of `argument`
is now inline with the rest of the XML system, i.e. the argument is
now stored in the XML nodes text not in attributes of the node.

Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes #4120 
User interface changes?: n/a

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -18,6 +18,7 @@ from CIME.utils import (
 from CIME.locked_files import lock_file, unlock_file
 from collections import OrderedDict
 import stat, re, math
+import pathlib
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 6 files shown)


jasonb5:
@jedwards4b @jgfouca I've written this fix to be backwards compatible, the normal `arg` node still works but I've added an `argument` node that has the new behavior and added queue selector feature. The intention was to not cause disruptions while system maintainers update the `config_batch.xml`, using the old `arg` node just prints a warning about being deprecated.

Are we fine with this behavior or should I remove the `argument` node and just update `arg` and we'll deal with it being a breaking change?

jedwards4b (approved):

PR #4315, opened 2022-09-30
PEM: fix potential invalid rootpe

jgfouca:
We had a case where rootpe was > ntasks after ntasks was halved for the second run.

Test suite: PEM by hand
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


billsacks (approved):
This seems like a good change - thank you.

jgfouca:
Note, this approach mimics what is already in ERP.

jonbob (commented):
Approved based on testing -- thanks for the quick implementation!

PR #4326, opened 2022-10-13
Add SLIM as a valid land component

### LINKED ISSUE #4191: Add SLIM as a recognized component
SLIM is a simple land model that can be run in place of CLM. In order to allow it to appear in a CESM tag alongside CLM we need to recognize it as a different component name. Initially we just need this to work in a branch off of the maint-5.6 branch as it will be in the CESM2.1 versions. We just need to bring in the version that works with the latest cime version.

### LINKED ISSUE #4292: Add "set_value" method to CIME.BuildTools.configure.FakeCase so that user can add needed variables
In principle this is a simple change, but I want to make sure it's acceptable. I'd like to add a set_value method to the FakeCase class in CIME.BuiltTools.configure so that the user can add the variables they want to have in their FakeCase. The utility I see is for unit-testing of buildnml. But, I think there may be other cases where you want to create a simpler Fake case without doing all the normal needed setup to create a real case. There are some offline scripts we are thinking about having this kind of utility where we know we need to setup some type of fake case, but we want to hide it from the user, and it doesn't need to be fully functional like a real case.

``` python
diff --git a/scripts/lib/CIME/BuildTools/configure.py b/scripts/lib/CIME/BuildTools/configure.py
index a74c087a5..fd7067ed7 100644
--- a/scripts/lib/CIME/BuildTools/configure.py
+++ b/scripts/lib/CIME/BuildTools/configure.py
@@ -74,6 +74,9 @@ def get_value(self, attrib):
         expect(attrib in self._vals, "FakeCase does not support getting value of '%s'" % attrib)
         return self._vals[attrib]
 
+    def set_value(self, attrib, value):
+       self._vals[attrib] = value
+
 def _generate_env_mach_specific(output_dir, machobj, compiler, mpilib, debug,
                                 sysos, unit_testing):
     """
```

@jedwards4b does this sound acceptable? If so I'll submit a simple PR for it.
@billsacks would this kind of thing possibly make LILAC easier to setup?

ekluzek:
Add SLIM as a valid LAND component option

Testing:  Ran a simple Slim test case: 
   SMS_Vmct_D.f19_g17.IHistSlimQianRsGs.cheyenne_intel.slim-realistic_fromCLM5_1850
   This is with a branch version of SLIM that works with the latest cime
slim hash: a3ceeff39e90dc2f2b10e75caf7d8888472aeccb
tag = ccs_config_cesm0.0.36
tag = cpl7.0.14
tag = share1.0.12
tag = MCT_2.11.0
tag = pio2_5_7

Fixes #4191 
Fixes #4292

User interface changes?:

### DIFF
--- a/CIME/BuildTools/configure.py
+++ b/CIME/BuildTools/configure.py
@@ -165,8 +165,9 @@ class FakeCase(object):
         return self._vals[attrib]
 
     def set_value(self, attrib, value):
-       """ Sets a given variable value for the case"""
-       self._vals[attrib] = value
+        """Sets a given variable value for the case"""
+        self._vals[attrib] = value
+
 
 def generate_env_mach_specific(
     output_dir,
--- a/CIME/BuildTools/configure.py
+++ b/CIME/BuildTools/configure.py
@@ -153,6 +153,10 @@ class FakeCase(object):
     def get_build_threaded(self):
         return self.get_value("SMP_PRESENT")
 
+    def get_case_root(self):
+        """Returns the root directory for this case."""
+        return self.get_value("CASEROOT")
+
     def get_value(self, attrib):
         expect(
             attrib in self._vals,
@@ -160,6 +164,9 @@ class FakeCase(object):
         )
         return self._vals[attrib]
 
+    def set_value(self, attrib, value):
+       """ Sets a given variable value for the case"""
+       self._vals[attrib] = value
 
 def generate_env_mach_specific(
     output_dir,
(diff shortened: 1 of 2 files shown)


ekluzek:
@fischer-ncar let me know if you'd like me to do more testing here.

jedwards4b (approved):

jedwards4b:
Looks like the failing tests are related to #4325 @jasonb5 can you have a look?

fischer-ncar (approved):

jedwards4b:
@ekluzek I notice that this test is using mct - what is the plan for using nuopc?  Are you adding a test to the ctsm test-suite?

ekluzek:
Good questions @jedwards4b. SLIM is a separate component and right now it's only implemented with MCT. But I'm going to start working on moving it over to NUOPC shortly.

jedwards4b:
@ekluzek can you try merging master into this branch - I think that may fix the failing docs test.

ekluzek:
@jedwards4b OK, I updated to the latest and pushed it and that seems to have resolved the failing test.

PR #4334, opened 2022-11-03
Gpu improvements

jedwards4b:
Create new variables GPU_TYPE and GPU_OFFLOAD in cime so that special -gpu compiler definitions are no longer needed. 

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/build.py
+++ b/CIME/build.py
@@ -239,6 +239,20 @@ def get_standard_cmake_args(case, sharedpath, shared_lib=False):
     cmake_args += " -Dcompile_threaded={} ".format(
         stringify_bool(case.get_build_threaded())
     )
+    gpu_type = case.get_value("GPU_TYPE")
+    gpu_offload = case.get_value("GPU_OFFLOAD")
+
+    if gpu_type != "none":
+        expect(
+            gpu_offload != "none",
+            "Both GPU_TYPE and GPU_OFFLOAD must be defined if either is",
+        )
+        cmake_args += f" -DGPU_TYPE={gpu_type} -DGPU_OFFLOAD={gpu_offload}"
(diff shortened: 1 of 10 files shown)


CIME/build.py
```diff
@@ -239,6 +239,20 @@ def get_standard_cmake_args(case, sharedpath, shared_lib=False):
     cmake_args += " -Dcompile_threaded={} ".format(
         stringify_bool(case.get_build_threaded())
     )
+    gpu_type = case.get_value("GPU_TYPE")
+    gpu_offload = case.get_value("GPU_OFFLOAD")
+
+    if gpu_type != "none":
+        expect(
+            gpu_offload != "none",
+            "Both GPU_TYPE and GPU_OFFLOAD must be defined if either is",
```
sarats, line 248: Starting a discussion from E3SM perspective:
As this mandates GPU_OFFLOAD, we probably have to add a third option to handle the case for C++ codebases that use a library approach as it won't be OpenMP or OpenACC.

CIME/case/case.py
```diff
@@ -1509,47 +1511,35 @@ def configure(
...
-                    ),
-                )
-            else:
-                expect(
-                    ngpus_per_node == 0,
-                    " ngpus_per_node is expected = 0 for compiler {}; current value is {}".format(
-                        compiler, ngpus_per_node
-                    ),
-                )
+        if gpu_type:
```
sarats, line 1520: +1 on removing this init stub for ngpus as it makes the definition explicit.

jedwards4b:
superceeded

PR #4332, opened 2022-11-01
add logic for aprun for narwhal, warhawk, onyx

apcraig:
Add support for narwhal, warhawk, and onxy for use of aprun.

This should ultimately NOT be required for porting cime, it would be nice if whatever this piece of code is doing were generalized or fixed, so modifications per machine were not needed.

A few test cases were run on narwhal, warhawk, and onyx.

Test suite:
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?: N

Update gh-pages html (Y/N)?: N

PR #4339, opened 2022-11-24
Updates aprun behavior

### LINKED ISSUE #2024: aprun sub needs to handle all options

jasonb5:
This PR update's CIME to be more flexible when using aprun. Currently
there is a specific case for the `theta` machine that uses `aprun` but 
cannot utilize CIME's aprun capabilities due to the `-j` flag not being 
supported. A new element `aprun_mode` is added to control the CIME's
behavior when a machine uses `aprun`. The `ignore` mode will skip using
CIME's aprun module and concatenate the `arg` elements, this is the
behavior for `theta` and the machines from #4332. The `default` mode 
will ignore the `arg` elements and utilize CIME's aprun module to generate
the arguments. Lastly the `override` mode will use CIME's aprun module
and mutate the output using `arg` elements.

Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes #2024 
User interface changes?: N

### DIFF
--- a/CIME/XML/env_mach_specific.py
+++ b/CIME/XML/env_mach_specific.py
@@ -581,16 +581,12 @@ class EnvMachSpecific(EnvBase):
             else None
         )
 
-    def get_mpirun(self, case, attribs, job, exe_only=False, overrides=None):
-        """
-        Find best match, return (executable, {arg_name : text})
-        """
+    def _find_best_mpirun_match(self, attribs):
         mpirun_nodes = self.get_children("mpirun")
         best_match = None
         best_num_matched = -1
         default_match = None
         best_num_matched_default = -1
-        args = []
         for mpirun_node in mpirun_nodes:
             xml_attribs = self.attrib(mpirun_node)
             all_match = True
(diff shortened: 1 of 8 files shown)


jasonb5:
TODO:
- [ ] Add new `aprun_mode` to control behavior per machine config
- [ ] Add `ignore` and `default` mode
     1. The `ignore` mode will use the default behavior for `theta` and
         concatenate all of the `arg` element values not utilizing CIME's
         aprun module.
     2. `default` will ignore `arg` elements under `mpirun` and utilize
         CIME's aprun module to generate the arguments for aprun.
- [ ] Add additional mode `override` which will use CIME's aprun module
      but utilize `arg` elements to mutate the output of the aprun function.
- [ ] Add documentation.

jasonb5:
Leaving as draft as I'm still finishing up the `override` mode.

PR #4342, opened 2022-12-07
improve float detection in argument resolution of the batch environment

### LINKED ISSUE #4341: Provided project string is altered in job submission command
We use project codes of the form `<int>_<int>`. This is altered by cime, which seemingly treats the project string as a number. For instance, a project ID such as `2022_123` gets converted to `2022123`.

All job submission commands then fail due to the wrong project ID.

This happens by setting the project ID through the option `--project` or environment variable `$PROJECT` and then passed down to the batch system as a submit argument in `config_batch.xml` as follows:
```
<batch_system MACH="hydra" type="slurm">
    <submit_args>
        <arg flag="-A" name="$PROJECT"/>
    </submit_args>
</batch_system>
```

Steps to reproduce:
1. Create new case with `--project 2022_123`
    ``` 
    $ ./create_newcase --case test --res f19_g17 --compset I2000Clm50BgcCrop --project 2022_123
    ```
3. Execute `./case.setup`
4. Execute `./preview_run` to check the resulting `SUBMIT CMD`:
    ```
    [...]
        SUBMIT CMD:
          sbatch -t 120:00:00 -p skylake_mpi -A 2022123 .case.run --resubmit
    ```

lexming:
Fixes #4341 

The underlying cause of the issue is that all arguments are converted to integers. Since a project string of the form `2022_123` is a valid number in Python syntax, it gets converted to `2022123` without going through this exception.

I suggest to replace this indiscriminate conversion of arguments into a more specific test on floats and only convert to integers those arguments that are floats.

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -675,11 +675,9 @@ class EnvBatch(EnvBase):
             else:
                 rval = val
 
-            # We don't want floating-point data
-            try:
+            # We don't want floating-point data (ignore anything else)
+            if str(rval).replace(".", "", 1).isdigit():
                 rval = int(round(float(rval)))
-            except ValueError:
-                pass
 
             # need a correction for tasks per node
             if flag == "-n" and rval <= 0:


jedwards4b (approved):

lexming:
updated to fix the failed tests

PR #4343, opened 2022-12-07
Updates aprun behavior

jasonb5:
This PR update's CIME to be more flexible when using aprun. Currently
there is a specific case for the theta machine that uses aprun but
cannot utilize CIME's aprun capabilities due to the -j flag not being
supported. A new element aprun_mode is added to control the CIME's
behavior when a machine uses aprun. The ignore mode will skip using
CIME's aprun module and concatenate the arg elements, this is the
behavior for theta and the machines from https://github.com/ESMCI/cime/pull/4332. The default mode
will ignore the arg elements and utilize CIME's aprun module to generate
the arguments. Lastly the override mode will use CIME's aprun module
and mutate the output using arg elements.

Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes https://github.com/ESMCI/cime/issues/2024
Fixes #4332
User interface changes?: N

### DIFF
--- a/doc/source/users_guide/machine.rst
+++ b/doc/source/users_guide/machine.rst
@@ -57,9 +57,35 @@ Each ``<machine>`` tag requires the following input:
   * May have optional attributes of ``compiler``, ``mpilib`` and/or ``threaded``
   * May have an optional ``<arguments>`` element which in turn contains one or more ``<arg>`` elements.
     These specify the arguments to the mpi executable and are dependent on your mpi library implementation.
-  * May have an option ``<run_exe>`` element which overrides the ``default_run_exe``
-  * May have an option ``<run_misc_suffix>`` element which overrides the ``default_run_misc_suffix``
+  * May have an optional ``<run_exe>`` element which overrides the ``default_run_exe``
+  * May have an optional ``<run_misc_suffix>`` element which overrides the ``default_run_misc_suffix``
+  * May have an optional ``<aprun_mode>`` element which controls how CIME generates arguments when ``<executable>`` contains ``aprun``. 
 
+  The ``<aprun_mode>`` element can be one of the following. The default value is ``ignore``.
+
+  * ``ignore`` will cause CIME to ignore it's aprun module and join the values found in ``<arguments>``.
+  * ``default`` will use CIME's aprun module to generate arguments.
+  * ``override`` behaves the same as ``default`` expect it will use ``<arguments>`` to mutate the generated arguments. When using this mode a ``position`` attribute can be placed on ``<arg>`` tags to specify how it's used.
+
+  The ``position`` attribute on ``<arg>`` can take one of the following values. The default value is ``per``.
+
+  * ``global`` causes the value of the ``<arg>`` element to be used as a global argument for ``aprun``.
+  * ``per`` causes the value of the ``<arg>`` element to be appended to each separate binaries arguments.
+
+  Example using ``override``:
+  ::
+
+    <executable>aprun</executable>
+    <aprun_mode>override</aprun_mode>
+    <arguments>
+      <arg position="global">-e DEBUG=true</arg>
+      <arg>-j 20</arg>
+    </arguments>
+
+  Sample command output:
+  ::
+
+    aprun -e DEBUG=true ... -j 20 e3sm.exe : ... -j 20 e3sm.exe
 
 * ``module_system``: How and what modules to load on this system. Module systems allow you to easily load multiple compiler environments on a machine. CIME provides support for two types of module tools: `module <http://www.tacc.utexas.edu/tacc-projects/mclay/lmod>`_ and `soft  <http://www.mcs.anl.gov/hs/software/systems/softenv/softenv-intro.html>`_. If neither of these is available on your machine, simply set ``<module_system type="none"\>``.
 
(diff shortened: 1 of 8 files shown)


jasonb5:
CIME's aprun module utilizes the PE layout to support aprun's MPMD (Multiple Program Multiple Data) mode. The override mode will use attributes on the arg elements to control which program has their arguments mutated.

TODO:

Add new aprun_mode to control behavior per machine config
Add ignore and default mode

    The ignore mode will use the default behavior for theta and
    concatenate all of the arg element values not utilizing CIME's
    aprun module.
    default will ignore arg elements under mpirun and utilize
    CIME's aprun module to generate the arguments for aprun.

Add additional mode override which will use CIME's aprun module
but utilize arg elements to mutate the output of the aprun function.
Add documentation.

jgfouca (approved):
Looks good!

jasonb5:
@jedbrown @jgfouca The new default behavior is to skip using the aprun module, which is what we did for `theta` and the machines from #4332. Are we fine with this being the default, or do we want machines to automatically use the aprun module when `<executable>` is set to aprun, and explicitly disable it (this would require adding `<aprun_mode>ignore</aprun_mode>` to the config_machine.xml for theta, etc)?

jasonb5:
You can find the documentation changes [here](https://esmci.github.io/cime/branch/fixes_aprun/html/users_guide/machine.html#config-machines-xml-machine-specific-file)

jgfouca:
@jedwards4b will have to comment on that. I don't think we (E3SM) have any actively-used aprun machines.

jedwards4b:
I think that we would want the default behavior to be default wouldn't we?

doc/source/users_guide/machine.rst
```diff
@@ -57,9 +57,35 @@ Each ``<machine>`` tag requires the following input:
...
-  * May have an option ``<run_misc_suffix>`` element which overrides the ``default_run_misc_suffix``
+  * May have an optional ``<run_exe>`` element which overrides the ``default_run_exe``
+  * May have an optional ``<run_misc_suffix>`` element which overrides the ``default_run_misc_suffix``
+  * May have an optional ``<aprun_mode>`` element which controls how CIME generates arguments when ``<executable>`` contains ``aprun``. 
 
+  The ``<aprun_mode>`` element can be one of the following. The default value is ``ignore``.
+
+  * ``ignore`` will cause CIME to ignore it's aprun module and join the values found in ``<arguments>``.
+  * ``default`` will use CIME's aprun module to generate arguments.
+  * ``override`` behaves the same as ``default`` expect it will use ``<arguments>`` to mutate the generated arguments. When using this mode a ``position`` attribute can be placed on ``<arg>`` tags to specify how it's used.
```
jedwards4b, line 68: 'expect' should be "except"     Shouldn't the default be "default"?

jasonb5:
@jedwards4b That makes sense. I know there aren't a lot of machines using aprun so I was trying to keep backward compatibility. For `theta` we'll need to add `<aprun_mode>ignore</aprun_mode>` to keep the current behavior and the machines from #4332 will need to do the same.

jedwards4b:
I would hope that this allows #4332 to be redone in a better way.

jedwards4b:
@apcraig do you want to try to redo #4332 with the new aprun functionality introduced here and give us feedback?

jasonb5:
The override mode should allow support for those machines, the aprun module should generate most of the arguments and any additional can be passed through `<arguments>`. 

If there's a standard way to derive the `-j` argument I can add that to the aprun module's functionality.

apcraig:
I can try to test this.  Do you think I should just merge this PR into my sandbox (associated with #4332).  Or should I update CIME to the latest version and then merge this PR into that sandbox?

jasonb5:
@apcraig You could just merge this into your sandbox, it doesn't depend on anything specific.

apcraig:
I pulled these changes into my sandbox and removed my modifications to aprun, created a new case, built and submitted.  The case ran fine.  I will test on a couple other platforms.  I did not change anything in my xml files.  Is the default that it should run fine or is the default going to change?  Regardless, should I be explicitly setting the "aprun_mode" value in my xml file for my machines just to be certain?

jedwards4b:
I think that it would be wise to explicitly set the aprun mode.  Thanks for testing.

apcraig:
Where does this belong in the xml file, what is the correct setting?  Can you provide the exact syntax.  Thanks.

jedwards4b:
<aprun_mode>override</aprun_mode>

Goes in config_machines.xml:
   ```
<mpirun mpilib="openmpi">
    <aprun_mode>override</aprun_mode>
    <executable>aprun</executable>
```
All documented in this PR.

apcraig:
OK, so that all seems to work OK.  Thanks.  Please close #4332 when this is merged.  Or I will.

jedwards4b (approved):

jasonb5:
@jedwards4b I'm going to change the default value to `default` so for `theta` you'll need to set `aprun_mode` to `ignore` or `override` and pass `-j` through `arguments`.

jedwards4b:
Sounds good - I know we have theta in our machine definitions but I don't think anyone is using cesm there at this time.

PR #4345, opened 2022-12-15
correct the rpointer name for nuopc cases

### LINKED ISSUE #4346: NUOPC restarts broken in cime6.0.64 - 6.0.80
This was missed because the code in question is explicitly not being tested.  To clarify, restarts work in 
testing but not in a regular case.  A workaround is to set MULTI_DRIVER=FALSE.

jedwards4b:
Restarts are broken for nuopc cases since tag cime6.0.64 due to code which is explicitly not being tested. 
I have removed the conditional that causes this code block not to be tested so that this doesn't happen again - 
but I'm not sure why that block was there in the first place - can we be more specific about tests that should not use this code block?

Test suite:
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #4346 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/case/case_submit.py
+++ b/CIME/case/case_submit.py
@@ -52,7 +52,12 @@ def _submit(
     # Check if CONTINUE_RUN value makes sense
     # if submitted with a prereq don't do this check
     testcase = case.get_value("TESTCASE")
-    if (not testcase or testcase not in ("PET", "NCK")) and case.get_value("CONTINUE_RUN") and hasMediator and not prereq:
+    if (
+        (not testcase or testcase not in ("PET", "NCK"))
+        and case.get_value("CONTINUE_RUN")
+        and hasMediator
+        and not prereq
+    ):
         rundir = case.get_value("RUNDIR")
         expect(
             os.path.isdir(rundir),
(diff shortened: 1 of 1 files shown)


mvertens:
@jedwards4b - what restarts are broken for nuopc? It cannot be all restarts since we do a lot of testing. Can you please clarify this.

jedwards4b:
All restarts except those with NINST > 1.   We missed it because [this code block ](https://github.com/ESMCI/cime/blob/master/CIME/case/case_submit.py#L55)is explicitly not tested, I want to understand why not and fix it in the PR.

fischer-ncar (approved):

CIME/case/case_submit.py
```diff
@@ -51,8 +51,9 @@ def _submit(
 
     # Check if CONTINUE_RUN value makes sense
     # if submitted with a prereq don't do this check
+    testcase = case.get_value("TESTCASE")
     if (
-        job != "case.test"
+        (not testcase or testcase not in ("PET", "NCK"))
```
billsacks, line 56: Do you know why these test cases are exceptions? If so, a comment here would be helpful for future maintenance.

jedwards4b:
I ended up combining these fixes in #4348, closing this and using that one.

PR #4348, opened 2022-12-16
error out if run length is shorter than longest coupling interval

jedwards4b:
For Nuopc we need to make sure that the run length exceeds the longest component coupling interval to avoid a runtime error with a cryptic error message. 

Test baseline:
Test namelist changes:
Test status: bit for bit
Fixes https://github.com/ESCOMP/CMEPS/issues/328

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/case/case_submit.py
+++ b/CIME/case/case_submit.py
@@ -8,7 +8,7 @@ submit, check_case and check_da_settings are members of class Case in file case.
 """
 import configparser
 from CIME.XML.standard_module_setup import *
-from CIME.utils import expect, run_and_log_case_status, CIMEError
+from CIME.utils import expect, run_and_log_case_status, CIMEError, get_time_in_seconds
 from CIME.locked_files import unlock_file, lock_file
 from CIME.test_status import *
 
@@ -296,6 +296,37 @@ def check_case(self, skip_pnl=False, chksum=False):
     if self.get_value("COMP_WAV") == "ww":
         # the ww3 buildnml has dependencies on inputdata so we must run it again
         self.create_namelists(component="WAV")
+
+    if self._comp_interface == "nuopc":
+        # Check that run length exceeds longest component coupling interval
+        # longest interval is smallest NCPL value
+        maxncpl = 10000
+        maxcomp = None
+        for comp in self.get_values("COMP_CLASSES"):
+            if comp == "CPL":
+                continue
+            ncpl = self.get_value("{}_NCPL".format(comp))
+            if ncpl and maxncpl > ncpl:
+                maxncpl = ncpl
+                maxcomp = comp
+        ncpl_base_period = self.get_value("NCPL_BASE_PERIOD")
+        if ncpl_base_period == "hour":
+            coupling_secs = 3600 / maxncpl
+        elif ncpl_base_period == "day":
+            coupling_secs = 86400 / maxncpl
+        elif ncpl_base_period == "year":
+            coupling_secs = 3.154e7 / maxncpl
+        elif ncpl_base_period == "decade":
+            coupling_secs = 3.154e8 / maxncpl
+        runtime = get_time_in_seconds(
+            self.get_value("STOP_N"), self.get_value("STOP_OPTION")
+        )
+        expect(
+            runtime > coupling_secs,
+            " Runtime ({} s) is shorter than the longest component coupling interval ({} s), increase runtime or increase {}_NCPL and try again.  ".format(
+                runtime, coupling_secs, maxcomp
+            ),
+        )
 
     expect(
         self.get_value("BUILD_COMPLETE"),


ekluzek (approved):
This is a nice addition for usability. This looks good to me.

cacraigucar (approved):
As long as the error message is writing out seconds (which I believe it is) for both places, I approve this.

rljacob:
Is this check done for all CIME-coupled models?   I hope so.

jedwards4b:
No, it's only done for nuopc cases because they will fail at initialization with a cryptic error message.  
Do you want me to make it independent of comp_interface?

rljacob:
Yes please.

rljacob:
Hang on.  I just realized some of our tests are actually doing this.  They run for a few atmosphere time steps when the coupling interval with river is a few hours.  So I guess this should be done just for nuopc.

jedwards4b:
Right - the mct dycore will only error out later if allowed to run long enough so we do those short tests without the correction, okay - I'll put the nuopc restriction back.

CIME/case/case_submit.py
```diff
@@ -297,6 +293,54 @@ def check_case(self, skip_pnl=False, chksum=False):
...
+            timestep = 3600 / minncpl
+        elif ncpl_base_period == "day":
+            coupling_secs = 86400 / maxncpl
+            timestep = 86400 / minncpl
+        elif ncpl_base_period == "year":
+            coupling_secs = 3.154e7 / maxncpl
+            timestep = 3.154e7 / minncpl
+        elif ncpl_base_period == "decade":
+            coupling_secs = 3.154e8 / maxncpl
+            timestep = 3.154e8 / minncpl
```
billsacks, line 329: Is this going to work robustly given that you're using floats rather than ints? I wonder if you should change this to either use ints or round the result to the nearest int???

CIME/case/case_submit.py
```diff
@@ -51,12 +51,7 @@ def _submit(
 
     # Check if CONTINUE_RUN value makes sense
     # if submitted with a prereq don't do this check
-    if (
-        job != "case.test"
-        and case.get_value("CONTINUE_RUN")
-        and hasMediator
-        and not prereq
-    ):
+    if case.get_value("CONTINUE_RUN") and hasMediator and not prereq:
```
billsacks, line 54: Was it intentional that you built this on top of the changes in #4345? It looks like your fix in #4345 has evolved from this, so this may cause problems with the merge.

billsacks (commented):
Thanks a lot for putting this in, @jedwards4b ! I have a couple of questions below:

CIME/case/case_submit.py
billsacks: Oh, I see you just closed #4345 . So then I guess the question becomes which one you want to use: this one that applies this code to all test cases, or what you had in #4345, which had exceptions for PET and NCK. If the latter, can you add a comment about why you need an exception for those test types (if you know)?

jedwards4b: I thought that an exception was needed then retested and didn't need an exception, currently I have removed all exceptions and would like to pass it off to someone else to test independently.

CIME/case/case_submit.py
jedwards4b: These values are not being used for anything other than this message so I don't thing it matters if they are real or integer.

CIME/case/case_submit.py
billsacks: Sounds good, thanks.

CIME/case/case_submit.py
billsacks: They're used below in `runtime % coupling_secs`. I just tested this and I think it would give a problem currently for ncpl_base_period == "year" or ncpl_base_period == "decade" - e.g.:

```python
maxncpl = 365
runtime = 86400
coupling_secs = 3.154e7 / maxncpl
runtime % coupling_secs
```

gives 86400.0.

Whereas this (corrected):

```python
maxncpl = 365
runtime = 86400
coupling_secs = 31536000 / maxncpl
runtime % coupling_secs
```

gives 0.0, correctly.

billsacks:
@jedwards4b - I was a bit uncomfortable with the trial-and-error changes we made to get the unit tests passing, so I dug into this further to try to understand why we needed these changes and convince myself that these changes seem right. My conclusions are that these changes to the unit test mostly seem reasonable / expected, but:

(1) I'm thinking that, if we want the unit tests to most accurately reflect what would happen in reality, then we should change line 146 of the unit test file to return something like "$ENV{TEST}" and then change `my_get_resolved_value` so that if it sees "$ENV{TEST}" it returns "/test". I'm not positive about this, though: This arises from calling case.get_value on `$$ENV{TEST}`; in this case would you expect the return result to be `$ENV{TEST}` and then this would be fed into case.get_resolved_value? If so, I think my suggestion is right. (The unit test passes either way, but it could be best to keep the unit test most closely reflecting what would happen with a real case.)

(2) From the extra prints I added, I can't tell if the production code is working as you intend. Maybe it is... it seems a little funny that we're feeding a lot of fully-resolved return values into case.get_resolved_value, but maybe that's okay.

Specifically, I ran the unit tests with these diffs:

```diff
diff --git a/CIME/XML/env_batch.py b/CIME/XML/env_batch.py
index dd6bf3bea..73d330690 100644
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -654,6 +654,8 @@ def _get_argument(self, case, arg):
 
     def _resolve_argument(self, case, flag, name, job):
         submitargs = ""
+        print("_resolve_argument: name = {}".format(name))
+
         logger.debug("name is {}".format(name))
         # if name.startswith("$"):
         #    name = name[1:]
@@ -666,6 +668,7 @@ def _resolve_argument(self, case, flag, name, job):
                 if part != "":
                     logger.debug("part is {}".format(part))
                     resolved = case.get_value(part, subgroup=job)
+                    print("With $ in name, new loop: get_value: {}".format(resolved))
                     if resolved:
                         name += resolved
                     else:
@@ -674,8 +677,10 @@ def _resolve_argument(self, case, flag, name, job):
             # We have a complex expression and must rely on get_resolved_value.
             # Hopefully, none of the values require subgroup
             val = case.get_resolved_value(name)
+            print("With $ in name: get_resolved_value: {}".format(val))
         else:
             val = case.get_value(name, subgroup=job)
+            print("Without $ in name: get_value: {}".format(val))
 
         if val is not None and len(str(val)) > 0 and val != "None":
             # Try to evaluate val if it contains any whitespace
diff --git a/CIME/tests/test_unit_xml_env_batch.py b/CIME/tests/test_unit_xml_env_batch.py
index 0086b3308..0e5c74c09 100755
--- a/CIME/tests/test_unit_xml_env_batch.py
+++ b/CIME/tests/test_unit_xml_env_batch.py
@@ -62,6 +62,7 @@ def test_get_submit_args_job_queue(self):
 
             case.filename = mock.PropertyMock(return_value=tfile.name)
 
+            print("In test_get_submit_args_job_queue:")
             submit_args = batch.get_submit_args(case, ".case.run")
 
             expected_args = "  -w default -w long"
@@ -154,6 +155,7 @@ def my_get_resolved_value(val):
 
             case.filename = mock.PropertyMock(return_value=tfile.name)
 
+            print("In test_get_submit_args:")
             submit_args = batch.get_submit_args(case, ".case.run")
 
             expected_args = "  --time 00:30:00 -p long --account CIME --no-arg --path /test -w docker"
```

Giving this output:

```
In test_get_submit_args:
_resolve_argument: name = $JOB_WALLCLOCK_TIME
With $ in name, new loop: get_value: 00:30:00
With $ in name: get_resolved_value: 00:30:00
_resolve_argument: name = $JOB_QUEUE
With $ in name, new loop: get_value: long
With $ in name: get_resolved_value: long
_resolve_argument: name = $PROJECT
With $ in name, new loop: get_value: CIME
With $ in name: get_resolved_value: CIME
_resolve_argument: name = $$ENV{TEST}
With $ in name, new loop: get_value: /test
With $ in name: get_resolved_value: /test

In test_get_submit_args_job_queue:
_resolve_argument: name = $VARIABLE_THAT_DOES_NOT_EXIST
With $ in name, new loop: get_value: None
With $ in name: get_resolved_value: None
```

jedwards4b:
I was thinking that that last call to get_resolved_value is no longer needed.   @billsacks Thank you so much for your help on this.

jedwards4b:
A quick check confirms that that final get_resolved_value is needed for the case where name = "-A $VARIABLE_THAT_DOES_NOT_EXIST" and the result should be "".

PR #4349, opened 2022-12-16
Check number of hist file comparisons in system_test_common

jgfouca:
Every file that gets copied by copy_histfiles should later be compared by compare_test. We ran into a situation where the hist_file_extension for a component worked when suffix="" but not when suffix!="". This caused hist files for that component to get silently skipped in various system tests that do internal compares.

This PR will make the scenario above be reported as a failed comparison so it can't go unnoticed.

Test suite: script_regression_tests
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/SystemTests/system_tests_common.py
+++ b/CIME/SystemTests/system_tests_common.py
@@ -59,6 +59,7 @@ class SystemTestsCommon(object):
         self._ninja = False
         self._dry_run = False
         self._user_separate_builds = False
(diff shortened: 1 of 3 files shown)


jedwards4b (approved):

jasonb5 (approved):
LGTM

PR #4358, opened 2023-01-24
Switch file compare count mismatch to a warning

jgfouca:
... instead of an error.

Too many E3SM components have problems in their config_archive.xml settings.

Test suite: by-hand.
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jedwards4b (approved):
I was just looking at this problem in CESM.

PR #4360, opened 2023-02-10
Improve cmake build system for sharedlibs

jgfouca:
Sending COMP_NAMES to the macros allow for checking for the presence of certain components which may influence how the sharedlibs are build.

Also, the shared_lib flag to get_standard_cmake_args goes away since the cmake_args are now the same either way.

Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jgfouca:
@sarats

jedwards4b (approved):

jedwards4b:
@jasonb5 The cdeps repository has been updated and this test should now pass, but I don't see how the cdeps repo is checked out in testing.yml.  I just pushed a change with the cdeps tag updated in Externals.cfg but I'm not sure that that's what needs to happen.

jasonb5 (approved):

PR #4372, opened 2023-03-16
Adding support for ADIOS for E3SM

jayeshkrishna:
Adding support for the ADIOS I/O library for E3SM.

* The build script for SCORPIO (spio) is available in E3SM
* All I/O post processing logic/scripts for ADIOS is available in E3SM

Test suite: scripts_regression
Test baseline: None
Test namelist changes: None
Test status: bit for bit

User interface changes: None


jayeshkrishna:
@jedwards4b, this PR separates the build script for SCORPIO (buildlib.spio) from PIO (buildlib.pio). So these changes should not affect CESM+PIO builds.

CIME/build_scripts/buildlib.spio
```diff
@@ -0,0 +1,267 @@
+#!/usr/bin/env python3
```
jedwards4b, line 1: Using config_files you can relocate this file to a path within the scorpio repository, that way you can make further changes without the need for a cime PR - I suggest that you move this file.

CIME/case/case_post_run_io.py
```diff
@@ -0,0 +1,113 @@
+"""
```
jedwards4b, line 1: I think that this file can also be located in scorpio and it makes more sense to have it there.   I have a couple of examples of how to do this if you are interested.

jedwards4b (changes requested):

jayeshkrishna:
Note that this change requires recent versions of E3SM master for testing with E3SM (latest E3SM master - https://github.com/E3SM-Project/E3SM/commit/8266141d31bd661f9d5625b545669b43e4e155e4 - should have all the required changes). Also, the I/O post processing is not turned on and requires some further changes (https://github.com/E3SM-Project/E3SM/tree/jayeshkrishna/cime_config/add_adios_conv_job) in subsequent E3SM PRs.

CIME/case/case_post_run_io.py
jayeshkrishna: Sure, please let me know.

CIME/build_scripts/buildlib.spio
jayeshkrishna: Looks like @jgfouca also agrees with you about the location of the file, let me move the file and then you can review the PR.

CIME/case/case_post_run_io.py
jgfouca: @jedwards4b said it could go in scorpio. I'm not sure how to make that happen such that python knows to search that directory for imports. My thought would be to put it in `E3SM/cime_config/customize`

jedwards4b: I must admit that I would need to experiment a bit to come up with a functional solution.

jayeshkrishna: ok, I will try moving the file to $E3SM_SRC/cime_config/customize for now

CIME/build_scripts/buildlib.spio
jayeshkrishna: buildlib.spio is now moved out of CIME to $E3SM_SRC/share/build

CIME/case/case_post_run_io.py
jayeshkrishna: case_post_run_io.py has been moved out of CIME to $E3SM_SRC/cime_config/customize

jayeshkrishna:
As discussed above, E3SM specific scripts/logic has now been moved to E3SM.

This PR requires https://github.com/E3SM-Project/E3SM/tree/jayeshkrishna/cime_config/add_adios_conv_job for testing with E3SM.

jgfouca (approved):

jgfouca:
It looks like this PR won't pass AT until the E3SM PR is merged.

jayeshkrishna:
This PR should now (E3SM-Project/E3SM#5530) work with E3SM master

jgfouca:
@jayeshkrishna , yep! I forced a retest and it appears to have passed. Merging.

PR #4380, opened 2023-03-27
should not test timestep of stub components

jedwards4b:
This test was over conservative giving an error in aquaplanet cases for ROF_NCPL, since
ROF is not a part of an aquaplanet case we don't need this check.

Test suite:
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/case/case_submit.py
+++ b/CIME/case/case_submit.py
@@ -307,6 +307,10 @@ def check_case(self, skip_pnl=False, chksum=False):
         for comp in self.get_values("COMP_CLASSES"):
             if comp == "CPL":
                 continue
+            compname = self.get_value("COMP_{}".format(comp))
+            # ignore stub components in this test.
+            if compname.startswith("S"):
+                continue
             ncpl = self.get_value("{}_NCPL".format(comp))
             if ncpl and maxncpl > ncpl:
                 maxncpl = ncpl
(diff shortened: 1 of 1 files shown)


fischer-ncar (approved):

CIME/case/case_submit.py
```diff
@@ -307,6 +307,10 @@ def check_case(self, skip_pnl=False, chksum=False):
         for comp in self.get_values("COMP_CLASSES"):
             if comp == "CPL":
                 continue
+            compname = self.get_value("COMP_{}".format(comp))
+            # ignore stub components in this test.
+            if compname.startswith("S"):
```
rljacob: I'm concerned this might be too general but not sure how else to identify stub models in case_submit.

jedwards4b: We have a convention that if a component name begins with 's' it is a stub component.
This is an assumption already used in many places in cime.

jedwards4b: Actually I could refine this a little: 
if compname == "s{comp}".format(comp)

I'll do that.

fischer-ncar (approved):

rljacob (approved):

PR #4384, opened 2023-04-04
check_input_data: Expand SharedArea coverage to entire implementation

jgfouca:
Users are reporting directories in the input file area that have locked-down permissions. I did see several os.makedirs calls in check_input_data.py that fell outside SharedArea contexts. I think the easiest thing is to just do the entire implementation in a SharedArea.

Test suite: by-hand, pylint
Test baseline:
Test namelist changes:
Test status: bfb

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/case/check_input_data.py
+++ b/CIME/case/check_input_data.py
@@ -66,25 +66,24 @@ def _download_checksum_file(rundir):
             os.rename(full_path, tmpfile)
         # Use umask to make sure files are group read/writable. As long as parent directories
         # have +s, then everything should work.
-        with SharedArea():
-            success = server.getfile(rel_path, new_file)
-            if success:
-                _reformat_chksum_file(full_path, new_file)
-                if tmpfile:
-                    _merge_chksum_files(full_path, tmpfile)
-                chksum_hash.clear()
+        success = server.getfile(rel_path, new_file)
+        if success:
+            _reformat_chksum_file(full_path, new_file)
+            if tmpfile:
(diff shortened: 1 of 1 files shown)


jgfouca:
@jonbob ^^^

jedwards4b (approved):
I don't have any problem with this change but I don't think that it will necessarily solve issues with permissions in inputdata.   For example on perlmutter group write permissions are set but many cesm and e3sm users are not in the group, and I don't think that we want everyone to have write permission to this directory tree.

jgfouca:
@jedwards4b , yeah the root of the input dir will require the correct group ownership and chmod +s.

jedwards4b:
I guess that what I'm saying is that I don't think that we can get to a place where any user can write to inputdata without intervention from someone with more privilege.   One way that we could workaround this is to create a temporary inputdata directory in the users space, linking files that are in the default inputdata location and downloading files which are not.   I was working on something like this in another context - on TACC systems access to the $WORK dir can be quite slow from compute nodes so I created a method to stage inputdata into the $SCRATCH dir prior to running.  We lost our TACC project though and I didn't bring this branch change into master - I think I can find it though if you are interested.

jgfouca:
@jedwards4b , i think it may actually be this PR. I'm seeing similar failures when I run cime_tiny by hand that go away when I remove my changes.

jgfouca:
@jedwards4b , I forgot to do the returns in the wrappers 😆

Issue #3690, opened 2020-08-26
Faster way to generate multi-instance modelio namelists


kdraeder:
The ${component}\_modelio.nml\_${INST} namelists are written serially on 1 task, 
while all the rest wait. For 80 instances (720 files) that takes 40 seconds, 
while in the data assimilation context the 6 hour hindcast takes ~3 minutes 
and the data assimilation another ~3 minutes.
The only thing that's changing in those files is the name of the log file,
so now we just look for an existing modelio namelist file and change
the log file name in it.  Now it takes 0.6 seconds for all 720.
A faster version of buildnml can be found in
/glade/work/raeder/Models/cesm2_1_relsd_m5.6/cime/src/drivers/mct/cime_config
and I've attached it here.
[buildnml_faster_modelio.txt](https://github.com/ESMCI/cime/files/5127182/buildnml_faster_modelio.txt)

PR #4391, opened 2023-04-08
Cesm dart integ; modelio.nml creation

### LINKED ISSUE #4383: Slow modelio namelist generation when NINST is large
Data assimilation (DA) jobs require 20 < NINST < 1000
and typically run multiple cycles (calls to ESP) in each job.
Each cycle requires a new run of CESM, which prompts calls to preview_namelist
and the generation of namelists for all of the instances of all of the components,
serially (while a large number of nodes needed for the job sit idle).
The namelists are either identical from cycle to cycle, 
or have minor changes such as time stamps in file names.
The ${component}\_in\_${inst} creation can be turned off by 
   `case.submit --skip-preview-namelist`
but the modelio namelists can't.

We've modified 
   `src/drivers/mct/cime_config/buildnml`
to check for the existence of a modelio namelist.
If it exists, merely replace some of the contents with new data,
otherwise, create the namelist as usual.
It reduced the time required to write the modelio files for 80 instances
from 40 s to 0.6 s, out of a total cycle time of \~400 s.
I've attached a modified version (with .txt added so that github will accept it), 
from a cime branch from origin/maint-5.6  (\~ Mar 12 05:20:53 2019)
170db73e00130294bac98fde5dbbcdcde06a33d6

> \> git describe --tags
>     cime5.6.11-87-g0a8bce9f1
> 

Tell me if I need to import this into a more recent CIME in order
for it to be reasonable to work with.
I understand that this will need to be implemented in NUOPC/CMEPS.

This may not be the solution that's optimal for CIME as a whole, 
but it was sufficient for our use and has a small footprint.
I'll submit it as a pull request, if it looks like the right strategy.
[buildnml.txt](https://github.com/ESMCI/cime/files/11114117/buildnml.txt)

kdraeder:
Commit 746f5ac71f4e7e gives the context for this upgrade.
The upgrade accelerates the creation of modelio.nml files when NINST is large
by replacing the log file names in existing modelio.nml files, if they exist,
instead of recreating the files from scratch.

I created a draft PR because the commit list in this comparison 
shows commits that were made in 2018 to the branch 
which is the source of this branch (cesm_dart_integ).
The 2018 commits were either incorporated in maint-5.6 or reverted,
so the net changes in this PR are all in one file and fix one issue.
Do I need to clean that up before making this an actual PR?

Test suite:
I have tested it in the latest maint-5.6 CIME 
(cime_cesm2_1_rel_02-478-gab29a7a32) in cesm2_2_alpha01c-4-g6ebfe0b.
The case uses NINST=3 and the multidriver feature.
As intended, when the case is built, the modelio.nml files are created 
from scratch in Buildconf/cplconf.  When the job is submitted,
those files are modified to contain the new log file names
and then copied to $RUNDIR as usual.
Test baseline: 
not a standard test
Test namelist changes:
No external changes.  The modelio namelists are created by a modified buildnml.
Bit for bit.  

Fixes  #4383

User interface changes?:
None

Update gh-pages html (Y/N)?:
Not needed


kdraeder:
I don't have much context for understanding these test failures.
My first suspicion would be that they're related to this test using maint-5.6 with cesm2_2_alpha01c
instead of a cesm2_1.  But the case built and the test ran, so I'm a little skeptical of that.

Do they need to be resolved before making this an actual PR?

kdraeder:
@jedwards4b 
I haven't made a draft PR before, so I don't know whether reviewers (aka you) 
get notifications about comments.

jedwards4b (approved):

kdraeder:
Also, this PR would justify closing #3690.
The final entry in that issue is that @billsacks moved this issue to Todo ~ Longer term
in https://github.com/orgs/ESCOMP/projects/2

jedwards4b:
@kdraeder https://github.com/ESMCI/cime/pull/4395 fixes the github workflow issues that you are encountering - once it has been reviewed and merged you should be able to update your PR and see the tests pass.

jedwards4b:
@kdraeder  Testing indicates that this PR is causing the xml values not to be set in the modelio namelists and
all values are set to defaults.  This is causing several tests to fail and so we had to back this PR out of the 2.1.4 release candidate.   I take responsibility for not finding this issue earlier.   I think an alternate fix may be just to move the default value assignment lines out of the inst loop.   This reduces the time to create modelio files for 80 members from 6s to < 1s.
```diff --git a/src/drivers/mct/cime_config/buildnml b/src/drivers/mct/cime_config/buildnml
index aecd915cc..83dffa514 100755
--- a/src/drivers/mct/cime_config/buildnml
+++ b/src/drivers/mct/cime_config/buildnml
@@ -321,22 +321,22 @@ def _create_component_modelio_namelists(case, files):
 
             inst_string = ""
             inst_index = 1
+            # set default values
+            for entry in entries:
+                nmlgen.add_default(entry)
+            # overwrite defaults
+            moddiri = case.get_value('EXEROOT') + "/" + model
+            nmlgen.set_value('diri', moddiri)
+
+            moddiro = case.get_value('RUNDIR')
+            nmlgen.set_value('diro', moddiro)
+
             while inst_index <= inst_count:
+
                 # determine instance string
                 if inst_count > 1:
                     inst_string = '_{:04d}'.format(inst_index)
 
-                # set default values
-                for entry in entries:
-                    nmlgen.add_default(entry)
-
-                # overwrite defaults
-                moddiri = case.get_value('EXEROOT') + "/" + model
-                nmlgen.set_value('diri', moddiri)
-
-                moddiro = case.get_value('RUNDIR')
-                nmlgen.set_value('diro', moddiro)
-
                 logfile = model + inst_string + ".log." + str(lid)
                 nmlgen.set_value('logfile', logfile)
```

PR #4396, opened 2023-04-17
Add flexible controls of GPU configuration

sjsprecious:
This PR updates the configurations for the GPU offload based on the PRs in CMEPS (https://github.com/ESCOMP/CMEPS/pull/363) and ccs_config_cesm (https://github.com/ESMCI/ccs_config_cesm/pull/97).

An example to build a GPU case on Derecho with these new GPU options by using **create_newcase** looks:
`./create_newcase --case /path_to_case_dir --mach derecho --compiler nvhpc --mpilib mpich --compset F2000dev --res f19_f19_mg17 --queue main --ngpus-per-node 4 --gpu-type a100 --gpu-offload openacc`

An example to build a GPU case on Derecho with these new GPU options by using **create_test** looks:
`./create_test ERP_Ln9_G4-a100-openacc.f19_f19_mg17.F2000dev.gust_nvhpc.cam-outfrq9s --test-root /path_to_case_dir --output-root /path_to_output -q main`


sjsprecious:
@jedwards4b will issue a separate PR to introduce these changes.

CIME/config.py
```diff
@@ -180,7 +180,7 @@ def __init__(self):
         self._set_attribute(
             "gpus_use_set_device_rank",
             True,
-            desc="If set to `True` and NGPUS_PER_NODE > 0 then `$RUNDIR/set_device_rank.sh` is appended when the MPI run command is generated.",
+            desc="If set to `True` and NGPUS_PER_NODE > 0 then `$RUNDIR/set_device_rank.sh` or `get_local_rank` (a global script on Derecho/Gust) is appended when the MPI run command is generated.",
```
jedwards4b, line 183: Is there a more general way to do this?  One that would apply to any system and generic script name?

sjsprecious: Maybe we could copy the `get_local_rank` script on Derecho and save it under the `ccs_config_cesm/machines` folder?

jedwards4b: I was thinking more along the lines of a new xml variable: MPIEXEC_HELPER_SCRIPT=
or something like that.

sjsprecious: Oh I see. So add `MPIEXEC_HELPER_SCRIPT` in CMEPS and use it in CIME like what we did for `GPU_TYPE`?

sjsprecious: @jedwards4b I just added a new XML variable named `MPI_GPU_WRAPPER_SCRIPT` in https://github.com/ESMCI/ccs_config_cesm/pull/97 and https://github.com/ESCOMP/CMEPS/pull/363, and update the settings in CIME here as well. I just built a case on Derecho and the `preview_run` indicated that the `get_local_rank` script was attached to the MPI command line as expected. See the output below:
 ```
   MPIRUN (job=case.run):
      mpiexec  --label  --line-buffer  -n 64 get_local_rank /glade/derecho/scratch/sunjian/F2000dev.f19_f19_mg17.derecho.nvhpc.pumas/bld/cesm.exe   >> cesm.log.$LID 2>&1 
```

jedwards4b (approved):

rljacob:
I believe this works independently of the E3SM approach to define gpu-specific compilers?   And I think we would handle a machine like derecho with separate machine entries for each node type.

jedwards4b:
that's correct - we are departing from that approach.

jedwards4b:
@sjsprecious It looks like you may need to add something to make sure this only happens when driver=nuopc

sjsprecious:
Thanks @jedwards4b . Do you mean that we should only use these new GPU options for the NUOPC driver? Why can't we use it for MCT?

jedwards4b:
If you want use them for MCT then you will need to make changes in 
components/cpl7/driver/cime_config/config_component.xml
similar to what you did in cmeps.

sjsprecious:
Thanks @jedwards4b for your explanation.

So what is the best way you think to handle the MCT case? Currently generating a case with MCT will fail due to the missing new XML variables as expected. Would you suggest bypassing those errors and continuing to build anyway (probably ignoring the given GPU options and always building a CPU case)?

jedwards4b:
probably the easiest is to just add them to the mct config_component.xml file.

sjsprecious:
Thanks @jedwards4b . I just opened a PR to introduce these new XML variables in CPL7 (https://github.com/ESCOMP/CESM_CPL7andDataComps/pull/25). Could you please review and merge it if the PR looks good to you?

jedwards4b:
@sjsprecious The workflow test for e3sm, mct above needs to pass.  Remember that these variables are not defined for e3sm.

sjsprecious:
Thanks @jedwards4b . Since I am not familiar with the E3SM build process, could you please let me know where I can add those new XML variables for the E3SM build workflow?

PR #4397, opened 2023-04-18
Update pFUnit support to version 4

### LINKED ISSUE #4394: Upgrade pFUnit support to version 4
I tried running CTSM's pFUnit-based unit tests on my Mac today, using pFUnit v3.3.3 (which is the version I have used for the last few years) and they failed due to issues in this older version of pFUnit that appear with recent python versions:

```
AttributeError: module 'collections' has no attribute 'Iterable'
```

This led me to want to update my pFUnit to version 4. However, the pFUnit support in CIME is tied somewhat to pFUnit version 3. So I have been working today at upgrading this support to version 4. There are a number of incompatibilities between version 3 and 4, so this is turning out to be non-trivial and the changes I make will lead us to be incompatible with version 3. I will keep plugging away at these changes, but wanted to get feedback on:
- Will there be any problems (e.g., for E3SM) with requiring pFUnit version 4.x moving forward?
- Have others already done work towards this goal?

@rljacob

billsacks:
pFUnit4 is backwards incompatible with v3 in a few ways. We had been using pFUnit3; this PR allows us to upgrade to v4, which is important because v3 has problems with recent python versions.

This needs to be coordinated with updates to the following other repositories; I have updated Externals.cfg with the necessary updates:
- share
- cpl7
- ccs_config

Some key elements of this are:

- pFUnit now provides its own CMake utilities that we can use in place of our pFUnit_utils.cmake. In fact, our utilities in pFUnit_utils.cmake no longer worked with pFUnit4, so this replacement is necessary.

- FindpFUnit.cmake wasn't working correctly and is not necessary as long as we explicitly point to a PFUNIT path, which we have been doing, so I removed it and changed the mechanism for pointing to this PFUNIT path in the CMake build.

- I have removed mpi support from run_tests.py because I think the way this needs to be done has changed and we're not currently using any pFUnit tests that require mpi support (at least in CESM).

- In test_sys_unittest.py, I have updated the Macros search to allow CIME's pFUnit testing to run on a user-defined machine.

Test suite:
- pytest CIME/tests/test_sys_unittest.py on cheyenne
- pytest -vvv on cheyenne

Test baseline: N/A
Test namelist changes: None
Test status: bit for bit

Fixes #4394 

User interface changes?: N


billsacks:
@rljacob - you recently mentioned that E3SM is moving to adopt the pFUnit-based unit testing framework. If so, this PR should be of interest. Is there someone on the E3SM side who should review this or are you comfortable leaving it to the CESM side to review for now?

rljacob:
I'm fine with the CESM side handling it.  No one has trained up on it yet.

billsacks:
`pytest -vvv` completed on cheyenne. The only error was one that I'm also getting on cime master:

```
CIME/tests/base.py:141: in run_cmd_assert_result
    self.assertEqual(stat, expected_stat, msg=msg)
E   AssertionError: 1 != 0 :
E       COMMAND: ./case.submit --skip-preview-namelist
E       FROM_DIR: /glade/scratch/sacks/scripts_regression_test.20230418_173359/PET_Ln9_P4.f19_f19.A.cheyenne_intel.fake_testing_only_
20230418_181300
E       SHOULD HAVE WORKED, INSTEAD GOT STAT 1
E       OUTPUT:
E       ERRPUT: ERROR: CONTINUE_RUN is true but this case does not appear to have restart files staged in /glade/scratch/sacks/script
s_regression_test.20230418_173359/PET_Ln9_P4.f19_f19.A.cheyenne_intel.fake_testing_only_20230418_181300/run rpointer.cpl
```

billsacks:
I see that the system testing for e3sm failed, but it looks like the failure is unrelated to the changes in this PR. Let me know if this is something that needs to be investigated.

jasonb5:
@billsacks The E3SM failure has been resolved, #4399. I'll re-run the workflow now.

fischer-ncar (approved):
I've merged the other PRs.  So you just need to update the Externals.

billsacks:
I reran `pytest -vvv` on cheyenne with the updated Externals and all pass, so merging.

PR #4401, opened 2023-04-20
Fix create_test --single-exe option

### LINKED ISSUE #4376: SystemTestsCompareTwo and --single-exe option
These appear to be incompatible at this time.   If I run: 
` ./create_test SMS_D_Ln9_Vnuopc.f10_f10_mg37.QPC5.cheyenne_intel.cam-ttrac_loadbal1 SMS_D_Ln9_Vnuopc.f10_f10_mg37.QPC5.cheyenne_intel.cam-ttrac_loadbal3 SMS_D_Ln9_Vnuopc.f10_f10_mg37.QPC5.cheyenne_intel.cam-ttrac_loadbal0 --single-exe`
this works fine.   However using 
`./create_test ERR_D_Ln9_Vnuopc.f10_f10_mg37.QPC5.cheyenne_intel.cam-ttrac_loadbal1 ERR_D_Ln9_Vnuopc.f10_f10_mg37.QPC5.cheyenne_intel.cam-ttrac_loadbal3 ERR_D_Ln9_Vnuopc.f10_f10_mg37.QPC5.cheyenne_intel.cam-ttrac_loadbal0 --single-exe`
causes two of the three tests to fail because variable BUILD_COMPLETE is FALSE in two of the three tests.  I have also tried other variants of the SystemTestsCompareTwo 
method and found the same problem.   @jgfouca any ideas?

jasonb5:
Adds a check in the `test_scheduler` for a tests ability to work with the 
`--single-exe` option. Checks if the test is not a compare two/n or does
not require separate builds.

Fixes `BUILD_COMPLETE` resetting to False when `--single-exe` is used
with SystemTestsCompareTwo and SystemTestsCompareN.

Test suite: pytest
Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes #4376 

User interface changes?:
Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/SystemTests/dae.py
+++ b/CIME/SystemTests/dae.py
@@ -27,7 +27,7 @@ class DAE(SystemTestsCompareTwo):
     """
 
     ###########################################################################
-    def __init__(self, case):
+    def __init__(self, case, **kwargs):
         ###########################################################################
         SystemTestsCompareTwo.__init__(
             self,
@@ -36,6 +36,7 @@ class DAE(SystemTestsCompareTwo):
             run_two_suffix="da",
             run_one_description="no data assimilation",
             run_two_description="data assimilation",
+            **kwargs,
         )
 
     ###########################################################################
--- a/CIME/SystemTests/eri.py
+++ b/CIME/SystemTests/eri.py
@@ -38,11 +38,11 @@ def _helper(dout_sr, refdate, refsec, rundir):
 
 
 class ERI(SystemTestsCommon):
-    def __init__(self, case):
+    def __init__(self, case, **kwargs):
         """
         initialize an object interface to the ERI system test
         """
-        SystemTestsCommon.__init__(self, case)
+        SystemTestsCommon.__init__(self, case, **kwargs)
         self._testname = "ERI"
 
     def run_phase(self):
--- a/CIME/SystemTests/erio.py
+++ b/CIME/SystemTests/erio.py
@@ -10,11 +10,11 @@ logger = logging.getLogger(__name__)
 
 
 class ERIO(SystemTestsCommon):
-    def __init__(self, case):
+    def __init__(self, case, **kwargs):
         """
         initialize an object interface to file env_test.xml in the case directory
         """
-        SystemTestsCommon.__init__(self, case, expected=["TEST"])
+        SystemTestsCommon.__init__(self, case, expected=["TEST"], **kwargs)
 
         self._pio_types = self._case.get_env("run").get_valid_values("PIO_TYPENAME")
         self._stop_n = self._case.get_value("STOP_N")
--- a/CIME/SystemTests/erp.py
+++ b/CIME/SystemTests/erp.py
@@ -15,7 +15,7 @@ logger = logging.getLogger(__name__)
 
 
 class ERP(RestartTest):
-    def __init__(self, case):
+    def __init__(self, case, **kwargs):
         """
         initialize a test object
         """
@@ -26,6 +26,7 @@ class ERP(RestartTest):
             run_two_suffix="rest",
             run_one_description="initial",
             run_two_description="restart",
+            **kwargs
         )
 
     def _case_two_setup(self):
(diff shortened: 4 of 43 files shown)


jedwards4b (approved):

jedwards4b:
I tested with 

` ./create_test PLB_D_Ln9_Vnuopc.f10_f10_mg37.QPC5.cheyenne_intel.cam-ttrac_loadbal1 PLB_D_Ln9_Vnuopc.f10_f10_mg37.QPC5.cheyenne_intel.cam-ttrac_loadbal3  PLB_D_Ln9_Vnuopc.f10_f10_mg37.QPC5.cheyenne_intel.cam-ttrac_loadbal0 --single-exe`

I added `<SINGLE_EXE>TRUE</SINGLE_EXE>` to the PLB test in components/cam/cime_config/config_tests.xml  
However the test still fails.   Looking at the TestStatus.log for one of the failed tests I see that case.submit is called with flag --skip-preview-namelist 
and there are no namelists created, the run fails trying to read docn_in.   

I'll try to figure out why this happens.

jedwards4b:
If I understand correctly the way that this is done EXEROOT is correctly set to point to the same executable for all cases when the --single-exe flag is used.   However I think that once the first case completes it's build all of the other cases need to call case.build (even though there is nothing to build, case.build calls preview_namelists which is necessary for all cases.)  

I think that this may be an issue in cam where the buildnml script only calls some of it's routines once and its logic for that is based on whether LockedFiles/env_build.xml exists.   

I've worked on the logic in cam's buildnml and I have solved that problem.  But I notice that the xml variable SINGLE_EXE is always true for a given test regardless of whether the --single-exe flag was used in create_test or not.   I think that we should rename this flag to be consistent with its intent - how about ALLOW_SINGLE_EXE?

jedwards4b (changes requested):
I think that the SINGLE_EXE xml variable can lead to confusion and would like it renamed to ALLOW_SINGLE_EXE.   Also I found the following change was required:

```
diff --git a/CIME/test_scheduler.py b/CIME/test_scheduler.py
index a657e2a6b..5bf8dee09 100644
--- a/CIME/test_scheduler.py
+++ b/CIME/test_scheduler.py
@@ -224,7 +224,10 @@ class TestScheduler(object):
         self._input_dir = input_dir
         self._pesfile = pesfile
         self._allow_baseline_overwrite = allow_baseline_overwrite
-        self._allow_pnl = allow_pnl
+        if single_exe:
+            self._allow_pnl = True
+        else:
+            self._allow_pnl = allow_pnl 
         self._non_local = non_local

```

jedwards4b:
@jasonb5 Please see comments in #4403.   I think that we should make sure all of the (case2 )  EXEROOT variables point to the same executable, none should be pointing to case2.

jedwards4b (approved):

jasonb5:
@jedwards4b @jgfouca I'm not sure this check is sufficient, currently the only tests that don't have `ALLOW_SINGLE_EXE` are compare-n tests with separate builds. My assumption was tests requiring separate builds would not be able to share a single exe. But with #4402 making the `PET` test set `separate_build` conditionally based on components, the current PR would not allow `PET` to be used with others tests and `--single-exe` when a compset allows.

Here are some options:
- Add a flag to allow users to bypass the `support_single_exe` check, something like `--force-tests`.
- Improve the current check to consider `separate_builds` in tests that don't have `ALLOW_SINGLE_EXE` defined.
- Completely redesign the check to be more dynamic, removing the explicit `ALLOW_SINGLE_EXE`.

jgfouca:
@jasonb5 , it sounds like allow_single_exe needs to be dynamic and set to !separate_build.

jgfouca:
@jasonb5 , any thoughts on my comment above?

jasonb5:
@jgfouca I'm reverting the changes to `config_tests.xml` and improving the logic in `support_single_exe` to consider the subclass and `separate_builds` setting.

jasonb5:
@jgfouca @jedwards4b I've moved the check back to be more dynamic.

jgfouca (approved):

PR #4402, opened 2023-04-20
Default to single-build in PEM tests

amametjanov:
Change the default from 2 separate builds in PEM tests to single build,
except for when there is `cice` or `pop` among compset components,
that require a separate build when MPI tasks change.

Test suite: PEM.T62_oQU120.CMPASO-NYF.chrysalis_intel
Test baseline: same test with separate builds
Test namelist changes: none
Test status: bit for bit

User interface changes?: none


amametjanov:
None of the other components in E3SM (except for `cice`) require separate builds when changing MPI tasks between runs in a PEM test: first run with default PEs, second run with halved PEs.
This is just a draft request for comments to see if there are any other components, which require separate builds.
If there are just a couple more, can we make an exception for them like here?
Avoiding the unnecessary second build can save substantial build time when running existing test-suites like `e3sm_integration` or new suites like E3SM-Project/E3SM#5576 (with multiple PEMs).

rljacob:
POP is another component that needs to be re-built when procs change.

amametjanov:
> Adding @jedwards4b since this could impact CESM.

@jedwards4b , please take a look at this PR ?

jgfouca:
@jedwards4b , do you approve?

jedwards4b (approved):

jgfouca:
Did an upstream merge to resolve conflicts. I will let testing hit it and then merge.

PR #4403, opened 2023-04-20
Preserve test ordering in suites with shared builds

amametjanov:
If all tests in a suite can share a single executable, it's desirable
to know which test's exe will be used by other tests in the suite: e.g.
```
    "e3sm_superbfb_ocn_opt" : {
        "share"   : True,
        "PEM.T62_oQU120.CMPASO-NYF.pemod-omp1",
        "ERS.T62_oQU120.CMPASO-NYF.pemod-omp1",
```
Currently, test scheduler re-orders tests alphanumerically and further sorts
by expected test run-times. It leads to unexpected pick of which
test gets built. This PR preserves test ordering in shared build suites:
build once, build the test listed first.

Test suite: `e3sm_superbfb_ocn`
Test baseline: same suite with `"share" : False`
Test namelist changes: none
Test status: bit for bit

User interface changes?: none

### DIFF
--- a/CIME/get_tests.py
+++ b/CIME/get_tests.py
@@ -381,8 +381,8 @@ def get_full_test_names(testargs, machine, compiler):
     expect(compiler is not None, "Must define a compiler")
     e3sm_test_suites = get_test_suites()
 
-    tests_to_run = set()
-    negations = set()
+    tests_to_run = []
+    negations = []
 
     for testarg in testargs:
         # remove any whitespace in name
@@ -390,10 +390,10 @@ def get_full_test_names(testargs, machine, compiler):
         if testarg.startswith("^"):
             negations.add(testarg[1:])
         elif testarg in e3sm_test_suites:
-            tests_to_run.update(get_test_suite(testarg, machine, compiler))
+            tests_to_run += get_test_suite(testarg, machine, compiler)
         else:
             try:
-                tests_to_run.add(
+                tests_to_run.append(
                     CIME.utils.get_full_test_name(
                         testarg, machine=machine, compiler=compiler
                     )
@@ -406,7 +406,7 @@ def get_full_test_names(testargs, machine, compiler):
 
     for negation in negations:
         if negation in e3sm_test_suites:
-            tests_to_run -= set(get_test_suite(negation, machine, compiler))
+            tests_to_run -= get_test_suite(negation, machine, compiler)
         else:
             fullname = CIME.utils.get_full_test_name(
                 negation, machine=machine, compiler=compiler
@@ -414,7 +414,7 @@ def get_full_test_names(testargs, machine, compiler):
             if fullname in tests_to_run:
                 tests_to_run.remove(fullname)
 
-    return list(sorted(tests_to_run))
+    return tests_to_run
 
 
 ###############################################################################
--- a/CIME/get_tests.py
+++ b/CIME/get_tests.py
@@ -129,6 +129,7 @@ _CIME_TESTS = {
 
 _ALL_TESTS.update(_CIME_TESTS)
 
+
 ###############################################################################
 def _get_key_data(raw_dict, key, the_type):
     ###############################################################################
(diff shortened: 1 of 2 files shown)


amametjanov:
@jgfouca this disables the call to `_order_tests_by_runtime`. AFAIK, all E3SM suites can now finish in given Jenkins wall-time time-outs: whether ordered by shortest run-time or not. I think this ordering is a remnant from when that was important. Please remind me if/when it's still necessary.

rljacob:
>  If you have a set of tasks that you can do in parallel, it's optimal to do the longer ones first.

Can you explain more about how you expect that to work in a queue system?

jgfouca:
@rljacob , unless the machine is big enough to run the entire suite simultaneously, the same theory applies. You don't want the longest job to be the last submission.

amametjanov:
> I'd like to hear why nondeterminism in the selection of the built case is a problem. The theory of a shared build test suite is that it shouldn't matter which build you use.

Yes and in my case I'd like ERS and PEM tests (same-grid, same-compset) to share a single build. It's a problem for PEM, if ERS gets picked for the shared-exe, because then there is no `case2bld` sub-dir in ERS' build-dir: ERS is a single-build test, PEM is a separate-build test (I'm trying to make PEM a single-build test by default in ESMCI/cime#4402, to mitigate this partially). So when PEM's 2nd run starts, it looks for 2nd exe and it's not there in ERS' `EXEROOT`. This would not happen if PEM gets picked for the shared-exe.

OTOH, there could be a flag to note which test should be scheduled to build in shared-exe suites. But I think deterministic builds might be more important in certain cases.

> I'd be OK with disabling time-based sorting within a shared suite

Okay, I'll look into adding a mod to disable time-based sort just for shared suites. Thanks.

jgfouca:
@amametjanov , OK, I think I understand the issue. I should have looked more closely at the tests in your suite. It was never intended to have tests, like PEM, that require a second build to be in a shared build suite. If you can figure out a way to have PEM be single build, that would be a much better solution. @boutte I know you have been looking at similar issues regarding single-exe.

jedwards4b:
With the pending update of the single-exe test it is possible to have two build tests in a build suite.   But all of the EXEROOTS should point to the same exe file, nothing should be pointing to a case2 bld directory.   Can you try reproducing this using #4401 ?

amametjanov:
> Can you try reproducing this using https://github.com/ESMCI/cime/pull/4401 ?

I merged 4401 into my local cime `master` branch. Also had to add this mod to get past an early error-check:
```
diff --git a/CIME/data/config/config_tests.xml b/CIME/data/config/config_tests.xml
index 07d693e98..b9c174d0c 100644
--- a/CIME/data/config/config_tests.xml
+++ b/CIME/data/config/config_tests.xml
@@ -594,6 +594,7 @@ NODEFAIL          Tests restart upon detected node failure. Generates fake failu
     <HIST_OPTION>$STOP_OPTION</HIST_OPTION>
     <HIST_N>$STOP_N</HIST_N>
     <CONTINUE_RUN>FALSE</CONTINUE_RUN>
+    <SINGLE_EXE>TRUE</SINGLE_EXE>
   </test>
 
   <test NAME="PET">
```
i.e. allow single-exe for PEM tests. Then
```
./cime/scripts/create_test e3sm_superbfb_ocn_opt --single-exe
```
where
```
    "e3sm_superbfb_ocn_opt" : { # opt + pureMPI
        "share"   : True,
        "time"    : "00:15:00",
        "tests"   : (
            "ERS_Lh3.T62_oQU120.CMPASO-NYF.pemod-omp1",
            "PEM_Lh3.T62_oQU120.CMPASO-NYF.pemod-omp1",
            "ERS_Ld3.T62_EC30to60E2r2.CMPASO-NYF.pemod-omp1",
            "PEM_Lh3.T62_EC30to60E2r2.CMPASO-NYF.pemod-omp1",
        )
    },
```
The scheduler picked `ERS_Ld3.T62_EC30to60E2r2.CMPASO-NYF.pemod-omp1` in alphanumeric order and built a single exe. ERS tests passed. Initial runs of PEM tests (with default PEs) succeeded. Second run (with halved PEs) failed to start with
```
  0: slurmstepd: error: execve(): /lcrc/group/e3sm/azamat/scratch/chrys/superbfb/20230404/cmp-share-exe-f-cime-newsingleexe/ERS_Ld3.T62_EC30to60E2r2.CMPASO-NYF.chrysalis_intel.pemod-omp1.C.f/bld/case2bld/e3sm.exe: No such file or directory
```

jasonb5 (approved):
LGTM

CIME/get_tests.py
```diff
@@ -406,15 +407,15 @@ def get_full_test_names(testargs, machine, compiler):
 
     for negation in negations:
         if negation in e3sm_test_suites:
-            tests_to_run -= set(get_test_suite(negation, machine, compiler))
+            tests_to_run -= get_test_suite(negation, machine, compiler)
```
jgfouca, line 410: `-=` is not supported for lists.

jgfouca (changes requested):
@jasonb5 , @amametjanov , I still have concerns with this. I don't think we should be depending on test order in order for things to work. Also, the line I commented on is not correct because list type does not support - operator. What happens if we have a  test suite with sharing on (not global sharing) via the `shared` setting? I thought the plan was to make PEM not do multiple builds in some cases, which makes test order not matter?

amametjanov:
I still cannot run `share=True` test-suites, after rebasing #4402 onto latest master: i.e. without this PR as well.
Now getting
```
2023-05-19 18:43:43: ERROR: Cannot modify case, read_only. Case must be opened with read_only=False and can only be modified within a context manager
```
for PET test-cases: initial run succeeds, second run fails to start with this error. @jasonb5 , can you take a look ?

jasonb5:
@amametjanov I'll take a look.

jasonb5:
@amametjanov #4429 will fix the `2023-05-19 18:43:43: ERROR: Cannot modify case, read_only. Case must be opened with read_only=False and can only be modified within a context manager` error.

amametjanov:
ESMCI/cime#4402 is working now. Closing this in favor of ESMCI/cime#4402.

PR #4407, opened 2023-04-21
jenkins_script: Remove from CIME

jgfouca:
This E3SM-specific jenkins test driver script does not belong in CIME.
I am moving it to E3SM.

Test suite: None
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jedwards4b (approved):

jasonb5 (approved):
LGTM

rljacob:
When we get around to designating "e3sm" and "cesm" testing as "python dict" and "xml list" testing, we will want to have some of the Jenkins pieces in case other CCS users want to set up their own Jenkins testing.   It should be generic and not tied to SNL/E3SM of course.

PR #4408, opened 2023-04-24
Reduce whitespace from append_to_user_nl_files()

samsrabin:
`append_to_user_nl_files()` can now accept a list of lines to append.
This allows many new lines to be appended without a blank line between
each.

Testing: None yet. Tagging @billsacks , who suggested I could try updating the unit test for this function.

Fixes [CIME Github issue #]: n/a

User interface changes?: No


ekluzek (approved):
Nice, simple change that looks helpful/useful.

CIME/SystemTests/test_utils/user_nl_utils.py
```diff
@@ -25,8 +25,12 @@ def append_to_user_nl_files(caseroot, component, contents):
             matching the pattern 'user_nl_clm*'. (We do a wildcard match to
             handle multi-instance tests.)
 
-        contents (str): Contents to append to the end of each user_nl file
+        contents (str or list): Contents to append to the end of each user_nl
+            file. If list, each item will be appended on its own line.
     """
+    
+    if not isinstance(contents, list):
```
billsacks, line 32: I think it would be more robust to change this to `if isinstance(contents, str)`: this way, tuples and other list-like objects will be handled the same as lists.

billsacks (changes requested):
Thanks a lot @samsrabin ! This is a nice addition and a good implementation. I have one minor request for a small logic change below - let me know if you disagree with it for any reason.

It would also be great if you could add a unit test for this new feature. There are unit tests in cime/CIME/tests/test_unit_user_nl_utils.py . See TestUserNLCopier.test_append for an example you could follow to add a new unit test. You can then run this unit test by running the following from the cime root:

pytest CIME/tests/test_unit_user_nl_utils.py

billsacks:
Also note that you will need to run this through black (via `pre-commit run -a`... see also https://github.com/ESMCI/cime/wiki/Testing) to get the checks passing.

jedwards4b:
@samsrabin you need to run black on your script and then push back the update.   One way to do this is to install pre-commit in your local repo.

samsrabin:
Hi Jim and Bill, thanks for the reminder on this. I'm having trouble running pre-commit. `pip install pre_commit` seems to work, but then `pre-commit run -a` gives me "command not found." This happens both on my Mac and on Cheyenne. What am I doing wrong?

I also tried installing pre-commit on my Mac via homebrew, and that works to a point but fails in the pylint step:
```
ImportError: cannot import name 'formatargspec' from 'inspect' (/opt/homebrew/Cellar/python@3.11/3.11.3/Frameworks/Python.framework/Versions/3.11/lib/python3.11/inspect.py)
```

samsrabin:
Oh, I guess it doesn't matter here, since the test says that pylint passed. Never mind!

CIME/SystemTests/test_utils/user_nl_utils.py
samsrabin: Changed to `if isinstance(contents, str)` in commit 1e31efc.
Unit test added in commit ab8e713.

samsrabin:
Thanks, both! I've made the requested changes, and the new test I added passes.

billsacks (approved):
This looks great - thanks a lot, @samsrabin !

PR #4416, opened 2023-05-03
case_clone: Fix bug in srcroot determination

### LINKED ISSUE #4393: create_test --srcroot problem
I have encountered an issue with the create_test script while running PEM and PET tests:

1) The following worked for these tests:
${CODE_ROOT}/cime/scripts/create_test  --output-root "${CASE_ROOT}"  --machine "${MACHINE}" --project "${PROJECT}" --wait "e3sm_ice_developer"

2) This did not work:
${CODE_ROOT}/cime/scripts/create_test  --output-root "${CASE_ROOT}"  --srcroot "${CODE_ROOT}" --machine "${MACHINE}" --project "${PROJECT}" --wait "e3sm_ice_developer"

Am I using --srcroot incorrectly?

Both of the above examples worked for ERS and SMS tests, but not PEM and PET tests, for which only the first example worked.  The tests failed to build.

jgfouca:
SRCROOT should match the original case if possible. This bug was causing multi-build tests, like PEM, to not respect a custom --srcroot argument to create_test.

This was a pretty serious bug since it, in many cases, silently built files from the wrong case when using create_test --srcroot.

Test suite: by-hand
Test baseline:
Test namelist changes:

Fixes #4393

### DIFF
--- a/CIME/case/case_clone.py
+++ b/CIME/case/case_clone.py
@@ -54,7 +54,10 @@ def create_clone(
     if os.path.isdir(os.path.join(newcase_cimeroot, "share")) and get_model() == "cesm":
         srcroot = newcase_cimeroot
     else:
-        srcroot = os.path.join(newcase_cimeroot, "..")
+        srcroot = self.get_value("SRCROOT")
+        if not srcroot:
+            srcroot = os.path.join(newcase_cimeroot, "..")
+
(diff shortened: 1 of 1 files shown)


jedwards4b (approved):
Good find.

jasonb5 (approved):
LGTM

PR #4417, opened 2023-05-03
Use compiler info when looking for test status files

jgfouca:
When test_id was being provided, compiler was being ignored. This was causing bless_test_results to match multiple sets of cases when the machine was running the same suite with multiple compilers.

Test suite: test_sys_bless_tests_results.TestBlessTestResults
Test baseline:
Test namelist changes:
Test status: bit for bit


jasonb5 (approved):
LGTM

jedwards4b (approved):

PR #4435, opened 2023-05-26
Add gpu gust

jedwards4b:
Add GPU capability for NCAR test platform gust.

Test suite: scripts_regression_tests.py SMS_Ln9_G4-a100-openacc.f19_f19_mg17.F2000dev.gust_nvhpc.cam-outfrq9s_mg3_default SMS_Ln9.f19_f19_mg17.F2000dev.gust_nvhpc.cam-outfrq9s_mg3_default
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes 
User interface changes?:

Update gh-pages html (Y/N)?:


CIME/case/case.py
```diff
@@ -1509,47 +1534,62 @@ def configure(
...
+                False,
+                "Both gpu-type and gpu-offload must be defined if either is defined",
+            )
+
+        # Set these two GPU XML variables here to overwrite the default values
+        if gpu_type:
+            self.set_value("GPU_TYPE", gpu_type)
+
+        if gpu_offload and gpu_offload != "none":
+            self.set_value("GPU_OFFLOAD", gpu_offload)
```
sjsprecious, line 1592: Interesting. In my version, I do not have these `if` statements (see https://github.com/sjsprecious/cime/blob/e3607d1ff76e78f2d41203e88267ba8d36b6bf1f/CIME/case/case.py#L1584) and I always set `GPU_TYPE` and `GPU_OFFLOAD` based on the input values. Will my version be more straightforward?

jedwards4b: Have you tested cases in which you are not using gpu's at all?  I ran into this when compiling on gust using nvhpc but for a cpu case.

sjsprecious: Hi @jedwards4b , thanks for your quick reply. My version works for NVHPC on CPU on Gust (i.e., send the job to a 128-CPU node rather than 64-CPU node). My command looks `./create_newcase --case XXXXX --mach gust --compiler nvhpc --mpilib mpich --compset F2000dev --res f19_f19_mg17 --walltime 01:00:00 --queue main --project XXXX`. Could you share what error you encountered in your case?

PR #4440, opened 2023-06-05
always use BATCH_COMMAND_FLAGS

### LINKED ISSUE #4439: Appending Arguments to `BATCH_COMMAND_FLAGS` Overwrites Existing PBS Batch Arguments in CESM Scripts
Hello, 
I would like to add a new PBS argument to my CESM job submission scripts. However, when I use `./xmlchange` to add a new argument to `BATCH_COMMAND_FLAGS`, it seems to ignore the other PBS batch arguments, rather than appending the new one to them.

Here's an example of what I'm experiencing to replicate this:
```
./create_newcase --case ~/cases/b.day1.0 --res f19_g17 --compset B1850
qcmd -- ./case.build
```
Running `./preview_run` then shows the following submission command:

```
qsub -q regular -l walltime=12:00:00 -A SCSG0002 -v ARGS_FOR_SCRIPT='--resubmit --silent' .case.run
```
-------
However, when I use `./xmlchange` to add a new `BATCH_COMMAND_FLAGS`:
```
./xmlchange --subgroup=case.run BATCH_COMMAND_FLAGS="-l place=group=rack"
```
The `./preview_run` shows:

```
qsub -l place=group=rack -v ARGS_FOR_SCRIPT='--resubmit' .case.run
```

As a result, the previous flags `-q regular -l walltime=12:00:00 -A SCSG0002` are removed, causing PBS to fail when submitting the job.

I understand that it's possible to manually include all necessary PBS arguments in `BATCH_COMMAND_FLAGS`. However, this approach doesn't seem to be the most efficient or intuitive. By default, the `BATCH_COMMAND_FLAGS` is empty and CIME automatically finds the necessary PBS information (i.e. job queue, walltime, project code). 

**More importantly, it seems that once `BATCH_COMMAND_FLAGS` is modified, CIME no longer pulls queue, project values, or walltime from the case information, which can potentially lead to confusion and errors.**

Is there a different way to append a new PBS argument to the CIME job scripts? 
If not, it could be helpful to the user community to add this functionality and address the issue with `BATCH_COMMAND_FLAGS`.

jedwards4b:
Always use the BATCH_COMMAND_FLAGS argument.
This allows users to alter with xmlchange and append with --append

Test baseline:
Test namelist changes:

Fixes #4439 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -961,10 +961,7 @@ class EnvBatch(EnvBase):
 
             return
 
-        submitargs = self.get_submit_args(case, job)
-        args_override = case.get_value("BATCH_COMMAND_FLAGS", subgroup=job)
-        if args_override:
-            submitargs = args_override
+        submitargs = case.get_value("BATCH_COMMAND_FLAGS", subgroup=job)
 
         if dep_jobs is not None and len(dep_jobs) > 0:
             logger.debug("dependencies: {}".format(dep_jobs))
--- a/CIME/case/case.py
+++ b/CIME/case/case.py
@@ -1586,6 +1586,13 @@ class Case(object):
             )
 
         env_batch.set_job_defaults(bjobs, self)
+        # Set BATCH_COMMAND_FLAGS to the default values
+
+        for job in bjobs:
+            if test and job[0] == "case.run" or not test and job[0] == "case.test":
+                continue
+            submitargs = env_batch.get_submit_args(self, job[0])
+            self.set_value("BATCH_COMMAND_FLAGS", submitargs, subgroup=job[0])
 
         # Make sure that parallel IO is not specified if total_tasks==1
         if self.total_tasks == 1:


jgfouca:
This makes me a little nervous that things will not behave as they did before. If I am understanding this right, this PR will set BATCH_COMMAND_FLAGS upon case creation. If the user xmlchanges something that would have impacted the batch command, like JOB_QUEUE, those changes won't be reflected in BATCH_COMMAND_FLAGS and so won't have the intended impact.

jedwards4b:
You are correct - It should not resolve the flags when setting BATCH_COMMAND_FLAGS - I think that currently it does.   I'll convert to draft until I fix that.

jgfouca:
@jedwards4b , thanks, this looks better! Since this is a fairly critical CIME capability, would you mind adding a scripts_regression_tests for this? I'd like to see two usecases covered:
1) User changes something like JOB_QUEUE after case.setup, confirm this change is reflected in the batch command.
2) User does `xmlchange --append BATCH_COMMAND_FLAGS XXX` to add a flag, confirm this flag is added to batch command.

jedwards4b:
```
+        # Test that changes to BATCH_COMMAND_FLAGS work
+        with Case(testdir, read_only=False) as case:
+            batch_command = case.get_value("BATCH_COMMAND_FLAGS", subgroup="case.run")
+            case.set_value("JOB_QUEUE","premium",subgroup="case.run")
+            new_batch_command = case.get_value("BATCH_COMMAND_FLAGS",subgroup="case.run")
```
I can see that new_batch_command and batch_command differ in the way that I expect, but I'm not sure how best to do that comparison programmatically.

jgfouca:
@jedwards4b , I would use the case.preview_run method and grab the batch submission string which should be the first line after `SUBMIT CMD`.

jedwards4b:
@jgfouca tests added and working on cheyenne and github.

jgfouca (approved):
Thanks!

PR #4446, opened 2023-06-14
Do not re-invoke make on repeat case.build calls

jgfouca:
Re-invoking with a full arg list can mess up the cmake cache.

Test suite: by-hand
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jasonb5 (approved):
@jgfouca LGTM, just need to fix the pre-commit issues.

Issue #4387, opened 2023-04-06
Issue with get_latest_hist_files


bartgol:
I noticed that I _had_ to add `$` to the end of my `hist_file_extension`. Take the following
```
<comp_archive_spec compname="foo" compclass="atm">
  <hist_file_extension>my\.hist</hist_file_extension>
</comp_archive_spec>
```
and consider what CIME does for an ERS test:

1. run model for phase run1
2. get latest hist file, and copy it to $file.base
3. run model for phase run2
4. get latest hist file, and copy it to $file.rest
5. compare $file.base with $file.rest

Now, during step 4, CIME calls `get_latest_hist_files` _without_ suffix (cause it's fishing for the model output files). In that routine, [these](https://github.com/ESMCI/cime/blob/b6c527d28228821bbafaf7f8ca4f6f18cdf4566b/CIME/XML/archive_base.py#L85) lines are in charge of finding the "latest" file. However, they would think that `foo.my.hist.nc.base` comes _after_ `foo.my.hist.nc`, since it does come after in lexicographical order. Since the file does not end in `.nc`, [these lines](https://github.com/ESMCI/cime/blob/b6c527d28228821bbafaf7f8ca4f6f18cdf4566b/CIME/hist_utils.py#L81) would cause it to be skipped (in a way, thank god, or else we would get false positives due to self-comparison). Therefore, I would be left without a `.rest` file for step 5.

I think the assumption of in `get_latest_hist_files` is that the latest file will contain "the latest" timestamp, so lexicographical ordering should be fine. However, for ERS tests, this seems to be a broken assumption. Unless, the user regex _ends_ with `$`, but that forces the user to put a regex that includes _all_ the file content, and this is also not great (see issue #4382).

I think the fix should be easy: since `copy_histfiles` already _assumes_ that valid files end with `.nc`, perhaps we could do the same in `get_latest_hist_files`, and only consider files ending with `.nc` (therefore, skipping `blah.nc.base`).

Thoughts?

jasonb5:
FAIL ERS.f09_g16_g.MALISIA.mappy_gnu (phase COMPARE_base_rest)
FAIL ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.mappy_gnu.mosart-rof_ocn_2way (phase COMPARE_base_rest)
FAIL ERS.r05_r05.IELM.mappy_gnu.elm-V2_ELM_MOSART_features (phase COMPARE_base_rest)
[...]

PR #4452, opened 2023-07-05
Fixes hist utils

### LINKED ISSUE #4438: Fix for `get_all_hist _files` breaks some tests
After https://github.com/ESMCI/cime/pull/4389 was introduced some tests on mappy broke.

FAIL ERS.f09_g16_g.MALISIA.mappy_gnu (phase COMPARE_base_rest)
FAIL ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.mappy_gnu.mosart-rof_ocn_2way (phase COMPARE_base_rest)
FAIL ERS.r05_r05.IELM.mappy_gnu.elm-V2_ELM_MOSART_features (phase COMPARE_base_rest)
FAIL ERS_Ld5.T62_oQU120.CMPASO-NYF.mappy_gnu (phase COMPARE_base_rest)
FAIL ERS_Ld5.T62_oQU240.DTESTM.mappy_gnu (phase COMPARE_base_rest)
FAIL PEM_Ln5.T62_oQU240wLI.DTESTM.mappy_gnu (phase COMPARE_base_modpes)

The fix caused additional `hist` files to be picked up during the `COMPARE` phase. These additional files come from the active components and are causing comparison fails. The `cpl` history files are always matched and checked before and after the fix.

Here are the files being matched when running [compare_test](https://github.com/ESMCI/cime/files/11639879/hists_match_output.txt). https://github.com/ESMCI/cime/blob/80428aa2277d7ed853b7c41c784008e39cd405f7/CIME/hist_utils.py#L383 
As an example `ERS.f09_g16_g.MALISIA` uses the following regex before
```python
'mali\\d?_?(\\d{4})?\\.hist.base$'
```
and after
```python
'mali\\d?_?(\\d{4})?\\.hist.*?\\.base$'
```
@jgfouca I'm trying to understand the correct behavior here. The first question, is it correct that CIME is checking these additional files?
If it is correct to check these files then the tests missing `.rest` either need to have their  `hist_file_extension` updated to include the `\.nc` or we need to update `get_latest_hist_files` to only process only `.nc` as discussed in #4387. This would prevent `get_latest_hist_files` from matching the `.nc.base` as the latest thus not creating the `.rest` file. The two other fails might actually be failures.

Summary of the fails:
- ERS.f09_g16_g.MALISIA.chrysalis_intel
  - Miss match between `mali.hist.am.0001-01-01_00000.nc.base` and `mali.hist.am.0001-01-07_00000.nc.rest`
- ERS.ne30pg2_r05_EC30to60E2r2.GPMPAS-JRA.chrysalis_intel.mosart-rof_ocn_2way
  - Missing `.rest` files for `mpassi`, `mpasso` and `mosart` 
- ERS.r05_r05.IELM.chrysalis_intel.elm-V2_ELM_MOSART_features
  - Missing `.rest` files for `mosart`
- ERS_Ld5.T62_oQU120.CMPASO-NYF.chrysalis_intel
  - Missing `.rest` files for `mpasso`
- ERS_Ld5.T62_oQU240.DTESTM.chrysalis_intel
  - Missing `.rest` files for `mpassi`
- PEM_Ln5.T62_oQU240wLI.DTESTM.chrysalis_intel
  - Matched `.base` and `.rest` but has actual diff on `mpassi` files

jasonb5:
Fixes regex pattern used to match history files. Only considers `.nc`
files when copying history files for a tests COMPARE_PHASE. Adds
a new optional attribute `exclude_testing` for `comp_archive_spec`
which allows a components history files to be archived but excluded
from the COMPARE_PHASE of tests when active.

Test suite: pytest CIME/tests/test*
Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes #4438 #4387 

User interface changes?: N

### DIFF
--- a/CIME/SystemTests/system_tests_common.py
+++ b/CIME/SystemTests/system_tests_common.py
@@ -489,7 +489,8 @@ class SystemTestsCommon(object):
         return allgood == 0
 
     def _component_compare_copy(self, suffix):
-        comments, num_copied = copy_histfiles(self._case, suffix)
+        # Only match .nc files
+        comments, num_copied = copy_histfiles(self._case, suffix, match_suffix="nc")
         self._expected_num_cmp = num_copied
 
         append_testlog(comments, self._orig_caseroot)
--- a/CIME/XML/archive_base.py
+++ b/CIME/XML/archive_base.py
@@ -3,6 +3,7 @@ Base class for archive files.  This class inherits from generic_xml.py
 """
 from CIME.XML.standard_module_setup import *
 from CIME.XML.generic_xml import GenericXML
+from CIME.utils import convert_to_type
 
 logger = logging.getLogger(__name__)
 
@@ -15,6 +16,25 @@ class ArchiveBase(GenericXML):
         return self.scan_optional_child(
             "comp_archive_spec", attributes={"compname": compname}
         )
+
+    def exclude_testing(self, compname):
+        """
+        Checks if component should be excluded from testing.
+        """
+
+        entry = self.get_entry(compname)
+
+        if entry is None:
+            return False
+
+        attrs = self.attrib(entry)
+
+        value = attrs.get("exclude_testing", None)
+
+        if not value:
+            return False
+
+        return convert_to_type(value, "logical")
 
     def _get_file_node_text(self, attnames, archive_entry):
         """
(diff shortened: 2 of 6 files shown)


jgfouca (approved):
Nice! This looks good to me.

jasonb5:
The behavior should still work the same as before. The fixed regex may cause new files to be picked up during the `COMPARE_PHASE`.

@jgfouca There's an incoming PR for E3SM to exclude `mpassi`, `mpaso`, `mali` and `mosart` from being compared, I'll link it this PR. There may also be other components whose history files are matched that need to be excluded as well.

PR #4461, opened 2023-07-17
replace PROJECT arg with nothing if PROJECT not required

### LINKED ISSUE #4460: Test cases failing on izumi with latest cime
I see this in cime6.0.132. Running for example: SMS_D_Ld5_PS_Vmct.f10_f10_mg37.I2000Clm50Sp.izumi_nag.clm-default, with ctsm5.1.dev130 on izumi I get the following error.

```
    Case dir: /fs/cgd/data0/erik/ctsm_worktree/ctsm_maindev/cime/scripts/SMS_D_Ld5_PS_Vmct.f10_f10_mg37.I2000Clm50Sp.izumi_nag.clm-default.20230716_014111_wujjde
    Errors were:
        env_batch.xml appears to have changed, regenerating batch scripts
        manual edits to these file will be lost!
        
        submit_jobs case.test
        Submit job case.test
        ERROR: Command: 'qsub -q medium -l walltime=08:00:00 -A -v ARGS_FOR_SCRIPT='--skip-preview-namelist' .case.test' failed with error 'qsub: script file 'ARGS_FOR_SCRIPT=--skip-preview-namelist' cannot be loaded - No such file or directory' from dir '/fs/cgd/data0/erik/ctsm_worktree/ctsm_maindev/cime/scripts/SMS_D_Ld5_PS_Vmct.f10_f10_mg37.I2000Clm50Sp.izumi_nag.clm-default.20230716_014111_wujjde'

Waiting for tests to finish
```

jedwards4b:
Strip out PROJECT flag if PROJECT is not required for system 

Test suite:  SMS_D_Ld5_PS_Vmct.f10_f10_mg37.I2000Clm50Sp.izumi_nag.clm-default 
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #4460 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -970,12 +970,11 @@ class EnvBatch(EnvBase):
         if not project:
             # If there is no project then we need to remove the project flag
             # slurm defines --account only on machines that require it, so this strip isn't required
-            if batch_system == "pbs" or batch_system=="cobalt":
-                submitargs = submitargs.replace("-A","")
+            if batch_system == "pbs" or batch_system == "cobalt":
+                submitargs = submitargs.replace("-A", "")
             elif batch_system == "lsf":
-                submitargs = submitargs.replace("-P","")
-            
-            
+                submitargs = submitargs.replace("-P", "")
+
         if dep_jobs is not None and len(dep_jobs) > 0:
             logger.debug("dependencies: {}".format(dep_jobs))
             if allow_fail:
@@ -1033,8 +1032,7 @@ class EnvBatch(EnvBase):
 
             if mail_type:
                 mail_type = mail_type.split(",")  # pylint: disable=no-member
-        
-                
+
         if mail_type:
             mail_type_flag = self.get_value("batch_mail_type_flag", subgroup=None)
             if mail_type_flag is not None:


jedwards4b:
Slurm systems only use the account flag on systems where account is required so no special action is needed.

CIME/XML/env_batch.py
```diff
@@ -964,7 +963,17 @@ def _submit_single_job(
...
             return
 
-        submitargs = case.get_value("BATCH_COMMAND_FLAGS", subgroup=job)
+        submitargs = case.get_value("BATCH_COMMAND_FLAGS", subgroup=job, resolved=False)
+
+        project = case.get_value("PROJECT", subgroup=job)
+
+        if not project:
+            # If there is no project then we need to remove the project flag
+            # slurm defines --account only on machines that require it, so this strip isn't required
```
ekluzek, line 972: Ahhh, OK, slurm is already handled, because it doesn't need it.

CIME/XML/env_batch.py
```diff
@@ -964,7 +963,17 @@ def _submit_single_job(
...
+            if batch_system == "pbs" or batch_system == "cobalt":
```
ekluzek, line 973: There's at least one cobalt variant "cobart_theta", maybe the match should be a pattern match rather than an exact match?

CIME/XML/env_batch.py
ekluzek: There are some slurm variants, but hopefully they work the same way.

ekluzek (approved):
This does work for me.

I'm concerned about having batch specific stuff here in cime, but I don't know of an alternative. And this is specific to batch systems that require the project to be removed, so it's pretty specific. So it seems reasonable that this this is the best solution.

I do wonder if the match for cobalt should be a pattern match rather than an exact one. But, I'll let you decide on that.

Other than that this is great. I'm really glad you were able to get a quick fix for this.

CIME/XML/env_batch.py
jedwards4b: cobalt_theta is only defined for one machine and that machine has PROJECT_REQUIRED=TRUE.

ekluzek: Awesome.

ekluzek:
Awesome. Would love to see this merged now.

PR #4463, opened 2023-07-19
Adds query_config driver argument

### LINKED ISSUE #4456: query_config --grids doesn't show new MOM grid unless --comp_interface nuopc is specified
When I run `query_config --grids` with cesm2_3_beta15, there is no mention of the recently added MOM grid `tx2_3v2`.
It shows up if I add the argument `--comp_interface nuopc` to the command line invocation.
It looks like this is because the default for `--comp_interface` is `mct`:https://github.com/ESMCI/cime/blob/bc19d7150c9b87384de578d3ef67f0b9e0b627b0/CIME/scripts/query_config.py#L315-L320
For CESM, this default differs from the default for `--driver` in `create_newcase`,
so CESM users see output from `query_config` that doesn't match `create_newcase` (by default).
Would it make sense to make the default for `--comp_interface` in `query_config` be `get_cime_default_driver()`,
as is done in `create_newcase`? https://github.com/ESMCI/cime/blob/bc19d7150c9b87384de578d3ef67f0b9e0b627b0/CIME/scripts/create_newcase.py#L240-L245

(Is there a reason for this option to be named `--driver` in `create_newcase` and `--comp_interface` in `query_config`? I find it confusing, but perhaps there is a distinction that I'm not aware of.)

jasonb5:
- Adds --driver to query_config
- Deprecates --comp_interface for query_config
- Fixes driver_choices in create_test

Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes #4456 
User interface changes?: N

### DIFF
--- a/CIME/scripts/create_newcase.py
+++ b/CIME/scripts/create_newcase.py
@@ -239,6 +239,8 @@ def parse_command_line(args, cimeroot, description):
 
     parser.add_argument(
         "--driver",
+        # use get_cime_default_driver rather than config.driver_default as it considers
+        # environment, user config then config.driver_default
         default=get_cime_default_driver(),
         choices=drv_choices,
         help=drv_help,
--- a/CIME/scripts/create_test.py
+++ b/CIME/scripts/create_test.py
@@ -219,7 +219,7 @@ def parse_command_line(args, description):
 
         parser.add_argument(
             "--driver",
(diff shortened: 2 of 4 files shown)


ekluzek (approved):

klindsay28:
Thanks @jasonb5!

PR #4464, opened 2023-07-24
Add support for MPASO_USE_PETSC

jgfouca:
Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/build.py
+++ b/CIME/build.py
@@ -1141,7 +1141,7 @@ def _case_build_impl(
     # the future there may be others -- so USE_PETSC will be true if
     # ANY of those are true.
 
-    use_petsc = clm_use_petsc or mpaso_use_petsc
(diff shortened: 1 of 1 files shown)


jasonb5 (approved):
LGTM

rljacob:
@bishtgautam I notice ELM uses  "USE_PETSC_LIB" as a cpp in its code.  How is that set?

bishtgautam:
@rljacob Currently, the code to add `-DUSE_PETSC_LIB` is broken in E3SM for a long time.

rljacob:
Ok.  The logic we are going for is to allow a user to specify which components need PETSc.  Any combination of on/off should be allowed.  If any of them do, add the PETSc library to the linker.

PR #4471, opened 2023-08-02
Update create-a-case.rst with how to use SourceMods

daianzhuo:
Emphasised two points on how to use SourceMods to modify a case.

1. The name of the xmlchange shell file should be shell_commands
2. SourceMods changes do not need to have the same strucutre as the component source codes

Test suite: n/a
Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jedwards4b (approved):
Thanks

jasonb5:
@jedwards4b The workflow failed because the PR was made from a fork, we can just merge it and the docs will rebuild.

PR #4474, opened 2023-08-15
Removes DAE.ww3a.ADWAV from cime_developers

jasonb5:
Remove DAE.ww3a.ADWAV from cime_developers. E3SM has removed
the ww3a grids, which caused this to fail.

Test suite: ./scripts/create_test cime_developer
Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes n/a
User interface changes?: N

### DIFF
--- a/CIME/get_tests.py
+++ b/CIME/get_tests.py
@@ -115,7 +115,6 @@ _CIME_TESTS = {
             "ERR_Ln9.f45_g37_rx1.A",
             "ERP_Ln9.f45_g37_rx1.A",
             "SMS_D_Ln9_Mmpi-serial.f19_g16_rx1.A",
-            "DAE.ww3a.ADWAV",
             "PET_Ln9_P4.f19_f19.A",
             "PEM_Ln9_P4.f19_f19.A",
             "SMS_Ln3.T42_T42.S",


jasonb5:
Removed from E3SM with https://github.com/E3SM-Project/E3SM/commit/352f73762f2cbc6d254f571640ccd3bc0f7522f8.

rljacob:
@sbrus89 did you remove those just to clean up?   If we need that case in cime_developers, we could probably add it back.

sbrus89:
@rljacob, yes. I removed the ww3a grids as a clean-up since we weren't planning to support them. We can add them back in if necessary.

jasonb5:
@rljacob I'm not sure if it's needed. @jgfouca are we fine to remove this or just replace the grid?

Looks like CESM had it disabled https://github.com/ESMCI/cime/blob/02fad90a379cdbd3c1106cbd63324480f0bf7a22/CIME/tests/test_sys_full_system.py#L44-L50 and https://github.com/ESMCI/cime/blob/02fad90a379cdbd3c1106cbd63324480f0bf7a22/CIME/tests/base.py#L212-L215

jgfouca:
@jasonb5 , we don't have any other DAE tests, so maybe just use a more common grid?

jgfouca:
This raises the question of whether the DAE test type is needed. E3SM has never used it.

jedwards4b (approved):

jedwards4b:
We are currently at a stall in the development of that component, however we do have plans to bring it in.

jgfouca:
This will leave DAE untested, but I'm fine with that.

PR #4475, opened 2023-08-17
Fixes hist_files excluding components during actual compare

jasonb5:
We were only excluding components during the `copy_histfiles`, 
need to exclude during `_compare_hists` as well otherwise
`compare_phase` complains it cannot find counterpart of old
baseline files.

Test suite: pytest  CIME/tests/*
Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes n/a
User interface changes?: N

### DIFF
--- a/CIME/hist_utils.py
+++ b/CIME/hist_utils.py
@@ -72,6 +72,8 @@ def copy_histfiles(case, suffix, match_suffix=None):
     num_copied = 0
     for model in _iter_model_file_substrs(case):
         if case.get_value("TEST") and archive.exclude_testing(model):
+            logger.info("Case is a test and component %r is excluded from comparison")
+
             continue
         comments += "  Copying hist files for model '{}'\n".format(model)
         test_hists = archive.get_latest_hist_files(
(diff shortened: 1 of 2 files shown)


CIME/hist_utils.py
```diff
@@ -72,6 +72,8 @@ def copy_histfiles(case, suffix, match_suffix=None):
     num_copied = 0
     for model in _iter_model_file_substrs(case):
         if case.get_value("TEST") and archive.exclude_testing(model):
+            logger.info("Case is a test and component %r is excluded from comparison")
```
jgfouca, line 75: What does `%r` mean?

CIME/hist_utils.py
```diff
@@ -289,6 +291,10 @@ def _compare_hists(
     archive = case.get_env("archive")
     ref_case = case.get_value("RUN_REFCASE")
     for model in _iter_model_file_substrs(case):
+        if case.get_value("TEST") and archive.exclude_testing(model):
+            logger.info("Case is a test and component %r is excluded from comparison")
```
jgfouca, line 295: Same.

jgfouca (changes requested):

CIME/hist_utils.py
jasonb5: @jgfouca Fixed, forgot to add the `model` argument.

jgfouca (approved):

PR #4477, opened 2023-08-17
gpu_enabled needs to be true only when gpu is used

### LINKED ISSUE #4476: gpu_enabled flag
This issue appears on derecho after cime6.0.151.

It appears that the gpu_enabled flag is always true and thus cpu jobs are getting  ncpus set to half the value it should be. 

-lselect=8:ncpus=64:mpiprocs=128:ompthreads=1:mem=480GB:ngpus=0:mps=1

jedwards4b:
Only set gpu_enabled if gpus are actually used. 

Test suite: PFS_Ld5.ne30pg3_ne30pg3_mg17.FLTHIST_v0d.derecho_intel
Test baseline:
Test namelist changes:
Test status: bit for bit
Fixes #4476 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/case/case.py
+++ b/CIME/case/case.py
@@ -458,7 +458,10 @@ class Case(object):
     def get_value(self, item, attribute=None, resolved=True, subgroup=None):
         if item == "GPU_ENABLED":
             if not self.gpu_enabled:
-                if self.get_value("GPU_TYPE") != "none":
+                if (
+                    self.get_value("GPU_TYPE") != "none"
+                    and self.get_value("NGPUS_PER_NODE") > 0
+                ):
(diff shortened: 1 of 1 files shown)


sjsprecious (approved):
Thanks @jedwards4b for finding this issue. Previously I can assign the CPU job to the CPU node correctly by always setting the `GPU_TYPE` variable at https://github.com/ESMCI/cime/blob/master/CIME/case/case.py#L1588 (i.e., remove the `if` check). But I tried your fix and it works as well. I do not know which way is better and I will leave it to you to make the decision. Sorry that I did not catch this error earlier.

PR #4478, opened 2023-08-21
support wildcard for machine and compiler name with meaning current d…

### LINKED ISSUE #2107: Request for test names with mods to not require machine/compiler strings
I posted as E3SM issue, but Jim F suggested I post here.

For example, these two tests:

```
ERS.f19_g16.I1850CNECACNTBC.cori-knl_gnu.clm-eca
SMS_D_Ln5.ne4_ne4.FC5AV1C-L.cori-knl_intel18.cam-cosplite_nhtfrq5

would be better if it were:

ERS.f19_g16.I1850CNECACNTBC.clm-eca
SMS_D_Ln5.ne4_ne4.FC5AV1C-L.cam-cosplite_nhtfrq5

cori07% ./create_test ERS.f19_g16.I1850CNECACNTBC.clm-eca --machine=cori-knl
ERROR: Expected 4th item of 'ERS.f19_g16.I1850CNECACNTBC.clm-eca' ('clm-eca') to be in form machine_compiler
```

If a test is not using any xml mods (i.e. ERS.f19_g16.I1850CNECACNTBC.cori-knl_gnu) then one can simply run a test by ./create_test ERS.f19_g16.I1850CNECACNTBC (i.e. no need to specify machine + compiler option). But, if a test is using test mod (ERS.f19_g16.I1850CNECACNTBC.cori-knl_gnu.clm-eca), then one needs to specify the full test name 

`./create_test ERS.f19_g16.I1850CNECACNTBC.cori-knl_gnu.clm-eca`

jedwards4b:
…efault is used

Allow machine and compiler names * in testname syntax to indicate default machine and/or compiler name.   

Test suite:
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #2107 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/utils.py
+++ b/CIME/utils.py
@@ -1149,9 +1149,6 @@ def get_full_test_name(
     ]
 
     result = partial_test
-    print(
-        f"partial_testcase {partial_testcase} partial_machine {partial_machine} partial_compiler {partial_compiler}"
-    )
     for partial_val, arg_val, name in required_fields:
         if partial_val is None:
             # Add to result based on args
@@ -1161,7 +1158,6 @@ def get_full_test_name(
                     partial_test, name
                 ),
             )
-            print(f"name = {name} result = {result}")
             if name == "machine" and "*_" in result:
                 result = result.replace("*_", arg_val + "_")
             elif name == "compiler" and "_*" in result:
@@ -1220,9 +1216,7 @@ def get_full_test_name(
                 partial_test, partial_caseopts, caseopts
             ),
         )
-    print(
-        f" partial_test {partial_test} machine {machine} compiler {compiler} result {result}"
-    )
+
     return result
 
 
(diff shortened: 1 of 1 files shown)


billsacks (approved):
Thanks, @jedwards4b ! I haven't gotten my head around the details of the implementation (let me know if you were hoping for a more careful review of the implementation itself), but I like the idea here.

Will the actual test name - in the case directory and elsewhere - keep the asterisks, or will they be replaced with the actual machine / compiler in the case name, etc.? My feeling is that it would be better for them to be replaced with the actual machine / compiler, but I don't feel strongly about this (without yet having experience with this). Some reasons are:
- Directories with `*` in the name could lead to some issues since `*` has special meaning in Unix for globbing.
- If you run these tests from multiple machines that share a file system (e.g., cheyenne and derecho) it could be good if the actual test name has the actual machine and compiler used for a given test.

jedwards4b:
The name passed to create_newcase is fully resolved and does not contain *.

billsacks:
> The name passed to create_newcase is fully resolved and does not contain *.

Awesome, thanks!

jgfouca (approved):
Nice! This was easier than I thought it would be.

PR #4484, opened 2023-09-06
Removes capture all between hist_file_extension and suffix

jasonb5:
Removes `.*` between `hist_file_extension` and `suffix` in `get_all_hist_files`.
This would cause ERIO test to fail by not finding the correct files to compare
due to the extra text allowed between the extension and suffix. If there is 
expected text between the extension and suffix this should be defined in
`hist_file_extension`.

Test suite: pytest -vvv
Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes n/a
User interface changes?: n/a

### DIFF
--- a/CIME/XML/archive_base.py
+++ b/CIME/XML/archive_base.py
@@ -146,7 +146,7 @@ class ArchiveBase(GenericXML):
             string = model + r"\d?_?(\d{4})?\." + ext
             if has_suffix:
                 if not suffix in string:
-                    string += r".*\." + suffix + "$"
+                    string += r"\." + suffix + "$"
 
                 if not string.endswith("$"):
                     string += "$"
--- a/CIME/tests/test_unit_xml_archive_base.py
+++ b/CIME/tests/test_unit_xml_archive_base.py
@@ -12,7 +12,7 @@ from CIME.XML.archive_base import ArchiveBase
 
 TEST_CONFIG = """<components version="2.0">
   <comp_archive_spec compname="eam" compclass="atm">
-    <hist_file_extension>unique\.name\.unique</hist_file_extension>
+    <hist_file_extension>unique\.name\.unique.*</hist_file_extension>
   </comp_archive_spec>
 </components>"""
 


billsacks:
Is there a potential that this will require updates to various components' `hist_file_extension` – e.g., throughout CESM? If so, what should we look for to determine if we need to change this?

jasonb5:
> Is there a potential that this will require updates to various components' `hist_file_extension` – e.g., throughout CESM? If so, what should we look for to determine if we need to change this?

There is a potential need for updates to `hist_file_extension`.

If the `hist_file_extension` was written with the assumption that extra text could appear between it's value and the suffix then it will need to be updated. You'll most likely see the need for an update if you begin seeing fails when comparing tests to their baselines or files are missing when `case.st_archive` runs.

billsacks:
What does "suffix" mean in this context? Can you give some examples of how this changes behavior so that we can tell CESM SEs what to look out for?

In case it's helpful, a common pattern in CESM components is something like:

```xml
<hist_file_extension>h\d*.*\.nc$</hist_file_extension>
```

So it allows arbitrary text before the `.nc` but not after the `.nc`. Will something like that be a problem?

jasonb5:
> What does "suffix" mean in this context? Can you give some examples of how this changes behavior so that we can tell CESM SEs what to look out for?
> 
> In case it's helpful, a common pattern in CESM components is something like:
> 
> ```
> <hist_file_extension>h\d*.*\.nc$</hist_file_extension>
> ```
> 
> So it allows arbitrary text before the `.nc` but not after the `.nc`. Will something like that be a problem?

Here's some additional context, `suffix` is referring to the postfix SystemTests uses to identify the various outputs of a test, see: https://github.com/ESMCI/cime/blob/28195b5f883f5ab1cf723725108abb1792ecf7b5/CIME/SystemTests/system_tests_common.py#L504-L509

Your example `<hist_file_extension>h\d*.*\.nc$</hist_file_extension>` should not be affected as long as the model outputs files matching that exact pattern e.g. `h012345-extra.nc`. This pattern would fail after this fix if the model outputs `h012345-extra.nc-anything` as the `-anything` would no longer be matched.

billsacks:
Thanks for this additional information. So my sense at this point is that most of our patterns in CESM are probably okay, but I'll point other CESM SEs to this discussion so that they can look into this themselves.

jgfouca (approved):

jedwards4b (approved):

PR #4485, opened 2023-09-08
env_batch: Fix flag/val separator

jgfouca:
One of the code paths for computing submit args was assuming a whitespace separator was needed, which is not right when the flag ends with '=' or ':'. This PR adds a new function that encapsulates this concept and calls it in the appropriate places in env_batch.

Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -15,6 +15,7 @@ from CIME.utils import (
     get_batch_script_for_job,
     get_logging_options,
     format_time,
+    add_flag_to_cmd,
 )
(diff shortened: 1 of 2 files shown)


jgfouca:
@amametjanov , I switched one of my machines over to pretend to be `pbspro` and was able to demonstrate the error and that this PR fixes the error. Once this PR is merged, I will update CIME in E3SM.

jedwards4b (approved):

PR #4488, opened 2023-09-19
update instructions for building cprnc with cmake

### LINKED ISSUE #3249: RCE Support in CIME for E3SM and CESM
Turns out there was redundant work going on to build in RCE support by myself,  @cacraigucar, @brianpm, @kevinareed, and @jedwards4b . Comparing my version to the one that is already merged in with CIME from NCAR, there's at least one change I'd like to propose, and there might be other after I get a chance to dig into it more. There also might be more unmerged work at NCAR, so we definitely need to coordinate to avoid this problem again in the future.

The main thing I'd like to propose is having the constant SST mode be specified with a more descriptive name rather than AQP mode "11". I think "AQPCONST" makes more sense. In my version I added an XML variable to env_run.xml that also controlled a new docn_in namelist variable so that the SST value could be controlled by the compset (something like F-CAM-RCE300 vs F-CAM-RCE295).

### LINKED ISSUE #4249: cprnc README out of date
The instructions to build cprnc are not consistent with the new directory structure.

jedwards4b:
Update README for cprnc build instructions

Test suite:
Test baseline:
Test namelist changes:
Test status: bit for bit
Fixes #4249

User interface changes?:

Update gh-pages html (Y/N)?:


CIME/non_py/cprnc/README
```diff
@@ -13,14 +13,15 @@ Quick Start Guide:
 ------------------
 
 On cime supported systems you can generate a cmake Macros file using the following
-(assuming you are running the command from the directory CIME/data/cprnc):
+(assuming you are running the command from the directory CIME/non_py/cprnc):
 
 export CIMEROOT=../..
```
billsacks, line 18: I think this line (`export CIMEROOT=../..`) is incorrect, and at least on my Mac seems to be unnecessary.

billsacks (approved):
Thanks a lot for this @jedwards4b . This procedure worked for me on my Mac (after adjusting the various arguments in the cmake call). As I noted in a comment, the export of CIMEROOT was unnecessary for me.

I would have expected not to need to explicitly set some of these various arguments to cmake - that the configure script would set them automatically. But that's probably a bigger discussion, and for now this process works. Thanks!

jedwards4b:
@mvertens this was still in a PR ready to be merged - it's on master now.

PR #4492, opened 2023-09-25
Copy only necessary cmake macros to case

jgfouca:
This implementation makes some assumptions about the context of /.../Macros.cmake, so I am only turning this on for E3SM.

I think this is a nice improvement. This will reduce the number of macros that get copied from nearly 100 to just a handful, making it easier for the user to view and edit their macros from the case.

Test suite: by-hand
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/case/case_setup.py
+++ b/CIME/case/case_setup.py
@@ -142,27 +142,26 @@ def _create_macros_cmake(
     if not os.path.isfile(os.path.join(caseroot, "Macros.cmake")):
         safe_copy(os.path.join(cmake_macros_dir, "Macros.cmake"), caseroot)
 
-    mach = mach_obj.get_machine_name()
-    if Config.instance().copy_all_cmake_macros:
-        if not os.path.exists(case_cmake_path):
-            shutil.copytree(cmake_macros_dir, case_cmake_path)
-    else:
-        if not os.path.exists(case_cmake_path):
-            os.mkdir(case_cmake_path)
+    if not os.path.exists(case_cmake_path):
+        os.mkdir(case_cmake_path)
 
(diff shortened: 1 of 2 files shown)


CIME/case/case_setup.py
```diff
@@ -141,12 +141,30 @@ def _create_macros_cmake(
...
+    if Config.instance().copy_all_cmake_macros:
+        if not os.path.exists(case_cmake_path):
+            shutil.copytree(cmake_macros_dir, case_cmake_path)
+    else:
+        if not os.path.exists(case_cmake_path):
+            os.mkdir(case_cmake_path)
+
+        # This impl is coupled to contents of Macros.cmake
+        macros = [
+            "universal.cmake",
```
jedwards4b, line 155: I think that you should include cesm.  To do that you just need to add a couple more files:
${OS}.cmake
${COMPILER}_${OS}.cmake

jedwards4b (changes requested):
I don't see anything here that limits this to e3sm.  Please just add two more files and this will apply to both models.

CIME/case/case_setup.py
jedwards4b: our Macros.cmake file is generated in cime/CIME/build.py

jedwards4b: Yes - I just tried a build with this PR and it works fine.

jgfouca: @jedwards4b , great. So do you approve of this PR now?

jedwards4b (approved):

PR #4500, opened 2023-10-16
Remove deprecated 'distutils' package from ParamGen.

### LINKED ISSUE #4499: ParamGen is using a deprecated package
It appears that the `distutils` package has been deprecated in Python 3.12, which is used in one location in ParamGen.  This is causing problems with some of the unit tests in CAM-SIMA, so this needs to be replaced with the still-supported `shutil` package.  A PR for this fix is incoming.

nusbaume:
Replaces the `find_executable` command, which is no longer available in python 3.12, with `shutil.which`, which appears to provide the same functionality (and is supported in 3.12).

Test suite:
Test baseline:
Test namelist changes:

Fixes #4499

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/ParamGen/paramgen.py
+++ b/CIME/ParamGen/paramgen.py
@@ -4,6 +4,7 @@ import re
 from copy import deepcopy
 import logging
 import subprocess
+import shutil
 
 try:
     from paramgen_utils import is_logical_expr, is_formula, has_unexpanded_var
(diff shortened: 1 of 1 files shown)


jedwards4b (approved):

alperaltuntas (approved):

nusbaume:
@jedwards4b Just sending a nudge about getting this merged soon-ish (unless there is more I need to do)?

PR #4502, opened 2023-10-20
Fix bless test results

jasonb5:
- Fixes requiring `--*-only` flag
- Fixes logic for default behavior of performing all bless types or only a single type
- Adds exclude feature

Test suite: pytest -vvv CIME/tests/test_unit_*
Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes n/a
User interface changes?:
Update gh-pages html (Y/N)?:


jasonb5:
Still needs to be tested.

jgfouca (approved):

PR #4504, opened 2023-10-20
reduce the number of times env_batch is rewriten

### LINKED ISSUE #4503: question on case_submit.py (Thomas Toniazzo)
I think that the instruction
      unlock_file(os.path.basename(env_batch.filename),caseroot=caseroot)
in line 97 might be wrong, since it seems to delete
LockFiles/env_batch.xml just before running check_lockfile; which throws
an error and thus forces env_submit.py to regenerate all batch scripts
every time.

Is this correct?

If so would it make sense to remove that line or mv it to just before
the subsequent lockfile line (although I'm not sure if that should be
part of the if clause rather than done every time).

jedwards4b:
Check if BATCH_SYSTEM needs to be updated before unlocking env_batch.xml and updating.
This should reduce the number of times env_batch.xml is overwritten.

Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #4503 

User interface changes?: No

### DIFF
--- a/CIME/case/case_submit.py
+++ b/CIME/case/case_submit.py
@@ -94,8 +94,10 @@ def _submit(
         batch_system = "none"
     else:
         batch_system = env_batch.get_batch_system_type()
-    unlock_file(os.path.basename(env_batch.filename), caseroot=caseroot)
-    case.set_value("BATCH_SYSTEM", batch_system)
+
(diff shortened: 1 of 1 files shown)


jasonb5 (approved):
LGTM

PR #4507, opened 2023-10-27
Yet another btr update

jasonb5:
- Reverts to original behavior, `bless_test_results` with no flags will bless hist and nml
- Adds performance flags `--bless-tput`, `--bless-mem`, and `--bles-perf` (both tput and mem)

Test suite: pytest -vvv CIME/tests/test_unit*
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]
User interface changes?:
Update gh-pages html (Y/N)?:


jasonb5:
Can `--bless-history` be used along with `--bless-throughput` or are they mutually exclusive?

PR #4514, opened 2023-11-13
Fixes separating std and performance blessing

jasonb5:
When blessing performance diffs, the default hist and namelists would
be run as well. This PR separates these, there is now two distinct blessing
paths; standard (hist and namelists) and performance.

Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes n/a
User interface changes?: N


jgfouca:
Thanks, @jasonb5

PR #4497, opened 2023-10-13
Add call in case_setup to run CAM shell script

lizziel:
The update is needed for use with GEOS-Chem chemistry so that run-time configuration files can be copied to the case directory upon case setup. It goes with CAM PR https://github.com/ESCOMP/CAM/pull/484 which is currently under review.

This is currently draft. I will fill in the below requirements soon.

[ Description of the changes in this Pull Request. It should be enough
information for someone not following this development to understand.
Lines should be wrapped at about 72 characters. Please also update
the CIME documentation, if necessary, in doc/source/rst and indicate
below if you need to have the gh-pages html regenerated.]

Test suite:
Test baseline:
Test namelist changes:
Test status: [bit for bit, roundoff, climate changing]

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

PR #4540, opened 2023-12-08
Add back changes from PR4497

fischer-ncar:
Put back in changes for PR #4497.  

From PR #4497.
The update is needed for use with GEOS-Chem chemistry so that run-time configuration files can be copied to the case directory upon case setup. **This PR should be merged along with CAM PR https://github.com/ESCOMP/CAM/pull/484 at the same time.**

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/case/case_setup.py
+++ b/CIME/case/case_setup.py
@@ -422,6 +422,14 @@ def _case_setup_impl(
                 run_cmd_no_fail(
                     "{}/cime_config/cism.template {}".format(glcroot, caseroot)
                 )
+            if comp == "cam":
+                camroot = case.get_value("COMP_ROOT_DIR_ATM")
+                logger.debug("Running cam.case_setup.py")
+                run_cmd_no_fail(
+                    "python {cam}/cime_config/cam.case_setup.py {cam} {case}".format(
+                    cam=camroot, case=caseroot
+                    ) 
+                )
 
         _build_usernl_files(case, "drv", "cpl")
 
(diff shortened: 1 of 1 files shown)


lizziel:
This update will be needed soon as the GEOS-Chem is in the final test phase. See https://github.com/ESCOMP/CAM/pull/484.
@fischer-ncar, @fvitt, @jedwards4b

jedwards4b (approved):

gold2718:
> Why can't cam copy files over? Why do we need cime to do this?

I think the issue is when will CAM have a chance to copy these files and is it soon enough? These files are similar in purpose and timing to the user_nl_xxx files such as user_nl_cam. In the CIME world, users are supposed to be able to edit these files before any CAM script (buildlib or buildnml) has been run which is why CIME is installing the user_nl_xxx files and not each individual component. This is the same reason that the same script installs the cism_template file.

gold2718:
> Will the copy happen only for GEOS-Chem compsets or all compsets?

If you look at cam.case_setup.py in the [CAM PR](https://github.com/ESCOMP/CAM/pull/484), you can see that it only copies files if `-chem geoschem` is specified.

lizziel:
The original implementation had the copy done by CAM within buildnml. See https://github.com/CESM-GC/CAM/blob/e2f974419dcb08ef7cbb49b2d4ebb757f3128dc9/cime_config/buildnml#L179-L199. The CIME update seems more elegant because it allows invokes component-specific instructions for case setup. Perhaps it could be updated to not require a case setup script in CAM.

fvitt:
OK. Thanks for the explanations.

jedwards4b:
@fischer-ncar I think that if you merge cime master into this branch and push a new commit the unit-testing will pass, e3sm is currently a known failure.

PR #4542, opened 2023-12-11
Reformat config machines

jedwards4b:
Reformats config machines so that the primary file contains only the regex field for machine name matching.   The
details for each machine are then read from files in machines/_machinename_/config_machines.xml
This is v3 of config_machines.xml, this change is backward compatible with v2.

Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/generic_xml.py
+++ b/CIME/XML/generic_xml.py
@@ -105,7 +105,8 @@ class GenericXML(object):
 
     def read(self, infile, schema=None):
         """
-        Read and parse an xml file into the object
+        Read and parse an xml file into the object.  The schema variable can either be a path to an xsd schema file or
+        a dictionary of paths to files by version.
         """
         cached_read = False
         if not self.DISABLE_CACHING and infile in self._FILEMAP:
--- a/CIME/XML/generic_xml.py
+++ b/CIME/XML/generic_xml.py
@@ -690,8 +690,13 @@ class GenericXML(object):
         """
         validate an XML file against a provided schema file using pylint
         """
-        expect(os.path.isfile(filename), "xml file not found {}".format(filename))
-        expect(os.path.isfile(schema), "schema file not found {}".format(schema))
+        expect(
+            filename and os.path.isfile(filename),
+            "xml file not found {}".format(filename),
+        )
+        expect(
+            schema and os.path.isfile(schema), "schema file not found {}".format(schema)
+        )
         xmllint = which("xmllint")
 
         expect(
--- a/CIME/XML/generic_xml.py
+++ b/CIME/XML/generic_xml.py
@@ -8,7 +8,7 @@ from CIME.utils import safe_copy, get_src_root
 import xml.etree.ElementTree as ET
 
 # pylint: disable=import-error
-from distutils.spawn import find_executable
+from shutil import which
 import getpass
 from copy import deepcopy
 from collections import namedtuple
@@ -126,8 +126,10 @@ class GenericXML(object):
             logger.debug("read: {}".format(infile))
             with open(infile, "r", encoding="utf-8") as fd:
                 self.read_fd(fd)
-
-            if schema is not None and self.get_version() > 1.0:
+            version = str(self.get_version())
+            if type(schema) is dict:
+                self.validate_xml_file(infile, schema[version])
+            elif schema is not None and self.get_version() > 1.0:
                 self.validate_xml_file(infile, schema)
 
             logger.debug("File version is {}".format(str(self.get_version())))
@@ -472,7 +474,7 @@ class GenericXML(object):
         xmlstr = self.get_raw_record()
 
         # xmllint provides a better format option for the output file
-        xmllint = find_executable("xmllint")
+        xmllint = which("xmllint")
 
         if xmllint:
             if isinstance(outfile, str):
@@ -690,7 +692,7 @@ class GenericXML(object):
         """
         expect(os.path.isfile(filename), "xml file not found {}".format(filename))
         expect(os.path.isfile(schema), "schema file not found {}".format(schema))
-        xmllint = find_executable("xmllint")
+        xmllint = which("xmllint")
 
         expect(
             xmllint and os.path.isfile(xmllint),
--- a/CIME/data/config/cesm/config_files.xml
+++ b/CIME/data/config/cesm/config_files.xml
@@ -43,7 +43,8 @@
     <group>case_last</group>
     <file>env_case.xml</file>
     <desc>file containing machine specifications for target model primary component (for documentation only - DO NOT EDIT)</desc>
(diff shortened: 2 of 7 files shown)


jasonb5 (changes requested):
The `config_machines_version3.xsd` schema file is missing, otherwise this looks good.

jedwards4b:
Added the config_machines_version3.xsd and the code to pass both schemas in and check the one based on the file version.

CIME/XML/generic_xml.py
```diff
@@ -126,8 +126,10 @@ def read(self, infile, schema=None):
             logger.debug("read: {}".format(infile))
             with open(infile, "r", encoding="utf-8") as fd:
                 self.read_fd(fd)
-
-            if schema is not None and self.get_version() > 1.0:
+            version = str(self.get_version())
+            if type(schema) is dict:
+                self.validate_xml_file(infile, schema[version])
+            elif schema is not None and self.get_version() > 1.0:
```
billsacks, line 132: As we discussed over lunch, I am happy with this approach. But can you please add some documentation in the docstring for this function describing the possible types for schema (i.e., that it can now be a string(?) or dict, and if the latter, what the keys are)?

billsacks (changes requested):
Awesome. I like the new format of storing machine information! I have just one inline request for a bit more documentation on a function you changed. Some other comments / possible requests:

- There are a few changes in here unrelated to the main changes here (cprnc and some test stuff). Would it make sense to separate these into their own PR? I don't feel strongly about this.
- Does this lead to changes in the recommended (or possibly required) way that people store stuff in their personal `.cime` directory? If so, is there somewhere where this should be documented?

CIME/data/config/cesm/config_files.xml
```diff
@@ -43,7 +43,8 @@
     <group>case_last</group>
     <file>env_case.xml</file>
     <desc>file containing machine specifications for target model primary component (for documentation only - DO NOT EDIT)</desc>
-    <schema>$CIMEROOT/CIME/data/config/xml_schemas/config_machines.xsd</schema>
+    <schema version="2.0">$CIMEROOT/CIME/data/config/xml_schemas/config_machines.xsd</schema>
```
rljacob, line 46: Would we need to make a similar change in E3SM's config_files.xml?

CIME/non_py/cprnc
rljacob, line 1: Are these changes intended?

rljacob:
This is welcome since config_machines.xml can become a giant file.  Does the code in the per-machine files look just like the XML blocks in the current combined file?

CIME/non_py/cprnc
jedwards4b: These changes were due to a merge to master.

CIME/data/config/cesm/config_files.xml
jedwards4b: no - this change is backward compatible

jedwards4b:
@rljacob I made it backward compatible so no changes are required in e3sm.  Yes the 
format of individual files is nearly the same as the current block - the only difference is that the NODENAME_REGEX has been moved out of  the block to the top level file.
[Here is a link](https://github.com/ESMCI/ccs_config_cesm/tree/main/machines) to the reformated cesm machines files.

CIME/non_py/cprnc
billsacks: Weird - usually a merge to master won't show up in PR diffs like this. Is the situation different for submodules like this?

billsacks (approved):
Thanks for the latest change!

rljacob:
Is there a way to do one subdir with a uniquely named file per machine instead of a uniquely named subdir for each machine?

jedwards4b:
What is your argument for that change?   I considered it, but plan to treat config_batch.xml in the same way - so there will be a config_batch.xml file in each machine subdir.

CIME/non_py/cprnc
jedwards4b: I agree this is a little strange and will make sure that the merge works correctly.

billsacks: Great, thanks!

rljacob:
If there's going to be more then one file in each subdir, then it makes more sense.  Does the subdir name have to match the string you pass to --machine ?

jedwards4b:
Yes, currently the subdirectory name is expected to match the string in MACH=

jedwards4b:
The cprnc submodule does seem to be out of sync with master, my attempts to fix the issue have failed  - @jgfouca can you have a look?  There should not be any changes in cprnc due to this PR.

jgfouca:
@jedwards4b , I just merged https://github.com/ESMCI/cime/pull/4545 , which should put us on the latest cprnc main commit. You should rebase or upstream this branch, take the cprnc commit that's on master, and the conflict will resolve.

jedwards4b:
@jgfouca I just did a rebase again - it still doesn't seem to have worked.

jgfouca:
It looks like the conflict is gone?

jgfouca:
Oh, I see this PR is still trying to change the submodule. Try this:
`git checkout origin/master $path_to_cprnc ; git add $path_to_cprnc; git commit`.  This assumes origin is the main repo, not your fork.

jedwards4b:
Still no help:
git checkout origin/master CIME/non_py/cprnc ; git add CIME/non_py/cprnc; git commit
Updated 0 paths from b6c797123
Check Xml............................................(no files to check)Skipped
Fix End of Files.....................................(no files to check)Skipped
Trim Trailing Whitespace.............................(no files to check)Skipped
black................................................(no files to check)Skipped
pylint...............................................(no files to check)Skipped
On branch reformat_config_machines
Your branch is up to date with 'mydev/reformat_config_machines'.

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	#DiagsCase.py#
	#testcase.py#
	CIME/#test_status.py#
	CIME/.#test_status.py
	CIME/ParamGen/requirements.txt
	CIME/case/#check_input_data.py#
	DiagsCase.py
	ccs_config/
	scripts/create_newcase.py.log
	scripts/create_test.log
	scripts/create_test.out
	scripts/create_test2.out
	scripts/test.out
	testcase.py

nothing added to commit but untracked files present (use "git add" to track)
derecho5: ~/sandboxes/cime
:) git push mydev
Everything up-to-date

jgfouca:
@jedwards4b , can you push this branch to the main repo? I'd like to check it out and modify it.

jedwards4b:
I could but I shouldn't have to, you can clone or add my fork with
git remote add

jgfouca:
I checked out your branch and somehow your cprnc submodule is still set to an older commit. This should fix it:'

```
git fetch $esmci_remote
git cherry-pick $esmci_remote/jgfouca/jime_branch
git submodule update
git push
```

jedwards4b:
That did it - thank you!

jgfouca (approved):

PR #4543, opened 2023-12-11
Update to latest cprnc

jgfouca:
Something is up with the old sha, d17494561ace91fc86387c83591c0382a2c79566. Jenkins jobs are having trouble fetching it and I don't see it in the commit history of the main branch. I also don't see it when I do a fresh clone. This commit updates cprnc to the latest HEAD on `main`.

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jedwards4b:
Sorry about that - a force push was done to restore the cprnc history from the cime repo, that wiped out existing hashes.
All should be good now and we don't expect anything like this to happen again.

jedwards4b (approved):

PR #4555, opened 2023-12-29
Remove manage externals

jedwards4b:
Removes the need for manage_externals from CESM.  We are still early in the discussion stage, so will set to draft. 

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jedwards4b:
@jasonb5 I'm going to need your help again.  I need to update entrypoint.sh for the tests to run on this branch, I made the changes I think I need on the entrypoint.sh of this branch, but it doesn't use that one.

jasonb5:
@jedwards4b I see the issue, I'll need to manually build and push the container. I'll work on fixing these workflows shortly.
@jgfouca Do we need submodules to be in the `git@github.com:` form?

jgfouca:
@jasonb5 , yes, it is helpful to have submodules in ssh form.

jedwards4b:
@jasonb5 I think I already did that here: https://github.com/ESMCI/cime/pull/4556

jasonb5:
@jedwards4b I'm referring to [.gitmodules](https://github.com/ESMCI/cime/blob/0dac4775041351c3d06d14d6610e474aadc92e05/.gitmodules) in this branch.

jedwards4b:
I just realized that this is going to cause e3sm problems because you will get more than you want.   I may need to think of another way to do this.

rljacob:
Maybe put all those in a .gitmodules in https://github.com/ESCOMP/CESM ?

jedwards4b:
@rljacob They are already there - the point of this .gitmodules was just to be able to do github testing without 
getting all of cesm.  I can figure out another way to do that.

jedwards4b:
I think that I just need to withdraw this PR until the CESM developers decide if they want to remove manage externals or not.  If they do decide to remove manage externals then I will put the new cesm on the trunk before reopening this PR.

PR #4558, opened 2024-01-04
Fix .env_mach_specific.sh when unsetting env

jgfouca:
When an env var value is None, this tells CIME to unset the environment. This use case was not being handled in the creation of .env_mach_specific.sh.

Test suite: by-hand
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jedwards4b (approved):
Thanks

jgfouca:
Test fail is due to a cdash network issue, merging.

PR #4562, opened 2024-01-16
Remove manage externals

jedwards4b:
This is to support a proposed change to the CESM distribution method, removing the manage externals tool and adding instead a new tool [git-fleximod](https://github.com/jedwards4b/git-fleximod).   Because E3SM does not use the fleximod tool we cannot add the submodules to cime in the same way we do the rest of the CESM repositories without populating E3SM cime with unused and confusing source code.   I am proposing a file in the docker directory indicating which repos and tags need to be cloned for cesm testing of cime.
@jasonb5 @jgfouca thoughts?   

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jedwards4b (commented):
thanks

jedwards4b:
- This change was merged in another PR

PR #4563, opened 2024-01-16
Test using dynamic container

jasonb5:
- Fix workflow to always build containers (utilizes cache for speedup)
- Testing always uses newest container
- Fix recursion error from `case.preview` call for E3SM

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]
User interface changes?
Update gh-pages html (Y/N)?:


jasonb5:
Looks like there is still a "maximum recursion depth exceeded in comparison" error introduced by #4572 and is still specific to E3SM.

jasonb5:
The recursion error was caused by `CIME.XML.env_batch._submit_single_job` not correctly handling the `dry_run` argument for tests.

E3SM system testing exposed the issue during it's pre-run provenance capture where it calls `case.preview_run` which calls `_submit_single_job` with the `dry_run` argument. Since the `dry_run` argument was not handled for tests it would cause `_submit_single_job` to submit the job again, causing the recursion error.

jasonb5:
TODO
- [ ] Update docs

jgfouca (approved):
LGTM

.github/scripts/ghcr-prune.py
```diff
@@ -0,0 +1,171 @@
...
+)
+
+kwargs = vars(parser.parse_args())
+
+logging.basicConfig(level=kwargs["log_level"])
+
+logger = logging.getLogger("ghcr-prune")
+
+
+class GitHubPaginate:
```
jedwards4b, line 37: Can you document this a little - what is it intended to do?

docker/Dockerfile
```diff
@@ -84,10 +77,38 @@ RUN curl -L -k -o "${PWD}/pnetcdf.tar.gz" \
             make install && \
             rm -rf "${PWD}/pnetcdf"
 
-RUN mkdir -p /root/.cime /storage/timings
+# CESM dependencies
```
jedwards4b, line 80: Is this the only place we need to update dependencies for cesm/cime?

jedwards4b: Should they be passed in as the external dependancies (eg ESMF_VERSION) are?

jedwards4b (approved):

docker/Dockerfile
jasonb5: Yes this is the only place you need to update dependencies for cesm/cime.

I can update these to build arguments as well but I would like to **always** have the defaults specified in the Dockerfile and not the workflow.

jedwards4b: If I understand you correctly then you should not make them arguments.

jedwards4b (approved):
Thanks!

.github/scripts/ghcr-prune.py
jasonb5: Adds description to tool and docstring for class. See the ghcr-prune workflow for usage.

PR #4565, opened 2024-01-17
Loosen type for module command to allow options to be passed before subcommand

mvdebolskiy:
This is needed for some versions of lua, since otherwise it will not recognize options like `--force` and `--quiet` which will not allow a case to be created. NCName, on the other hand does not allow to `--` to be start of the `name`.

Similar to #4375

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:
N
Update gh-pages html (Y/N)?:
N

### DIFF
--- a/config/xml_schemas/config_machines.xsd
+++ b/config/xml_schemas/config_machines.xsd
@@ -217,7 +217,7 @@
   </xs:element>
   <xs:element name="command">
     <xs:complexType mixed="true">
-      <xs:attribute ref="name" use="required"/>
+      <xs:attribute name="name" use="required" type="xs:string"/>
     </xs:complexType>
   </xs:element>
   <xs:element name="environment_variables">


jedwards4b:
@mvdebolskiy we already have a mechanism to provide this.  
<module_system type="module" allow_error="true">
Will allow lua to return errors without triggering a cime failure.

jedwards4b:
I guess maybe I can see a case where the allow_error feature may not be adequate but could you confirm please.

mvdebolskiy:
That option works. The model builds. The only concern I have is when sigma2 upgrades the machines and deprecates some toolchains, it will be harder to debug. But all good. Should I close this?

jedwards4b:
This is a concern on every machine, not in our control.

Issue #4528, opened 2023-11-27
CIME refactor


jasonb5:
After extensive analysis of CIME’s codebase using the prospector tool, it is my recommendation that the code is in need of a refactor. With this, CIME’s codebase can be simplified, improve error handling, readability and maintainability, increase unit test coverage, remove any remaining model specific code, and convert CIME to a standard package/release cycle.

The prospector tool was used to perform the analysis, it analyzes Python code and outputs information about errors, potential problems, convention violations and complexity. The tool found 2,587 instances when scanning the CIME codebase. There were 115 instances of code being marked as complex according to the cyclomatic complexity.

The first goal of the refactor is to reduce the complexity of CIME’s codebase, this will significantly improve the readability and maintainability of the code. This first step will be to remove any unused code, followed by reducing the complexity of functions identified by the prospector tool. While the code is being refactored unit tests will be added to ensure the correct function. Currently we have 67% test coverage; 36% coverage from unit tests and 31% from end-to-end tests, I would like to get this at least above 80%. Some unit tests need to be fixed as they rely on previous test output, this is not a good practice as it makes running some individual tests impossible. Fixing these tests will allow the test suite to run individual tests in parallel, decreasing the testing time and better utilizing the resources from our automated testing. Next is to improve error handling within CIME, there are many instances where generic builtin python errors are raised and not caught leaving users unaware of the true cause. After improving error handling, there is still some model specific code that went unaddressed in the first pass and will be removed. Lastly the user facing code will be cleaned up so CIME can be used like a standard 3rd party package and distributed through PyPi and Conda. Projects will still be able to use CIME as a submodule but have an installable option. With releasing a more standardized package there will be more control in what features go out to the user and enhance CIME by providing the ability to utilize it within the python ecosystem.

ekluzek:
@jasonb5 this tool you mention above sounds very useful, and very useful for other python projects we work on. Is prospector easy enough to use that I can just Google about it? Or could you give us a few notes on how to use it? You could just give a few notes here on how you used it for cime. Thanks for any insights you have. And thanks for posting about this.

Also agree with @jedwards4b that there should be some discussions on this.

jasonb5:
@jedwards4b This can all be done incrementally. To ensure CIME keeps functioning, I'll be filling in unit tests for each portion of code before I refactor, the goal is no disruptions to projects using it. I'll be tackling this all in small chunks so it shouldn't hold up any development outside of it either.

[...]

PR #4569, opened 2024-01-22
check .cime for a machine directory first

### LINKED ISSUE #4568: V3 config_machines.xml doesn't consider user `.cime` directory for machine configs
If there's a V3 `config_machines.xml` under `~/.cime/`, CIME will check for `$SRCROOT/ccs_config/machines/docker/config_machines.xml` and not `~/.cime/docker/config_machines.xml`.

Issue is in `set_machine` function: https://github.com/ESMCI/cime/blob/d2ec81a4e5ee4bc558c46f31b407c02afd759b6b/CIME/XML/machines.py#L320-L328

Should also consider `~/.cime/docker/config_machines.xml`.

jedwards4b:
Add read of .cime/machine/config_machines.xml file if it exists. 

Test suite: Hand testing with and without .cime/_machine_/config_machines.xml
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #4568 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/machines.py
+++ b/CIME/XML/machines.py
@@ -318,7 +318,7 @@ class Machines(GenericXML):
             return machine
         elif self.get_version() == 3:
             machines_file = os.path.join(
-                self.machines_dir, machine, "config_machines.xml"
+                os.environ.get("HOME"), ".cime", machine, "config_machines.xml"
             )
             if os.path.isfile(machines_file):
                 GenericXML.read(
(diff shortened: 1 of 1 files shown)


CIME/XML/machines.py
```diff
@@ -318,14 +318,24 @@ def set_machine(self, machine, schema=None):
             return machine
         elif self.get_version() == 3:
             machines_file = os.path.join(
-                self.machines_dir, machine, "config_machines.xml"
+                os.environ.get("HOME"), ".cime", machine, "config_machines.xml"
```
jasonb5, line 321: @jedwards4b Can we switch this to use `Pathlib`?

jedwards4b: We haven't been using Pathlib as a rule, but sure - do you want to do that or do you want me to?

jasonb5: Part of the refactor (https://github.com/ESMCI/cime/issues/4528, still need to update with the exact tasks) I have planned is to migrate to Pathlib, now that our minimum supported python is above 3.4.

I can make the changes.

PR #4575, opened 2024-02-02
case.submit: Fix no-batch mode.

jgfouca:
No output fron the batch env is provided to get_job_id when no_batch is on. Treat that case the same as if batchtype is none.

Also, update cprnc submodule.

We also never want file_util.copy_file to be verbose. It was cluttering the output coming from create_test on some platforms.

Test suite: by-hand.
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -1124,8 +1124,14 @@ class EnvBatch(EnvBase):
                 jobid_pattern is not None,
                 "Could not find jobid_pattern in env_batch.xml",
             )
+
+            # If no output was provided, skip the search. This could
+            # be because --no-batch was provided.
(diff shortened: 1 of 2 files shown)


jedwards4b (changes requested):
In file .gitmodules please update the fxtag to the new cprnc tag name.

jedwards4b:
Oh nevermind - I'm jumping ahead of myself, there is no fxtag in .gitmodules (yet).

jasonb5 (approved):
LGTM
Failing E3SM tests are fine, I'm still working on a fix.

PR #4577, opened 2024-02-06
Update cprnc submodule, to bring in some cmake fixes

bartgol:
There are a couple of cmake fixes in the new version of cprnc that help. For instance, I can't build e3sm on my laptop without using those fixes.

I'm not sure what the following should be

Test suite: ?
Test baseline: ?
Test namelist changes: ?
Test status: [bit for bit, roundoff, climate changing] ?

but I ran the `scripts_regression_tests.py` script as recommended on the Contributing page, on both master and the branch: master had 33 failures, the branch 34. The extra fail seems to be unrelated though:

```
FAIL: test_wait_for_test_cdash_pass (test_sys_wait_for_tests.TestWaitForTests.test_wait_for_test_cdash_pass)
  File "/home/lbertag/workdir/scream/scream-src/master/cime/CIME/tests/test_sys_wait_for_tests.py", line 260, in test_wait_for_test_cdash_pass
    self.assertFalse(
AssertionError: True is not false : wait_for_tests should have finished
```

User interface changes?: No

Update gh-pages html (Y/N)?: I don't think so?


jgfouca:
Closing. A newer cupric is already on master.

PR #4580, opened 2024-02-16
Standalone component

jedwards4b:
standalone component tests do not produce a cpl log file and so looking there for test completion of these components is a mistake.  This change will cause case_run.py to look in the component log for the completion stamp and will cause memory and performance baseline comparisons to be skipped since memory and performance data is not collected.

Test suite: cesm2.3.alpha17b
Test baseline:
Test namelist changes:
Test status: bit for bit,

Fixes 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/case/case_run.py
+++ b/CIME/case/case_run.py
@@ -320,9 +320,6 @@ def _post_run_check(case, lid):
     cpl_logfile = cpl_logs[0]
     # find the last model.log and cpl.log
     model_logfile = os.path.join(rundir, model + ".log." + lid)
-    print(
-        f"driver {driver} cpl_logfile {cpl_logfile} model_logfile {model_logfile} comp_standalone {comp_standalone}"
-    )
     if not os.path.isfile(model_logfile):
         expect(False, "Model did not complete, no {} log file ".format(model_logfile))
     elif os.stat(model_logfile).st_size == 0:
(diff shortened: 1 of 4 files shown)


CIME/case/case_run.py
```diff
@@ -322,9 +318,11 @@ def _post_run_check(case, lid):
         cpl_logs = [os.path.join(rundir, file_prefix + ".log." + lid)]
 
     cpl_logfile = cpl_logs[0]
-
     # find the last model.log and cpl.log
     model_logfile = os.path.join(rundir, model + ".log." + lid)
+    print(
```
jgfouca, line 323: Do we want to keep this print statement? If so, maybe it should be a logger call?

fischer-ncar (approved):
Looks good to me, except the print statement James pointed out.

CIME/case/case_run.py
jedwards4b: Thanks for pointing this out.  I have removed it along with another I found.

jgfouca (approved):

jedwards4b:
@jasonb5 Can I assume this is an expected failure until #4563 is merged?

jasonb5:
Yes these failures are expected and fixed in #4563.

CIME/case/case_run.py
```diff
@@ -292,16 +292,12 @@ def _post_run_check(case, lid):
...
-    model = case.get_value("MODEL")
     driver = case.get_value("COMP_INTERFACE")
-    model = get_model()
 
-    fv3_standalone = False
+    comp_standalone, model = is_comp_standalone(case)
 
-    if "CPL" not in case.get_values("COMP_CLASSES"):
-        fv3_standalone = True
     if driver == "nuopc":
```
jedwards4b, line 299: @rljacob perhaps moving the line 297 to after 299 will solve the problem?

rljacob: I can try that.

PR #4581, opened 2024-02-20
case.submit: Always try to download input data

jgfouca:
We don't want test cases downloading data while on a compute node if it can be avoided. It's harmless to call check_all_input_data twice, so let's have the case.submit call handle the 99% of cases where the main case will download all the inputs needed for the case/test. We keep the download calls in system_tests_common.py just in case case2 of a compare-two test needs more inputs.

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/case/case_submit.py
+++ b/CIME/case/case_submit.py
@@ -290,9 +290,9 @@ def check_case(self, skip_pnl=False, chksum=False):
     self.check_lockedfiles()
     if not skip_pnl:
         self.create_namelists()  # Must be called before check_all_input_data
+
     logger.info("Checking that inputdata is available as part of case submission")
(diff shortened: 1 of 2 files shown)


jgfouca:
@jasonb5 , you were the one that put the TEST check in, so please take a look at this.

jasonb5:
@jgfouca This issue is the same a https://github.com/ESMCI/cime/pull/4584. It's due to a bug that I missed in https://github.com/ESMCI/cime/pull/4580. I'll have a fix shortly.

ndkeen (approved):

PR #4582, opened 2024-02-21
need to make sure the directory exists before linking to it

jedwards4b:
The ERI test was failing after running the first ref case, its because the second run directory wasn't created and thus it wasn't possible to create links within it, I added code to create this directory if it doesn't already exist.  

Test suite:scripts_regression_tests
Test baseline:
Test namelist changes:
Test status: bit for bit
Fixes 

User interface changes?:

Update gh-pages html (Y/N)?:


jedwards4b:
@jasonb5 What is the current status of testing and how do we move forward?

fischer-ncar (approved):

jasonb5:
@jedwards4b working on a fix https://github.com/ESMCI/cime/pull/4584

PR #4607, opened 2024-03-27
fix an issue with single component runs which do use mediator

jedwards4b:
sometimes cmeps single component runs do include a mediator, this is a fix for tests of that nature.  

Test suite:
Test baseline:
Test namelist changes:
Test status: bit for bit
Fixes 

User interface changes?:

Update gh-pages html (Y/N)?:


jasonb5 (approved):
LGTM

PR #4608, opened 2024-03-28
Changes needed to add dglc component for cdeps

mvertens:
This provides changes to the CESM config_files and Makefile that permits a new dglc component to be added to cdeps.
This change will only effect CESM and NorESM.

Test suite: Testing was done with the new dglc component in place and will not effect any other CESM/NorESM configurations.
Test baseline:
Test namelist changes:
Test status: bit-for-bit

Fixes: None

User interface changes?: No


CIME/Tools/Makefile
```diff
@@ -300,6 +300,10 @@ ifeq ($(ULIBDEP),$(null))
       ULIBDEP += $(LIBROOT)/libatm.a
       CPPDEFS += -DATM_PRESENT
     endif
+    ifeq ($(GLC_PRESENT),TRUE)
```
jedwards4b, line 303: Did this need to be moved?  If not can you restore original?

jedwards4b (approved):

CIME/Tools/Makefile
mvertens: I don't know what you mean by 'moved'. I thought we needed a matching section for DGLC. Is this not the case? I think it built without this change. I will restore the original for now.

PR #4624, opened 2024-05-07
Update documentation for basic usage

rljacob:
Update the documentation for the basic usage of CCS.
Define more key concepts.  Add table for compset alias one-letter case names.
Make sure files created after create_newcase and ./case.setup match current function.
Change titles of two major sections to be about using CCS and configuring CCS.

Test suite: build locally using sphinx.
Test baseline: N/A
Test namelist changes: N/A
Test status: bit for bit

User interface changes?:  N

Update gh-pages html Y.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -4,6 +4,7 @@ buildnmlc
 buildlib.*c
 CIME.egg-info/
 build/
+_build/
 dist/
 .coverage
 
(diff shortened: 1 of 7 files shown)


rljacob:
This is just to get things in shape for the E3SM tutorial.  More cleanup is needed.

rljacob:
Merging so its live for tutorial starting in a few minutes.

PR #4627, opened 2024-05-08
Remove the standard_name length restriction for CAM-SIMA 

### LINKED ISSUE #4625: Increase standard_name length in entry_id_pg.xsd
I'm running into a build-namelist error for my new CCPP routine dadadj which contains a namelist entry with a standard name greater than the current 63 character limit.  Existing CCPP standard names can be much longer than the 63 character limit imposed by the schema file.  When searching the CCPP documentation I didn't see a length limit for standard names. If there is no standard name max length I would recommend removing the 63 character limit.  The one line mod to remove the length check would be:

12  -    <xs:pattern value="[A-Za-z][A-Za-z0-9_]{0,63}"/>
12  +   <xs:pattern value="[A-Za-z][A-Za-z0-9_]+"/>

jtruesdal:
Remove the standard_name 64 character length restriction and allow strings of 1 or more characters. This corrects a CAM-SIMA error for the new dadadj CCPP routine which contains a namelist entry with a standard name greater than the current 63 character limit. The new pattern match removes a length restriction but is consistent with the [Guidelines for Construction of CF Standard Names](https://cfconventions.org/Data/cf-standard-names/docs/guidelines.html#id2797697) which is adopted in the [CCPP Standard Name Rules](https://github.com/ESCOMP/CCPPStandardNames/blob/main/StandardNamesRules.rst#ccpp-standard-name-rules) (rule 2).

Test suite: Built and ran a CAM-SIMA snapshot test with the new routine and I also ran a create_test case with CAM on izumi.
Test baseline: cime6.0.236_httpsbranch01
Test namelist changes: N/A
Test status: bit for bit

Fixes #4625 

User interface changes?:Remove string length restriction for standard_name type.

Update gh-pages html (Y/N)?:N

### DIFF
--- a/CIME/ParamGen/xml_schema/entry_id_pg.xsd
+++ b/CIME/ParamGen/xml_schema/entry_id_pg.xsd
@@ -9,7 +9,7 @@
   <!-- simple types -->
   <xs:simpleType name="identifier_type">
     <xs:restriction base="xs:string">
-      <xs:pattern value="[A-Za-z][A-Za-z0-9_]+"/>
+      <xs:pattern value="[A-Za-z][A-Za-z0-9_]*"/>
     </xs:restriction>
   </xs:simpleType>
 
--- a/CIME/ParamGen/xml_schema/entry_id_pg.xsd
+++ b/CIME/ParamGen/xml_schema/entry_id_pg.xsd
@@ -9,7 +9,7 @@
   <!-- simple types -->
   <xs:simpleType name="identifier_type">
     <xs:restriction base="xs:string">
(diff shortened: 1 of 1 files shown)


jtruesdal:
@jedwards4b  What is the procedure for getting this reviewed and merged?  I looked at the workflow doc but didn't see any specifics about who to assign ie. who is authorized to merge? Just FYI Steve G. did comment on the fix via the associated issue.  Do I pick reviewers?

jedwards4b (approved):

PR #4629, opened 2024-05-10
Major update to testing section

rljacob:
Re-org and add to documentation on the create_test system. 
Better seperate the xml-specific parts and the python dictionary parts.
Add better explanation of testmods.

Test suite: built locally
Test baseline: N/A
Test namelist changes: N/A
Test status:  bit for bit

User interface changes?: N

### DIFF
--- a/doc/source/users_guide/testing.rst
+++ b/doc/source/users_guide/testing.rst
@@ -1,18 +1,13 @@
 .. _testing:
 
-**********
-Testing
-**********
+**************
+Testing Cases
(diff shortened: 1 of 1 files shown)


rljacob:
Someone from CESM please check the "Test control with XML files " section for accuracy.   Everything outside of that and "Test control with python dictionaries" should be applicable to both models.

rljacob:
Need this to be live for tutorial people so merging.

PR #4644, opened 2024-06-20
no longer required, remove from unit testing

jedwards4b:
esmf_wrf_time_manager was code to replace esmf in certain tests, 
it is no longer needed by cesm and these tests were never used in e3sm 

Test suite: github workflow
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 2.8)
+cmake_minimum_required(VERSION 3.14)
 include(ExternalProject)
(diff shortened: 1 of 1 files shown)


jedwards4b:
Do you use this unit testing infrastructure?  Currently the unit tests give an error if its not found, if you do use it I will change so that it will give a warning instead of an error - would that be okay?

rljacob:
No we don't use the unit testing infrastructure.

PR #4655, opened 2024-08-07
Add case git

jedwards4b:
This PR adds a case interface to git.  It creates a local git repository when case.setup is run and updates that repository with each action that triggers an update of the CaseStatus file.  Optionally a remote repository can be attached by setting CASE_GIT_REPOSITORY to the name of the remote repository.  The case name is used as the repository branch name and if the branch already exists in the remote an error is generated.  So multiple cases can exist in the same remote repository and each case is a branch of that repo.   The new file gitinterface.py can interact with git in two ways:

1. If the python git module is found it will be used.
2. If the python git module is not found git commands will be run through shell commands.

Note that the variable CASE_GIT_REPOSITORY must be defined in your driver config_component.xml in order to use this feature, it should be defined as follows:
```
<entry id="CASE_GIT_REPOSITORY">
  <type>char</type>
   <group>case_git</group>
   <file>env_build.xml</file>
   <desc>Remote git repository used for this case</desc>
</entry>
```

I have tested this code on three systems: derecho (with python git module)
izumi (without python git module) 
TACC vista (grace-hopper system no python git module)

Test suite: scripts_regression_tests on the three systems listed above
Test baseline:
Test namelist changes:
Test status: bit for bit

User interface changes?: Users who wish to use the CASE_GIT_REPOSITORY feature must have write permissions to the remote repository and it is recommended that the ssh interface to that repository be used to avoid frequent prompts for github tokens.

Update gh-pages html (Y/N)?:


billsacks (approved):
This is awesome, @jedwards4b ! This is a really excellent feature to have! Thank you for implementing it!

jgfouca (approved):
Very cool idea for a feature!

CIME/non_py/gitignore.template
jasonb5, line 1: Lets move this to `CIME/data/templates/` with the other templates.

jasonb5 (changes requested):

CIME/non_py/gitignore.template
jedwards4b: Makes sense, thanks.

jasonb5:
I ran into one issue which is more of an edge case.
I was in a new environment, created a case and went to run `./case.setup`.
image
Is there a better way to handle this?

jedwards4b:
I'm not sure how to fix that - were you on a github server or in a local environment?  I think that our github scripts set those git config variables don't they?

jasonb5:
It was an environment in a container, but it could happen on any system where the user is new and hasn't setup git, which I would consider an edge case.

Should have a fix for this, soon.
```
Looking for: ['python=3.8']

Could not solve for environment specs
The following packages are incompatible
├─ python 3.8**  is requested and can be installed;
└─ truststore is not installable because it requires
   └─ python >=3.10  but there are no viable options
      ├─ python [3.10.0|3.10.1|...|3.12.5] conflicts with any installable versions previously reported;
      └─ python 3.12.0rc3 would require
         └─ _python_rc, which does not exist (perhaps a missing channel).
```

jasonb5:
This PR https://github.com/ESMCI/cime/pull/4658 should fix that error.

jedwards4b:
Replaced by #4660

PR #4656, opened 2024-08-08
add support for new ccs_config structure - backward compatible

jedwards4b:
This allows the machine specific part of config_batch.xml to be split into an individual machine directory.  It is a backward compatible change.

Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/BuildTools/configure.py
+++ b/CIME/BuildTools/configure.py
@@ -28,7 +28,7 @@ from CIME.XML.env_mach_specific import EnvMachSpecific
 from CIME.XML.files import Files
 from CIME.build import CmakeTmpBuildDir
 
-import shutil
+import shutil, glob
 
 logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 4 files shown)


jasonb5:
Once  #4658 is merged it should fix this error.
```
Looking for: ['python=3.9']

Could not solve for environment specs
The following packages are incompatible
├─ python 3.9**  is requested and can be installed;
└─ truststore is not installable because it requires
   └─ python >=3.10  but there are no viable options
      ├─ python [3.10.0|3.10.1|...|3.12.4] conflicts with any installable versions previously reported;
      └─ python 3.12.0rc3 would require
         └─ _python_rc, which does not exist (perhaps a missing channel).
```

jedwards4b:
replaced by #4659

PR #4660, opened 2024-08-09
Add case git

jedwards4b:
Had to be reopened as a branch of escomp to run docker tests.  replaces #4655 

This PR adds a case interface to git. It creates a local git repository when case.setup is run and updates that repository with each action that triggers an update of the CaseStatus file. Optionally a remote repository can be attached by setting CASE_GIT_REPOSITORY to the name of the remote repository. The case name is used as the repository branch name and if the branch already exists in the remote an error is generated. So multiple cases can exist in the same remote repository and each case is a branch of that repo. The new file gitinterface.py can interact with git in two ways:

    If the python git module is found it will be used.
    If the python git module is not found git commands will be run through shell commands.

Note that the variable CASE_GIT_REPOSITORY must be defined in your driver config_component.xml in order to use this feature, it should be defined as follows:

<entry id="CASE_GIT_REPOSITORY">
  <type>char</type>
   <group>case_git</group>
   <file>env_build.xml</file>
   <desc>Remote git repository used for this case</desc>
</entry>

I have tested this code on three systems: derecho (with python git module)
izumi (without python git module)
TACC vista (grace-hopper system no python git module)

Test suite: scripts_regression_tests on the three systems listed above
Test baseline:
Test namelist changes:
Test status: bit for bit

User interface changes?: Users who wish to use the CASE_GIT_REPOSITORY feature must have write permissions to the remote repository and it is recommended that the ssh interface to that repository be used to avoid frequent prompts for github tokens.

Update gh-pages html (Y/N)?:


jedwards4b:
The e3sm fail in this PR is expected.  Some of the routines from file CIME/utils.py had to be moved to a new file CIME/status.py to avoid circular dependancies.  E3SM will have to update import for these routines.

jasonb5:
@jedwards4b Go ahead and merge, I'll push a change to the e3sm repo to fix the failing test.

rljacob:
@jasonb5 did you fix the failing test mentioned above?

jasonb5:
@rljacob The failing tests were fixed by https://github.com/E3SM-Project/E3SM/pull/6550.

PR #4667, opened 2024-08-21
Fix/rest n in tests

### LINKED ISSUE #4666: REST_N specification in env_tests.xml
The setting of the REST_N variable for model configurations in which different components have different timesteps is more complicated than the `$STOP_N / 2 +1 ` currently used in env_tests.py - I think that REST_N needs to be removed from config_tests.py and calculated in CIME/SystemTests/

jedwards4b:
Setting REST_N=STOP_N/2 + 1 in config_tests.xml was too simplified and required a lot of comprimises to get tests to run.  Here we move the computation of REST_N into the python 
so that tests with different NCPL times for different components work correctly.  

Test suite:
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes #4666 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/Servers/__init__.py
+++ b/CIME/Servers/__init__.py
@@ -1,9 +1,9 @@
 # pylint: disable=import-error
-from distutils.spawn import find_executable
+from shutil import which
 
-has_gftp = find_executable("globus-url-copy")
-has_svn = find_executable("svn")
-has_wget = find_executable("wget")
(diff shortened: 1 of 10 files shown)


jgfouca (approved):

jasonb5:
@jedwards4b LGTM, ignore the pre-commit, I'll address it after this has been merged.

PR #4673, opened 2024-09-03
create_test: add --driver support for e3sm

jgfouca:
Also, fix a couple issues in driver handling in test_scheduler.

Test suite:
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/test_scheduler.py
+++ b/CIME/test_scheduler.py
@@ -209,11 +209,12 @@ class TestScheduler(object):
         workflow=None,
         chksum=False,
         force_rebuild=False,
+        driver=None,
     ):
         ###########################################################################
         self._cime_root = get_cime_root()
         self._cime_model = get_model()
-        self._cime_driver = get_cime_default_driver()
+        self._cime_driver = driver if driver is not None else get_cime_default_driver()
         self._save_timing = save_timing
         self._queue = queue
         self._test_data = (
@@ -649,6 +650,7 @@ class TestScheduler(object):
         mpilib = None
         ninst = 1
         ncpl = 1
+        driver = self._cime_driver
         if case_opts is not None:
             for case_opt in case_opts:  # pylint: disable=not-an-iterable
                 if case_opt.startswith("M"):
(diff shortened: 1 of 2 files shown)


CIME/test_scheduler.py
```diff
@@ -681,15 +683,16 @@ def _create_newcase_phase(self, test):
                         )
                     )
                 elif case_opt.startswith("V"):
-                    self._cime_driver = case_opt[1:]
```
jgfouca, line 684: The original code looks like a bug. The presence of one test with Vdriver would change the driver for all the tests.

CIME/test_scheduler.py
```diff
@@ -924,7 +932,9 @@ def _xml_phase(self, test):
                 elif opt.startswith("A"):
                     # A option is for testing in ASYNC IO mode, only available with nuopc driver and pio2
                     envtest.set_test_parameter("PIO_ASYNC_INTERFACE", "TRUE")
-                    envtest.set_test_parameter("CIME_DRIVER", "nuopc")
```
jgfouca, line 927: This also looked really problematic. Setting the driver to nuopc even if that doesn't match cime_driver?

jgfouca:
@jedwards4b , is this error `ERROR: ESMF version should be 8.6.1 or newer` expected in CESM?

CIME/test_scheduler.py
jedwards4b: That's been there for a long time, seems like we would have seen problems by now if that were the case.   On the other hand, I agree.

CIME/test_scheduler.py
jedwards4b: PIO async mode is only supported in nuopc - but since cesm only supports the nuopc driver now, this is okay to remove.

jedwards4b (commented):
No, I don't think that this is expected - this was passing in the previous several PR's - 
` ESMF version should be 8.6.1 or newer expected in CESM?`. Did something in docker regress somehow?

jgfouca:
@jedwards4b , would you mind trying a CESM test on this branch? I don't see how this PR is causing an ESMF problem.

jedwards4b:
According to the docker README: `docker/README.md:| ESMF_VERSION | Version of ESMF, the default will install the latest | 8.2.0`
I don't see it set anywhere so you think that the latest would be used - the latest is 8.6.1.
https://anaconda.org/conda-forge/esmf says v8.6.1

jedwards4b:
@jgfouca I don't either and I think that we may need to rethink this docker testing framework - it seems like it's been a problem in nearly every cime PR since it was introduced.  Testing on derecho now...

jedwards4b:
cesm tests are passing locally -

jedwards4b (approved):
Ran cesm tests on derecho with no issues - the .github fail seems to be a docker issue.

CIME/test_scheduler.py
rljacob: You mean adding one Vdriver a test in a suite would change all tests in that suite?

jgfouca: @rljacob , yes. `self._cime_driver` is shared for all tests.

jgfouca:
I see the same failure on a PR that changed nothing: https://github.com/ESMCI/cime/pull/4674

Merging.

PR #4675, opened 2024-09-04
Fix shutils typo

mkstratos:
Import the shutil module rather than the "shutils" module typo

User interface changes?: N


jedwards4b (approved):
Thanks

PR #4676, opened 2024-09-04
Fix imports of copytree

jgfouca:
Somehow, this mistake slipped through testing. I would have expected that pylint would catch this.

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jedwards4b:
#4675 already addressed this...

PR #4677, opened 2024-09-09
make the hidden status of batch jobs xml dependent 

jedwards4b:
Currently this is harded for case.st_archive only.  This change
makes the hidden status of job scripts an xml variable in the config_workflow.xml file.
The default is hidden and case.st_archive is still an exception so this change is backward
compatible.

Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -32,6 +32,7 @@ class EnvBatch(EnvBase):
         initialize an object interface to file env_batch.xml in the case directory
         """
         self._batchtype = None
+        self._hidden_batch_script = {}
         # This arbitrary setting should always be overwritten
         self._default_walltime = "00:20:00"
         schema = os.path.join(utils.get_schema_path(), "env_batch.xsd")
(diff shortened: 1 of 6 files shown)


jgfouca (approved):

mnlevy1981 (changes requested):
[I think this comment is due to inadvertently typing in the comment box when trying to leave several individual comments]

mnlevy1981 (approved):
Looks good to me, and I verified that updating my `env_workflow.xml` to include

```diff
diff --git a/machines/config_workflow.xml b/machines/config_workflow.xml
index 7753f0f..117b470 100644
--- a/machines/config_workflow.xml
+++ b/machines/config_workflow.xml
@@ -71,6 +71,7 @@
   <workflow_jobs id="case.cupid" prepend="default">
     <job name="case.cupid">
       <template>template.cupid</template>
+      <hidden>False</hidden>
       <dependency>case.st_archive</dependency>
       <prereq>1</prereq>
       <runtime_parameters MACH="derecho">
```

works as expected. Thanks!

CIME/XML/env_batch.py
```diff
@@ -257,7 +258,31 @@ def make_batch_script(self, input_template, job, case, outfile=None):
...
+        hidden = env_workflow.get_value("hidden", subgroup=job)
+        # case.st_archive is not hidden for backward compatibility
+        if (
+            (job != "case.st_archive" and hidden is None)
+            or hidden == "True"
+            or hidden == "true"
+        ):
+            self._hidden_batch_script[job] = True
+        else:
+            self._hidden_batch_script[job] = False
```
mnlevy1981, line 272: How about

```py3
self._hidden_batch_script[job] = ((hidden is None and job != "case.st_archive") or (hidden is not None and hidden.lower() == "true"))
```

(The `hidden is not None` is needed because `AttributeError: 'NoneType' object has no attribute 'lower'`)
instead of the if statement?

PR #4678, opened 2024-09-09
Fix BFAILs, they need to be considered DIFF

jgfouca:
Having BFAILs resolve to overall FAIL prevent bless_test_results and other tools from working properly. For E3SM, a BFAIL has always been considered a DIFF.

Test suite: by-hand
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/test_status.py
+++ b/CIME/test_status.py
@@ -468,8 +468,8 @@ class TestStatus(object):
                     if rv in [NAMELIST_FAIL_STATUS, TEST_PASS_STATUS]:
                         phase_responsible_for_status = phase
                         # need to further inspect message to determine
-                        # phase status
-                        if "DIFF" in data[1]:
+                        # phase status. BFAILs need to be a DIFF
+                        if "DIFF" in data[1] or TEST_NO_BASELINES_COMMENT in data[1]:
                             rv = TEST_DIFF_STATUS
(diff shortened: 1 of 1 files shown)


jedwards4b (approved):

ekluzek:
Since, this changes behavior in how testing works for us in CESM, I'm highlighting this. I have found it important and helpful to have a way to distinguish between baselines that truly fail and those that the baseline doesn't exist. I'm not saying I can't live with the change, but need to think about it. I'll bring it to the attention of others in CESM.

jgfouca:
@ekluzek , this is how CIME has always worked until about a month ago. Are you guys (CESM) already dependent on the new behavior? We (E3SM) were dependent on the old behavior.

jgfouca:
@ekluzek , this PR changed the behavior:
https://github.com/ESMCI/cime/pull/4653

I did my best to keep the spirit of that PR but restoring the BFAIL behavior.

cacraigucar:
The CAM folks would like the old implementation restored (this PR).  We have not updated our cime external to the one where the change was introduced.  I'm not sure if it would break our workflow or not, so we'll just wait for this PR to be brought in before we update cime next.

PR #4679, opened 2024-09-12
improve functionality of hidden workflow flag

jedwards4b:
Improve the functionality of the hidden xml flag in workflow

Test suite: workflow test
Test baseline:
Test namelist changes:
Test status: bit for bit
Fixes 
User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -32,7 +32,6 @@ class EnvBatch(EnvBase):
         initialize an object interface to file env_batch.xml in the case directory
         """
         self._batchtype = None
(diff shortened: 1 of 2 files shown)


mnlevy1981 (approved):
Everything worked as expected when I introduce a new job to a workflow and use `<hidden>false</hidden>`: the new job created `case.cupid` instead of `.case.cupid`, `preview_run` showed `case.run`, `case.st_archive`, and `case.cupid`, and everything got submitted to the queue with the correct dependencies.

I didn't look closely at the python code or do more extensive testing.

PR #4684, opened 2024-09-26
refine the git interface to handle no git and old git

jedwards4b:
It may be that the installed git is too old for the cesm git interface or that
git is missing on a system (for example not available on compute nodes).  This PR makes the git interface conditional on git being present and above a minimum version

Test suite:scripts_regression_tests
Test baseline:
Test namelist changes:
Test status: bit for bit
Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jgfouca (approved):

jedwards4b:
cesm testing is still an expected failure.  Hoping @jasonb5 can work on that soon...

PR #4686, opened 2024-09-27
Fixes test status reporting errors during successful baseline comparison

### LINKED ISSUE #4685: Confusing BASELINE result: PASS but also "Could not interpret CPRNC output"
I'm confused by the TestStatus I'm seeing in a test I just ran with baseline comparisons:

```
PASS SMS_Ld2.ne30pg3_t232.BLT1850.derecho_intel.allactive-defaultio BASELINE cesm3_0_alpha03c.derecho: Could not interpret CPRNC output
```

There are two things that I find confusing about this:

(1) It's not clear to me what CPRNC output is leading to this "Could not interpret" message. I'm attaching the TestStatus.log file here. At a glance, it looks like all cprnc output is showing typical "IDENTICAL" lines.

[TestStatus.log](https://github.com/user-attachments/files/17157424/TestStatus.log)

(2) If it's really the case that there's at least one CPRNC file for which the code can't interpret the output, I feel like it would be safest to mark the result as a FAIL rather than a PASS, so that this isn't overlooked.

It looks like @jasonb5 introduced this in https://github.com/ESMCI/cime/commit/e33bedee161b8ba844b7109420799cb0316cba52, so Jason, I'm assigning this to you in hopes that you can help clarify this for me.

jasonb5:
The `get_ts_synopsis` function's catch-all would report an error when
the baseline comparison was actually success and had a `PASS` state.

This PR fixes the issue by explicitly checking if the the CPRNC output
reports identical files first. This allows the catch-all to still function for
unknown errors that may appear in the CPRNC output.

Test suite: doctests
Test baseline: n/a
Test namelist changes: n/a
Test status: n/a

Fixes #4685 
User interface changes?: n

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -39,20 +39,20 @@ jobs:
       packages: write
     steps:
       - name: Checkout code
-        uses: actions/checkout@v3
+        uses: actions/checkout@v4
       - name: Set up QEMU
-        uses: docker/setup-qemu-action@v2
+        uses: docker/setup-qemu-action@v3
       - name: Set up Docker Buildx
-        uses: docker/setup-buildx-action@v2
+        uses: docker/setup-buildx-action@v3
       - name: Login to DockerHub
-        uses: docker/login-action@v2
+        uses: docker/login-action@v3
         with:
           registry: ghcr.io
           username: ${{ github.actor }}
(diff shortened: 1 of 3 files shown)


billsacks (approved):
Thank you for this fix @jasonb5 - looks good!

jedwards4b (approved):

PR #4692, opened 2024-10-09
Add mem variables

jedwards4b:
PBS on derecho requires specifying memory requirement per node, this pr provides that capability.  MEM_PER_TASK and MAX_MEM_PER_NODE are defined in cmeps, and supported here.  This is easily extendable to other systems if needed.

Test suite: scripts_regression_tests, ERP_Ln9_P24x3.f45_f45_mg37.QPWmaC6.derecho_intel.cam-outfrq9s_mee_fluxes
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes 
User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -224,11 +224,32 @@ class EnvBatch(EnvBase):
             if thread_count:
                 overrides["thread_count"] = thread_count
         else:
-            total_tasks = case.get_value("TOTALPES") * int(case.thread_count)
+            total_tasks = case.get_value("TOTALPES") 
             thread_count = case.thread_count
         if int(total_tasks) * int(thread_count) < case.get_value("MAX_TASKS_PER_NODE"):
             overrides["max_tasks_per_node"] = int(total_tasks)
+        
+        # when developed this variable was only needed on derecho, but I have tried to
+        # make it general enough that it can be used on other systems by defining MEM_PER_TASK and MAX_MEM_PER_NODE in config_machines.xml
+        # and adding {{ mem_per_node }} in config_batch.xml
+        try:
+            mem_per_task = case.get_value("MEM_PER_TASK")
+            max_mem_per_node = case.get_value("MAX_MEM_PER_NODE")
+            mem_per_node = total_tasks 
 
+            if mem_per_node < mem_per_task:
+                mem_per_node = mem_per_task
+            elif mem_per_node > max_mem_per_node:
+                mem_per_node = max_mem_per_node
+            print(f"mem_per_node={mem_per_node} total_tasks={total_tasks} thread_count={thread_count}")
+            overrides["mem_per_node"] = mem_per_node
+        except TypeError as error:
+            #ignore this, the variables are not defined for this machine
+            pass
+        except Exception as error:
+            print("An exception occured:", error)
+        
+            
         overrides["ngpus_per_node"] = ngpus_per_node
         overrides["mpirun"] = case.get_mpirun_cmd(job=job, overrides=overrides)
         return overrides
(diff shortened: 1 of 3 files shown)


CIME/XML/env_batch.py
```diff
@@ -224,11 +224,33 @@ def get_job_overrides(self, job, case):
             if thread_count:
                 overrides["thread_count"] = thread_count
         else:
-            total_tasks = case.get_value("TOTALPES") * int(case.thread_count)
+            total_tasks = case.get_value("TOTALPES")
```
jgfouca, line 227: This change scares me a lot. Looking at a couple lines below, I see total_tasks being multiplied by thread_count. It makes no sense how that ever worked because that would have the thread_count multiplied twice. So I would approve of the change if that was the only use of total_tasks, but it isn't.

CIME/XML/env_batch.py
```diff
@@ -224,11 +224,33 @@ def get_job_overrides(self, job, case):
...
             thread_count = case.thread_count
         if int(total_tasks) * int(thread_count) < case.get_value("MAX_TASKS_PER_NODE"):
```
jgfouca, line 229: Maybe change this to:
```
if int(total_tasks) < case.get_value("MAX_TASKS_PER_NODE"):
```

CIME/XML/env_batch.py
```diff
@@ -224,11 +224,33 @@ def get_job_overrides(self, job, case):
...
             overrides["max_tasks_per_node"] = int(total_tasks)
 
+        # when developed this variable was only needed on derecho, but I have tried to
+        # make it general enough that it can be used on other systems by defining MEM_PER_TASK and MAX_MEM_PER_NODE in config_machines.xml
+        # and adding {{ mem_per_node }} in config_batch.xml
+        try:
+            mem_per_task = case.get_value("MEM_PER_TASK")
+            max_mem_per_node = case.get_value("MAX_MEM_PER_NODE")
+            mem_per_node = total_tasks
```
jgfouca, line 238: And this to:
`mem_per_node = case.get_value("TOTALPES")`

CIME/XML/env_batch.py
jedwards4b: I removed it because I the behavior I saw was that total_tasks was totalpes*thread_count*thread_count (twice).  With this removed, total_tasks=totalpes*thread_count which I think is the intent.

CIME/XML/env_batch.py
jedwards4b: I think that's correct.

CIME/XML/env_batch.py
jedwards4b: Can't do that because total_tasks may be the product of an override - for example the case.st_archive script overrides total_tasks and changes it to 1.

CIME/XML/env_batch.py
jasonb5: Ah gotcha so this was just a bug then. So `task_count` from a jobs override is equivalent to `TOTALPES`?

jedwards4b: Yes - that's correct.

jgfouca: @jedwards4b , I do see the multiply below on line 229 is definitely wrong if we keep the original code for line 227. Are you saying case.get_value("TOTALPES") already takes threads into account?

CIME/XML/env_batch.py
jgfouca: How about `mem_per_node = total_tasks / thread_count` ?

CIME/XML/env_batch.py
jasonb5: If `task_count` and `TOTALPES` are equivalent then this makes more sense.
```python
            total_tasks = case.get_value("TOTALPES")
            thread_count = case.thread_count
        
        total_tasks *= thread_count
        
        if int(total_tasks) < case.get_value("MAX_TASKS_PER_NODE"):
            overrides["max_tasks_per_node"] = int(total_tasks)
```

PR #4694, opened 2024-10-10
clarify prereq option in case.submit

### LINKED ISSUE #4693: job dependencies using --prereq and --batch-args behave differently
Two jobs A and B, B with `CONTINUE_RUN=.TRUE.` depends on A and A will generate restart files for B. 

- using `--batch-args="--dependency=afterok:JOBID"` job B failed with an error "ERROR: CONTINUE_RUN is true but this case does not appear to have restart files staged in "
- using `--prereq JOBID`  job B was submitted successfully

jedwards4b:
Clarify --prereq and --batch-args options in case.submit

Test suite:
Test baseline:
Test namelist changes:

Fixes #4693 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/Tools/case.submit
+++ b/CIME/Tools/case.submit
@@ -62,7 +62,9 @@ def parse_command_line(args, description):
     parser.add_argument(
         "--prereq",
         help="Specify a prerequisite job id, this job will not start until the\n"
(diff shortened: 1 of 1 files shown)


jasonb5 (approved):
LGTM

PR #4696, opened 2024-10-17
correct the cmake path for configuring tools like fortran_unit_tests

jedwards4b:
The cmake path for external tools was incorrect based on ccs_config_cesm1.0.4 and newer.  This should not adversely affect e3sm.

Test suite: This problem was found when running the ctsm unit test suite,
that suite now works correctly. cime/scripts/fortran_unit_testing/run_tests.py --build-dir unit_tests 
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


billsacks (approved):
Thank you for this fix, @jedwards4b !

PR #4726, opened 2024-12-30
corrected leap year correction logic

jedwards4b:
This fixes the logic for leap year correction in the case in which the final year is a leap year but the final date is prior to March 1.  

Test suite:
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes 

User interface changes?:

Update gh-pages html (Y/N)?:


peverwhee (approved):
thanks @jedwards4b ! this fixes my failing test

PR #4727, opened 2025-01-02
add try except clauses to reduce errors in testing

jedwards4b:
Avoid errors by not attempting to change read only case and using try except on git commit so that if there is nothing to commit no error is issued.  Fixes CTSM SSP tests.

Test suite: SSP_Ld4.f09_g17.I1850Clm50BgcCrop.derecho_intel.clm-ciso_rtmColdSSP
Test baseline:
Test namelist changes:
Test status: bit for bit
Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


ekluzek:
Thanks for figuring this out and bringing this in @jedwards4b! I'll do some testing with this in CTSM with the update.

PR #4746, opened 2025-02-06
Fix blesses for non-bfb testing suite

mkstratos:
When blessing the non-bfb suite, not all history files were archived.
This generates an incomplete baseline, as these test depend on ensembles.

Test suite: e3sm_nbfb


jgfouca (approved):

jgfouca:
Looks like this may have broken unit tests. If you can fix it quickly, I'll add this to the CIME update I'm doing for E3SM.

mkstratos:
This update fixes the unit tests (apart from the conda error on 3.8). I didn't fully understand how the unit tests passed the values, but I think this puts parameters in the right place now.

PR #4748, opened 2025-02-10
Details of test-mod searching should not be printed to info

jgfouca:
It clutters-up the output when running create_test.

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


rljacob:
Why don't you want these messages?

jgfouca:
@rljacob , I didn't like how create_test output, which usually only contains case phase information, had these test-mod messages mixed in.

PR #4750, opened 2025-02-14
Mac update

### LINKED ISSUE #4733: resource.setrlimit (CIME/XML/env_mach_specific.py) always fails on Darwin and should be skipped
This call does not work on macOS/Darwin and will return an error which should be caught and ignored.

https://github.com/ESMCI/cime/blob/master/CIME/XML/env_mach_specific.py#L162:

```
def _get_resources_for_case(self, case):
    resource_nodes = self.get_children("resource_limits")
    if resource_nodes is not None:
        nodes = self._compute_resource_actions(resource_nodes, case)
        for name, val in nodes:
            attr = getattr(resource, name)
            limits = resource.getrlimit(attr)
            logger.info(
                "Setting resource.{} to {} from {}".format(name, val, limits)
            )
            limits = (int(val), limits[1])
            resource.setrlimit(attr, limits)
```

If the resource node is identified as macOS, the resource.setrlimit() call should be skipped. 

`config_machines.xml` does not include an architecture or OS parameter, but one can use `platform` to identify the OS:
```
>>> import platform
>>> platform.system()
'Darwin'
```

Some context courtesy of Terry Lambert: 

https://ai-fast-downward.dmi.unibas.ch/msg7484

> > The corresponding information is hard to find, but Mac OS seems to
> *intentionally* provide these APIs (ulimit on the shell, resource.setrlimit in
> Python) but ignore them. There is an article on this on lists.apple.com, which
> is currently only available on the Wayback Machine:
> 
> https://web.archive.org/web/20150908044241/https://lists.apple.com/archives/unix-porting/2005/Jun/msg00115.html
> 
> In case this will no longer be available in the future, here is the most
> relevant part:
> 
> ======================================================================================
> 
> The setrlimit() administrative limits are not enforced except for certain ones
> which are enforcible in BSD code. Those which would require enforcement in Mach
> code (VM-based enforcement or scheduler- based enforcement) are not currently
> enforced at all.
> 
> These administrative limits were never intended to save the system from fragile
> code; the way to do that is to make the code less fragile. They historically
> date back to the days when CPU time and memory usage were billable resources,
> and process accounting and system accounting were used on old timeshare machines
> that were purchased by groups, such as several University departments, because
> there was no way for a single University department to be able to afford its own
> PDP-8. Today, they are basically knobs that have to be there so old code can be
> ported, but which don't actually have to do anything, for the most part.
> 
> The limits are considered administrative, as they are not intended to permit the
> creation of security policy dependencies, nor are they intended to mitigate
> problems related to application memory leaks.
> 
> The following is the current(0) MacOS X resource limit support table:
> 
> 
>     ------------- ---------    ---------
>     Limit name    Standard?    Enforced?
>     ------------- ---------    ---------
>     RLIMIT_CPU      YES          NO
>     RLIMIT_FSIZE    YES          YES(1)
>     RLIMIT_DATA     YES          NO
>     RLIMIT_STACK    YES          YES(1)
>     RLIMIT_CORE     YES          YES
>     RLIMIT_AS       YES          NO
>     RLIMIT_RSS      NO(2)        NO
>     RLIMIT_MEMLOCK  NO           NO
>     RLIMIT_NPROC    NO           YES(3)
>     RLIMIT_NOFILE   YES          YES
>     ------------- ---------    ---------
> 
> 
> (0) This table is subject to change.
> 
> 
> (1) Not entirely correctly enforced; specifically, boundary conditions are
> clipped before or after the boundary, rather than exactly at it.
> 
> (2) RLIMIT_RSS is a MACOS X-specific alias for RLIMIT_AS.
> 
> 
> (3) Whether this limit is the upper bounds depends on current sysctl settings on
> the machine.
> 
> See also:
> 
> 
> http://www.opengroup.org/onlinepubs/009695399/functions/ setrlimit.html
> 
> To determine whether or not a limit is enforced by an OS, rather than using the
> return code of setrlimit(), you must first use getrlimit() and compare the
> result to RLIM_INFINITY; if it's RLIM_INFINITY, then you should not set a limit
> (you can, but you will confuse child processes which call getrlimit() themselves
> to check for enforcement, if you do).

jedwards4b:
rlimit support is not available on macos
Fixes #4733

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -25,7 +25,7 @@ jobs:
     runs-on: ubuntu-latest
     if: github.event_name == 'push' || github.event_name == 'workflow_dispatch'
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           ref: 'gh-pages'
           fetch-depth: 0
@@ -64,17 +64,17 @@ jobs:
     name: Build and deploy documentation
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v3
+      - uses: actions/checkout@v4
         with:
           fetch-depth: 0
           lfs: true
       - name: Install python 3.x
-        uses: actions/setup-python@v2
+        uses: actions/setup-python@v5
         with:
           python-version: '3.x'
       # https://github.com/actions/cache/blob/main/examples.md#python---pip
       - name: Cache pip
-        uses: actions/cache@v3
+        uses: actions/cache@v4
         with:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pip-${{ hashFiles('doc/requirements.txt') }}
@@ -91,7 +91,7 @@ jobs:
         if: |
           github.event_name == 'pull_request' &&
           github.event.pull_request.head.repo.full_name == github.repository
-        uses: peaceiris/actions-gh-pages@v3
+        uses: peaceiris/actions-gh-pages@v4
         with:
           github_token: ${{secrets.GITHUB_TOKEN}}
           publish_dir: './_build/html'
--- a/CIME/XML/env_mach_specific.py
+++ b/CIME/XML/env_mach_specific.py
@@ -6,7 +6,7 @@ from CIME.XML.standard_module_setup import *
 from CIME.XML.env_base import EnvBase
 from CIME import utils
 from CIME.utils import transform_vars, get_cime_root
-import string, resource
+import string, resource, platform
 from collections import OrderedDict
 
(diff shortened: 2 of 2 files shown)


jedwards4b:
I accidentally merged in the docs_update branch, I backed it out to only have my small change but it still contains the workflow changes, I don't think that hurts anything.

jasonb5 (approved):
LGTM

PR #4760, opened 2025-03-06
Allow blessing if test only failed GENERATE

samsrabin:
Sometimes a test might fail only in the GENERATE step; for example, if the user doesn't have write permissions in the baseline directory. This PR makes it so that `bless_test_results` does try to bless things in such cases.

Test suite:
Test baseline:
Test namelist changes:

**Fixes:** None

**User interface changes?:** No

**Update gh-pages html (Y/N)?:** No

### DIFF
--- a/CIME/bless_test_results.py
+++ b/CIME/bless_test_results.py
@@ -513,7 +513,9 @@ def is_hist_bless_needed(
                 if phase_result is TEST_FAIL_STATUS:
                     only_failed_generate = False
                     break
-        if not only_failed_generate:
+        if only_failed_generate:
+            needed = True
+        else:
             broken_blesses.append((test_name, "test did not pass"))
             logger.warning(
                 "Test '{}' did not pass due to phase {}, not safe to bless, test status = {}".format(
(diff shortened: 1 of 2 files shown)


jedwards4b (approved):
I would like to have a review from one of the e3sm developers to make sure that this doesn't adversely affect them.

samsrabin:
The one failing unit test is `test_is_bless_needed_overall_fail`. Seems like its name and assertions should be changed to reflect that a test failing only in GENERATE should pass blessing. Then another test should be added to check for overall failure due to something other than just GENERATE.

jasonb5:
@jgfouca What do you think about this functionality for E3SM?

jgfouca (approved):
I think this change is fine for E3SM.

jedwards4b:
@samsrabin does the unit-testing need to be fixed?

jasonb5:
Just need testing to pass and I'll review.

jasonb5:
@samsrabin The test is setup by mocking `TestStatus` object which is passed to `is_hist_bless_needed`.

I would update `ts.get_status.side_effect = ["PASS", "FAIL"] + ["PASS"] * (len(ALL_PHASES) - 1)` to `ts.get_status.side_effect = ["PASS",]*len(ALL_PHASES)+1` so the function will still test the original intended behavior.

Then can copy it and update it with `ts.get_status.side_effect = ["PASS",]*10 + ["FAIL",] + ["PASS",]*5`. This should test when the `GENERATE_PHASE` is marked as `FAIL`.

samsrabin:
All tests are now working!

jasonb5 (approved):
LGTM

PR #4761, opened 2025-03-06
update python minimum to 3.8 and update cprnc module.

### LINKED ISSUE #4032: Need to update minimum python version check
This line

https://github.com/esmci/cime/blob/master/scripts/Tools/standard_script_setup.py#L17

needs to be updated.

What is the minimum version now? 3.5? I know that, according to https://github.com/ESMCI/cime/issues/3965, we want 3.6 or later, but I'm not sure if we have any 3.6-specific features yet - though we could change that to 3.6 if we think we'll have 3.6-specific features soon.

I'm labeling this critical because I think we're going to start getting help requests from confused users soon unless we do this.

jedwards4b:
Update min python version to 3.8
Also updates to latest cprnc version

Test suite:
Test baseline:
Test namelist changes:

Fixes #4032 

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/SystemTests/system_tests_common.py
+++ b/CIME/SystemTests/system_tests_common.py
@@ -201,7 +201,7 @@ class SystemTestsCommon(object):
         if not startdate:
             startdate = self._case.get_value("RUN_STARTDATE")
         if "-" in startdate:
-            startdatetime = datetime.fromisoformat(startdate) + timedelta(
+            startdatetime = datetime.strptime(startdate, "%Y-%m-%d") + timedelta(
                 seconds=int(starttime)
             )
         else:
--- a/CIME/Tools/standard_script_setup.py
+++ b/CIME/Tools/standard_script_setup.py
@@ -20,7 +20,7 @@ def check_minimum_python_version(major, minor):
         + str(major)
         + ", minor version "
         + str(minor)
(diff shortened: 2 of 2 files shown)


rljacob:
Can you also update the error message as below?
```
   msg = (
        "Python "
        + str(major)
        + ", minor version "
        + str(minor)
        + " is required to run CIME. You have "
        + str(sys.version_info[0])
        + "."
        + str(sys.version_info[1])
```

jedwards4b:
Yes, I did intentionally update cprnc to the latest as well.

rljacob:
Since we test as far back as 3.8, does that mean those tests will fail as soon as we use some python thing that is NOT in 3.8?  That would then tell us to either update this string again or try a different 3.8-friendly approach.

jedwards4b:
Updated datetime to be backward compatible.

jgfouca:
@jedwards4b , can we merge this?

rljacob:
@jedwards4b since you "Updated datetime to be backward compatible." can it work with python 3.6?   Do we want to lower the minimum back to 3.6?

jedwards4b:
I don't think so - python 3.6 is already end of life, as is python 3.7.

PR #4763, opened 2025-03-07
safe_copy: Use /bin/cp as last resort

jgfouca:
We are seeing some very strange file system errors on frontier.

/bin/cp seems to be more reliable there for unknown reasons.

Recommend viewing PR with whitespace changes off. I only added a couple lines but I put a large block of code within a try statement.

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/utils.py
+++ b/CIME/utils.py
@@ -1367,6 +1367,39 @@ def copy_globs(globs_to_copy, output_directory, lid=None):
             )
 
 
+def copy_over_file(src_path, tgt_path):
+    """
+    Copy a file over a file that already exists
+    """
+    st = os.stat(tgt_path)
+    owner_uid = st.st_uid
+
+    # Handle read-only files if possible
+    if not os.access(tgt_path, os.W_OK):
+        if owner_uid == os.getuid():
+            # I am the owner, make writeable
+            os.chmod(tgt_path, st.st_mode | statlib.S_IWRITE)
+        else:
+            # I won't be able to copy this file
+            raise OSError(
+                "Cannot copy over file {}, it is readonly and you are not the owner".format(
+                    tgt_path
+                )
+            )
+
+    if owner_uid == os.getuid():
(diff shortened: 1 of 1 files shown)


CIME/utils.py
```diff
@@ -1391,49 +1391,54 @@ def safe_copy(src_path, tgt_path, preserve_meta=True):
...
-        )
-    else:
-        shutil.copy(
-            src_path,
-            tgt_path,
-        )
+            shutil.copy(
+                src_path,
+                tgt_path,
+            )
```
jasonb5, line 1437: This should be put into a function to keep the try/catch body easy to ready.

CIME/utils.py
```diff
@@ -1391,49 +1391,54 @@ def safe_copy(src_path, tgt_path, preserve_meta=True):
...
+    except OSError:
+        # Some systems get weird OSErrors when using shutil copy, try an
+        # old-fashioned cp as a last resort
+        run_cmd_no_fail(f"/bin/cp -f {src_path} {tgt_path}")
```
jasonb5, line 1441: Should we handle an edge case for `cp` not being in `/bin`?

jasonb5 (changes requested):

jedwards4b (approved):
I agree with the suggested change that Jason made - at least make sure that /bin/cp exists before using it.

jgfouca:
I updated the PR to clean up the implementation a bit, pulling out most of the code from the try/catch block. Also, used shut.which for a more-robust way to find `cp`.

jasonb5:
@jgfouca I know this is an edge case but if all fails and `cp` doesn't exist should we warn the user?

jgfouca:
@jasonb5 , done

jasonb5 (approved):
LGTM, thanks @jgfouca

PR #4765, opened 2025-03-13
Fix ERR test for E3SM

jgfouca:
E3SM does not use DRV_RESTART_POINTER.

Test suite: by hand, ERR case
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


jgfouca:
CI fail is not related to this PR. merging.

PR #4772, opened 2025-03-27
Build Kokkos in the CAM workflow

sjsprecious:
The PR adds the changes to allow the CAM workflow to build Kokkos library from scratch.

The `buildlib.kokkos` and `buildlib_util.py` scripts are borrowed from the [E3SM reop](https://github.com/E3SM-Project/E3SM/tree/master/share/build).

This PR is expected to work together with an updated `ccs_config` tag once [this PR](https://github.com/ESMCI/ccs_config_cesm/pull/225) is merged.

Test suite: FADIAB.ne5_ne5_mg37.derecho.intel
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


CIME/build_scripts/buildlib.kokkos
```diff
@@ -48,62 +50,102 @@ OR
...
+        print("no value found in env for Kokkos_ROOT. building from scratch")
+
     srcroot = case.get_value("SRCROOT")
-    kokkos_dir = os.path.join(srcroot, "externals", "kokkos")
-    expect(os.path.isdir(kokkos_dir), "Missing kokkos submodule")
+    cime_model = case.get_value("MODEL")
+    if cime_model == "e3sm":
+        ekat_dir = os.path.join(srcroot, "externals", "ekat")
+    else:
+        ekat_dir = os.path.join(srcroot, "libraries", "ekat")
```
jasonb5, line 92: Can we remove the model check and just check which folder exists? @jedwards4b Would we encounter a case where both a present?

jasonb5 (changes requested):
We want to keep model checks like this out of CIME.

CIME/build.py
```diff
@@ -215,10 +215,11 @@ def generate_makefile_macro(case, caseroot):
...
     make_args += " SMP={} ".format(stringify_bool(case.get_build_threaded()))
-    expect(
-        not (uses_kokkos(case) and not shared_lib),
-        "Kokkos is not supported for classic Makefile build system",
-    )
+    if case.get_value("MODEL") != "cesm":
+        expect(
+            not (uses_kokkos(case) and not shared_lib),
+            "Kokkos is not supported for classic Makefile build system",
+        )
```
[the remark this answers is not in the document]
jedwards4b: I think that the cesm part was added because we found that we didn't need this check, the kokkos build works fine with the classic Makefile system.   It can be removed.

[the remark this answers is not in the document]
sjsprecious: Thanks @jasonb5 and @jedwards4b . Since we can build Kokkos in CAM now, I just removed this check as Jim suggested.

CIME/build_scripts/buildlib.kokkos
sjsprecious: Actually I think only CAM will use this script to build kokkos (E3SM should use a different one at https://github.com/sjsprecious/E3SM/blob/master/share/build/buildlib.kokkos). So I may just need to set the path for CAM here? @jedwards4b @jasonb5

CIME/build.py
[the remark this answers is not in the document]
sjsprecious: @jedwards4b I also removed the `shared_lib` argument since it seems no longer needed. But this will fail the CI test because we need to change the [buildlib.csm_share](https://github.com/ESCOMP/CESM_share/blob/14338bef3fa604d49160e376257264db1d3313e5/buildlib.csm_share#L51) as well. Is it ok to leave the CESM CI test failed here or do you just want to keep the `shared_lib` argument to do some fake checks (to pass the pylint check)?

CIME/build_scripts/buildlib.kokkos
```diff
@@ -48,62 +50,98 @@ OR
...
+        # cmake build system will do that soon enough, so any error will be caught there.
+        print(f"Using pre-installed Kokkos_ROOT: {kokkos_root}")
+        return
+    else:
+        print("no value found in env for Kokkos_ROOT. building from scratch")
+
     srcroot = case.get_value("SRCROOT")
-    kokkos_dir = os.path.join(srcroot, "externals", "kokkos")
-    expect(os.path.isdir(kokkos_dir), "Missing kokkos submodule")
+    ekat_dir = os.path.join(srcroot, "libraries", "ekat")
```
bartgol, line 88: This seems to be *very* e3sm-specific. I don't think other CIME customers may need/want/have ekat (and should not be forced to clone ekat just to get kokkos). Also, ekat is in `externals/ekat` in e3sm.

bartgol: Maybe we could do something like
```py
kokkos_dir = None
kokkos_subdirs = {'core', 'container', 'simd', 'algorithm'}
for dirpath, dirnames, filenames in os.walk(os.path.join(srcroot,"externals")):
        if os.path.basename(dirpath) == 'kokkos' and 'LICENSE' in filenames and kokkos_subdirs.issubset(set(dirnames)):
		kokkos_dir = dirpath
```
This does not rely on a particular location of the kokkos folder (other than being in `ROOT/externals` somewhere). Note: checking `kokkos` dir content is necessary, as ekat has other `kokkos` subdirs.

sjsprecious: Thanks @bartgol . This change is actually CAM-specific because there is an ongoing project to bring the theta-l and theta-l_kokkos dycores into CAM, which requires the Kokkos library. Note that CAM does not have `$SRCROOT/externals` but `$SRCROOT/libraries` instead, where we typically save the external libraries.

I thought I needed both ekat and kokkos repo to build Kokkos library and get theta-l/theta-l_kokkos dycores compiled. But if ekat is not needed, I can remove it here to reduce complexity (also not cloned in CAM as well).

CIME/Tools/Makefile
```diff
@@ -656,6 +655,14 @@ ULIBS += -L$(INSTALL_SHAREDPATH)/lib
 
 ULIBS += -lcsm_share -L$(INSTALL_SHAREDPATH)/lib $(PIOLIBNAME) -lgptl
 
+# Add Kokkos header files and libraries for SE-NH dycore
```
jgfouca, line 658: I'm wondering if this small make block is sufficient to introduce Kokkos into the Makefile build system? On the E3SM side, we use CMake, so we can just do:
`target_link_libraries($target PUBLIC Kokkos::kokkos)`

and we get everything we need, including whatever build and link flags Kokkos thinks it needs to inject for the current platform.

CIME/build_scripts/buildlib.kokkos
jgfouca: You guys should feel free to use EKAT if you want, but it may make more sense to just add a Kokkos submodule in `srcroot/libraries/kokkos` instead of putting EKAT there. If you guys plan to actually use EKAT features, then I withdraw this comment.

jgfouca (approved):
Tentative approval but I have questions.

CIME/Tools/Makefile
sjsprecious: Thanks @jgfouca . Currently we only plan to use kokkos for the `theta-l` and `theta-l_kokkos` dycores in CAM, and I confirm that this block is sufficient.

CIME/build_scripts/buildlib.kokkos
bartgol: I agree with Jim. Also, you guys may want to keep up with kokkos releases at a different pace than us (faster or slower). So it is a better approach to just have `github.com/kokoks/kokkos` as a submodule (e3sm uses a fork of kokkos, following a policy of using in-house forks for tpls).

bartgol: Re: `externals` vs `libraries`. Should CIME offer this as an XML model config setting? Something like `external_tpls_src_root`?

jgfouca: One thing you should consider, especially if you are going to assume CESM things like the location of the kokkos submodule, you can have this script live directly in CESM like we do for E3SM.

sjsprecious: Thanks @jgfouca . I looked at the `pm-cpu` machine file for EAMxx (https://github.com/E3SM-Project/E3SM/blob/master/components/eamxx/cmake/machine-files/pm-cpu.cmake#L1-L5) and it includes EKAT there. Thus I assume that we need EKAT in order to build theta-l and theta-l_kokkos dycores in CAM?

CIME/build_scripts/buildlib.kokkos
jgfouca: Yes, E3SM will never use this. It's just a minor design flaw to have CESM-specific stuff in CIME file.

CIME/build_scripts/buildlib.kokkos
bartgol: No, the dycore does not _need_ ekat. Right now, the cmake build system relies on some ekat's cmake script to look for and possibly build kokkos. But since CAM does not use CMake, you can do away with ekat altogether. You should be able to just use the kokkos repo.

CIME/build.py
```diff
@@ -211,7 +211,7 @@ def generate_makefile_macro(case, caseroot):
         )
         fd.write(all_output)
 
-
+# pylint:disable=shared_lib
```
jedwards4b, line 214: I think that this is wrong.  It should be something like 
disable=unused-argument

CIME/build_scripts/buildlib.kokkos
sjsprecious: Thanks @jgfouca and @bartgol . After talking with @jedwards4b , we decide to revoke the kokkos changes here and move this script to live in a CAM-specific place instead as you all suggest.

For @bartgol , so do I understand correctly that EKAT is needed in order to build Kokkos? And this is the reason that kokkos stays insides EKAT in the E3SM structure like `externals/ekat/extern/kokkos`?

CIME/build.py
sjsprecious: Ah I am just gonna ask you why this does not work. Thanks and commit the fix!

CIME/build_scripts/buildlib.kokkos
sjsprecious: Thanks @jgfouca . I have moved this script to a CAM-specific place now.

CIME/build_scripts/buildlib.kokkos
bartgol: No, you don't need EKAT to build kokkos. Ekat provides some convenience cmake scripts (which may be convenient for us, just b/c we wrote them with US in mind, they may not be great for everyone).

The reason kokkos sits in ekat in e3sm is that ekat is its own pkg, which other projects used, so it made sense for ekat to ship it. At that point, e3sm had 2 kokkos submodules (which were not necessarily in sync): `externals/kokkos` and `externals/ekat/extern/kokkos`. We decided to rm the former, so we only had to keep track of one...

But, again, you don't need ekat.

bartgol: I should mention that the SE kokkos dycore has some duplicated code with ekat, so _at some point_ we may ekat-ify homme, so that we are more uniform across eamxx (e.g., hommexx has its own "Vector" structure, which is a simd-like struct very similar to the ekat::Pack one). But this should not affect CESM.

sjsprecious: Ah, thanks @bartgol for your clarification. That is clear to me now.

Maybe this is a dumb question: what is EKAT? And what is its relationship with Kokkos?

CIME/build.py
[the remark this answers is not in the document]
sjsprecious: After talking with @jedwards4b , we can bypass the check of `shared_lib` argument rather than removing this argument.

CIME/build_scripts/buildlib.kokkos
bartgol: EKAT started as a bunch of utilities _inside_ eamxx (which used to be just called "scream"). When other projects decided they could use some of those utilities, we extracted them in a pkg, which we called E3SM Kokkos Application Toolkit (EKAT), since at the time it contained mostly simd utilities and stuff related to kokkos (e.g., a tridiag solver, kokkos policies utilities, kokkos views manipulation, etc).

It then eveloved into a bunch of utilities, not just kokkos related. E.g., cmake utilities for unit testing generation, interface to a logging library, yaml parser/writer, string manipulation, etc. We never bothered changing the name, even though it should just be "E3SM C++ utils miscellanea" :)

sjsprecious: Gotcha! That is really good to know. Thanks for explaining it.

jedwards4b (approved):

bartgol (approved):

jasonb5 (approved):
LGTM

PR #4787, opened 2025-04-16
Revert mct guard for kokkos

jgfouca:
We have MOAB tests that use Kokkos, so we need to revert this change.

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/build.py
+++ b/CIME/build.py
@@ -755,7 +755,7 @@ def _build_libraries(
     if mpilib == "mpi-serial":
         libs.insert(0, mpilib)
 
-    if uses_kokkos(case) and comp_interface == "mct":
+    if uses_kokkos(case):
         libs.append("kokkos")
 
     # Build shared code of CDEPS nuopc data models
--- a/CIME/build.py
+++ b/CIME/build.py
@@ -755,7 +755,7 @@ def _build_libraries(
     if mpilib == "mpi-serial":
         libs.insert(0, mpilib)
 
-    if uses_kokkos(case):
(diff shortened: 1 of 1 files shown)


CIME/build.py
```diff
@@ -755,7 +755,7 @@ def _build_libraries(
     if mpilib == "mpi-serial":
         libs.insert(0, mpilib)
 
-    if uses_kokkos(case) and comp_interface == "mct":
+    if uses_kokkos(case):
```
rljacob, line 758: If you just remove it, does that break something in CESM?  Maybe we need "comp_interface="mct OR moab" "

sjsprecious: Thanks @jgfouca . Rob is correct that removing this will break the build of Kokkos through the CESM workflow. It is better to use Rob's suggested check or `comp_interface != "nuopc"` (CESM3 only uses NUOPC coupler and I assume E3SM won't use NUOPC in the future).

sjsprecious (changes requested):

CIME/build.py
jgfouca: @sjsprecious , that's a good idea. I have changed the code to check for nuopc.

jedwards4b (approved):
LGTM

rljacob (approved):

CIME/build.py
jgfouca: That's a good thought. I think a simple workaround is fine for now.

Issue #4451, opened 2023-06-28
Need a complete set of files to test st_archive


kdraeder:
DART creates quite a few additional file types, which requires changes
to the st_archive modules, even when they are named according to
the [CESM naming convention](https://www.cesm.ucar.edu/models/cesm2/naming-conventions#modelOutputFilenames).

Before opening a PR I need a test set of output files, which is minimal, 
but tests all of archiving needs of all of the components, and the corresponding DART files.
Is there a existing CESM file set for that?
I've found references to test files that have 1979-01-01-00000
in their names, but have not found the files.
Do they need to be generated by running CESM in some predefined way?
A related issue is whether the st_archive (test) can handle multidriver cases.
That's essential for DART runs.
#1931 discusses this, and has been (partly) addressed in CESM2.3.

Here is what I'm guessing needs to happen to make CESM2.3 archive DART's files.

1. Make a components/dart/cime_config directory.  "dart" is an instance of the generic component "esp".
2. Extract the "esp" section from CESM2.1's cime/config/cesm/config_archive.xml into a new components/dart/cime_config/config_archive.xml.
3. Expand cmeps/cime_config/config_archive.xml to handle all cmeps history file extensions, including auxiliary files (hx),  and gzipped files (for disk management in large DART runs).  It looks like the regular expression could be made more focused, since other components handle their history files.
4. Some components, which have DART interfaces, don't handle the '.e.' file type.
5. While I'm fixing those, it could be worthwhile to add .e. to the components that don't yet have DART interfaces.

I'm also looking for the right place to enable the archiving of a DART input.nml file.
This was done in  CESM2.1:cime/scripts/lib/CIME/case/case_st_archive.py,
but the place that worked in there may have been moved to the
CESM2.3: cime/CIME/XML/archive_base.py:get_all_hist_files

The [exampleStaFilenames](https://www.cesm.ucar.edu/models/cesm2/naming-conventions#exampleStaFilenames) web page  could be updated with some of these files as an example of esp files.

A question, which may be a separate issue;
when running CAM+DART assimilations we want to archive the initial files with the restart set.
The code (components/cam/cime_config/config_archive.xml) puts them in cam/hist,
which is inconsistent with the exampleStaFilenames which says that they go to the restart directory.
Is there a compelling reason for them to go to the hist directory?

I'm probably missing, or misunderstanding, something about the current st_archive.

jedwards4b:
[...]

PR #4788, opened 2025-04-22
Modified code to make st-archive handle DART files

kdraeder:
Doing data assimilation with CESM+DART generates many new types of output files,
which st-archive should archive, assuming they are named according to CESM conventions.  
Many of the new files will be handled by the esp component code, but others are 
more closely associated  with the geophysical components and should be archived 
with their files. We've chosen new file names in such a way that this can be done
with very small changes to
   $component/cime_config/config_archive.xml
   cime/CIME/XML/archive_base.py
A second small change is to make st-archive handle compressed files (*.gz).
This has no effect on model calculations or existing output.

Test suite: 
cesm3.0_alphabranch (cesm3_0_beta03-121-g0a50f32) in a B compset 
(HISTC_CAM70%LT_CLM60%BGC-CROP_CICE_MOM6_MOSART_SGLC_SWAV,
a%ne30np4.pg3_l%ne30np4.pg3_oi%tx2_3v2_r%r05_g%null_w%null_z%null_m%tx2_3v2).
The test used the tags listed in .gitmodules (fxtag = cime6.1.87) 
with no development branches checked out. 
It was set up to create the greatest variety of files possible (I probably missed some).  
The interface to DART is not complete, so files that will result from running DART 
were created artificially, since the contents don't matter to st_archive (except for 
the restart files which contain the names of restart history files).   The script used to create
these files is attached.
[dart_files2arch.csh.txt](https://github.com/user-attachments/files/19855993/dart_files2arch.csh.txt)
The test consisted of running a 24 hour forecast, adding the DART files to RUNDIR, 
running st_archive, and running a second day to check the restart capability.

The changes were added to a feature branch based on this component's development branch.

Test namelist changes: None
Test status: bit for bit

Partially fixes [[CIME Github issue #4451]](https://github.com/ESMCI/cime/issues/4451)

User interface changes?: None

Update gh-pages html (Y/N)?: None yet.  I don't know if they're needed.
The capability is increased, but it requires nothing from users.

### DIFF
--- a/CIME/XML/archive_base.py
+++ b/CIME/XML/archive_base.py
@@ -139,7 +139,7 @@ class ArchiveBase(GenericXML):
         for ext in extensions:
             if ext.endswith("$") and has_suffix:
                 ext = ext[:-1]
-            string = model + r"\d?_?(\d{4})?\." + ext
+            string = model + r"\d?_?(\d{4})?(_d\d{2})?\." + ext
             if has_suffix:
                 if not suffix in string:
                     string += r"\." + suffix + "$"


jasonb5:
@kdraeder @jedbrown Would adding another/modifying `hist_file_extension` in https://github.com/ESMCI/ccs_config_cesm/blob/main/config_archive.xml support this?

kdraeder:
@jasonb5 Thanks for looking into this so quickly!
DART creates files with names like ${CASE}.cam_0001_d01.e.analysis.1850-01-02-00000.nc.
These were ignored by st_archive (archive_base.py) because the _d01 
was not an acceptable part of the  component name.  
It looks to me like the xml <hist_file_extension> entries only describe 
what st_archive should do after the name part has been found to be acceptable.

An alternative solution is to move the 'domain' description to a different part of the file name,
but the other places don't seem to be a good fit.

jasonb5:
@kdraeder ah right, you are correct that wouldn't work.

I'm in favor for adding a feature to make this customizable, rather than modifying it for this specific case. This way if other components decide to use a different naming it will only require a config change and not another PR.

@jedbrown @jgfouca We could add something like `hist_unique_name` and default it to the existing regex. We could also do this for restart files so we can remove this component specific code (https://github.com/ESMCI/cime/blob/2343dba8a347a57a41f7bf9e650ae98104209554/CIME/case/case_st_archive.py#L576-L592).

jedwards4b:
poor jedbrown gets hit again!   I'm all for your suggested change, but I don't have time to address it right now - do you?

kdraeder:
The [MOM_interface issue #238](https://github.com/ESCOMP/MOM_interface/issues/238)
might be related enough to include @alperaltuntas view in this decision.

jasonb5:
@jedwards4b Opps. Yea I can work on this feature.

kdraeder:
The [MOM_interface issue #238](https://github.com/ESCOMP/MOM_interface/issues/238)
might be related enough to include @alperaltuntas view in this decision.

jedwards4b (approved):

PR #4789, opened 2025-04-23
bless_test_results: fix stale file use case

jgfouca:
Sometimes, files reside in the baseline area that are no longer relevant. Btr needs to remove these files when blessing so that the "no original counterpart" error goes away.

Test suite: by hand
Test baseline:
Test namelist changes:
Test status: bit for bit

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/bless_test_results.py
+++ b/CIME/bless_test_results.py
@@ -187,7 +187,9 @@ def bless_history(test_name, case, baseline_name, baseline_root, report_only, fo
                         file_to_remove = (
                             line.split(NO_ORIGINAL)[0].split()[-1].strip("'")
                         )
-                        logger.info(f"Remove stale baseline file {file_to_remove}")
+                        logger.info(
+                            "Removing stale baseline file {}".format(file_to_remove)
+                        )
                         os.remove(os.path.join(baseline_full_dir, file_to_remove))
 
                 gen_result, gen_comments = generate_baseline(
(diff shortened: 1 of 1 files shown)


CIME/bless_test_results.py
```diff
@@ -180,6 +180,16 @@ def bless_history(test_name, case, baseline_name, baseline_root, report_only, fo
...
                 force or input("Update this diff (y/n)? ").upper() in ["Y", "YES"]
             ):
+                # Sometimes, in order to get things passing, files have to be removed
+                # from the baseline area.
+                for line in cmp_comments.splitlines():
+                    if NO_ORIGINAL in line:
+                        file_to_remove = (
+                            line.split(NO_ORIGINAL)[0].split()[-1].strip("'")
+                        )
+                        logger.info(f"Remove stale baseline file {file_to_remove}")
```
jedwards4b, line 190: You can't use f style formatting here.

jgfouca: Why not?

jgfouca: Oh, looks like the CI version of pylint doesn't like it. The one I used had no issue with it, but I can change it.

gold2718: Is this an optimization issue?
https://docs.python.org/3/howto/logging.html#optimization

jasonb5 (approved):
LGTM

PR #4795, opened 2025-05-09
Updates container and workflow

jasonb5:
- Updates and simplifies the container
- Moves model setup from container to workflow
- Adds vscode tasks for development activities


.github/workflows/testing.yml
```diff
@@ -140,7 +163,13 @@ jobs:
...
       matrix:
-        model: ["e3sm", "cesm"]
+        model:
+          - name: e3sm
+            repo: E3SM-Project/E3SM
+            branch: master
+          - name: cesm
+            repo: ESCOMP/CESM
+            branch: cesm3.0-alphabranch
         driver: ["mct", "nuopc"]
```
jedwards4b, line 173: We can probably restructure a little here to avoid this unnecessary loop - e3sm only uses mct and cesm only uses nuopc.

jedwards4b (approved):
This is great - only one suggestion.   Thank you so much for your work to improve testing.

jasonb5:
Had to remove the full system tests, which were taking ~1 hour.

PR #4803, opened 2025-05-22
Revert "Output GPTL timing info when using NUOPC driver"

jedwards4b:
Reverts ESMCI/cime#4791


sjsprecious (approved):
Thanks @jedwards4b for reverting this.

The `MODEL_THETA_L` and `HOMME_ENABLE_COMPOSE` variables are not applied by default when building the GPTL library. Even if I add them manually, no ESMF or GPTL timing file is generated when using the `theta-l` dycore in CAM (both ESMF and GPTL timing files are generated correctly when using the `theta-l_kokkos` dycore in CAM). Thus I think it is better to manually revise this block as needed for different dycores.

PR #4804, opened 2025-06-03
Fix failing tests

jasonb5:
Fixes failing tests introduced by syncing https://github.com/ESMCI/cime/pull/4749 to master.


jasonb5:
This may not longer be needed after testing update.

PR #4809, opened 2025-06-20
support multiple config directories

jedwards4b:
Allow for a config directory in .cesm_v_ or .e3sm_v_ instead of .cime this is to allow some version compatibility while being backward compatible with the original .cime

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:


billsacks:
From discussion on cime call: We like the idea of this, but aren't sure about the details of the implementation - e.g., how to determine the version. Until we have specific use cases, we're going to close this for now.

Issue #4792, opened 2025-05-06
Update config_archive.xml/case_st_archive.py


jasonb5:
The regex patterns for matching history/restart files for archiving are partial hard-coded. Updates require PRs like #4788 to be made when the naming changes. We should update these patterns to be fully definable in `config_archive.xml`. 

- https://github.com/ESMCI/cime/blob/2343dba8a347a57a41f7bf9e650ae98104209554/CIME/XML/archive_base.py#L142
- https://github.com/ESMCI/cime/blob/2343dba8a347a57a41f7bf9e650ae98104209554/CIME/XML/archive_base.py#L253

This would also be a good opportunity to refactor `case_st_archive.py`, some of this code belongs in `CIME/XML/archive_base.py` and `CIME/XML/archive.py`. There's also a bit of component specific code that should be moved out into `config_archive.xml`.

- https://github.com/ESMCI/cime/blob/c63fd0b46486ec6d85d7138e92c54b58d9149a14/CIME/case/case_st_archive.py#L320-L326
- https://github.com/ESMCI/cime/blob/c63fd0b46486ec6d85d7138e92c54b58d9149a14/CIME/case/case_st_archive.py#L576-L591
- https://github.com/ESMCI/cime/blob/c63fd0b46486ec6d85d7138e92c54b58d9149a14/CIME/case/case_st_archive.py#L686-L697

jasonb5:
Tracking this under #5000 (Slice 4: Case decomposition), part of #4950.
`case/case_st_archive.py` (1395 lines, ~20 procedural functions) is the
largest procedural module and is on the Slice 4 decomposition list.

PR #4814, opened 2025-07-01
archiver: Make entire special regex optional

jgfouca:
The problem with the original line of code:
`string = model + r"\d?_?(\d{4})?(_d\d{2})?\." + ext`

... is that the `.` at the end is not optional, which will mean files like:

`ERS_Ln3.ne4_ne4.F2010-SCREAMv1.mappy_gnu.20250701_125457_1nse3o.scream.h.AVERAGE.nsteps_x3.0001-01-01-00000.nc`

Will not match regexes like

`<hist_file_extension>.*\.h\.(?!rhist\.).*\.nc$</hist_file_extension>`

Because there's only one `.` between the model (scream) and the hist extension `h`.

I think, since the added regex is intended to be optional, is to make all of it optional via another `?`.

Also, add wait_for_tests to the list of tools that gets copied to the case.

Test suite: by-hand
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

User interface changes?:

Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/archive_base.py
+++ b/CIME/XML/archive_base.py
@@ -139,21 +139,28 @@ class ArchiveBase(GenericXML):
         for ext in extensions:
             if ext.endswith("$") and has_suffix:
                 ext = ext[:-1]
-            string = model + r"\d?_?(\d{4})?(_d\d{2})?\.?" + ext
+
+            # Build the regex that will identify history files
+            re_string = model  # We expect the model name to be in the filename
+            re_string += (
+                r"\d?_?(\d{4})?(_d\d{2})?"  # Need this for potential DART files?
+            )
+            re_string += r"\." + ext  # We expect a dot to precede the extension
+
             if has_suffix:
-                if not suffix in string:
-                    string += r"\." + suffix + "$"
+                if not suffix in re_string:
+                    re_string += r"\." + suffix + "$"
 
-                if not string.endswith("$"):
-                    string += "$"
+                if not re_string.endswith("$"):
+                    re_string += "$"
 
-            logger.debug("Regex is {}".format(string))
-            pfile = re.compile(string)
+            logger.debug("Regex is {}".format(re_string))
+            regex = re.compile(re_string)
             hist_files.extend(
                 [
                     f
                     for f in os.listdir(from_dir)
-                    if pfile.search(f)
+                    if regex.search(f)
                     and (
                         (f.startswith(casename) or f.startswith(model))
                         and not f.endswith("cprnc.out")
--- a/CIME/XML/archive_base.py
+++ b/CIME/XML/archive_base.py
@@ -139,7 +139,7 @@ class ArchiveBase(GenericXML):
         for ext in extensions:
             if ext.endswith("$") and has_suffix:
                 ext = ext[:-1]
-            string = model + r"\d?_?(\d{4})?(_d\d{2})?\." + ext
+            string = model + r"\d?_?(\d{4})?(_d\d{2})?\.?" + ext
             if has_suffix:
                 if not suffix in string:
                     string += r"\." + suffix + "$"
--- a/CIME/case/case.py
+++ b/CIME/case/case.py
@@ -1696,6 +1696,7 @@ class Case(object):
             os.path.join(toolsdir, "xmlchange"),
             os.path.join(toolsdir, "xmlquery"),
             os.path.join(toolsdir, "pelayout"),
+            os.path.join(toolsdir, "wait_for_tests"),
         )
         try:
             for exefile in exefiles:


jedwards4b (approved):

jasonb5 (approved):
LGTM

kdraeder:
This change will fix the scream problem that prompted the issue,
and it will allow history files that adhere to the (outdated?) [CESM2 file naming convention](https://www.cesm.ucar.edu/models/cesm2/naming-conventions) to match.

But I think that it also will match files which have no '.' separator between the model string 
and the extension  string ("file type" in the CESM2 link).  Do we want to count on no model ever doing that?
In CESM CISM seems to be the only component that puts characters in front of the '.h'., 
but its filenames do have a '.' separator between model and extention.
Is it a common thing to do in scream (E3SM?), and do all the filenames have the '.' separator?
Is the separator something that CESM3 (cime #?) will enforce?

jgfouca:
@kdraeder , it looks like almost all the E3SM components do not explicitly state the `.` in the extension. For example, I see lots of entries like this:

`<hist_file_extension>h\d*.*\.nc$</hist_file_extension>`

So, maybe the real issue is that scream did not follow this convention. Let me re think this a bit.

rljacob:
Can we move to a system where the file name pattern for a model is stored with the model in its cime_config dir?  Then these kind of details wouldn't have to live in CIME itself.

jgfouca:
OK, after a bit a thinking, my opinion is that it was actually the scream extension setting that was wrong, and not the CIME code. I did do a bit of refactoring to make the code clearer as to what parts of the regex mangling were needed for optional DART stuff and what part is required (the dot preceding the extension).

CIME/XML/archive_base.py
```diff
@@ -139,21 +139,28 @@ def get_all_hist_files(self, casename, model, from_dir, suffix="", ref_case=None
         for ext in extensions:
             if ext.endswith("$") and has_suffix:
                 ext = ext[:-1]
-            string = model + r"\d?_?(\d{4})?(_d\d{2})?\." + ext
+
+            # Build the regex that will identify history files
+            re_string = model  # We expect the model name to be in the filename
+            re_string += (
+                r"\d?_?(\d{4})?(_d\d{2})?"  # Need this for potential DART files?
```
jgfouca, line 146: It's not entirely clear to me why this weird digit stuff has to be injected, but this has been in there in one form or another for a very long time.

jgfouca:
@rljacob , I agree. The current situation is a bit of a mess. We currently have (mostly) what you describe in a model's config_archive.xml. The problem is there's a bit of magic added to the regex by CIME that you (model developer) cannot see and cannot opt-out of. The situation with restart files is even worse as there appear to be hardcoded special-cases in CIME itself for mpas, mali, and nemo, in addition to general magic.

jgfouca:
Ah crap, now scream is back to not finding history files. The issue is that scream wants to have hist files with names like

`stuff.scream.anything_might_here.h.stuff.nc`

And the current magic injections make this impossible.

jedwards4b:
I wonder if we can head in the direction that Rob suggests incrementally - that is if the model has a complete regex in its cime_config/config_archive.xml then the one in cime is ignored?

jgfouca:
@jedwards4b , yeah, or we could have a version=3.0 mode that doesn't do any magic.

bartgol:
> I wonder if we can head in the direction that Rob suggests incrementally - that is if the model has a complete regex in its cime_config/config_archive.xml then the one in cime is ignored?

I agree. But I don't really understand why CIME does any regex manipulation to start with. My naive idea would be that "the user gave a regex, let's use it", but maybe there are additional manipulations needed to distinguish files in, say, ERS runs?

But yeah, I agree that it seems reasonable to _just_ use the user-provided regex...

jasonb5:
I agree that the regex should live with the model/component. I've also identified other areas with baked in regex (#4792), that should refactored and move from the CIME code.

I think it's a fine option for CIME to have sane defaults and allow the model to override.

CIME/XML/archive_base.py
kdraeder: The first \d is for models/components like clm2, I believe.
The _ and 4 digits are the instance number, which are needed for multi-instance runs (like DART).
The _d and 2 digits I think are for 'domains' of a model, which I haven't had to deal with (CAM and CLM).
All of these are considered to be part of the "model" description.
I like having them be strictly defined, to prevent mass confusion and impossible to decipher regexes.

The next pieces, provided by config_archive.xml, are file "types" or extensions.
I'm in favor of regularity in the extensions too, but I realize some accommodation of components may be needed.

PR #4815, opened 2025-07-01
Fix resolving reference variables in mixed subgroups

### LINKED ISSUE #4646: xmlquery issue when values are set to match another variable
I'm getting this error with --listall (or when query the HIST_N variable):

```bash
$ ./xmlquery HIST_N
ERROR: Entry HIST_N was listed as type int but value '$STOP_N' is not valid int
$ ./xmlquery --listall
ERROR: Entry HIST_N was listed as type int but value '$STOP_N' is not valid int
```

Did something change in CIME related to xmlquery? I don't think we recently changed the default for HIST_N, so I'm guessing this is a new thing? I'm using the tag cime6.0.226 (438850986).

jasonb5:
Querying a variable whose value is a reference that exists
in a different `subgroup` would fail with `ERROR: Entry
HIST_N was listed as type int but value '$STOP_N' is not
valid int`. 

This PR changes the behavior so a variable that cannot
be resolved within the current subgroup will try to resolve
from any subgroup. Finally if still not resolvable an exception
is raised.

Test baseline:
Test namelist changes:

Fixes #4646 

User interface changes?:
Update gh-pages html (Y/N)?:

### DIFF
--- a/CIME/XML/entry_id.py
+++ b/CIME/XML/entry_id.py
@@ -394,7 +394,7 @@ class EntryID(GenericXML):
             val = self.get_default_value(node)
 
         if resolved:
-            val = self.get_resolved_value(val)
+            val = self.get_resolved_value(val, subgroup=subgroup)
 
         return val
 
--- a/CIME/XML/entry_id.py
+++ b/CIME/XML/entry_id.py
@@ -313,7 +313,7 @@ class EntryID(GenericXML):
         )
         node = self.get_optional_child("entry", {"id": vid}, root=root)
         if node is not None:
-            val = self._set_value(node, value, vid, subgroup, ignore_type)
+            val = self._set_value(node, value, vid, None, ignore_type)
         return val
 
     def get_values(self, vid, attribute=None, resolved=True, subgroup=None):
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -100,9 +100,9 @@ class EnvBatch(EnvBase):
 
     def get_type_info(self, vid):
         gnodes = self.get_children("group")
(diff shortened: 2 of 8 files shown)


jasonb5:
@jedwards4b @jgfouca The original solution discussed for this issue was to "namespace" the value e.g. `HIST_N=<subgroup>::STOP_N`. I'm wondering if this complexity is required? The change I have right now simply falls back to resolving the reference without a subgroup. Is there a scenario when a variable would be defined more than twice?

jedwards4b (changes requested):
I would suggest adding a test for a variable which is in a subgroup, for example 
JOB_WALLCLOCK_TIME 
```
 ./xmlquery JOB_WALLCLOCK_TIME
Results in group case.cupid
	JOB_WALLCLOCK_TIME: 48:00:00
	JOB_WALLCLOCK_TIME: 00:20:00
Results in group case.test
	JOB_WALLCLOCK_TIME: 48:00:00

```

jgfouca (approved):

jasonb5:
Looks like a subgroup reference would be needed for the following example.

``` shell
Results in group case.cupid
	JOB_WALLCLOCK_TIME: $JOB_WALLCLOCK_TIME@case.cupid
	JOB_WALLCLOCK_TIME: 00:20:00
Results in group case.test
	JOB_WALLCLOCK_TIME: 48:00:00
```

@jedwards4b @jgfouca Any recommendations for the delimiter? Originally it was suggested to use `::` like c++, another alternative could be `@`.

jgfouca:
@jasonb5 , I think `::` would be best. We use that in Eamxx tools already.

PR #4820, opened 2025-07-11
test_scheduler.py: Error if Izumi nag without debug

samsrabin:
There is a long-standing bug, _extremely_ difficult to diagnose, where building CESM will fail using the nag compiler on Izumi if you didn't enable debug mode. This PR makes it so that tests with such a configuration will fail with a useful error message.

Test suite:
Test baseline:
Test namelist changes:

Fixes [CIME Github issue #]

**User interface changes?:** No

**Update gh-pages html (Y/N)?:** No


samsrabin:
It looks like there is at least one CESM test that uses izumi nag without debug. From `testlist_cam.xml`:
```xml
  <test compset="QPC6" grid="ne5pg3_ne5pg3_mg37" name="ERP_Ln9" testmods="cam/outfrq9s_clubbmf">
    <machines>
      <machine name="izumi" compiler="nag" category="aux_cam"/>
    </machines>
    <options>
      <option name="wallclock">00:10:00</option>
    </options>
  </test>
```

I'm not seeing this test in any expected failure list, so maybe it'd be better to do this check in CTSM instead. @ekluzek

ekluzek (commented):
The change looks good. It looks like it makes it easy to see what to do. But, it is really specific to a specific CESM case.

PR #4831, opened 2025-08-13
correct math in rest_dir calculation

jedwards4b:
## Description
Incorrect math in err.py leads to occasional crash.

- Closes #<ISSUE_NUMBER_HERE>

## Checklist
- [X] I have added tests that excerise my feature/fix and existing tests continue to pass
- [X] I have made corresponding additions and changes to the documentation

### DIFF
--- a/CIME/SystemTests/err.py
+++ b/CIME/SystemTests/err.py
@@ -39,6 +39,7 @@ class ERR(RestartTest):
         rest_root = os.path.abspath(os.path.join(dout_s_root, "rest"))
         restart_list = ls_sorted_by_fname(rest_root)
         rest_cnt = len(restart_list)
+        expect(rest_cnt >= 1, "No restart files found in {}".format(rest_root))
         rest_dir = restart_list[max(0, rest_cnt // 2)]
         self._case.restore_from_archive(rest_dir=os.path.join(rest_root, rest_dir))
         logger.info("Setting restart to {}".format(rest_dir))


CIME/SystemTests/err.py
```diff
@@ -39,9 +39,9 @@ def _case_two_custom_prerun_action(self):
         rest_root = os.path.abspath(os.path.join(dout_s_root, "rest"))
         restart_list = ls_sorted_by_fname(rest_root)
         rest_cnt = len(restart_list)
-        expect(rest_cnt >= 1, "No restart files found in {}".format(rest_root))
```
ekluzek, line 42: Is this line meant to be removed? Seems like this should remain, unless it wasn't actually working anyway.

CIME/SystemTests/err.py
```diff
@@ -39,9 +39,9 @@ def _case_two_custom_prerun_action(self):
...
-        rest_dir = restart_list[max(1, rest_cnt // 2)]
+        rest_dir = restart_list[max(0, rest_cnt // 2)]
```
ekluzek, line 43: This makes sense to make sure it at least correctly gets the one restart directory if that's all that exists.

CIME/SystemTests/err.py
```diff
@@ -39,9 +39,9 @@ def _case_two_custom_prerun_action(self):
...
         self._case.restore_from_archive(rest_dir=os.path.join(rest_root, rest_dir))
+        logger.info("Setting restart to {}".format(rest_dir))
```
ekluzek, line 45: Yes, adding this would be helpful in debugging the current ERR test we've been working on. So nice addition here.

ekluzek (approved):
Changes make sense to me. I add a couple comments that don't need to be handled. Just confirming behavior, and then praising the addition of something to the log.

I do ask a question about the expect line that's being removed, because it looks to me like it shouldn't be. But, I might be missing something.

Pushing approve, so you don't have to get back to me.

PR #4834, opened 2025-08-19
Change Kokkos sharedlib to ekat

jgfouca:
## Description

E3SM wants to use Kokkos through EKAT, so we want EKAT to be the sharedlib that gets built by CIME, not Kokkos directly.

I don't know if this will impact anyone outside of E3SM, I don't think it will..

It seems like it might be better to have the host model have a config somewhere that describes the sharedlibs it needs rather than having this hardcoded in CIME/build.py.

## Checklist
- [ x ] My code follows the style guidlines of this proejct (black formatting)
- [ x ] I have performed a self-review of my own code
- [ x ] My changes generate no new warnings
- [ x ] I have added tests that excerise my feature/fix and existing tests continue to pass
- [ x ] I have commented my code, particularly in hard-to-understand areas
- [ x ] I have made corresponding additions and changes to the documentation

### DIFF
--- a/CIME/build.py
+++ b/CIME/build.py
@@ -746,6 +746,9 @@ def _build_libraries(
             cpl_in_complist = True
     if ufs_driver:
         logger.info("UFS_DRIVER is set to {}".format(ufs_driver))
+
+    # This is a bit hacky. The host model should define whatever
+    # shared libs it might need.
     if ufs_driver and ufs_driver == "nems" and not cpl_in_complist:
         libs = []
     elif case.get_value("MODEL") == "cesm":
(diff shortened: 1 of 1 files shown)


jedwards4b (approved):
It makes sense that this would be defined in xml instead of hardcoded.

sjsprecious:
> It makes sense that this would be defined in xml instead of hardcoded.

@jedwards4b are you suggesting something like:
```
./xmlchange BUILD_PIO=TRUE
./xmlchange BUILD_CSM_SHARE=TRUE
./xmlchange BUILD_KOKKOS=TRUE
...
```
to decide what libs should be built rather than using the `MODEL` or `comp_interface` here?

jedwards4b:
No, it would be something like a new variable in the driver config_component.xml file, say
CASE_SUPPORT_LIBRARIES and this variable would be a list of strings naming support libraries to 
be built with the case, so for example CASE_SUPPORT_LIBRARIES="gptl","pio2","csm_share","FMS","Kokkos"
this would replace the lists currently hard coded [here.](https://github.com/ESMCI/cime/blob/master/CIME/build.py#L752)

PR #4836, opened 2025-08-20
moves the lib list to the driver buildnml with backward compatibility

### LINKED ISSUE #4835: move support libraries to driver
I'm thinking that the variable would be populated by the driver in it's buildnml script - moving most of the logic from this part of build.py to there.  Does this make sense?

jedwards4b:
## Description
Proposed change to handle list of support libraries in the driver rather than in cime.

- Closes #4835 

## Checklist
- [X] I have added tests that excerise my feature/fix and existing tests continue to pass
- [X] I have made corresponding additions and changes to the documentation


sjsprecious (approved):
Thanks @jedwards4b for working on this PR. I could confirm that it builds and runs the Kokkos lib/kernels successfully in CAM, together with your other PR (https://github.com/ESCOMP/CMEPS/pull/584).

CIME/build.py
```diff
@@ -738,44 +738,48 @@ def _build_libraries(
...
-        libs.append("ekat")
+        if uses_kokkos(case) and comp_interface != "nuopc":
+            libs.append("ekat")
 
-    # Build shared code of CDEPS nuopc data models
-    build_script = {}
-    if comp_interface == "nuopc" and (not ufs_driver or ufs_driver != "nems"):
-        libs.append("CDEPS")
+        # Build shared code of CDEPS nuopc data models
+        build_script = {}
```
bartgol: I think this line is pointless, no? It's already set before the if clause.

jedwards4b: Looks like a merge error, I'll clean it up - thanks.

CIME/build.py
```diff
@@ -738,44 +738,47 @@ def _build_libraries(
...
+        libs.append("FTorch")
 
-    if mpilib == "mpi-serial":
-        libs.insert(0, mpilib)
+        if mpilib == "mpi-serial":
+            libs.insert(0, mpilib)
 
-    if uses_kokkos(case) and comp_interface != "nuopc":
-        libs.append("ekat")
+        if uses_kokkos(case) and comp_interface != "nuopc":
```
sjsprecious, line 776: @bartgol if I understood your suggestion from https://github.com/ESMCI/cime/issues/4837 correctly, the `uses_kokkos` function can be deprecated since the host model will define the shared libs in its config_component.xml?

jedwards4b: That's correct, this code is only still here for backward compatibility.

bartgol: Yeah, that was my idea. I don't want to inject knowledge about host model tpls into CIME. All CIME should care about is a list of libs names, and where to find their build scripts.

jedwards4b:
An issue that I think we still need to resolve is that some support libraries depend on others and
when all the libraries are defined in one place as is currently done it is easy to maintain an order so
that libraries are built in the correct order.   How do we do this if several different buildnml scripts contribute
to the overall variable?   One thought is that each buildnml include the dependencies specifying the order locally,
then in build.py in cime `use libs = list(dict.fromkeys(libs))`  to get an ordered list of unique libraries.

jedwards4b:
@bartgol I have added https://github.com/ESCOMP/MOM_interface/pull/276 to show what this looks like in a particular component and have tested this together with the CMEPS and cime changes.

PR #4853, opened 2025-09-22
compare_namelists: Add support for YAML namelist files

jgfouca:
## Description

Components are beginning to use yaml files, instead of nml files, as inputs. This PR adds support for YAML files to compare_namelists so that CIME will detect changes to YAML input files.

## Checklist

### DIFF
--- a/CIME/compare_namelists.py
+++ b/CIME/compare_namelists.py
@@ -680,7 +680,7 @@ def _compare_namelists(gold_namelists, comp_namelists, case):
 ###############################################################################
 def normalized_dict_compare(keyname, data1, data2, case):
     ###############################################################################
-    """
+    r"""
     Recursively compare contents of dicts, normalize values (but not keys)
 
     >>> data1 = {'averaging_type': 'average', 'fields': {'physics_gll': {'field_names': ['T_mid', 'tracers', 'horiz_winds', 'p_int', 'pseudo_density', 'p_mid']}}, 'filename_prefix': 'ERS_Ln9.ne4_ne4.F2000-SCREAMv1-AQP1.mappy_gnu.eamxx-output-preset-2.G.20250922_110259_eeywbj.scream.phys.h', 'output_control': {'frequency': 9, 'frequency_units': 'nsteps', 'skip_t0_output': False}, 'skip_t0_output': False}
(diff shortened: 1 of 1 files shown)


CIME/compare_namelists.py
```diff
@@ -677,6 +677,140 @@ def _compare_namelists(gold_namelists, comp_namelists, case):
...
+    else:
+        if data1 != data2:
+            comments += (
+                f"{keyname} had mismatched values, baseline={data1}, case={data2}\n"
+            )
+
+    return comments
+
+
+class Array(list):
```
bartgol, line 749: For reference: these user-defined classes are used in EAMxx to allow having yaml files containing stuff like `my_ints: []` (i.e., an empty list) and still have our generic parser be able to store in a container where the inner type has the proper type (e.g., with my_ints being a `std::vector<int>`). The `!ints` tag (and similar others) allows our parsing utilities to store `my_ints` in a `std::vector<int>` inside our data containers structures.

I understand this is a bit of an injection of downstream knowledge. We are trying to think of alternative ways for our generic yaml parser routines to parse files with empty lists without assuming the wrong type (e.g., string).

If we manage to do away with these user-defined tags, we could revisit this utility.

CIME/compare_namelists.py
```diff
@@ -686,9 +820,15 @@ def compare_namelist_files(gold_file, compare_file, case=None):
     expect(os.path.exists(gold_file), "File not found: {}".format(gold_file))
     expect(os.path.exists(compare_file), "File not found: {}".format(compare_file))
 
-    gold_namelists = _parse_namelists(open(gold_file, "r").readlines(), gold_file)
-    comp_namelists = _parse_namelists(open(compare_file, "r").readlines(), compare_file)
-    comments = _compare_namelists(gold_namelists, comp_namelists, case)
+    if gold_file.endswith(".yaml"):
```
bartgol: I think some folks also use `.yml` as extension, so maybe we should check both?

bartgol (approved):

CIME/compare_namelists.py
jgfouca: Good idea. Done

CIME/compare_namelists.py
```diff
@@ -677,6 +677,140 @@ def _compare_namelists(gold_namelists, comp_namelists, case):
...
+
+###############################################################################
+def _compare_yamls(gold_file, compare_file, case):
+    ###############################################################################
+    """
+    Compare contents of two YAML files
+    """
+    # If we have cases producing YAML files, we can probably safely assume we have
+    # pyyaml available to us.
+    import yaml
```
jedwards4b, line 798: Perhaps
```
try:
  import yaml 
except: 
    ModuleLoadError("pyyaml module not found")
```

bartgol: Yeah, this is prob a good idea. Since pyyaml is not a def py pkg (*), we could print a useful error msg explaining the user how to install it, and why we need it.

(*) It is a bit underwhelming that py _does_ ship XML parsing support with its std library but does not ship any yaml support (nor does it plan to in the near future). It is not a niche language, so it'd be nice if we did not have to do this kind of checks...

jgfouca: Hey @jedwards4b , sorry, I clicked merge before I saw your review. Python will automatically raise a `ModuleNotFoundError` if the import fails, so I'm wondering why you want to handle it manually?

jedwards4b: no worries, just want to make sure that the error and the error location are clear.

PR #4877, opened 2025-10-13
Allow namelist variables whose names vary from what's in namelist_definition

billsacks:
Previously, any namelist variable needed to appear in the namelist_definition xml file. With the changes here, the namelist_definition file can contain a generic variable name that gives metadata for one or more final namelist variables with different names. This is done by adding multiple copies of a namelist_definition entry, one for each final name.

The use case for this is having configuration variables whose names are not known ahead of time, but instead are determined at build-namelist time based on some xml variables. Specifically, I'm planning to use this to support the addition of arbitrary water tracers, each of which will have one or more configuration variables whose name contains the name of that water tracer, where the list of water tracer names is set in an xml variable. The type and default value will be the same for each water tracer, so we can take this metadata from a single, generic entry in the namelist_definition file, but then we need to support an add_default call for each individual, dynamically-named variable.

Specifically, in CMEPS, I plan to have a namelist_definition entry like this:

```xml
  <entry id="wtracer_initial_ratio" multi_variable_entry="true">
    <type>real</type>
    <category>water_tracers</category>
    <group>ALLCOMP_attributes</group>
    <desc>
      Initial tracer ratio for each water tracer in this simulation.

      Note that there will be a separate entry for each water tracer, with a name like
      wtracer_initial_ratio_MyWaterTracer.
    </desc>
    <values>
      <value>0.0</value>
    </values>
  </entry>
```

But there won't actually be a `wtracer_initial_ratio` variable in the nuopc.runconfig file. Instead, for each water tracer in this simulation (defined via a list in an XML variable in env_run.xml), the nuopc.runconfig file will have a variable like `wtracer_initial_ratio_MyWaterTracer`. This will be accomplished in the CMEPS buildnml file by setting up a dict like:

```python
multi_variable_mappings = {"wtracer_initial_ratio": ["wtracer_initial_ratio_MyWaterTracer", "wtracer_initial_ratio_AnotherWaterTracer"]}
```

(though this will be done dynamically based on the entries in the new XML variable).

Without the changes in this PR, I believe that the main way to accomplish something like this would be to have an array of values. However, that feels more error-prone from a user perspective, and perhaps more importantly, would lead to less consistency with how I plan to set things up for water isotopes - for which we will have predefined names, following a more standard approach.

## Checklist

## Testing
I ran scripts_regression_tests on my Mac. Failures are the same as on master.


CIME/XML/namelist_definition.py
```diff
@@ -89,44 +89,80 @@ def __init__(self, infile, files=None):
         super(NamelistDefinition, self).__init__(infile, schema=schema)
 
         self._attributes = {}
-        self._entry_nodes = []
-        self._entry_ids = []
```
billsacks, line 93: Note that I have removed _entry_nodes from NamelistDefinition: this wasn't being used, and would be problematic now that there could be the same node with different names, since the name might not match the id.

I have also renamed _entry_ids to _var_names, since this no longer necessarily gives the "id" attribute but instead may give the rename.

CIME/XML/namelist_definition.py
```diff
@@ -89,44 +89,80 @@ def __init__(self, infile, files=None):
...
-    def set_node_values(self, name, node):
-        self._entry_nodes.append(node)
-        self._entry_ids.append(name)
-        self._nodes[name] = node
-        self._entry_types[name] = self._get_type(node)
-        self._valid_values[name] = self._get_valid_values(node)
-        self._group_names[name] = self.get_group_name(node)
-
-    def set_nodes(self, skip_groups=None):
+    def _set_node_values(self, names, node):
```
billsacks, line 98: Note that I have made this function private. I have changed its interface (changing the `name` argument to now be a list of names - which will often be a list of a single item, but could be more) and want to make sure it isn't being used externally (which, as far as I can tell, it isn't)… and it seems like it's not meant to be used externally.

CIME/XML/namelist_definition.py
```diff
@@ -89,44 +89,80 @@ def __init__(self, infile, files=None):
...
+            # returned default_nodes are just the nodes themselves, without the mappings
+            # in multi_variable_mappings - so users of these default_nodes examine their
+            # ids, which is problematic for a multi_variable_entry. The implication is
+            # that add_default will need to be called explicitly for the final names in a
+            # multi_variable_entry.
+            add_default = (
+                not skip_default_entry
+                and not per_stream_entry
+                and not multi_variable_entry
+            )
```
billsacks, line 145: Note that I don't return variables with `multi_variable_entry` in the list of defaults because use of these default nodes relies on the `id` matching the name. My original hope was that I could change the `id` in the node, but my sense from reading some code is that any modification of nodes assumes you want to write back to their originating file. If I'm wrong about that and it's actually possible to modify the nodes in memory without modifying the original xml file, I'd be happy to hear about that. Without that ability, I think returning these in the default list would require a change to how this default list is encoded and used, such as making it a dictionary mapping a name to a node, but it felt like that would be a pretty invasive change.

At least for my purposes, it's not a big deal that these variables don't appear in the default list: We need to know all of the names in buildnml anyway in order to add them to the multi_variable_mappings argument, so it's easy enough to loop through them and explicitly call add_default on each of them.

billsacks:
Thanks for your questions, @gold2718 .

Following some additional discussion with @nusbaume and @jiang-zhu , I have more clarity on the requirements for water isotopes/tracers, and it turns out that this feature seems relatively unimportant, so I'm closing this PR.

@gold2718 let me know if you're still interested in answers to your questions... I'm happy to try to answer them if you are, but I figure they might be irrelevant since I'm closing this.

PR #4881, opened 2025-10-18
Fix handling of namelist variables with uppercase characters

### LINKED ISSUE #4873: Cannot set variables containing uppercase letters in user_nl files
If a namelist variable contains any uppercase letters, it seems impossible to set its value in the user_nl file.

Reproducer: From `cesm3_0_alpha07d`, I created `SMS.f19_g17.X.derecho_intel` and then tried to override the variable `logFilePostFix` (which appears in the nuopc.config file). I tried setting both `logFilePostFix = ".mylog"` and `logfilepostfix = ".mylog"` in `user_nl_cpl`. Both attempts gave the same error: `ERROR: Variable 'logfilepostfix' from file 'user_nl_cpl' is not in the namelist definition.`

I am working on some other changes to `namelist_definition.py` (for which I'll submit a PR soon). I'll put in a fix for this bug on top of my other changes, since my other changes affect related code. (I came across this bug in testing those other changes; at first I thought I had broken something, but then found that I was running into the same problem even before my changes.)

billsacks:
## Description

Prior to this PR, variables with uppercase characters in the namelist definition file could not be modified via user_nl files (regardless of whether the entry in the user_nl file matched the correct case or was all lowercase).

This PR fixes this behavior so that entries in the user_nl file are case-insensitive, while preserving the case as defined in the namelist definition file in the final output. (Note that the case of variable names in any explicit `add_default` calls will still need to match the case of the variable in the namelist definition file.)

- Closes #4873 

## Checklist

## Testing performed
(1) scripts_regression_tests on my Mac

Failures were the same as on main

(2) create_test with baseline comparisons, namelist generation/comparison only (`-n` flag) for these tests in a CESM context:
```
SMS.ne30_g17.A.green_gnu
SMS.ne30pg3_t232.B1850C_LTso.green_gnu
SMS.ne30pg3_t232.B1850C_LTso.green_gnu.allactive-defaultio--drv-asyncio1node
```

(3) Manual testing of setting a mixed-case variable in the `user_nl` file: set `logFilePostFix` in `user_nl_cpl`.

Set this as `logFilePostFix`, `logfilepostfix` and `logfilepostFIX`. Confirmed that all worked, and all led to the final output in `nuopc.runconfig` matching the case in the namelist definition file (i.e., `logFilePostFix`).

### DIFF
--- a/CIME/tests/test_unit_xml_namelist_definition.py
+++ b/CIME/tests/test_unit_xml_namelist_definition.py
@@ -150,5 +150,6 @@ class TestXMLNamelistDefinition(unittest.TestCase):
         assert "MyTestVar(3)" in nml_vars
         assert nml_vars["MyTestVar(3)"] == "1.2"
 
+
 if __name__ == "__main__":
     unittest.main()


CIME/XML/namelist_definition.py
```diff
@@ -215,7 +219,7 @@ def get_value_match(self, vid, attributes=None, exact_match=True, entry_node=Non
         # NOTE(wjs, 2021-06-04) In the following call, replacement_for_none='' may not
         # actually be needed, but I'm setting it to maintain some old logic, to be safe.
         value = super(NamelistDefinition, self).get_value_match(
-            vid.lower(),
+            vid,
```
billsacks, line 222: The removal of `.lower()` here and in the call from `get_default_value` aren't needed to get things to work, but I think could be needed in principle in the future: I think these weren't needed because (1) `get_default_value` doesn't seem to ever be called, and (2) it happens that the `vid` here is irrelevant because we always pass an `entry_node` into this call.

I tested removing the setting of `entry_node` a couple of lines above. In this case, I got an error in `get_value_match` unless I removed the `lower` here, when working with mixed-case variable names.

PR #4899, opened 2025-11-18
Made ERI F1850C test work with gregorian calendar.

### LINKED ISSUE #4811: testmodifier overdefined
@kdraeder has pointed out that the test modifier _C is overdefined in cime.  In [test_scheduler.py at line 681](https://github.com/ESMCI/cime/blob/master/CIME/test_scheduler.py#L681). C# is used to specify the number of instances in a multi-coupler ensemble case.  But then [at line 884](https://github.com/ESMCI/cime/blob/master/CIME/test_scheduler.py#L884) it is used to specify that a Gregorian calendar should be used.   

I suspect that we don't actually use the _CG in any tests or this problem would have presented itself some time ago.   We can allow for this by changing line 681 to be
               ` elif case_opt.startswith("C") and not case_opt.startswith("CG"):`

kdraeder:
Several changes are needed in cam, which will have a PR with the same title.  
I'll provide the PR number after it's assigned.

This was tested using
ERI_D_C3_CG_Ld8.ne30pg3_g17.F1850C_LTso.derecho_intel.cam-outfrq9s_leapday

CIME/SystemTests/eri.py

- Changed the year increment from 2 to 4; all runs will be in leap years,
- which have Feb 29, which is the start date for outfrq9s_leapday.

CIME/test_scheduler.py

- Distinguish carefully between the _C# and _CG test modifiers so that _CG does not confuse it.

## Description
The ERI test cannot use both the _CG (gregorian calendar)  _C# (multi-instance) modifiers
because some conditionals refer only to _C.  This PR makes the scripts be more careful
in using the modifiers, and other changes to make the test work for leap days.
It should require no changes to other uses of the ERI test.
It requires a few changes in CAM, which will be handled in a CAM PR and referenced here.

- Closes #4811 

- Will lead to baseline failures for ERI tests due to a year difference (so will get failures due to "no original counterpart" / "no compare counterpart")

## Checklist
- [y] My code follows the style guidlines of this proejct (black formatting)
- [y] I have performed a self-review of my own code
- [?] My changes generate no new warnings
- [y] I have added tests that excerise my feature/fix and existing tests continue to pass
        I have not tried existing tests with this modified code.  I don't know yet which to run.
- [y] I have commented my code, particularly in hard-to-understand areas
- [y] I have made corresponding additions and changes to the documentation
        I plan to submit documentation changes in a separate PR.  They are somewhat broader than this PR.
        I'll add it to this PR if that's preferable.

### DIFF
--- a/CIME/Servers/wget.py
+++ b/CIME/Servers/wget.py
@@ -26,20 +26,21 @@ class WGET(GenericServer):
             args += "--password {} ".format(passwd)
 
         try:
-            err = run_cmd("wget {} --spider {}".format(args, address), timeout=60)[0]
-        except:
+            err, _, errstr = run_cmd(
+                "wget {} --spider {}".format(args, address), timeout=60
+            )
+        except RuntimeError:
             logger.warning(
-                "Could not connect to repo '{0}'\nThis is most likely either a proxy, or network issue .(location 1)".format(
+                "Could not connect to repo '{0}'\nThis is most likely either a proxy, or network issue .".format(
                     address
                 )
             )
             return None
-
-        if err and not "storage.neonscience.org" in address:
+        if "Connecting to " in errstr and "... connected" in errstr:
+            logger.warning("Connection established with nonzero code %s", err)
+        elif err:
             logger.warning(
-                "Could not connect to repo '{0}'\nThis is most likely either a proxy, or network issue .(location 2)".format(
-                    address
-                )
+                "Could not connect to repo '{0}'\n{1}".format(address, errstr)
             )
             return None
 
--- a/CIME/SystemTests/funit.py
+++ b/CIME/SystemTests/funit.py
@@ -35,6 +35,13 @@ class FUNIT(SystemTestsCommon):
         """
         return get_cime_root()
 
+    def get_extra_run_tests_args(self):
+        """
+        Override this to return a string containing extra command-line arguments to
+        run_tests.py
+        """
+        return ""
+
     def run_phase(self):
 
         rundir = self._case.get_value("RUNDIR")
@@ -51,8 +58,9 @@ class FUNIT(SystemTestsCommon):
                 get_cime_root(), "scripts", "fortran_unit_testing", "run_tests.py"
             )
         )
-        args = "--build-dir {} --test-spec-dir {} --machine {}".format(
-            exeroot, test_spec_dir, mach
+        extra_args = self.get_extra_run_tests_args()
+        args = "--build-dir {} --test-spec-dir {} --machine {} {}".format(
+            exeroot, test_spec_dir, mach, extra_args
         )
 
         stat = run_cmd(
--- a/CIME/XML/tests.py
+++ b/CIME/XML/tests.py
@@ -22,13 +22,20 @@ class Tests(GenericXML):
                 files = Files()
             infile = files.get_value("CONFIG_TESTS_FILE")
         GenericXML.__init__(self, infile)
-        # append any component specific config_tests.xml files
+
+        # Append any component-specific config_tests.xml files. We take care to only add a
+        # given file once, since adding a given file multiple times creates a "multiple
+        # matches" error. (This can happen if multiple CONFIG_TESTS_FILEs resolve to the
+        # same path.)
+        files_added = set()
         for comp in files.get_components("CONFIG_TESTS_FILE"):
             if comp is None:
                 continue
             infile = files.get_value("CONFIG_TESTS_FILE", attribute={"component": comp})
-            if os.path.isfile(infile):
+            infile_abspath = os.path.abspath(infile)
+            if os.path.isfile(infile) and infile_abspath not in files_added:
                 self.read(infile)
+                files_added.add(infile_abspath)
 
     def support_single_exe(self, case):
         """Checks if case supports --single-exe.
--- a/CIME/data/config/cesm/config_files.xml
+++ b/CIME/data/config/cesm/config_files.xml
@@ -112,6 +112,7 @@
     <values>
       <value>$CIMEROOT/CIME/data/config/config_tests.xml</value>
       <!-- component specific config_tests files -->
+      <value component="allactive">$SRCROOT/cime_config/config_tests.xml</value>
       <value component="clm">$COMP_ROOT_DIR_LND/cime_config/config_tests.xml</value>
       <value component="slim">$COMP_ROOT_DIR_LND/cime_config/config_tests.xml</value>
       <value component="cam">$COMP_ROOT_DIR_ATM/cime_config/config_tests.xml</value>
@@ -371,6 +372,7 @@
     <type>char</type>
     <values>
       <value component="any"       >$CIMEROOT/CIME/SystemTests</value>
+      <value component="allactive" >$SRCROOT/cime_config/SystemTests</value>
       <value component="clm"       >$COMP_ROOT_DIR_LND/cime_config/SystemTests</value>
       <value component="cam"       >$COMP_ROOT_DIR_ATM/cime_config/SystemTests</value>
       <value component="pop"       >$COMP_ROOT_DIR_OCN/cime_config/SystemTests</value>
--- a/CIME/tests/test_unit_xml_tests.py
+++ b/CIME/tests/test_unit_xml_tests.py
@@ -32,16 +32,27 @@ class TestXMLTests(unittest.TestCase):
 
             case.get_compset_components.return_value = ()
 
-            case.get_value.side_effect = (
-                "SMS",
-                tdir,
-                f"{caseroot}",
-                "SMS.f19_g16.S",
-                "cpl",
-                "SMS.f19_g16.S",
-                f"{caseroot}",
-                "SMS.f19_g16.S",
-            )
+            def fake_get_value(item, attribute=None):
+                simple_lookup = {
+                    "TESTCASE": "SMS",
+                    "CASEROOT": f"{caseroot}",
+                    "CASEBASEID": "SMS.f19_g16.S",
+                    "COMP_INTERFACE": "cpl",
+                    "DRV_RESTART_POINTER": None,
+                }
+                if item in simple_lookup:
+                    return simple_lookup[item]
+                elif item == "SYSTEM_TESTS_DIR":
+                    if attribute["component"] == "any":
+                        return tdir
+                    else:
+                        return None
+
+                raise KeyError(
+                    f"Unmocked call: case.get_value({item}, attribute={attribute})"
+                )
+
+            case.get_value.side_effect = fake_get_value
 
             tests = Tests()
 
@@ -65,17 +76,28 @@ class TestXMLTests(unittest.TestCase):
 
             case.get_compset_components.return_value = ()
 
-            case.get_value.side_effect = (
-                "ERP",
-                tdir,
-                f"{caseroot}",
-                "ERP.f19_g16.S",
-                "cpl",
-                None,
-                "ERP.f19_g16.S",
-                f"{caseroot}",
-                "ERP.f19_g16.S",
-            )
+            def fake_get_value(item, attribute=None):
+                simple_lookup = {
+                    "TESTCASE": "ERP",
+                    "CASEROOT": f"{caseroot}",
+                    "CASEBASEID": "ERP.f19_g16.S",
+                    "CASE": "ERP.f19_g16.S",
+                    "COMP_INTERFACE": "cpl",
+                    "DRV_RESTART_POINTER": None,
+                }
+                if item in simple_lookup:
+                    return simple_lookup[item]
+                elif item == "SYSTEM_TESTS_DIR":
+                    if attribute["component"] == "any":
+                        return tdir
+                    else:
+                        return None
+
+                raise KeyError(
+                    f"Unmocked call: case.get_value({item}, attribute={attribute})"
+                )
+
+            case.get_value.side_effect = fake_get_value
 
             tests = Tests()
 
--- a/CIME/utils.py
+++ b/CIME/utils.py
@@ -2267,7 +2267,7 @@ def find_system_test(testname, case):
     if testname.startswith("TEST"):
         system_test_path = "CIME.SystemTests.system_tests_common.{}".format(testname)
     else:
-        components = ["any"]
+        components = ["any", "allactive"]
         components.extend(case.get_compset_components())
         fdir = []
         for component in components:
@@ -2276,6 +2276,11 @@ def find_system_test(testname, case):
             )
             if tdir is not None:
                 tdir = os.path.abspath(tdir)
+                if tdir in fdir:
+                    # This can happen if multiple SYSTEM_TESTS_DIRs resolve to the same
+                    # path; in this case, we just want to handle the first occurrence and
+                    # skip the rest.
+                    continue
                 system_test_file = os.path.join(tdir, "{}.py".format(testname.lower()))
                 if os.path.isfile(system_test_file):
                     fdir.append(tdir)


CIME/test_scheduler.py
```diff
@@ -674,12 +674,12 @@ def _create_newcase_phase(self, test):
...
                 elif case_opt.startswith("N"):
-                    expect(ncpl == 1, "Cannot combine _C and _N options")
+                    expect(ncpl == 1, "Cannot combine _C# and _N options")
                     ninst = case_opt[1:]
                     create_newcase_cmd += " --ninst {}".format(ninst)
                     logger.debug(" NINST set to {}".format(ninst))
-                elif case_opt.startswith("C"):
-                    expect(ninst == 1, "Cannot combine _C and _N options")
+                elif case_opt.startswith("C") and not case_opt.startswith("CG"):
+                    expect(ninst == 1, "Cannot combine _C# and _N options")
```
billsacks: Thanks for this fix, @kdraeder !

@kdraeder and @jedwards4b - What would you think about changing the `CG` modifier to instead be `cG` (lowercase `G`) to further remove some of this ambiguity? It doesn't look like we use any `CG` test modifiers in any CESM test list, so this shouldn't break anything... and it would be easier to change this now before we start getting tests with this modifier. This would be a departure from the current use of all uppercase as the first letter for a test modifier, so I don't love this inconsistency, but in my mind it's still better to have this disambiguation even if it means for some inconsistency in this respect. What do you both think?

CIME/SystemTests/eri.py
```diff
@@ -90,7 +90,9 @@ def run_phase(self):
         start_1_year, start_1_month, start_1_day = [
             int(item) for item in start_1.split("-")
         ]
-        start_2_year = start_1_year + 2
+        # KDR The default start date is a leap day, so 2-29 2 years later doesn't exist; run failure.
+        # start_2_year = start_1_year + 2
```
billsacks: I'm happy with this fix, but the comment is confusing, since I think it only applies to a specific case, not a general issue with the ERI test. What about changing it to something like:

```suggestion
        # Change the year for the hybrid case to make sure the system can handle this
        # change in year.
        #
        # Note: When using a Gregorian calendar, it can be important to have consistency
        # between whether the two years are leap years or not because of how this test is
        # set up; so change year by 4 so that these will generally be consistent.
```

(Note that I have also removed your initials and removed the commented-out old code.)

CIME/test_scheduler.py
jedwards4b: okay by me.

kdraeder:
> Use pip or conda to install the pre-commit tool,
Pip didn't work, and neither does conda.
\>  conda install -c conda-forge pre_commit
```
EnvironmentNotWritableError: The current user does not have write permissions to the target environment.
  environment location: /glade/u/apps/opt/conda
  uid: 4241
  gid: 69284
```

CIME/test_scheduler.py
kdraeder: > What would you think about changing the `CG` modifier to instead be `cG` (lowercase `G`)

I'm guessing that you meant "lower case 'C'" not 'G'.
I wondered about that possibility, but didn't know how to see whether it would be disruptive, 
so thanks for looking into it!
I agree that it would be clearer to not use _C for 2 modifiers.
How about using just _G (currently unused) for gregorian and _C# for the ensemble size?
That would preserve the use of only capitol letters for modifiers too.

CIME/SystemTests/eri.py
kdraeder: Yes, that comment is mostly left over from debugging.
Your suggestion looks good.
We could consider being more specific about "because of how this test is set up".
"because a testmod which tests the gregorian calendar starts the runs on leap day, 
which must exist in all of the start years."

kdraeder:
> You need to use: `pip install --user`
> 
> --user Install to the Python user install directory for your platform. Typically ~/.local/, or %APPDATA%\Python on Windows. (See the Python documentation for site.USER_BASE for full details.)

I'm sorry that I don't have more competence with python.
site.USER_BASE seems to be a path name variable in a python module.
The documentation seems to tell me that I can set it by environment variable PYTHONUSERBASE.

\> setenv PYTHONUSERBASE ~/.local
\> ls $PYTHONUSERBASE
etc  lib  share
\> pip install --user pre_commit
ERROR: Can not perform a '--user' install. User site-packages are not visible in this virtualenv.


Since my changes are so limited, maybe I should skip using a tool to fix the formatting
and just look at a document that tells me how it should look.
But I suppose it will be useful to be able to use black in the future . . .

CIME/test_scheduler.py
billsacks: Yes, sorry, I meant lower case 'C', not 'G' - sorry for the confusion.

I also thought about using _G. I think that would be a reasonable option, though with these downsides / caveats:
- I wondered if there might be a future desire to use some other calendar besides Gregorian; if so, that argues for keeping the "c" syntax rather than having completely separate options for each calendar. This was my main rationale that led me to the thought of keeping "_cG" rather than just "_G".
- This line implies that "G" is a currently-accepted option, though from a quick look I can't see where it's used so maybe this is some historical code that isn't true anymore??? In any case, a little more digging should be done if we want to go with "_G":

https://github.com/ESMCI/cime/blob/2f3f2216a7717abd136bec63e42611f6e54f7056/CIME/test_scheduler.py#L941

In all, I still lean towards "_cG", but don't feel strongly about it. So, @kdraeder, @jedwards4b or anyone else - if you have even moderately strong opinions about it, I'm happy to go with your feelings.

CIME/SystemTests/eri.py
billsacks: I like your updated wording - my ambiguity came from my not fully understanding the situation.

Can you change this wording to your suggested version on top of mine?

billsacks:
> just look at a document that tells me how it should look.

I don't think such a document exists... the guidance at this point for CIME, as for many python projects, is "just run black and use whatever formatting it provides". But if you don't want to go down the rabbit hole of getting it set up this time, I can run it for you once you are done making other changes.

CIME/test_scheduler.py
kdraeder: Doh, I missed that one.
I can see some value in being able to provide other calendars.
I don't have a deep background in what testing is or needs to be done,
so I don't have a strong opinion keeping the capital letter convention or adding modifiers.
A more thorough search leads me to believe that F, H, J, K, O, Q, S, T, U, W-Z are not being used.
_T for time?  _K for kalender?

billsacks: I still vote for lowercase "_c": I'd prefer something more mnemonic even if it breaks with the current uppercase convention.

@jedwards4b , @jgfouca @jasonb5 - any opinions on using a lowercase `_c` as a test modifier to specify calendar, replacing the current uppercase `_C`, which is currently used for two different purposes? It would be the first use of a lowercase test option, though I personally feel okay about this precedent as we get more and more test modifiers and start running out of letters.

jgfouca: That's fine with me.

kdraeder:
> 2\. get out of conda altogether and then run pip install --user

I ran `conda deactivate`  several times, but `pip install --user {pre_commit}` fails with the same message.
\> conda env list 
shows a dozen environments
Is there some other kind of deactivation I should use?

kdraeder:
I removed the conda parts of my .tcshrc file, and I don't see it in any other dot files.
I opened  a new xterm on derecho, which is not aware of a conda command, so that seems good.
But the pip command, with and without --user, fails in the same ways.
You probably have much more pressing things to do, so if you'd like me to get help from the consultants,
I'll do that.

jedwards4b:
I guess that you should try that.  Sorry I wasn't more helpful.

CIME/test_scheduler.py
billsacks: @kdraeder is it okay with you if I push a change to your branch to change to using `_c` as the test modifier for calendar?

kdraeder: > @kdraeder is it okay with you if I push a change to your branch to change to using `_c` as the test modifier for calendar?

Yes, that should work well.  
In which case the line that tests for _C and not _CG can be reverted.

billsacks:
@kdraeder - I think the question in your last paragraph (about making the use case know about the calendar) may be best to ask in your CAM PR, since it mainly relates to how things should be done in CAM. That said, based on my possibly-incorrect understanding of CAM, I think it may get the grid and other variables from the config_cache.xml file written by its configure script. If I'm right about that, then you would need to add the calendar there.

I don't know the answer to your question about using CALENDAR in config_component.xml, but I'd be hesitant about adding a dependence on CALENDAR even if it's possible because CALENDAR can be set any time before case.build whereas I think config_component is parsed at create_newcase time – so you risk picking up the wrong thing when parsing config_component if CALENDAR changes later. (Though I may be wrong about this ordering.)

Given the growing complexity here, I think it's worth stepping back and asking if there's another way to accomplish what you're trying to do. My initial question is: How much of this feels important to work out of the box for a user setting up a case, vs. mainly being important for the sake of creating a test? If the main point of all of this is to create a test with the Gregorian calendar, then I wonder if the best approach would be to create a testmod, or modify an existing testmod, with the changes you need for the test to work (e.g., see https://github.com/ESCOMP/CAM/tree/cam_development/cime_config/testdefs/testmods_dirs/cam/dartcambigens). Then I'm hoping you wouldn't need to bother with a new CAM use case, but could instead use an existing use case and just modify select variables via the testmod. (In fact, with that approach, you wouldn't even need `_cG`, but could change to using a GREGORIAN calendar in the testmod - though using `_cG` is also a fine way to go.)

That said, if it's important to have things work more cleanly out of the box for users creating new cases, and not just for the test system, then it could be worth continuing down the path you're on. In that case, I think engaging the CAM developers via your PR there would be best.

billsacks:
@kdraeder (and others) - I have pushed the change we discussed, using lowercase `_c` for calendar.

I tested by creating the following tests and checking the relevant xml variables (I didn't build or run the tests):
- SMS_Ld3_D_P8x1_C3.f10_f10_mg37.X.green_gnu
- SMS_Ld3_D_P8x1_cG.f10_f10_mg37.X.green_gnu
- SMS_Ld3_D_P8x1_cG_C3.f10_f10_mg37.X.green_gnu

Note that I reverted some of the error message changes that had changed "_C" to "_C#": I felt it was inconsistent to refer to this as "_C#" when we just refer to "_N" without the "#", and now it's unnecessary with the change from _C to _c for calendar. I'd be fine returning to referring to this as "_C#", but if so, feel we should also refer to "_N#" for consistency.

@kdraeder - whenever you get a chance, let me know if you feel this is ready to go; if so, I can merge it.

billsacks:
@jgfouca - I'm requesting a formal approval from you on this before merging since it changes behavior of the ERI test and will lead to baseline comparison failures for the ERI test due to "no original counterpart" / "no compare counterpart", because now we add 4 years instead of 2. Let us know if that's a problem from the E3SM side.

jgfouca (approved):
We (E3SM) don't have any ERI tests in any of our suites, so I'm fine with a behavior change if you guys need it.

jedwards4b (approved):

CIME/test_scheduler.py
```diff
@@ -688,6 +691,15 @@ def _create_newcase_phase(self, test):
...
                     driver = case_opt[1:]
+                elif case_opt.startswith("cG"):
+                    # Need to examine the long name.
+                    # spinup cases are 1850, 2000, 2010
+                    # not HIST, SSP###
+                    spinup_po = re.compile(r'([\d]{1,4})')
+                    spinup = spinup_po.match(compset)
+                    expect(spinup == None,
+                        "Gregorian calendar (_cG) cannot be used with spinup compset \n  {}".format(compset)
+                    )
```
billsacks: I would prefer that this check **not** be here, for a few reasons:

First, I don't believe that there is a fundamental limitation of the system that you can't use the Gregorian calendar with spinup compsets. My impression from the discussion here is that there are problems in some components (e.g., CAM) when you try to use a Gregorian calendar in a spinup compset because of the way they have set up streams, but this is not a fundamental limitation of the system, and in the future we may want to fix that component behavior or have Gregorian spinup tests of a compset with components that don't have that issue – for example, I would expect a Gregorian X compset (coupler-test compset) test to pass currently; I haven't tried that, but if it fails I would like to be able to see that failure in whatever way it shows up rather than having CIME prevent me from running it at all.

An additional concern I have is that the mechanism for checking whether it's a spinup compset, by checking whether the compset name starts with a number, is - I believe - model-specific, not something that's prescribed by CIME, and so it doesn't feel like this logic belongs in CIME (though I could be wrong about that).

Also, if people feel it's important to have this check, then it should probably be done in a way that catches (probably more common) alternative ways of setting CALENDAR – i.e., by setting the CALENDAR XML variable after creating a case. And if this check is kept, I would prefer to see it moved to the component(s) where it is actually an issue. So, for example, this could be moved to the CAM build-namelist. 

@kdraeder and/or @jedwards4b - if, after all this, you still feel it's best to have this check in CIME, I can accept it, but I then have a few specific questions / requests about the implementation of this check:
- Can it be put in the block that already parses the `cG` option (further down in this file) rather than having the `cG` option checked in two places?
- Is this regex really what's intended... in particular the inclusion of `1` in the `{1,4}`? It looks like this will match any compset that starts with a single digit... and if that's the intent then a simpler regex of just '\d' would be sufficient. I think the intent may be to only match a compset that starts with 4 digits, in which case I think the regex you want is `r'\d{4}'`
- If this is kept, I'd like a more detailed comment explaining what the problem is with having a Gregorian test of a spinup case, with a goal of explaining whether / when this check could be removed in the future.

jedwards4b: I agree that the assumption that if the compset name starts with a digit it's a spinup compset is misguided and this check should be removed.  However I do not understand the utility or need to support a Gregorian calendar in a spinup compset.

kdraeder: > I don't believe that there is a fundamental limitation of the system that you can't use the Gregorian calendar with spinup compsets.

@billsacks Thanks for your insights!

Maybe there should be no gregorian modifier and it should be handled in the place(s) Bill suggests.
If it is handled in create_test, and if there's an actual limitation, there should be a check somewhere.
I agree that it should be clear, focused, and easy to remove.

I agree that there's not a fundamental limitation, but in addition to @jedwards point, we discussed (above) the difficulty of specifying the correct (external forcing) namelist entries via the use_cases.  It's beyond me to make the use_case aware of the calendar, so that it could specify suitable year(s) to extract from the forcing files (like it does with the grid).  It could probably be done through a new testmod(s?).  I haven't figured out why both use_cases and test_mods can alter the namelists, but one is user-selectable and the other isn't.  That might answer the question of whether and how ERI for spinup compsets should handle gregorian.
If this were moved to the components' build-namelist, would the user have control over it in a way that's clear to people who run these tests?

If this limitation will stay for the time being, then:
I wasn't completely comfortable relying on the compset starting with 4 digits to flag spinup cases,
so if there are better solutions we should use them.  But that seemed to be the convention.
Yes, the {1,4} should be {4} or {4,4}.

I tried to put this check in a place where _cG was already being handled, but it seemed that the compset was not available there.  If it can be made available, then moving the check is good.  Moving it into the component(s) would be fine too.  It will take me a while to figure out how to do that.

billsacks: > However I do not understand the utility or need to support a Gregorian calendar in a spinup compset.

I also don't see enough scientific need for this to justify spending time on getting it working right now, but I don't think it should be fundamentally dis-allowed. Two things pop to mind: First, the use of a 365-day calendar by default is a model-specific thing, not a CIME thing; in principle, a model using CIME could decide that it wants to use a Gregorian calendar for all of its cases by default. (So I guess this is just a further argument that, if there is any logic about this, it should be in a component, not in CIME.) Second, I can at least see some theoretical reasons why someone might want to use a Gregorian calendar for a spinup run in CESM. One big one is for consistency: If you know you're going to use a Gregorian calendar for the later transient run, then you may want your spinup run to be consistent with that. I could see there being other reasons, too, that I / we aren't thinking of now.

I **am** generally in favor of adding error checking like this along with meaningful error messages to catch things that we know can't work together. I just want to see this localized to the place where the problem actually occurs so we don't end up preventing things that *should* actually work.

> Maybe there should be no gregorian modifier and it should be handled in the place(s) Bill suggests.

I don't have strong opinions one way or another on this. But I'd love to see this PR move towards resolution, so unless anyone else has strong-ish feelings, I'd vote for keeping this modifier in place as it is in this PR now.

> we discussed (above) the difficulty of specifying the correct (external forcing) namelist entries via the use_cases

Yes, agreed... but I feel this is a CAM limitation, not a CIME limitation, so if there is a check, it belongs in CAM's build-namelist.

> I haven't figured out why both use_cases and test_mods can alter the namelists, but one is user-selectable and the other isn't.

Don't get me started on use_cases... the short answer is that I'm not a fan of them, but CAM uses them extensively, so... 🤷🏻‍♂️

> If this were moved to the components' build-namelist, would the user have control over it in a way that's clear to people who run these tests?

I'm not sure what you're asking here; can you clarify?

> I wasn't completely comfortable relying on the compset starting with 4 digits to flag spinup cases,
so if there are better solutions we should use them.

There may be some additional options opened up if this check is moved into CAM's build-namelist, which I think is where it belongs. At the very least, it would at least be somewhat better to have this assumption about compset name be there rather than in CIME.

> I tried to put this check in a place where _cG was already being handled, but it seemed that the compset was not available there. If it can be made available, then moving the check is good. Moving it into the component(s) would be fine too. It will take me a while to figure out how to do that.

Ah, okay, thanks for clarifying.

I have appreciated your initiative in trying to get all of the needed changes in place yourself. That said, my suggestion at this point would be to take this check out of CIME and open an issue in CAM requesting a more clear error message when someone tries to use a GREGORIAN calendar in a non-transient case, and thus starting a discussion with the CAM developers about the best way to implement the check there.

kdraeder: > > If this were moved to the components' build-namelist, would the user have control over it in a way that's clear to people who run these tests?
> 
> I'm not sure what you're asking here; can you clarify?

I confused myself here.  I was thinking that the _cG might be moved from create_case to build-namelist, but we're talking about where the check should be, so _cG would stay in create_case
and be passed to build-namelist.

billsacks: Ah, yes. The `_cG` would stay here. This is simply a convenience mechanism to set the CALENDAR XML variable at create_test time. The CALENDAR variable can be queried by any component's build-namelist.

kdraeder: > The CALENDAR variable can be queried by any component's build-namelist.

DART will want to use Gregorian for all of the components to which it has interfaces: CAM, CLM, CICE, MOM, and components to be named later.  So a check of compset+calendar would need to be in several build-namelists.  I'll look into CAM.

Should this be a separate PR?  If it is, then we'll be closing this PR with code that still has
one of the problems that led me to open the issue.  I guess we could close it with a note about that.
And would the second PR be another partial fix of issue #4811, or a new issue?

billsacks: > a check of compset+calendar would need to be in several build-namelists

It's possible that other components don't have this issue and can safely be run with a GREGORIAN calendar. Actually, now I'm confused about why you ran into issues with this in CAM: digging into the CDEPS code, it looks like streams should be set up to handle discrepancies between the model calendar and the streams calendar:

https://github.com/ESCOMP/CDEPS/blob/95e0bac4a723496b0da57d9d290c86e449c97238/streams/dshr_strdata_mod.F90#L807-L854

And indeed, I just successfully ran this CLM-only test: `ERI_D_Ld9.f10_f10_mg37.I1850Clm60Bgc.derecho_gnu.clm-gregorian` (where I introduced a new test mod that simply set CALENDAR=GREGORIAN to avoid depending on the changes in this cime PR)

This is a further argument to me that this error check should only exist in the component(s) that actually have an issue running this case.

Maybe the issue is that CAM doesn't use CDEPS's streams code in many places, but instead implements its own stream reading? From one spot-check, it looks like attempts were made even there to handle this discrepancy as well... though maybe this doesn't handle all cases, and maybe not all of CAM's streams have similar handling?:

https://github.com/kdraeder/CAM/blob/fc28fce1ff3f2a33840536e7128cb1eb8337b277/src/chemistry/utils/mo_flbc.F90#L130-L137

> Should this be a separate PR? If it is, then we'll be closing this PR with code that still has
one of the problems that led me to open the issue. I guess we could close it with a note about that.
And would the second PR be another partial fix of issue https://github.com/ESMCI/cime/issues/4811, or a new issue?

It would need to be a PR in CAM - you could ask over in your existing CAM PR whether they want you to include it in your existing CAM PR or open a new CAM PR. It's fine to merge this CIME PR without everything from #4811 being fully resolved, especially because some of the issues in #4811 are really outside of CIME. The CAM PR(s) would be a fix of a new issue, in CAM.

billsacks: Just following up from my last comment: I meant to add:

Based on what I see in code and my being able to run an 1850 Gregorian case in CLM, I think you should only worry about putting an error check about Gregorian+spinup compset into CAM, unless you have run into issues running this combination in other components. And even in CAM, it may be that this is a bug and that in fact this combination should work... I'd suggest discussing this with the CAM developers.

kdraeder: I've removed this inconsistency check from my clone by resetting back to @billsacks latest commit (0afcb5fb2).  
`> git reset --hard 0afcb5fb2`
My clone (and branch) is now 8 commits behind my github kdraeder fork.  Git suggests that I update my clone, but I want to change my fork instead, to make 0afcb5fb2 be the HEAD(?).  Can I do that by pushing my clone to my fork?  Or do I need to do something directly in my fork?

kdraeder: > Maybe the issue is that CAM doesn't use CDEPS's streams code in many places, but instead implements its own stream reading?

My early 1850 test failed with:

> PET00 ESMCI_Calendar.C:1059 ESMCI::Calendar::convertToTime()
> Input argument out of range  - ; 
>  Gregorian: for February 1850, dd=29 > 28 days in the month.

Which led me to  
> &chem_surfvals_nl
   flbc_cycle_yr          = 1850

It looked like I should just change these years to leap years (in the 1850_cam_lt use_case).  
I didn't track down any potential coding causes of the failure; it seemed more likely that the namelist values were set incorrectly than that the code is insufficient.
I'll describe this more in the CAM issue that I open.

billsacks: First: I realized that my CTSM test wasn't a good one because it started Jan 1. I'm retrying with a test that starts in late Feb of 1852 and will post back here when I see those test results.

> I've removed this inconsistency check from my clone by resetting back to @billsacks latest commit ([0afcb5f](https://github.com/ESMCI/cime/commit/0afcb5fb2795d49bf84e8b37f1cc340295d5741d)).
> `> git reset --hard 0afcb5fb2`
> My clone (and branch) is now 8 commits behind my github kdraeder fork. Git suggests that I update my clone, but I want to change my fork instead, to make [0afcb5f](https://github.com/ESMCI/cime/commit/0afcb5fb2795d49bf84e8b37f1cc340295d5741d) be the HEAD(?). Can I do that by pushing my clone to my fork? Or do I need to do something directly in my fork?

It can be very problematic to use `git reset --hard` on a branch that has already been pushed, since pushing after that would be a destructive operation that rewrites history. As a general rule, you should only use that on your own work before pushing it to a place shared with other people, and even then it should be used with extreme caution because you can lose work with that command.

Here's what you should do at this point:

(1) Update to the latest version: `git pull`

(2) Manually delete your recently-added check, commit and re-push. There are ways you could do this with git, but in this case, since there are multiple commits touching it and it's just a single self-contained block of code, it will be easier to just remove this manually and do an additional commit.

> It looked like I should just change these years to leap years (in the 1850_cam_lt use_case).
> I didn't track down any potential coding causes of the failure; it seemed more likely that the namelist values were set incorrectly than that the code is insufficient.
> I'll describe this more in the CAM issue that I open.

Sounds good. So it looks like the failure is in the module where I did find some handling of leap-year discrepancies, but maybe some other handling is needed... I haven't dug into this code very far.

billsacks (approved):
Looks good to me now. Thank you for all of your iterations on this, @kdraeder !

I see there is a failing test that seems unrelated to this PR. I'm going to try updating to latest master and rerunning the tests.

PR #4901, opened 2025-12-03
wget server spider mode disabled

### LINKED ISSUE #4900: wget --spider fails on new gdex inputdata server
```
 wget -v --no-check-certificate --spider https://ftp.cgd.ucar.edu/cesm/inputdata/atm
Spider mode enabled. Check if remote file exists.
--2025-12-03 12:53:32--  https://ftp.cgd.ucar.edu/cesm/inputdata/atm
Resolving ftp.cgd.ucar.edu (ftp.cgd.ucar.edu)... 128.117.13.41
Connecting to ftp.cgd.ucar.edu (ftp.cgd.ucar.edu)|128.117.13.41|:443... connected.
HTTP request sent, awaiting response... 301 Moved Permanently
Location: https://ftp.cgd.ucar.edu/cesm/inputdata/atm/ [following]
Spider mode enabled. Check if remote file exists.
--2025-12-03 12:53:32--  https://ftp.cgd.ucar.edu/cesm/inputdata/atm/
Connecting to ftp.cgd.ucar.edu (ftp.cgd.ucar.edu)|128.117.13.41|:443... connected.
HTTP request sent, awaiting response... 200 OK
Length: unspecified [text/html]
Remote file exists and could contain further links,
but recursion is disabled -- not retrieving.
```

jedwards4b:
wget --spider is used by cesm to establish that the server exists without downloading anything
## Description
, but some servers have spider mode disabled so they exist and can connect but still
return an error with this command.   We can look for the key word 'connected' in the stderr of the wget command to determine if this is the issue and proceed without failing.   I think that this was the reason there was a special case for the NEON server and I have removed that special case, replacing it with this more generic handler.  

- Closes #4900 

## Checklist

### DIFF
--- a/CIME/Servers/wget.py
+++ b/CIME/Servers/wget.py
@@ -31,13 +31,13 @@ class WGET(GenericServer):
             )
         except RuntimeError:
             logger.warning(
-                "Could not connect to repo '{0}'\nThis is most likely either a proxy, or network issue .(location 1) ".format(
+                "Could not connect to repo '{0}'\nThis is most likely either a proxy, or network issue .".format(
                     address
                 )
             )
             return None
         if "Connecting to " in errstr and "... connected" in errstr:
-            logger.warning("spider mode disabled, server otherwise okay")
+            logger.warning("Connection established with nonzero code %s", err)
         elif err:
             logger.warning(
                 "Could not connect to repo '{0}'\n{1}".format(address, errstr)


CIME/Servers/wget.py
```diff
@@ -26,20 +26,21 @@ def wget_login(cls, address, user="", passwd=""):
             args += "--password {} ".format(passwd)
 
         try:
-            err = run_cmd("wget {} --spider {}".format(args, address), timeout=60)[0]
-        except:
+            err, _, errstr = run_cmd(
```
ekluzek, line 29: Ahh, so this gets both the error code and the error string "errstr"

CIME/Servers/wget.py
```diff
@@ -26,20 +26,21 @@ def wget_login(cls, address, user="", passwd=""):
...
+        except RuntimeError:
             logger.warning(
-                "Could not connect to repo '{0}'\nThis is most likely either a proxy, or network issue .(location 1)".format(
+                "Could not connect to repo '{0}'\nThis is most likely either a proxy, or network issue .(location 1) ".format(
                     address
                 )
             )
             return None
-
-        if err and not "storage.neonscience.org" in address:
```
ekluzek, line 38: There still needs to be handling for "storage.neonscience.org" though right? So another elf branch?

CIME/Servers/wget.py
```diff
@@ -26,20 +26,21 @@ def wget_login(cls, address, user="", passwd=""):
...
+        if "connected" in errstr:
```
ekluzek: I think "connected" is too short of a string to use here, and could be in other error strings.

I think you should get this in errstr, based on the error we see:

recursion is disabled -- not retrieving

That might be too long and too specific though, in case the error string from wget changes in the future. So an alternative could be the first three words, which is probably long enough. I thought of disregarding the case of the string as well, but that seems unlikely it would change from wget.

CIME/Servers/wget.py
```diff
@@ -26,20 +26,21 @@ def wget_login(cls, address, user="", passwd=""):
...
+            logger.warning("spider mode disabled, server otherwise okay")
```
ekluzek: The warning won't mean anything to most users. So maybe something to clarify what spider mode means?

```suggestion
            logger.warning("spider mode (to check that the server is available) is disabled, but server is otherwise okay")
```

CIME/Servers/wget.py
```diff
@@ -26,20 +26,21 @@ def wget_login(cls, address, user="", passwd=""):
...
+        elif err:
             logger.warning(
-                "Could not connect to repo '{0}'\nThis is most likely either a proxy, or network issue .(location 2)".format(
```
ekluzek, line 40: Is the bit about "a proxy, or network issue" no longer true? It does seem like a network problem going to be common case, and you just need to retry it . And someone can screw it up with a proxy, so might still be helpful to still suggest it as a possible culprit.

CIME/Servers/wget.py
jedwards4b: You misunderstand the meaning of wget --spider.  It is not getting the file tree of the inputdata server, it is simply establishing that the server exists and will accept connections.

CIME/Servers/wget.py
jedwards4b: I've augmented the string to be more precise.

CIME/Servers/wget.py
jedwards4b: You need to test this against NEON data but I'm pretty sure that this more general handler is going to work there.

CIME/Servers/wget.py
ekluzek: Ahh ok. Thanks for the correction. I propose something different above then.

CIME/Servers/wget.py
ekluzek: Ahh, OK you added that back in again. Cool.

CIME/Servers/wget.py
ekluzek: I like what you came up with.

CIME/Servers/wget.py
ekluzek: Yes, that looks much better. Looks to be more robust and unlikely to need changes if the wget error string changes.

CIME/Servers/wget.py
ekluzek: OK, I'll try it out and see what happens.

ekluzek: OK, I tried it for a NEON case on both Izumi and Derecho and it worked just fine. The --spider business seems to work fine for the NEON data and it just reports that the server is fine.

SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60SpRs.derecho_intel.clm-default--clm-NEON-TOOL--clm-nofireemis
SMS_Ld10_D_Mmpi-serial.CLM_USRDAT.I1PtClm60SpRs.izumi_nag.clm-default
--clm-NEON-TOOL--clm-nofireemis

So looks good to go.

PR #4907, opened 2025-12-23
Fixes rpointer archiving

jasonb5:
## Description

When using the `--copy-only` or `--last_date` option, the latest rpointer files would be moved from the run directory breaking the ability to restart a run. The first archived rpointer file would be incorrect as CIME would move the current rpointer into the archive directory and then create all subsequent rpointer files (these would be correct).

- Fixes latest rpointer files being moved from run directory
- Fixes first archived rpointer file contents
- Refactors `_archive_rpointer_files
- Adds unittests

## Checklist
- [ ] My code follows the style guidlines of this proejct (black formatting)

### DIFF
--- a/CIME/XML/env_archive.py
+++ b/CIME/XML/env_archive.py
@@ -16,23 +16,5 @@ class EnvArchive(ArchiveBase, EnvBase):
         schema = os.path.join(utils.get_schema_path(), "env_archive.xsd")
         EnvBase.__init__(self, case_root, infile, schema=schema, read_only=read_only)
 
-    def get_archive_specs(self):
-        components_element = self.get_child("components")
-
-        return self.get_children("comp_archive_spec", root=components_element)
-
-    def get_rpointer_nodes(self, root):
-        assert root.name == "comp_archive_spec"
-
-        return self.get_children("rpointer", root=root)
-
-    def get_rpointers(self, root):
-        for node in self.get_rpointer_nodes(root):
-            file = self.get_child("rpointer_file", root=node).text
-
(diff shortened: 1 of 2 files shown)


CIME/XML/env_archive.py
```diff
@@ -16,14 +16,35 @@ def __init__(self, case_root=None, infile="env_archive.xml", read_only=False):
...
+
+    def get_rpointers(self, root):
+        for node in self.get_rpointer_nodes(root):
+            file = self.get_child("rpointer_file", root=node).text
+
+            content = self.get_child("rpointer_content", root=node).text
+
+            yield file, content
+
+    # TODO check if used otherwise remove
```
jedwards4b: I could not find any usage of any of get_entries, get_entry_info or get_rpointer_contents in CESM.

jedwards4b:
Regarding the test failures I found that a check of the object 'archive' at line 200 of case_st_archive.py
reveals it to be of type: <CIME.XML.archive.Archive object at 0x7f40e60e9f90>
not of type env_archive as expected.

jedwards4b (changes requested):
Running `case.st_archive --test-all` reveals a problem with your new code.  The 
archive variable is of type Archive and not env_archive as you expect.

CIME/XML/env_archive.py
jasonb5: I could not find any usage either, I'll go ahead and remove them.

jasonb5:
@jedwards4b Thanks, I see the issue. I'll adjust the code and have you review again.

Issue #4858, opened 2025-10-02
Recent pylint complains about raising a general Exception


billsacks:
In trying to update to the latest version of pylint (https://github.com/ESMCI/cime/issues/4856 ), in addition to the issues mentioned in #4857 , I got one other error:

```
************* Module CIME.XML.tests
CIME/XML/tests.py:56:12: W0719: Raising too general exception: Exception (broad-exception-raised)
```

We should either raise a more specific exception or bypass this check here. @jasonb5 it looks like you added this. Do you have any insight into whether it would make sense to raise a more specific exception here?

bartgol:
What about just using `CIMEError`?

billsacks:
I think this will be resolved by @jedwards4b 's #4908

PR #4908, opened 2026-01-02
Update/pylint

### LINKED ISSUE #4856: Update pylint version in pre-commit
I had trouble running pre-commit on my Mac with python 3.13:

<details>

```
$ pre-commit run -a
[INFO] Installing environment for https://github.com/PyCQA/pylint.
[INFO] Once installed this environment will be reused.
[INFO] This may take a few minutes...
An unexpected error has occurred: CalledProcessError: command: ('/Users/sacks/.cache/pre-commit/repofyy_uh61/py_env-python3.13/bin/python', '-mpip', 'install', '.')
return code: 1
stdout:
    Processing /Users/sacks/.cache/pre-commit/repofyy_uh61
      Installing build dependencies: started
      Installing build dependencies: finished with status 'done'
      Getting requirements to build wheel: started
      Getting requirements to build wheel: finished with status 'done'
      Preparing metadata (pyproject.toml): started
      Preparing metadata (pyproject.toml): finished with status 'done'
    Collecting platformdirs>=2.2.0 (from pylint==2.11.1)
      Using cached platformdirs-4.4.0-py3-none-any.whl.metadata (12 kB)
    Collecting astroid<2.9,>=2.8.0 (from pylint==2.11.1)
      Using cached astroid-2.8.6-py3-none-any.whl.metadata (4.7 kB)
    Collecting isort<6,>=4.2.5 (from pylint==2.11.1)
      Using cached isort-5.13.2-py3-none-any.whl.metadata (12 kB)
    Collecting mccabe<0.7,>=0.6 (from pylint==2.11.1)
      Using cached mccabe-0.6.1-py2.py3-none-any.whl.metadata (4.3 kB)
    Collecting toml>=0.7.1 (from pylint==2.11.1)
      Using cached toml-0.10.2-py2.py3-none-any.whl.metadata (7.1 kB)
    Collecting lazy-object-proxy>=1.4.0 (from astroid<2.9,>=2.8.0->pylint==2.11.1)
      Using cached lazy_object_proxy-1.12.0-cp313-cp313-macosx_11_0_arm64.whl.metadata (5.1 kB)
    Collecting wrapt<1.14,>=1.11 (from astroid<2.9,>=2.8.0->pylint==2.11.1)
      Using cached wrapt-1.13.3.tar.gz (48 kB)
      Installing build dependencies: started
      Installing build dependencies: finished with status 'done'
      Getting requirements to build wheel: started
      Getting requirements to build wheel: finished with status 'done'
      Preparing metadata (pyproject.toml): started
      Preparing metadata (pyproject.toml): finished with status 'error'
stderr:
      error: subprocess-exited-with-error

      × Preparing metadata (pyproject.toml) did not run successfully.
      │ exit code: 1
      ╰─> [98 lines of output]
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/expand.py", line 71, in __getattr__
              return next(
                  ast.literal_eval(value)
                  for target, value in self._find_assignments()
                  if isinstance(target, ast.Name) and target.id == attr
              )
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/expand.py", line 72, in <genexpr>
              ast.literal_eval(value)
              ~~~~~~~~~~~~~~~~^^^^^^^
            File "/opt/homebrew/Cellar/python@3.13/3.13.7/Frameworks/Python.framework/Versions/3.13/lib/python3.13/ast.py", line 110, in literal_eval
              return _convert(node_or_string)
            File "/opt/homebrew/Cellar/python@3.13/3.13.7/Frameworks/Python.framework/Versions/3.13/lib/python3.13/ast.py", line 109, in _convert
              return _convert_signed_num(node)
            File "/opt/homebrew/Cellar/python@3.13/3.13.7/Frameworks/Python.framework/Versions/3.13/lib/python3.13/ast.py", line 83, in _convert_signed_num
              return _convert_num(node)
            File "/opt/homebrew/Cellar/python@3.13/3.13.7/Frameworks/Python.framework/Versions/3.13/lib/python3.13/ast.py", line 74, in _convert_num
              _raise_malformed_node(node)
              ~~~~~~~~~~~~~~~~~~~~~^^^^^^
            File "/opt/homebrew/Cellar/python@3.13/3.13.7/Frameworks/Python.framework/Versions/3.13/lib/python3.13/ast.py", line 71, in _raise_malformed_node
              raise ValueError(msg + f': {node!r}')
          ValueError: malformed node or string on line 2: <ast.Call object at 0x102783dd0>

          The above exception was the direct cause of the following exception:

            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/expand.py", line 185, in read_attr
              value = getattr(StaticModule(module_name, spec), attr_name)
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/expand.py", line 77, in __getattr__
              raise AttributeError(f"{self.name} has no attribute {attr}") from e
          AttributeError: wrapt has no attribute __version__

          During handling of the above exception, another exception occurred:

            File "/Users/sacks/.cache/pre-commit/repofyy_uh61/py_env-python3.13/lib/python3.13/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 389, in <module>
              main()
              ~~~~^^
            File "/Users/sacks/.cache/pre-commit/repofyy_uh61/py_env-python3.13/lib/python3.13/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 373, in main
              json_out["return_val"] = hook(**hook_input["kwargs"])
                                       ~~~~^^^^^^^^^^^^^^^^^^^^^^^^
            File "/Users/sacks/.cache/pre-commit/repofyy_uh61/py_env-python3.13/lib/python3.13/site-packages/pip/_vendor/pyproject_hooks/_in_process/_in_process.py", line 175, in prepare_metadata_for_build_wheel
              return hook(metadata_directory, config_settings)
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/build_meta.py", line 374, in prepare_metadata_for_build_wheel
              self.run_setup()
              ~~~~~~~~~~~~~~^^
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/build_meta.py", line 512, in run_setup
              super().run_setup(setup_script=setup_script)
              ~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/build_meta.py", line 317, in run_setup
              exec(code, locals())
              ~~~~^^^^^^^^^^^^^^^^
            File "<string>", line 41, in <module>
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/__init__.py", line 115, in setup
              return distutils.core.setup(**attrs)
                     ~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/_distutils/core.py", line 160, in setup
              dist.parse_config_files()
              ~~~~~~~~~~~~~~~~~~~~~~~^^
            File "/Users/sacks/.cache/pre-commit/repofyy_uh61/py_env-python3.13/lib/python3.13/site-packages/_virtualenv.py", line 22, in parse_config_files
              result = old_parse_config_files(self, *args, **kwargs)
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/dist.py", line 752, in parse_config_files
              setupcfg.parse_configuration(
              ~~~~~~~~~~~~~~~~~~~~~~~~~~~~^
                  self, self.command_options, ignore_option_errors=ignore_option_errors
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
              )
              ^
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/setupcfg.py", line 188, in parse_configuration
              meta.parse()
              ~~~~~~~~~~^^
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/setupcfg.py", line 502, in parse
              section_parser_method(section_options)
              ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/setupcfg.py", line 477, in parse_section
              self[name] = value
              ~~~~^^^^^^
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/setupcfg.py", line 294, in __setitem__
              parsed = self.parsers.get(option_name, lambda x: x)(value)
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/setupcfg.py", line 600, in _parse_version
              return expand.version(self._parse_attr(value, self.package_dir, self.root_dir))
                                    ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/setupcfg.py", line 419, in _parse_attr
              return expand.read_attr(attr_desc, package_dir, root_dir)
                     ~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/expand.py", line 190, in read_attr
              module = _load_spec(spec, module_name)
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-build-env-hnbpvhbk/overlay/lib/python3.13/site-packages/setuptools/config/expand.py", line 211, in _load_spec
              spec.loader.exec_module(module)
              ~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^
            File "<frozen importlib._bootstrap_external>", line 1026, in exec_module
            File "<frozen importlib._bootstrap>", line 488, in _call_with_frames_removed
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-install-dwq0vdza/wrapt_3dd7183d94544ffba170c187878610d3/src/wrapt/__init__.py", line 10, in <module>
              from .decorators import (adapter_factory, AdapterFactory, decorator,
                      synchronized)
            File "/private/var/folders/mp/rlk6q70n1jl_syyghjrhsv7w0000gp/T/pip-install-dwq0vdza/wrapt_3dd7183d94544ffba170c187878610d3/src/wrapt/decorators.py", line 34, in <module>
              from inspect import ismethod, isclass, formatargspec
          ImportError: cannot import name 'formatargspec' from 'inspect' (/opt/homebrew/Cellar/python@3.13/3.13.7/Frameworks/Python.framework/Versions/3.13/lib/python3.13/inspect.py). Did you mean: 'formatargvalues'?
          [end of output]

      note: This error originates from a subprocess, and is likely not a problem with pip.
    error: metadata-generation-failed

    × Encountered error while generating package metadata.
    ╰─> See above for output.

    note: This is an issue with the package mentioned above, not pip.
    hint: See above for details.
Check the log at /Users/sacks/.cache/pre-commit/pre-commit.log
```

</details>

I think this is a dependency issue. Updating pylint to the latest (v3.3.8; we're currently pointing to the old v2.11.1) in `.pre-commit-config.yaml` solves this issue.

However, this pylint update will require either fixing (ideally) or disabling (short-term) some checks that are now failing. I'll open separate issues for these.

@jasonb5 and others: does it seem reasonable to update pylint to the latest in our pre-commit config?

jedwards4b:
## Description
Removes all pylint errors for versions up to and including v4.0.4
note that I did not update the default from v2.11.1
Fix some spelling in the github templates.
- Closes #4856

## Checklist

### DIFF
--- a/.github/workflows/testing.yml
+++ b/.github/workflows/testing.yml
@@ -99,7 +99,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        python-version: ["3.8", "3.10", "3.12"]
+        python-version: ["3.9", "3.10", "3.12", "3.13", "3.14"]
     steps:
       - name: Checkout model code
         uses: actions/checkout@v4
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -18,7 +18,7 @@ repos:
       - id: black
         files: CIME
   - repo: https://github.com/PyCQA/pylint
-    rev: v2.11.1
+    rev: v4.0.4
     hooks:
       - id: pylint
         args:
--- a/CIME/Tools/standard_script_setup.py
+++ b/CIME/Tools/standard_script_setup.py
@@ -36,10 +36,10 @@ def check_minimum_python_version(major, minor, warn_only=False):
     raise RuntimeError(msg + " - please use a newer version of Python.")
 
 
-# Require users to be using >=3.6
-check_minimum_python_version(3, 6)
-# Warn users if they are using <3.8
-check_minimum_python_version(3, 8, warn_only=True)
+# Require users to be using >=3.9
+check_minimum_python_version(3, 9)
+# Warn users if they are using <3.10
+check_minimum_python_version(3, 10, warn_only=True)
 
 real_file_dir = os.path.dirname(os.path.realpath(__file__))
 cimeroot = os.path.abspath(os.path.join(real_file_dir, "..", ".."))


CIME/case/case_setup.py
```diff
@@ -311,13 +311,14 @@ def _case_setup_impl(
             ninst = case.get_value("NINST")
```
billsacks, line 315:
```suggestion
else
    # This default value will be overridden below for each component
    ninst = 1

```

CIME/case/case_setup.py
```diff
@@ -311,13 +311,14 @@ def _case_setup_impl(
...
 
         multi_driver = case.get_value("MULTI_DRIVER")
-
+        ninst = 0
```
billsacks: I'd find this more clear with my suggestion above (putting its setting with the other ninst setting, along with a comment... and setting the default to 1, which feels like a more reasonable fall-back default).

CIME/case/case_setup.py
billsacks: See my comment a few lines down. It could also be reasonable to set ninst to None at first.

CIME/case/case_submit.py
```diff
@@ -355,6 +355,9 @@ def check_case(self, skip_pnl=False, chksum=False):
         elif ncpl_base_period == "decade":
             coupling_secs = 315360000 / maxncpl
             timestep = 315360000 / minncpl
+        else:
+            coupling_secs = None
+            timestep = None
```
billsacks: It feels like this `else` clause should raise an exception like "unhandled ncpl_base_period".

CIME/SystemTests/system_tests_common.py
```diff
@@ -185,6 +185,9 @@ def _set_restart_interval(
         elif ncpl_base_period == "decade":
             coupling_secs = 315360000 / maxncpl
             timestep = 315360000 / minncpl
+        else:
+            coupling_secs = None
+            timestep = None
```
billsacks: As above: It feels like this else clause should raise an exception like "unhandled ncpl_base_period".

CIME/XML/compsets.py
```diff
@@ -85,7 +87,8 @@ def get_value(self, name, attribute=None, resolved=False, subgroup=None):
                         alias = self.text(child)
                     if self.name(child) == "lname":
                         lname = self.text(child)
-                compsets[alias] = lname
+                if alias is not None and lname is not None:
```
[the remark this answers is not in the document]
billsacks: It seems like this Copilot suggestion is valid: It seems like the settings to `None` should happen inside the `for node` loop. Though I don't understand this completely, so maybe I'm reading this wrong?

In addition, if it should always be the case that alias and lname are found/set at this point (based on the old code, I'm thinking that's the case?), then there should be an `else` here that aborts with an error message.

CIME/XML/namelist_definition.py
```diff
@@ -140,13 +140,17 @@ def get_group_name(self, node=None):
             group = self.get(node, "group")
         elif self.get_version() >= 2.0:
             group = self.get_element_text("group", root=node)
+        else:
+            group = None
```
billsacks: It seems like this should raise an exception like "Unhandled version".

CIME/XML/namelist_definition.py
```diff
@@ -140,13 +140,17 @@ def get_group_name(self, node=None):
...
         return group
 
     def _get_type(self, node):
         if self.get_version() == 1.0:
             type_info = self.get(node, "type")
         elif self.get_version() >= 2.0:
             type_info = self._get_type_info(node)
+        else:
+            type_info = None
```
billsacks: Again, it seems like this should raise an exception like "Unhandled version"

CIME/XML/namelist_definition.py
```diff
@@ -534,6 +538,8 @@ def get_input_pathname(self, name):
             input_pathname = self.get(node, "input_pathname")
         elif self.get_version() >= 2.0:
             input_pathname = self._get_node_element_info(node, "input_pathname")
+        else:
+            input_pathname = None
```
billsacks: Again, it seems like this should raise an exception like "Unhandled version"

CIME/compare_test_results.py
```diff
@@ -151,7 +151,8 @@ def compare_test_results(
                     logfile_name,
                     test_dir,
                 )
-
+            compare_result = None
+            compare_comment = None
```
billsacks: I spent some time trying to understand this one. I think the fix needed here might be different from this one, but I don't understand all of the possible permutations well enough to know for sure what the right fix is.

Specifically: I think the problem here comes from a discrepancy in the conditional on line 114 (`if not namelists_only and not build_only`) and the conditional on line 186 of the unmodified code (`if not namelists_only`). The problem is: If `namelists_only` is False and `build_only` is True, then the first conditional will lead to `do_compare = False`. But then we'll still enter the second conditional and try to write `compare_result` to `brief_result`. In the old code, this would lead to an error (trying to write an unset `compare_result`); with the change in this PR it would lead to writing `None` to `brief_result`, which seems wrong.

The pylint error goes away if we change the conditional on line 186 to `if not namelists_only and not build_only`, removing the need for the above change (setting `compare_result` and `compare_comment` to `None`), but I'm not sure if that's actually the correct fix. What is the expected behavior in the case where `namelists_only` is False and `build_only` is True?

CIME/get_timing.py
```diff
@@ -390,7 +390,8 @@ def _getTiming(self, inst=0):
         inittype = "FALSE"
         if (run_type == "startup" or run_type == "hybrid") and not continue_run:
             inittype = "TRUE"
-
+        binfilename = None
+        finfilename = None
```
billsacks: It seems like the conditional starting on line 399 should have an `else` clause that aborts with a message like "Unknown driver". These settings to `None` could then be made in that else clause if needed to satisfy pylint.

CIME/get_timing.py
```diff
@@ -450,6 +451,9 @@ def _getTiming(self, inst=0):
             logger.warning("Unknown NCPL_BASE_PERIOD={}".format(ncpl_base_period))
 
         # at this point the routine becomes driver specific
+        nsteps = 0
+        nmax = 0
+        fmax = 0
```
billsacks: Similarly to my above comment in this file, I'd find this more clear (and possibly more robust long-term) if this were changed a bit:

(1) In the conditional just below this, add an `else` that aborts with an error like "unknown driver"; in that conditional, set nsteps = 0 (or None) if needed by pylint.

(2) In the conditional around line 536 in the original code (`if self._driver == "nuopc"`): change the `if` on line 550 (`if self._driver == "mct" or self._driver == "moab"`) to an `elif`, then add an `else` that aborts with an error like "unknown driver", and if needed by pylint, put the settings of nmax and fmax to 0 (or None) in that else.

CIME/namelist.py
```diff
@@ -1214,6 +1214,7 @@ def write(
 
     def _write(self, out_file, groups, format_, sorted_groups):
         """Unwrapped version of `write` assuming that a file object is input."""
+        equals = None
```
billsacks: It seems like maybe there should be an `else` clause that aborts if `format_` doesn't match one of the expectations (following the conditional lines 1220-1223). Then this setting of `equals = None` could be moved into that `else` clause or put just before the start of that conditional block for clarity.

CIME/test_utils.py
```diff
@@ -26,6 +26,8 @@ def get_tests_from_xml(
     """
     listoftests = []
     testlistfiles = []
+    thismach = None
+    thiscompiler = None
```
billsacks: This fix seems okay, but I'd find the following fix more clear:

```diff
diff --git a/CIME/test_utils.py b/CIME/test_utils.py
index d4daa13149..84bb7bbea9 100644
--- a/CIME/test_utils.py
+++ b/CIME/test_utils.py
@@ -26,10 +26,6 @@ def get_tests_from_xml(
     """
     listoftests = []
     testlistfiles = []
-    if machine is not None:
-        thismach = machine
-    if compiler is not None:
-        thiscompiler = compiler
 
     if xml_testlist is not None:
         expect(
@@ -62,8 +58,12 @@ def get_tests_from_xml(
         for test in newtests:
             if machine is None:
                 thismach = test["machine"]
+            else:
+                thismach = machine
             if compiler is None:
                 thiscompiler = test["compiler"]
+            else:
+                thiscompiler = compiler
             test["name"] = CIME.utils.get_full_test_name(
                 test["testname"],
                 grid=test["grid"],
```

(I have checked, and this satisfies pylint.)

CIME/utils.py
```diff
@@ -2697,6 +2697,7 @@ def is_comp_standalone(case):
     """
     stubcnt = 0
     classes = case.get_values("COMP_CLASSES")
+    model = None
```
billsacks: I'm not sure about this change. It seems like this is relevant if the system is all stubs (or there is one non-stub? is this use of `numclasses - 2` instead of `numclasses - 1` accounting for the driver/mediator being an element of COMP_CLASSES?). It seems like returning `None` for `model` in this case could cause problems for callers. Should this perhaps return False in this case?? Or return some reasonable default for model??

billsacks (changes requested):
Thanks a lot for working on fixing these pylint errors, @jedwards4b . I ended up doing a pretty careful look through all of these fixes and have a number of suggestions and requests. In a couple of cases, the current fix doesn't feel right to me but I don't understand the expected usage well enough to know what the right fix seems to be off-hand. I'm happy to talk through any of this.

It seems like a lot of these fixes are needed because pylint doesn't understand that `expect(False, ...)` will abort. I spent a few minutes trying to give pylint more information through something like the `NoReturn` from the `typing` module, but couldn't get that to work. So it seems like the fixes here, which set some default value in the case where we're going to abort, may be the best we can do.

Thanks again - I appreciate your taking the time to get these pylint errors fixed!

CIME/case/case_setup.py
```diff
@@ -311,13 +311,14 @@ def _case_setup_impl(
...
         multi_driver = case.get_value("MULTI_DRIVER")
-        ninst = 0
         for comp in models:
             ntasks = case.get_value("NTASKS_{}".format(comp))
             if comp == "CPL":
                 continue
             if comp_interface != "nuopc":
                 ninst = case.get_value("NINST_{}".format(comp))
+            else:
+                ninst = 1
```
billsacks: I think this is the wrong place for this `else` clause to appear: I think this will override the ninst value for nuopc cases that is set on line 311. I think the correct placement of this else block is on line 312.

CIME/utils.py
```diff
@@ -2697,6 +2697,7 @@ def is_comp_standalone(case):
     """
     stubcnt = 0
     classes = case.get_values("COMP_CLASSES")
+    model = False
```
billsacks: I don't think setting `model = False` is the right thing here. I wonder if my earlier comment was worded confusingly (re-reading it, I think it was - sorry):

> I'm not sure about this change. It seems like this is relevant if the system is all stubs (or there is one non-stub? is this use of `numclasses - 2` instead of `numclasses - 1` accounting for the driver/mediator being an element of COMP_CLASSES?). It seems like returning `None` for `model` in this case could cause problems for callers. Should this perhaps return False in this case?? Or return some reasonable default for model??

When I suggested returning `False` in this case, I meant for the first element of the returned tuple, not for the `model` element of the tuple. I'm not sure what should be returned for the `model` element of the tuple in this case. I think the correct behavior probably depends on how this function is used, which I haven't looked into carefully. Whatever is done, this should probably be added to the function documentation.

billsacks (changes requested):
Thanks a lot for these follow-up changes. Most of them look good to me, but there are two that don't look right to me - see comments below.

CIME/case/case_run.py
```diff
@@ -305,19 +305,17 @@ def _post_run_check(case, lid):
...
     driver = case.get_value("COMP_INTERFACE")
 
-    comp_standalone, model = is_comp_standalone(case)
-
     if driver == "nuopc":
-        if comp_standalone:
-            file_prefix = model
-        else:
+        comp_standalone, file_prefix = is_comp_standalone(case)
+        if not comp_standalone:
```
billsacks, line 310: Thank you for this fix in case_run.py. I had been confused by the use of is_comp_standalone here when I looked at it last week, and your fix helps clarify this!

billsacks (approved):
Just marking this as re-approved after reviewing the latest set of changes. Thank you for your continued work on this @jedwards4b !

billsacks:
@jedwards4b - do you plan to also update the default pylint version? I realized that you said that this closes #4856 , but that issue is actually about updating the default pylint version. The changes you made here are important ones to enable that, but this doesn't actually close that issue unless we update the default pylint version.

billsacks:
I think this also resolves https://github.com/ESMCI/cime/issues/4858

PR #4920, opened 2026-01-16
Use different suffixes for the two ERI comparisons

### LINKED ISSUE #4896: Misleading failure message from ERI test
For a recent CESM alpha tag, we had the following failure in `TestStatus`:

```
FAIL ERI.TL319_t232.G_JRA.derecho_intel.cice-default COMPARE_base_hybrid
```

However, the issue was that one part of MOM6 was not initializing from a restart in a branch run... the hybrid run was fine.

### LINKED ISSUE #4912: ERI cprnc.out files from base-hybrid comparison are clobbered by the cprnc.out files from base-rest comparison
As noted [here](https://github.com/ESMCI/cime/issues/4911#issuecomment-3726141369) , the cprnc.out files from the base-hybrid comparison of the ERI test are clobbered by the cprnc.out files from the base-rest comparison (at least, I think that's what's happening). I've been aware of this issue (e.g., [here](https://github.com/ESCOMP/CTSM/pull/3299#issuecomment-3046410700)), but since it seems to repeatedly bite people, I think this is worth solving.

I can look into this.

billsacks:
## Description
Prior to this change, ERI cprnc.out files from the base-hybrid comparison were clobbered by the cprnc.out files from the base-rest comparison. This PR fixes this issue. It also renames the comparison steps and the cprnc.out files to try to be more intuitive - labeling these cprnc.out files with the name of the case that's being tested, either "branch" (for the branch-vs-hybrid test) or "rest" (for the rest-vs-branch test). The phases are now named `COMPARE_branch_hybrid` and `COMPARE_rest_branch`.

- Closes #4912 
- Closes #4896 

## Checklist

### DIFF
--- a/CIME/SystemTests/eri.py
+++ b/CIME/SystemTests/eri.py
@@ -302,13 +302,17 @@ class ERI(SystemTestsCommon):
         # do the restart run (short term archiving is off)
         self.run_indv(suffix="rest")
 
-        # Note that, for both of these comparisons, the "control" case comes first: the
-        # branch case is compared against the hybrid case (which it branched off of, and
-        # so serves as its "control"); the "rest" run is a restart from the branch case
-        # and so is compared against this branch case. The order of the two suffixes in
-        # each call isn't very important, but it *is* somewhat important that the first
-        # suffix differs between these two comparisons: otherwise, the cprnc output files
-        # from the second comparison overwrite the files from the first comparison (since
-        # the cprnc output file names are derived from the first suffix).
-        self._component_compare_test("hybrid", "branch")
-        self._component_compare_test("branch", "rest")
+        # Note that, for both of these comparisons, the "test" case comes first and the
+        # "control" case comes second: the branch case is compared against the hybrid case
+        # (which it branched off of, and so serves as its "control"); the "rest" run is a
+        # restart from the branch case and so is compared against this branch case. We
+        # make this choice because the cprnc output file names are derived from the first
+        # suffix, so:
+        # - Listing the "test" case as the first suffix means that the cprnc files are
+        #   named with the name of the case we're testing in that cprnc comparison, which
+        #   is more intuitive.
+        # - Having the first suffix differ between the two comparisons is important to
+        #   avoid having the cprnc output files from the second comparison overwrite the
+        #   files from the first comparison.
+        self._component_compare_test("branch", "hybrid")
+        self._component_compare_test("rest", "branch")


billsacks:
@ekluzek and @mnlevy1981 - this is now ready for review. See the edited description in the top-level comment in this PR. I have redone the testing that I noted above with the latest version of this code to ensure that an ERI passes when it should and fails in the correct comparison step when it should.

CIME/SystemTests/eri.py
```diff
@@ -302,5 +302,17 @@ def run_phase(self):
...
+        # restart from the branch case and so is compared against this branch case. We
+        # make this choice because the cprnc output file names are derived from the first
+        # suffix, so:
+        # - Listing the "test" case as the first suffix means that the cprnc files are
+        #   named with the name of the case we're testing in that cprnc comparison, which
+        #   is more intuitive.
+        # - Having the first suffix differ between the two comparisons is important to
+        #   avoid having the cprnc output files from the second comparison overwrite the
+        #   files from the first comparison.
+        self._component_compare_test("branch", "hybrid")
```
ekluzek, line 317: OK, so there will be two cprnc series created one lableled "branch" that compares the branch to hybrid. And one labeled "rest" that compares the restart along the final branch segment.

This makes sense. And it doesn't overwrite the cprnc.log file for "base" since the first name in both of these calls is different.

So this is good.

CIME/SystemTests/eri.py
```diff
@@ -302,5 +302,17 @@ def run_phase(self):
         # do the restart run (short term archiving is off)
         self.run_indv(suffix="rest")
```
ekluzek, line 303: I had to look at this carefully to figure it out enough. So this is mostly thinking out loud.

I found 4 instances of run_indv. The first one has None, then hybrid, branch and finally rest.

Since the first one isn't used for comparisons I suppose that's OK to use None. But, since there's so many cases here, I think it would be good to label it as "startup" or "base". Although I might be wrong and the test setup needs you to label the first case as None.

billsacks: Thanks a lot for your careful look at this, @ekluzek ! Yes, I agree: the ERI test is complicated enough that it takes some careful analysis to understand what's going on. I think I briefly considered adding a suffix to the ref1 case... I know that, in general, it works to give a suffix to the original case, but I think what held me back was a concern that I might break something in the ERI case - and specifically in the setup of the ref1 files as refcase files for the ref2 case - by adding a suffix. I'm pretty sure it could be made to work, but my initial thinking was that the benefits are small (since the ref1 case stays isolated from the other cases and isn't used in any comparisons) and the possible time and risks outweighed those small benefits. That said, I'm definitely open to digging into that if you or others think it's worth doing. Let me know how you feel about that.

ekluzek: It doesn't sound worth pursuing. So let's just bring this in as is.

PR #4929, opened 2026-02-09
Update container

jasonb5:
## Description

- Migrates from conda to spack
- Fix mixed perl environment (conda/system)
- Moves to UV to manage python
- Simplifies usage and testing workflow

## Checklist

### DIFF
--- a/docker/entrypoint.sh
+++ b/docker/entrypoint.sh
@@ -3,8 +3,6 @@
 # Set up basic user, logname, and default group/user IDs
 export USER="$(id -nu)"
 export LOGNAME="${USER}"
-export USER_ID="${USER_ID:-1000}"
-export GROUP_ID="${GROUP_ID:-1000}"
 
 # Set static home path where .cime exists and container entrypoint options
 SKIP_ENTRYPOINT="${SKIP_ENTRYPOINT:-false}"
@@ -34,13 +32,15 @@ function build_cprnc() {
 # Download input data needed for model setup
 # required for grid generation tests
 function download_input_data() {
-    mkdir -p "${HOME}/inputdata/cpl/gridmaps/oQU240" \
-        "${HOME}/inputdata/share/domains" \
-        "${HOME}/timings" \
-        "${HOME}/cases" \
-        "${HOME}/archive" \
-        "${HOME}/baselines" \
-        "${HOME}/tools"
+    local storage="${HOME}/storage"
+
+    mkdir -p "${HOME}/inputdata/cpl/gridmaps/oQU240"
+    mkdir -p "${HOME}/inputdata/share/domains"
+    mkdir -p "${storage}/cases"
+    mkdir -p "${storage}/timings"
+    mkdir -p "${storage}/archive"
+    mkdir -p "${storage}/baselines"
+    mkdir -p "${storage}/tools"
 
     wget -O "${HOME}/inputdata/cpl/gridmaps/oQU240/map_oQU240_to_ne4np4_aave.160614.nc" \
         https://portal.nersc.gov/project/e3sm/inputdata/cpl/gridmaps/oQU240/map_oQU240_to_ne4np4_aave.160614.nc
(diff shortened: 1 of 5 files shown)


docker/Dockerfile
```diff
@@ -1,46 +1,28 @@
-ARG BASE_TAG=latest
-FROM condaforge/miniforge3:${BASE_TAG} AS base
+ARG BASE_TAG=25.11.0-1
```
jedwards4b: Is this now something we will need to maintain and manually update?

docker/entrypoint.sh
```diff
@@ -21,38 +21,94 @@ function fix_arflags() {
...
         export ESMFMKFILE=/opt/conda/envs/cesm/lib/esmf.mk
 
-        ln -sf /home/cime/.cime/config_machines.v3.xml /home/cime/.cime/config_machines.xml
+        ln -sf ${1}/.cime/config_machines.v3.xml ${1}/.cime/config_machines.xml
     fi
+}
 
-    if [[ "${USER_ID}" == "0" ]]; then
-        cp -rf /home/cime/.cime /root/
+user_home=/home/cime
```
jedwards4b: Why not define this at line 62 and use it in 63 as well?

jedwards4b (approved):
There is a variable user_home which it seems like should replace the hardcoded
/home/cime throughout - but then you are also handling a case of root running so I'm a little lost as to the intent.

docker/Dockerfile
jasonb5: I'm going to move it back to latest, this was just for testing.

docker/entrypoint.sh
jasonb5: I'm still flushing this out. I need to roll in the CI specific stuff, I want to simplify the workflow itself.

docker/Dockerfile
```diff
@@ -1,46 +1,28 @@
...
 
 RUN mamba create -n cesm -y --file cime.yaml python=3.10 'libnetcdf>=4.9.2=*openmpi*' \
-    && conda remove -n cesm --force perl \
     && conda clean -afy \
     && rm -rf /opt/conda/pkgs/*
 
 COPY .cime .cime
 COPY entrypoint.sh /entrypoint.sh
 
+RUN bash -c "SKIP_ENTRYPOINT=true . /entrypoint.sh && download_input_data"
```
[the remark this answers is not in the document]
jasonb5: Added wget/curl/subversion to dockerfile.

docker/entrypoint.sh
```diff
@@ -21,37 +40,100 @@ function fix_arflags() {
...
+{
+    echo "source /opt/conda/etc/profile.d/conda.sh"
+    echo "conda activate base"
+    echo "if [[ ${CIME_MODEL} == 'e3sm' ]]; then"
+    echo "  conda activate e3sm"
+    echo "elif [[ ${CIME_MODEL} == 'cesm' ]]; then"
+    echo "  conda activate cesm"
+    echo "fi"
+    # Clobber PATH so system perl is used instead of conda perl
+    echo "export PATH=/usr/bin:\$PATH"
```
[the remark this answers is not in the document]
jasonb5: Used both hash and PERL variable.

docker/spack.yaml
```diff
@@ -0,0 +1,18 @@
+spack:
+  specs:
+    - hdf5@1.12.3+cxx+fortran+hl+mpi+shared %gcc@13.3.0
+    - netcdf-c@4.9.2+mpi %gcc@13.3.0
+    - netcdf-fortran@4.6.1 %gcc@13.3.0
+    - parallel-netcdf@1.12.3 %gcc@13.3.0
```
rljacob, line 6: @jayeshkrishna can you comment on these versions?

billsacks:
@jasonb5 - Thanks - I agree that the CESM failure is due to a CICE change - we just saw that in our recent testing as well without this CIME change.

PR #4944, opened 2026-02-27
Skip MOM6 .ic. history file comparison in multi-instance runs 

alperaltuntas:
Currently, `copy_histfiles` skips the mom6 initial conditions (ic) files for single-instance files, but not for multi-instance files. This PR fixes the logic by skipping these files for both single instance and multi instance runs.

- Closes https://github.com/NCAR/MOM6/issues/407

This PR should be evaluated in conjunction with:
https://github.com/ESCOMP/FMS/pull/8
https://github.com/NCAR/MOM6/pull/409
https://github.com/ESCOMP/MOM_interface/pull/311

## Checklist

### DIFF
--- a/CIME/hist_utils.py
+++ b/CIME/hist_utils.py
@@ -101,7 +101,7 @@ def copy_histfiles(case, suffix, match_suffix=None):
                 if "ocean_geometry" in test_hist:
                     comments += "    skipping '{}'\n".format(test_hist)
                     continue
-                if "mom6.ic" in test_hist:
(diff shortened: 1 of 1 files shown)


billsacks (approved):
This change looks good.

I'd love if we could find a way to move model-specific things like this to xml (e.g., the config_archive.xml file, where other regexes are given related to file name extensions). I'm also guilty of putting logic in hist_utils.py that ideally would be in xml - the specific check for "initial" elsewhere in hist_utils.py, which is needed for CISM - so I'm not meaning this as a criticism, just planting this thought.

PR #4946, opened 2026-03-03
Move xcpl_comps_nuopc to CMEPS

billsacks:
## Description

Move xcpl_comps_nuopc out of CIME. These will now be in CMEPS - see https://github.com/ESCOMP/CMEPS/pull/637

Also remove some things from gitignore that I think shouldn't be there

Depends on the changes in https://github.com/ESCOMP/CMEPS/pull/637 for X compset tests to pass in CESM.

- Closes https://github.com/ESMCI/cime/discussions/4943

## Checklist

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -23,10 +23,7 @@ scripts/Tools/JENKINS*
 # Ignore anything that are produced under scripts "cases" directory
 /scripts/cases/
 
-#Ignore Externals
-components
-libraries
-share
+# Ignore some other files
 test_coverage/**
 *.bak
 


billsacks:
I think the CESM system testing will pass once https://github.com/ESCOMP/CMEPS/pull/637 has been merged to the cesm alphabranch. And it should be safe to bring that CMEPS change in before this CIME change. So I think the order of business here should be:
- merge the CMEPS change to CMEPS main
- bring that tag into the CESM alphabranch
- rerun the failing GitHub tests here
- then bring in this CIME change

I think it will be fine for the CMEPS change to come into an earlier alpha tag than this CIME change – it would mean that the xcomps are defined in two places for that alpha tag, but I don't think that's a big deal.

.gitignore
```diff
@@ -23,10 +23,7 @@ scripts/Tools/JENKINS*
 # Ignore anything that are produced under scripts "cases" directory
 /scripts/cases/
 
-#Ignore Externals
-components
-libraries
-share
```
fischer-ncar, line 29: These were added back in the days we used manage_externals to checkout cime.  But since we are no longer using manage_externals it should be fine removing these externals.

billsacks: Thanks for confirming this - I thought that might be the case.

PR #4948, opened 2026-03-04
Flip cmake macro names to match case

jgfouca:
## Description

Flips cmake macro names to match case (use ${MACHINE}_${COMPILER}.cmake instead of the reverse).

This is related to E3SM PR: https://github.com/E3SM-Project/E3SM/pull/8129

I don't know why github is saying logger warnings that I thought were already on CIME master.

EDIT: Since ccs_config_cesm depends on the previous behavior, add a new model config setting `build_model_invert_macro_name`.

## Checklist

### DIFF
--- a/CIME/case/case_setup.py
+++ b/CIME/case/case_setup.py
@@ -152,6 +152,7 @@ def _create_macros_cmake(
     # This impl is coupled to contents of Macros.cmake
     os_ = mach_obj.get_value("OS")
     mach = mach_obj.get_machine_name()
+    deprecated = "{}_{}.cmake".format(compiler, mach)
     macros = [
         "universal.cmake",
         os_ + ".cmake",
@@ -159,6 +160,7 @@ def _create_macros_cmake(
         "{}_{}.cmake".format(compiler, os),
         mach + ".cmake",
         "{}_{}.cmake".format(mach, compiler),
+        deprecated,
         "CMakeLists.txt",
     ]
     for macro in macros:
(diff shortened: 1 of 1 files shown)


CIME/case/case_setup.py
```diff
@@ -150,15 +150,12 @@ def _create_macros_cmake(
...
     # This impl is coupled to contents of Macros.cmake
-    os_ = mach_obj.get_value("OS")
     mach = mach_obj.get_machine_name()
     macros = [
         "universal.cmake",
-        os_ + ".cmake",
         compiler + ".cmake",
-        "{}_{}.cmake".format(compiler, os),
         mach + ".cmake",
-        "{}_{}.cmake".format(compiler, mach),
```
jedwards4b, line 161: This change would require a corresponding and extensive change in the https://github.com/ESMCI/ccs_config_cesm/ repository.

jgfouca: I just poked around that repo. It looks like there's only a couple macros that need to be flipped. The bigger concern is that this repo appears to be using the OS concept in the macro hierarchy, which this PR drops support for.

CIME/case/case_setup.py
```diff
@@ -150,15 +150,12 @@ def _create_macros_cmake(
         os.mkdir(case_cmake_path)
 
     # This impl is coupled to contents of Macros.cmake
```
jedwards4b, line 152: The _os.cmake file was completely removed in this PR but it is still used in CESM.  For example https://github.com/ESMCI/ccs_config_cesm/blob/main/machines/cmake_macros/Darwin.cmake

CIME/case/case_setup.py
```diff
@@ -152,24 +152,35 @@ def _create_macros_cmake(
...
+        "{}_{}.cmake".format(mach, compiler),
+        deprecated,
         "CMakeLists.txt",
     ]
     for macro in macros:
         repo_macro = os.path.join(cmake_macros_dir, macro)
         mach_repo_macro = os.path.join(cmake_macros_dir, "..", mach, macro)
         case_macro = os.path.join(case_cmake_path, macro)
         if not os.path.exists(case_macro):
+            copied = False
```
jedwards4b, line 171: If I understand this addition - if the user provides the file locally in their case directory they don't get the warning?  Only if its in the repository - why?

jgfouca: That's correct. I think it's very unlikely that a user would add new macros to their case, but I suppose it's possible. Much more likely is that it will come from the repo.

PR #4949, opened 2026-03-06
Add cupid to preview namelists

### LINKED ISSUE #4939: Add run of buildnml for POSTPROCESSING if postprocssing is turned on and available
I'm mainly thinking of CUPiD here, but this could be useful for any post-processing type workflow. "preview_namelists" could check if postprocssing files are available, and if so and RUN_POSTPROCESSING is true and if so run the buildnml for the post-processing suite.

A CUPiD issue that talks about this is here:

https://github.com/NCAR/CUPiD/issues/371

### LINKED ISSUE #4942: Add creation of src.cupid for CESM
For CESM cases we'd like to have a $CASEROOT/SourceMods/src.cupid source mods directory created so that user  config YAML files could be used rather than the templates from CUPiD itself.

the issue on the CUPiD side is here:

https://github.com/NCAR/CUPiD/issues/378

Looking at the code on how to implement this.

CIME/case/case.py

creates the SourceMod directories in the _create_caseroot_sourcemods subroutine. 

Probably the way to do this would be to have an if statement that extends the component array with cupid if the model is CESM.

``` python
if case.get_value("MODEL") == "cesm":
  components.extend(["cupid"])
```

This is related to #4939

ekluzek:
## Description

If RUN_POSTPROCESSING is TRUE and the postprocessing spec file is setup for a case, and the buildnml script exists for it -- run it as part of preview_namelists.

- Closes #4939 
- Closes #4942

The CUPiD PR that goes along with this is here: https://github.com/NCAR/CUPiD/pull/395

These changes can come in first though and don't require the CUPiD change at the same time.

## Checklist

### DIFF
--- a/CIME/case/preview_namelists.py
+++ b/CIME/case/preview_namelists.py
@@ -128,19 +128,20 @@ def create_namelists(self, component=None):
         if have_postprocessing:
             config_dir = os.path.dirname(_postprocessing_spec_file)
             cmd = os.path.join(config_dir, "buildnml")
-            # NOTE: In order to name the specific postprocessing component, there would need to be
-            # a new XML variable with something like POSTPROC_NAME
-            compname = "postprocessor"
-            logger.info("Create namelist for postprocessing component {}".format(compname))
-            import_and_run_sub_or_cmd(
-                cmd,
-                (caseroot),
-                "buildnml",
-                (self, caseroot, compname),
-                config_dir,
-                compname,
-                case=self,
-            )
+            if ( os.path.exists(cmd) ):
+                # NOTE: In order to name the specific postprocessing component, there would need to be
+                # a new XML variable with something like POSTPROC_NAME
+                compname = "postprocessor"
+                logger.info("Create namelist for postprocessing component {}".format(compname))
+                import_and_run_sub_or_cmd(
+                    cmd,
+                    (caseroot),
+                    "buildnml",
+                    (self, caseroot, compname),
+                    config_dir,
+                    compname,
+                    case=self,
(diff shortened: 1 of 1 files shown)


CIME/case/preview_namelists.py
```diff
@@ -109,6 +109,39 @@ def create_namelists(self, component=None):
...
+        # Run buildnml for the post-processing component if buildnml exists
+        # NOTE: This copies the handling of buildnml for components in the section above
+        # TODO: Make it into a little subroutine?
+        if have_postprocessing:
+            config_dir = os.path.dirname(_postprocessing_spec_file)
+            cmd = os.path.join(config_dir, "buildnml")
+            if ( os.path.exists(cmd) ):
+                # NOTE: In order to name the specific postprocessing component, there would need to be
+                # a new XML variable with something like POSTPROC_NAME
+                compname = "postprocessor"
```
ekluzek, line 134: We plan to add POSTPROCESSING_TOOL_NAME to the CUPiD confit_tool.xml file which will set the specific name of the postprocessing tool (just like components set their names). Here we will get that variable and then set compname to it.

CIME/case/preview_namelists.py
```diff
@@ -128,6 +161,7 @@ def create_namelists(self, component=None):
         "nuopc.runconfig",
         "*streams*txt*",
         "*streams.xml",
+        "*.yml",
```
ekluzek, line 164: The CUPiD config yaml files should be copied to the CaseDocs subdirectory to keep the record of what they are doing.

PR #4981, opened 2026-05-19
wait_for_tests: Improve cdash description

jgfouca:
This info was already in notes, but it was easy to miss. With this PR, all you have to do is click the dropdown to expand the build information, and you'll see this info (the commit that was tested and the time it took).


jasonb5 (approved):
LGTM

Issue #4950, opened 2026-03-13
Refactor and API changes


jasonb5:
This will happen in small stages. May require external script updates.

Scripts will stay in place, will not break current usage.

## New internal structure

`CIME/api/` -> User facing classes e.g. Case, Downloader, etc
`CIME/cli/`
`CIME/core/`
`CIME/core/build/`
`CIME/core/baseline/`
`CIME/core/batch/`
`CIME/core/config/`
`CIME/core/compare/`
`CIME/core/downloader/`
`CIME/core/locking/`
`CIME/core/mods/`
`CIME/core/namelist/`
`CIME/core/status/`
`CIME/core/timing/`
`CIME/core/tools/`
`CIME/core/xml/`
`CIME/data/`
`CIME/non_py`
`CIME/build_scripts`
`CIME/SystemsTests/`
`CIME/tests/`

## Restructuring

`CIME/BuildTools/configure.py` -> `CIME/core/build/configure.py`
`CIME/ParamGen/` -> `CIME/core/tools/paramgen/`
`CIME/Servers/` -> `CIME/core/downloader/`
`CIME/SystemTests` will be retained, some logic will be moved into `CIME/core/`
`CIME/baselines/` -> `CIME/core/baseline/`
`CIME/build_scripts/` will be retained, logic moves to `CIME/core/build/`
`CIME/case` -> `CIME/core/case.py` multiple files consolidated, split out core logic, will be thin wrapper
`CIME/scripts` will be retained, all logic moved to core
`CIME/aprun.py` -> `CIME/core/batch/`
`CIME/bless_test_results.py` will retain, move logic into core.
`CIME/build.py` -> `CIME/core/build/make.py`
                -> `CIME/core/build/cmake.py`
                -> `CIME/core/build/libraries.py`
`CIME/buildlib.py` -> `CIME/core/build/`
`CIME/buildnml.py` -> `CIME/core/namelist/`
`CIME/code_checker.py` remove, no longer used
`CIME/compare_namelists.py` -> `CIME/core/namelist/`
`CIME/compare_test_results.py` -> `CIME/core/baseline/`
`CIME/config.py` -> `CIME/core/config`
`CIME/cs_status.py` -> `CIME/core/status/case.py`
`CIME/cs_satus_creator.py` -> `CIME/core/status/case.py`
`CIME/date.py` -> `CIME/core/date.py`
`CIME/expected_fails.py` -> `CIME/core/status/expected.py`
[...]

PR #4982, opened 2026-05-27
Refactor core foundation (exceptions and bootstrap)

jasonb5:
## Description

- Adds `CIME/core/exceptions.py` with typed exception hierarchy extending `CIMEError`
- Adds `CIME/core/config/bootstrap.py` for centralized sys.path management
- Migrates internal `CIMEError` imports to use new location
- Adds 26 unit tests for exception hierarchy

## Motivation

CIME's foundation has accumulated two recurring pain points that this PR addresses as the first slice of a broader `CIME/core/` refactor:

**Typed exceptions.** Today there is a single `CIMEError` in `CIME/utils.py`, so callers can't distinguish a configuration problem from a build failure from a transient subprocess error without string-matching on messages. The new hierarchy (`ConfigurationError`, `BuildError`, `SubmitError`, `SystemTestError`, `LockingError`, `ExternalCommandError`, `RetryableExternalCommandError`, `InputError`) enables:
- Granular `except` clauses in callers (e.g. retry only transient subprocess errors)
- Clearer log and traceback context
- `ExternalCommandError` modeled on `subprocess.CalledProcessError`, carrying `returncode`, `cmd`, `output`, and `stderr` so `run_cmd` failures preserve diagnostic context

`CIME.utils.CIMEError` is re-exported from the new location so existing imports (internal or downstream) continue to work.

**Centralized sys.path bootstrap.** CIME is not pip-installed; every entry script (`xmlchange`, `case.setup`, `create_newcase`, etc.) does its own `sys.path.insert` to make `import CIME...` work. This duplicated logic has drifted in subtle ways. `bootstrap_cime()` provides:
- A single source of truth for path setup
- Three-way CIMEROOT resolution (explicit arg, env var, walk-up from `__file__`)
- Idempotent insertion (no duplicates, deterministic ordering)
- The Python version check previously in `standard_script_setup.py`

Existing call sites are intentionally unchanged in this PR — the module is standalone so subsequent slices can adopt it incrementally.

**Why these two pieces first.** They are dependency-free building blocks that the planned `CIME/core/` DI-based abstractions (filesystem, process, env, clock) will rely on, so they need to land before the higher-level work.

## Related

- Part of #4996 (Slice 1: Foundation — typed exceptions, bootstrap, kill star imports). This PR delivers the typed-exception hierarchy and the bootstrap module; the star-imports cleanup, `GLOBAL` removal, and conftest fix remain.
- Part of #4950 (refactor master tracker).

## Checklist


jgfouca (approved):
Looks good to me. Can you add motivation for this change in the description?

CIME/core/config/bootstrap.py
```diff
@@ -0,0 +1,150 @@
...
+
+    _ensure_paths(paths_to_add)
+
+    if set_env:
+        os.environ["CIMEROOT"] = root
+
+    return root
+
+
+def _ensure_paths(paths: Sequence[str]) -> None:
```
billsacks: The name of this function implies checking, not modification of sys.path. I was confused when reading `bootstrap_cime`, wondering, "where does the actual modification of sys.path happen?" It would be more clear if this were renamed to something like `_add_paths_to_sys_path`

CIME/core/config/bootstrap.py
```diff
@@ -0,0 +1,150 @@
...
+
+    Paths already present are moved to the correct position rather than
+    duplicated.
+    """
+    for i, p in enumerate(paths):
+        absp = os.path.abspath(p)
+        # Remove existing entry if present (to re-position)
+        if absp in sys.path:
+            sys.path.remove(absp)
+        sys.path.insert(i, absp)
```
billsacks, line 149: This might be a weird edge case, and the consequences are probably minor, but I noticed that, with this algorithm, things will work in an unexpected way if there are duplicates in the input `paths` (e.g., if `extra_paths` in `bootstrap_cime` somehow included the path to cime itself).

For example, imagine that `paths = ["/my/path", "/my/path"]`. The first time through the loop, "/my/path" would get inserted at the front, as intended. The second time through, the first "/my/path" would be removed, and then the second would be inserted as the *second* list element, after whatever the first element was in the pre-existing sys.path. If there were even more duplicates, each subsequent time through the loop would result in the path being inserted one element further along the pre-existing sys.path, until - in the extreme case - the new element would actually appear at the end rather than at the start of sys.path.

Although this might be a paranoid edge case, it seems like the solution is simple: change the loop to `for p in reversed(paths):` and then change the insertion to `sys.path.insert(0, absp)`. (This probably warrants a comment explaining the reason for the `reversed`: this is done so that order in `sys.path` matches the order in `paths`, given that each element is prepended to the current sys path.)

CIME/core/__init__.py
```diff
@@ -0,0 +1,7 @@
+"""
+CIME core package — DI-based abstractions for filesystem, process, environment,
```
billsacks: Is DI Dependency Injection? Probably worth spelling out.

CIME/tests/test_unit_core_bootstrap.py
```diff
@@ -0,0 +1,137 @@
...
+
+    def teardown_method(self):
+        sys.path[:] = self._orig_path
+
+    def test_inserts_in_order(self, tmp_path):
+        a = str(tmp_path / "a")
+        b = str(tmp_path / "b")
+        _ensure_paths([a, b])
+        assert sys.path[0] == a
+        assert sys.path[1] == b
```
billsacks, line 115: (Optional) Another assertion that would be helpful here is that the remaining entries of sys.path match the original sys.path. (i.e., that `_ensure_paths` didn't inadvertently remove anything that it shouldn't have.)

billsacks (changes requested):
I like the direction you're taking things here - thank you for this work! I have reviewed the production code and did a cursory look at the unit tests. I have a few comments. I appreciate the added testing here.

CIME/core/config/bootstrap.py
jasonb5: Renamed to `_prepend_sys_path`.

CIME/core/config/bootstrap.py
jasonb5: @billsacks Confirmed this is a bug. Reversing the list won't work, see the example below. I think just deduping `paths` (first-occurrence wins) should work, adding a warning as well.
```
sys.path = ["A", "B", "C"]
paths = ["C", "D", "C"]
sys.path = ["C", "A", "B"]
sys.path = ["C", "D", "A", "B"]
sys.path = ["D", "A", "C", "B"]
```

billsacks: I may be thinking about this wrong, but I think your example didn't capture my suggestion: note that, in addition to reversing the list, I also suggested changing the insertion to do a prepend, via `sys.path.insert(0, absp)`. I *think* this would result in:

```
sys.path = ["A", "B", "C"]
paths = ["C", "D", "C"]
sys.path = ["C", "A", "B"]
sys.path = ["D", "C", "A", "B"]
sys.path = ["C", "D", "A", "B"]
```

Your solution of first deduplicating the paths seems reasonable, too. It's more complex, but does add the ability to give a warning for duplicated paths. I'm fine with either approach.

billsacks (approved):
Thank you for the fixes! See the comment I made about the algorithm for inserting paths, but as I note there, I'm fine with either approach. So I'm approving here and will let you decide which way to go.

CIME/core/config/bootstrap.py
jasonb5: @billsacks Missed that detail, that would make it work. Probably better to keep the method that can warn for user interface.

billsacks: > @billsacks Missed that detail, that would make it work. Probably better to keep the method that can warn for user interface.

Sounds good.

jasonb5:
Test failures are unrelated.

PR #4989, opened 2026-06-03
compare_namelists.py: allow comparing multiple namelist groups with same name

### LINKED ISSUE #4987: Namelist groups with the same group name will fail NLCOMP
CIME's `compare_namelists.py` fails if there are multiple namelist groups with the same name:
```
&hist_file_config_nl
    hist_volume = 'h1'
    hist_inst_fields = 'T', 'Q'
    hist_max_frames = 30
    hist_output_frequency = '1*nsteps'
    hist_precision = 'REAL32'
    hist_file_type = 'history'
    hist_filename_spec = '%c.cam.%u%f.%y-%m-%d-%s.nc'
    hist_write_nstep0 = .true.
/

&hist_file_config_nl
    hist_volume = 'h2'
    hist_inst_fields = 'U', 'V'
    hist_max_frames = 30
    hist_output_frequency = '1*nsteps'
    hist_precision = 'REAL32'
    hist_file_type = 'history'
    hist_filename_spec = '%c.cam.%u%f.%y-%m-%d-%s.nc'
    hist_write_nstep0 = .true.
/
```

The error is
```
CIMEError: ERROR: In file 'atm_in', Duplicate namelist 'hist_file_config_nl'
```

The Fortran standard is not very clear on this, but it is not prohibited and it is inferred that the read pointer continues after the last group read, so the group name being a duplicate should not be an issue. It is just read in sequentially. A pull request will be submitted to handle this case.

Companion host model side issue: https://github.com/ESCOMP/CAM-SIMA/issues/430

Thanks!

jimmielin:
## Description
This changes compare_namelists.py changes to support comparing multiple namelist groups of the same name. This namelist configuration is used in CAM-SIMA for separate history tapes but currently triggers a NLCOMP failure when multiple tapes are specified. Please refer to issue.

- Closes #4987

## Checklist
- [ ] I have made corresponding additions and changes to the documentation - not applicable unless there's somewhere documenting this behavior?

### DIFF
--- a/CIME/compare_namelists.py
+++ b/CIME/compare_namelists.py
@@ -288,11 +288,10 @@ def _parse_namelists(namelist_lines, filename):
             # same name (read sequentially). Use positional suffixes to
             # distinguish them while preserving order for comparison.
             if current_namelist in rv:
-                base_name = current_namelist.split("__")[0]
                 count = 2
-                while "{}__{}".format(base_name, count) in rv:
+                while "{}__{}".format(current_namelist, count) in rv:
                     count += 1
-                current_namelist = "{}__{}".format(base_name, count)
+                current_namelist = "{}__{}".format(current_namelist, count)
 
             rv[current_namelist] = {}
 
@@ -651,11 +650,45 @@ def _compare_namelists(gold_namelists, comp_namelists, case):
       BASE: csw_specifier dict item DMS = 1.0*value.nc
       COMP: csw_specifier dict item DMS = 1.0*other.nc
     <BLANKLINE>
+
+    >>> gold = '''&nml
+    ... val = 1
+    ... /
+    ... &nml
+    ... val = 2
+    ... /
+    ... &nml
+    ... val = 3
+    ... /'''
+    >>> comp = '''&nml
+    ... val = 1
+    ... /
+    ... &nml
+    ... val = 20
+    ... /'''
+    >>> comments = _compare_namelists(_parse_namelists(gold.splitlines(), 'g'),\
+    _parse_namelists(comp.splitlines(), 'c'), None)
+    >>> print(comments)
+      BASE: val = 2
+      COMP: val = 20
+    Missing namelist: nml (occurrence 3)
+    <BLANKLINE>
     """
+
+    def _display_name(namelist):
+        """Translate internal 'nml__3' to 'nml (occurrence 3)'."""
+        if "__" in namelist:
+            parts = namelist.rsplit("__", 1)
+            if parts[1].isdigit():
+                return "{} (occurrence {})".format(parts[0], parts[1])
+        return namelist
+
     different_namelists = {}
     for namelist, gold_names in gold_namelists.items():
         if namelist not in comp_namelists:
-            different_namelists[namelist] = ["Missing namelist: {}\n".format(namelist)]
+            different_namelists[namelist] = [
+                "Missing namelist: {}\n".format(_display_name(namelist))
+            ]
         else:
             comp_names = comp_namelists[namelist]
             for name, gold_value in gold_names.items():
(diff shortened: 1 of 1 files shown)


billsacks (approved):
Sorry for letting this one slip through the cracks! This looks great to me, and thanks for the added tests to verify and illustrate the new functionality!

billsacks:
@jasonb5 - would you like to review this so it has some E3SM eyes on it (or assign it to someone else from E3SM to reviiew)? From my review, it looks like this is behavior-preserving in existing cases... it just allows a situation that previously caused an error in compare_namelists.

ekluzek:
@jimmielin the thing about this is that I've always felt that having the same namelist name in the same file is really bad to do. So I've used the limitation to make sure we don't have the same name used multiple times in the same file. The order that they will be read can change with how initialization is done, and it's just easy to get it confused. In CTSM at least the infrastructure doesn't allow this to happen. And CAM didn't used to allow this either. Has this changed in CAM-SIMA? Can you give some of your thinking about why this is important in CAM-SIMA?

jimmielin:
Thanks @ekluzek! 

Yes, this has changed in CAM-SIMA. CAM-SIMA uses a new history tape configuration system where the user-facing format is a set of `hist_*` directives that are parsed by Python at build time and post-processed into `atm_in`, generating one `&hist_file_config_nl` group per history tape:

For example, a user writes in `user_nl_cam`:
```
hist_add_avg_fields;h0:QRL

hist_add_inst_fields;h1: T, U, V, PS
hist_output_frequency;h1: 2*nsteps
hist_max_frames;h1: 24
```

The Python post-processor then generates in `atm_in`:
```
&hist_file_config_nl
 hist_volume = 'h0'
 hist_avg_fields = 'QRL'
 hist_output_frequency = '1*nmonths'
/

&hist_file_config_nl
 hist_volume = 'h1'
 hist_inst_fields = 'T', 'U', 'V', 'PS'
 hist_output_frequency = '2*nsteps'
 hist_max_frames = 24
/
```

Each group shares the same namelist group name, but is self-identifying (`hist_volume = 'h0'`, `'h1'`, etc.) so the order in which they are read does not matter since the Fortran reader on the model side can handle it.

I do agree that repeated same-named groups can be a source of confusion. The Fortran standard is just silent on this (it just sets rules that namelist groups are read sequentially, so it only implies that the behavior is not prohibited.) I am not entirely sure of the reason of the switch to the new format (it predates me joining the project) but @peverwhee, @nusbaume may be able to chime in on this!

gold2718:
> I am not entirely sure of the reason of the switch to the new format (it predates me joining the project)

That was me. The idea was to migrate from the inscrutable syntax of CAM (and CLM) history (e.g., `nhtfrq`? `mflit`?, counting history file numbers to make sure all the array items are lined up) into something easy to use and human readable both in user_nl_cam and in atm_in.

This syntax is certainly standard Fortran as pointed out by @jimmielin .

This is what I came up with, I'm sure the world is open to better ideas (I am not a UX designer).

ekluzek:
@jimmielin and @gold2718 thanks for the explainations.

> Thanks @ekluzek!
> 
> Yes, this has changed in CAM-SIMA. CAM-SIMA uses a new history tape configuration system where the user-facing format is a set of `hist_*` directives that are parsed by Python at build time and post-processed into `atm_in`, generating one `&hist_file_config_nl` group per history tape:
> 
> For example, a user writes in `user_nl_cam`:
> 
.
.
.
> ```
> 
> Each group shares the same namelist group name, but is self-identifying (`hist_volume = 'h0'`, `'h1'`, etc.) so the order in which they are read does not matter since the Fortran reader on the model side can handle it.
> 
 
The self-identifying aspect here seems sufficiently robust to me that I think this is OK. I see the point in this case of keeping the namelist name the same. And obviously the namelist handling has changed both writing the namelist files and reading them have changed from what it used to be. And I don't think I have a suggestion on how to improve this. @gold2718 is right there is a lot of confusion about the history options that are done through seperate arrays. I think this might be better? It'll be interesting to hear what people think about this interface.

ekluzek (approved):
LGTM, now that I had my questions answered.

PR #4991, opened 2026-06-04
Add ability to offload CIME builds to batch system

jgfouca:
## Description

We've been having issues with our CIME test suite builds oversubscribing the login nodes on some machines. I believe it's better to do builds on compute nodes if possible.

This feature was tested on mappy with an E3SM branch with the same name as this one and it seemed to work well.

Changes:
1) Main change: adds ability for build.py to do batched builds via `_submit_build_as_batch`.
2) Add batched_build_active / no_batch_build to case.build. This is to avoid an infinite recursion of batch submissions.
3) Add new config_machines optional entry `BATCHED_BUILD` which defaults to False.

Fixes:
1) env_batch had a somewhat serious bug involving get_status and cancel_job where only the last batch_system block was being checked for settings. This meant any settings inherited from the generic batch_system block were ignored.
2) In build_impl, need to ensure that build problems leave the system in a state the reflects the failure of the build (BUILD_COMPLETE should be False and BUILD_STATUS should be something other than zero).

## Checklist

### DIFF
--- a/CIME/Tools/case.build
+++ b/CIME/Tools/case.build
@@ -228,8 +228,6 @@ def _main_func(description):
         no_batch_build,
     ) = parse_command_line(sys.argv, description)
 
-    # If --no-batch-build is specified, pass batched_build_active=True to
-    # build.case_build so it skips batch submission even when BATCHED_BUILD=TRUE.
     success = True
     with Case(caseroot, read_only=False, record=True) as case:
         testname = case.get_value("TESTCASE")
--- a/CIME/Tools/case.build
+++ b/CIME/Tools/case.build
@@ -86,6 +86,13 @@ def parse_command_line(args, description):
         )
 
     parser.add_argument(
+        "--no-batch-build",
+        action="store_true",
+        help="Skip batch submission even if BATCHED_BUILD is TRUE for this machine. "
+        "Use this when you are already on a compute node or want to build interactively.",
+    )
+
+    parser.add_argument(
         "--skip-submit",
         action="store_true",
         help="Sets the current test phase to RUN, skipping the SUBMIT phase. This "
@@ -197,6 +204,7 @@ def parse_command_line(args, description):
         args.ninja,
         args.dry_run,
         args.skip_submit,
+        args.no_batch_build,
     )
 
 
@@ -217,7 +225,13 @@ def _main_func(description):
         ninja,
         dry_run,
         skip_submit,
+        no_batch_build,
     ) = parse_command_line(sys.argv, description)
+
+    # If --no-batch-build is specified, signal to build.py to skip batch submission
+    # even if BATCHED_BUILD=TRUE is set for this machine.
+    if no_batch_build:
+        os.environ["CIME_BATCHED_BUILD_ACTIVE"] = "TRUE"
 
     success = True
     with Case(caseroot, read_only=False, record=True) as case:
--- a/CIME/Tools/case.build
+++ b/CIME/Tools/case.build
@@ -228,11 +228,8 @@ def _main_func(description):
         no_batch_build,
     ) = parse_command_line(sys.argv, description)
 
-    # If --no-batch-build is specified, signal to build.py to skip batch submission
-    # even if BATCHED_BUILD=TRUE is set for this machine.
-    if no_batch_build:
-        os.environ["CIME_BATCHED_BUILD_ACTIVE"] = "TRUE"
-
+    # If --no-batch-build is specified, pass batched_build_active=True to
+    # build.case_build so it skips batch submission even when BATCHED_BUILD=TRUE.
     success = True
     with Case(caseroot, read_only=False, record=True) as case:
         testname = case.get_value("TESTCASE")
@@ -283,6 +280,7 @@ def _main_func(description):
                 dry_run=dry_run,
                 separate_builds=separate_builds,
                 skip_submit=skip_submit,
+                batched_build_active=no_batch_build,
             )
 
         else:
@@ -296,6 +294,7 @@ def _main_func(description):
                 separate_builds=separate_builds,
                 ninja=ninja,
                 dry_run=dry_run,
+                batched_build_active=no_batch_build,
             )
 
     sys.exit(0 if success else 1)
--- a/CIME/build.py
+++ b/CIME/build.py
@@ -1169,10 +1169,10 @@ def _submit_build_as_batch(
 
     while True:
         status = env_batch.get_status(jobid)
-        if status is not None:
+        if status is not None and jobid in status:
             job_seen_in_queue = True
             not_seen_count = 0
-            logger.info(
+            logger.debug(
                 "Build job {} in queue (status: {}). Next check in {} s.".format(
                     jobid, status.strip(), poll_interval
                 )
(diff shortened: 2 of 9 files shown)


CIME/data/config/xml_schemas/config_machines_template.xml
```diff
@@ -72,6 +72,12 @@
     <!-- GMAKE_J: optional number of threads to pass to the gmake flag -->
     <GMAKE_J>8</GMAKE_J>
 
+    <!-- BATCHED_BUILD: optional, if TRUE case.build will submit the build
+      as a batch job instead of building on the current (login) node.
+      Requires a batch system to be configured (BATCH_SYSTEM != none).
+      Default is FALSE. -->
+    <!-- <BATCHED_BUILD>FALSE</BATCHED_BUILD> -->
```
bartgol, line 79: Was this line meant to be uncommented?

bartgol (approved):

CIME/Tools/case.build
```diff
@@ -217,8 +225,11 @@ def _main_func(description):
         ninja,
         dry_run,
         skip_submit,
+        no_batch_build,
     ) = parse_command_line(sys.argv, description)
 
+    # If --no-batch-build is specified, pass batched_build_active=True to
+    # build.case_build so it skips batch submission even when BATCHED_BUILD=TRUE.
```
billsacks: This comment seems far-removed from the lines it's referring to. Particularly given the unintuitive nature of the lines it's referring to (where batched_build_active is set to no_batch_build), I'd be inclined to move the comment down to those lines - even though it means duplicating the comment in two places.

CIME/build.py
```diff
@@ -1085,12 +1085,140 @@ def _clean_impl(case, cleanlist, clean_all, clean_depends):
...
+    )
+
+    jobid = depid.get("case.build") if depid else None
+    expect(jobid, "Batch build job submission did not return a job ID.")
+
+    logger.info(
+        "Build submitted as batch job {}. Polling for completion...".format(jobid)
+    )
+
+    poll_interval = 30
```
billsacks, line 1170: If I understand right, this means that there will be up to a 30 second delay between the completion of the build and whatever is happening next. Is that right? 30 seconds feels like a long delay, particularly if this kicks in for a rebuild that might otherwise just take a few seconds. But are there problems with shortening it?

billsacks (commented):
This seems like a great idea, and from my quick skim of the implementation, it looks great to me (and I appreciate the comments throughout!) - though I haven't done a careful review.

For CESM, we've handled this situation by requiring users to explicitly submit the build through a batch job (and there's an easy-to-use wrapper script on our main system that facilitates that), but this seems like a cleaner mechanism.

If there are multiple jobs in a create_test invocation, does this new mechanism lead to simultaneous builds or would they still be done sequentially?

I'm also curious how hard it would be to force a batch build using this mechanism on a system that didn't have BATCHED_BUILD set in config_machines - e.g., if you don't usually want to do a batched build on a given machine but might want to under some conditions. I'm not requesting that for this PR (it can come later if we want it), but I'm curious how hard that would be to add later.

I have a couple of inline comments, though I don't feel strongly about them.

CIME/build.py
billsacks: > If I understand right, this means that there will be up to a 30 second delay between the completion of the build and whatever is happening next. Is that right? 30 seconds feels like a long delay, particularly if this kicks in for a rebuild that might otherwise just take a few seconds. But are there problems with shortening it?

Actually, I heard from others that we've had problems in the past with even having 30-second polling of the batch system - that this polling has in the past caused slowdowns in the batch system. Now, on our main system, there is apparently caching in place to prevent this kind of issue, but which also could mean that more frequent polling also may not be effective because the cache could be stale (I'm not sure of this... I don't know the details of how the caching works). Anyway, maybe 30 seconds is a reasonable compromise after all.

CIME/data/config/xml_schemas/config_machines_template.xml
jgfouca: No, it defaults to FALSE.

jgfouca:
@billsacks ,

> If there are multiple jobs in a create_test invocation, does this new mechanism lead to simultaneous builds or would they still be done sequentially?

Yes, they will be done in parallel on multiple compute nodes. This should speed up the build process. Before I integrate into E3SM, I will test the speed up.

> I'm also curious how hard it would be to force a batch build using this mechanism on a system that didn't have BATCHED_BUILD set in config_machines - e.g., if you don't usually want to do a batched build on a given machine but might want to under some conditions. I'm not requesting that for this PR (it can come later if we want it), but I'm curious how hard that would be to add later.

Good question. For now, I think you'd just need to xmlchange BATCHED_BUILD to TRUE after you've set up the case.

billsacks:
> they will be done in parallel on multiple compute nodes

Thanks for clarifying. That seems good in many respects, though I can't remember: do you know how that would interact with sharedlib builds, where the same build is shared across multiple cases in a create_test job?

> Good question. For now, I think you'd just need to xmlchange BATCHED_BUILD to TRUE after you've set up the case.

Thanks!

jgfouca:
@billsacks , I think create_test/test_scheduler knows to wait for the sharedlib build.

CIME/build.py
```diff
@@ -1085,13 +1085,155 @@ def _clean_impl(case, cleanlist, clean_all, clean_depends):
...
 
     # reset following values in xml files
-    case.set_value("SMP_BUILD", str(0))
-    case.set_value("NINST_BUILD", str(0))
-    case.set_value("BUILD_STATUS", str(0))
-    case.set_value("BUILD_COMPLETE", "FALSE")
+    case.set_value("SMP_BUILD", 0)
+    case.set_value("NINST_BUILD", 0)
+    case.set_value("BUILD_STATUS", 0)
+    case.set_value("BUILD_COMPLETE", False)
```
jasonb5: This needs to be changed back. The testing is failing with `ERRPUT: ERROR: Wrong type for entry id 'SMP_BUILD'`.

jasonb5 (changes requested):
The code looks good, just the one change to resolve failing tests. I'm going to test this on a machine locally and will report back. Thanks!

billsacks:
> I think create_test/test_scheduler knows to wait for the sharedlib build.

Thanks for that reminder!

jasonb5:
@jgfouca Can you also document this feature.

CIME/build.py
jgfouca: Ah, I see that SMP_BUILD and NINST_BUILD are both defined as type `char`. The documentation says these are "coded strings"... what does that mean? Seems like they should be integers.

jgfouca:
@jasonb5 , where should I add documentation?

jasonb5:
I would update the this [section](https://github.com/ESMCI/cime/blob/master/doc/source/ccs/model-configuration/variables/machine.rst) with enabling it for a machine and this [section](https://github.com/ESMCI/cime/blob/master/doc/source/ccs/building-a-case.rst) with user usage.

I tried running this locally and added `<BATCHED_BUILD>TRUE</BATCHED_BUILD>` to the machine config. Looks like neither `create_test` or `create_newcase` ever set the XML variable `BATCHED_BUILD`, which results in the build never being submitted to the batch system and `xmlquyer`/`xmlchange` cannot manipulate the variable.

jgfouca:
@jasonb5 , you have to add it as a env XML var too. If you look at the branch `jgfouca/build_on_compute` for E3SM, you can see some additional steps were needed for this to work. The host model needs a .case.build workflow, changes to config_machines.xml to add BATCHED_BUILD=TRUE, a template.case.build file, and driver env XML changes to add BATCHED_BUILD as a known env XML entry.

jgfouca:
@jasonb5 , I have added documentation.

jasonb5:
@jgfouca Ok I see so it's pretty involved on the model side, we should document this all too. Adding a new document (batched_build.rst) under `https://github.com/ESMCI/cime/tree/master/doc/source/ccs/model-configuration` might be a good place. If you'd like I could put it together.

rljacob:
If this is E3SM specific, we should probably put it in the E3SM docs.

jgfouca:
@rljacob , this feature will work on any model if the prereqs I mentioned earlier are met.

jgfouca:
@jasonb5 , I tried to document the model prereqs a bit in building-a-case.rst . Let me know if more is needed.

jgfouca:
I'm eager to begin experimenting with this for E3SM, merging.

PR #4994, opened 2026-06-16
A couple minor fixes

jgfouca:
## Description

This reverts a PR from a couple months ago that added the build time and sha to the cdash build description (it was already being added to notes). What I did not realize is that this updated the description for all builds, not just the most recent one, so the information was incorrect for all but the most recent build. The notes file does not have this issue, so lets just use that like we were doing before.

Also, a minor upgrade to the widely-used `expect` function. The prior implementation only worked when `error_msg` was a `str`. By using `f"` instead of `+`, `expect` will now work with any stringifiable object.

## Checklist


billsacks (approved):

jasonb5 (approved):
LGTM

PR #4995, opened 2026-06-17
For share-exe test suites, allow specification of which case to use

jgfouca:
This will allow for test suites to specify which case's build to use as the shared build, which could be important if you have a suite with some tests that have a generic, share-able build and others that do not.

## Checklist

### DIFF
--- a/CIME/get_tests.py
+++ b/CIME/get_tests.py
@@ -323,12 +323,13 @@ def get_build_groups(tests):
     >>> get_build_groups(tests2)
     [('SMS_P8.f45_g37.A.melvin_gnu', 'SMS_P2.f45_g37.A.melvin_gnu', 'SMS_P4.f45_g37.A.melvin_gnu', 'SMS_P16.f45_g37.A.melvin_gnu')]
     """
-    build_groups = []  # list of tuples ([tests], set(suites))
-
+    #
     # Get share suites and their designated build leaders (if any)
+    #
     suites = get_test_suites()
-    share_suites = []
-    share_suite_leaders = {}  # suite -> short test name for string share values
+    share_suites = []  # list of suites with exe sharing on
+    # suite -> short test name for specified leader if provided
+    share_suite_leaders = {}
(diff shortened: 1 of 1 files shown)


CIME/get_tests.py
```diff
@@ -311,7 +359,31 @@ def get_build_groups(tests):
...
 
-    return [tuple(item[0]) for item in build_groups]
+    # Reorder each group to put the designated build leader first (string share)
+    final_groups = []
+    for group_tests, group_suites in build_groups:
+        leader_short = None
+        for suite in group_suites:
+            if suite in share_suite_leaders:
+                leader_short = share_suite_leaders[suite]
+                break
```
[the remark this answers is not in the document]
jgfouca: @copilot , why not just add sorted, `for suite in sorted(group_suites):`

CIME/get_tests.py
```diff
@@ -311,7 +369,34 @@ def get_build_groups(tests):
         if not matched:
             build_groups.append(([test], my_share_suites))
 
-    return [tuple(item[0]) for item in build_groups]
+    #
+    # Reorder each group to put the designated build leader first (string share)
+    #
+    final_groups = []
+    for group_tests, group_suites in build_groups:
+        # Due to transitivity, if there are multiple build leaders, we can
```
jgfouca, line 377: I'm not 100% sure about this design decsion.

CIME/get_tests.py
```diff
@@ -90,6 +90,16 @@
             "SMS_P16.f19_g16.X",
         ),
     },
+    "cime_test_share3": {
+        "time": "0:10:00",
+        "share": "SMS_P8.f45_g37.A",
```
rljacob, line 95: So the presence of a testname means that share is True.

rljacob (approved):

CIME/get_tests.py
jgfouca: Yep!

PR #5002, opened 2026-06-22
Fix get_ts_synopsis catch-all triggered by bless trailer after PASS

### LINKED ISSUE #4932: Still seeing "Could not interpret CPRNC output" in TestStatus.log
I looked at a random number of the 149 tests in our integration suite and they all have this at the bottom of the TestStatus.log file:

```
============ BASELINE COMPARE SYNOPSIS =============
master: ERROR Could not interpret CPRNC output
====================================================
```

Doesn't impact the pass/fail but what is supposed to go here and how do we make it happen?

jasonb5:
## Description

Fixes the `master: ERROR Could not interpret CPRNC output` message that has been appearing in `TestStatus.log` for passing baseline comparisons across entire integration suites.

### Root cause

`compare_baseline` (`CIME/hist_utils.py:592-598`) appends a bless trailer after the `PASS` token emitted by `_compare_hists`:

```
Comparing hists for case '...'
  ...matched...
PASS
  Most recent bless: <sha>
```

`get_ts_synopsis` only handled the case where `comments` ended exactly with `"PASS"`, so once any trailer was present the function fell through every recognized pattern and landed on the catch-all `ERROR Could not interpret CPRNC output`.

This became a near-universal symptom after #4952 (`Preserve TestStatus and bless_log`) added `BLESS_LOG_NAME` to the preserve list, so prior bless lines now persist across baseline regeneration and the trailer is appended on essentially every comparison.

### Fix

Replace the brittle `endswith("PASS")` check with a single multiline regex that recognizes the verbatim `PASS` token regardless of where it sits in the comments and tolerates CRLF line endings without weakening the exact-match guarantee:

```python
# Comparison passed if a line consisting of exactly "PASS" appears.
# _compare_hists writes this token verbatim; any other casing (Pass, pass,
# PASSING, ...) is not a status marker and must fall through. The optional
# \r tolerates CRLF line endings without weakening the exact-match check.
if re.search(r"^PASS\r?$", comments, re.MULTILINE):
    return ""
```

### Tests

- New doctest in `get_ts_synopsis` covering the realistic production string `'Comparing hists\nPASS\n  Most recent bless: sha:abc\n'`.
- Four unit tests in `CIME/tests/test_unit_hist_utils.py`:
  - `test_get_ts_synopsis_pass_at_end` — PASS at end-of-string.
  - `test_get_ts_synopsis_pass_on_own_line_with_multiple_lines_after` — PASS followed by a bless trailer.
  - `test_get_ts_synopsis_pass_is_case_sensitive_and_exact` — pins that variants like `Pass`, `pass`, `PASSING`, `passed`, `  PASS`, and `PASS extra` do NOT short-circuit.
  - `test_get_ts_synopsis_pass_handles_crlf_line_endings` — PASS on a CRLF-terminated line is recognized, both standalone and with a bless trailer.
- All hist_utils unit tests + doctests pass locally.

- Closes #4932

## Checklist

### DIFF
--- a/CIME/hist_utils.py
+++ b/CIME/hist_utils.py
@@ -786,13 +786,18 @@ def get_ts_synopsis(comments):
     'ERROR Could not interpret CPRNC output'
     >>> get_ts_synopsis('file1=\nfile2=\n  diff_test: the two files seem to be IDENTICAL \n')
     ''
+    >>> get_ts_synopsis('Comparing hists\nPASS\n  Most recent bless: sha:abc\n')
+    ''
     """
     comments = comments.strip()
 
     if comments == "" or "\n" not in comments:
         return comments
 
-    if comments.endswith("PASS"):
+    # Check if comments indicate a passing comparison
+    # Comments end with "PASS" directly, or have "PASS" on its own line
+    # (the latter happens when bless messages are appended after PASS)
+    if comments.endswith("PASS") or re.search(r"\nPASS\n", comments):
         return ""
 
     # Empty synopsis when files are identicial
--- a/CIME/hist_utils.py
+++ b/CIME/hist_utils.py
@@ -794,10 +794,11 @@ def get_ts_synopsis(comments):
     if comments == "" or "\n" not in comments:
         return comments
 
-    # Check if comments indicate a passing comparison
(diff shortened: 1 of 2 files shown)


jgfouca (approved):
Looks great! thanks

PR #5003, opened 2026-06-24
For batched builds, use the whole node

jgfouca:
GMAKE_J is generally used as a way to throttle builds so as to not overwhelm a login node. There is no such concern for a batched build, so just use the entire node.

Also, minor cleanup of batched build resource calculation in test_scheduler.

Also, add a check that the ninja exe is usable if the user asked for it.

Also, add support for communicating to sharedlib buildlib that ninja was requested.

## Checklist


jgfouca:
Sorry for the churn. This is ready for final review now.

CIME/build.py
```diff
@@ -812,6 +833,14 @@ def _build_libraries(
     # generate Makefile macro
     generate_makefile_macro(case, caseroot)
 
+    if ninja:
+        check_ninja(case)
+        ninja = False
```
billsacks: Is this correct? It looks like `check_ninja` returns a logical, so I would expect something like `ninja = check_ninja(case)`.

Either way, a comment would be helpful here. The current code is definitely unintuitive, even if it's correct: why do you set `ninja = False` if it's originally True? Even with my suggested change, a comment would help clarify what's going on... something like (if I'm understanding correctly): "User has requested a ninja-based build, but we need to check that this is possible; if not, we'll fall back to a non-ninja build"

CIME/build.py
```diff
@@ -812,6 +833,14 @@ def _build_libraries(
...
+    else:
+        # We cannot know if the various buildlib scripts support ninja.
+        # Just set an env var to let them know ninja was requested.
+        os.environ["CIME_SHAREDLIB_NINJA"] = "TRUE"
```
billsacks: This is definitely unintuitive, even if it's correct. It looks like this says: if `ninja` is false (so we're in the `else` block) then we're setting `CIME_SHAREDLIB_NINJA` to `TRUE`? That feels backwards.

My expectation is that this whole block would look like:

```python
if ninja:
    # Explanatory comment here...
    ninja = check_ninja(case)

# If ninja is still true after the check....
if ninja:
    # We cannot know if the various buildlib scripts support ninja.
    # Just set an env var to let them know ninja was requested.
    os.environ["CIME_SHAREDLIB_NINJA"] = "TRUE"
```

If I'm wrong and the current logic is correct, it warrants some comments explaining what right now feels like unintuitive behavior.

billsacks (changes requested):

CIME/build.py
jgfouca: Good catch. Not sure what I was thinking here.

billsacks (approved):
Looks good now - I like the way you have set up the logic.

PR #5004, opened 2026-06-25
Do NOT auto-pass first job

jgfouca:
## Description

Even though this is a small code change, this is a significant change to how CIME works. For reasons I don't understand, we had the first job in the workflow list skip the prereq check. I had no idea why we did this since it's easy to not provide a prereq if you don't want one. This design lead to surprises, like my case.build running at SUBMIT even though I had prereq set to False (since I never want to run it during SUBMIT).

This needs to be merged quickly so I can update E3SM's cime because a lot of stuff is broken there.

CESM folks may want to double-check their workflow XML to make sure this won't break you.

## Checklist

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -816,7 +816,6 @@ class EnvBatch(EnvBase):
 
         startindex = 0
         jobs = []
-        firstjob = job
         if job is not None:
             expect(job in alljobs, "Do not know about batch job {}".format(job))
             startindex = alljobs.index(job)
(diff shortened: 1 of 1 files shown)


rljacob (approved):

jgfouca:
@billsacks , can you take a quick look at this? It could potentially impact CESM.

billsacks:
@jgfouca - Do you have a sense of what, specifically, we should look for in our [workflow xml file](https://github.com/ESMCI/ccs_config_cesm/blob/8fe3339bd2b75c2090e06b054972bcd805c9d408/machines/config_workflow.xml) to make sure this won't break anything for us?

@mnlevy1981 and @fischer-ncar - also looping you two in, in the hopes that you might be more familiar with this workflow stuff than I am.

jgfouca:
@billsacks , the CESM file looks OK to me. I don't see a workflow that won't run when it should.

billsacks:
Thanks for looking & confirming, @jgfouca !

PR #5005, opened 2026-07-01
Much better support for ninja cmake backend

jgfouca:
## Description

Add much better support for different cmake backends, specifically ninja.

Change list:
1) Instead of hardcoding gmake invocations, use `cmake --build .` which will flexibly work for any backend
2) Add arguments for picking a specific backend to case.build and create_test
3) env_mach will automatically add ninja to PATH if it's there.
4) Instead of using `CIME_SHAREDLIB_NINJA` as a magic var, use the env that is known to CMake (`CMAKE_GENERATOR`).
5) Add machine setting / case env var `CMAKE_BACKEND`

## Checklist

### DIFF
--- a/CIME/test_scheduler.py
+++ b/CIME/test_scheduler.py
@@ -215,6 +215,8 @@ class TestScheduler(object):
         chksum=False,
         force_rebuild=False,
         no_batch_build=False,
+        ninja=False,
+        gmake=False,
         driver=None,
     ):
         ###########################################################################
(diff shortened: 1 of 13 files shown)


CIME/test_scheduler.py
```diff
@@ -310,6 +312,9 @@ def __init__(
 
         self._namelists_only = namelists_only
 
+        self._ninja = ninja
+        self._gmake = gmake
```
bartgol, line 316: Should we check that only one between gmake and ninja is true?

bartgol (approved):
Looks good to me. I would also consider renaming GMAKE_J to something like BUILD_J...

CIME/test_scheduler.py
jgfouca: Yes, that happens at the case.build level. I can add it here too.

fischer-ncar (approved):
Looks good.

PR #5006, opened 2026-07-06
Add wait_for_tests to the list of linked Tools

jgfouca:
It's pretty useful to be able to wait on something you just submitted.


jgfouca:
Now that I think about, `wait_for_tests` should probably go under `$case/Tools` instead of `$case`, especially since it won't work for `$case` if it's not a test.

billsacks (approved):
Cool, I didn't realize this was available as a standalone tool. This does seem useful. Thanks.

PR #5010, opened 2026-07-15
Selectors should work at the machine object level automatically

jgfouca:
## Description

Selectors should work at the machine object level automatically so that the user doesn't have to remember to pass attributes every time they call get_value.

This change makes me a bit nervous since it potentially impacts all CIME users in a non backwards-compatible way.

The problem was exposed by a machine config that looks like this:

```
    <COMPILERS>gnu,oneapi,gnugpu</COMPILERS>
    ...
    <MAX_TASKS_PER_NODE>1</MAX_TASKS_PER_NODE>
    <MAX_TASKS_PER_NODE compiler="gnu">32</MAX_TASKS_PER_NODE>
    <MAX_TASKS_PER_NODE compiler="gnugpu">2</MAX_TASKS_PER_NODE>
    <MAX_TASKS_PER_NODE compiler="oneapi">192</MAX_TASKS_PER_NODE>
```

This appears to work as intended when you have a case object, but does not work if you are dealing with a machine obj. There are many places where CIME must query machine properties before any case exists. One example of this is test_scheduler. The default pool size for create_test/test_scheduler is derived from MAX_TASKS_PER_NODE. Before this PR, the machine described above would return `1` even if the compiler was set to `gnu`. This is confusing behavior in my opinion. I could have added compiler attribute to all `machobj.get_value` calls in test_scheduler, but it seems like bad design for the developer to have to remember to pass attributes every time.

## Checklist

### DIFF
--- a/CIME/XML/machines.py
+++ b/CIME/XML/machines.py
@@ -380,13 +380,18 @@ class Machines(GenericXML):
     # pylint: disable=arguments-differ
     def get_value(self, name, resolved=True, subgroup=None):
         """
-        Get Value of fields in the config_machines.xml file
+        Get Value of fields in the config_machines.xml file. Note that most
+        config machines fields support compiler="..." and mpilib="..."
+        selectors. These are handled automatically but be sure to set_value
+        COMPILER or MPILIB correctly.
         """
         if self.machine_node is None:
             logger.debug("Machine object has no machine defined")
             return None
 
+        # Make sure no client is trying to pass attributes
         expect(type(resolved) is bool, "Wrong type for resolved")
+
         expect(subgroup is None, "This class does not support subgroups")
         value = None
 
--- a/CIME/XML/machines.py
+++ b/CIME/XML/machines.py
@@ -399,11 +399,25 @@ class Machines(GenericXML):
         elif name == "MPILIB":
             value = self.get_default_MPIlib(attributes)
         else:
-            node = self.get_optional_child(
-                name, root=self.machine_node, attributes=attributes
-            )
-            if node is not None:
-                value = self.text(node)
+            # Automatically add compiler and mpi selectors
+            if attributes is None:
+                attribute_list = [{"compiler" : self.get_value("COMPILER"), "mpilib" : self.get_value("MPILIB")},
+                                  {"compiler" : self.get_value("COMPILER")},
+                                  {"mpilib" : self.get_value("MPILIB")},
+                                  {}]
+                for attributes in attribute_list:
+                    node = self.get_optional_child(
+                        name, root=self.machine_node, attributes=attributes
+                    )
+                    if node is not None:
+                        value = self.text(node)
+                        break
+            else:
+                node = self.get_optional_child(
+                    name, root=self.machine_node, attributes=attributes
+                )
+                if node is not None:
+                    value = self.text(node)
 
         if resolved:
             if value is not None:
@@ -430,7 +444,7 @@ class Machines(GenericXML):
         )
         # if no match with attributes, try without
         if supported_values is None:
-            supported_values = self.get_value(listname, attributes=None)
+            supported_values = self.get_value(listname, attributes={})
 
         expect(
             supported_values is not None,
@@ -460,25 +474,30 @@ class Machines(GenericXML):
                 ),
             )
         else:
-            value = self.get_field_from_list("COMPILERS")
+            value = self.get_field_from_list("COMPILERS", attributes={})
         return value
 
     def get_default_MPIlib(self, attributes=None):
         """
         Get the MPILIB to use from the list of MPILIBS
         """
+        if attributes is None:
+            attributes = {"compiler": self.get_value("COMPILER")}
         return self.get_field_from_list("MPILIBS", attributes=attributes)
 
     def is_valid_compiler(self, compiler):
         """
         Check the compiler is valid for the current machine
         """
-        return self.get_field_from_list("COMPILERS", reqval=compiler) is not None
+        return self.get_field_from_list("COMPILERS", attributes={}, reqval=compiler) is not None
 
     def is_valid_MPIlib(self, mpilib, attributes=None):
         """
         Check the MPILIB is valid for the current machine
         """
+        if attributes is None:
+            attributes = {"compiler": self.get_value("COMPILER")}
+
         return (
             mpilib == "mpi-serial"
             or self.get_field_from_list("MPILIBS", reqval=mpilib, attributes=attributes)
--- a/CIME/XML/machines.py
+++ b/CIME/XML/machines.py
@@ -401,10 +401,15 @@ class Machines(GenericXML):
         else:
             # Automatically add compiler and mpi selectors
             if attributes is None:
-                attribute_list = [{"compiler" : self.get_value("COMPILER"), "mpilib" : self.get_value("MPILIB")},
-                                  {"compiler" : self.get_value("COMPILER")},
-                                  {"mpilib" : self.get_value("MPILIB")},
-                                  {}]
+                attribute_list = [
+                    {
+                        "compiler": self.get_value("COMPILER"),
+                        "mpilib": self.get_value("MPILIB"),
+                    },
+                    {"compiler": self.get_value("COMPILER")},
+                    {"mpilib": self.get_value("MPILIB")},
+                    {},
+                ]
                 for attributes in attribute_list:
                     node = self.get_optional_child(
                         name, root=self.machine_node, attributes=attributes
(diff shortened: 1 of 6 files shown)


CIME/XML/machines.py
```diff
@@ -431,7 +453,7 @@ def get_field_from_list(self, listname, reqval=None, attributes=None):
         )
         # if no match with attributes, try without
         if supported_values is None:
-            supported_values = self.get_value(listname, attributes=None)
+            supported_values = self.get_value(listname, attributes={})
```
bartgol: Here (and below) you are _intentionally_ passing an empty dict, to prevent it from populating it with COMPILER/MPILIB, right?

bartgol (approved):
I lack the true expertise to know the ramifications. However, the changes seem to make sense to me...

billsacks (changes requested):
### General thoughts

I like the change here: I agree that it feels really error-prone to require callers to remember to pass exactly the right attributes. Thank you for fixing this!!

It seems like this PR fixes some cases of this, but now I'm nervous about other, similar cases. Is there any way for get_value to see if there are any attribute selectors in the xml for the given thing you're trying to get and give an error if those attributes aren't in the attributes argument? Or are there cases where you'd want to leave out some attributes intentionally? (In those cases, could you require explicitly saying that you don't care about those attributes somehow? Basically, force the caller of get_value to specify all attributes needed for this variable or explicitly say that they're not specifying a given attribute.) I have no idea how much this would involve, and I'm fine with this idea being deferred to later or rejected, I just wanted to raise this question.

> This change makes me a bit nervous since it potentially impacts all CIME users in a non backwards-compatible way.

Is the non-backwards-compatibility just in cases that are probably currently buggy? Or are there other possible behavior changes in ways where things are working right currently but will somehow break with this PR? Oh, maybe answering my own question: I guess this could break in cases where there is a query done where you explicitly *don't* want the compiler or mpilib attribute to be part of the match? Are you able to give a pattern that we could look for in code outside of CIME that might break due to this PR, and then I could raise this possible issue with CESM people.

### Possible specific change

Currently it looks like compiler and mpilib are added to the attributes only if attributes=None. A more intuitive behavior for me would be if the compiler and mpilib were always added to the attributes. I'm thinking: what if you need to match something that has some other attribute, `foo`. So the developer adds `attributes={"foo": "whatever"}`. Now suddenly you no longer add compiler and mpilib. That feels counter-intuitive to me, and could cause problems if that match also needs compiler and mpilib. For example, imagine that there was already a call to get_value that didn't specify any attributes (so implicitly added compiler and mpilib), and now you just added a "foo" selector to that variable. The intuitive thing would be to just add "foo" to the (currently non-existent) list of attributes, but that would lead to incorrect behavior; instead you need to know that, since you're adding an attribute, you also need to explicitly add compiler and mpilib.

It seems like one problem with my proposal is that you sometimes want to tell it NOT to add compiler and mpilib, and you currently do that by giving an empty attribute list. That mechanism wouldn't work with my proposal. But maybe this can be handled with an explicit argument like `no_attributes` (which defaults to False but could be set to True) (there is probably a better name for this): if set, this would skip adding the compiler and mpilib attributes.

### And a specific request that I'd like

Whether you stick with the current implementation or take my above suggestion, I'd like to see better documentation of the changed routines that automatically add attributes. These should document the attributes that are added and the conditions under which they are / aren't added. (e.g., that the current behavior is to add compiler / mpilib attributes if attributes is left as None, but *not* to add any attributes if attributes is given... and so you can avoid having compiler / mpilib attributes by specifying `attributes={}`. The challenge of documenting this clearly is partly what led me to my above suggested behavior change.)

CIME/XML/machines.py
jgfouca: Yes

jgfouca:
@billsacks , the first thing to check is if CESM is using selectors for fields like MAX_TASKS_PER_NODE. Selectors are commonly used for environment stuff, and that should work the same as before. I suspect you guys are not using selectors for other fields. We weren't really doing that either until @bartgol started enhancing some of the CI machine configurations.

I agree that the current implementation is confusing. Let me think about it a bit.

jasonb5:
Automatically injecting the compiler/mpilib doesn't seem very clean, could have side effects if the developer doesn't know to pass `{}` to get the value with no selector.

Maybe instead of basing it off of the `attributes` value we add an explicit argument that would trigger injecting these values. If they're already provided in attributes we can just fill in what is missing. This way the developer has a clearer path on controlling the functionality and the code reads better.

jgfouca:
Upon further review, I decided to just remove support for passing custom attributes to machine queries. It's just too confusing. compiler and mpilib will be the supported selectors and they will be automatically added to queries. It is up to the developer to set COMPILER and MPILIB if they want their queries to reflect non-default settings.

CIME/case/case.py
```diff
@@ -1456,24 +1456,24 @@ def configure(
...
         for name in (
             "MAX_TASKS_PER_NODE",
             "MAX_MPITASKS_PER_NODE",
             "MAX_CPUTASKS_PER_GPU_NODE",
             "MAX_GPUS_PER_NODE",
         ):
-            dmax = machobj.get_value(name, {"compiler": compiler})
-            if not dmax:
-                dmax = machobj.get_value(name)
+            dmax = machobj.get_value(name)
```
jgfouca, line 1476: You can see here that very hacky code is eliminated by the design adjustment in this PR.

billsacks:
> the first thing to check is if CESM is using selectors for fields like MAX_TASKS_PER_NODE

I did the following git grep in [CESM's ccs_config](https://github.com/esmci/ccs_config_cesm/) to look for any selectors on any fields that start with a capital letter (there are a lot of selectors on fields that are lowercase, like `batch_system`, `queue`, `mpirun`, etc., but my sense is that they might not be problematic... but I'm not sure??)

```
$ git grep '^ *<[A-Z].*='
machines/cheyenne/config_machines.xml:    <MPILIBS compiler="intel" >mpt,openmpi</MPILIBS>
machines/cheyenne/config_machines.xml:    <MPILIBS compiler="pgi" >openmpi,mpt</MPILIBS>
machines/cheyenne/config_machines.xml:    <MPILIBS compiler="nvhpc" >openmpi,mpt</MPILIBS>
machines/cheyenne/config_machines.xml:    <MPILIBS compiler="gnu" >mpt,openmpi</MPILIBS>
```

This is on a no-longer-supported machine, and compiler seems like an acceptable attribute, so based on this check I think this is okay?? -- though I'm not clear on whether it was safe to ignore all lowercase variables in this check.

CIME/XML/machines.py
```diff
@@ -379,32 +378,55 @@ def set_machine(self, machine, schema=None):
...
-    def get_value(self, name, attributes=None, resolved=True, subgroup=None):
+    def get_value(self, name, resolved=True, subgroup=None):
         """
         Get Value of fields in the config_machines.xml file
         """
         if self.machine_node is None:
             logger.debug("Machine object has no machine defined")
             return None
 
+        expect(type(resolved) is bool, "Wrong type for resolved")
```
billsacks, line 393: Since (I think) we don't normally have assertions like this on the type of arguments, this line probably warrants a comment - that this is here to catch problems with old code that may have passed the now-removed attributes argument as a positional argument that would now inappropriately land in the resolved argument. (There's probably a better way to say that.)

billsacks (changes requested):
I like this implementation better - thanks!

I have one request for a comment.

In addition, it feels like maybe this new behavior should be documented somewhere? (i.e., that COMPILER and MPILIB are automatically considered as selectors.) Maybe in the get_value docstring? I'm not sure where someone would be most likely to look to see this, so if you think that documenting this wouldn't have much value, I'm okay letting that go.

jgfouca:
@jasonb5 , any idea why I'm getting so many file not found errors in the CI?

jgfouca:
@billsacks , done.

jasonb5:
@jgfouca They're all mpi build errors, so not related to your changes.
```
/__w/cime/cime/e3sm/externals/scorpio/src/clib/core/rearr/pio_rearr_contig.cpp:391:17: error:  MPI_Type_hvector  was not declared in this scope; did you mean  MPI_Type_vector ?
  391 |           ret = MPI_Type_hvector(nvars, 1, stride_between_vars, agg_gs_info_.rtypes[i], &agg_rtype);
      |                 ^~~~~~~~~~~~~~~~
      |                 MPI_Type_vector
```
I'll need to dive deeper, maybe we're using too old of an mpi version.

billsacks (approved):
Thanks!

jasonb5 (approved):
LGTM

PR #5011, opened 2026-07-16
Test

jgfouca:
No-op PR just to confirm e3sm CI is broken.

- Closes #<ISSUE_NUMBER_HERE>

## Checklist

### DIFF
--- a/CIME/build.py
+++ b/CIME/build.py
@@ -2,6 +2,7 @@
 functions for building CIME models
 """
 
+
 import glob, shutil, time, threading, subprocess
 from pathlib import Path
 from CIME.XML.standard_module_setup import *


jasonb5:
@jgfouca Right now it looks like scorpio added some code that uses `MPI_Type_hvector`  which was removed in >openmpi 4.x.x. There is another reference to `MPI_Type_hvector` in scorpio but its only used when building with mpiserial which we don't touch. The other thing that could have happened was openmpi going from 3->4 in the container, but that seems unlikely since its already at 4.1.6.

jgfouca:
@jayeshkrishna , can we make scorpio compatible with recent openmpi versions?

PR #5019, opened 2026-07-28
feat(flux): convert walltime to minutes/FSD for batch_system="flux"

### LINKED ISSUE #5017: feat(flux): convert walltime to minutes/FSD for batch_system="flux"
### Summary
Flux's `-t/--time-limit` accepts **minutes or Flux Standard Duration (FSD, RFC 23)** — not `HH:MM:SS`. Tuolumne is currently the only E3SM machine running Flux, so this is handled as a **Flux-specific code path** in `env_batch.py`, leaving the generic `format_time()`/`walltime_format` mechanism untouched (used as-is by other Slurm/PBS-style machines).

### Reference: Flux's accepted format
Per `flux-submit(1)`:
> `-t, --time-limit=MINUTES|FSD`
> Set a time limit for the job in either minutes or Flux standard duration (RFC 23). FSD is a floating point number with a single character units suffix (`s`, `m`, `h`, or `d`). The default unit is minutes when no unit is specified.

### Root cause recap
`CIME/utils.py::format_time()` re-arranges `H`/`M`/`S` fields positionally; it does not sum across units. E.g. `walltime_format="%M"` on `01:10:00` yields `10`, not `70`. This isn't a bug in `format_time()` itself — it's simply the wrong tool for Flux's minutes/FSD requirement, and other machines rely on its current field-rearranging behavior, so it should not be changed generically.

### Proposed fix
In `CIME/XML/env_batch.py::get_job_overrides()` (~line 484), add a Flux-specific branch that converts the resolved walltime directly, rather than reusing the generic `format_time()`/`walltime_format` path for this scheduler:

```python
if batch_system == "flux":
    seconds = convert_to_seconds(walltime)
    # convert to minutes and/or FSD as appropriate
    walltime = <flux-appropriate conversion>
else:
    walltime_format = self.get_value("walltime_format")
    if walltime_format:
        ...  # existing format_time() path, unchanged
```

Implementation should support **both** Flux-accepted formats:
- **Minutes** (plain integer, e.g. `70`)
- **FSD** (floating point + unit suffix `s`/`m`/`h`/`d`, e.g. `70m`, `1.17h`)

Exact choice of which format to emit by default (and any rounding behavior for partial minutes) is an implementation detail to work out in the PR.

### Out of scope
- Changing the generic `format_time()` field-rearranging behavior used by other (non-flux) machines.
- Rounding/ceiling-division correctness improvements to the general `walltime_format` path.

### Validation
- Unit tests covering `batch_system="flux"` walltime conversion for both minutes and FSD output paths, e.g. `01:10:00` → `70` (minutes) and equivalent FSD form.
- Confirm non-flux batch systems are unaffected (existing `format_time()`/`walltime_format` behavior unchanged).
- End-to-end: submit a Tuolumne test case with `env_batch.xml` walltime `01:10:00`; confirm the generated `flux batch`/`flux submit -t` value is valid and the enforced job time limit matches.

### References
- [Project](https://flux-framework.org/)
- [GitHub](https://github.com/flux-framework)
- [Docs](https://flux-framework.readthedocs.io/en/latest/)

jasonb5:
## Summary
- Adds a flux-specific walltime conversion in `EnvBatch.set_job_defaults()`: Flux's `-t/--time-limit` accepts minutes or Flux Standard Duration (FSD, RFC 23), not `HH:MM:SS`
- On the flux path, `%H`/`%M`/`%S`(/`%D`) in `walltime_format` expand to the **total** duration in that unit, so a matching FSD suffix yields a valid value (e.g. `%Mm` on `01:10:00` → `70m`, `%Hh` → `1.17h`); no format defaults to FSD minutes (`70m`)
- Accepted formats are whitelisted (`%M`, `%Ss`, `%Mm`, `%Hh`, `%Dd`); anything else (e.g. `%H` — a bare number flux would misread as minutes — or `%H:%M:%S`) fails loudly at case setup instead of submitting a wrong time limit
- Partial values round up at two decimal places so a job is never allotted less time than requested
- The generic `format_time()`/`walltime_format` path for all other batch systems is untouched
- Existing deployed flux config (`<walltime_format>%Mm</walltime_format>` on Tuolumne) works unchanged

## Validation
- New parametrized unit tests for `_flux_walltime` covering all accepted formats, rounding edge cases, and rejected formats, plus mock-based `set_job_defaults` tests for the flux branch (`01:10:00` → `70m`/`70`/`1.17h`, `00:10:30` → `11m`)
- All existing `test_unit_xml_env_batch.py` tests pass, confirming non-flux behavior is unchanged
- Docs build cleanly; new "Walltime Format" section documents both the positional and flux semantics
- End-to-end validation on Tuolumne still to be confirmed by a maintainer with access

Closes #5017

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -482,7 +482,9 @@ class EnvBatch(EnvBase):
                         )
 
             walltime_format = self.get_value("walltime_format")
-            if walltime_format:
+            if self._batchtype == "flux":
+                walltime = self._flux_walltime(walltime, walltime_format)
+            elif walltime_format:
                 seconds = convert_to_seconds(walltime)
                 full_bab_time = convert_to_babylonian_time(seconds)
                 walltime = format_time(walltime_format, "%H:%M:%S", full_bab_time)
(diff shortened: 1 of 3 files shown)


jasonb5:
Needs testing on actual machine still.

PR #5020, opened 2026-07-28
feat(flux): support omit_in_job attribute on submit args to drop args invalid inside a job

### LINKED ISSUE #5018: feat(flux): support omit_in_job attribute on submit_args to drop args invalid when submitting from inside a job
### Summary
On Tuolumne, `-p <partition>` is required for the **initial** `flux batch`/`flux submit` call (issued from outside any Flux instance), but must be **dropped** when a submission is issued **from inside an already-running job** (a true in-job resubmit), since Flux gives each job its own ephemeral nested cluster instance with no partitions defined. Resubmitting with a stale `-p <partition>` from within a running job fails because that partition doesn't exist in the new instance.

### Current CIME behavior
- `CIME/XML/env_batch.py::get_submit_args()` / `_resolve_argument()` (~lines 621-770) resolves `<submit_args>` from `config_batch.xml` identically regardless of submission context.
- The resubmit path (`submit_jobs`, `resubmit_immediate`/`submit_resubmits`, ~lines 855-998) doesn't currently distinguish "submitted from outside a job" vs. "submitted from within a job's own resubmit step."

### Proposed design
1. **New `<arg>` attribute: `omit_in_job`** (boolean, default `false`) — add to the `<arg>` schema in `config_batch.xsd`, settable on individual submit args in `config_batch.xml`. When `omit_in_job="true"`, the arg is dropped from the resolved submit command if the submission is happening from inside an already-running batch job.
2. **In-job detection via `batch_env`:** add support for checking common, well-known batch-system environment variables that indicate an active job (e.g. `SLURM_JOB_ID`, `PBS_JOBID`, `FLUX_JOB_ID`) as part of/alongside the existing `batch_env` handling in `env_batch.py`. Since batch schedulers change infrequently, this scheduler→env-var mapping can be a small **static/hardcoded list in CIME** rather than a fully config-driven mapping — no need for per-machine config to declare which env var to check.
3. In `get_submit_args()`/`_resolve_argument()`, use this in-job check to decide whether to skip any arg marked `omit_in_job="true"`.
4. Apply `omit_in_job="true"` to the `-p` arg in Tuolumne's flux `config_batch.xml` (model-config change, once schema/mechanism support lands in CIME).

### Scope of affected args
`-p` is the only confirmed case so far. Other submit args scoped to the outer/initial allocation may also need `omit_in_job="true"`, but this hasn't been tested yet on Tuolumne — the mechanism should be generic/attribute-driven (not hardcoded to `-p`) so additional args can opt in later without further code changes.

### Validation
- Unit test in `CIME/tests/test_unit_xml_env_batch.py`: resolve submit args in both "in-job" (env var present, e.g. `FLUX_JOB_ID` set) and "not-in-job" (env var absent) contexts against a flux-like config with an `omit_in_job="true"` arg; assert it's included only in the not-in-job case.
- End-to-end on Tuolumne with `RESUBMIT>=1`: confirm the initial `flux batch` includes `-p`, and the in-job resubmit call omits it.
- **Follow-up investigation (fast-follow, not blocking initial fix):** test on Tuolumne to identify whether other submit args besides `-p` need `omit_in_job="true"`.

### Open questions
- Confirm the static list of env vars to check per batch system (e.g. `SLURM_JOB_ID`, `PBS_JOBID`, `LSB_JOBID`, `FLUX_JOB_ID`) covers all currently supported CIME batch systems.
- Confirm exact integration point within `batch_env` handling vs. a small standalone helper reusing similar logic.

### References
- [Project](https://flux-framework.org/)
- [GitHub](https://github.com/flux-framework)
- [Docs](https://flux-framework.readthedocs.io/en/latest/)

jasonb5:
## Summary
- Adds an `omit_in_job` boolean attribute to `<arg>` and `<argument>` submit args in `config_batch.xsd`; args marked `omit_in_job="true"` are dropped when submission happens from inside an already-running batch job.
- Adds `EnvBatch.is_in_batch_job()`, detecting in-job context via a small static scheduler→env-var mapping (`FLUX_JOB_ID`, `SLURM_JOB_ID`, `PBS_JOBID`, `LSB_JOBID`, `COBALT_JOBID`). Only the current batch system's env var is considered.
- Motivation: on Tuolumne, flux gives each job an ephemeral nested instance with no partitions, so a stale `-p <partition>` fails on in-job resubmit. The mechanism is generic/attribute-driven so other args can opt in without code changes.

## Testing
- New pytest unit tests in `CIME/tests/test_unit_xml_env_batch.py` covering in-job, not-in-job, other-scheduler env var present (ignored), and unknown batch system cases; full file passes (29 tests).

## Follow-up
- E3SM model config: add `omit_in_job="true"` to `-p` for Tuolumne's flux `config_batch.xml`.
- End-to-end validation on Tuolumne with `RESUBMIT>=1`.

Closes #5018

### DIFF
--- a/CIME/XML/env_batch.py
+++ b/CIME/XML/env_batch.py
@@ -24,6 +24,22 @@ from itertools import zip_longest
 
 logger = logging.getLogger(__name__)
 
+# Static mapping of batch system type to the well-known environment
+# variable that indicates the current process is running inside an
+# active job for that scheduler. Batch schedulers change infrequently
+# so this is maintained in code rather than in per-machine config.
+IN_JOB_ENVIRONMENT_VARIABLES = {
+    "flux": "FLUX_JOB_ID",
+    "lsf": "LSB_JOBID",
+    "pbs": "PBS_JOBID",
+    "pbspro": "PBS_JOBID",
+    "moab": "PBS_JOBID",
+    "slurm": "SLURM_JOB_ID",
+    "slurm_single_node": "SLURM_JOB_ID",
+    "cobalt": "COBALT_JOBID",
+    "cobalt_theta": "COBALT_JOBID",
+}
+
 # pragma pylint: disable=attribute-defined-outside-init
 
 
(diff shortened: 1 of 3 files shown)


jasonb5:
Needs testing on actual machine still.

Issue #5023, opened 2026-08-06
Case postrun fails when there is a single component (is_comp_standalone == TRUE) for multi-instance cases


ekluzek:
This started in:

https://github.com/ESCOMP/CDEPS/issues/426

where I ran a single component case with DGLC with two instances. It marks the case as failing because it doesn't find a single instance glc.log file as they are split into glc_0001.log and glc_0002.log.

The fail looks like this:

```
ERROR: Model did not complete, no /glade/derecho/scratch/erik/cases/SMS_C2_D_Lh12.f10_f10_mg37.2000_SATM_SLND_SICE_SOCN_SROF_DGLC%NOEVOLVE_SWAV.derecho_intel.clm-pauseResume.GC.ctsm545delist/run/glc.log.7026886.desched1.260805-205554 log file
```

ekluzek:
This is because of this bit of logic in case_run.py doesn't recognize multi-instance when comp_standalone==TRUE.

``` python
    # find the last model.log and cpl.log
    if comp_standalone:
        model_logfile = os.path.join(rundir, file_prefix + ".log." + lid)
        if not os.path.isfile(model_logfile):
            expect(
                False, "Model did not complete, no {} log file ".format(model_logfile)
            )
        elif os.stat(model_logfile).st_size == 0:
            expect(False, "Run FAILED")
```

billsacks:
I agree that this seems like a bug. I think there are actually multiple bugs here:

(1) When CIME detects that this is a "standalone" single-component run, the multi-instance logic isn't used.

(2) The logic for `is_comp_standalone` seems to capture situations that should not be captured.

Ideally, both of these issues should be fixed. But resolving just one of these should be sufficient to fix the issue @ekluzek ran into. I actually think (2) is the more important one to fix.

Some relevant old PRs:

https://github.com/ESMCI/cime/pull/4917

https://github.com/esmci/cime/pull/4580

@rljacob raised the point in #4580 that the logic for "standalone" component was breaking tests in E3SM, and I think the same comment applies to CESM: even if (single instance) tests haven't been failing for us, I think this logic is wrong. I'm not sure, though, what the logic is supposed to be. Specifically, what is special about a case like FKESSLER that leads it to not have a mediator log file? How can we distinguish that situation from the situation where we have a single component but things otherwise work normally, with a mediator log file?

ekluzek:
[...]

PR #5031, opened 2026-08-17
fix incorrect finalization of aborted CAM standalone runs (bad)

adamrher:
## Description

CAM standalone compsets (all stub components except CAM) are treated as a successful completion even when the model aborts (e.g., tar's all the log files). These compsets include FKESSLER and FHS94. Included a fix here that nests logic inside the  ```if comp_standalone``` block that reads the ```model_logfile``` and checks for ```TERMINATION_TEXT```. 

- Closes #<ISSUE_NUMBER_HERE>

## Checklist


adamrher:
@ekluzek I don't think this closes your issue https://github.com/ESMCI/cime/issues/5023, but it does address an issue with CAM standalone finalization in case_run.py.

adamrher:
This PR isn't robust. While aborted runs don't finalize, it seems that now successfully completed runs aren't being finalized. I'll debug this shortly.

billsacks:
Thanks for this contribution @adamrher . If I understand correctly, it is not yet working right in all cases. I'm converting it to draft as a reminder that this isn't really ready for review yet.

If you get this to a point where it flags both successful and unsuccessful standalone runs correctly, it will be great to have that. And since I *think* the only scenarios impacted by that (at least following a change I plan to make) are F compsets in CESM (since the logic in this routine only sets standalone to true for nuopc, so this doesn't impact e3sm, and I *think* the only CESM standalone configurations are F compsets), if it works robustly for these standalone F compsets I'm happy to accept it.

billsacks:
@adamrher - I noticed that the history of this branch is weird, including an unrelated commit from many years ago. I normally don't like pushing a rebase to an open PR, but in this case I think it's warranted to clean this up. Let me know if you want help with this (at least, if you plan to move ahead with this PR).

adamrher:
> @adamrher - I noticed that the history of this branch is weird, including an unrelated commit from many years ago. I normally don't like pushing a rebase to an open PR, but in this case I think it's warranted to clean this up. Let me know if you want help with this (at least, if you plan to move ahead with this PR).

Hmm ... It's an old CIME fork that I merged up, but there must have been some non-pushed changes or something else going on. Is there an easy way to clean this up? Otherwise I'm thinking I'll just delete my fork and get a fresh new CIME fork and then re-create the PR / PR-branch.

I now have a fix for CAM standalone that works for both aborted and completed cases.

billsacks:
You shouldn't have to delete your entire fork, but you can. You could keep the same fork and just start a new branch off of latest ESMCI master (or something close to it).

adamrher:
closing PR due to odd commit history. Will re-open with this cleaned up.

PR #5032, opened 2026-08-19
For fake system tests, only do filesystem and IO on rank 0

jgfouca:
## Description

I tried running some fake tests with _P$N where N > 1 and there were issues due to multiple procs trying to copy files etc.

## Checklist
- [x] My code follows the style guidelines of this project (black formatting)
- [x] I have performed a self-review of my own code
- [x] My changes generate no new warnings
- [ ] I have added tests that exercise my feature/fix and existing tests continue to pass
- [x] I have commented my code, particularly in hard-to-understand areas
- [ ] I have made corresponding additions and changes to the documentation

PR #5033, opened 2026-08-20
fix incorrect finalization of aborted cam_standalone runs

adamrher:
## Description
CAM standalone compsets (all stub components except CAM) are treated as a successful completion even when the model aborts (e.g., tar's all the log files). These compsets include FKESSLER and FHS94. Included a fix here that nests logic inside the  ```if comp_standalone``` block that reads the ```model_logfile``` and checks for ```TERMINATION_TEXT```.

- Addresses part of #5023 

## Checklist

### DIFF
--- a/CIME/case/case_run.py
+++ b/CIME/case/case_run.py
@@ -343,6 +343,16 @@ def _post_run_check(case, lid):
             )
         elif os.stat(model_logfile).st_size == 0:
             expect(False, "Run FAILED")
+        else:
+            # The existence/size checks above pass even when the model aborts
+            # mid-run, so also require a termination message in the component
+            # log. CAM writes "END OF MODEL RUN" on successful completion.
+            with open(model_logfile, "r") as fd:
+                logfile = fd.read()
+            expect(
+                any([x in logfile for x in TERMINATION_TEXT]),
+                "Model did not complete - see {} \n ".format(model_logfile),
+            )
     else:
         count_ok = 0
         for cpl_logfile in cpl_logs:
--- a/CIME/case/case_run.py
+++ b/CIME/case/case_run.py
@@ -346,7 +346,8 @@ def _post_run_check(case, lid):
         else:
             # The existence/size checks above pass even when the model aborts
             # mid-run, so also require a termination message in the component
-            # log. CAM writes "END OF MODEL RUN" on successful completion.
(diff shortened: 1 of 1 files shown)


CIME/case/case_run.py
```diff
@@ -343,6 +343,16 @@ def _post_run_check(case, lid):
             )
         elif os.stat(model_logfile).st_size == 0:
             expect(False, "Run FAILED")
+        else:
+            # The existence/size checks above pass even when the model aborts
+            # mid-run, so also require a termination message in the component
+            # log. CAM writes "END OF MODEL RUN" on successful completion.
```
billsacks:
```suggestion
            # log. This comp_standalone logic is currently designed for CAM,
            # which writes "END OF MODEL RUN" on successful completion.
```

billsacks (changes requested):
Looks good to me, though I have a minor suggestion for an improvement to the comment. Thank you for submitting this fix!

billsacks:
Thanks a lot, @adamrher ! I just want to confirm: is this one working in both the pass and fail cases now?

I think the CI failure is unrelated (it also failed in https://github.com/ESMCI/cime/pull/5032).

adamrher:
> Thanks a lot, @adamrher ! I just want to confirm: is this one working in both the pass and fail cases now?

Yes, I just ran two tests showing it working for both cases.

billsacks (approved):

billsacks:
@jasonb5 - I'm adding you as reviewer in case you'd like to look at this before we merge it. I'm pretty sure the block of code that's changed doesn't ever apply for E3SM, since comp_standalone is only true here for the nuopc driver, but I still want to give you a chance to look at it before we merge it if you want.

The new logic isn't completely general, but it works in the cases where it matters for us, so I'm okay with it - and it is an improvement in practice from what was there before, which wasn't working robustly.

billsacks:
@jasonb5 gave me the go-ahead to merge.

billsacks:
It turns out that bringing in these changes caused aux_cdeps tests to fail. I realized that I'll need to bring in other fixes for #5023 on top of these changes so that tests can continue to pass. For now I'm backing out the changes in this PR, but I'll plan to reintroduce them next week - see #5062

PR #5035, opened 2026-08-20
Make jenkins_generic_job glob more strict by adding mach_comp

jgfouca:
Also, transition jenkins and wait_for_tests to modern py string formatting.

We hit a very unlikely edge case where two PM jobs kicked off near the same time and ended up picking the same timestamp, which messed a lot of things up. Making the glob have the mach_comp in addition to the timestamp will prevent this from happening again.

## Checklist

### DIFF
--- a/CIME/jenkins_generic_job.py
+++ b/CIME/jenkins_generic_job.py
@@ -38,9 +38,7 @@ def delete_old_test_data(
     ###############################################################################
     # Remove old dirs
     for clutter_area in [scratch_root, test_root, run_area, build_area, archive_area]:
-        for old_file in glob.glob(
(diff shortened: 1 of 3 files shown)


CIME/jenkins_generic_job.py
```diff
@@ -415,15 +383,15 @@ def jenkins_generic_job(
 
     os.environ["CIME_MACHINE"] = machine.get_machine_name()
 
+    mach_comp = f"{machine.get_machine_name()}_{compiler}"
```
jgfouca, line 386: These are the only changes that weren't just string format updating.

jasonb5 (changes requested):
I think we just need some tests, then I'd be fine approving.

PR #5044, opened 2026-08-29
For cesm add coupling_mode to the list of attributes that can be used by component config_component.xml files

### LINKED ISSUE #5043: For CESM add coupling_mode to the list of attributes that can be used in components config_component.xml files
The list of attributes that can be used in components config_component.xml files is: compset, grid, and cime_model. For CESM we'd like to add coupling_mode from the CMEPS variable COUPLING_MODE so it can be used in component models config_component.xml files. Specifically we need it for DATM in CDEPS. COUPLING_MODE is currently how NorESM is distinquished from CESM.

ekluzek:
## Description

For cesm add coupling_mode to the list of attributes that can be used by components config_component.xml. 

Note, this happens before shell_commands is called, so you can't change coupling_mode for a case with a user or test-mod shell_commands file.

- Closes #5043

## Checklist


CIME/case/case.py
```diff
@@ -1123,6 +1123,20 @@ def _get_component_config_data(self, files):
         if comp_root_dir is not None:
             self.set_value(root_dir_node_name, comp_root_dir)
 
+        #
+        # For CESM add coupling_mode to the list of attributes
+        # This has to be done after the driver env variables where done above
```
billsacks: Should "where done" be "were done"?

billsacks:
I've been thinking about this. Your point that you can't change this in testmods raises what I think is a broader concern, at least in my mind: I feel like it's problematic to have the defaults for xml variables depend on other xml variables. I've had the mental model: Compset sets default values of xml variables; xml settings set default values of namelist variables. The change in this PR breaks this by having some xml defaults depend on other xml defaults. Besides messing with the mental model, this also seems problematic in that it introduces an order dependence to the parsing, and raises problematic questions about what should happen if a user changes the value of COUPLING_MODE after possibly making other changes to other xml variables whose default depends on it. At the very least, I think this would require moving COUPLING_MODE to env_case.xml, but then we'd probably want a way to set it at create_newcase time, e.g., via a new flag. This is all feeling very messy.

Therefore, I think we should *not* bring in this change. The differentiation that you're trying to allow with this seems like it would best be done in some other way. Some ways that come to mind are: (1) differentiating on cime_model (I know that's the long-term plan); (2) adding some compset-level differentiators.

mvertens:
@billsacks - I totally agree with the points you raise and don't think we should bring in this change as well. I think at this point if we want to have xml variables be noresm dependent we need to have noresm be parallel to cesm in cime.

ekluzek:
OK, we decided to not do this approach. And have something possible in CDEPS. So closing.

PR #5050, opened 2026-09-07
Fix for ESMF-aware threading ompthread option for threaded CESM runs

### LINKED ISSUE #5049: NUOPC's ESMF-aware threading gives incorrect resource line to job system (when using cores as threads)
When running CESM with ESMF_AWARE_THREADING=true, the 'ompthreads' line would specify the number of threads desired, vs letting ESMF's managed threading handle them.  This results in ESMF PETs being assigned strided CPU cores (by the number of threads specified), and no ability for it to 'map' those skipped CPU cores into its threading logic.

It might be this was desired behavior with 'hyperthreads' (eg, map to a single core, but two hyperthreads on it), but is not desired behavior for full core-based threading.

briandobbins:
## Description
These changes fix how threads are allocated in a job when using NUOPC and ESMF_AWARE_THREADING=true.  A test is included too.

This alters that behavior only when ESMF_AWARE_THREADING=true, so that both non-ESMF threading and ESMF threading behave correctly.

I don't generally contribute to CIME much, so I welcome feedback on the process so I can do better next time!  If it looks good, I'll update the documentation, even though in practice we don't expect folks to run with threading much -- for now, it's primarily needed for a CMIP run with complex chemistry.

- Closes #5049

## Checklist
- [X ] I have performed a self-review of my own code
- [X] I have added tests that exercise my feature/fix and existing tests continue to pass


billsacks (approved):
Thanks for this fix @briandobbins ! This seems subtle but important!

billsacks:
@briandobbins - I noticed that there is handling of ESMF_AWARE_THREADING in env_mach_pes.py. I am very unfamiliar with this code, so I ask this naively: Did you consider whether this could be done via a change / fix to the ESMF_AWARE_THREADING logic there rather than via this override? I really have no idea if that's possible, but at a glance it looks like there is some intent to handle ESMF_AWARE_THREADING in env_mach_pes.py....

briandobbins:
This is a really good point, Bill -- honestly, being unfamiliar with the code, I hadn't really looked too deep.  But I bet it can be moved there, and it's arguably more 'clean' since it keeps all that logic in one place.   I can restructure this tomorrow, I imagine.

billsacks (approved):
I can't say I understand this very well, but it looks reasonable. Thanks for getting this working @briandobbins !

rljacob:
No impact on E3SM. go ahead.

PR #5054, opened 2026-09-13
Feature/merge to cime6.5.12

mvertens:
## Description
Brings the noresm branch up to date with upstream CIME tag cime6.5.12.

This is a straight merge with no NorESM changes of its own — it is a prerequisite for the upcoming PR to make noresm it's own model in CIME.  Once that PR is in - we can make a PR back to ESMCI with these changes and ESCOMP can then bring it in as well.

The only conflict was the cprnc submodule, which git cannot merge on its own. Resolved by taking the upstream commit (8aa01c2), which has our previous commit (025932e) as an ancestor, so nothing is lost.

Testing: 
Running prealpha tests on olivia using noresm3_0_beta24 and comparing to baselines.


mvertens:
Sorry - wrong PR.

PR #5057, opened 2026-09-15
add new noresm cime model

mvertens:
## Description
NorESM and CESM share the same git-fleximod layout, so a NorESM checkout was previously detected as `cesm`. This PR adds `noresm` as a model in its own right. This PR should have no impact on E3SM CIME functionality.

- new file `CIME/data/config/noresm/config_files.xml` — this new file is based on the cesm file, with  `MODEL` set to `noresm`, components NorESM does not use removed.

- `get_model()` now reads a new file -  `$SRCROOT/.cime_model_id` ***if it is present***. 
  - If the string inside this file is noresm - then the noresm model is used. 
  - **This takes precedence over `CIME_MODEL`; a conflicting value is warned about.** 
  - Note that the accepted values for the string are the names of the files in `cime/CIME/data/config `(currently cesm, e3sm, noresm and ufs)
- a new variable `CESM_LIKE_MODELS` collects the places where CIME branches on the model name   and noresm must behave as cesm: shared library list, unsupported compset check, describe_version tag lookup, case cloning, build provenance.
- **Backwards compatible.** Nothing is modified without the presence of the `.cime_model_id` file, which lives in the NorESM repo. 

A few other changes:
- git@ (ssh) ==> https: for the cprnc submodule (universal access to CIME)
- Fix typo in Makefile (CESM/NorESM specific)
- shebang fix in find_circular_dependencies.py (necessary in modern environments, e.g., conda) (CESM/NorESM specific)
- Taking out the "mira" special case (I don't think this machine exists)

Testing: verified that this change is completel backwards compatible with noresm checkouts
  
## Checklist
- [x] I have tested that exercise my feature/fix and existing tests continue to pass


CIME/utils.py
```diff
@@ -518,7 +576,28 @@ def get_model():
     >>> get_model()
     'e3sm'
     >>> reset_cime_config()
+    >>> del os.environ["SRCROOT"]
```
billsacks: It seems like, instead of deleting the SRCROOT environment variable, this doctest should save any pre-existing SRCROOT environment variable and then, if it existed previously, reset it to the original rather than deleting it.

billsacks (changes requested):
This is great! I really like this approach - both the CESM_LIKE_MODELS and the check for a model id file at the top level. Thanks a lot for doing this, @mvertens - this is going to really help with coordination of CMEPS, CDEPS and other repositories between NorESM and CESM!

I have one request for a fix to the doctest. (Sorry that I missed this in my review on NorESMhub.) Other than that, this looks great.

billsacks:
@jasonb5 and @jgfouca - There are a couple of design points here that I discussed with @mvertens and am satisfied with myself, but would be happy for your input on:

(1) The new `$SRCROOT/.cime_model_id` file takes precedence over everything else. This felt like the right approach to me, but can you see any problems with this?

(2) Do we still have tests that use a standalone cime, outside of any model? If so, can you see any problems with this check for `.cime_model_id` for standalone cime usages? I couldn't see any, but wanted to run this by you.

CIME/utils.py
mvertens: @billsacks - thanks for catching this. I have fixed this problem.

jgfouca (approved):

jgfouca:
@billsacks , I'm ok with `.cime_model_id`.

billsacks (approved):
Looks good now - thank you!

mvertens:
@jgfouca @billsacks - thanks for your reviews. Is it okay to merge this? Who normally does this?

